Issue #100, opened 2018-07-11
Directives with host: {'[class]'} in the declaration overwrites any user-defined classes


mikerodonnell89:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal
Directives do not allow the user to set any custom classes because host: {'[class]'} overwrites them.  See Deno's solution for https://github.com/SAP/fundamental-ngx/issues/93

#### What versions of Angular and Fundamental NGX? (If this is a feature request, use current version)
Angular 6.0.0, fundamental-ngx 0.0.6-beta

#### If this is a bug, please provide steps for reproducing it
`<div class="some-class" fd-navbar-group [position]="'left'">`
"some-class" is not applied to the div after rendering

PR #107, opened 2018-07-12
fix: update directives that use host classes so that they can accept …

mikerodonnell89:
…custom classes from the user on the directive element

#### Please provide a link to the associated issue
#100 

#### Please provide a brief summary of this pull request
Using "host: {'[class]'}" in the directive declaration overwrites any class the user sets on the directive's HTML element.  In this pull request we add the necessary styles to the component one-by-one without overwriting any classes the user has added on the element.  Each directive that requires this functionality extends an abstract class that handles the construction of the element and will reset and repopulate the classList when any changes are made to the directive's Inputs.

### DIFF
--- a/projects/fundamental-ngx/src/lib/utils/custom-class-base-component.ts
+++ b/projects/fundamental-ngx/src/lib/utils/custom-class-base-component.ts
@@ -1,0 +1,33 @@
+import { ElementRef, OnChanges, OnInit, Input } from '@angular/core';
+
+export abstract class CustomClassBaseComponent implements OnInit, OnChanges {
+    private _elementRef: ElementRef;
+
+    @Input() class; // user's custom classes
+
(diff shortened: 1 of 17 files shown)


projects/fundamental-ngx/src/lib/utils/custom-class-base-component.ts
```diff
@@ -0,0 +1,33 @@
+import { ElementRef, OnChanges, OnInit, Input } from '@angular/core';
+
+export abstract class CustomClassBaseComponent implements OnInit, OnChanges {
+    private _elementRef: ElementRef;
+
+    @Input() class; // user's custom classes
+
+    abstract _setProperties(): void;
```
SAPMontreal: maybe we should document the purpose of that function and the one after for future

Issue #103, opened 2018-07-12
implement component time


droshev:
#### Is this a bug, enhancement, or feature request?
feature request

#### Briefly describe your proposal
implement [time](https://sap.github.io/fundamental/components/time.html) component released in fundamental-ui v. 1.1

#### What versions of Angular and Fundamental NGX? (If this is a feature request, use current version)
Angular 6.0.0, fundamental-ngx 0.0.6-beta

PR #166, opened 2018-08-01
time component implementation

mikerodonnell89:
#103


droshev:
Maybe we shouldn't advertise the disabled validation example
screen shot 2018-08-08 at 9 24 32 am
I would just document it as part of the input type

droshev:
I think when one of the fields is not valid then swapping pm/am should not manipulate that field. It should stay invalid
[2018-08-08_09-28-00.mp4.zip](https://github.com/SAP/fundamental-ngx/files/2270403/2018-08-08_09-28-00.mp4.zip)

droshev:
I still see the last issue:
when the hour has a number that is not valid and change pm/am the number gets updated. I think once a field has invalid value we should not change it

droshev:
I don't see value of showing disabled validation showcase. even the opposite it looks weird case

Issue #208, opened 2018-08-23
Add rowSpan input to fd-panel


mikerodonnell89:
#### Is this a bug, enhancement, or feature request?

Enhancement

#### Briefly describe your proposal.

Panels support row and column span, currently the panels are only accepting column span and the user must manually set the row span class on the component.  Row span should be an input (rowSpan) and column span input should be renamed to colSpan

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)

Angular 6.0.0, fundamental-ngx 0.0.16-beta

mikerodonnell89:
https://github.com/SAP/fundamental/issues/630

droshev:
@mikerodonnell89 is this still valid?

PR #219, opened 2018-09-18
Panel row col span

mikerodonnell89:

https://github.com/SAP/fundamental-ngx/issues/208

Panels should support row and column spans.  Documentation was also incorrect for panels (col is a property of fd-panel-grid, not fd-panel)


projects/fundamental-ngx/src/lib/panel/panel.component.ts
```diff
@@ -6,14 +6,19 @@ import { AbstractFdNgxClass } from '../utils/abstract-fd-ngx-class';
...
+
+    @Input() rowSpan: number;
 
     @Input() backgroundImage: string;
 
     _setProperties() {
         this._addClassToElement('fd-panel');
-        if (this.span) {
-            this._addClassToElement('fd-has-grid-column-span-' + this.span);
+        if (this.columnSpan) {
```
droshev: can we have unit test for those changes?

PR #243, opened 2018-09-25
input group should implement ControlValueAccessor

mikerodonnell89:

Input Group should implement ControlValueAccessor, enabling the use of ngModel, form state/validity, etc


projects/fundamental-ngx/src/lib/input-group/input-group-number.component.spec.ts
```diff
@@ -26,10 +26,10 @@ describe('InputGroupNumberComponent', () => {
 
     it('should get the input and increment/decrement properly', () => {
         component.inputText = 0;
-        expect(component.getInput()).toBe(0);
+        expect(component.inputText).toBe(0);
```
droshev: can we add a test that an event is being fired on change?

Issue #286, opened 2018-11-01
*Fiori-Fundamentals v.1.3.1*  Button 


InnaAtanasova:
#### Is this a bug, enhancement, or feature request?
code refactoring
#### Briefly describe your proposal.
The Button component needs code refactoring: 
- Button Options - emphasized, regular, light
- Types - actions, standard, positive, medium, negative
- Sizes - default and compact
- RTL support
- Improve documentation

PR #298, opened 2018-11-08
Enhancement/286

mikerodonnell89:

#286

### DIFF
--- a/docs/modules/documentation/containers/button/button-docs.component.html
+++ b/docs/modules/documentation/containers/button/button-docs.component.html
@@ -8,40 +8,53 @@
 
 <properties [properties]="{
   inputs: [
+    { name: 'options', description: '(Optional) String - Button Options.  Options include \'emphasized\' and \'light\'.  Leave empty for default.'},
     { name: 'type', description: '(Optional) String - The type of the button.  Options include \'primary\', \'main\', \'secondary\', and \'toolbar\'. Leave empty for default.'},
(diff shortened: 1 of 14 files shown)


docs/modules/documentation/containers/button/button-docs.component.html
```diff
@@ -8,40 +8,53 @@ <h1>&lt;button fd-button&gt;</h1>
 
 <properties [properties]="{
   inputs: [
+    { name: 'options', description: '(Optional) String - Button Options.  Options include \'emphasized\' and \'light\'.  Leave empty for default.'},
     { name: 'type', description: '(Optional) String - The type of the button.  Options include \'primary\', \'main\', \'secondary\', and \'toolbar\'. Leave empty for default.'},
-    { name: 'semantic', description: '(Optional) String - Semantic buttons. Semantic modifiers include \'positive\' and \'negative\'. Leave empty for no modifier.'},
-    { name: 'size', description: '(Optional) String - The size of the button.  Options include \'xs\', \'s\', \'compact\', default, and \'l\'. Leave empty for default.'},
+    { name: 'semantic', description: '(Optional) String - Semantic buttons. Semantic modifiers include \'standard\', \'positive\', \'medium\', and \'negative\'. Leave empty for no modifier.'},
+    { name: 'size', description: '(Optional) String - The size of the button. Leave empty for default, or specify \'compact\'.'},
```
InnaAtanasova: Since only 2 options are available, what do you think of renaming the property to 'compact' and change the type to boolean? I think it would be less confusing. We can even go further and make it a directive???

mikerodonnell89: Good idea.  I think it should be a boolean input property rather than a separate directive

Issue #332, opened 2018-11-27
Calendar - globalisation - first day of the week in UI


GauSim:
#### Is this a bug, enhancement, or feature request?

enhancement, or feature request

#### Briefly describe your proposal.

add support for a `first day of the week`-setting for render. 
depending on with locale the user is in the first day of the week might be different.
Example: 
in Germany the first day of a week is considered a Monday. 
in US the first day of a week is considered a Sunday.

this should be reflected in the calendar layout. 

note: i like how momentJs deals with this - here you can set a locale and the lib sets values like first day of the week, or first week of the year (used to display week numbers in calendars) etc. 

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
n/a

PR #334, opened 2018-11-27
add option for calendar weeks to start on Monday instead of Sunday

mikerodonnell89:

#332

### DIFF
--- a/library/src/lib/calendar/calendar.component.ts
+++ b/library/src/lib/calendar/calendar.component.ts
@@ -61,6 +61,9 @@ export class CalendarComponent implements OnInit, OnChanges, AfterViewChecked {
         return false;
     };
 
+    @Input()
+    mondayStartOfWeek: boolean = false;
+
     @Output()
     updateDatePickerInput: EventEmitter<any> = new EventEmitter();
     @Output()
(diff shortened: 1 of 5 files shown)


library/src/lib/calendar/calendar.component.ts
```diff
@@ -100,7 +103,7 @@ export class CalendarComponent implements OnInit, OnChanges, AfterViewChecked {
         'November',
         'December'
     ];
-    weekDays: string[] = ['S', 'M', 'T', 'W', 'T', 'F', 'S'];
+    weekDays: string[];
     daysPerMonth = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
```
droshev: what if February is 29 days

library/src/lib/calendar/calendar.component.ts
```diff
@@ -694,6 +702,7 @@ export class CalendarComponent implements OnInit, OnChanges, AfterViewChecked {
             this.constructCalendar();
             this.constructCalendarYearsList();
         }
+        this.weekDays = this.mondayStartOfWeek ? ['M', 'T', 'W', 'T', 'F', 'S', 'S'] : ['S', 'M', 'T', 'W', 'T', 'F', 'S'];
```
droshev: do we need to make this translatable?

library/src/lib/calendar/calendar.component.ts
InnaAtanasova: there is a function `determineDaysInMonth` that determines if Feb has 28 or 29 days

library/src/lib/calendar/calendar.component.ts
mikerodonnell89: I think so, however I think i18n should be a separate story

Issue #404, opened 2019-01-11
docs: hovering over components is clunky


MattL75:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
- [x] Hovering over a component's name in the sidebar and then exiting its box causes the 'selected' teal bar to leave, which brings the component with it and causes a clunky/bouncy effect.
- [x] As part of this issue, I would suggest adding an 'active' teal bar, which is persists besides the component after it is clicked to indicate that the page is open.

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
Angular 7.1
Fundamental-ngx 0.0.26-beta

#### If this is a bug, please provide steps for reproducing it.
Hover over a component name in the sidebar, then hover over a different component.

#### Is there anything else we should know?
Can probably be fixed by adding a negative margin when the teal bar disappears, or by using a box-shadow instead of a border.

Also present in the React docs.

PR #407, opened 2019-01-11
Documentation/#404 smoother navigation

MattL75:
#404 

Added a router-link to the <a> tags in the HTML. This activates the 'active' router-link so the CSS is applied to the active element.

Changed the teal side border to an inner box-shadow, the effect is a much smoother navigation and no bouncy effect. I kept the pulling out of the text on hover since I think it looks nice. Could change if needed.

### DIFF
--- a/docs/modules/documentation/components/documentation/documentation.component.scss
+++ b/docs/modules/documentation/components/documentation/documentation.component.scss
@@ -63,7 +63,8 @@
 .nav-item:focus,
 .nav-item--active {
     color: #424e5a;
-    border-left: 6px solid #2fd0c8;
+    //border-left: 6px solid #2fd0c8;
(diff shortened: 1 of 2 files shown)


docs/modules/documentation/components/documentation/documentation.component.scss
```diff
@@ -63,7 +63,8 @@
 .nav-item:focus,
 .nav-item--active {
     color: #424e5a;
-    border-left: 6px solid #2fd0c8;
+    //border-left: 6px solid #2fd0c8;
```
droshev: you can remove it completely

MattL75: Oops

Issue #421, opened 2019-01-16
Form Input component enhanhacement


fkolar:
#### Briefly describe your proposal.

Input components in general here does not support template driven (TDF) or reactive driven forms (RDF).  Even 2-way binding for value could work for simple cases but this is not angular way.

When using them in angular form and trying to levarange  `ngModel` or `formControl`  I can not. 

This applies for any inputs based component starting from input fields, dropdown, date pickers, etc..

fkolar:
Toggle is just the simple state YES/NO, True/false soprobably not much. 
I know I would probably add the onChange event but that's about it.

MattL75:
@mikerodonnell89 Can we consider this done?

MattL75:
Closing this. Most (if not all) components that qualify as form components are now compatible with Angular forms.

PR #428, opened 2019-01-18
toggle handles ngModel

mikerodonnell89:

#421 

Add support for ngModel on toggle

### DIFF
--- a/library/src/lib/toggle/toggle.component.ts
+++ b/library/src/lib/toggle/toggle.component.ts
@@ -30,11 +30,11 @@ export class ToggleComponent implements OnInit, ControlValueAccessor {
     onChange: any = () => {};
     onTouched: any = () => {};
 
-    get inputText() {
+    get isChecked() {
         return this.checked;
     }
 
-    set inputText(value) {
+    set isChecked(value) {
         this.checked = value;
         this.onChange(value);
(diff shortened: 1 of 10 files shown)


library/src/lib/toggle/toggle.component.ts
```diff
@@ -15,13 +23,33 @@ export class ToggleComponent implements OnInit {
...
-    @Input()
-    id: string;
+    onChange: any = () => {};
+    onTouched: any = () => {};
+
+    get inputText() {
+        return this.checked;
+    }
+
+    set inputText(value) {
```
MattL75: Maybe a bit confusing to have get and set InputText on a toggle?

fkolar:
I think you dont really need this part

```
if (this.size && this.size !== 'xs' && this.size !== 's' && this.size !== 'l') {
            this.size = null;
        }
```

if you do something like:
```
size: InputSize = 'small';
export type InputSize = 'large' | 'normal' | 'small';
```

fkolar:
But I guess all this should be generic thing common for all comments.

library/src/lib/toggle/toggle.component.ts
mikerodonnell89: fixed

Issue #422, opened 2019-01-16
List component


fkolar:
#### Briefly describe your proposal.
Let's thing how to bring list component into enterprise level. 
 - Usually list needs to handle large number of items,  
 - They need to support single and multiselection. 
 - Developers needs to be able to customize the way item is rendered `ng-template `
 - Virtual scrolling and lazy loading. 

We dont need to reinvent all this @angular/cdk its good patern library .

MattL75:
- [x] Usually list needs to handle large number of items
- [x] They need to support single and multiselection
- [x] Developers needs to be able to customize the way item is rendered ng-template (already doable)
- [ ] Virtual scrolling and lazy loading.

Considering that virtual scrolling has its own issue thread, I will close this

PR #440, opened 2019-01-23
Enhancement/#422 added single selection lists

MattL75:

Part of #422 

The code isn't exactly pretty, but it serves as an automated single selection list. It could also be implemented by the user, so I'm not sure what direction we want to go in. Doesn't hurt to implement it ourselves though, saves the user some time.

@mikerodonnell89 If you have an idea of how to accomplish this besides my method pls let me know 😶 

yes


mikerodonnell89:
Failing unit test, `Can't bind to 'ngModel' since it isn't a known property of 'input'`, probably means the test needs to import `FormsModule`

mikerodonnell89:
Also I think, if we are going to have single-select lists, the checkbox should actually be a radio button, thoughts @fkolar @xak ?

fkolar:
> Failing unit test, `Can't bind to 'ngModel' since it isn't a known property of 'input'`, probably means the test needs to import `FormsModule`

It just means that for ngModel you haven't imported FormModule

fkolar:
Yes if you turn on single select mode have it backed up by radio is the easiest way to track what is selected. Even you probably will not show the actual radio button

MattL75:
Yeah we're not sure if the radio button from fundamental would work here

fkolar:
I think any radio group would work. I think radio buttons will not be visible user will see only a list no? The specs for Controls is not finalized yet so not sure what is the final state.

MattL75:
Ideally we would want to give some visual feedback to the user about what is selected or not with the radio buttons themselves

fkolar:
I think this is good inspiration in terms how the list should be customizable. 
https://sapui5.hana.ondemand.com/#/entity/sap.m.StandardListItem

with icons, description, titles, etc..

MattL75:
That can all be done with our current list, it's very flexible though requires you to code some of it. I know you are an advocate of stricter api so that might not be what you're looking for though

Issue #439, opened 2019-01-23
Table should support selectable elements


MattL75:
#### Is this a bug, enhancement, or feature request?
Feature

#### Briefly describe your proposal.
I believe we should rework our implementation of table. What we do currently gives the users a lot of freedom when it comes to the actual cells, but at the cost of not being able to natively provide a selectable feature, sorting... etc. Users would need to implement it themselves with this setup. This is needed by an internal team.

Issue will be updated as we find new requirements.

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
0.0.29-beta
Angular 7.1

PR #446, opened 2019-01-24
Docs/#439 tables with checkboxes

MattL75:
#439 

Added documentation on using the table with checkboxes, including a check-all feature. @mikerodonnell89 do you think our API should eventually provide something similar natively? It's not exactly straightforward to implement and would for sure be blocking to some people, if only because of the ViewChildren. Thoughts?

### DIFF
--- a/docs/modules/documentation/containers/table/examples/table-checkboxes-example.component.html
+++ b/docs/modules/documentation/containers/table/examples/table-checkboxes-example.component.html
@@ -10,7 +10,7 @@
     <fd-table-body>
         <tr *ngFor="let row of tableRows; let i = index">
             <td>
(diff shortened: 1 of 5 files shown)


mikerodonnell89:
I think when we define our own table API we could handle the checking logic on our end.  Also could the checked property just be a property of each table row?  Like this: https://github.com/SAP/fundamental-ngx/tree/docs/%23439-table-checkbox-no-viewchild

MattL75:
Yeah that's actually simpler way probably, if they're okay with having the checked property in their rows, which I'm not sure why they would be against it.

Some stuff like auto-deselect of the check-all when something is deselected doesn't work though so I'll rework it and push some commits in a sec

Issue #458, opened 2019-01-25
Search-Input option for glyph


jakeleesap:
Can you please support the option to use another icon in search-input for glyph ?

thanks

droshev:
@jakeleesap what is your scenario or use case?

jakeleesap:
Hi,
Base on our UX design, this field is default as a  drop down with a default list, until the user type in any value, it will do a round trip to search the value. So, since the original drop down has already contains a list, it should allow us to default the icon as a Arrow-Down so the user would first click on it. Hope my explanation clear :(

MattL75:
I'm not sure I completely understand what it is you're trying to do, but if the baseline requirement is control over the icon in search-input, I'm sure we can accommodate. 

We can still have the search icon as default, so adding the option to change it will only make the component more flexible.

jakeleesap:
Thanks for the response, maybe I explain once more time that is more simple.

Lets say I have many other field that I used "arrow-down" from glyph - and if I want to be consistent for  the same field like "arrow-down" instead of the default "search" icon, it would nice to have the same icon.

thanks

MattL75:
Hi @jakeleesap control over the search input icon will be available in the next release which will hopefully be soon! :)

Issue #456, opened 2019-01-25
Support more match options with search-input component


cstechfoodie:
#### Feature request details
Currently, the search-input component only has one pipe function to match the prefix of string with input. However, this is not alway the case we want. We also need to have a pipe function to match the substring of a string with input, no matter what index the substring starts at. This helps much with filtering ids whose prefix are the same.

#### Below are source code

https://github.com/SAP/fundamental-ngx/blob/develop/library/src/lib/search-input/search-input.component.ts

https://github.com/SAP/fundamental-ngx/blob/develop/library/src/lib/search-input/search-input.component.html

#### Is there anything else we should know?
It is nice to provide multiple pipe options and let user to chose which pipe function to use.

cstechfoodie:
And maybe an option to disable the DropdownValues to be added

PR #466, opened 2019-01-29
Enhancement/#456 & #458 search input revamp

MattL75:
#456 #458 

Added a glyph input to search input to change the icon at will. @mikerodonnell89 I think there's an issue with the Fundamental aspect though. It seems like with any icon other than search, you can't close the popup with the button after you open it. You can still close it with other methods.

Made search-input a bit more flexible.

- Allow custom filters, added a docs example with this
- Made the highlighting apply to the searchTerm no matter where in the result it is
- Made highlighting optional
- other minor fixes

There should be no breaking changes here. The old search input should work the same as it always did.

Yes

**Custom Filter**
[Custom Filter](https://gyazo.com/69245f8eca55f3ea340c8dffa2ce594c)

**Standard No Highlighting**
[Image from Gyazo](https://gyazo.com/afeaf2d297ab1f9d0ee26e13970c24ec)


mikerodonnell89:
The issue with the search not closing when providing a custom glyph was actually an issue with our popover code, just pushed a fix to this branch

MattL75:
@mikerodonnell89 ah cool makes sense

Issue #454, opened 2019-01-25
Alert should open as an overlay


mikerodonnell89:
Currently the alert doesn't have any positioning, it should be opened as an overlay like on this demo page: https://sap.github.io/fundamental/demo-pages/alert-overlay-demo-page.html

This will be similar to a modal but without the darkening overlay and without blocking interaction with the rest of the page.

PR #468, opened 2019-01-30
refactor alert to use overlay and service

mikerodonnell89:

#454


library/src/lib/alert/alert.component.ts
```diff
@@ -42,7 +51,45 @@ export class AlertComponent implements OnInit {
...
+            this.alertService.popAlert();
+        }
+    }
+
+    open() {
+        // check to make sure this alert is not already opened
+        if (this.elRef.nativeElement.style.display !== 'block') {
+            const top = this.getTop();
+            this.elRef.nativeElement.style.display = 'block';
+            const alertDiv = <HTMLElement>this.elRef.nativeElement.querySelector('.fd-alert');
```
MattL75: Are we OK with using querySelector? I always try to avoid it in favour of ViewChild/ViewChildren if I can.

mikerodonnell89: Good question - in this specific example, we're querying for the div that contains the classname `'.fd-alert'`, which is actually a child element of the `<fd-alert>` component.  ViewChild would be useful if we were trying to select a child `<fd-alert>`component, but, `this` is the alert component itself, so ViewChild wouldn't work here

Issue #470, opened 2019-01-31
Readme not updated on npm


mikerodonnell89:
Update the library build script in package.json to copy over all necessary readme documentation and assets

PR #471, opened 2019-01-31
update readme for npm

mikerodonnell89:
#470


MattL75:
So there are two readmes 😅

Issue #464, opened 2019-01-29
Structural directive on <fd-tab> casues ExpressionChangedAfterItHasBeenCheckedError


schaefermichael:
#### Is this a bug, enhancement, or feature request?
Bug

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
Angular 7, fundamental-ngx 0.0.30

#### If this is a bug, please provide steps for reproducing it.
- Create the first fd-tab in a fd-tab-list using a structural directive
- In Dev mode, the following error will occur:
`
ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'aria-expanded: null'. Current value: 'aria-expanded: true'.
`

#### Please provide relevant source code if applicable.
See https://stackblitz.com/edit/angular-nhewct

PR #472, opened 2019-01-31
fix expressionchanged error when using ngIf on tabs

mikerodonnell89:
#464 

Feels hacky but check the official docs: https://angular.io/guide/component-interaction#parent-calls-an-viewchild


MattL75:
Ugh. I do not like that but I get it

PR #485, opened 2019-02-05
remove playground for alert

mikerodonnell89:
Alert playground broken with #454, each option is demonstrated earlier so I think we should just remove it


MattL75:
would it be useful to also have an inline alert that we could then show in the playground? Just curious

mikerodonnell89:
We would need a separate component with its own CSS rules and it would have to work outside of the alert service so that the vertical stacking works correctly for the demo alerts.  I think that'd be overkill when we can demonstrate each modal state in the examples

Issue #482, opened 2019-02-05
Implement file input system


MattL75:
#### Is this a bug, enhancement, or feature request?
Feature

#### Briefly describe your proposal.

TODO
- [x] Multi-input
- [x] Allow only a set of extensions
- [x] Drag and drop files
- [x] Return the files via event
- [x] public api for everything
- [ ] tests for the component
- [x] docs basics
- [x] docs examples (~~basic~~, ~~dropzone with custom styling~~, ~~without ngModel + concat/remove files as they come in~~, disabled dragndrop, only accept certain extensions + dealing with invalid files, disabled input, clear files)
- [x] Utilities + Layout section for documentation sidenav

https://experience.sap.com/fiori-design-web/upload-collection/#behavior-and-interaction

Needed by an internal team

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
Current

PR #492, opened 2019-02-07
Feature/#482 file input

MattL75:
#482 

Pretty big PR with quite a few things.

Added layout and utils section to docs, fixed a AfterChecked error datepicker/calendar bug that had slipped through the pr on openSelectDays.

Yes


mikerodonnell89:
How about size limit as an input that would add a file that is too large to the invalid list?  Or should that be handled by the dev or maybe server side?

MattL75:
Hm. I hadn't considered that. I'm pretty sure there's no way to 100% foolproof stop with a size limit, but there's probably a way to stop most cases, which leaves the server to handle the edges. I'll take a look

PR #507, opened 2019-02-11
Master merge conflicts

mikerodonnell89:
This fixes merge conflicts between develop and master.  After this is merged, another pull request from develop to master should be opened and then we should not see any more develop/master conflicts


droshev:
@mikerodonnell89 can you bring master into develop after?

PR #523, opened 2019-02-14
feat: add height/width inputs to modal

### LINKED ISSUE #522: Modal various improvements
Enhancement

- Users should be able to pass the width/height of the modal manually if they so choose. Currently the only way to do that is to use /deep/.
- Maybe as part of that ticket we can offer full screen mode too?
- allow having modals without header so applications can have free style modal

MattL75:
fixes #522 

Allows more control over modal sizing for users.

### DIFF
--- a/docs/modules/documentation/containers/modal/examples/modal-example.component.ts
+++ b/docs/modules/documentation/containers/modal/examples/modal-example.component.ts
@@ -4,7 +4,7 @@ import { ModalService } from '../../../../../../library/src/lib/modal/modal.serv
 @Component({
     selector: 'fd-modal-example',
     template: `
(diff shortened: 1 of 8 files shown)


droshev:
@MattL75 I added 2 minor improvements to #522. Would you mind doing them as well in the same PR?

MattL75:
Full screen should already work by just using % or vw/vh units. 

Also pretty sure modals without a header is already possible.

droshev:
@MattL75 This is great! Would you mind adding examples in the playground? It would be more obvious for the developers. Also, how about unit testing the changes?

MattL75:
Done

PR #528, opened 2019-02-15
fix: dynamically generated modals should destroy on close

### LINKED ISSUE #524: Modal components are not being destroyed
Bug

Angular 7.1.1, fundamental-ngx 0.0.32 (Also occurs in the current docs)

https://sap.github.io/fundamental-ngx/modal -> Component as content example
When opening a modal, it is appended to the DOM. But after closing, the modal service does not destroy it.
In our specific case, it caused some weird behavior in a reactive form.

https://sap.github.io/fundamental-ngx/modal -> Component as content example

MattL75:
fixes #524 

Added a Subject to the modal component, which is subscribed to when the modal is first opened. When the modal is closed, it triggers the subscription and removes the dom element from the body.

I need thoughts on this solution. @mikerodonnell89 for next week when you're done skiing!

### DIFF
--- a/library/src/lib/modal/modal.component.ts
+++ b/library/src/lib/modal/modal.component.ts
@@ -6,6 +6,7 @@ import {
     OnInit
 } from '@angular/core';
 import { ModalService } from './modal.service';
+import { Subject } from 'rxjs';
 
 @Component({
     selector: 'fd-modal',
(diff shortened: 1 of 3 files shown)


MattL75:
Not sure how I would go about testing this this since it stems from a modal/service interaction. I'll ask Mike on Monday what he thinks. Maybe add a modal instance to the test would work.

PR #547, opened 2019-02-19
chore: add dependency

jbadan:

angular-cli-ghpages was missing from the dev dependencies, needed to publish docs site from Travis

### DIFF
--- a/ci-scripts/publish.sh
+++ b/ci-scripts/publish.sh
@@ -31,5 +31,7 @@ npm publish
 
 cd ../..
 
+npm install
+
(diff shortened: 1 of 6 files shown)


ci-scripts/publish.sh
```diff
@@ -31,5 +31,7 @@ npm publish
 
 cd ../..
 
+npm install
+
```
greg-a-smith: I think this can be left out — I don’t think this was an issue with `npm install` getting called, it was that it was never defined as a dependency.

PR #562, opened 2019-02-21
fix: bug/#529 remove closed alert dom

### LINKED ISSUE #529: alert doesn't work properly since v 0.0.31
bug

After v 0.0.31 the alert elements after dismiss stay in the dom

angular v 7.3
fundamental-ngx 0.0.31 and later

mikerodonnell89:
fixes #529 

Remove component as content alerts from the DOM when they are closed.  Still need to figure out how to do this for template reference alerts and modals...


MattL75:
I don't think we even should do that for template reference modals & alerts. They have a place in the dom as defined by the user, so they should be removed from the dom by the user with an ngIf. If needed, we could apply the animation on leave/enter to make it look the same. Let me know what you think

PR #565, opened 2019-02-22
multi-input better object support

MattL75:
Allows users to provide a displayFn function, which then handles the display. With this method, users can display the text however they like. Pretty flexible compared to the previous [displayWith] property that only allowed 1-level deep object support.


MattL75:
Removed the 'fix:' prefix since this should not appear in the release notes. Instead it will be under the 'multi-input' feature point umbrella.

PR #578, opened 2019-02-25
fix: scroll spy threshold is customizable

MattL75:
Enhances the scroll spy to support a percentage scroll detection threshold.


library/src/lib/scroll-spy/scroll-spy.directive.ts
```diff
@@ -11,32 +11,35 @@ export class ScrollSpyDirective {
     @Input()
     public fireEmpty: boolean = false;
 
+    @Input()
+    public targetPercent: number = 0;
+
     @Output()
-    public topSpyChange: EventEmitter<string> = new EventEmitter<string>();
+    public spyChange: EventEmitter<any> = new EventEmitter<any>();
```
Algatron: It looks like this EventEmitter<`any`> and `currentActive` and `spiedTag` and `element` should all be typed as HTMLElement?

MattL75: Good point

PR #588, opened 2019-02-26
feat: calendar keyboard improvements and calendar tests

### LINKED ISSUE #535: Upgrade Datepicker/Calendar keyboard navigation
Enhancement

Current

mikerodonnell89:
fixes #535


droshev:
At the moment the datepicker opens up automatically when the focus ends up on the input field. I think it would be more logical the following scenario:

1. focus goes on the button of the datepicker
2. the user click space and then the pickers shows

Let's see what the others think

mikerodonnell89:
@droshev made those changes

library/src/lib/calendar/calendar.component.ts
```diff
@@ -157,6 +158,9 @@ export class CalendarComponent implements OnInit, OnDestroy, AfterViewChecked, C
         selectedLastDay: this.selectedRangeLast
     };
 
+    @Output()
+    closeCalendar = new EventEmitter<any>();
```
Algatron: should any be undefined here?

droshev:
@mikerodonnell89 now the focus traps works but in datetime picker it doesn't go to the time part

library/src/lib/calendar/calendar.component.ts
mikerodonnell89: I think in this example, users may want to configure their own functionality when the calendar closes so we should leave them the option to pass an event along

Algatron: Sorry, I am still adjusting to angular, but isn't it the owner of the event emitter (this component) the only person that defines the type of information emitted?

droshev: maybe we defined as `new EventEmitter<void>()` in case we don't pass anything

Algatron: I'm just pretty confused apparently? :) I think the one who **creates** the event emitter does `myEmitter.emit(X)`
And the type of X defines the generics T of EventEmitter<T>

The **user** of the component will hook onto the event with `<cmp (event)=Y`
and here in the Y the user can choose to consume the value (if any) that was emitted, or do whatever they want when there is something emitted, but AFAIK they can't control when/how/what is emitted?

Algatron: @droshev only reason I suggested undefined is that is the default value 
https://angular.io/api/core/EventEmitter#emit

ghiobi: Hello,

Let's take a step back. I think, what Mr. `Algatron` wanted was to enforce typing instead of using `any`.

Instead of `closeCalendar = new EventEmitter<any>();` it would be `closeCalendar = new EventEmitter<void>();`

When you called `.emit()`, the type is `void`, because there was nothing in the arguments.

Regards,

fkolar: I would not worry here about the this too much. It really depends of you want to just inform the app that the overlay was closed and not pass a value (which in this case you dont really need to ) or you want to pass a value.


So you can define a Emitter in the way:

```
@Output() readonly something = new EventEmitter();
```

or event like this:

```
@Output() readonly something = new EventEmitter<null>();
```

Algatron: Sort of, basically yes, only that default emitted value is `undefined` if none are provided so I would prob type as that

mikerodonnell89: @Algatron you are correct, I initially misunderstood the purpose of this event, apologies. At the moment this event is meant to be used only internally to inform the date picker to close the calendar.  However I do think a public and documented close event could potentially be useful for developers? @MattL75

MattL75: Yeah that could be useful. If it never returns any value, we should have it typed to undefined or null instead of any.

droshev:
🚢 looks good

PR #589, opened 2019-02-27
fix: remove router dependency from library

### LINKED ISSUE #580: Abstract routing
#### enhancement
Abstract routing

As a consumer, I should be able to use either ngRoute, UIRouter, or expressJS (or some anything else) for my routing. As such, routing should not be a direct dependency on this UI component library, as seen here:

https://github.com/SAP/fundamental-ngx/blob/master/library/package.json#L18

mikerodonnell89:
fixes #580


droshev:
@mikerodonnell89 I think the README to Home doesn't work with the latest changes in master

Issue #596, opened 2019-03-03
Documentation Overhaul


MattL75:
#### Is this a bug, enhancement, or feature request?
Enhancement

#### Briefly describe your proposal.

- [x] Style like fd-react's next version
- [x] Display version number
- [x] Table-like layout for inputs/outputs etc
- [x] Maybe automatically scrape the files for the properties/methods and display the descriptions using comments in the file? [TSDoc](https://github.com/Microsoft/tsdoc) has a parser. We could use it in a per-file basis. Ideally, we'd do it once when we compile. This needs to be investigated further.
- [x] Tabs for Examples/API at the top. API page should contain tables representing all related child component/service/model etc.
- [x] Tabs to switch between the files of an example, also display file name.
- [x] Use StackBlitz as a playground.
- [x] Have a way to hide/display the code.
- [x] Anchors on all <h> tags, with a link icon to easily share a section of the docs.
- [ ] Table of contents on the right side (as with fd-react). They use [TocBot](https://tscanlin.github.io/tocbot/).

More may be added later

droshev:
i think the dynamically generated inputs/outputs would be great

PR #600, opened 2019-03-04
feat: alert component overhaul

### LINKED ISSUE #549: Alert stacking logic needs work

Bug

Depending on when alerts are opened, closed, reopened etc. they can show up stacking over/under each other.

screen shot 2019-02-19 at 9 00 19 pm

### LINKED ISSUE #581: AlertService improvements
enhancement

- `alertService.open()` should accept 
       - as first argument both a `component` class and `templateRef` that only pertain to the alert content, not the alert itself.
       - as second argument a config object with at least a `duration?: number` property with a default behaviour
- both component class and templateRef can invoke `dismiss()` method (no arguments) from within their template.
- `alertService.open()` should return an object that can be invoked without arguments to dismiss the alert immediately.
- `alertService` should have a `dismissAll()` method
- `alertService` must be strongly typed internally and from an API stand point.

**Remark:** When the applications pass a component the markup should be free markup and not passing `fd-alert`. Example:
```
<div class="my-application-class">
  <span>The application is not responding</span>
</div>
```

**Additional ACC**:
- update the alert documentation to reflect the latest API including all the Types/Classes being introduced.
- document if there are breaking changes and what the developers would need to change
- write unit tests

MattL75:
mostly fixes #581 
fixes #549 

- Alerts can now be opened by providing a ComponentType, TemplateRef or string to the open() method of the AlertService.
- An AlertConfig can be passed along to the open() method to customize the alert component.
- Can pass data through the config to the component content or template ref.
- Can inject an AlertRef to the component content to control close/access data.
- Can access AlertRef in the templateRef using let-alert or similar. Call alert.dismiss() for example.
- alertService.open() returns an AlertRef which allows calling .dismiss() to close right away, or subscribe to afterClosed.
- alertService has a dismissAll() method

Documentation needs to be better, but that should also wait on the documentation overhaul that is explained in #596 .

Yes, though the documentation is a little bland. It will be updated with the public API for the service, config and more when we get around to doing the documentation overhaul which is in a separate issue.

When that occurs, we'll add an API tab. There, we'll have available methods for every related service, component or model.


docs/modules/documentation/containers/alert/alert-docs.component.html
```diff
@@ -1,56 +1,72 @@
...
 
 <properties [properties]="{
   inputs: [
-    { name: 'type', description: 'Determines the type of alert - \'error\' (red) or \'warning\' (orange). Defaults to white if no type is provided.' },
-    { name: 'id', description: 'The id of the alert. If omitted, an id will be generated by a HashService.' },
-    { name: 'dismissible', description: 'Shows a dismissible button if set to true. Default is false.' },
-    { name: 'persist', description: 'Boolean, when set to true, the alert will not disappear after a set amount of time.  Default is false.'},
-    { name: 'mousePersist', description: 'Boolean, when set to true, the alert will remain visible as long as the mouse is hovered over it.  The alert will disappear when the mouse leaves the alert.  Default is false.'},
-    { name: 'visibleTime', description: 'Number dictating how long (in milliseconds) the alert will be visible.  Default is 10000.'}
+    { name: 'type', description: 'String. Determines the type of alert. View example below for options. Defaults to white if no type is provided.' },
```
Algatron: More of a "high level" question about the library. @droshev This `width`, `ariaLabel` etc... is some pretty specific options that will tempt people to start playing around with them, but I guess one of the purposes of this library is to make sure users have a consistent look and feel. Completely opposite to smartedit and our usual hybris type extensible mind-set, wouldn't it be better to try and restrict customization here?

docs/modules/documentation/containers/alert/alert-docs.component.html
```diff
@@ -1,56 +1,72 @@
...
+    { name: 'id', description: 'String. The id of the alert. If omitted, an id will be generated.' },
+    { name: 'dismissible', description: 'Boolean. Shows a dismiss button. Default is true.' },
+    { name: 'persist', description: 'Boolean. When set to true, the alert will not disappear after a set amount of time.  Default is false.'},
```
Algatron: Maybe we should combine this with duration? Maybe duration 0?

MattL75: We definitely could. Thoughts on this @mikerodonnell89 @droshev ?

docs/modules/documentation/containers/alert/alert-docs.component.html
droshev: @Algatron you are right. On the other side we shouldn't prevent passing those attributes especially if the applications want to be accessible.

droshev:
@MattL75 I tried the last example from the doc. there is a button to open an alert which gets dismissed after few seconds. After that i still see the `fd-alert` in the DOM

MattL75:
> @MattL75 I tried the last example from the doc. there is a button to open an alert which gets dismissed after few seconds. After that i still see the `fd-alert` in the DOM

If you look above the example there's a note that says that alerts that are defined inline are not removed from the dom

docs/modules/documentation/containers/alert/alert-docs.component.html
droshev: Is there a place where we can see the current predefined types?

docs/modules/documentation/containers/alert/alert-docs.component.html
droshev: I was about to raise the same question as @Algatron

docs/modules/documentation/containers/alert/alert-docs.component.html
```diff
@@ -1,56 +1,72 @@
...
+    { name: 'mousePersist', description: 'Boolean. When set to true, the alert will remain visible as long as the mouse is hovered over it. Default is false.'},
+    { name: 'duration', description: 'Number. Dictates how long in milliseconds the alert will be visible. Default is 10000.'},
+    { name: 'ariaLabel', description: 'String. Aria-label attribute for the alert component.' },
+    { name: 'ariaLabelledBy', description: 'String. Aria-labelledby attribute for the alert component.' },
+    { name: 'width', description: 'String. Width of the alert. Supports all css units.' },
```
droshev: maybe we can show some examples

droshev:
Is there a way(maybe as interim solution) to show the AlertService API documentation?

docs/modules/documentation/containers/alert/alert-docs.component.html
```diff
@@ -1,56 +1,72 @@
...
 <component-example [name]="'ex1'">
   <fd-alert-example></fd-alert-example>
 </component-example>
 <code-example [code]="alertHtml" [language]="'HTML'"></code-example>
-<code-example [code]="alertTs" [language]="'TypeScript'"></code-example>
 
 <separator></separator>
 
-<h3>Alert opened by component reference as content</h3>
+<h2>Alert as Overlay</h2>
```
Algatron: What do you mean as Overlay?
It would be more clear for me if it was something like `Opening Alerts from Services` or `Controlling alerts Programmatically` or `The Alert Service` or... I dunno something like this.

docs/modules/documentation/containers/alert/alert-docs.component.html
mikerodonnell89: I would think a value of -1 for indefinite

docs/modules/documentation/containers/alert/alert-docs.component.html
MattL75: I'll think about a better way to phrase it, wasn't really sure how to put it

docs/modules/documentation/containers/alert/alert-docs.component.html
```diff
@@ -1,56 +1,72 @@
...
 <properties [properties]="{
   inputs: [
-    { name: 'type', description: 'Determines the type of alert - \'error\' (red) or \'warning\' (orange). Defaults to white if no type is provided.' },
-    { name: 'id', description: 'The id of the alert. If omitted, an id will be generated by a HashService.' },
-    { name: 'dismissible', description: 'Shows a dismissible button if set to true. Default is false.' },
-    { name: 'persist', description: 'Boolean, when set to true, the alert will not disappear after a set amount of time.  Default is false.'},
-    { name: 'mousePersist', description: 'Boolean, when set to true, the alert will remain visible as long as the mouse is hovered over it.  The alert will disappear when the mouse leaves the alert.  Default is false.'},
-    { name: 'visibleTime', description: 'Number dictating how long (in milliseconds) the alert will be visible.  Default is 10000.'}
+    { name: 'type', description: 'String. Determines the type of alert. View example below for options. Defaults to white if no type is provided.' },
+    { name: 'id', description: 'String. The id of the alert. If omitted, an id will be generated.' },
```
Algatron: What is the purpose of ID here?

docs/modules/documentation/containers/alert/alert-docs.component.html
```diff
@@ -1,56 +1,72 @@
...
 <code-example [code]="alertHtml" [language]="'HTML'"></code-example>
-<code-example [code]="alertTs" [language]="'TypeScript'"></code-example>
 
 <separator></separator>
 
-<h3>Alert opened by component reference as content</h3>
+<h2>Alert as Overlay</h2>
+<description>
+    The <code>AlertService</code> is available for injection and handles most of the work for this usage. It accepts a component type, a template ref or a string as the alert content.
+    If opening through a component as content, the component should be added to the <code>entryComponents</code> of your application.
```
Algatron: Suggesting rewording: If a component is provided as content, it should be added....

docs/modules/documentation/containers/alert/alert-docs.component.html
MattL75: Well this is left over from the previous alert. I was thinking that without it you wouldn't be able to set an id if you're opening the alert with a service. Might be an issue for e2e tests or other things later on?

docs/modules/documentation/containers/alert/alert-docs.component.html
MattL75: Not sure if that's super useful, it just changes the width?

docs/modules/documentation/containers/alert/alert-docs.component.html
MattL75: we need a doc overhaul

docs/modules/documentation/containers/alert/alert-docs.component.html
MattL75: Agreed, fixed

docs/modules/documentation/containers/alert/alert-docs.component.html
```diff
@@ -1,56 +1,72 @@
...
-    { name: 'mousePersist', description: 'Boolean, when set to true, the alert will remain visible as long as the mouse is hovered over it.  The alert will disappear when the mouse leaves the alert.  Default is false.'},
-    { name: 'visibleTime', description: 'Number dictating how long (in milliseconds) the alert will be visible.  Default is 10000.'}
+    { name: 'type', description: 'String. Determines the type of alert. View example below for options. Defaults to white if no type is provided.' },
+    { name: 'id', description: 'String. The id of the alert. If omitted, an id will be generated.' },
+    { name: 'dismissible', description: 'Boolean. Shows a dismiss button. Default is true.' },
+    { name: 'persist', description: 'Boolean. When set to true, the alert will not disappear after a set amount of time.  Default is false.'},
+    { name: 'mousePersist', description: 'Boolean. When set to true, the alert will remain visible as long as the mouse is hovered over it. Default is false.'},
+    { name: 'duration', description: 'Number. Dictates how long in milliseconds the alert will be visible. Default is 10000.'},
```
Algatron: 10s seems long for a default. Can we expose this, so its a configurable system wide default?

docs/modules/documentation/containers/alert/examples/alert-example.component.html
```diff
@@ -1,28 +1,15 @@
-<fd-alert [dismissible]="true" [type]="'warning'" (close)="closeAlert($event)" #warningAlert>
-    A dismissible warning type alert.
+<fd-alert [type]="'warning'">
+    A dismissible warning alert.
```
Algatron: For all these alert types, should we put the actually string value of the type in bold or quotes or something?
It never actually lists the values you can provide to "type", just part of the sentence of the example.

MattL75: Not sure I understand what you mean. You want to put the 'warning' in the content in bold?

library/src/lib/alert/alert-ref.ts
```diff
@@ -0,0 +1,13 @@
+import { Observable, Subject } from 'rxjs';
+
+export class AlertRef {
+    
+    private readonly _afterDismissed: Subject<undefined> = new Subject<undefined>();
+    public afterDismissed: Observable<undefined> = this._afterDismissed.asObservable();
```
Algatron: is `close` is the output of the fd-alert, should is also be called close here? (or `onClose`)

library/src/lib/alert/alert.component.ts
```diff
@@ -1,137 +1,175 @@
...
-        const openAlerts = Array.prototype.filter.call(alerts, function(alert) {
-            return alert.style.display === 'block';
-        });
-        let totalOffsetHeight = 10;
-        openAlerts.forEach((alert) => {
-            totalOffsetHeight = totalOffsetHeight + alert.offsetHeight + 10;
-        });
-        return totalOffsetHeight + 'px';
+        if (!this.persist) {
+            setTimeout(() => {
```
Algatron: this timeout should short-circuit when someone calls dismiss on the alertRef

docs/modules/documentation/containers/alert/alert-docs.component.html
Algatron: I think if a user really needs a handle on a specific alert, it should be up tot hem to manage from the alertRef

Algatron: (if you mean for selenium you can always do textContaining for ex)

docs/modules/documentation/containers/alert/examples/alert-example.component.html
Algatron: Or something like that, we never actually say what values you can provide for `type`
It should be obvious from the example, but I never trust things to be obvious, would be better if it was explicit :)

library/src/lib/alert/alert.service.ts
```diff
@@ -1,53 +1,111 @@
...
+        const componentRef = factory.create(this.injector);
+        this.appRef.attachView(componentRef.hostView);
+
+        const domElement = (componentRef.hostView as EmbeddedViewRef<any>).rootNodes[0] as HTMLElement;
+        document.body.appendChild(domElement);
+        this.alertContainerRef = componentRef;
+    }
+
+    private destroyAlertContainer(): void {
+        this.appRef.detachView(this.alertContainerRef.hostView);
```
Algatron: this will be called twice actually, is alert is manually dismissed throwing an exception with undefined `alertContainerRef`

docs/modules/documentation/containers/alert/alert-docs.component.html
MattL75: We're gonna talk about it, but it feels slightly outside the scope of the ticket. Maybe if you ad another separate issue?

library/src/lib/alert/alert-ref.ts
MattL75: I'll make the close event onDismiss

library/src/lib/alert/alert.service.ts
MattL75: Not sure what you mean. In what case is this called twice?

library/src/lib/alert/alert.component.ts
MattL75: Can you please clarify?

library/src/lib/alert/alert.service.ts
Algatron: if you leave the default 10s timeout, but dimiss it after 2 sec, it will fire after 2, and then again after 10

docs/modules/documentation/containers/alert/alert-docs.component.html
Algatron: If you don't feel its necessary thats ok with me, I have no requirement for it. I was just thinking if you wanted the default to be 5sec for ex, it would be useful to be able to control the alert timeout from 1 place instead of having to put a new value in every alert invocation in your app.

library/src/lib/alert/alert.service.ts
MattL75: Ah I get it, nice catch. I'll just add a check in the service for an undefined container

docs/modules/documentation/containers/alert/alert-docs.component.html
MattL75: I totally agree that it would be useful, we'll just talk about how we want to go about it etc. So possibly for a future ticket

library/src/lib/alert/alert.component.ts
Algatron: I thought/think maybe this was the cause of the twice-called destraoyContainer.
If you keep a reference to the timeout `const timeoutKey = window.setTimeout....` you can cancel it with `window.clearTimeout(timeoutKey)`
And just intercept someone calling dismiss on the alertRef, and stop this timeout right away

MattL75: Well it **is** the cause of the twice-called destroyContainer, but I don't really want to manage a setTimeout reference. I would need to keep it inside a component attribute, and that just seems like a bad practice. Like I mentioned below, I added a check for the container and it's sufficient for this. There could be a better solution though.

docs/modules/documentation/containers/alert/alert-docs.component.html
droshev: How about we see the predefined types listed like:
`Determines the type of alert. The predefined types are "info", "warning, ''error", and default if no input is provided`

droshev:
@MattL75 I think it is a good idea to bring back the alert playground

MattL75:
> @MattL75 I think it is a good idea to bring back the alert playground

done

droshev:
At the moment the inline alert doesn't accept class attribute. This can be allowed by extending `AbstractFdNgxClass`

MattL75:
> At the moment the inline alert doesn't accept class attribute. This can be allowed by extending `AbstractFdNgxClass`

Done. Not a fan of the AbstractFdNgxClass concept, but absent a better way it'll do for now.

PR #603, opened 2019-03-04
fix: useless menu components are now directives

### LINKED ISSUE #597: menu markup looks odd
enhancement

The current `menu` markup is:
```
<fd-menu>
    <fd-menu-list>
        <li fd-menu-item>
            <a [attr.href]="'#'">Option 1</a>
        </li>
        <li fd-menu-item>
            <a [routerLink]="'#'">Option 2</a>
        </li>
        <li fd-menu-item>
            Option 3
        </li>
    </fd-menu-list>
</fd-menu>
```
I think there is a room for simplifying/flattening it and instead of component `fd-menu-list` we can have an attribute directive on a `<ul>` element. Especially since we do the same for `<li>` with the directive `fd-menu-list`

angular 7.3
fundamental-ngx 0.2.0

MattL75:
fixes #597 

fd-menu-list, fd-menu-title and fd-menu-item have been turned to directives. It allows more control over the elements while conserving functionality and style.

@droshev This is a breaking change and should be documented somewhere. Maybe on the wiki?


droshev:
@MattL75 can we avoid having the wrapping element `fd-menu` and rely only on directives? It will flatten the markup.

library/src/lib/menu/menu-list.directive.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { MenuListDirective } from './menu-list.directive';
+
+describe('MenuListDirective', () => {
+    it('should create an instance', () => {
```
droshev: is there a way we can add more test than just instantiating the directive?

MattL75:
> @MattL75 can we avoid having the wrapping element `fd-menu` and rely only on directives? It will flatten the markup.

We could, but personally I think it would be confusing to use. Having the top-level element as a directive seems much clearer.

library/src/lib/menu/menu-list.directive.spec.ts
MattL75: all it does is add a class to the host

droshev: can we check that? otherwise the test looks like just targeting simple code coverage.

droshev:
🚢 it looks good.

PR #605, opened 2019-03-04
fix: clean up pagination component

### LINKED ISSUE #598: pagination markup can be simplified
enhancement

the current pagination example looks:
```
<fd-pagination [class]="'fd-demo-pagination'" [pagination]="{totalItems: totalItems, itemsPerPage:  itemsPerPage, currentPage: currentPage}"></fd-pagination>
```
The first thing that looks odd is the combination of component and setting a class.
On top of that the markup produced looks like that can be flattened

angular 7.3
fundamental-ngx 0.2.0

mikerodonnell89:
fixes #598 

@droshev I was also thinking about removing the [pagination] input and replacing it with individual [totalItems], [currentPage] and [itemsPerPage] inputs, thoughts?

### DIFF
--- a/library/src/lib/pagination/pagination.component.ts
+++ b/library/src/lib/pagination/pagination.component.ts
@@ -32,8 +32,11 @@ export class PaginationComponent implements OnChanges {
 
     ngOnChanges() {
         this.pages = this.paginationService.getPages(this.getPaginationObject());
-        if (!this.currentPage) {
+        const totalPages = this.paginationService.getTotalPages(this.getPaginationObject());
+        if (!this.currentPage || this.currentPage < 1) {
             this.currentPage = 1;
+        } else if (this.currentPage > totalPages) {
+            this.currentPage = totalPages;
         }
     }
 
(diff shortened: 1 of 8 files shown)


droshev:
@mikerodonnell89 I think this is a good idea and will align with the other components.

droshev:
@mikerodonnell89 one more thing - can we avoid the wrapping `div` in the template and set the class `fd-pagination` on the `<fd-pagination>` node?

droshev:
@mikerodonnell89 Is there a way to do better validation:
screen shot 2019-03-05 at 10 14 52 pm
screen shot 2019-03-05 at 10 15 00 pm

library/src/lib/pagination/pagination.component.ts
```diff
@@ -19,11 +31,14 @@ export class PaginationComponent implements OnChanges {
     constructor(private paginationService: PaginationService) {}
 
     ngOnChanges() {
-        this.pages = this.paginationService.getPages(this.pagination);
+        this.pages = this.paginationService.getPages(this.getPaginationObject());
+        if (!this.currentPage) {
```
droshev: maybe the logic can be extend so it handles the cases when the input is out of the current page ranges

droshev:
@Algatron raised a point to have `Outputs` related to `currentPage`

PR #617, opened 2019-03-10
chore: updated/cleaned readme

MattL75:
- Fixed some punctuation issues.
- Replaced text in the getting started section with actual code.
- Replaced the reference to [(ngModel)] in the getting started section with [(checked)]. The ngModel would not work without FormsModule and we weren't mentioning that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ For an existing Angular application,
     import { FundamentalNgxModule } from 'fundamental-ngx';
     
     @NgModule({
+        ...
         imports: [FundamentalNgxModule],
     })
     ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,28 +45,32 @@ For an existing Angular application,
...
 3. **Import the modules you want to use.**
 
-    To add the entire library, add
+    To add the entire library, add the following code to your main application module.
 
     ```javascript
     import { FundamentalNgxModule } from 'fundamental-ngx';
+    
+    @NgModule({
+        imports: [FundamentalNgxModule],
```
droshev: does this mean they have to replace their `@NgModule` or just add in the imports?

MattL75: Seems obvious to me that it just requires adding the imports. Is it not? I could add some ellipsis to symbolize other stuff around the module import

PR #619, opened 2019-03-11
chore: updated documentation style

MattL75:
Part of #596 

Updated the styling for the documentation to be more like fd-react's.

I made some improvements to their design (at least in my opinion). First, elements aren't fixed-positioned. Instead we rely on a more robust flex layout. We also always show the collapse button, something which came recommended by designers. Otherwise, it's pretty much the same.

Docs are still a heavy WIP, but at least this makes it look nice


docs/styles.scss
```diff
@@ -170,14 +168,6 @@ pre {
     padding: 10px 10px 0 10px;
 }
 
-.skip-navigation {
```
mikerodonnell89: This is for screenreader users

mikerodonnell89:
Seeing two scrollbars in the sidenav and neither of them work.  This is with Chrome
Screen Shot 2019-03-11 at 4 12 24 PM

MattL75:
Hmm I can't see the two scrollbars with chrome, I'll see what I can do

MattL75:
Issues fixed

bcullman:
Consider hiding the "hamburger" menu until you hit the breakpoint to hide the left nav. If you leave it always visible like you do now, you will also need to track it's state. 

For example, on a large screen, hide the left nav by clicking the hamburger, then resize your window. The left nav returns because this is not stateful. 

Simpler to just set its visibility based on media queries alone.

MattL75:
@bcullman I had that behaviour initially with the same reasoning, but our designers recommended keeping the hamburger. After hearing them out I see there's some sense in hiding the sidebar for some users that want to see only the doc text when they read, even on large screens.

droshev:
🚢 looks good

PR #623, opened 2019-03-13
feat: update pagination component to handle promises and provide documentat…

### LINKED ISSUE #611: Pagination component - no way to prevent page changed
bug

With fd-pagination component, when end-user triggers a page change, there is no way for the application developer to prevent the component visually switching to that page. Even if the app developer decides to "fail" the page change operation (maybe the data failed to load?) and return to previous page.
So the app developer is left with simply setting the current page back to the original value, which might not always be straightforward to know what the previous value was, and causes a UI flicker of current page.

current version

mikerodonnell89:
fixes #611


MattL75:
For some reason the build is stuck

PR #625, opened 2019-03-14
chore: keep .github files together

bcullman:
keep .github files together


droshev:
in .github folder are files that are interpreted by github

PR #655, opened 2019-03-29
fix: allow null datepicker model, calendar event/subject refactor

### LINKED ISSUE #629: Clearing datepicker dates has no effect in model

bug

Clearing datepicker dates (via keyboard) from UIs input has no effect in model

latest

1. Set date
2. Check model data
3. Clear date input via keyboard
4. Check model data

At point 4 I would expected to see `null` in model

https://www.useloom.com/share/6fb91055c2cd49d19b2f06cf6945d9b5 (video reproducing problem)

mikerodonnell89:
fixes #629


droshev:
@mikerodonnell89 the datetime picker doesn't update the date if you change it from the calendar

PR #656, opened 2019-03-29
feat: combobox keyboard navigation

### LINKED ISSUE #604: Navigate Combobox with arrow keys
enhancement (accessibility bug)

navigate in combobox with arrow keys and select with enter

Release v0.2.0 (current version)

https://sap.github.io/fundamental-ngx/comboboxInput

mikerodonnell89:
fixes #604


MattL75:
I've been thinking, and I feel like it might be better for us to handle custom keyboard navigation like this **without** focus. So we'd have like a **selected** item that would represent whichever element is currently highlighted, and then changed that depending on keyboard input. Sortof like what Material Design does. Can merge this though, and I guess we can revisit later.

PR #676, opened 2019-04-03
fix: timepicker should call onChange when time model changes

### LINKED ISSUE #675: Timepicker: changing time with the spinner or dropdown inputs does not trigger form validator
```
<fd-time-picker
  required
  formControlName="startTime"
  [displaySeconds]="false">
</fd-time-picker>
```

In this example, `startTime` has a validator, which is properly triggered when the main input is changed, but not when using the spinners or updating the inputs in the dropdown.

mikerodonnell89:
fixes #675

### DIFF
--- a/library/src/lib/time-picker/time-picker.component.html
+++ b/library/src/lib/time-picker/time-picker.component.html
@@ -30,6 +30,7 @@
         <fd-time [disabled]="disabled"
                  [meridian]="meridian"
                  [(ngModel)]="time"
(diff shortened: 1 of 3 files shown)


MattL75:
PR titles can only be feat/fix/chore. Also no need to add the issue number as it's linked automatically in the changelog.

MattL75:
> Also no need to add the issue number as it's linked automatically in the changelog.

PR #678, opened 2019-04-04
feat: allow null values for datetime picker

### LINKED ISSUE #674: Datetime picker should allow null values
User should be able to clear the datetime picker input, nullifying the ngModel

see #665

mikerodonnell89:
fixes #674


docs/modules/documentation/containers/datetime-picker/examples/datetime-example/datetime-example.component.html
```diff
@@ -2,5 +2,5 @@
 <fd-datetime-picker [(ngModel)]="this.date"></fd-datetime-picker>
 
 <span style="padding-left: 20px;">
-    Selected: {{date.toLocaleString()}}
+    Selected: {{getSelectedDate()}}
```
MattL75: How about instead you do `Selected: {{date ? date.toLocaleString() : 'null'}}` ?

That allows removing the method in the ts file, simplifying the whole thing.

library/src/lib/datetime-picker/datetime-picker.component.ts
```diff
@@ -58,6 +58,9 @@ export class DatetimePickerComponent implements OnInit, ControlValueAccessor {
     @Output()
     timeChange: EventEmitter<Date> = new EventEmitter<Date>();
 
+    @ViewChild(TimeComponent)
+    timeComponent;
```
MattL75: Can you think of a way to avoid this? I'd rather not slowly return to a convoluted datetime/date picker. Preferably, they'd only interact with inputs/events but sometimes I know its difficult to avoid so shrug

mikerodonnell89: We can't use the time component's inputs here because the displayed value for the hour does not automatically update when the model changes, and events won't work because we need to pass the value from parent to child and not child to parent.

PR #681, opened 2019-04-05
fix: bug where datepicker events fire when calendar is opened/closed

### LINKED ISSUE #595: Datepicker events problem
Bug

Datepicker ngModelChange and selectedDayChange are slightly broken, and sometimes fire for no reason. We should fix these events, which would then allow people to use the .closeCalendar() method using a template reference variable.

Basically: onChange and selectedDayChange events fire when the calendar is closed/opened. This should not be the case. They should only fire if the values have changed.

Current

mikerodonnell89:
fixes #595

### DIFF
--- a/library/src/lib/date-picker/date-picker.component.ts
+++ b/library/src/lib/date-picker/date-picker.component.ts
@@ -132,7 +132,9 @@ export class DatePickerComponent implements OnInit, OnDestroy, ControlValueAcces
     }
 
     getInputValue(e) {
-        this.dateFromDatePicker.next(e);
+        if (e !== this.inputFieldDate) {
+            this.dateFromDatePicker.next(e);
+        }
     }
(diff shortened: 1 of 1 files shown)


library/src/lib/date-picker/date-picker.component.ts
```diff
@@ -132,7 +132,9 @@ export class DatePickerComponent implements OnInit, OnDestroy, ControlValueAcces
     }
 
     getInputValue(e) {
-        this.dateFromDatePicker.next(e);
+        if (e !== this.inputFieldDate) {
```
MattL75: does this work? usually I compare dates with getTime() but if this works then good

mikerodonnell89: Yes it works

PR #683, opened 2019-04-05
fix: make dropdown position static inside modals

### LINKED ISSUE #680: dropdown position: relative style prevents correct display in modals
Bug

The position: relative style needs to be removed/overriden from the dropdown fundamentals class. It interferes with fd-popovers in modal behaviour.

mikerodonnell89:
fixes #680

### DIFF
--- a/library/src/lib/popover/popover-directive/popover-container.scss
+++ b/library/src/lib/popover/popover-directive/popover-container.scss
@@ -166,3 +166,6 @@
         }
     }
 }
+::ng-deep .fd-dropdown {
+    position: static;
+}
--- a/library/src/lib/popover/popover-directive/popover-container.scss
+++ b/library/src/lib/popover/popover-directive/popover-container.scss
@@ -166,6 +166,3 @@
         }
     }
 }
-::ng-deep .fd-dropdown {
-    position: static;
-}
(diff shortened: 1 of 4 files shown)


MattL75:
@mikerodonnell89 Pretty sure you can make it an override in the popover scss file. The position: relative just isn't necessary with Popper.js.

library/src/lib/popover/popover-directive/popover-container.scss
```diff
@@ -166,3 +166,6 @@
         }
     }
 }
+::ng-deep .fd-dropdown {
+    position: static;
```
MattL75: I was more thinking of putting it inside popover.component.scss (if that doesn't exist, we probs should have one). That way its on the same level as where we actually use the class, and it would also **not** be needed to put ::ng-deep I think

library/src/lib/popover/popover-directive/popover-container.scss
```diff
@@ -166,3 +166,6 @@
         }
     }
 }
+::ng-deep .fd-dropdown {
```
MattL75: I was more thinking of putting it inside popover.component.scss (if that doesn't exist, we probs should have one). That way its on the same level as where we actually use the class, and it would also not be needed to put ::ng-deep I think

PR #689, opened 2019-04-08
fix: refactor table to remove custom elements

### LINKED ISSUE #396: <tr> in fd-table causes Firefox tab to crash
Bug

When I view the documentation for tables (https://sap.github.io/fundamental-ngx/docs/table) or create a <tr>-Element in a fd-table-body, the tab crashes in Firefox (Version 64). Restarting the browser didn't help, but it worked sometime inbetween - so it might not be fully reproducable.
Crash report: https://crash-stats.mozilla.com/report/index/a67c3529-cdae-4c55-9af1-3d01a0190107
We were not able to reproduce it with the current FF Developer Edition

Angular 7.0.0, fundamental-ngx 0.0.26-beta

The crash only occured when crating a <tr>-Element in the table body (<tr> in the table-header actually worked)

mikerodonnell89:
(possibly) fixes #396 

can't reproduce the issue on my machine so we will have to ask around and see if this fixes it.


library/src/lib/table/table-body.component.ts
```diff
@@ -1,7 +1,9 @@
 import { Component } from '@angular/core';
 
 @Component({
-    selector: 'fd-table-body',
+    // TODO to be discussed
+    // tslint:disable-next-line:component-selector
+    selector: '[fd-table-body]',
     templateUrl: './table-body.component.html'
 })
 export class TableBodyComponent {}
```
MattL75: I think this will fix the issue good stuff

These components should be directives though, no? Files should be renamed as well in that case

library/src/lib/table/table.component.html
```diff
@@ -1,8 +1,4 @@
 <table class="fd-table">
-  <thead>
-  <tr>
-    <ng-content select="fd-table-header"></ng-content>
-  </tr>
-  </thead>
-  <ng-content select="fd-table-body"></ng-content>
+  <ng-content select="thead"></ng-content>
+  <ng-content select="tbody"></ng-content>
```
MattL75: maybe add an ng-content for custom stuff? might as well

MattL75:
as mentioned I think you can remove all the css in the tables scss file

PR #726, opened 2019-04-11
docs: breadcrumb typedocs

### LINKED ISSUE #725: add breadcrumb typedocs

mikerodonnell89:
fixes #725


library/src/lib/breadcrumb/breadcrumb.directive.ts
```diff
@@ -1,5 +1,16 @@
 import { Directive } from '@angular/core';
 
+/**
+ * Breadcrumb parent wrapper directive. Must have breadcrumb item child directives.
```
MattL75: why is this directive used as a component

PR #770, opened 2019-05-01
fix: construct calendar when disable function changes

### LINKED ISSUE #744: Unable to trigger Date Picker's disableFunction post initialization

Bug/enchancement

`disableFunction` is fired on component initialization and on nexMonth() prevMonth() but is not fired upon popover open

latest

1) log the `disablefunction`
2) open calendar
3) see that fn is not called
4) unable to call it programatically

n/a

I find it inconsistent that `disableFunction` is being checked upon switching months/years, but is not being fired upon calendar (popover) opening.

More generic solution would be to allow triggering event programatically

mikerodonnell89:
fixes #744


mikerodonnell89:
> Does this also work when used through the datepicker?

Tested this, yes it does, no changes to datepicker code necessary

PR #772, opened 2019-05-02
fix: allow user to set any day as starting day of week

### LINKED ISSUE #761: Allow any weekday as first day of calendar
Bug

For some reason we can only set the start day to be Sunday or Monday.

mikerodonnell89:
…datepicker and datetimepicker.  Also fixes bug where days of the week did not show up in the date selection grid.

fixes #761

### DIFF
--- a/library/src/lib/datetime-picker/datetime-picker.component.ts
+++ b/library/src/lib/datetime-picker/datetime-picker.component.ts
@@ -62,6 +62,9 @@ export class DatetimePickerComponent implements OnInit, ControlValueAccessor {
     timeComponent;
 
     @Input()
+    startingDayOfWeek: number = 0;
+
+    @Input()
     disableFunction = function(d): boolean {
         return false;
     };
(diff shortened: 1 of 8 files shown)


library/src/lib/datetime-picker/datetime-picker.component.ts
```diff
@@ -61,6 +61,9 @@ export class DatetimePickerComponent implements OnInit, ControlValueAccessor {
     @ViewChild(TimeComponent)
     timeComponent;
 
+    @Input()
+    startingDayOfWeek: number = 0;
```
MattL75: Need to ensure this is between 0 and 6, no?

mikerodonnell89: There's a note in the docs about 1 being monday, 2 being tuesday etc.  But will limit this to 0-6

MattL75: ideally they should match internally & externally

mikerodonnell89: Can you elaborate, not sure what you mean by this

MattL75: like the value being input to the component should be the same as the one being used internally

MattL75:
image
image

Days are different. In the calendar below April 30th does not appear.

PR #779, opened 2019-05-03
feat: calendar range disable functions

### LINKED ISSUE #746: Range Datepicker's disableFunction behaviour

enhancement

`disableFunction` makes sanse for single datepicker/calendar, but if range value is used - validators should be separated. Start date can have different logic than end date. Its not currently possible to handle that.

latest

1) Use multi range picker
2) Try to create scenario where user:
- cannot set start after 10.04
- cannot set end before 20.04
3) Its impossible
By allowing 8.04 and 9.04 for start date, you also implicitly allow it for end date because these share the constraints (one, shared disableFunction) -> you can then select start 8.04 and end 9.04 with no problem

mikerodonnell89:
fixes #746

### DIFF
--- a/docs/modules/documentation/containers/calendar/calendar-docs.component.html
+++ b/docs/modules/documentation/containers/calendar/calendar-docs.component.html
@@ -17,7 +17,12 @@
 <separator></separator>
 
 <h2>Range Selection Calendar</h2>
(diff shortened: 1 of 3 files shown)


mikerodonnell89:
Closing, need to add this to datepicker and datetimepicker and write tests

PR #781, opened 2019-05-04
chore: removed useless generated components and fixed dependencies

MattL75:
For some reason we still had
* fundamental-ngx.component.ts
* fundamental-ngx.component.spec.ts
* fundamental-ngx.service.ts
* fundamental-ngx.service.spec.ts

which are files generated when you first build an angular library through the CLI. 

Also added @angular/animations to peer dependencies.


mikerodonnell89:
@MattL75 merge conflicts

PR #801, opened 2019-05-10
feat: i18n labels for remaining components

### LINKED ISSUE #426: i18n support
Feature

One of the key requirements is `i18n`. Even Angular now provides some limitted support along with its CLI but its not yet ready. Currently it relies heavily on static strings which you must have already in HTML templates. For most of the simple usecases it works, but in enterprise you must be able to retrieve your translation string at the runtime.

If you need dynamicity then people use ngx-translate, where the guy behind it is now at Google and working runtime version. 

Not sure if there is any other solution to this?

mikerodonnell89:
fixes #426


MattL75:
Looks good! Can you add an 'Internationalization' example to the Time component? I know it won't show any difference, but just to show that it's there I guess?

Issue #527, opened 2019-02-15
Date/time localization and/or specific formatting


droshev:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
allow the applications to pass specific time/date format for display

MattL75:
Date formatting has been added to the datepicker. Formatting to time and datetime are being tracked in separate issues.

PR #807, opened 2019-05-13
feat: added formatting support to datepicker

MattL75:
Part of #527

Adds a dependency provider option to calendar, which also has been adapted to support datepicker. 

There's only one 'bug' in this that I can think of, and its that with the Date object, 20 can = 2020, so if you try to delete 2019 manually, you'll keep getting reset to 2020. Not too much we can do about that without being inconsistent. Calendar2 will fix this.

Datetime picker is a TODO, and probably the next thing on the list. After that, I'll look into making Calendar2, which should be fun.


docs/app/documentation/component-docs/date-picker/examples/date-picker-format-example.component.ts
```diff
@@ -0,0 +1,60 @@
...
+        return new Date(Number(values[0]), values[1] - 1, values[2]);
+    }
+
+    public format(date: Date): string {
+        let monthStr = (date.getMonth() + 1).toLocaleString();
+        if (monthStr.length === 1) {
+            monthStr = '0' + monthStr;
+        }
+
+        return date.getFullYear() + '-' + monthStr + '-' + date.getDate();
```
InnaAtanasova: I think it's a good idea to have more details/comments about this line of code and the whole `format` function in order to avoid any confusion for the user.

library/src/lib/calendar/calendar.component.ts
```diff
@@ -689,29 +690,26 @@ export class CalendarComponent implements OnInit, OnDestroy, AfterViewChecked, C
         }
     }
 
-    validateDateFromDatePicker(date: Array<any>): boolean {
-        let isInvalid: boolean = false;
-        const month = date[0];
-        const day = date[1];
-        const year = date[2];
-        let numOfDaysInMonth = 0;
+    validateDateFromDatePicker(date: Date): boolean {
```
InnaAtanasova: Since we are returning a bool value, it's probably a good idea to give a proper name to this function.

MattL75: we're blowing it up anyways 🤷‍♀

Issue #687, opened 2019-04-08
Add documentation typedoc comments for all components


MattL75:
#### Is this a bug, enhancement, or feature request?
Enhancement

#### Briefly describe your proposal.
Add documentation typedoc comments to all components. See Alert and Modal for examples.

List of done components: 

- [x] Action Bar (#722)
- [x] Alert (#684)
- [x] Badge (#724)
- [x] Breadcrumb (#726)
- [x] Button (#728) 
- [x] Button Group (#782) 
- [x] Calendar (#867) 
- [x] Combobox Input (#799)
- [x] Date Picker (#867) 
- [x] Datetime Picker (#778) 
- [x] Dropdown
- [x] File Input (#711)
- [x] Form (#777) 
- [x] Icon (#794)
- [x] Identifier (#796)
- [x] Infinite Scroll (#767)
- [x] Image (#797)
- [x] Inline Help (#798) 
- [x] Input Group (#799)
- [x] List (#802 )
- [x] Loading Spinner (#810)
- [x] Menu (#751) 
- [x] Modal (#684)
- [x] Multi Input (#738) 
- [x] Pagination (#811)
- [x] Panel (#765) 
- [x] Popover (#718) 
- [x] Popover Directive (#718) 
- [x] Scroll Spy (#812)
- [x] Search Input (#739) 
- [x] Shellbar (#815)
- [x] Side Navigation (#814)
- [x] Table (#816)
- [x] Tabs (#688) 
- [x] Tile (#809)
- [x] Time (#707)
- [x] Time Picker (#708)
- [x] Toggle (#755) 
- [x] Token (#766)

droshev:
Can we try to wrap up this ticket?

PR #812, opened 2019-05-14
docs: added typedocs for scrollspy component

InnaAtanasova:

part of [#687](https://github.com/SAP/fundamental-ngx/issues/687)

added documentation typedoc comments for scrollspy component

### DIFF
--- a/library/src/lib/scroll-spy/scroll-spy.directive.ts
+++ b/library/src/lib/scroll-spy/scroll-spy.directive.ts
@@ -1,26 +1,46 @@
 import { Directive, ElementRef, EventEmitter, HostListener, Input, Output } from '@angular/core';
 
+/**
+ * A directive designed to help navigation elements determine the element currently in view of the user. 
+ */
 @Directive({
     selector: '[fdScrollSpy]'
 })
 export class ScrollSpyDirective {
-
+    /**
+     * An array of tags to track.
+     */
     @Input()
     public trackedTags: string[] = [];
 
+    /**
+     * Whether events are still fired if there is no tag present on the user's screen.
+     */
     @Input()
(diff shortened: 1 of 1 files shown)


library/src/lib/scroll-spy/scroll-spy.directive.ts
```diff
@@ -1,26 +1,46 @@
...
+     */
     @Input()
     public trackedTags: string[] = [];
 
+    /**
+     * Whether an arrow is displayed when the user scrolls to the absolute top or bottom.
+     * To hide the arrow, set the property to *true*.
+     */
     @Input()
     public fireEmpty: boolean = false;
```
MattL75: There is no visual aspect to this component. All this input does is determine whether events are still fired if there is no tag present on the user's screen. It *can* be used to do as you describe, but that's up to the user.

library/src/lib/scroll-spy/scroll-spy.directive.ts
```diff
@@ -1,26 +1,46 @@
...
+     * 0 for the top and 1 for the bottom.
+     */
     @Input()
     public targetPercent: number = 0;
 
+    /** 
+     * Event fired on the scroll element when a new item becomes activated by the scrollspy . 
+     */
     @Output()
     public spyChange: EventEmitter<HTMLElement> = new EventEmitter<HTMLElement>();
```
MattL75: add readonly pls.

also maybe say that the returned value (through $event) is the HTMLElement itself

library/src/lib/scroll-spy/scroll-spy.directive.ts
InnaAtanasova: Done

PR #819, opened 2019-05-16
fix: only trigger onChange if input value changes

### LINKED ISSUE #817: [Datepicker] clicking on calendar icon sets the date

bug

Clicking on calendar icon sets the date (to null) triggering validation. The same will not happen if you enter calendar by clicking on the input rather than calendar icon.

latest

calendar

mikerodonnell89:
fixes #817

### DIFF
--- a/library/src/lib/date-picker/date-picker.component.html
+++ b/library/src/lib/date-picker/date-picker.component.html
@@ -9,7 +9,7 @@
                    [value]="inputFieldDate"
                    [placeholder]="placeholder"
                    (keyup)="getInputValue(datePicker.value)"
-                   (click)="isOpen = true"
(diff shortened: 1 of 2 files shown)


MattL75:
This fixes one issue, but not the second.

Now the initial date is not populated to the calendar on first open :\

mikerodonnell89:
> This fixes one issue, but not the second.
> 
> Now the initial date is not populated to the calendar on first open :\

Fixed, good catch

MattL75:
Looking good! One last thing: maybe make clicking on input keep the popover open, while the button keeps toggling it. This will probably mean making the Open method an actual open method (not a toggle) and creating a toggle method

PR #820, opened 2019-05-16
fix: bug where safari can't parse dates with hyphens

mikerodonnell89:


MattL75:
image

maybe add a check?

PR #822, opened 2019-05-16
fix: datetime invalid time in input should not nullify ngModel

### LINKED ISSUE #821: Datetimepicker - invalid time in input should not nullify model

Bug.  An invalid time input should not nullify the datetime picker ngModel

mikerodonnell89:
fixes #821


MattL75:
Think you can make the date populate on when you click the input? I imagine its the same issue as with the datepicker before where the onChange event fires even if there's no change, and the input if clicked doesn't populate the calendar.

PR #825, opened 2019-05-17
show how to use cdk-table with fd-table

mikerodonnell89:


package.json
```diff
@@ -30,6 +30,7 @@
     },
     "dependencies": {
         "@angular/animations": "^7.2.7",
+        "@angular/cdk": "^7.3.7",
```
MattL75: I think we wanted to include the cdk as a dependency of the actual library. I still think we should run this by the guys in the fundamentals meeting.

PR #826, opened 2019-05-18
chore: fixed build

MattL75:


droshev:
i think it would be better to be more descriptive.

MattL75:
@droshev nah

PR #832, opened 2019-05-20
fix: time/date/datetime picker should all handle null values the same

### LINKED ISSUE #831: An empty string in the datetime picker component is incorrectly being deemed invalid
Bug

7.2.11

1. Open the datetime picker component
2. Select a date
3. Click outside to close popover
4. Empty datetimer picker input field
5. Notice that the validation marks the input with a red border

We believe this is incorrect as emptying the input field should nullify the the date object. We think that an empty field should be valid.

https://github.com/SAP/fundamental-ngx/blob/337f0471517f894847342f6eedc6105f149fe043/library/src/lib/datetime-picker/datetime-picker.component.ts#L233-L261

Note that `new Date("")`  give `Invalid Date` on [L236](https://github.com/SAP/fundamental-ngx/blob/337f0471517f894847342f6eedc6105f149fe043/library/src/lib/datetime-picker/datetime-picker.component.ts#L236) so the field is marked invalid by the if statement on [L251](https://github.com/SAP/fundamental-ngx/blob/337f0471517f894847342f6eedc6105f149fe043/library/src/lib/datetime-picker/datetime-picker.component.ts#L251)

Observed on chrome and firefox (both latest).

mikerodonnell89:
fixes #831 

also adds a boolean "allowNull" input to these components, so the dev can decide if a null input would be considered valid or not

### DIFF
--- a/library/src/lib/date-picker/date-picker.component.ts
+++ b/library/src/lib/date-picker/date-picker.component.ts
@@ -86,6 +86,10 @@ export class DatePickerComponent implements OnInit, OnDestroy, ControlValueAcces
     @Input()
     displayCalendarToggleLabel: string = 'Display calendar toggle';
 
+    /** Whether a null input is considered valid. */
+    @Input()
+    allowNull: boolean = true;
+
     @Input()
     disableFunction = function(d): boolean {
         return false;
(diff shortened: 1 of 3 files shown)


mikerodonnell89:
Closing, input automatically repopulates with the past date when the user enters a null value, leaves the input, and re-focuses it.  Also the input automatically clears when the user re-focuses the input after entering a partially valid value - it should allow the user to finish entering a value

PR #833, opened 2019-05-21
docs: demonstrate how to use angular cdk with fd-ngx table

mikerodonnell89:


docs/app/documentation/component-docs/panel/examples/panel-edge-bleed-example.component.html
```diff
@@ -15,6 +15,6 @@
                     </td>
                 </tr>
             </tbody>
-        </fd-table>
+        </fd-table-->
```
MattL75: comment?

library/src/lib/table/table.component.ts
```diff
@@ -1,12 +1,13 @@
...
  */
-@Component({
-    selector: 'fd-table',
-    templateUrl: './table.component.html',
-    encapsulation: ViewEncapsulation.None
+@Directive({
+    // tslint:disable-next-line:directive-selector
+    selector: '[fd-table]',
+    host: {
+        class: 'fd-table'
```
MattL75: use hostbinding and document this as a breaking change in the wiki page for it, under 0.5.1

library/src/lib/table/table.module.ts
```diff
@@ -1,12 +1,13 @@
...
-import { TableComponent } from './table.component';
+import { TableDirective } from './table.component';
 import { ColumnSortableDirective } from './column-sortable.directive';
+import { CdkTableModule } from '@angular/cdk/table';
 
 @NgModule({
-    imports: [CommonModule],
-    declarations: [TableComponent, ColumnSortableDirective],
-    exports: [TableComponent, ColumnSortableDirective]
+    imports: [CommonModule, CdkTableModule],
```
MattL75: should not import CdkTable module in the lib itself, it should only be for the docs

MattL75:
As I described here: https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes, only major version bumps (fix bumps since we are under 0) are given their own pages. You can just list the change in the wiki I just linked. If we make a page for every minor version we'll get a much too crowded wiki real quick

PR #837, opened 2019-05-22
fix: allow datetime null

### LINKED ISSUE #831: An empty string in the datetime picker component is incorrectly being deemed invalid

mikerodonnell89:
fixes #831 

also added an allowNull input if the dev would like to consider null values invalid


library/src/lib/date-picker/date-picker.component.ts
```diff
@@ -86,6 +86,10 @@ export class DatePickerComponent implements OnInit, OnDestroy, ControlValueAcces
     @Input()
     displayCalendarToggleLabel: string = 'Display calendar toggle';
 
+    /** Whether a null input is considered valid. */
+    @Input()
+    allowNull: boolean = true;
```
MattL75: How about **required**

library/src/lib/datetime-picker/datetime-picker.component.ts
```diff
@@ -92,6 +92,10 @@ export class DatetimePickerComponent implements OnInit, OnDestroy, ControlValueA
     @Input()
     displayDatetimeToggleLabel: string = 'Display calendar toggle';
 
+    /** Whether a null input is considered valid. */
+    @Input()
+    allowNull: boolean = true;
```
MattL75: How about **required**

library/src/lib/datetime-picker/datetime-picker.component.ts
```diff
@@ -174,12 +178,11 @@ export class DatetimePickerComponent implements OnInit, OnDestroy, ControlValueA
     }
 
     /** Opens the popover. */
-    openPopover(): void {
+    openPopover(inputFieldDate?): void {
```
MattL75: specify argument type

library/src/lib/time-picker/time-picker.component.ts
```diff
@@ -66,6 +66,10 @@ export class TimePickerComponent implements ControlValueAccessor, OnInit {
     @Input()
     timePickerInputLabel: string = 'Time picker input';
 
+    /** Whether a null input is considered valid. */
+    @Input()
+    allowNull: boolean = true;
```
MattL75: How about **required**

library/src/lib/date-picker/date-picker.component.ts
MattL75: actually nvm all these, allowNull is fine

PR #841, opened 2019-05-23
docs: demonstrate how to use drag and drop with CDK table

mikerodonnell89:
Demonstrate how to use drag and drop with the CDK table


MattL75:
Drag row 1 to a spot, then try to drag it to another spot and the behaviour seems to break.

Maybe show a better drag/drop styling? Right now its not clear that a row is being dragged or that one is being replaced except by the number

mikerodonnell89:
> Drag row 1 to a spot, then try to drag it to another spot and the behaviour seems to break.
> 
> Maybe show a better drag/drop styling? Right now its not clear that a row is being dragged or that one is being replaced except by the number

Changed the cursor to grab for the table rows, fundamental has a hover background color for the rows, I think it's a little more pronounced on some monitors than others.

MattL75:
This is nice, but as an addition maybe you could add a background color (white?) to the row being dragged? Just so it isn't transparent. And maybe make the placeholder in the table just like grey with an outline? Or something similar to indicate that its being dragged

PR #845, opened 2019-05-27
chore: lock update

MattL75:
Updated lock


droshev:
again?

MattL75:
Yep, for some reason npm installed on the new master with the updated lock and it triggered some changes so 🤷‍♀

PR #848, opened 2019-05-28
fix: dismissing alerts display: none style fix

### LINKED ISSUE #840: Encapsulation none breaks static alerts
Bug

We currently add a display: none style directly to the element in the alert component. This doesn't work with no style encapsulation. We should instead define a class and apply that to the element.

mikerodonnell89:
fixes #840


MattL75:
breaks dynamic inline alerts
image

PR #861, opened 2019-06-04
Date parser data/month switch fix

JKMarkowski:
It's possible to reproduce error on docs:  {{baseurl}}/datetime-picker
Every time the datepicker popup is opened the days and months chenge their places
It's caused by wrong usage of parsing function.

`public parse(value: string): Date {
        console.log(value); // 5.06.2019
        console.log(new Date(value)); // Mon May 06 2019 00:00:00 GMT+0200
        return new Date(value);
    }`

### DIFF
--- a/library/src/lib/calendar/format/date-parser.ts
+++ b/library/src/lib/calendar/format/date-parser.ts
@@ -42,7 +42,7 @@ export class DateFormatParserDefault extends DateFormatParser {
      * @param value String to convert to a date.
      */
     public parse(value: string): Date {
-        return new Date(value);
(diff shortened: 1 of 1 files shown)


mikerodonnell89:
The abstract parse() function is meant to be implemented by you if you do not want to use our default.  The function you're changing in this pull request is the default implementation we've chosen, which is JavaScript's default way of parsing strings to dates (mm/dd/yyyy or mm.dd.yyyy).  If you'd like to parse dates passed in as dd.mm.yyyy you must implement a class that extends DateFormatParser with your own parsing function

PR #862, opened 2019-06-04
fix: month gets skipped when the current date is the 31st and the user clicks next month button

### LINKED ISSUE #855: Bug: Calendar component next month click

Fundamental-react was having some problems with the Calendar component this morning due to it being the 31st, which shows up in the next month on our calendar. I was looking at ngx to see how you handled it and found that the next button skips June. I'm betting this is due to the same problem as in fundamental-react - having previous months days show up in the calendar. 

0.5.2

mikerodonnell89:
fixes #855 

Fixed bug where if the 31st day of the month is selected and the user clicks the next month button, and the next month has fewer than 31 days, the next month will be skipped and the one after that will be displayed.  So for example if Jan 31 is selected and the next month button is clicked, March will be displayed.

### DIFF
--- a/library/src/lib/calendar/calendar.component.ts
+++ b/library/src/lib/calendar/calendar.component.ts
@@ -614,7 +614,19 @@ export class CalendarComponent implements OnInit, OnDestroy, AfterViewChecked, C
     }
 
     setCurrentMonth(month: number) {
+        // get the current date of the month
+        const currentDate = this.date.getDate();
+        /*
+         set the date to the first for now, to prevent skipping a month
+         in the event that the currentDate is 31 and the next month has 30 days
+         */
+        this.date.setDate(1);
+        // set the month
         this.date.setMonth(month);
+        // get the number of days in this month
+        const daysInMonth = new Date(this.date.getFullYear(), this.date.getMonth() + 1, 0).getDate();
+        // restore the date to whichever number is lower, today's date or the number of days in this month
+        this.date.setDate(Math.min(currentDate, daysInMonth));
         this.month = this.date.getMonth();
         this.monthName = this.monthsFullName[this.date.getMonth()];
         this.year = this.date.getFullYear();
--- a/library/src/lib/calendar/calendar.component.ts
+++ b/library/src/lib/calendar/calendar.component.ts
@@ -616,17 +616,21 @@ export class CalendarComponent implements OnInit, OnDestroy, AfterViewChecked, C
     setCurrentMonth(month: number) {
         // get the current date of the month
         const currentDate = this.date.getDate();
+        // get the number of days in the new month
+        const daysInNewMonth = new Date(this.date.getFullYear(), month + 1, 0).getDate();
         /*
-         set the date to the first for now, to prevent skipping a month
+         if the currentDate > daysInNewMonth, set the date to the first for now, to prevent skipping a month
          in the event that the currentDate is 31 and the next month has 30 days
          */
-        this.date.setDate(1);
+        if (currentDate > daysInNewMonth) {
+            this.date.setDate(1);
+        }
         // set the month
         this.date.setMonth(month);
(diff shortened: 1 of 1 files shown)


library/src/lib/calendar/calendar.component.ts
```diff
@@ -614,7 +614,23 @@ export class CalendarComponent implements OnInit, OnDestroy, AfterViewChecked, C
     }
 
     setCurrentMonth(month: number) {
+        // get the current date of the month
+        const currentDate = this.date.getDate();
+        // get the number of days in the new month
+        const daysInNewMonth = new Date(this.date.getFullYear(), month + 1, 0).getDate();
```
MattL75: if this is december, will the Date object automatically go to January of next year?

library/src/lib/calendar/calendar.component.ts
```diff
@@ -614,7 +614,23 @@ export class CalendarComponent implements OnInit, OnDestroy, AfterViewChecked, C
...
+        /*
+         if the currentDate > daysInNewMonth, set the date to the first for now, to prevent skipping a month
+         in the event that the currentDate is 31 and the next month has 30 days
+         */
+        if (currentDate > daysInNewMonth) {
+            this.date.setDate(1);
```
MattL75: Does this go to the next month?

library/src/lib/calendar/calendar.component.ts
mikerodonnell89: yes
```
new Date(2019, 12, 1) = Wed Jan 01 2020 00:00:00 GMT-0700 (Mountain Standard Time)
```

library/src/lib/calendar/calendar.component.ts
mikerodonnell89: No, it sets the date for the current month to 1, but sets it back later

MattL75: hmm not sure how friendly this for ux

mikerodonnell89: This is all happening internally, not with the actual ngModel the user sees

MattL75: Yeah I know, but so if the user is on the 31st and clicks next, it goes to date(1) of the same month that was currently displayed? Or the next month?

mikerodonnell89: The month that is currently displayed, but I don't think this really matters because it reverts back a few lines later

MattL75: Can you describe the PR in the description? And also exactly the flow of the bug? Like for instance, if the user presses the next month button, this and this happens.

mikerodonnell89: Added a little description.  The original ticket is linked in the description with a gif demonstration

MattL75:
Why does it skip February if you press next in January 31st?

mikerodonnell89:
> Why does it skip February if you press next in January 31st?

It's just the way dates in JS work... if date===jan 31 and you say date.setMonth(feb) it'll skip over Feb and go straight to March.  Because there is no Feb 31.

You can easily test this bug on master, and the big fix on this branch, by setting a specific selectedDay.date in the calendar-single-example.component.ts example code

MattL75:
Yeah but I'm saying instead of doing that, why not set it to Feb 1st?

mikerodonnell89:
> Yeah but I'm saying instead of doing that, why not set it to Feb 1st?

What I'm doing here is setting it to the latest date of the new month when the current month has more days than the new month and one of those additional days is the set date. Because this is only setting the internal date the calendar is using to know which month to display, not the actual "selectedDate", I guess it doesn't really matter what date is being set and it would simplify the logic to just set the 1st date of the new month.  Is that what you're getting at?

Issue #654, opened 2019-03-29
Improve Alert Responsiveness


MattL75:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
Alert default size is currently 33% of the screen size. We should have this default behaviour be responsive, or at least make it responsive in the docs and show how it's done.

PR #863, opened 2019-06-05
fix: Add min-width option on alerts with default value

JKMarkowski:
https://github.com/SAP/fundamental-ngx/issues/654
I included minimum width option to alert component which handles responsive problems. By default right now min-width is set to 300px, which means that it will look good on every device.
I put this option usage into one example

### DIFF
--- a/docs/app/documentation/component-docs/alert/examples/alert-component-as-content-example.component.ts
+++ b/docs/app/documentation/component-docs/alert/examples/alert-component-as-content-example.component.ts
@@ -14,7 +14,7 @@ export class AlertComponentAsContentExampleComponent {
     openFromComponent() {
         this.alertService.open(AlertContentComponent, {
             type: 'warning',
-            minWidth: '400px',
+            minWidth: '300px',
             mousePersist: true,
             duration: 7500,
             data: {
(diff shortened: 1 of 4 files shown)


docs/app/documentation/component-docs/alert/examples/alert-component-as-content-example.component.ts
```diff
@@ -14,6 +14,7 @@ export class AlertComponentAsContentExampleComponent {
     openFromComponent() {
         this.alertService.open(AlertContentComponent, {
             type: 'warning',
+            minWidth: '400px',
```
MattL75: Maybe just put this one down to 300, even if it's the default. 400px will break on some phone screens which defeats the purpose of this feature. The rest seems pretty good to me though. Eventually we'll want to address #865 but that can wait for now.

JKMarkowski: Hi @MattL75 I changed 400px to 300px in example.

library/src/lib/alert/alert-service/alert.service.ts
```diff
@@ -51,6 +51,11 @@ export class AlertService {
             alertConfig.width = '33vw';
         }
 
+        // Ensure default minimum width
+        if (alertConfig && !alertConfig.minWidth) {
```
MattL75: When your pr #868 is merged you will be able to remove this check 👍

Issue #865, opened 2019-06-05
Make alert service follow the modal service pattern


MattL75:
#### Is this a bug, enhancement, or feature request?
Enhancement

Currently, the alert service doesn't have **real** default values coming from the config object. Instead, the default values are manually set in the service. This is not scalable. ModalService has a better option.

```typescript
modalConfig = Object.assign(new ModalConfig(), modalConfig);
```

Using this method, we assign all the properties of the user's modalConfig to a default config, overriding only the user's parameters.

PR #868, opened 2019-06-06
fix: make alert service follow the modal service pattern

JKMarkowski:
https://github.com/SAP/fundamental-ngx/issues/865
Put default config to alert open function
Not needed

### DIFF
--- a/library/src/lib/alert/alert-service/alert.service.ts
+++ b/library/src/lib/alert/alert-service/alert.service.ts
@@ -46,10 +46,8 @@ export class AlertService {
             this.openAlertContainer();
         }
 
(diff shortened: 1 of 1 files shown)


MattL75:
It's not mentioned anywhere that I know of, but we have the following guidelines for PR titles (if we don't follow them the automatic rc versions are not released).

Prefix with...
```
'fix:' --> increments 0.0.x, used for non-breaking changes and bugfixes
'feat:' --> increments 0.x.0, used for features or breaking changes (for now since we are under version 0)
'chore:' --> for stuff like updating a dependency
'docs:' --> for documentation updates
```

Issue #704, opened 2019-04-09
Table Component: inner scroll


ditiashova:
#### Inner scroll for the table component
```
"@angular/core": "^7.0.1",
"fundamental-ngx": "^0.2.0",
```

Hi,

currently table doesn't have inner scroll and it can be an issue in some cases like the following:
1) 1250px is fine for it:
![image](https://user-images.githubusercontent.com/25684357/55808346-9465ec00-5aec-11e9-9fea-73e3e436ce0f.png)
2) then I narrow the screen and get this:
![image](https://user-images.githubusercontent.com/25684357/55808423-b7909b80-5aec-11e9-95f4-9a2ed09051a4.png)
![image](https://user-images.githubusercontent.com/25684357/55808535-ef97de80-5aec-11e9-99fc-2d638593ea2a.png)
And at some point I'm not able to access my information anymore.

PR #869, opened 2019-06-06
fix: add possibility to include scrollable table + example in docs

JKMarkowski:
…n docs

https://github.com/SAP/fundamental-ngx/issues/704
I added directive which makes scrolling table horizontally possible.
Yes, the documentation is updated and contains one of usage example.


docs/app/documentation/component-docs/table/table-docs.component.ts
```diff
@@ -8,10 +8,17 @@ import * as tableCheckHtml from '!raw-loader!./examples/table-checkboxes-example
...
+import * as tableResponsiveHtml from '!raw-loader!./examples/table-responsive-example.component.html';
+import * as tableResponsiveTs from '!raw-loader!./examples/table-responsive-example.component.ts';
 
 @Component({
     selector: 'app-table',
-    templateUrl: './table-docs.component.html'
+    templateUrl: './table-docs.component.html',
+    styles: [`
+        ::ng-deep app-table .fd-tile {
+            display: block;
```
MattL75: hmm kind of annoying that this is needed in the first place, but I get it and it's fine to keep it. Maybe something we can consider improving later

mikerodonnell89:
This works really good, but I am wondering if we want to test for div width and apply the fixed class to the parent based on that rather than have it as a directive?  Could use the scrollbar width changes for frame of reference but this might be something we should check with the fundamental design team

JKMarkowski:
Actually in fiori fundamental library there is no such thing allowing to scroll horizontally. --fixed class is more useful on adding fixed cell, so it's not needed to be used in this case.

MattL75:
Yeah I'm good to merge this - maybe we could add the fixed cell at some point, or even make it toggle-able in the example. I'll wait for word from @mikerodonnell89 though.

Issue #864, opened 2019-06-05
fd-panel-grid class not properly updated after changing col attribute


ahoracek:
#### Is this a bug, enhancement, or feature request?

Bug

#### Briefly describe your proposal.

`fd-panel-grid` `class` value is not properly updated after changing `[col]` attribute

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)

Angular 7.2.0
fiori-fundamentals: 1.5.7
fundamental-ngx: 0.5.1
using SCSS

#### If this is a bug, please provide steps for reproducing it.

`fd-panel-grid` with 4 columns, content in one column has "Hide" button.

when page is loaded
```html
<fd-panel-grid class="fd-panel-grid--4col fd-panel-grid--nogap fd-panel-grid" ng-reflect-col="4" ng-reflect-nogap="true">
```
is generated. When clicking Hide button, change in the `[col]` property causes the grid is updated as 
```html
<fd-panel-grid class="fd-panel-grid--3col" ng-reflect-col="3" ng-reflect-nogap="true">
```

The updated `class` is missing `fd-panel-grid--nogap fd-panel-grid` classes so content of the grid is not properly organised (cols stack vertically instead of horizontally).

#### Please provide relevant source code if applicable.

- test.component.html

```html
<fd-panel-grid [nogap]="true" [col]="col">
    <fd-panel *ngIf="expanded">
        <button fd-button (click)="expanded = !expanded">Hide</button>
    </fd-panel>
    <fd-panel>Col 2</fd-panel>
    <fd-panel>Col 3</fd-panel>
    <fd-panel>Col 4</fd-panel>
</fd-panel-grid>
```

- test.component.ts

```ts
@Component({
    selector: 'app-test',
    templateUrl: './test.component.html',
    styleUrls: ['./test.component.scss']
})
export class TestComponent {
    expanded: boolean = true;
    get col() {
        return this.expanded ? 4: 3
    }
    constructor() { }
}
```

- test.component.scss
```scss
// empty
```

#### Is there anything else we should know?

setting style
```html
<fd-panel-grid class="fd-panel-grid--nogap fd-panel-grid"> ... </fd-panel-grid>
```
solves the problem, however it's workaround.

ahoracek:
[...]

PR #881, opened 2019-06-10
fix: Add removed classes while changed on grid panel

JKMarkowski:
https://github.com/SAP/fundamental-ngx/issues/864
In issue there is noticed that, basic fd-grid-panel classes dissapear, when number of columns is changed 
Not needed.

### DIFF
--- a/library/src/lib/panel/panel-grid/panel-grid.component.ts
+++ b/library/src/lib/panel/panel-grid/panel-grid.component.ts
@@ -15,7 +15,7 @@ export class PanelGridComponent extends AbstractFdNgxClass {
     @Input()
     col: number;
 
-    /** Whether the grid shoul have a gap. */
+    /** Whether the grid should have a gap. */
     @Input()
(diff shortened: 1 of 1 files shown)


library/src/lib/panel/panel-grid/panel-grid.component.ts
```diff
@@ -15,7 +15,7 @@ export class PanelGridComponent extends AbstractFdNgxClass {
     @Input()
     col: number;
 
-    /** Whether the grid shoul have a gap. */
+    /** Whether the grid should have a gap. */
     @Input()
     @HostBinding('class.fd-panel-grid--nogap')
```
MattL75: is there a point in keeping this line if we're just adding the class in setProperties?

JKMarkowski: No, there is no point in adding classes in HostBinding.

Issue #805, opened 2019-05-10
add time formatting to time picker


MattL75:
Follow pattern set by date picker. Use the service provider methodology to allow developers to format the time as they want. Use the time object that is already defined.

Possibly connected #866

PR #894, opened 2019-06-11
fix: Add timepicker formating following datepicker

JKMarkowski:
fixes https://github.com/SAP/fundamental-ngx/issues/805
I added timepicker component formating which is very similar to datepicker's formating, there are default parse/format functions which can be overwritten by providinf other service.
I added 1 example on documentation and 6 unit tests.


JKMarkowski:
I removed range delimiter for time parser/format function, the example is also changed to be DDhDDmDDs .

docs/app/documentation/component-docs/time-picker/examples/time-picker-other-delimiter-example.component.ts
```diff
@@ -0,0 +1,148 @@
+import { Component, Injectable } from '@angular/core';
+import { TimeFormatParser } from '../../../../../../library/src/lib/time-picker/format/time-parser';
+import { TimeObject } from '../../../../../../library/src/lib/time/time-object';
+
+
+@Injectable()
+export class TimeFormatDashes extends TimeFormatParser {
```
MattL75: Rename the class since this isn't dashes anymore - maybe something like 'CustomTimeFormat'?

Otherwise this PR looks pretty good

Issue #784, opened 2019-05-07
Split public_api


MattL75:
Our public_api is growing too large, and it can be difficult to isolate which file belongs to which component.

We should have a public_api file in each component file, and then export them all together in the main public_api (slight research required on this).

PR #907, opened 2019-06-12
fix: split public_api

JKMarkowski:
fixes https://github.com/SAP/fundamental-ngx/issues/784
I added index.d.ts file on every single component which exports all required file. 
--


MattL75:
Good stuff @JKMarkowski! 

I think the CI is failing because of the barrel imports. If you import explicitly for example `export * from './lib/popover/index.d.ts'` it might fix it. We've had issues with barrel imports in libraries in another project before.

JKMarkowski:
I found the information that .d.ts files are already  compiled and ready to be used for javascript, we shouldn't name files like that directly. After compilation in dist directory all exported files with .ts extension are being changed to .d.ts. Changing naming did job.

PR #933, opened 2019-06-14
fix: fd-tree throwing error when used without header

migg24:
No issue, found the issue while using fd-tree.

When using the `<fd-tree>` without a `[header]` this part of the code throws errors because `this.header` is undefined, practically making `[header]` a mandatory attribute even if you do not need headers (currently you need to provide an empty array).

As a side note the documentation for <fd-tree> I could not find on https://sap.github.io/fundamental-ngx/home.


mikerodonnell89:
While this is currently a prototype, undocumented component at the moment, we're liking this PR. Thanks @migg24

Issue #888, opened 2019-06-10
Cleanup Badge/Label API


MattL75:
#### Is this a bug, enhancement, or feature request?
Enhancement

#### Briefly describe your proposal.

- [x] Replace @Inject annotation by just private
- [x] Use HostBinding, possibly remove extension of AbstractFdNgxClass if possible
- [x] Do not use same template for both components. Can potentially define the template in the ts file since it will be very small
- [x] Rename docs to Status Indicators

PR #937, opened 2019-06-17
fix: remove isComponent

JKMarkowski:
#### Please provide a link to the associated issue.
https://github.com/SAP/fundamental-ngx/issues/771
#### Please provide a brief summary of this pull request.
I added popover-dropdown component and removed isDateTimePicker from time component,
changed input flag on calendar component to make it more clear.
#### If this is a new feature, have you updated the documentation?
I changed documentation due to dropdown changes.

PR #946, opened 2019-06-17
fix: cleaned up badge/label component

SaiShan:
#888 

Completed 4/5 tasks. As for the 5th one, Find alternative way of dealing with the isStatusLabel input that doesn't seem so hacky, could not find a way without causing breaking changes.


mikerodonnell89:
Maybe a separate component `fd-status-label` is warranted, like how @JKMarkowski refactored the isDropdown into its on fd-dropdown component in this pull request: https://github.com/SAP/fundamental-ngx/pull/937/files

MattL75:
@mikerodonnell89 Tbh these could all be directives but whatever

mikerodonnell89:
@MattL75 agreed

Issue #892, opened 2019-06-10
Add navigation mode for tabs


MattL75:
#### Is this a bug, enhancement, or feature request?
Enhancement
#### Briefly describe your proposal.
Currently, our tabs cannot be used with a router-outlet. In the documentation we are currently using the css classes from fundamental-ui to accomplish this pattern, not the fd-ngx component.

We should have a special mode for the tabs to be used inside a `<nav>` element with `<a>` tags. The solution should look something like the following.

```html
<nav fd-tab-nav>
    <a fd-tab-link
         [routerLink]="link.path"
         routerLinkActive #rla="routerLinkActive"
         [active]="rla.isActive">
             Link
    </a>
</nav>
```

Note that the routerLink, routerLinkActive and associated stuff is not to be handled by fd-ngx components. The only thing we should handle is an `[active]` input which controls the applied class.

JKMarkowski:
just FYI, I'm on it

PR #953, opened 2019-06-19
fix: Refactor Tabs

JKMarkowski:
https://github.com/SAP/fundamental-ngx/issues/892
There was needed big refactor for this issue, including adding some directives/components, but now tab component is fully customizable and generic. I added also some unit tests. 
Yes, documentation has been updated to show new way to use this component.


MattL75:
This increases the required markup for tabs significantly. What I had in mind was to keep the directives shown in the issue and the current component as separate. Doing so would keep the markup for both methods of using tabs very simple, and at the same time keep each component/directive simple in their implementation. The ultimate goal of the issue was just to enable using the tabs with a router-outlet. What do you think?

JKMarkowski:
Right now there is possibility to put `[routerLink]` to `<a>` element with `fd-tab-link` directive. To make this fully working and also keep possibility to put normal content like it was before, I had to add additional markups. But yes for using router-outlet it wasn't really needed. I could just add 2 other little directives which would be used only to show nav without panel.

MattL75:
I think it might be simpler to have the two directives and make the two different uses completely separated. This will make it easier for users to consume tabs in both uses, while also making easier for us to maintain these since they will both be simpler individually. @mikerodonnell89 @droshev thoughts?

mikerodonnell89:
In my opinion this is an example of extra markup being a good thing.  The extra flexibility here is crucial.  And having proper HTML semantics (`<nav>, <ul>, <li>`) is nice for accessibility (though I guess that could be improved upon with the previous more strict approach).  Having two directives with different uses would be confusing to someone approaching the tab for the first time.

MattL75:
@mikerodonnell89 I disagree. Having two directives is not confusing at all if properly documented (see Material). In terms of accessibility, including extra markup is not a good thing. See the [guidelines](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel), tabs aren't strictly paired with navigation and lists. The extra markup is **very** significant here, we're talking almost twice as much. Maintaining a component with two different modes is also not very practical, with the code often becoming bloated over time.

JKMarkowski:
So @MattL75 , @mikerodonnell89 , @droshev  what is the conclusion, do we keep it as it is on this PR, or should I create new one with 2 additional directives?

mikerodonnell89:
@JKMarkowski consensus was to go the 2 directives route

PR #970, opened 2019-06-21
fix: only handle input change on keyup.enter or blur for datepicker/dateti…

### LINKED ISSUE #967: [Datepicker] manual date entering is challenging

bug / bad ux

It requires extra effort to input dates manually to datepicker

latest

manual2

mikerodonnell89:
…mepicker

fixes #967 

Easy fix for this issue.  I started working on a more complex fix, where we trigger validation checks at each keystroke (for the red outline) but not attempt to update the model until the user presses enter or blurs the input - this would require separating the validation and model updating logic.  However after reading a lot of UX stuff on stackoverflow, it sounds like it is best to handle validation only on submit or blur.

Another thing is if the user just enters "1" in the input and presses enter or blurs, the date input will automatically set the selected date to 1/1/2001, if the user sets the input to "1/1/20" in the input the date will be set to 1/1/2020... is this something we want?  I could also work on having the date input check if the formatting of the input matches the dateAdapter format function

### DIFF
--- a/docs/app/documentation/component-docs/date-picker/date-picker-docs.component.html
+++ b/docs/app/documentation/component-docs/date-picker/date-picker-docs.component.html
@@ -9,6 +9,10 @@
 <description>
   The default format for the input is MM/DD/YYYY. See the 'Formatting' section for details on how to change this.
 </description>
+<description>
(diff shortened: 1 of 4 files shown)


JKMarkowski:
Look good for me, it's the best way we can handle problems with date changes on keyups.

PR #976, opened 2019-06-25
feat: Add navigation mode for tabs

JKMarkowski:
fixes https://github.com/SAP/fundamental-ngx/issues/892
There are added 3 instead of 2 directives, I needed to add fd-tab-item also (it's not mandatory to include it, but then we lose one fior fundamentalsi class). I also included key handler to support arrows and enter/space keys for list. The whole functionallity for keydowns is in parent directive, that's why I had to wrap nav inside.
There is one additional example of usage.


docs/app/documentation/component-docs/tabs/tabs-docs.component.html
```diff
@@ -25,6 +25,17 @@ <h2>Complex Tab Titles</h2>
 </component-example>
 <code-example [code]="complexH" [language]="'HTML'"></code-example>
 
+<separator></separator>
+<h2>Tab Navigation Mode</h2>
+<description>
+    There is possibility to change tab mode to navigation, it means that there is not possibility to put content, but there is possibility to add some
```
MattL75: It is possible to use the tabs in navigation mode. While there is no option to provide some content in this mode, it does allow the use of a router-outlet with <code>[routerLink]</code> or <code>[href]</code> attributes.

docs/app/documentation/component-docs/tabs/examples/tabs-navigation-mode-example.component.html
```diff
@@ -0,0 +1,30 @@
+<fd-tab-list>
```
MattL75: Is there a reason we need fd-tab-list as a wrapper component?

docs/app/documentation/component-docs/tabs/tabs-docs.component.html
mikerodonnell89: Also I think we could use a more detailed description of a practical use case here

JKMarkowski: Docs changed.

docs/app/documentation/component-docs/tabs/examples/tabs-navigation-mode-example.component.html
JKMarkowski: There was a reason why we needed it, because of keyboard functionality, that was provided by wrapper. Right now It's changed and external service keeps this functionality, so wrapping is not needed anymore. :)

library/src/lib/tabs/tab-link/tab-link.directive.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    /** Whether the link is active */
+    @HostBinding('attr.aria-selected')
+    get _active(): boolean {
+        return this.active;
+    }
+
+    /** Only visual Thing, to disable routerLink does not respect preventDefault/stopPropagation */
+    @HostBinding('attr.aria-disabled')
+    get _disabled(): boolean {
```
MattL75: Can you try to merge the inputs and hostbindings here? In Angular 2 this wasn't possible, but I think it was a bug which might be fixed now.

```typescript
@Input()
@HostBinding('attr.aria-selected')
active: boolean;
```

library/src/lib/tabs/tab-list.component.html
```diff
@@ -1,17 +1,15 @@
...
+        *ngFor="let tab of panelTabs.toArray(); let i = index;">
         <a #tabLink
-           role="tab"
-           class="fd-tabs__link"
+           fd-tab-link
+           [disabled]="tab.disabled"
+           [active]="tab.expanded"
            [attr.tabindex]="tab.disabled ? -1 : 0"
            [attr.aria-controls]="tab.id"
-           [attr.aria-disabled]="tab.disabled"
```
MattL75: The aria stuff serves as more than just a visual thing here, it's for accessibility. Though I haven't really tested this component with a screenreader so not sure how much of an impact they have

JKMarkowski:
Ok, bug fixed. Also I merged hostbindins with inputs and it works well. I removed `aria-disabled` because right now we have `nav-link` directive which automatically adds `aria-disabled` and `aria-selected` attributes.

MattL75:
Need a better way to showcase the tabs with navigation, not very nice to the user if we go to /home when they test the click... Maybe setup some router outlet inside the example?

JKMarkowski:
@MattL75 Sure, I was thinking about it. I will prepare something :)

JKMarkowski:
Example with router-outlet added

Issue #890, opened 2019-06-10
Icon component cleanup


MattL75:
#### Is this a bug, enhancement, or feature request?
Enhancement/Code refactor
#### Briefly describe your proposal.
Currently, the icon is a directive but is being used as a component. We should make it a component outright. We could also consider getting the text content of the host element and loading the icon based on that as opposed to the current method of using an input called glyph.

PR #979, opened 2019-06-26
fix: Change icon to component, add tests

JKMarkowski:
fixes https://github.com/SAP/fundamental-ngx/issues/890
I changed icon directive to component and added some tests. Also tried to add support for adding icon name inside component brackets, but it's hard to deal with it on any changes. Other libraries got different system, where font-style interpretate whole words as a icons.
-

### DIFF
--- a/library/src/lib/icon/icon.component.ts
+++ b/library/src/lib/icon/icon.component.ts
@@ -14,15 +14,13 @@ const BASE_ICON_CLASS = 'sap-icon';
 const PREFIX_ICON_CLASS = BASE_ICON_CLASS + '--';
 
 /**
- * The Component that represents an icon.
+ * The component that represents an icon.
  *
  * ```html
  * <fd-icon [glyph]="cart-approval" [size]="'l'"></fd-icon>
  * ```
  */
 @Component({
-    // TODO to be discussed
-    // tslint:disable-next-line:directive-selector
     selector: 'fd-icon',
     template: ``,
     host: {
@@ -32,7 +30,9 @@ const PREFIX_ICON_CLASS = BASE_ICON_CLASS + '--';
 })
 export class IconComponent extends AbstractFdNgxClass {
 
-    /** The glyph name  */
+    /** The icon name to display. See the icon page for the list of icons
+     * here: https://sap.github.io/fundamental-ngx/icon
+     * */
     @Input() glyph;
 
     /** 
--- a/library/src/lib/icon/icon.component.ts
+++ b/library/src/lib/icon/icon.component.ts
@@ -1,4 +1,4 @@
-import { Directive, Input, ElementRef, Inject } from '@angular/core';
+import { Directive, Input, ElementRef, Inject, AfterContentInit, AfterViewInit, Component } from '@angular/core';
 import { AbstractFdNgxClass } from '../utils/abstract-fd-ngx-class';
 
 /** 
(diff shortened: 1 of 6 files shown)


library/src/lib/icon/icon.component.ts
```diff
@@ -14,22 +14,27 @@ const BASE_ICON_CLASS = 'sap-icon';
...
  */
-@Directive({
+@Component({
     // TODO to be discussed
     // tslint:disable-next-line:directive-selector
     selector: 'fd-icon',
+    template: ``,
     host: {
         role: 'presentation'
     }
```
MattL75: Need to set view encapsulation to none since this is a component now

library/src/lib/icon/icon.component.ts
```diff
@@ -1,4 +1,4 @@
-import { Directive, Input, ElementRef, Inject } from '@angular/core';
+import { Directive, Input, ElementRef, Inject, AfterContentInit, AfterViewInit, Component } from '@angular/core';
```
mikerodonnell89: some unused inputs up here

library/src/lib/icon/icon.component.ts
mikerodonnell89: What is the motivation for making the icon a component rather than using it as a directive like we do for other widgets that don't need a template definition?  Such as the new `<span fd-badge>`, etc.

MattL75: Eventually I'm planning on having a Material-like experience where you put the icon text inside the icon's content.

library/src/lib/icon/icon.component.ts
```diff
@@ -14,23 +14,29 @@ const BASE_ICON_CLASS = 'sap-icon';
...
- * The directive that represents an icon. 
+ * The Component that represents an icon.
  *
  * ```html
  * <fd-icon [glyph]="cart-approval" [size]="'l'"></fd-icon>
- * ```
+ // * ```
+ // * Or
+ // * ```html
+ // * <fd-icon [size]="'l'">cart-approval</fd-icon>
```
MattL75: This actually displays in the docs sortof. I'm guessing you commented it out because it's not currently working? We should envision implementing this in the future

image

JKMarkowski:
I removed commented template. In future we should have possibility to add icons like that `<icon>iconName</icon>`, but it would require from us having own font-style which will interpret word as a icon. Actually I managed to include it by reading `innerHTML`, but it was impossible to detect if any icon has been changed, or it required huge workaround.

MattL75:
@JKMarkowski What if you read from something like `(this.elRef.nativeElement as HTMLElement).textContent.trim()` or something? And then bind that to the glyph input? Haven't tested or tried this though

JKMarkowski:
@MattL75 I can check if `ngDoCheck()` will work, but as I know it's not very good for performance.

JKMarkowski:
@MattL75 For now, unfortunately I can't find any good solution for it.

library/src/lib/icon/icon.component.ts
```diff
@@ -14,23 +14,25 @@ const BASE_ICON_CLASS = 'sap-icon';
...
+ * The Component that represents an icon.
  *
  * ```html
  * <fd-icon [glyph]="cart-approval" [size]="'l'"></fd-icon>
  * ```
  */
-@Directive({
+@Component({
     // TODO to be discussed
     // tslint:disable-next-line:directive-selector
```
MattL75: Can remove this tslint disable

library/src/lib/icon/icon.component.ts
```diff
@@ -14,23 +14,25 @@ const BASE_ICON_CLASS = 'sap-icon';
 const PREFIX_ICON_CLASS = BASE_ICON_CLASS + '--';
 
 /**
- * The directive that represents an icon. 
+ * The Component that represents an icon.
```
MattL75: remove capital letter

library/src/lib/icon/icon.component.ts
```diff
@@ -14,23 +14,25 @@ const BASE_ICON_CLASS = 'sap-icon';
...
-    }
+    },
+    encapsulation: ViewEncapsulation.None
 })
-export class IconDirective extends AbstractFdNgxClass {
-    
-    /** The glyph name */
+export class IconComponent extends AbstractFdNgxClass {
+
+    /** The glyph name  */
```
MattL75: Should be more descriptive, maybe add a link to the list of glyphs

Issue #620, opened 2019-03-12
convert inline-help component to use popover


droshev:
#### Is this a bug, enhancement, or feature request?
enhancement

help’s popover content should do it’s best (with help from popperjs) to position itself out of the way of the input itself.

consumers that want an alternate icon or no icon at all for help should be directed to use the base icon and base popover components on their own.

Basically, the inline-help should make use of the Popover Component instead of its own inline-popover mechanic.

MattL75:
@mikerodonnell89 No rush on this one, some significant changes coming to popover soon

PR #986, opened 2019-06-27
fix: make inline help component using popover component

JKMarkowski:
fixes https://github.com/SAP/fundamental-ngx/issues/620
I made inline help component using popover component inside, the popover control contains only the help icon. I adde dalso possibility to change trigger event. Had to put some styles, cause popover for example already got some borders/positions. I added also some styles which simulates the fact that the inline help component is always hovered, cause right now popover is controlling view.
I added one example.

### DIFF
--- a/docs/app/documentation/component-docs/inline-help/inline-help-header/inline-help-header.component.html
+++ b/docs/app/documentation/component-docs/inline-help/inline-help-header/inline-help-header.component.html
@@ -1,6 +1,6 @@
 <header>Inline Help</header>
 <description>Inline help is used to display help text in a popover, often inline with headers, body text and form
-    labels. Also it uses popover component</description>
+    labels. It also uses the fd-popover component and supports much of its functionality.</description>
 <import module="InlineHelpModule" path="fundamental-ngx"></import>
 
 <fd-header-tabs></fd-header-tabs>
--- a/docs/app/documentation/component-docs/inline-help/inline-help-header/inline-help-header.component.html
+++ b/docs/app/documentation/component-docs/inline-help/inline-help-header/inline-help-header.component.html
@@ -1,6 +1,6 @@
 <header>Inline Help</header>
 <description>Inline help is used to display help text in a popover, often inline with headers, body text and form
-    labels.</description>
+    labels. Also it uses popover component</description>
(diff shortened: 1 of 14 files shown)


library/src/lib/inline-help/inline-help.component.ts
```diff
@@ -1,29 +1,35 @@
...
+ * <fd-inline-help [placement]="'bottom-left'">
  *      Lorem ipsum dolor sit amet, consectetur adipiscing.
  * </fd-inline-help>
  * ```
  */
 @Component({
     selector: 'fd-inline-help',
     templateUrl: './inline-help.component.html',
-    host: {
-        class: 'fd-inline-help',
```
mikerodonnell89: A very minor detail, but because this class is no longer applied, the text of the inline help is grey and not black as designed in fundamental-styles

JKMarkowski: Generally this class cannot be applied to parent, because it has some css styling that will break component inside popover. I had to interfere styling, what is not best thing I could do, because of possible future problems with following any fiori styles changes. What do you think about it @MattL75 ?

docs/app/documentation/component-docs/inline-help/inline-help-header/inline-help-header.component.html
```diff
@@ -1,6 +1,6 @@
 <header>Inline Help</header>
 <description>Inline help is used to display help text in a popover, often inline with headers, body text and form
-    labels.</description>
+    labels. Also it uses popover component</description>
```
MattL75: "It also uses the fd-popover component and supports much of its functionality."

library/src/lib/inline-help/inline-help.component.ts
MattL75: I don't think there's a clear way around this. We can't even go and remove it from Fundamental Styles because the component needs to work without popper.js. I suppose one option would be to include a 'hidden' class in Styles which wouldn't incorporate the style-popovers and would let us develop our own. As a workaround, maybe you can add the same text color variable to your custom scss?

JKMarkowski:
Added text color change based on variable. Corrected documentation description due to your proposition.

mikerodonnell89:
The element has lost its fade-in/fade-out as well

Issue #267, opened 2018-10-15
Menu sub-items should be navigable with the up/down arrow keys


mikerodonnell89:
#### Is this a bug, enhancement, or feature request?

Enhancement

#### Briefly describe your proposal.

As per: https://www.w3.org/WAI/tutorials/menus/application/

Top level menus should be navigable with the tab key, and sub items should be navigable with the arrow keys.

For our menu component we should be able to navigate inside the component with the arrows.
There should be implemented some methods in this component. to easy communicate from outside. 
For example: focusFirstItem, also some event thrown when element is clicked/focused. It will simplify adding keyboard support to future/existing components.

droshev:
@JKMarkowski will update this ticket reflecting our last discussion

PR #1005, opened 2019-07-05
fix: add keyboard support to menu component

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/267
I added keyboard support to menu component, right now we can switch items with arrowup/arrowdown, also there is way to disable handling and provide custom functionality, when escape focus.
-


library/src/lib/menu/menu-group.component.ts
```diff
@@ -8,12 +9,10 @@ import { Component, HostBinding, ViewEncapsulation } from '@angular/core';
     templateUrl: './menu-group.component.html',
     encapsulation: ViewEncapsulation.None,
     host: {
-        '[class.fd-has-display-block]': 'true'
+        'class': 'fd-has-display-block fd-menu__group',
```
MattL75: The problem with this (if I recall correctly) is that if a user adds a class, it causes some problems. That's why we used the class.[classname] thing initially

library/src/lib/menu/menu-item.directive.ts
```diff
@@ -1,41 +1,83 @@
...
     }
 
+    public focus() {
+        if (this.getChildrenElements() && this.getChildrenElements()[0]) {
+            this.getChildrenElements()[0].focus();
+        }
+    }
+
+    @HostListener('keydown', ['$event'])
+    keyDown(event: KeyboardEvent): void {
```
MattL75: Is there a reason why these (+ subject) aren't just events? Technically events are observables too

MattL75: just seems a bit convoluted to have this stuff

library/src/lib/menu/menu-item.directive.ts
```diff
@@ -1,41 +1,83 @@
...
+
+    public click() {
+        if (this.getChildrenElements() && this.getChildrenElements()[0]) {
+            this.getChildrenElements()[0].click();
+        }
+    }
+
     /** @hidden */
     ngOnChanges() {
         this.ngAfterViewInit();
```
MattL75: prefer to avoid calling life cycle hooks explicitly

library/src/lib/menu/menu-item.directive.ts
```diff
@@ -1,41 +1,83 @@
...
     }
+
+    public isChildElementAnchor(): boolean {
+        return this.getChildrenElements() &&
+        this.getChildrenElements()[0] &&
+        this.getChildrenElements()[0].tagName === 'A'
```
MattL75: As mentioned above, this is super not flexible. I know it was there before, but can you think of a better way we could handle this? Not a huge deal though.

library/src/lib/menu/menu-list.directive.ts
```diff
@@ -1,15 +1,40 @@
...
 /**
  * The directive that represents a listing structure of the menu.
  */
 @Directive({
     // TODO to be discussed
     // tslint:disable-next-line:directive-selector
-    selector: '[fd-menu-list]'
+    selector: '[fd-menu-list]',
+    host: {
+        'class': 'fd-menu__list'
```
MattL75: prefer hostbinding

library/src/lib/menu/menu-list.directive.ts
```diff
@@ -1,15 +1,40 @@
...
+    }
 })
-export class MenuListDirective {
+export class MenuListDirective implements AfterContentInit, OnDestroy {
     /** @hidden */
-    @HostBinding('class')
-    elementClass = 'fd-menu__list';
+    @ContentChildren(MenuItemDirective) menuItems: QueryList<MenuItemDirective>;
```
MattL75: next line

library/src/lib/menu/menu-list.directive.ts
```diff
@@ -1,15 +1,40 @@
...
+    private onMenuItemsChangeSubscription: Subscription;
+
+    /** @hidden */
+    public ngAfterContentInit(): void {
+        this.onMenuItemsChangeSubscription = this.menuItems.changes.subscribe(() => this.listRefresh.emit());
+    }
+
+    /** @hidden */
+    public ngOnDestroy(): void {
+        this.onMenuItemsChangeSubscription.unsubscribe();
```
MattL75: need to add a null check for the subscription

library/src/lib/menu/menu-list.directive.ts
```diff
@@ -1,15 +1,40 @@
...
-    elementClass = 'fd-menu__list';
+    @ContentChildren(MenuItemDirective) menuItems: QueryList<MenuItemDirective>;
+
+    /** @hidden */
+    public get menuItemsWithAnchors(): MenuItemDirective[] {
+        return this.menuItems.filter(item => item.isChildElementAnchor());
+    }
+
+    /** @hidden */
+    @Output() listRefresh = new EventEmitter();
```
MattL75: Could you add a docs description to these? Even if not shown in the docs.

library/src/lib/menu/menu-list.directive.ts
```diff
@@ -1,15 +1,40 @@
...
 })
-export class MenuListDirective {
+export class MenuListDirective implements AfterContentInit, OnDestroy {
     /** @hidden */
-    @HostBinding('class')
-    elementClass = 'fd-menu__list';
+    @ContentChildren(MenuItemDirective) menuItems: QueryList<MenuItemDirective>;
+
+    /** @hidden */
+    public get menuItemsWithAnchors(): MenuItemDirective[] {
```
MattL75: Put the Output and the subscription reference before this method

library/src/lib/menu/menu.component.ts
```diff
@@ -1,16 +1,147 @@
...
- * The component that represents a menu.
+ * The component that represents a menu. Provides some keyboard event default handler.
  */
 @Component({
     selector: 'fd-menu',
     templateUrl: './menu.component.html',
-    encapsulation: ViewEncapsulation.None
+    encapsulation: ViewEncapsulation.None,
+    host: {
+        'class': 'fd-menu'
```
MattL75: prefer hostbinding

library/src/lib/menu/menu.component.ts
```diff
@@ -1,16 +1,147 @@
...
+    }
 })
-export class MenuComponent {
+export class MenuComponent implements AfterContentInit, OnDestroy {
 
     /** @hidden */
-    @HostBinding('class.fd-menu')
-    fdMenuClass: boolean = true;
+    @ContentChildren(MenuGroupComponent) menuGroup: QueryList<MenuGroupComponent>;
```
MattL75: next line

library/src/lib/menu/menu.component.ts
```diff
@@ -1,16 +1,147 @@
...
+
+    /** @hidden */
+    @ContentChildren(MenuListDirective) menuList: QueryList<MenuListDirective>;
```
MattL75: next line

library/src/lib/menu/menu.component.ts
```diff
@@ -1,16 +1,147 @@
...
+
+    /** Event thrown always, when item link is clicked */
+    @Output() itemClicked = new EventEmitter<number>();
```
MattL75: next line

MattL75: also explicit type please
itemClicked: EventEmitter<number> = new EventEmitter<number>()

MattL75: Also please make all Outputs **readonly**

library/src/lib/menu/menu.component.ts
```diff
@@ -1,16 +1,147 @@
...
+
+    /** Whether user wants to remove default keyboard handling */
+    @Input() disableKeydownHandling: boolean = false;
```
MattL75: next line

library/src/lib/menu/menu.component.ts
```diff
@@ -1,16 +1,147 @@
...
 
     /** @hidden */
-    @HostBinding('class.fd-menu')
-    fdMenuClass: boolean = true;
+    @ContentChildren(MenuGroupComponent) menuGroup: QueryList<MenuGroupComponent>;
+
+    /** @hidden */
+    @ContentChildren(MenuListDirective) menuList: QueryList<MenuListDirective>;
+
+    /** Event thrown always, when item link is clicked */
```
MattL75: Event emitted when an item link is clicked.

also specify what the return is

library/src/lib/menu/menu.component.ts
```diff
@@ -1,16 +1,147 @@
...
+    @Output() itemClicked = new EventEmitter<number>();
+
+    /** Whether user wants to remove default keyboard handling */
+    @Input() disableKeydownHandling: boolean = false;
+
+    /** Whether user wants to define own function, that is supposed to be called, when focus escape before list*/
+    @Input() focusEscapeBeforeList: Function;
```
MattL75: next line... I'll skip writing next line from here on out

library/src/lib/menu/menu.component.ts
```diff
@@ -1,16 +1,147 @@
...
+    /** @hidden */
+    @ContentChildren(MenuListDirective) menuList: QueryList<MenuListDirective>;
+
+    /** Event thrown always, when item link is clicked */
+    @Output() itemClicked = new EventEmitter<number>();
+
+    /** Whether user wants to remove default keyboard handling */
+    @Input() disableKeydownHandling: boolean = false;
+
+    /** Whether user wants to define own function, that is supposed to be called, when focus escape before list*/
```
MattL75: Not a very clear description, tbh I'm not sure myself what this is for from looking just at the description?

library/src/lib/menu/menu.component.ts
```diff
@@ -1,16 +1,147 @@
...
+    /** Whether user wants to define own function, that is supposed to be called, when focus escape after list*/
+    @Input() focusEscapeAfterList: Function;
+
+    private onItemKeyDownSubscription: Subscription[];
+    private onItemClickSubscription: Subscription[];
+
+    /** @hidden */
+    public ngAfterContentInit(): void {
+        this.refreshList();
+        this.menuList.forEach(list => list.listRefresh.subscribe(() => this.refreshList()));
```
MattL75: where are these subscriptions unsubscribed?

library/src/lib/menu/menu.component.ts
```diff
@@ -1,16 +1,147 @@
...
+        this.menuGroup.forEach(list => list.menuList.listRefresh.subscribe(() => this.refreshList()));
+    }
+
+    /** @hidden */
+    public ngOnDestroy(): void {
+        this.onItemKeyDownSubscription = [];
```
MattL75: For this case, I think it's better to have a subject called when the component is destroyed, and have every dependent subscription takeUntil(destroy).

see https://github.com/SAP/fundamental-ngx/blob/feat/dropdown/library/src/lib/select/select.component.ts#L120

and https://github.com/SAP/fundamental-ngx/blob/feat/dropdown/library/src/lib/select/select.component.ts#L71

and https://github.com/SAP/fundamental-ngx/blob/feat/dropdown/library/src/lib/select/select.component.ts#L112

MattL75:
There is also a one-off error with the logic for the keyboard navigation. It always skips the last element.

library/src/lib/menu/menu-item.directive.ts
JKMarkowski: Hmm... The only thing that come up to my mind, is creating new directive which would be named fd-menu-link. Then we would be able to check if there is any link inside menu item.

JKMarkowski:
It's not really bug, I can say that it's even the feature. I was thinking about it and decided, to allow focusing only item which has anchor inside. We should talk about it and about some other stuff related to anchors in this component.

library/src/lib/menu/menu-item.directive.ts
```diff
@@ -8,34 +8,22 @@ import { AfterViewInit, Directive, ElementRef, OnChanges } from '@angular/core';
...
-            // if the menu item does not contain a child element, apply 'fd-menu__item' class to the fd-menu-item component
-            this.itemEl &&
-            this.itemEl.nativeElement &&
-            this.itemEl.nativeElement.children &&
-            this.itemEl.nativeElement.children.length === 0
-        ) {
-            this.itemEl.nativeElement.classList.add('fd-menu__item');
-        }
+    ngAfterViewInit(): void {
+        this.itemEl.nativeElement.classList.add('fd-menu__item');
```
MattL75: What's the point of adding this afterViewInit?

docs/app/documentation/component-docs/menu/examples/menu-keyboard-support-example.component.ts
```diff
@@ -0,0 +1,38 @@
...
+    public onClick() {
+        alert('Element clicked');
+    }
+
+    public handleKeyDown(event: KeyboardEvent, index: number): void {
+        this.menuKeyboardService.keyDownHandler(event, index, this.menuItems.toArray());
+    }
+
+    ngAfterViewInit(): void {
+        this.menuKeyboardService.focusEscapeAfterList = this.escapeAfterListFunction;
```
MattL75: Not sure I like this approach to the keyboard navigation, it's a bit confusing for developers to use I think. Maybe we're better off without it completely and just implementing it for individual components, I'm not sure.

JKMarkowski: I'm sure these functions would be easy for developers to use. It's completely optional, so devs can handle events by themselves. I would rather keep it, this approach will greatly reduce amount of code in individual components.

library/src/lib/menu/menu-item.directive.ts
JKMarkowski: Sure, no point for keepnig it, removed.

docs/app/documentation/component-docs/menu/examples/menu-keyboard-support-example.component.ts
JKMarkowski: @MattL75

docs/app/documentation/component-docs/menu/menu-docs.component.html
```diff
@@ -15,3 +15,18 @@ <h2>Menu with Group</h2>
   </p>
 </component-example>
 <code-example [code]="menuGroupHtml" [language]="'HTML'"></code-example>
+
+<separator></separator>
+
+<h2>Menu with keyboard support</h2>
```
MattL75: Keyboard Support

docs/app/documentation/component-docs/menu/menu-docs.component.html
```diff
@@ -15,3 +15,18 @@ <h2>Menu with Group</h2>
...
+<description>
+    Menu keyboard support can be added on every component, which is using <code>MenuComponent</code>. It simulates
```
MattL75: Keyboard support can be added on every component which uses the `<code>MenuComponent</code>`.

docs/app/documentation/component-docs/menu/menu-docs.component.html
```diff
@@ -15,3 +15,18 @@ <h2>Menu with Group</h2>
...
+    click event on <code>Enter</code> or <code>Space</code> key down. Also it's possible to overwrite functions, which
+    are responsible for situation, when focus escape before <code>focusEscapeBeforeList</code> or after
+    <code>focusEscapeAfterList</code> list. To try it, use arrow up/down.
```
MattL75: It is also possible to override functions for a variety of situations. For example, when the focus should escape the element. Use the up and down arrows to test the component below.

docs/app/documentation/utilities/api-files.ts
```diff
@@ -1,10 +1,10 @@
-
 /**
  * Files to display in the API tab of each component.
  * Names should be without hyphens, and capitalized where hyphens occur normally.
  * Include the suffix i.e. Directive or Component.
  * Names are sorted in the ApiComponent so order does not matter.
  */
+
```
MattL75: remove extra line

JKMarkowski:
It's about the fact, that we focus `<a>` element instead of `<li>` element. I also don't know what we should do in this situation

JKMarkowski:
@MattL75 All issues are solved, could you please take a look today if everything is ok?

PR #1022, opened 2019-07-11
[WIP] month-view implementation without the tests

InnaAtanasova:


library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.ts
```diff
@@ -1,17 +1,104 @@
...
+        const tempDate: Date = new Date();
+        return tempDate.getMonth() + 1;
     }
 
-    ngOnInit() {
+    /** @hidden */
+    selectMonth(selectedMonth: number, event?: MouseEvent): void {
+        if (event) {
+            event.stopPropagation();
+        }
```
JKMarkowski: We should think about it, I also used stopPropagation of click on dayView, because clicking inside component caused closing the popover.

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.ts
```diff
@@ -1,17 +1,104 @@
...
+                } else {
+                    newFocusedMonthId = '#' + this.id + '-fd-month-' + (month + 1);
+                }
+                break;
+            }
+            case 'Tab': {
+                if (!event.shiftKey && !this.allowFocusEscape) {
+                    event.preventDefault();
+                    this.focusElement('#arrowLeft');
+                }
```
JKMarkowski: Maybe it would be better to pass some function there, instead of passing flag. Default function shoould focus on arrowLeft, but it could be changed from outside.

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.html
```diff
@@ -1,3 +1,22 @@
-<p>
-  calendar2-month-view works!
-</p>
+<div class="fd-calendar__months">
+    <ul class="fd-calendar__list">
+        <li
+            class="fd-calendar__item"
+            *ngFor="let month of monthNames; let i = index"
+            [ngClass]="
+                (i + 1 === currentMonth ? ' fd-calendar__item--current' : '') +
```
MattL75: Avoid magic numbers

also, the better syntax for things like this would be 
`<some-element [ngClass]="{'firstclass': true, 'secondclass': true, 'thirdclass': false}">...</some-element>`

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.html
```diff
@@ -1,3 +1,22 @@
...
+                (i + 1 === monthSelected ? ' is-selected' : '')
+            "
+            (click)="selectMonth(i, $event)"
+        >
+            <span
```
MattL75: is this standard from fd? this is bad for a11y, a span with role button should just be a button.

also this needs an aria-label of the month name localized

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.ts
```diff
@@ -1,17 +1,104 @@
...
 @Component({
     selector: 'fd-calendar2-month-view',
     templateUrl: './calendar2-month-view.component.html',
     styleUrls: ['./calendar2-month-view.component.scss'],
     encapsulation: ViewEncapsulation.None
 })
-export class Calendar2MonthViewComponent implements OnInit {
+export class Calendar2MonthViewComponent {
+    /** @hidden The id of the calendar passed from the parent component. */
+    @Input() id: string;
```
MattL75: this should also modify the [attr.id] of the host element, concatenated with like 'month-view' or something

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.ts
```diff
@@ -1,17 +1,104 @@
...
 
 @Component({
     selector: 'fd-calendar2-month-view',
     templateUrl: './calendar2-month-view.component.html',
     styleUrls: ['./calendar2-month-view.component.scss'],
     encapsulation: ViewEncapsulation.None
 })
-export class Calendar2MonthViewComponent implements OnInit {
+export class Calendar2MonthViewComponent {
+    /** @hidden The id of the calendar passed from the parent component. */
```
MattL75: you can remove all the @hidden tags since we won't be displaying this component page to the users at all

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.ts
```diff
@@ -1,17 +1,104 @@
-import { Component, OnInit, ViewEncapsulation } from '@angular/core';
+import { Component, Input, Output, ViewEncapsulation, EventEmitter, ElementRef } from '@angular/core';
 
 @Component({
```
MattL75: needs a documentation at the top here

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.ts
```diff
@@ -1,17 +1,104 @@
...
     encapsulation: ViewEncapsulation.None
 })
-export class Calendar2MonthViewComponent implements OnInit {
+export class Calendar2MonthViewComponent {
+    /** @hidden The id of the calendar passed from the parent component. */
+    @Input() id: string;
 
-    constructor() {
+    /** @hidden A list of month names */
+    @Input() monthNames: string[];
```
MattL75: You should do 

```
@Input()
variable: type;
```

everywhere, but some places I see `Input() variable: type;`

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.ts
```diff
@@ -1,17 +1,104 @@
...
+    @Output()
+    monthClicked: EventEmitter<number> = new EventEmitter<number>();
+
+    constructor(private eRef: ElementRef) {
+    }
+
+    /** Get a number (1-12) representing the current month  */
+    get currentMonth(): number {
+        const tempDate: Date = new Date();
+        return tempDate.getMonth() + 1;
```
MattL75: You can use FdDate.getToday().month to not have to do this

we should avoid repeating work like this and instead have all the (+1) month logic be centralized in FdDate

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.ts
```diff
@@ -1,17 +1,104 @@
...
+            }
+        }
+        if (newFocusedMonthId) {
+            this.focusElement(newFocusedMonthId);
+        }
+    }
+
+    /** @hidden */
+    focusElement(elementSelector: string): void {
+        const elementToFocus: HTMLElement = this.eRef.nativeElement.querySelector(elementSelector);
```
MattL75: using querySelector just isn't very `Angular`, and usually indicates that the solution can be imroved

library/src/lib/calendar/calendar2/calendar2.component.ts
```diff
@@ -52,7 +56,17 @@ export type DaysOfWeek = 1 | 2 | 3 | 4 | 5 | 6 | 7;
         '[attr.id]': 'id'
     }
 })
-export class Calendar2Component implements OnInit, ControlValueAccessor {
+export class Calendar2Component implements OnInit, OnDestroy, ControlValueAccessor {
+
+    // --------------- CODE CONCERNING THE MONTH VIEW --------------- 
+    /** @hidden For i18n, the list of month names. Internal use. */
+    monthNames: string[];
```
MattL75: You should get these directly from the service and not need an intermediary variable

library/src/lib/calendar/calendar2/calendar2.component.ts
```diff
@@ -297,4 +313,24 @@ export class Calendar2Component implements OnInit, ControlValueAccessor {
...
+        this.i18nLocalSub = this.calendarI18n.i18nChange.subscribe(() => {
+            this.monthNames = this.calendarI18n.getAllShortMonthNames();
+            this.cd.detectChanges();
+        });
+    }
+
+    /** @hidden */
+    ngOnDestroy() {
+        if (this.i18nLocalSub) {
+            this.i18nLocalSub.unsubscribe();
```
MattL75: I prefer you use the methodology of the takeUntil, it just scales better for large components

library/src/lib/calendar/calendar2/calendar2.component.ts
```diff
@@ -183,13 +197,15 @@ export class Calendar2Component implements OnInit, ControlValueAccessor {
 
     /** @hidden */
     constructor(public calendarI18nLabels: CalendarI18nLabels,
-                public calendarI18n: CalendarI18n,
-                public dateAdapter: DateFormatParser,
-                private calendarService: Calendar2Service) {
+        public calendarI18n: CalendarI18n,
+        public dateAdapter: DateFormatParser,
+        private calendarService: Calendar2Service,
+        private cd: ChangeDetectorRef) {
```
MattL75: cdRef pls

library/src/lib/calendar/calendar2/calendar2.component.ts
InnaAtanasova: Me, too. But that's how you implemented it for the calendar and I wanted to avoid discussions :)

library/src/lib/calendar/calendar2/calendar2.component.ts
InnaAtanasova: Just followed the Architecture.

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.ts
InnaAtanasova: how?

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.html
InnaAtanasova: Screen Shot 2019-07-12 at 9 47 26 AM

InnaAtanasova: https://sap.github.io/fundamental-styles/components/calendar.html

MattL75: Ok well we should change that in FD to buttons probably lol

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.ts
MattL75: idk, gotta think about it

library/src/lib/calendar/calendar2/calendar2.component.ts
MattL75: should be discussed then, not sure why the monthNames is a part of that if you can just get it directly

library/src/lib/calendar/calendar2/calendar2.component.ts
MattL75: yes so let's change it since we're modernizing :D

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.html
InnaAtanasova: Agreed for the better syntax, will change it. Not sure what you mean with `magic numbers`. SInce we agreed on working with 1-12 months I'm not sure how to handle the indexing of the ngFor (0-11) without adding some additional logic in the component.

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.ts
InnaAtanasova: ok

library/src/lib/calendar/calendar2/calendar2.component.ts
InnaAtanasova: ok

library/src/lib/calendar/calendar2/calendar2-views/calendar2-month-view/calendar2-month-view.component.html
MattL75: You can just define a const on top of the component with value 1 which is called something like `loopMonthOffset` and also documented. It's just for clarity 🤷‍♀

PR #1058, opened 2019-07-26
fix: introduce menu built-in keyboard support

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1062

### Please provide a brief summary of this pull request.
I added menu built-in keyboard support to combobox. And changed a little bit split button examples, cause there was visual bug with anchors color.
I also modified the tests, deleted some useless code.

I updated documentation for split button.

### DIFF
--- a/library/src/lib/combobox/combobox.component.html
+++ b/library/src/lib/combobox/combobox.component.html
@@ -29,25 +29,18 @@
                  [style.maxHeight]="maxHeight">
             <ng-content></ng-content>
             <ul fd-menu-list>
-                <ng-container *ngIf="!itemTemplate">
-                    <li *ngFor="let term of displayedValues; let index = index;"
-                        (click)="onMenuClickHandler(index)"
-                        (keydown)="onMenuKeydownHandler($event, index)"
-                    >
-                        <a fd-menu-item tabindex="0"
-                           [innerHTML]="term | displayFnPipe:displayFn | highlight:inputText:highlighting">
-                        </a>
-                    </li>
-                </ng-container>
-                <ng-container *ngIf="itemTemplate">
-                    <li *ngFor="let term of displayedValues; let index = index;"
-                        (click)="onMenuClickHandler(index)"
-                        (keydown)="onMenuKeydownHandler($event, index)"
-                        fd-menu-item tabindex="0"
-                    >
+                <li *ngFor="let term of displayedValues; let index = index;"
+                    (click)="onMenuClickHandler(index)"
+                    (keydown)="onMenuKeydownHandler($event, index)"
+                    fd-menu-item
+                    tabindex="0">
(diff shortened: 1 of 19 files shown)


MattL75:
Keyboard support is broken on last combobox example

library/src/lib/combobox/combobox.component.html
```diff
@@ -29,15 +29,25 @@
...
+                           [innerHTML]="term | displayFnPipe:displayFn | highlight:inputText:highlighting">
+                        </a>
+                    </li>
+                </ng-container>
+                <ng-container *ngIf="itemTemplate">
+                    <li *ngFor="let term of displayedValues; let index = index;"
+                        (click)="onMenuClickHandler(index)"
+                        (keydown)="onMenuKeydownHandler($event, index)"
+                        fd-menu-item tabindex="0"
+                    >
```
MattL75: Why did you change this? Why does there need to be two different loops in two different ng-containers?

MattL75: Can just leave the fd-menu-item on the li? This change makes the file pretty bloaty and confusing

library/src/lib/combobox/combobox.component.ts
```diff
@@ -145,14 +155,27 @@ export class ComboboxComponent implements ControlValueAccessor, OnInit, OnChange
         }
     }
 
+    ngOnDestroy(): void {
+        this.onDestroy$.next();
```
MattL75: complete the observable after .next

library/src/lib/combobox/combobox.component.html
JKMarkowski: We can, but then the anchors won't have fd-menu-item class. So it will have blue color.

JKMarkowski: I will add @input to class host binding, to have a control on it, then add just a class to anchor.

MattL75: Do we even need the anchors? I'm not sure why we have them, doesn't seem like a11y would require it

MattL75: HostBinding is a decent solution in the mean time

JKMarkowski: I just changed anchors to spans in combobox.

JKMarkowski:
All comments are solved.

library/src/lib/menu/menu-item.directive.ts
```diff
@@ -1,4 +1,4 @@
-import { Directive, ElementRef, HostBinding } from '@angular/core';
+import { Directive, ElementRef, HostBinding, Input } from '@angular/core';
```
JKMarkowski: Unused Input - to remove

Issue #1068, opened 2019-07-30
datepicker focus trap issue


droshev:
1. open https://deploy-preview-1048--fundamental-ngx.netlify.com
2. go to datepicker
3. click on the first datepicker to open
4. tab twice so the focus ends up on the month
5. press `space` to open the month menu
6. tab 3 times so the focus goes on the current month down
7. use the arrows to focus on another month
8. press `space` then the focus is out of the datepicker

Simiar issue:
open the datepicker by clicking on the calendar icon
- if you keep tabbing then the focus trap stays in the calendar
- if you press `shift` + `tab` then the focus goes out

MattL75:
A lot of stuff in this issue is expected behaviour 😅

PR #1069, opened 2019-07-30
fix: improved focus on calendar

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1068
Issues with focus has been fixed. 
Right now:
- Shift tab escapes the focus on all views (that's how it's supposed to be)
- Tab doesn't escape the focus, by default.
- After changing month/year the active(selected > today > first) day is focused
- When year list is switched (+/- 12), it's possible to focus on active(selected > today > first) year, by tab key.
- Also added some statements which will provide protection from `undefined is not an object/function` errors.

### DIFF
--- a/library/src/lib/calendar/calendar-views/calendar-year-view/calendar-year-view.component.ts
+++ b/library/src/lib/calendar/calendar-views/calendar-year-view/calendar-year-view.component.ts
@@ -1,7 +1,6 @@
 import { Component, OnInit, ViewEncapsulation, Output, Input, EventEmitter, ElementRef, AfterViewChecked, OnDestroy } from '@angular/core';
 import { FdDate } from '../../models/fd-date';
 import { takeUntil } from 'rxjs/operators';
-import { CalendarI18n } from '../../i18n/calendar-i18n';
 import { CalendarService } from '../../calendar.service';
 import { Subject } from 'rxjs';
 
@@ -16,6 +15,11 @@ import { Subject } from 'rxjs';
     }
 })
 export class CalendarYearViewComponent implements AfterViewChecked, OnInit, OnDestroy {
+
+    /** @hidden
+     *  This variable is used to define which year should be possible to focus (tabindex = 0)
+     * */
+    activeYear: number;
 
     /** Parameter that stores the dozen of years that are currently being displayed. */
     calendarYearList: number[];
@@ -47,6 +51,10 @@ export class CalendarYearViewComponent implements AfterViewChecked, OnInit, OnDe
     /** Event fired when a year is selected. */
     @Output()
     readonly yearClicked: EventEmitter<number> = new EventEmitter<number>();
+
+    /** @hidden */
+    constructor(private eRef: ElementRef, private calendarService: CalendarService) {
+    }
 
     /** @hidden */
     ngAfterViewChecked(): void {
@@ -92,14 +100,19 @@ export class CalendarYearViewComponent implements AfterViewChecked, OnInit, OnDe
         this.onDestroy$.complete();
     }
 
-    /** @hidden */
-    constructor(private eRef: ElementRef, private calendarService: CalendarService) { }
-
-    /** @hidden */
-    private constructYearList(): void {
-        this.calendarYearList = [];
-        for (let x = 0; x < 12; x++) {
-            this.calendarYearList.push(this.firstYearInList + x);
+    /**
+     * Method that returns active cell, which means:
+     * if there is any selected year, return selected year
+     * if there is no selected year, but there is current year, return current year
+     * if there is no current year, or selected, return first one
+     */
+    private getActiveYear(): number {
+        if (this.calendarYearList.find(year => year === this.yearSelected)) {
+            return this.calendarYearList.find(year => year === this.yearSelected);
+        } else if (this.calendarYearList.find(year => year === this.currentYear)) {
+            return this.calendarYearList.find(year => year === this.currentYear);
+        } else {
+            return this.calendarYearList[0];
         }
     }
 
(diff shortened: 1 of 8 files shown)


MattL75:
Can we please keep in mind that we have **focus-trap** as a dependency?

droshev:
I thought we were using **focus-trap**?

MattL75:
@droshev We are, just not here - which was my point

JKMarkowski:
Hi @droshev @MattL75  I decided to not add focus-trap inside calendar, I tried to do it, but I encountered a lot of weird issues due to multiple calendars on same page. To introduce focus-trap, I would need to spend a lot of time and use some workarounds.

MattL75:
I was more thinking of using focus trap on datepicker, but this works so whatever I guess

MattL75:
I was more thinking of using focus trap on datepicker, but this works so whatever I guess

library/src/lib/calendar/calendar-views/calendar-year-view/calendar-year-view.component.ts
```diff
@@ -92,14 +100,19 @@ export class CalendarYearViewComponent implements AfterViewChecked, OnInit, OnDe
...
+     * if there is no current year, or selected, return first one
+     */
+    private getActiveYear(): number {
+        if (this.calendarYearList.find(year => year === this.yearSelected)) {
+            return this.calendarYearList.find(year => year === this.yearSelected);
+        } else if (this.calendarYearList.find(year => year === this.currentYear)) {
+            return this.calendarYearList.find(year => year === this.currentYear);
+        } else {
+            return this.calendarYearList[0];
         }
```
MattL75: Consider instead defining these as const values and then doing your if statements on those. Not very practical nor readable to call .find twice per if

library/src/lib/calendar/calendar-views/calendar-year-view/calendar-year-view.component.ts
```diff
@@ -136,4 +149,13 @@ export class CalendarYearViewComponent implements AfterViewChecked, OnInit, OnDe
         this.yearSelected = selectedYear;
         this.yearClicked.emit(this.yearSelected);
     }
+
+    /** @hidden */
+    private constructYearList(): void {
+        this.calendarYearList = [];
+        for (let x = 0; x < 12; x++) {
```
MattL75: Put the 12 in a const above the loop and call it something like "displayedYearsAmount" or something

MattL75: also use `++x` instead of `x++` pls

library/src/lib/calendar/models/fd-date.ts
```diff
@@ -62,16 +62,26 @@ export class FdDate {
 
     /**
      * Get amount of milliseconds from 01.01.1970
+     * 0 is thrown when some some of properties (day,month,year) are not defined
```
MattL75: Consider having -1 value instead?

library/src/lib/calendar/models/fd-date.ts
```diff
@@ -62,16 +62,26 @@ export class FdDate {
...
+        if (this.year && this.month && this.day) {
+            return this.toDate().getTime();
+        } else {
+            return 0;
+        }
     }
 
     /**
      * Get number of weekday ex. Sunday = 1, Monday = 2, Tuesday = 3 etc.
+     * 0 is thrown when some some of properties (day,month,year) are not defined
```
MattL75: consider -1

library/src/lib/calendar/models/fd-date.ts
```diff
@@ -62,16 +62,26 @@ export class FdDate {
...
      * */
     public getDay(): number {
-        return this.toDate().getDay() % 7 + 1;
+        if (this.year && this.month && this.day) {
+            return this.toDate().getDay() % 7 + 1;
```
MattL75: Define what the 7 and 1 mean, probably in a comment above is fine

library/src/lib/calendar/calendar-views/calendar-year-view/calendar-year-view.component.ts
```diff
@@ -17,6 +16,11 @@ import { Subject } from 'rxjs';
 })
 export class CalendarYearViewComponent implements AfterViewChecked, OnInit, OnDestroy {
 
+    /** @hidden
+     *  This variable is used to define which year should be possible to focus (tabindex = 0)
+     * */
+    activeYear: number;
```
MattL75: I don't understand the purpose of this just by reading the docs comment, consider improving the description a bit

MattL75: Actually I don't really get what this even does

library/src/lib/calendar/calendar-views/calendar-year-view/calendar-year-view.component.ts
JKMarkowski: I'm not sure what is the reason, for changing `x++ to `++x`, but it's fine.

library/src/lib/calendar/models/fd-date.ts
JKMarkowski: The % 7 is not really needed this is leftovers after other getDay function. Removed.

library/src/lib/calendar/calendar-views/calendar-year-view/calendar-year-view.component.ts
JKMarkowski: It defines which year can be focused by tab key. I could be a little unclear. Description changed.

library/src/lib/calendar/calendar-views/calendar-year-view/calendar-year-view.component.ts
MattL75: The reason is my OCD, and my teachers always used ++x so I'm used to it, and we might as well have the same everywhere

Issue #1065, opened 2019-07-29
Add way to change am/pm naming


JKMarkowski:
#### Is this a bug, enhancement, or feature request?
Enhancement
#### Briefly describe your proposal.
It should be possible to change am/pm naming on time component. It can be achieved by existing time I18n labels class.

MattL75:
^ @droshev Not related to datepicker

PR #1070, opened 2019-07-30
fix: Refactor Time Component, Add AM/PM and placeholder i18n

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1065
I added second i18n service to make possible adding other placeholders for hours, minutes and seconds. I added way to provide am/pm naming by user. Right now it's case sensitive. Also had to make refactor of time component and remove oldPeriod. I added also some comments, and cleaned methods used in component.

There is description added, with example of usage custom service which provides naming.


library/src/lib/time/time.component.html
```diff
@@ -107,13 +107,13 @@
                 [attr.aria-label]="timeI18nLabels?.increasePeriodLabel"></button>
     </div>
     <div class="fd-time__input">
-        <input [(ngModel)]="period"
+        <input [ngModel]="period"
```
MattL75: Why did you remove two-way binding here?

library/src/lib/time/time.component.html
```diff
@@ -107,13 +107,13 @@
...
                [ngClass]="{'is-disabled': disabled, 'is-invalid': periodInvalid}"
                class="fd-form__control"
                (blur)="inputBlur('period')"
-               (ngModelChange)="periodModelChange()"
+               (ngModelChange)="periodModelChange($event)"
                type="text"
-               placeholder="am"
+               placeholder="{{timeI18n.meridianPlaceholder}}"
```
MattL75: nullcheck `?`

library/src/lib/time/time.component.html
MattL75: Usually using ngModelChange on top of two-way binding doesn't hurt

library/src/lib/time/time.component.ts
```diff
@@ -51,14 +52,11 @@ export class TimeComponent implements OnChanges, ControlValueAccessor {
 
     /** @hidden */
     @Output()
-    focusArrowLeft: EventEmitter<any> = new EventEmitter<any>();
+    focusArrowLeft: EventEmitter<void> = new EventEmitter<void>();
```
MattL75: readonly

library/src/lib/time/time.component.html
JKMarkowski: I don't want to override period variable, when it's invalid.

JKMarkowski:
All comments are solved, also time component has been refactored, so it's required to make review rerun.

library/src/lib/time/time.component.ts
```diff
@@ -51,56 +52,103 @@ export class TimeComponent implements OnChanges, ControlValueAccessor {
 
     /** @hidden */
     @Output()
-    focusArrowLeft: EventEmitter<any> = new EventEmitter<any>();
+    readonly focusArrowLeft: EventEmitter<void> = new EventEmitter<void>();
 
-    /** @hidden */
+    /** @hidden
+     * Used only in meridian mode, stores information if it's AM or PM
```
MattL75: Used only in meridian mode. Stores information the current am/pm state.

library/src/lib/time/time.component.ts
```diff
@@ -51,56 +52,103 @@ export class TimeComponent implements OnChanges, ControlValueAccessor {
...
+     * */
     period: string;
 
+    /** @hidden
+     * Defines if actually written meridian is invalid
+     * */
```
MattL75: For all of these you can probably just use the format

/**
 *
 */

library/src/lib/time/time.component.ts
```diff
@@ -22,7 +23,7 @@ import { TimeI18nLabels } from './i18n/time-i18n-labels';
 })
 export class TimeComponent implements OnChanges, ControlValueAccessor {
 
-    /** @Input When set to true, uses the 24 hour clock (hours ranging from 0 to 23)
+    /** @Input When set to false, uses the 24 hour clock (hours ranging from 0 to 23)
      * and does not display a period control. */
     @Input() meridian: boolean = false;
```
MattL75: gotta document as breaking

library/src/lib/time/time.component.ts
```diff
@@ -51,56 +52,103 @@ export class TimeComponent implements OnChanges, ControlValueAccessor {
...
+    writeValue(time: TimeObject): void {
+        if (!time) {
+            return;
+        }
+        this.time = time;
+        this.setDisplayedHour();
+    }
+
+    /** @hidden */
+    ngOnChanges(): void {
```
MattL75: check for a specific type of changes? Or does this need to occur after every kind of change?

library/src/lib/time/time.component.ts
```diff
@@ -256,97 +277,85 @@ export class TimeComponent implements OnChanges, ControlValueAccessor {
...
+        return period === pmMeridian;
+    }
+
+    /**
+     * @hidden
+     * Defines if period is AM, Considers the fact that period should be case sensitive
+     * */
+    private isAm(period: string): boolean {
+        const amMeridian = this.timeI18n.meridianCaseSensitive ? this.timeI18n.meridianAm : this.timeI18n.meridianAm.toLowerCase();
+        period = this.timeI18n.meridianCaseSensitive ? period : period.toLowerCase();
```
MattL75: Anywhere you use `toLowerCase()` instead use `toLocaleLowerCase()`.

Actually, it might even be better to do the comparison in `toLocaleUpperCase()`

docs/app/documentation/component-docs/time/time-docs.component.html
```diff
@@ -50,8 +50,12 @@ <h2>Disabled State</h2>
 
 <h2>Internationalization</h2>
 <description>
-  The time component uses a custom service called TimeI18nLabels to give aria-labels to the buttons and inputs.
-  Inspect the HTML elements to see the Spanish aria-labels.
+    The time component uses a custom service called TimeI18nLabels to give aria-labels to the buttons and inputs.
+    Inspect the HTML elements to see the Spanish aria-labels.
+    The time component uses also a custom service called TimeI18n to give possibility, to change naming of meridian
```
MattL75: The time component also uses a custom service called TimeI18n to allow the possibility of changing the name of the meridian modifier. In the example below, the german vorm. and nach. are shown.

library/src/lib/time/i18n/time-i18n.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Placeholder for seconds input
+     * */
+    secondsPlaceholder: string = 'ss';
+
+    /**
+     * Defines if the meridian validation should be case sensitive.
+     * */
+    meridianCaseSensitive: boolean = false;
```
MattL75: For all these see my comment about docs comments format

library/src/lib/time/time.component.ts
JKMarkowski: The comment was just not correct. Meridian option works like before.

library/src/lib/time/time.component.ts
JKMarkowski: Right now it checks if changes are including meridian/time configuration,

JKMarkowski:
All comments are done, just FYI.

docs/app/documentation/component-docs/time/time-docs.component.html
```diff
@@ -50,8 +50,12 @@ <h2>Disabled State</h2>
...
-  Inspect the HTML elements to see the Spanish aria-labels.
+    The time component uses a custom service called TimeI18nLabels to give aria-labels to the buttons and inputs.
+    Inspect the HTML elements to see the Spanish aria-labels.
+    The time component also uses a custom service called TimeI18n to allow the possibility of changing the name of the
+    meridian modifier. In the example below, the german vorm. and nach. are shown.
+    Also it is possible to modify placeholders of the inputs, such as hours, minutes, seconds, and meridian. Besides
+    service allows to define if meridian modifier should be validated case sensitive.
 </description>
 <component-example [name]="'ex6'">
     <fd-time-i18n-example></fd-time-i18n-example>
```
InnaAtanasova: Missing the HTML tab with the html code

InnaAtanasova:
I know we have an example with i18n in the Time component documentation, but can we add also an example in the Time Picker component? Or at least a reference to the Time component i18n example?

library/src/lib/time/time.component.ts
```diff
@@ -256,97 +290,85 @@ export class TimeComponent implements OnChanges, ControlValueAccessor {
...
+     * Handles period model change. depending on current hour and new period changes hours by +/- 12
+     */
+    periodModelChange(): void {
+        if (this.time && !this.time.hour) {
+            this.time.hour = 0;
+        }
+        if (!this.period ||
+            (!this.isPm(this.period) && !this.isAm(this.period))
+        ) {
             this.periodInvalid = true;
```
InnaAtanasova: should we reset the invalid input like we do with hours/minutes/secs? 
Screen Shot 2019-08-08 at 4 16 46 PM

JKMarkowski: It's very easy to achieve, but I'm not sure if that's how it's supposed to be.

JKMarkowski: Actually it will improve UX, I'm gonna try it.

JKMarkowski:
@MattL75 , What do you think about it ?

Issue #1075, opened 2019-07-31
ComboBox component should return object for formControl value.


carlostekken:
#### Currently ComboBoxComponent with dropdownValues is synchronizing the selected text value to the formControl. For example if I have the following structure assigned to the comboBox:

  dropdownValues:  = [
       {
      id: 111,
      description: 'XXX'
    },
   {
      id: 222,
      description: 'YYY'
    }
  ];

When the value is selected the formControl is assigned with the description (assuming displayFn returns description)

this make sense to reflect the value selected in the input text However it does not make sense for the formControl value since there we need to have the whole object in order to get the "id" attribute.

I am aware you are returning the whole object in the "selected" property however this will not work with a reactive approach since selected attribute is not bonded with any formControl.

Find the example for the enhancement described here:

https://stackblitz.com/edit/angular-xmns77

#### 0.8.0

https://stackblitz.com/edit/angular-xmns77

Reactive Forms implementation value subscription requires the whole object no the the display value.

MattL75:
That's a good point. I'm not really a fan of complicating the component's usage within reactive forms, but there's really no other way to solve this issue that I can see.

PR #1077, opened 2019-08-01
feat: added option in combobox to return whole object; added forms example

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1075
I added dropdown mode to combobox component, thanks to it component can communicate with the outside not only by strings, but with whole objects that are passed as a options.  
I added new example with reactive forms and usages of 2 combobox components, to compare enabled and disabled dropdown mode.


MattL75:
I'm not a huge fan of this. Before I started working on fundamental, the guys has gone down this path of having a custom input for every "mode" that a component could have. This usually led to very confusing templates, or bloated functionality in general.

JKMarkowski:
I made some naming changes, so right now it's easier understand what is it for.

PR #1088, opened 2019-08-06
feat: added ng-add schematics

### LINKED ISSUE #1007: Switch fundamental-styles from peer-dependency to dependency
Enhancement

We need to make fd-styles an actual dependency because of all the breaking changes that will be coming to that library. We should force a certain version to make sure we stay functional on developer environments.

### LINKED ISSUE #1032: Make package ng-addable
Feature

We should make use of cli schematics to make the library ng-addable, and potentially extend this to generate project templates in the future.

Tasks to automate:

- [ ] Installing fd-styles
- [ ] Adding the styles to the styles array
- [ ] Configuring Angular animations

MattL75:
closes #1032 
closes #1007 
* Makes Fundamental-Styles a dependency.
* Adds schematics for ng-add with scaffolding.
* Deleted package-lock and re-installed because of a huge number of modifications to both package.json files.


README.md
```diff
@@ -27,43 +27,13 @@ Prior knowledge of Angular is recommended.
...
-      ...
-      imports: [BrowserAnimationsModule],
-      ...
-    })
-    export class DemoModule { }
-    ```
+    `ng add fundamental-ngx`
     
-    Alternatively, you can disable the animations by importing `NoopAnimationsModule` instead.
+    *If you do not use the Angular CLI or if this command does not work for you, please see the [full installation guide](https://github.com/SAP/fundamental-ngx/wiki/Full-Installation-Guide-%5B0.9.0%5D).*
```
droshev: I would rename the wiki(link) without having the version as part of it.
Maybe we can separate the content before `0.9` and after

library/package.json
```diff
@@ -1,29 +1,33 @@
...
-    "standard-version": "4.4.0"
-  },
-  "dependencies": {
-    "focus-trap": "^5.0.2",
-    "popper.js": "^1.15.0"
-  }
+    "name": "fundamental-ngx",
+    "version": "0.9.0-rc.9",
+    "schematics": "./schematics/collection.json",
+    "description": "SAP Fiori Fundamentals, implemented in Angular",
```
droshev: `Fiori` was removed in master

README.md
MattL75: Yeah it's just there to not confuse people. It will be removed after 0.9.0 is released.

library/package.json
MattL75: When I merge this will be fixed

README.md
droshev: then we need to update `README.md` again

MattL75: Okay let's confuse people 😀

PR #1113, opened 2019-08-12
fix: shellbar refactor

JKMarkowski:
.
There was a need to refactor shellbar component.
- I cleared it up a little bit, added typings for inputs, because before it was type: any.
- Also I removed search input usage on shellbar, replaced it by combobox.
- Added other behaviour on product menu component, when there are no product menus, it would behave like a subtitle.
- Added Shellbar mode to combobox (following search input)


droshev:
Screen Shot 2019-08-23 at 11 24 41 AM
It looks like the product menu doesn't work properly

droshev:
I think we should park that until we clean the shell-bar in `fundamental-styles`

JKMarkowski:
Sure, let's wait for `fundamental-styles` clean up. The product menu bug is fixed now.

PR #1114, opened 2019-08-12
fix: downgrade rxjs

MattL75:
For some reason the build errors with the latest merge. Not sure why but it seems like downgrading from 6.5.2 does the trick.


JKMarkowski:
Let's check if it works.

PR #1135, opened 2019-08-16
docs: Add a fixed position popover

droshev:
Add an example with fixed popover position(bottom) - using popper.js API

### DIFF
--- a/docs/app/documentation/component-docs/popover/examples/popover-simple/popover-example.component.html
+++ b/docs/app/documentation/component-docs/popover/examples/popover-simple/popover-example.component.html
@@ -1,13 +1,16 @@
 <div class="fd-docs-flex-display-helper">
     <fd-popover placement="left">
         <fd-popover-control>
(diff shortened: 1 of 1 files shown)


docs/app/documentation/component-docs/popover/examples/popover-simple/popover-example.component.html
```diff
@@ -1,13 +1,16 @@
 <div class="fd-docs-flex-display-helper">
     <fd-popover placement="left">
         <fd-popover-control>
-            <fd-image [size]="'l'" [circle]="true" [photo]="'https://placeimg.com/400/400/nature'"></fd-image>
+            <fd-image [size]="'l'"
+                      [circle]="true"
+                      [photo]="'https://placeimg.com/400/400/nature'"></fd-image>
```
InnaAtanasova: Maybe it's a good idea we all use the same settings in VS code to avoid this

droshev: I agree. what would you suggest?
maybe we can even make it part of the PR/Travis?

PR #1142, opened 2019-08-19
fix: datepicker update all fields after changing the locale

### LINKED ISSUE #1140: datepicker localization issue
bug

the datepicker weekdays on the day view(whole month) doesn't get localized

latest `fundamental-ngx` version `0.9.0`

1. open https://stackblitz.com/edit/angular-pajrvk-bqtile
2. open the datepicker - the default language is Bulgarian and the current month and the week days are properly in Bulgarian
3. click on the German or French button - it will change the i18n to German or French
4. open the datepicker: the current month name is properly translated but not the weekdays

droshev:
fixes #1140 build the fields after a new locale is being applied
add a localization examples for the calendar(3 languages)

### DIFF
--- a/docs/app/documentation/component-docs/date-picker/examples/date-picker-i18n-example.component.ts
+++ b/docs/app/documentation/component-docs/date-picker/examples/date-picker-i18n-example.component.ts
@@ -3,6 +3,7 @@ import { Component, Injectable } from '@angular/core';
 import { CalendarI18nLabels } from '../../../../../../library/src/lib/calendar/i18n/calendar-i18n-labels';
 import { FdDate } from '../../../../../../library/src/lib/calendar/models/fd-date';
 
+// The weekdays translations have to start with Sunday
 const localized_values = {
     'fr': {
         weekdays: ['Di', 'Lu', 'Ma', 'Me', 'Je', 'Ve', 'Sa'],
@@ -39,13 +40,8 @@ export class LanguageService {
 @Injectable()
 export class CustomCalendarI18n extends CalendarI18n {
 
-    // You could also define a custom service and inject it here
-    //language: string = 'fr';
-    languageService: LanguageService;
-
-    constructor(languageService: LanguageService) {
+    constructor(private languageService: LanguageService) {
         super();
-        this.languageService = languageService;
     }
 
     getDayAriaLabel(date: Date): string {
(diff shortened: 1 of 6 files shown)


docs/app/documentation/component-docs/date-picker/examples/date-picker-i18n-example.component.ts
```diff
@@ -9,29 +9,59 @@ const localized_values = {
...
+
 @Injectable()
 export class CustomCalendarI18n extends CalendarI18n {
 
     // You could also define a custom service and inject it here
-    language: string = 'fr';
+    //language: string = 'fr';
+    languageService: LanguageService;
+
+    constructor(languageService: LanguageService) {
```
MattL75: If you are using a service you should use it correctly with **dependency injection** to set a good example for users. In my opinion you should remove the service and use the method that was present before with the property. It's simpler, and actually fits pretty much every use case. But I don't really mind which is used as long as it's used correctly and has Angular best practices included.

docs/app/documentation/component-docs/date-picker/examples/date-picker-i18n-example.component.ts
```diff
@@ -3,35 +3,61 @@ import { Component, Injectable } from '@angular/core';
 import { CalendarI18nLabels } from '../../../../../../library/src/lib/calendar/i18n/calendar-i18n-labels';
 import { FdDate } from '../../../../../../library/src/lib/calendar/models/fd-date';
 
+//the weekdays translations have to start with Sunday
```
MattL75: `// The`

docs/app/documentation/component-docs/date-picker/examples/date-picker-i18n-example.component.ts
```diff
@@ -55,7 +81,15 @@ export class CustomI18nLabels extends CalendarI18nLabels {
 @Component({
     selector: 'fd-datepicker-i18n-example',
     template: `
-        <fd-date-picker [(ngModel)]="date"></fd-date-picker>`,
+        <label fd-form-label for="language">Select language:</label>
+        <fd-button-group id="language" style="margin-bottom:20px">
+            <button fd-button-grouped [size]="'xs'" (click)="setFrench()" [state]="selected === 1 ? 'selected' : ''">French</button>
```
MattL75: you should use get methods instead of boolean expressions in the template where possible. Doing so makes code self-documenting and easier to read/follow.

docs/app/documentation/component-docs/date-picker/examples/date-picker-i18n-example.component.ts
```diff
@@ -70,7 +104,26 @@ export class CustomI18nLabels extends CalendarI18nLabels {
     ]
 })
 export class DatePickerI18nExampleComponent {
+    languageService: LanguageService;
+    selected: number = 3;
+
+    constructor(languageService: LanguageService) {
+        this.languageService = languageService;
```
MattL75: This should usually never be in an angular app, you should provide the service instead

library/src/lib/calendar/calendar-views/calendar-day-view/calendar-day-view.component.ts
```diff
@@ -537,7 +529,7 @@ export class CalendarDayViewComponent implements OnInit, AfterViewChecked, OnCha
      * Method that returns first letter of every weekday, basing on CalendarI18nDefault. Can be changed by user by
      * providing other class which implements CalendarI18n
      */
-    private getShortWeekDays(): string[] {
+    get shortWeekDays(): string[] {
```
JKMarkowski: I'm not sure about this approach. Every time angular detects any change(scroll, mouse hover, clicks) these array methods will be called. It can be bad for performance. The same thing is used on month view and it wasn't changed.

PR #1150, opened 2019-08-21
fix: action bar fix for the responsiveness of menu list

manu-kr:
fixes https://github.com/SAP/fundamental-styles/issues/234

PR has changes to fix for the responsiveness of menu list in the action bar and also removed hard coded value for width of the mobile screen size.

### DIFF
--- a/docs/app/documentation/component-docs/action-bar/examples/action-bar-contextual-menu-example.component.html
+++ b/docs/app/documentation/component-docs/action-bar/examples/action-bar-contextual-menu-example.component.html
@@ -3,7 +3,7 @@
         <h3 fd-action-bar-title>Page Title</h3>
     </div>
     <div fd-action-bar-actions>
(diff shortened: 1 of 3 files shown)


docs/app/documentation/component-docs/action-bar/examples/action-bar-contextual-menu-example.component.html
```diff
@@ -3,13 +3,10 @@
         <h3 fd-action-bar-title>Page Title</h3>
     </div>
     <div fd-action-bar-actions>
-        <fd-popover>
+        <fd-popover placement="bottom">
```
droshev: how about bottom-end in case that button is on the right border?

droshev:
The title has to start with ‘fix:’

manu-kr:
fix for the responsiveness of menu list in the action bar

Issue #1139, opened 2019-08-16
Combobox - Emit event when button icon is clicked


KevinOkamoto-SAP:
#### Is this a bug, enhancement, or feature request?
Feature request.

#### Briefly describe your proposal.
When using combobox as a search input field, the user needs to be able to execute a "free-text" search where the search text may or may not be one of the items in the `dropdownValues`.

While the developer can currently capture a "free-text" search event by using the attribute `searchFunction` and binding their custom search function. The `searchFunction` only is invoked on keyboard 'enter'. This function also needs to be invoked when the "icon" button is clicked.

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
Angular 7
Fundamental 0.9.0

#### If this is a bug, please provide steps for reproducing it.
Steps:
- Enter text in input field, e.g. "Ap"
- Click "icon button"

Result:
No event is emitted, nor is the `searchFunction` executed.

Expected Result:
Event should be emitted or `searchFunction` gets executed

JKMarkowski:
Thanks for reporting this issue, that's a good point.

PR #1165, opened 2019-08-23
fix: combobox primary button click

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1139
I added search function call on primary button click.


InnaAtanasova:
Can we add an example in the documentation with this specific case? Like providing a searchFunction and the button is not the arrow but the Search icon? Let me know what you think about it @droshev @JKMarkowski 
Apart from this the code looks great!

JKMarkowski:
Sure, I will add example to this PR

Issue #1133, opened 2019-08-15
Menu items should have optional separator lines between them


InnaAtanasova:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
Currently the [Menu](https://sap.github.io/fundamental-ngx/menu) component has no separator lines between the menu items. There should be an option to have items with separators.

![Screen Shot 2019-08-15 at 4 11 28 PM](https://user-images.githubusercontent.com/39598672/63123658-7f580880-bf77-11e9-9b83-08e66763a21d.png)

kavya-b:
I am working on this issue at present, but unable to assign it myself due to some access issues.

sKudum:
Reassigning it to Kavya, As she is working on it and same has informed to JKMarkowski as well.

kavya-b:
Fixed in PR https://github.com/SAP/fundamental-ngx/pull/1180

PR #1180, opened 2019-08-27
fix: added optional separator line to menu list items

kavya-b:
#1133 
Added optional separator line to menu list items. On fd-menu tag, use **[separator] = "true"** to get the line separator on the list items.
Yes.

### DIFF
--- a/library/src/lib/menu/menu.component.scss
+++ b/library/src/lib/menu/menu.component.scss
@@ -3,7 +3,3 @@
         outline: var(--fd-color-action-focus) dotted 1px;
     }
 }
-.fd-menu__list-separator > ul > li,
-.fd-menu__list-separator > fd-menu-group > ul > li{
-    border-bottom: 1px #EDEDED solid;
-}
--- a/library/src/lib/menu/menu.component.scss
+++ b/library/src/lib/menu/menu.component.scss
@@ -3,3 +3,6 @@
         outline: var(--fd-color-action-focus) dotted 1px;
     }
 }
+.fd-menu__list-separator > li{
+    border-bottom: 1px #EDEDED solid;
+}
--- a/library/src/lib/menu/menu.component.scss
+++ b/library/src/lib/menu/menu.component.scss
@@ -3,6 +3,7 @@
         outline: var(--fd-color-action-focus) dotted 1px;
     }
 }
-.fd-menu__list-separator > li{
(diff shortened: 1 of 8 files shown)


library/src/lib/menu/menu.component.ts
```diff
@@ -12,9 +15,24 @@ import {
     encapsulation: ViewEncapsulation.None,
     styleUrls: ['menu.component.scss']
 })
-export class MenuComponent {
+export class MenuComponent extends AbstractFdNgxClass {
```
JKMarkowski: Could you please use `@HostBinding` along with `@Input` to bind class to input variable? There are some examples in code. There is no sense to use `AbstractFdNgxClass` anymore.

library/src/lib/menu/menu.component.ts
```diff
@@ -12,9 +15,24 @@ import {
...
+    separator: boolean = false;
+
+    /** @hidden */
+    constructor(public itemEl: ElementRef) {
+        super(itemEl);
+    }
+    /** @hidden */
+    _setProperties(): void {
+        if (this.separator) {
+            this._addClassToElement('fd-menu__list-separator');
```
JKMarkowski: Also this class shouldn't be added there. We need to add class `fd-menu__list--separated` into `fd-menu-list` component.

library/src/lib/menu/menu.component.scss
```diff
@@ -3,3 +3,7 @@
         outline: var(--fd-color-action-focus) dotted 1px;
     }
 }
+.fd-menu__list-separator > ul > li,
+.fd-menu__list-separator > fd-menu-group > ul > li{
+    border-bottom: 1px #EDEDED solid;
+}
```
JKMarkowski: The General problem of this solve is that, we would need to follow all of the changes from fundamental styles. So let's use prepared class and delete this one.

kavya-b:
Hi @JKMarkowski , as discussed, have made changes in accordance to your review comments. Please check.

library/src/lib/menu/menu.component.ts
kavya-b: done

library/src/lib/menu/menu.component.ts
kavya-b: done

PR #1181, opened 2019-08-27
docs: added example to disable the menu item in the action bar

manu-kr:
https://github.com/SAP/fundamental-styles/issues/236

Add a style disable the menu list item in the action bar example.

### DIFF
--- a/docs/styles.scss
+++ b/docs/styles.scss
@@ -105,6 +105,12 @@ pre {
         }
     }
 
+    li.is-disabled{
+        color: rgba(128, 128, 128, 0.4);
+        cursor: not-allowed;
+        text-decoration: none;
+    }
+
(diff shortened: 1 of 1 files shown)


docs/styles.scss
```diff
@@ -105,6 +105,12 @@ pre {
         }
     }
 
+    li.is-disabled{
+        color: rgba(128, 128, 128, 0.4);
+        cursor: not-allowed;
+        text-decoration: none;
+    }
+
```
JKMarkowski: It is actually wrong library, it resolves issue only inside docs. Please put there changes to [fundamental-styles](https://github.com/SAP/fundamental-styles) library.

manu-kr: I agree, I am making the changes in https://github.com/SAP/fundamental-styles

Issue #1182, opened 2019-08-27
feat: add link before header title


JYSW380:
#### Is this a bug, enhancement, or feature request?
feature
#### Briefly describe your proposal.
add link before header title for easy navigate.
see below screenshot for an example:
![Screen Shot 2019-08-27 at 9 21 59 AM](https://user-images.githubusercontent.com/41170611/63774883-332c9280-c8ac-11e9-9dbc-89cf2d04f587.png)

droshev:
We will merge the example with `input` and implement to for the whole documentation library later on

PR #1196, opened 2019-08-29
docs: Links on Input docs component sections

JYSW380:
#1182
Added link before header title for easy navigate. 
There is also new component named `fd-docs-section-title`, which is used to add titles before section, we should refactor whole docs page in future.
Only input group component updated so far.

### DIFF
--- a/docs/app/documentation/component-docs/input/input-docs.component.html
+++ b/docs/app/documentation/component-docs/input/input-docs.component.html
@@ -1,4 +1,9 @@
-<h2>Input</h2>
+<h2 id="input" class="docs-header-link">
+    <a class="docs-markdown-a" aria-describedby="input" href="/input#input">
+        <span class="sap-icon--chain-link"></span>
+    </a>
+        Input
+</h2>
 <component-example [name]="'ex1'">
     <fd-input-example></fd-input-example>
 </component-example>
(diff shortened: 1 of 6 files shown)


docs/styles.scss
```diff
@@ -221,3 +221,33 @@ a:focus {
...
+
+//this is the style for the link
+// .docs-markdown-a {
+//     visibility: hidden;
+//     margin-left: -25px;
+// }
+
+// .docs-header-link:hover .docs-markdown-a {
+//     visibility: visible;
+// }
```
JKMarkowski: I guess It is not needed anymore.

docs/app/documentation/component-docs/input/input-docs.component.html
```diff
@@ -1,28 +1,48 @@
...
 
 <separator></separator>
 
-<h2>Input Help elements</h2>
+<h2 id="input-help-elements" class="docs-header-link">
+    <a class="docs-markdown-a"aria-describedby="input-help-elements" href="/input#input-help-elements">
+        <span class="sap-icon--chain-link"></span>
+    </a>
+        Input Help elements
+</h2>
```
JKMarkowski: I think it will be better to create component which will be named `docs-section-title`. You would be able to pass `[siteUrl]` and `[sectionUrl]`. There can be also `<ng-content>` inside template for title section name. It will save a lot of space and time.

JKMarkowski:
I'm about to take care of this PR, since Jun is not here anymore. Just noticed some weird behaviour/styling of input group (I'm gonna remove display flex from doc's scss)
image

JKMarkowski:
@droshev @InnaAtanasova Could you take a look on it ?

PR #1209, opened 2019-09-02
feat: split modal alert service

JKMarkowski:
I added new `DynamicComponentService` files, which are used by modal and alert service. Also will be used by future notification component. This service takes whole responsibility for creating dynamic components. It allows also to inject services into dynamically created components..


library/src/lib/utils/dynamic-component/dynamic-component.service.ts
```diff
@@ -0,0 +1,77 @@
...
+    /** @hidden */
+    constructor(
+        private componentFactoryResolver: ComponentFactoryResolver,
+        private appRef: ApplicationRef,
+        private injector: Injector
+    ) {}
+
+    /**
+     * Function that creates dynamic component and injects services to allow communication between component and outside
+     * @param contentType Content of the component.
```
InnaAtanasova: Type of the component content

library/src/lib/utils/dynamic-component/dynamic-component.service.ts
```diff
@@ -0,0 +1,77 @@
...
+    Injector,
+    EmbeddedViewRef,
+    ComponentRef,
+    Type, TemplateRef
+} from '@angular/core';
+import { DynamicComponentInjector } from './dynamic-component-injector';
+import { DynamicComponentConfig } from './dynamic-component-config';
+
+/**
+ * Service used to dynamically generate a modal.
```
InnaAtanasova: Service used to dynamically generate components like modals/alerts/notifications

Issue #1202, opened 2019-08-30
Time control accepts non-numeric characters in Safari and Firefox


matt-williams8:
#### Is this a bug, enhancement, or feature request?

Bug

#### Briefly describe your proposal.

Replicate behaviour in Chrome where only numeric characters can be entered into time controls

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)

Latest, this can be seen in the example controls at https://sap.github.io/fundamental-ngx/time, https://sap.github.io/fundamental-ngx/timePicker and https://sap.github.io/fundamental-ngx/datetime-picker

#### If this is a bug, please provide steps for reproducing it.

Go to any of the example components listed above in Safari or Firefox and when editing a time control type in non-numeric characters. Whilst typing the time segment (hours, minutes or seconds) you are editing displays as `null` in the input field and when the popover is closed it reverts back to 0  

#### Is there anything else we should know?

In Chrome non-numeric characters are totally blocked and cannot be entered

droshev:
@matt-williams8 I couldn't reproduce that the non-numeric characters are accepted. It is true that you can type letters, but once you move to the next field or move the focus to other field then the input gets reset.

matt-williams8:
@droshev perhaps accept was the wrong choice of words, allows to be typed may have been a better choice. 

The input does reset, but that does not seem like the desired behaviour. For example if you have the hours set to 14 and the type some letters the input resets to 0, not 14. As mentioned in the original description, letters cannot even be typed in Chrome so ideally this behaviour would be replicated across all browsers

JKMarkowski:
Hi @matt-williams8, I will add some pipe which allows only numbers. It will work on every browser.

PR #1210, opened 2019-09-02
feat: added a directive that accepts only digits and special keys

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1202
I added only digits directive, which prevent from writing any non-digits chars on time component. It can be used anywhere.


matt-williams8:
This may be a bit late, but this implementation seems...excessive. Could the inputs not just be set to be HTML `input type="number"`? i.e. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number

PR #1221, opened 2019-09-04
chore: Added MultiLib & MultiApp support

fkolar:
refactor(MultiLib): Added full MultiLib,MultiApp Support

- This PR is first atempt to provide more suitable structure for the repository so it supports multiple libraries as well as multiple application. Even all this is possible with existing CLI but its painfull process as after each run you would need to always modify angular.json to achive something like this. 

- The main feature that was added is `https://github.com/nrwl/nx`, which enchances existing CLI schematics to provide real enterprise support for multi repo. 

To add additional app: 
 ` ng g app <name> <options>`
To Generate library
 ` ng g lib <name> <options>`

- App `docs` was completly refactored to use use relative scope package name to `'@fundamental-ngx/core'` instead of hardcoding path to library
- When creating new library you can have either shared library or publishable. (more in the nx website)
- This enforces running lint for the CI adn now we are using true angular cli tslint.json
- Refactored many files  under `apps/docs` to refer to new scope package `'@fundamental-ngx/core'` (We had many hardcoded links)
- make some changes in the in the imports where we work with moment js. (We should not import whole moment JS library only features that we use
 

**Current structure**

/apps/* - have `docs` app
/libs/*. => place is existing libray adn created scoped package `'@fundamental-ngx/core'`

- Updated  build scripts adn paths **This will be finetuned in this PR to see if CI can run properly**

- the Library package.json should not include any `dependencies` or `devDependencies` - only peer. Also the CI build shoudl `npm install from the main/root package.json`

How it knows to lookup the right classes when you import it like this `'@fundamental-ngx/core'` ?
 - in the root `tsconfig.json` you will see`PATHS` that setups the aliases so application can easily resolve this. 

To build

`ng build docs`
`ng build core`

To lint
`ng lint docs`
`ng lint core`

On this URL you can see how the new structure can look like:

https://github.com/fkolar/fundamental-ngx/tree/restructured


fkolar:
Please check this out as branch, to test it out locally and understand the structure. There will be conflict as I refactored whole repo.

package.json
```diff
@@ -91,7 +111,11 @@
         "tslib": "^1.10.0",
         "tslint": "^5.19.0",
         "typedoc": "^0.15.0",
-        "typescript": "3.2.4"
+        "typescript": "3.2.4",
+        "@ngrx/store-devtools": "7.2.0",
+        "ngrx-store-freeze": "0.2.4",
```
JKMarkowski: Hi @fkolar Do we need ngrx there ?

fkolar: No we will be able to remove this is not needed. It is jsut default behavior when you add nx schematics. But this is minor, we can remove this once, we are happy with this structure and I  can do 1 quick commit.

JKMarkowski: I went deeply into the, in case we want to have more, libs/docs it will be great, also the fact that we have paths shortcut is perfect. I'm happy with this structure so far, let's wait for other teammates opinion.

fkolar: I briefly run this structure with deno on one of the call, but yes we want to createanother layer on top of this called @xxxx/platform, to provide more simple and closed API , turing directives into components, introducing a form concept, adn some other thigns , plus the main motivation behind this is mainly this `@component` so we can instantiate things programatically. 

You guys, Inna, deno need to review this once more, to make sure existing functionality works as before, mainly CI, publishing etc.. 
Then we will have add manually all the commits  comes after 04/09/2019

libs/core/README.md
```diff
@@ -0,0 +1,25 @@
+# Core
```
droshev: do we need this file?

ci-scripts/publish.sh
```diff
@@ -8,7 +8,11 @@ git config --global user.name "fundamental-bot"
 # delete temp branch
 git push "https://$GH_TOKEN@github.com/$TRAVIS_REPO_SLUG" ":$TRAVIS_BRANCH" > /dev/null 2>&1;
 
-cd library
+# Frank Note:
+# library package.json should not have any dependencies or dev dependencies, only peer.
+# This should install from the root. s
```
droshev: we are publishing a npm package which is just the library, not the whole repo. That's why we are running it in the library

droshev:
> the Library package.json should not include any dependencies or devDependencies - only peer. Also the CI build shoudl npm install from the main/root package.json

@fkolar can you elaborate a little bit more this?

ci-scripts/publish.sh
fkolar: Yes understood, but all its is ok. But you dont need to have dependencies and devDep, it shoudl work fine with the global ones, and you use those globals ones just to compile you lib into DIST folder, and then npm publish takes it from there. Once you install a library, then those 2 sections became absolete as they are not used at all.

fkolar:
> > the Library package.json should not include any dependencies or devDependencies - only peer. Also the CI build shoudl npm install from the main/root package.json
> 
> @fkolar can you elaborate a little bit more this?

Already replied in my previous comment but the flow is always:

1. You need to have all available node packages so you can compile your code
2. You run ng-packagr using `ng build` (here when you publishing libraries you dont need to use --prod flag)
3. `dist/XXX` is created adn this binary is pushed to npm using npm, it has no effect if you use global node_modules, or you install local onces. you just duplicate this under libs. 
4. When somebody installs libs the dependencies/devDependencies are completly ignored as they target system is using only peerDependencies

   -- In older version of NPM people used to have dependencies cuz npm was able to automatically install it but now this functionality is not there anymore, you only alert developer using peerDependencies. If you want to install something, you use postInstal scripts.

PR #1228, opened 2019-09-06
chore: fix gh pages publishing

droshev:


droshev:
@jbadan i doubt this will work. netlify is empty

PR #1250, opened 2019-09-11
Refactor Input Group component, localization editor

JKMarkowski:
So, I did huge refactor on input group component, which caused a need to refactor localization editor. Right now the `new` input group component is used by
- Combobox
- Date Picker
- Date Time Picker
- Localization Editor
- Multi Input
- Time Picker
It is mandatory to implement self contained styles. Also styles are added to:
Loading Spinner,
Localization Editor,
Input Group


droshev:
> I havent noticed this before as I wasnt too much into actual code. but Is fundamentals using some custom naming conventions?
> 
> Usually component selectors are Kebab-case or we use dashes, directive selector are nonKebab. like `fdFormControl`.
> 
> Probably its too late I guess no, as we were not using lint at the beggining that much We might missed this one. But as I said I am not sure if we use some custom naming convension other then whar Angular recomends and Lint default rules are based on.

what we can do is to have both variants: `[fd-form-control], [fdFormControl]`

PR #1252, opened 2019-09-11
Change breadcrump tests naming, Ad styles to pagination

JKMarkowski:
Change breadcrump tests naming, Ad styles to pagination


libs/core/src/lib/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { BreadcrumbComponent } from './breadcrumb.component';
 
-describe('BreadcrumbDirective', () => {
+describe('BreadcrumbComponent', () => {
     it('should create an instance', () => {
         const directive = new BreadcrumbComponent();
         expect(directive).toBeTruthy();
```
InnaAtanasova: Maybe rename `directive` to `component`?

PR #1284, opened 2019-09-20
chore: Updated npm publish functionality

fkolar:
 - standard-version run now on root package.json
 - this updated version is then used to update both core,
    platform platform.json files
 - Added placeholder variables into package.json files to remove hardcoded
   dependencies
- npm publish now support both version core, platform


fkolar:
@droshev Is there any other way to test the publish automation other then running merge?

fkolar:
Just for the info:

- Publish script is now updated and expect 1 arguments rc or mastaer
 - You can find there two sections with specific code for master and rc
- you can find there this line `PACKAGES=(core platform)` which creates an array of all packages we support
 - then we use this to iterate and publish each package

```
cd "libs"

pwd
for P in ${PACKAGES[@]};
do
    echo publish "@fundamental-ngx/${P}"
    cd ${P}
    npm publish --access public
    cd ..
done

```

fkolar:
was the merge sucessful?  i will be back in 2 hours...

Issue #1281, opened 2019-09-19
Migrating to Angular 8


JKMarkowski:
#### Is this a bug, enhancement, or feature request?
Enhancement
#### Briefly describe your proposal.
We need to migrate `fundamental-ngx` library to angular 8.2.x. There will be listed some issues, caused during this migration.

fkolar:
It will be mostly smooth. There might a few issues running command ng update that will try to force to version 9, but if you specify purelly version 8 we shoudl be ok.

fkolar:
in case u wil run into the same issue, then it just run this specific command;

`ng update @angular/cli@8 @angular/core@8` instead of the generic one:
`ng update @angular/cli @angular/core` 

I will try to update to version 9 as I would like to test this with Ivy

JKMarkowski:
I just noticed that, some of libraries are no longer supported and won't be working with IVY. So I need to replace them in scope of this issue.

fkolar:
What libraries, what features are you refering to? There is breaking change as they now need to provide additional a flag for each View query, to tell if this needs to executed in ngInit or ngAfter... 

but the rest shoudl be ok.

JKMarkowski:
The libraries that are used on docs page.

fkolar:
If you find some issues, then please let me I know we will log issue on angular and I send message directly to Angular Team, as they are asking me to test migration to 9.

JKMarkowski:
`ngx-highlight.js` package had to be removed.

PR #1296, opened 2019-09-23
feat: angular 8 version

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1281
There is provided a lot of things, related to angular 8 changes. The most important one is docs, which uses IVY compilation and Query/Content reference strategy changes (Added {static: boolean} to every single).
I Added also some important components to public_api,ts.
The angular versions are changed to latest 8.* version.


fkolar:
The best is the load this to local branch and go thru it. Also Whey did you used `ViewChild` with `static:true` ?
I am asking cuz original behavior was more resolved in `ngAfterViewInit ` not in `ngInit`

fkolar:
- Let's add new node requirements to the package.json as there needs to be node 10+, so we can avoid issue peple are running this with older version of node


**Replace lazy modules with dynamic imports**

```
const routes: Routes = [
    { path: '', loadChildren: './documentation/documentation.module#DocumentationModule' },
    { path: '**', redirectTo: '/home' }
];

// vs
loadChildren: () => import('./documentation/documentation.module').then(m => m.DocumentationModule);

```

`ContentChild, ViewChild`= static option => Shoudl we keep original behaviour or change all to different one. 

`ViewChild` =>In versino 7 all view queries got resolve in ngAfterViewInit, but setting to true it is changed to ngInit

droshev:
If we want to have version `0.11.0` as angular 7 then we need to release `0.11.0` first and then merge this. Let's groom this.

fkolar:
Looks good, just have one question are you sure you want to enable IVY ?

JKMarkowski:
Sure, why shouldn't I ? :d

fkolar:
Its definitelly good move, just many depending apps might fail right away. and we just need to be aware of this. 

Does `docs` app work with this change ?

JKMarkowski:
There is not many apps depending on our `docs` app. `Docs` app works well, you can check preview

droshev:
I squashed the commits and tried to rebase but i ran into tons of errors.

Issue #1269, opened 2019-09-16
Dropdown: Maximum width and truncation


Lokanathan-k:
Is this a bug, enhancement, or feature request?
Enhancement

 **Briefly describe your proposal.**
If no fixed width is set, the width of the button is set automatically, depending on the text and the container. The maximum width of the dropdown button is 12 rem (192 px).

If the label exceeds the maximum width or the set fixed width, use an ellipsis to indicate that there's overflow text.

Since there is no text box area to display the selected text value need to introduce the text value in the DropDown component

fkolar:
Let's finish this one..

Issue #1268, opened 2019-09-16
Creating new diff types of DropDown Button types


Lokanathan-k:
#### Is this a bug, enhancement, or feature request?
enhancement 

**Briefly describe your proposal.**

in order to implement the different types of button types need make code change in such a way that it should be a minimal code change and need to allow the user to create a DropDown button in a configurable manner which allows the user to work effectively and easy development of the application

fkolar:
What is diff types ?

Lokanathan-k:
[cid:image001.png@01D56C7E.2F0ABE50]


With Regards,
Lokanathan K

PR #1298, opened 2019-09-24
platform select dropdown component 

Lokanathan-k:
committing the changes for the platform component with all the

https://github.com/SAP/fundamental-ngx/issues/1269

did enhancement of the code change compatible with Kit Spec added 4 different types of Dropdown select with border,no Border,split, and split with icon

and added a truncation of text in the select component

### DIFF
--- a/libs/platform/src/index.ts
+++ b/libs/platform/src/index.ts
@@ -1,1 +1,2 @@
 export * from './lib/platform.module';
+export * from './lib/components/select/public_api';
--- a/libs/platform/src/lib/components/select/option/option.component.html
+++ b/libs/platform/src/lib/components/select/option/option.component.html
@@ -1,0 +1,1 @@
+<ng-content></ng-content>
--- a/libs/platform/src/lib/components/select/option/option.component.scss
+++ b/libs/platform/src/lib/components/select/option/option.component.scss
@@ -1,0 +1,21 @@
+.fd-option-default-custom {
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow-x: hidden;
+
+    &:focus {
+        color: var(--fd-color-text-1);
+        background-color: var(--fd-color-background-hover);
+        outline: none;
+    }
+}
+
+.fd-option-default-custom.is-disabled {
+    // TODO
+}
+
+.fd-option-default-custom.is-selected {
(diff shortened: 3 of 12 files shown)


fkolar:
- Please follow guide I posted about commit message and PR 

- why we are not using existing component `<fd-select` and implementing everythign again? 

- `select/public_api.ts`
  - even you export inside you are missing record in the main `index.ts` otherwise its not visible to outside

- Please use formating in your IDE  
	- Some inconsistences. Select module as export, import always on new line platform
	module does not

-- Let;s get back to the design phase. The main reason why we decided to update select
is to hide code inside, so user does not need to deal with this. 
	- Select should have a binding, that provide list of options
		- Here we need some minimal design. What format, how.. 


`encapsulation: ViewEncapsulation.None,`
	- Please consult with Fundamentals latest changes as they are moving away from .none and they use encapsulated styles

- Our component needs to use change detection onPush	

- Why we have ` [(isOpen)]="isOpen" (isOpenChange)="isOpenChangeHandle($event)"`

	`[(isOpen)]="isOpen"` - are you passing in value or are you listening as well ? 

CSS - use formating - it looks liek hand made and we have tools for this. 
	 - we use cammel case, we dont use capital letters
	 - you should not be harding colors you should inherit them from css variables  - please consult how 
	    fundamentals does it. 

Keydown event we should redirect to /core - we dont want to duplicate functionality..

More coming..

fkolar:
Please when you create an issue pick the right label or more select a project so we can filter. 

As I wrote in my guide we make a different between commit message and PR title. I woudl start with this and contine with the other comments.

fkolar:
I think we are missing here elementary design and some understanding. Let's first finalize design, let;s have a review on it and let;s continue..

fkolar:
I think you should be more focusing on the specs, let;s start with this first, as you will not have to rewrite your major part of work.  In this the state like this this cannot be approved. 
Signature is not right + some other things. Week0

fkolar:
while looking at this, we havent aggred on the component itself in the design phase..  we have different comments that were not addressed at least I havent seen any message.  The major part was teh dropdown need to have internal implementation hidden from user and not try to do similar thigns like in core. 

also another part was how are you going to handle complex objects

PR #1352, opened 2019-10-03
feat(platform: new): button as standalone component (instantiable)

sKudum:
https://github.com/SAP/fundamental-ngx/wiki/Platform:-Button-Component-V1.0-Technical-Design
#### Please provide a brief summary of this pull request
 It has a Button component under platform project, created using a button directive of the core with more enhanced features.

It is a new component under platform project


fkolar:
please revert and follow posted guide. see slack

fkolar:
- you apply truncation everything do we want to really truncate everytime ? 


- Why do you have there this construction twice? 
```
[style.width]="width"
    [ngStyle]="{ width: width}">
```

- Why dou have duplicate declaration for _elementRef?  In TS when you do 
`constructor(private elementRef: ElementRef) {` you declare private member of the class you dont need to redefine it again. 


- Where are you using - `_elementRef`

- Let's be consitent 

```

    /** Whether to compress the button. */
    @Input() compact: boolean;

    /** Whether the button content should be in truncated form this width. */
    @Input()
    width: string;


```
When you start using things inlined keep using this inline if you want newline keep using new lines I like new line.

sKudum:
> * you apply truncation everything do we want to really truncate everytime ?

- By default a button will take the size based on the form container where it is placed even though tuncation class is applied.
Screen Shot 2019-10-04 at 9 31 14 AM

- If user don't want this behaiour and have a button with truncated conetent then they have to define the width.
Screen Shot 2019-10-04 at 9 33 01 AM

- Truncation will be applied only when the user adds contrarians on width as shown above.



> * Why do you have there this construction twice?
> 
> ```
> [style.width]="width"
>     [ngStyle]="{ width: width}">
> ```
> 

- Removed duplicate

>
> * Why dou have duplicate declaration for _elementRef?  In TS when you do
>   `constructor(private elementRef: ElementRef) {` you declare private member of the class you dont need to redefine it again.
> * Where are you using - `_elementRef`
> * Let's be consitent
> 

- _elementRef => Removed it, not required in current component.

> ```
> 
>     /** Whether to compress the button. */
>     @Input() compact: boolean;
> 
>     /** Whether the button content should be in truncated form this width. */
>     @Input()
>     width: string;
> ```
> 
> When you start using things inlined keep using this inline if you want newline keep using new lines I like new line.

> Kept new line where ever is required

libs/platform/src/lib/components/button/button.component.html
```diff
@@ -0,0 +1,13 @@
...
+    fd-button
+    [options]="options"
+    [compact]="compact"
+    [glyph]="glyph"
+    [disabled]="disabled"
+    class="truncate_btn__txt"
+    (click)="onBtnClick($event)"
+    [fdType]="type"
+    [ngStyle]="{ width: width }"
+>
```
fkolar: Put the width to a quetes

[ngStyle]="{'width': width }"

libs/platform/src/lib/components/button/button.component.scss
```diff
@@ -0,0 +1,6 @@
+.truncate_btn__txt {
```
fkolar: let;s keep the naming syntax like rest of hte fundamentals
`<libray>-<component>-<what>` we use dashes as well. 

`.fdp-btn-truncate__txt `

libs/platform/src/lib/platform.module.ts
```diff
@@ -1,12 +1,10 @@
...
 import { CommonModule } from '@angular/common';
 import { FundamentalNgxCoreModule } from '@fundamental-ngx/core';
+import { PlatformButtonModule } from './components/button/button.module';
 
 @NgModule({
-    imports: [
-        CommonModule,
-        FundamentalNgxCoreModule
-    ]
+    imports: [CommonModule, FundamentalNgxCoreModule],
```
fkolar: Dont change the existing formating to be inlined. 

Check your formating rules, they should not inline all like this it should be like before where arrays like imports are on the new lines, 

the same for class. declaratin

`{}` vs 

```{
}
``

libs/platform/src/lib/components/button/button.component.html
sKudum: formater is setting it with out '' automatically even tho I am adding '' . If I run with  [ngStyle]="{ 'width' : width }" then build is failing.

libs/platform/src/lib/components/button/button.component.scss
sKudum: Done

libs/platform/src/lib/platform.module.ts
sKudum: ng lint error is occurring if I place in a new line.

fkolar: it should not, what exactly is the linting error? we havent changed linting rules and they passed before..

sKudum: > it should not, what exactly is the linting error? we havent changed linting rules and they passed before..

Done, aligned in new line.

Issue #1356, opened 2019-10-03
Time Picker Without Minutes and Seconds issue


InnaAtanasova:
#### Is this a bug, enhancement, or feature request?
Bug
#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
Current
#### If this is a bug, please provide steps for reproducing it.
`<fd-time-picker [displaySeconds]="false"
                [(ngModel)]="timePickerOnlyHoursObject"></fd-time-picker>
<br />`

The input field still shows the seconds
![Screen Shot 2019-10-03 at 10 59 22 AM](https://user-images.githubusercontent.com/39598672/66139286-25b2a880-e5ce-11e9-805f-a698dbae4e1f.png)

PR #1357, opened 2019-10-04
fix: timepicker without minutes/seconds and add some tests

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1356
I changed parser/format functions to allow using only hours mode. There are also some tests added into. I fixed change detection between timepicker and time component. Also there is fixed validation on timepicker.


InnaAtanasova:
Screen Shot 2019-10-07 at 8 51 52 AM
Screen Shot 2019-10-07 at 8 51 30 AM

I know it's styling but it looks weird. Maybe we should add a min-width to the popover body, especially when we have only `hour`

JKMarkowski:
The padding is added here @InnaAtanasova

PR #1375, opened 2019-10-09
chore: Modify tsconfig, allow detecting changes form lib

JKMarkowski:
This changes on tsconfig provides detecting changes from library on `ng serve`. So if anything is changed from `libs`, the `docs` page will refresh with these changes.


JKMarkowski:
@fkolar Could you take a look on it ?

PR #1385, opened 2019-10-10
[NO FOR MERGE] IVY with @ng9

JKMarkowski:
That's testing PR, only to see the preview of ng9 with IVY performance


tsconfig.json
```diff
@@ -18,5 +18,8 @@
             "@fundamental-ngx/core": ["dist/libs/core",   "libs/core/src/index.ts"],
             "@fundamental-ngx/platform": ["libs/platform/src/index.ts"]
         }
-    }
+    },
+  "angularCompilerOptions": {
+    "enableIvy": false
```
droshev: this is not enabling it

JKMarkowski: I wanted to disabled IVY in general and enable only for `docs` page.

droshev: if i want to build the library with IVY then i should set it to true, right?

Issue #1246, opened 2019-09-10
Search Input Component in platform


KevinOkamoto:
#### Is this a bug, enhancement, or feature request?
Enhancement

#### Briefly describe your proposal.
Create an opinionated component for "Search Input" for the "platform" library.

The code should leverage the "ComboBoxComponent" from the "core" library, as it has 90% of the functionality needed.

Additional features which should be addressed in this component:
* Add "loading" state to inform user that a search initiated by component is in process
* Add "cancel" button in "loading" state, to allow user to trigger an event to interrupt search process
* Add "category" dropdown

**Loading State**
<img width="216" alt="Screen Shot 2019-08-27 at 6 41 17 AM" src="https://user-images.githubusercontent.com/48103491/64635966-8f67da00-d3b5-11e9-9e1c-72ee32896bd6.png">

**Cancel Button**
<img width="243" alt="Screen Shot 2019-08-27 at 6 41 27 AM" src="https://user-images.githubusercontent.com/48103491/64635989-9c84c900-d3b5-11e9-91a7-390c4fdd4283.png">
NOTE: The cancel button appears when the "loading state" spinner icon is hovered over.

**Category Dropdown**
<img width="214" alt="Screen Shot 2019-09-10 at 10 29 32 AM" src="https://user-images.githubusercontent.com/48103491/64636133-f08fad80-d3b5-11e9-8d11-14c0588c7793.png">

KevinOkamoto:
The current `ComboboxComponent` handles user input via `[(ngModel)]` binding for input values, and via binding of a user function to the `searchFunction` attribute.

I would prefer if this newer `SearchInputComponent` would handle these interactions using emitted events. So the interface would look like:

```
<fd-search-input
    [placeholder]="'Search...'"
    [dropdownValues]="dropdownValues"
    (inputChange)="onInputChange($event)"
    (searchSubmit)="onSearchSubmit($event)">
</fd-search-input>
```
where `inputChange` would be emitted whenever there was an input text change, and `searchSubmit` would be emitted whenever the user selected an option from the dropdown list or hit the "enter" key.

KevinOkamoto:
A fully-loaded `SearchInputComponent` will look like this:

```html
<fd-search-input
    [placeholder]="'Search...'"
    [dropdownValues]="dropdownValues"
    [categoryValues]="categoryValues"
    (inputChange)="onInputChange($event)"
    (searchSubmit)="onSearchSubmit($event)">
[...]

PR #1386, opened 2019-10-10
feat(platform): Creating search input component

KevinOkamoto:
#1246

This is an initial PR for the SearchInputComponent for `platform` as described here:

https://github.com/SAP/fundamental-ngx/wiki/Platform:-SearchInput-Technical-Design


fkolar:
### search-input.component.html 

`[triggers]="[]"` - there is some default value already. Or are you trying to have no event triggers at all ?

### PlatformSearchInputComponent

  - Is there a reason why we need to explicitelly call this `PlatformSearchInputComponent`? I woudl just call it `SearchInputComponent `, we can still import both from code, platform and give it alias. The import statement can do this for us. 

  - `import { ComboboxComponent, ComboboxItem } from '@fundamental-ngx/core';` The comboboxItem is not exported from core.

  - `ngAfterViewInit`  - cannot this `this.combobox` be null and cause problem at some cases ? 

 - Just to make sure this component will never be part of the form, its always meant to be page level component, so we dont need to support forms at all. 

 - Haven't we said that we will support both list of values and observable stream? 

 - We should have changedetection as `OnPush` to be more peformance driven and to guarantee that our component will not be cause perf. problems

- Let's put a directive as first thing in the `li` tag so its recognizable easier
```
<li *ngFor="let category of categories" (click)="setCurrentCategory(category)"><span
                            fd-menu-item>{{category.label}}</span></li> 

```

KevinOkamoto:
> `[triggers]="[]"` - there is some default value already. Or are you trying to have no event triggers at all ?

Yes, I'm purposely trying to disable the `triggers`. By default the combobox opens the dropdown on "click" of the input box - I did not want the dropdown box to open until the user starts typing and suggestions are provided.

KevinOkamoto:
> ### PlatformSearchInputComponent
> * Is there a reason why we need to explicitelly call this `PlatformSearchInputComponent`? I woudl just call it `SearchInputComponent `, we can still import both from code, platform and give it alias. The import statement can do this for us.

Initially I wanted to do this so that we could import both libraries wholely without having to import individual components. However, since core has removed SearchInputComponent from their library, I can make this name change.

> * `import { ComboboxComponent, ComboboxItem } from '@fundamental-ngx/core';` The comboboxItem is not exported from core.

I added the export of this interface in this PR.

> * `ngAfterViewInit`  - cannot this `this.combobox` be null and cause problem at some cases ?

I've already added "if-guards" if this is the case.

> * Just to make sure this component will never be part of the form, its always meant to be page level component, so we dont need to support forms at all.

As far as I know, this is the case.

> * Haven't we said that we will support both list of values and observable stream?

Yes, I need to implement this.

> * We should have changedetection as `OnPush` to be more peformance driven and to guarantee that our component will not be cause perf. problems

OK

> * Let's put a directive as first thing in the `li` tag so its recognizable easier
> 
> ```
> <li *ngFor="let category of categories" (click)="setCurrentCategory(category)"><span
>                             fd-menu-item>{{category.label}}</span></li> 
> ```

OK

Issue #1264, opened 2019-09-13
Menu as standalone instantiable component


kavya-b:
#### Is this a bug, enhancement, or feature request?
Enhancement

#### Briefly describe your proposal.

Core implementation of the menu is combination of `@Component` and `@Directive`: 
```html
<fd-menu>
    <ul fd-menu-list>
        <li fd-menu-item>
            Option 1
        </li>
        <li fd-menu-item>
            Option 2
        </li>
        <li fd-menu-item>
            Option 3
        </li>
    </ul>
</fd-menu>
```

And the the goal of this enhacement is to hide all the exposed code into component itself to make it standalone as well as provide some additional features

- Vertical scrolling with limits
- Menu item name must truncate if too long
- Tooltip should be displayed on hovering items
 - RTL mirroring of icons should work
- Keyboard focus should happen on non-actionable items

Where the component signature could something like this in order to provide best DX:

```html
<fdp...>
```

kavya-b:
Kevin has already worked on menu as part of shellbar development. As discussed with him, some of it can be reused here as well for a basic structure. We will need to make adjustments for removing custom element dependency.
Something like this is available currently:
```html
<fdp-menu #menuData separate-items="true" text-align="left"></fdp-menu>
```
where,

-  `#menuData` is a local reference to the data that will be passed as menu items.
- `separate-items` specify if line separator between menu items should be displayed.
- `text-align` specifies the alignment of item names

In our case, we don't really need `text-align` as ours will be strictly left-aligned for LTR mode.
Also, do we need to pass data through the template reference variable, or can we just pass an array of `MenuItem/MenuGroup` type where `MenuItem` consists of a label, icon, callback, selectable, selected fields and `MenuGroup` has `MenuItems` as children? Something that looks like this: 
```html
<fdp-menu 
       [menuData]="menuData" 
       separate-items="true" 
       scrolling="true" 
       scroll-limit=10>
</fdp-menu>
```
where,

- `[menuData]` is an array of MenuItems or MenuGroup
- `scrolling` specifies if the menu should scroll
- `scrollLimit` specifies the number of items to be displayed after which scrolling begins

kavya-b:
> Menu item name must truncate if too long

This has been fixed by Fundamental Styles team in PR https://github.com/SAP/fundamental-styles/pull/331 and issues SAP/fundamental-styles#317 and #1175 have now been closed.

PR #1390, opened 2019-10-11
feat(platform): Menu as a standalone component

kavya-b:
Fixes https://github.com/SAP/fundamental-ngx/issues/1264
This PR contains code for development of Menu as a standalone component, with varying levels of customization. The design is documented here: https://github.com/SAP/fundamental-ngx/wiki/Platform:-Menu-component-Technical-Design 
This is a new feature; documentation is pending.


libs/platform/src/lib/components/menu/menu-component/menu-item.component.html
```diff
@@ -0,0 +1,91 @@
...
+        [attr.disabled]="disabled"
+        [ngClass]="['normal-item-state', disabled ? 'is-disabled' : '']"
+    >
+        <div class="fd-menu__addon-before" *ngIf="icon; else noIcon">
+            <span [ngClass]="[icon]" data-tag="menu-item__icon-before" [attr.title]="icon"> </span>
+        </div>
+        <div class="fd-menu__label-truncation">
+            <span
+                class="fd-menu__item"
+                [ngStyle]="{ width: getItemWidth() }"
```
KevinOkamoto: Calling `getItemWidth()` within the property binding may cause performance issues. It is better to assign a variable here e.g. `[ngStyle]="{ width: finalItemWidth }"`.

And then in your component code, update the variable in the `OnChanges` cycle:

```ts
public finalItemWidth = '';

ngOnChanges(changes: SimpleChanges) {
  if (changes.itemWidth || changes.item) {
    this.finalItemWidth = this.getItemWidth();
  }
}

File: libs/platform/src/lib/components/menu/menu-component/menu-item.component.scss [hunk 2]
```diff
@@ -0,0 +1,104 @@
...
+        //don't apply hover properties on only label item, we need it over entire item
+        background-color: transparent;
+    }
+    text-overflow: ellipsis;
+    min-width: 192px;
+    height: 40px; //line=24 px. calc is 40-8-8(padding top/bottom)=24px
+    overflow: hidden;
+    background-color: transparent;
+    padding: 12px 16px;
+    color: var(--fd-menu-item-color);
```
KevinOkamoto: When using CSS variables, we need to also assign a hardcoded default color, since IE11 doesn't recognize the `var()` syntax.

So change:
```css
   color: var(--fd-menu-item-color);
```

to:

```scss
   color: #51555a; 
   color: var(--fd-menu-item-color);
```

Be sure to add the "hardcoded" value before the `var()` value. For browsers that recognize `var()`, the hardcoded value will be overwritten.

fkolar:
Can you please try to do rebase? we shoudl not see any Merge messgage in the commit

kavya-b:
> Can you please try to do rebase? we should not see any Merge messgage in the commit

Oops, forgot to do it this time. Currently branch is up-to-date with master, so it is not allowing me to rebase, Frank. I will do rebase as soon as someone pushes changes to ngx next time.

libs/platform/src/lib/components/menu/menu-component/menu-item.component.scss
kavya-b: done

libs/platform/src/lib/components/menu/menu-component/menu-item.component.html
kavya-b: done

kavya-b:
> Can you please try to do rebase? we shoudl not see any Merge messgage in the commit

done.

libs/platform/src/lib/components/menu/menu.module.ts
```diff
@@ -0,0 +1,13 @@
...
+import { MenuItemComponent } from './menu-component/menu-item.component';
+import { FundamentalNgxCoreModule, MenuKeyboardService } from '@fundamental-ngx/core';
+
+@NgModule({
+    declarations: [MenuComponent, MenuItemComponent],
+    imports: [CommonModule, FundamentalNgxCoreModule],
+    exports: [MenuComponent, MenuItemComponent],
+    providers: [MenuKeyboardService]
+})
+export class MenuModule {}
```
KevinOkamoto: There's already a `MenuModule` in core. Will this cause name conflicts?

kavya-b: It could, yes, if one is not careful while importing. Would you rather I rename it to PlatformMenuModule instead like it is done for Search and Button?

fkolar: Again we dont need to create some extra names as Menu is straight forward. If you want to use both of the for some reason you can import it like this:

```
import {MenuComponnet as PlatformMenu} from 'xxxx'
```

As we keep working on more components we will have similar names but it does not mean its bad , when we work on the e.g. dropdown, we want to still keep it a dropdown.

fkolar:
In general besides other comments we should follow the same path as `core` import `scss` from fundamental-style, and dont touch anything that is already define once, otherwise we will break theming. 

When working with styles and in the component we have set `viewencapsulation.none`, lets be aware that it means `Don't provide any template or style encapsulation`, in other words that style from your `menu-item scss`  will be put on top of the html file available for everybody and changing everything that other components sets (using the same styles).

kavya-b:
> In general besides other comments we should follow the same path as `core` import `scss` from fundamental-style, and dont touch anything that is already define once, otherwise we will break theming.
> 
> When working with styles and in the component we have set `viewencapsulation.none`, lets be aware that it means `Don't provide any template or style encapsulation`, in other words that style from your `menu-item scss` will be put on top of the html file available for everybody and changing everything that other components sets (using the same styles).

There is a huge gap between what our invision spec says and what fundamental-styles currently provides, Frank. I had already raised corresponding issues on fundamental-styles for this gap ([styles Menu open issues](https://github.com/SAP/fundamental-styles/issues?q=is%3Aissue+menu+author%3Akavya-b+is%3Aopen)) but only one has been addressed till now.

So, for now, what should be the way to proceed? Should I be dropping all css changes entirely?; this will break some unit and automation tests in addition to the entire look of the Menu component.

KevinOkamoto:
For most of our components, we should be trying to use the "core" directives and components in our templates - so we should be getting the styling automatically.

If there are gaps between our design and what fundamental-styles provides, then yes we need to add customized CSS rules to our components. We do need to be aware of theming however and try to use the CSS variable definitions of fundamental-styles, so that our color palate is the same.

libs/platform/src/lib/components/menu/menu.module.ts
KevinOkamoto: What happens when someone imports both "core" and "platform" libraries in their module?
```
@NgModule({
  declarations: [ ... ],
  imports: [
    ...
    FundamentalNgxCoreModule,
    FundamentalNgxPlatformModule,
    ...
  ],
  providers: [ ... ],
})
```

Will there be name conflicts in this case, because both libraries have imports with the same names?

kavya-b:
> For most of our components, we should be trying to use the "core" directives and components in our templates - so we should be getting the styling automatically.
> 
> If there are gaps between our design and what fundamental-styles provides, then yes we need to add customized CSS rules to our components. We do need to be aware of theming however and try to use the CSS variable definitions of fundamental-styles, so that our color palate is the same.

The "core" directives in Menu are mostly binding to the fundamental-styles classes such as `fd-menu__item`, `fd-menu__group`, `fd-menu--addon-before`, `fd-menu__list` etc. I am using these same classes directly in "platform" and wherever gaps are present, I am overriding them. I am also using the same CSS variable definitions wherever I can.

However, I am also adding additional classes for secondaryIcon such as `fd-menu--addon-after` which do not exist in fundamental-styles at all currently. Theming then could affect secondaryIcon and we should probably get fundamental-styles to also implement a secondaryIcon feature for consistency.

kavya-b:
@fkolar @KevinOkamoto changed name from `MenuModule` to `FdpMenuModule`.

fkolar:
> > For most of our components, we should be trying to use the "core" directives and components in our templates - so we should be getting the styling automatically.
> > If there are gaps between our design and what fundamental-styles provides, then yes we need to add customized CSS rules to our components. We do need to be aware of theming however and try to use the CSS variable definitions of fundamental-styles, so that our color palate is the same.
> 
> The "core" directives in Menu are mostly binding to the fundamental-styles classes such as `fd-menu__item`, `fd-menu__group`, `fd-menu--addon-before`, `fd-menu__list` etc. I am using these same classes directly in "platform" and wherever gaps are present, I am overriding them. I am also using the same CSS variable definitions wherever I can.
> 
> However, I am also adding additional classes for secondaryIcon such as `fd-menu--addon-after` which do not exist in fundamental-styles at all currently. Theming then could affect secondaryIcon and we should probably get fundamental-styles to also implement a secondaryIcon feature for consistency.

Even if they are just directive that adds class, we should still use them. The reason is that once styles do some new adjustments and some renamign, you dont want to change this on two places.  Let;'s use core directives as wrapper for the target class. . YOu dont want to be using directly class if some other component is wrapping them.

kavya-b:
> > > For most of our components, we should be trying to use the "core" directives and components in our templates - so we should be getting the styling automatically.
> > > If there are gaps between our design and what fundamental-styles provides, then yes we need to add customized CSS rules to our components. We do need to be aware of theming however and try to use the CSS variable definitions of fundamental-styles, so that our color palate is the same.
> > 
> > 
> > The "core" directives in Menu are mostly binding to the fundamental-styles classes such as `fd-menu__item`, `fd-menu__group`, `fd-menu--addon-before`, `fd-menu__list` etc. I am using these same classes directly in "platform" and wherever gaps are present, I am overriding them. I am also using the same CSS variable definitions wherever I can.
> > However, I am also adding additional classes for secondaryIcon such as `fd-menu--addon-after` which do not exist in fundamental-styles at all currently. Theming then could affect secondaryIcon and we should probably get fundamental-styles to also implement a secondaryIcon feature for consistency.
> 
> Even if they are just directive that adds class, we should still use them. The reason is that once styles do some new adjustments and some renamign, you dont want to change this on two places. Let;'s use core directives as wrapper for the target class. . YOu dont want to be using directly class if some other component is wrapping them.

Okay, yes, you're right. Made the changes accordingly. Please check.

droshev:
@kavya-b if you think the colors are not correct I would suggest to do a fix in styles instead of adding custom css here

kavya-b:
> @kavya-b if you think the colors are not correct I would suggest to do a fix in styles instead of adding custom css here

Hi Deno, are you suggesting to do this for only colors or any/all styling? What about adding new classes such as for a secondary icon?

kavya-b:
> Anything new on this PR ?

nothing new added, only rebased multiple times

Issue #1320, opened 2019-09-29
Review main README.md files


droshev:
We decided to create 2 packages in the repository `fundamental-ngx` so we would need to split and re-organize our `README.md` files.
The proposal:
- [ ] `README.md`
keeps the content related to both libraries; general information about the repository and contribution and adds links to the other 2 files
- [ ] `/libs/core/README.md`
contains information related to the package `@fundamental-ngx/core` - how to consume and how to contribute. This one gets published for npm package `@fundamental-ngx/core`
- [ ] `/libs/platform/README.md`
contains information related to the package `@fundamental-ngx/platform` - how to consume and how to contribute. This one gets published for npm package `@fundamental-ngx/platform`

JKMarkowski:
I added and copy pasted some information into Core lib README.md and changed a little infor on main README.md. It would be great if everyone can take a look on it

PR #1400, opened 2019-10-14
chore: #1320 change readme.md

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1320

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,15 @@
 The SAP Fundamental Styles library is a design system and HTML/CSS component library used to build modern product user experiences with the SAP look and feel.
 
 This repository contains 2 libraries, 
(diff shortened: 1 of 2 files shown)


fkolar:
how about something like this:

### API REFERENCE

@fundamental-ngx/core 

Fundamental-ngx  core provides base angular implementation of SAP Fundamental styles to offer developer rich set of components 
they can use when building angular application. `Fundamental-ngx core` angular components provides to end developer some degree of flexibility as most of the components allows finer customization directly on the HTML template level. 


@fundamental-ngx/platform 


`Fundamental-ngx platform` is built on top of core to both enhance existing functionality with additional features 
which are driven by application requirements as well as it provides higher abstraction for the components to hide most of the internal implementation detail to boost the productivity.

PR #1422, opened 2019-10-17
Docs(platform): Documentation structure creation in fundamental-ngx/platform

sKudum:
It has the platform documentation structure created along with button component templates added.


JKMarkowski:
You can uncomment a switcher 
image
To allow to switch to platform from UI.

PLATFORMREADME.md
```diff
@@ -0,0 +1,109 @@
...
+Prior knowledge of Angular is recommended.
+
+## Getting Started
+
+For an existing Angular CLI application,
+
+1. **Install Fundamental-NGX.**
+   `ng add @fundamental-ngx/platform`
+   For models prior to 0.10 use `fundamental-ngx`
+
```
JKMarkowski: Before 0.11.1 there wasn't platform library

PLATFORMREADME.md
```diff
@@ -0,0 +1,109 @@
...
+Run `ng test @fundamental-ngx/platform`. Append `--code-coverage` to generate code coverage documentation.
+
+For models prior to 0.10 use `fundamental-ngx`
+
+## Versioning
+
+The `@fundamental-ngx/platform` library follows [Semantic Versioning](https://semver.org/). These components strictly adhere to the `[MAJOR].[MINOR].[PATCH]` numbering system (also known as `[BREAKING].[FEATURE].[FIX]`).
+
+For models prior to 0.10 use `fundamental-ngx`
```
JKMarkowski: Like above, there wasn't platform lib on 0.10

apps/docs/src/app/documentation/services/api-docs.service.ts
```diff
@@ -1,16 +1,25 @@
...
     readonly BASE_URL = 'assets/typedoc/';
+    regex: RegExp;
 
-    constructor(private httpClient: HttpClient) {}
+    constructor(private httpClient: HttpClient, private router: Router) { }
 
     getComponentHtml(component: string): Observable<string> {
+        const regex = /(\/[^\/\s]+\/)/;
+        const currentLib = regex.exec(this.router.url)[0];
+        console.log(`Found match, group`, currentLib);
```
JKMarkowski: Let's remove all console logs

apps/docs/src/app/documentation/services/api-docs.service.ts
```diff
@@ -1,16 +1,25 @@
 import { Injectable } from '@angular/core';
+import { Router } from '@angular/router';
 import { HttpClient } from '@angular/common/http';
 import { Observable } from 'rxjs';
 
 @Injectable()
 export class ApiDocsService {
 
     readonly BASE_URL = 'assets/typedoc/';
+    regex: RegExp;
```
JKMarkowski: Is it used anywhere?

PLATFORMREADME.md
sKudum: updated

apps/docs/src/app/documentation/services/api-docs.service.ts
sKudum: done

apps/docs/src/app/documentation/services/api-docs.service.ts
sKudum: removed multiple decorations.

PLATFORMREADME.md
sKudum: updated

libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -9,19 +9,19 @@ export type ButtonOptions = 'light' | 'emphasized' | '';
...
+    /** Option to truncate content of the button based on width. */
     @Input()
     width: string;
 
     /** The icon to include in the button */
     @Input()
     glyph: string;
 
-    /** Whether the button is disabled. */
+    /** button is disabled on true. */
```
KevinOkamoto: Capitalize "button" -> "Button"

sKudum: Done

fkolar:
On the button page there is a button `Playground` => `click here`. It is not doing anything.

fkolar:
For the button the extensibility sections is not really formatted well, like all one big sentense in one paragraphs and it does not look that good, harder to read.

fkolar:
But over all it looks great. there are not really major issues, 

great work !!! our first Demo App well done. 

Btw. this can be a feature but it would be nice, if I am on the `platform-button` and there exists button in the `core` and if I click toggle to go to `CORE`, it would jump to button core demo page, and so on. 

What do you think @droshev .

sKudum:
> For the button the extensibility sections is not really formatted well, like all one big sentense in one paragraphs and it does not look that good, harder to read.

Done

sKudum:
> But over all it looks great. there are not really major issues,
> 
> great work !!! our first Demo App well done.
> Just is case let's go over content if we have meaniful text and we dont have some outdated text , numbers, versions, etc..
> 
> Btw. this can be a feature but it would be nice, if I am on the `platform-button` and there exists button in the `core` and if I click toggle to go to `CORE`, it would jump to button core demo page, and so on.
> 
> What do you think @droshev .

Got it Frank, Just have ☝️ dout. 
Are we trying to compare here core and platform? I think in this approach they are chances of confusion to the end user.

For example,
 I am a developer working on something referring button from the core and in between unintentionally I toggled it, without realizing the difference we may continue using whatever is displayed on the screen and I may end up with issues in my code.

sKudum:
Closing this PR as https://github.com/SAP/fundamental-ngx/pull/1481 handled the same changes.

Issue #1263, opened 2019-09-13
Convert action bar as component


manu-kr:
Enhancement

Briefly describe your proposal.

In order to be able to effectively work with with the component and have ability to instantiate them on the fly angular offers no other way than using components. With this in mind action bar needs to be enhanced in such way that most of the things will be hidden for developer and provide options to configure the values like title, description and actions.

The basic usage of this component with minimal setup need to represent the best practice to defined Fiori 3 design.

To clarify this further, and to provide the best user experience we will start from the end, meaning we will try to craft here different signature of the component how it will be used and viewed by the application developer, Later on we will execute actual implementation.

Use existing @fundamental-ngx/core action bar directives  in order to create a instantiable component.

#### Which versions of Angular and Fundamental NGX are affected? 

Angular 7

PR #1423, opened 2019-10-17
fix(platform): provide an option to make action bar sticky and fixing…

manu-kr:
… unit test case failures

https://github.com/SAP/fundamental-ngx/issues/1263

PR has changes to provide an option to configure sticky action bar.

### DIFF
--- a/libs/platform/src/lib/components/action-bar/action-bar.component.scss
+++ b/libs/platform/src/lib/components/action-bar/action-bar.component.scss
@@ -1,3 +1,10 @@
 .action-bar-display {
     display: -webkit-box;
 }
+.fdp-action-bar-sticky {
+    position: -webkit-sticky;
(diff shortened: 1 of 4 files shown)


libs/platform/src/lib/components/action-bar/action-bar.component.scss
```diff
@@ -1,3 +1,10 @@
 .action-bar-display {
     display: -webkit-box;
 }
+.fdp-action-bar-sticky {
+    position: -webkit-sticky;
+    position: sticky;
+    top: 0;
+    background-color: #ffffff;
+    background-color: var(--fd-color-background-2);
+}
```
KevinOkamoto: This solution will not work for IE11.
https://caniuse.com/#feat=css-sticky

PR #1437, opened 2019-10-21
feat(platform): adding new component select under platform

Lokanathan-k:

feat(platform): adding new component select under platform issue details:#1268


fkolar:
I think have already commented this one before, the main goal is ot hide complexity inside so we can programatically instantiate components pass values and we are dont with this:

```
   <fdp-select [(value)]="wrapperValue" formControlName="selectControl">
            <fdp-option id="fdtest1" [value]="'test1'">Test1</fdp-option>
            <fdp-option id="fdtest2" [value]="'test2'">Test2</fdp-option>
            <fdp-option id="viewValue1" [value]="'viewValue1'" [viewValue]="'testViewValue'">Test3</fdp-option>
            <div>
                <fdp-option id="nestedOption" [value]="'nestedOption'">Nested</fdp-option>
            </div>
        </fdp-select>

```

We are not able to do this, programatically. 

This is why initial discussion that we need to have 


```
<fdp-select [(value)]="wrapperValue" [options]="MyPurchaseOrderList">
</fdp-select>
```

And I am done. Please see original discussion in the dropdown technical design.

fkolar:
Also we discussused that dropdown (select) needs to have an option to select (no value).

To get some ideas, you can check out what we have already in platform, it could be done better if we look at this 1.5 year from now, but here are several components. one of them is dropdown where we hide all internally and you can just pass list of values, but still have freedom to customize:

https://github.com/SAP/angular-metaui/blob/master/packages/components/test/widgets/dropdown/dropdown.component.spec.ts

https://github.com/SAP/angular-metaui/blob/master/packages/components/src/widgets/dropdown/dropdown.component.ts

Lokanathan-k:
frank the above code implementation is on the platform side when we implement the code, in application only this is what we need to do. 

<fdp-select id="default_select" class="fd-select-example" [(value)]="selectThreeValue" [placeholder]="'Select from values'"[dropdownValues]="options">
</fdp-select>

this is implemented as per the comments shared previously

Issue #1442, opened 2019-10-21
Platform: Change loading icon on SearchInputComponent


KevinOkamoto:
#### Is this a bug, enhancement, or feature request?
Feature.

#### Briefly describe your proposal.

The current loading icon for the SearchInputComponent does not use the standard loading spinner for fundamental-styles.

***CURRENT:***
<img width="263" alt="Screen Shot 2019-10-21 at 12 18 00 PM" src="https://user-images.githubusercontent.com/48103491/67235898-e6bd8900-f3fc-11e9-8eeb-5830bc64261d.png">

***PROPOSED:***
<img width="234" alt="Screen Shot 2019-10-21 at 12 17 35 PM" src="https://user-images.githubusercontent.com/48103491/67235913-f1781e00-f3fc-11e9-8baa-16ac09c10aa9.png">

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)

Angular 8
Fundamental-NGX v0.12.0

harikusumba:
@KevinOkamoto FYI, We made a similar customization for the the multi-input component but are using the refresh icon - double arrow with states, blue: drop-down arrow, blue: processing with the spinner, red - error spinner. I would like to know if you also intend to add the error state.

KevinOkamoto:
This issue is obsolete as SearchInput has been replaced by SearchField. Also the Fiori 3 requirements do not feature a loading spinner.

PR #1460, opened 2019-10-23
fix(platform: search-input): Change loading icon to "loading spinner" graphic

KevinOkamoto:

#1442

Changing the loading icon of search input to the "loading spinner" graphic, found here: https://sap.github.io/fundamental-ngx/loadingSpinner

***BEFORE:***
Screen Shot 2019-10-23 at 12 07 42 PM

***AFTER:***
Screen Shot 2019-10-23 at 12 07 09 PM

- [ ] `README.md`
- [ ] Documentation Examples

### DIFF
--- a/libs/platform/src/lib/components/search-input/search-input.component.scss
+++ b/libs/platform/src/lib/components/search-input/search-input.component.scss
@@ -165,11 +165,6 @@
     animation: sap-icon-spin 2s infinite linear;
 }
 
-.is-loading ::ng-deep .fd-input-group__addon:hover .fd-button:before {
-    content: "\e03e";
-    animation: none;
-}
-
 @keyframes sap-icon-spin {
     0% {
         transform: rotate(0deg);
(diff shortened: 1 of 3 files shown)


libs/platform/src/lib/components/search-input/search-input.component.scss
```diff
@@ -161,23 +161,90 @@
...
+            width: $fd-spinner-width--bar;
+            background-color: #0a6ed1;
+        }
+
+        .fd-spinner__body {
+
+            &::before,
+            &::after {
+                width: $fd-spinner-width--bar;
+                background-color: #0a6ed1;
```
kavya-b: any reason why at some places background-colour is used like this `background-color: #0a6ed1;` but in places like line 224 it is `background-color: var(--fd-color-action-1, #0a6ed1);` being used as a fallback color? Just trying to understand.

droshev:
@KevinOkamoto  this could be a good feature in core. Maybe you can bring it there too

libs/platform/src/lib/components/search-input/search-input.component.scss
droshev: I would advice you if you set colors to use the css variables from https://github.com/SAP/theming-base-content/blob/master/content/Base/baseLib/sap_fiori_3/css_variables.css

KevinOkamoto:
> @KevinOkamoto this could be a good feature in core. Maybe you can bring it there too. Btw do you have the doc/playground working so i can see it live?

OK. I'll look into adding it to the core.

I don't have the documentation done yet. I believe we just got the initial "platform" doc folders merged.

libs/platform/src/lib/components/search-input/search-input.component.scss
KevinOkamoto: The CSS variables are applied further down in the file (line 218) starting with `@supports(--css: variables)`.

For IE, which does not support the `var()` syntax, you need to have a direct assignment, i.e.
```
            background-color: #0a6ed1;
```

For browser which can understand the `var()` syntax, we place the variable definition in the `@supports(--css: variables)` block. This variables block needs to be defined after (or below) the hardcoded definition, so that the "cascading" CSS variable rules can override.

KevinOkamoto: On second thought, I think I may try to import the SCSS color variable definitions from core, so that I don't have the hardcoded values here, and I still get IE compatible CSS after SCSS compilation.

KevinOkamoto: It looks like `fundamental-styles` doesn't distribute their SCSS source files, so I won't be able to import the SCSS color definitions.

droshev: @KevinOkamoto fundamental-styles doesn't own any colors. They are coming from the theme package https://github.com/SAP/theming-base-content/blob/master/content/Base/baseLib/sap_fiori_3/css_variables.css

If you want to have IE support - you can do the same way as we do in fundamental-styles - with PostCSS to compile 2 rows:
```
  color: #32363a;
  color: var(--sapTextColor,#32363a);
```

KevinOkamoto:
@droshev I've updated the SCSS to use SAP theming variables.

fkolar:
We will get back to this, SAP is re-doing loading spiner and we need to do some additional work on our side. 
Moving to backlong.

Issue #1461, opened 2019-10-23
Platform: Placeholder text in SearchInputComponent is not properly vertical aligned in Firefox


KevinOkamoto:
#### Is this a bug, enhancement, or feature request?
Bug.

#### Briefly describe your proposal.
Fix vertical alignment issue for placeholder text in Firefox.

***Screenshot:***
<img width="257" alt="Screen Shot 2019-10-23 at 12 31 00 PM" src="https://user-images.githubusercontent.com/48103491/67427453-01286b80-f591-11e9-98fe-c57e95db1721.png">

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
Angular 8
Fundamental NGX v0.12.0

#### If this is a bug, please provide steps for reproducing it.
Load page with SearchInputComponent in Firefox browser.

KevinOkamoto:
This issue is obsolete, as there is a new component SearchField.

PR #1462, opened 2019-10-23
fix(platform:search-input): Fix vertical alignment of placeholder text

KevinOkamoto:
Fixes: #1461 

Adjust CSS rules for input field placeholder, so that the `height` and `line-height` are properly synced.

***BEFORE:***
Screen Shot 2019-10-23 at 3 36 25 PM

***AFTER:***
Screen Shot 2019-10-23 at 3 35 48 PM

- [ ] `README.md`
- [ ] Documentation Examples

### DIFF
--- a/libs/platform/src/lib/components/search-input/search-input.component.scss
+++ b/libs/platform/src/lib/components/search-input/search-input.component.scss
@@ -34,6 +34,14 @@
         padding-left: 12px;
         border-color: #89919a;
         background: none;
+
+        &::placeholder {
+            line-height: 36px;
+        }
+
+        &.fd-input--compact::placeholder {
+            line-height: 28px;
+        }
     }
 
     ::ng-deep [type=search]+.fd-input-group__addon {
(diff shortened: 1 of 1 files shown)


droshev:
@KevinOkamoto I would try to fix those issues in `fundamental-styles` instead of here

KevinOkamoto:
> @KevinOkamoto would you try to fix those issues in `fundamental-styles` instead of here?

Yes, I'll see if I can fix it in fundamental styles.

fkolar:
@KevinOkamoto just explained me that he has this fix to this but it might he also have a fix to fundamentals styles to fix this at the source. 

When he is able to finalize the fix for `styles` he will be able to push in there. 

1) Do we need this specific fix now? Or can we wait untill we consume newer version of `styles` and have the fix ready? 

2) Do we want to go on with this fix and once we have it in styles remove it from here ?

@droshev which approach you would go with ?

droshev:
@KevinOkamoto can you move `(platform:search-input)` after `:`

KevinOkamoto:
This has been addressed in fundamental styles: https://github.com/SAP/fundamental-styles/pull/406

PR #1470, opened 2019-10-24
Docs(platform): Documentation structure creation in fundamental-ngx/platform rebased

sKudum:
It has the platform documentation structure created along with button component templates added.


apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
```diff
@@ -0,0 +1,106 @@
...
+</component-example>
+<code-example [exampleFiles]="buttonHtmlIcon"></code-example>
+<separator></separator>
+
+<fd-docs-section-title [id]="'Button-state'" [componentName]="'button'">
+    Button State
+</fd-docs-section-title>
+<description>
+    'disabled' true makes the button disabled when we make it false it won't be disabled any more, by default disabled
+    is false
```
JKMarkowski: Can you change this sentence to:
<code>[disabled]="true"</code> makes the button disabled, when we make it false, it won't be disabled any more, by default disabled is set to false.

sKudum:
> I see in the content only a button. Is this intentional?

Hi Deno, Yes it has button examples only as of now.

Issue #1474, opened 2019-10-24
Modal height and scrollable body issue


InnaAtanasova:
#### Is this a bug, enhancement, or feature request?
bug
#### Briefly describe your proposal.
When we set the height of the modal, the body doesn't reflect this change. We can either allow the user to set the height of the scrollable body or dynamically calculate it.
#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
current versions
#### If this is a bug, please provide steps for reproducing it.
![Screen Shot 2019-10-24 at 10 35 38 AM](https://user-images.githubusercontent.com/39598672/67496637-b3703980-f64a-11e9-82f1-e0cf4f4c2caa.png)
In this case the height of the modal is set to 350px, the modal body remains the same causing the modal footer to take the remaining space.

PR #1483, opened 2019-10-25
fix: Modal height example issues. Close icon fix.

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1474
I changed the way how there are added classes, when modal is generated `as component`. There is also fixed `max-height` value for whole modal. 
I added some classes to the icon element, following fundamental-styles. 
There is changed example a little bit, with height specified.

- [x] `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/modal/modal.component.ts
+++ b/libs/core/src/lib/modal/modal.component.ts
@@ -144,7 +144,7 @@ export class ModalComponent extends AbstractFdNgxClass implements OnInit, AfterV
         if (!elementRef) {
             return;
         }
-        this.HOST_COMPONENT_CLASS_LIST.forEach(className => elementRef.nativeElement.classList.add(className));
+        elementRef.nativeElement.classList.add(...this.HOST_COMPONENT_CLASS_LIST);
     }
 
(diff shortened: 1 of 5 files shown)


libs/core/src/lib/modal/modal.component.ts
```diff
@@ -132,6 +140,13 @@ export class ModalComponent extends AbstractFdNgxClass implements OnInit, AfterV
         this.componentRef = this.containerRef.createEmbeddedView(content, context);
     }
 
+    private addClassesToComponent(elementRef: ElementRef): void {
+        if (!elementRef) {
+            return;
+        }
+        this.HOST_COMPONENT_CLASS_LIST.forEach(className => elementRef.nativeElement.classList.add(className));
```
MattL75: Maybe consider using?
```
elementRef.nativeElement.classList.add(...this.HOST_COMPONENT_CLASS_LIST);
```

JKMarkowski: Actually, your example looks better.

PR #1504, opened 2019-10-30
docs: Change way how we recognize library, change switch library button

JKMarkowski:
I added providers with names of libraries in our modules, to detect, on which library user currently is. Before it was calculated by checking url string.
Also there is changed switcher between libraries

- [x] `README.md`- **not needed**
- [x] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes) **not needed**
- [x] Documentation Examples **not needed**

### DIFF
--- a/apps/docs/src/app/core/core-documentation.module.ts
+++ b/apps/docs/src/app/core/core-documentation.module.ts
@@ -756,6 +756,9 @@ import { ComboboxOpenControlExampleComponent } from './component-docs/combobox/e
         SharedDocumentationModule,
         MarkdownModule.forChild(),
         RouterModule.forChild(ROUTES)
+    ],
+    providers: [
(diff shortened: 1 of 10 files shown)


droshev:
how did we agree to move on with that?

MattL75:
I think this looks way better than the old toggle. Still kinda weird, but not sure what the best way would be. At least this looks clean

fkolar:
Looks good. Unless you employ some state management solution which we should do as well (another library in the ngx) then you dont have much choices, Provider here is good.  


You still can create some `@Injectable` AppConfig  which registers at root level and you will store, and comunicate all the things thru this service.

MattL75:
@JKMarkowski I like the dropdown picker. Stays out of your way and it's clear where you currently are.

PR #1518, opened 2019-10-31
feat: (platform and core)adding new component select under platform

Lokanathan-k:

feat(platform): adding new component select under platform issue details:#1268, 
its a wrapper component across the core select module to create a new component handle added with the new features.

in the core component there i have added option to add new select button type where user can add [selectType] and [gylph] assign diff types of select type and render the ui accordingly.

for this change i have added changes in select.component.ts and  select.component.scss

- [ ] the commit message follows the guideline:
- [ ] tests for the changes that have been done

Documentation checklist:
- [ ] update `README.md`
- [ ] Documentation Examples


libs/core/src/lib/select/select.component.html
```diff
@@ -1,31 +1,29 @@
...
-        class="fd-select-popover-body-custom"
-        [style.maxHeight]="maxHeight || (calculatedMaxHeight + 'px')">
-        <fd-menu>
-            <ul fd-menu-list>
-                <ng-content></ng-content>
-            </ul>
-        </fd-menu>
-    </fd-popover-body>
-</fd-popover>
+<fd-popover [(isOpen)]="isOpen" (isOpenChange)="isOpenChangeHandle($event)" [options]="popperOptions"
```
droshev: @Lokanathan-k Could you revert formatting changes so we can only see what was changes

libs/core/src/lib/select/select.component.ts
```diff
@@ -146,10 +152,10 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnDestroy,
     ngOnChanges(changes: SimpleChanges): void {
         if (changes.value) {
             setTimeout(() => {
-                if (this.value) {
+               if (this.value) {
```
JKMarkowski: Are you using `3 spaces` indent ?

libs/core/src/lib/select/select.component.scss
```diff
@@ -93,3 +94,29 @@
...
+}
+.fd-button--regularborder::after{
+    border: none !important;
+}
+.fd-button--splitborder{
+   border:1px solid !important;
+}
+.fd-button--splitborder::after{
+    border-left-style: solid !important;
+    border-left-width: 1px !important;
```
JKMarkowski: I'm not sure about `!important` here

libs/core/src/lib/select/select.component.ts
```diff
@@ -69,6 +69,12 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnDestroy,
     @Input()
     maxHeight: string;
 
+    @Input()
```
JKMarkowski: Comments are mandatory for any `non-private` property.

fkolar:
@Lokanathan-k  also when I check your PR and description it does not mention anything about modifying core select component. 
Can you add some brief description why are you changing core select and what you are adding to it? So we can focus our review accordingly. I

fkolar:
Common comments:
 1) When you see you have more then 4 comments under PR do not try to submit more, squash them locally (always rebase before this) here is good source: for rebasing: https://gist.github.com/Chaser324/ce0505fbed06b947d962

2) We want to have some naming conversion like @KevinOkamoto mentioned above but it does not mean we want to put a `platform` into the package name like here: `select platform`

Lokanathan-k:
> Common comments:
> 
> 1. When you see you have more then 4 comments under PR do not try to submit more, squash them locally (always rebase before this) here is good source: for rebasing: https://gist.github.com/Chaser324/ce0505fbed06b947d962
> 2. We want to have some naming conversion like @KevinOkamoto mentioned above but it does not mean we want to put a `platform` into the package name like here: `select platform`
_Done_

> 3. The same namving convetion applies to name of the components update this one too `SelectPlatformComponent`

_Yes i can see that but if i name the same component  again with the same name as select i coud see conflict in controllers so i have named it like this._ 

> 4. Missing comments for class and fields.
_done_
> 5. This is no-go. We discussed this in the call remember? I sent you recordings. We dont pass template as input. How would you do this on the application level? You just have bindings that you never tested? and you are submitting PR.
> 
> ```
> @Input()
>     itemTemplate: TemplateRef<any>;
> ```
> 
_Done_
> Please do not try to submit anything unless you are sure that every single piece you put in is tested and you can guarantee it make sense.
> 
> Plus if you have work in progress and you just want to share it with other do name this PR as feat name it as `WIP: title`
> 
> -> We discussed many things on the call and I dont see them here.

fkolar:
@Lokanathan-k  please use my implementation code on sent on slack, and jus try to test it and create docs, please.

libs/core/src/lib/select/select.component.scss
```diff
@@ -93,3 +94,37 @@
         display: block;
     }
 }
+
+// no border class csss overrides the border of the button after adding important 
+.fd-button--noborder{
+   border: none  !important;
```
droshev: can't we use light button instead?

libs/core/src/lib/select/select.component.ts
```diff
@@ -131,10 +139,10 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnDestroy,
     }) as Observable<OptionComponent>;
 
     /** @hidden */
-    onChange: Function = () => {};
+     onChange: Function = () => { };
```
droshev: can we do those code formatting in a separate PR?

libs/core/src/lib/select/select.component.scss
Lokanathan-k: i tried using the light button but css doesn't overrides the look and feel, and also we have 3 other types of button for this component

fkolar: @Lokanathan-k  why we even changing core class ?
 is this is the reason? 

`in the core component there i have added option to add new select button type where user can add [selectType] and [gylph] assign diff types of select type and render the ui accordingly.
`

Can we just use triggerTemplate for this ? Where application can define their own format do we need to change the core at all? 

Also we should not be adding any enhacements that comes from our products. This is new direction. 

This PR is getting little bit messy... It woudl be cleaner to create separate PRs that updates A, and PR that changes B.

Lokanathan-k: we are not changing the core class but just adding another class with the property that  needs to be overridden.

when we write the class on the wrapper component it doesn't get overridden it add another layer of styling.

fkolar:
Closing this one as there will be brand new pr just for the platform select to make it cleaner.

Issue #416, opened 2019-01-16
Add 'id' attribute, 'information' and 'success' types for alerts


InnaAtanasova:
#### Is this a bug, enhancement, or feature request?
v1.4.1 of Fiori Fundamentals support 'information' and 'success' types for alerts. 
The user should be able to provide their own 'id' for the alert.

PR #1549, opened 2019-11-06
fix: popover-placement-example fix for the control elements

InnaAtanasova:
### Please provide a brief summary of this pull request.
The example for popover placement is broken. All control elements are situated together. 
The styling for `fd-col-#` is in `layout.css but it doesn't bring the styling to the component. We need to import the whole lib to make it work. 
Until we find the problem in Fundamental-styles we are going to use this solution.

Before:
Screen Shot 2019-11-06 at 9 44 53 AM

After:
Screen Shot 2019-11-06 at 9 44 40 AM

Documentation checklist:
- [x] Documentation Examples


InnaAtanasova:
> Size of these popovers are `100vh` I guess. It's a little bit too big.
> it's solved here: [SAP/fundamental-styles#416](https://github.com/SAP/fundamental-styles/pull/416) So we can merge.

You are right, it's solved in the #416. The issue with the menu size was because of `display:table`. I changed it to `display: block`

Issue #1586, opened 2019-11-13
Add loading spinner when data is not available 


stefanoScalzo:
#### Is this a bug, enhancement, or feature request?
enhancement
#### Briefly describe your proposal.
For the select, when the data is not made available there should be a loading spinner to show that there will be nothing inside the dropdown container
#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
Angular 8
NGX "0.13.0-rc.29"

PR #1587, opened 2019-11-13
fix: (core)loading spinner for select

stefanoScalzo:
fixes: #1586
Adding a new @Input decorator called loading which receives a boolean true or false. This boolean will be sent from the parent component with "theirData==null". If the data has not arrived a spinner will be shown else it will show the navigation-down-arrow. By default if they do not use [loading] it will set default to showing buttons

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


libs/core/src/lib/select/select.component.ts
```diff
@@ -159,9 +167,12 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnDestroy,
         }
     }
 
+    ngOnInit() {
+        // console.log(this.loading);
+    }
+
```
JKMarkowski: Let's remove it

libs/core/src/lib/select/select.component.html
```diff
@@ -1,31 +1,24 @@
...
-            <button class="fd-select-button-custom"
-                    fd-button
-                    aria-haspopup="true"
-                    [ngClass]="{'fd-button--compact': compact}"
-                    [attr.aria-expanded]="isOpen"
-                    [disabled]="disabled">
-                <span class="fd-select-text-custom">{{triggerValue}}</span>
+            <button fd-button class="fd-select-button-custom" aria-haspopup="true" [ngClass]="{
+                    'fd-select-button-custom--loading': loading,
+                    'fd-button--compact': compact
```
JKMarkowski: Maybe let's change it to `[compact]="compact", it's been possible since we've got `fd-button` everywhere

stefanoScalzo: changed

libs/core/src/lib/select/select.component.ts
```diff
@@ -119,6 +126,9 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnDestroy,
     /** Subject triggered when the component is destroyed. */
     private readonly destroy$: Subject<void> = new Subject<void>();
 
+    @Input() loading: boolean = false;
```
MattL75: add doc comments?

MattL75: also might want to group this input with all the other inputs in this file, and follow the whole 2-line input thing, just for cleanliness

stefanoScalzo: Were waiting for designers for further implimentation

libs/core/src/lib/select/select.component.scss
```diff
@@ -80,6 +80,52 @@
                 }
             }
         }
+        &--loading {
+            &::after {
+                color: var(--sapUiContentIconColor, #0854a0);
```
droshev: we should use postCSS for the css variables - like in `fundamental-styles`

libs/core/src/lib/select/select.component.scss
```diff
@@ -80,6 +80,52 @@
...
+            @keyframes spin {
+                100% {
+                    -webkit-transform: rotate(360deg);
+                    transform: rotate(360deg);
+                }
+            }
+            &:focus {
+                &::after {
+                    border: none;
+                    color: var(--sapUiContentIconColor, #0854a0) !important;
```
droshev: same as above

libs/core/src/lib/select/select.component.scss
```diff
@@ -80,6 +80,52 @@
...
+                    background-color: var(--sapUiFieldBackground, #ffffff) !important;
```
droshev: same here

libs/core/src/lib/select/select.component.scss
```diff
@@ -80,6 +80,52 @@
...
+                }
+            }
+            &:hover {
+                &::after {
+                    border: none;
+                    background-color: var(--sapUiFieldBackground, #ffffff) !important;
```
droshev: same

libs/core/src/lib/select/select.component.scss
```diff
@@ -80,6 +80,52 @@
...
+                }
+            }
+            &:active {
+                &::after {
+                    border: none;
+                    background-color: var(--sapUiFieldBackground, #ffffff);
```
droshev: same

libs/core/src/lib/select/select.component.scss
```diff
@@ -80,6 +80,52 @@
...
+                }
+            }
+            &:active,
+            &.is-active,
+            &[aria-expanded='true'] {
+                &::after {
+                    color: var(--sapUiContentIconColor, #0854a0) !important;
```
droshev: same

InnaAtanasova:
It's labelled as feat. Could you describe the breaking changes in the Description section?

Issue #1633, opened 2019-11-20
Nested fd-popover element inherits predecessor's styles


jedrzejiwanicki:
#### Is this a bug, enhancement, or feature request?

Bug

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)

0.12.0

#### If this is a bug, please provide steps for reproducing it.

Nest two fd-popover inside each other providing different placement attributes. Second popover will be broken

#### Is there anything else we should know?

Additional request is to provide possibility to add a class to `fd-popover-container`. It is possible to select it via `fd-popover` itself but in case of nested popovers it also inherits prececessor's styles.

PR #1646, opened 2019-11-21
fix: (core) Add Additional class to popover component, introduce nested popover example

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1633
There is added new property on popover, to add custom classlist for popover container. It was missing mostly, when user wanted to put popover inside popover, to differentiate them.
I added also interesting use case with `nested-popover`, as example.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


InnaAtanasova:
To be honest I don't know how I feel about this nested popover...
In terms of code it looks good. About the look... I think we need some design before releasing this. 
@droshev what do you think about this?

Screen Shot 2019-11-25 at 2 54 24 PM

JKMarkowski:
Hi @InnaAtanasova, this was more to show how user can use class names injection.

droshev:
I agree with @InnaAtanasova. It looks odd the way it appears.

JKMarkowski:
Nested class example is removed.

apps/docs/src/app/core/core-documentation.module.ts
```diff
@@ -756,7 +756,7 @@ import { SelectTypesExampleComponent } from './component-docs/select/examples/se
         NotificationContentComponent,
         DatePickerComplexI18nExampleComponent,
         DatetimePickerComplexI18nExampleComponent,
-        PopoverDropdownComponent
+        PopoverDropdownComponent,
```
InnaAtanasova: unnecessary comma

Issue #1536, opened 2019-11-01
Update Calendar markup for a11y


jacobdevera:
#### Is this a bug, enhancement, or feature request?
Enhancement (to comply with a11y)

#### Briefly describe your proposal.
The rendered markup should be updated to reflect changes made in `fundamental-styles`.
https://github.com/SAP/fundamental-styles/pull/410
- Convert month/year selector `ul` into `table` with role=“grid”
- Currently selected date: `<td>`: aria-selected=“true”  and `<span>` role "button"

Additional changes:
- Month/year header: aria-live=“polite” (to announce month/year changes)
- Navigation buttons: Needs localizable `aria-label` (e.g. “Next Year”, "Next Month")
- Selectable dates: `aria-label` should be used to read out full date
- Disabled dates: `<td>`: `aria-disabled="true"`

PR #1649, opened 2019-11-21
fix: (core) Calendar A11y - change ul/li to table with grid

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1536
There is changed the way how we build the calendar month and year components. Also there is changed the way how the models are generated. There is also added one missing `aria-live`

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/calendar/calendar-header/calendar-header.component.html
+++ b/libs/core/src/lib/calendar/calendar-header/calendar-header.component.html
@@ -20,6 +20,7 @@
                     [attr.aria-label]="calendarI18nLabels.monthSelectionLabel"
                     [attr.aria-selected]="isOnMonthView()"
                     (click)="processViewChange('month')"
+                    aria-live="polite"
                     type="button">
                 {{monthLabel}}
             </button>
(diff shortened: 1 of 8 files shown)


libs/core/src/lib/calendar/calendar-header/calendar-header.component.html
```diff
@@ -20,6 +20,7 @@
                     [attr.aria-label]="calendarI18nLabels.monthSelectionLabel"
                     [attr.aria-selected]="isOnMonthView()"
                     (click)="processViewChange('month')"
+                    aria-live="polite"
```
jacobdevera: Having multiple `aria-live` can cause conflicts. There should just be one `aria-live` on a parent element (e.g. the `<header>` would work).

jacobdevera:
Awesome! Something I didn't catch: Disabled dates should have `aria-disabled="true"` for its `<td>`

JKMarkowski:
Hello @jacobdevera, thanks for pointing it out :) It helped a lot during implementation. Attribute `aria-disabled` is added to day's view td.

PR #1694, opened 2019-12-02
feat(platform): select component  for platform to enhance the core

Lokanathan-k:
feat(platform): adding new wrapper select component  for platform

wrapper component for the select component, with pluggable options and diff css for selectTypes

- [ ] the commit message follows the guideline:
- [ ] tests for the changes that have been done

Documentation checklist:
- [ ] update `README.md`
- [ ] Documentation Examples

### DIFF
--- a/libs/platform/src/lib/components/select/select.component.ts
+++ b/libs/platform/src/lib/components/select/select.component.ts
@@ -1,0 +1,134 @@
+import { 	
+    Component,
+    OnInit,
+    Input,
+    forwardRef,
+    ChangeDetectionStrategy,
+    ChangeDetectorRef,
+    ContentChild,
+    TemplateRef,
+    AfterContentInit } from '@angular/core';
+import { NG_VALUE_ACCESSOR } from '@angular/forms';
+import { OptionComponent, SelectComponent as fdSelect } from '@fundamental-ngx/core';
+/**
+ * Interface SelectItem is used to deal with complex object in order to be able to format custom label that is
+ * shown in the options.
+ */
+export interface SelectItem {
+    /**
+     * Item text shown in the popup
+     */
+     label: string;
+
+    /**
+     * References to the object instance
+     */
+    value: any;
+    disabled?: boolean;
+
+    /**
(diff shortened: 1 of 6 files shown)


libs/platform/src/lib/components/select/select.component.ts
```diff
@@ -0,0 +1,134 @@
...
+    styleUrls: ['./select.component.scss'],
+    providers: [
+        {
+            provide: NG_VALUE_ACCESSOR,
+            useExisting: forwardRef(() => SelectPlatformComponent),
+            multi: true
+        }],
+    changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class SelectPlatformComponent extends fdSelect implements OnInit, AfterContentInit  {
```
KevinOkamoto: Is there a reason we are "extending" the core SelectComponent? It seems like overkill since we are utilizing the core select component in the template.

libs/platform/src/lib/components/select/select.component.ts
```diff
@@ -0,0 +1,134 @@
...
+ /**
+     * Form element ID.
+     * Todo: This should be moved to higher class that will be common to all input fields
+     */
+    @Input()
+    id: string;
+
+    /** variable to input any type of object. */
+    @Input()
+    list: Array<any>;
```
KevinOkamoto: Shouldn't this be an array of SelectItems:
```ts
list: Array<SelectItem>
```

libs/platform/src/lib/components/select/select.component.ts
Lokanathan-k: @KevinOkamoto  it was frank suggestion not to recreate the variables instead overload the parent core class and use the variables in the platform component

libs/platform/src/lib/components/select/select.component.ts
Lokanathan-k: Yes, i will make the change and push it, thank you

fkolar:
Let's address those small changes otherwise it looks good.

Issue #1710, opened 2019-12-06
Platform: address Menu component library review comments


kavya-b:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
Address PR #1563 comments by Frank: https://github.com/SAP/fundamental-ngx/pull/1563#issuecomment-553385144
Additionally, 
1. Need to use fd-icon instead of span classes
2. Using dropdown popover with button instead of deprecated fd-dropdown-control

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
current

kavya-b:
Fixed in #1711

PR #1563, opened 2019-11-08
docs:(platform:menu) Documentation for Menu component

kavya-b:
#### Please provide a link to the associated issue.
#1562 
#### Please provide a brief summary of this pull request.
Adding documentation pages for Menu component and additionally,
1. using fiori css variables to refer to colors in scss files
2. fixed issue where focus style was not applying in Chrome browser correctly

#### Please check whether the PR fulfills the following requirements

- [x] the commit message follows the guideline:
https://github.com/SAP/fundamental-ngx/blob/master/CONTRIBUTING.md -done
- [x] tests for the changes that have been done -NA

Documentation checklist:
- [x] update `README.md`
- [x] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)-NA
- [x] Documentation Examples

PR #1711, opened 2019-12-06
fix(platform): menu component library review fixes

kavya-b:
https://github.com/SAP/fundamental-ngx/issues/1710

Address PR #1563 comments by Frank: https://github.com/SAP/fundamental-ngx/pull/1563#issuecomment-553385144
Additionally,
1. Need to use fd-icon instead of span classes
2. Using dropdown popover with button instead of deprecated fd-dropdown-control

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/platform/src/lib/components/menu/menu-component/menu-item.component.ts
+++ b/libs/platform/src/lib/components/menu/menu-component/menu-item.component.ts
@@ -69,7 +69,7 @@ export class MenuItemComponent implements DefaultMenuItem, OnChanges {
 
     /** @hidden */
     ngOnChanges(changes: SimpleChanges): void {
-        if (changes.itemWidth || changes.item) {
+        if (changes.itemWidth || changes.item || changes.selected) {
             this.finalItemWidth = this.getItemWidth();
         }
     }
@@ -80,7 +80,7 @@ export class MenuItemComponent implements DefaultMenuItem, OnChanges {
     getItemWidth(): string {
         // todo: handle em, rem etc.
         let finalItemWidth: string = '';
-        if (this.itemWidth !== undefined) {
+        if (this.itemWidth) {
             finalItemWidth = this.getOffsetItemWidth(finalItemWidth);
         } else {
             const itemElement = this.itemEl.nativeElement;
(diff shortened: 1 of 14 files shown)


libs/platform/src/lib/components/menu/menu-component/menu-item.component.ts
```diff
@@ -69,7 +69,7 @@ export class MenuItemComponent implements DefaultMenuItem, OnChanges {
 
     /** @hidden */
     ngOnChanges(changes: SimpleChanges): void {
-        if (changes.itemWidth || changes.item) {
+        if (changes.itemWidth || changes.item || changes.selected) {
             this.finalItemWidth = this.getItemWidth();
```
KevinOkamoto: I think you may be able to remove the `getItemWidth` method and the `finalItemWidth` property using the CSS `calc()` method: https://developer.mozilla.org/en-US/docs/Web/CSS/calc

The `calc()` method is compatible with all browsers and allows you to add together mixed units.

You could probably change line 31 in the `menu-item.component.html` template file to:

```
<div class="fd-menu__label-truncation" [ngStyle]="{ width: 'calc(' + itemWidth + ' + 46px )' }">
```

kavya-b: Okay, I will try this and update.

kavya-b: I have been trying this, Kevin, and it is not as straightforward to do it in this particular case because I have to take care of three things: 
1. the width of the item when an icon is selected or not selected(offset by 46px)
2. the width of the item when a normal icon(not selectable) is used(offset by 46px)
3. the width of the secondary icon if it is used. 
If both primary and secondary icons are used, I need to offset by 88px, but if any one icon is used, I need to offset by 46px, and for no icons there is no offset. 
I tried using calc in css and using ngClass instead to give different offsets, but I cannot access the [previous sibling](https://stackoverflow.com/questions/1817792/is-there-a-previous-sibling-selector) so getting the 3rd case to work seamlessly is causing some problems.  There are other edge cases as well which I had worked out programmatically but doing it through CSS alone would be all the more complicated.

KevinOkamoto: OK. It sounds like your use case is complicated enough that it needs this added javascript code. Keep your original solution.

PR #1743, opened 2019-12-18
fix: Add exports to public_api, add @angular/cdk to schematics

JKMarkowski:

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


libs/core/src/lib/popover/public_api.ts
```diff
@@ -1,2 +1,7 @@
 export * from './popover.module';
 export * from './popover-directive/popover.directive';
+export * from './popover-body/popover-body.component';
+export * from './popover-directive/popover-container';
+export * from './popover-directive/popover.directive';
```
MattL75: exporting popover-directive twice

JKMarkowski: Thanks for comment @MattL75

Issue #1758, opened 2019-12-23
Alert Component 0.4.0


JKMarkowski:
Alert Component has got different structure right now, there is `fd-button fd-button--light fd-button--compact ` class on `close` element. Check [here](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes#alert)

PR #1768, opened 2019-12-30
fix: (core:Alert) Add new button classes to alert component

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/1758
There was changes in `styles@0.4.0` about the close button on alert component. There are new classes added. 
Due to `AbstractFdxNgClass` usage, which changes the hierarchy of classes I had to add some additional styles, to make it look the same as on styles. 

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/alert/alert.component.scss
+++ b/libs/core/src/lib/alert/alert.component.scss
@@ -2,6 +2,7 @@
 
 .fd-alert {
     display: block;
+    line-height: 1rem;
 }
 
 
@@ -11,4 +12,10 @@
 
 .fd-has-display-none {
     display: none;
+}
+
+.fd-alert__close {
+    border-width: 1px;
+    border-style: solid;
+    border-color: transparent;
 }
--- a/libs/core/src/lib/alert/alert.component.scss
+++ b/libs/core/src/lib/alert/alert.component.scss
@@ -2,7 +2,6 @@
 
 .fd-alert {
     display: block;
-    line-height: 1rem;
 }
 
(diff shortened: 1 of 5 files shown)


libs/core/src/lib/alert/alert.component.scss
```diff
@@ -2,6 +2,7 @@
 
 .fd-alert {
     display: block;
+    line-height: 1rem;
```
InnaAtanasova: This should be coming from Fundamental-styles

libs/core/src/lib/alert/alert.component.scss
```diff
@@ -12,3 +13,9 @@
 .fd-has-display-none {
     display: none;
 }
+
+.fd-alert__close {
+    border-width: 1px;
+    border-style: solid;
+    border-color: transparent;
+}
```
InnaAtanasova: This should be coming from Fundamental-styles

libs/core/src/lib/alert/alert.component.scss
JKMarkowski: Your right, it needed a class `fd-alert__text` wrapper.

libs/core/src/lib/alert/alert.component.scss
JKMarkowski: There are some issues with button classes, it's solved in https://github.com/SAP/fundamental-ngx/pull/1770

Issue #1759, opened 2019-12-23
Weird class hierarchy behaviour, when AbstractFdNgxClass is used


JKMarkowski:
#### Is this a bug, enhancement, or feature request?
Bug
#### Briefly describe your proposal.
We should fix or get rid of `AbstractFdNgxClass`. It provides a unexpected changes on class hierarchy.
So having example of most used `fd-button` component. Sometimes we want to add different classes manually, by passing `class=""`. Classes added by `class=""` are shadowed, by classes from `AbstractFdNgxClass` method. We should keep the hierarchy of classes, because otherwise we will need to add some styles manually.
#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
Newest

rengare:
working on it

droshev:
@JKMarkowski could you document more information regarding the issue?

JKMarkowski:
@droshev Sure, It's on my way.

PR #1770, opened 2019-12-30
feat: added decorators and interfaces to replace AbstractFdNgxClass

rengare:
#1759 

- [ /] the commit message follows the guideline:
- [ /] tests for the changes that have been done


droshev:
@rengare can you update the PR title?

libs/core/src/lib/utils/decorators/apply-css-style.decorator.ts
```diff
@@ -0,0 +1,17 @@
+import { Hash, ELEMENT_REF_EXCEPTION } from '../public_api';
+
+export function applyCssStyle(target: any, propertyKey: string, descriptor: PropertyDescriptor) {
+    const originalMethod = descriptor.value
+    console.log(descriptor);
```
JKMarkowski: Please remove `console.log`

libs/core/src/lib/utils/decorators/apply-css-style.decorator.ts
```diff
@@ -0,0 +1,17 @@
+import { Hash, ELEMENT_REF_EXCEPTION } from '../public_api';
+
+export function applyCssStyle(target: any, propertyKey: string, descriptor: PropertyDescriptor) {
```
JKMarkowski: I know that in general this function is descriptive, but could you describe with  `/** */` a little why we should use this function?

libs/core/src/lib/utils/abstract-fd-ngx-class.ts
```diff
@@ -1,4 +1,5 @@
 import { ElementRef, OnChanges, OnInit, Input } from '@angular/core';
+import { debounceTime } from 'rxjs/operators';
```
JKMarkowski: Could you remove unused import?

libs/core/src/lib/button/button.component.ts
```diff
@@ -47,34 +55,45 @@ export class ButtonComponent extends AbstractFdNgxClass {
...
+
     }
 
     /** @hidden */
-    constructor(private elementRef: ElementRef) {
-        super(elementRef);
+    constructor(private _elementRef: ElementRef) {
+    }
+
+    ngOnInit() {
```
JKMarkowski: Let's add return type and some comments to functions on this file.

libs/core/src/lib/button/button.component.ts
```diff
@@ -21,17 +20,26 @@ export type ButtonOptions = 'light' | 'emphasized' | 'menu';
...
+        this._class = userClass;
+        this.buildComponentCssClass();
+    } // user's custom classes
+
+    /** The icon to include in the button. See the icon page for the list of icons. */
+    private _glyph: string
+    @Input() set glyph(icon: string) {
+        this._glyph = icon;
+        this.buildComponentCssClass()
+    };
```
JKMarkowski: This is great, but I'm not sure what about situation, when someone changes for ex. `compact` property or `fdType`. Will the new classes be applied on this element?

libs/core/src/lib/utils/abstract-fd-ngx-class.ts
```diff
@@ -39,14 +48,14 @@ export abstract class AbstractFdNgxClass implements OnInit, OnChanges {
...
-        this._setProperties();
+        // const classList = (this._elementRef.nativeElement as HTMLElement).classList;
+        // while (classList.length > 0) {
+        //     classList.remove(classList.item(0));
+        // }
+        // if (this.class) {
+        //     this._addClassToElement(this.class);
+        // }
+        // this._setProperties();
     }
```
JKMarkowski: Let's uncomment it.

libs/core/src/lib/utils/decorators/apply-css-class.decorator.ts
```diff
@@ -0,0 +1,14 @@
+import { ELEMENT_REF_EXCEPTION } from '../public_api';
+
+export function applyCssClass(target: any, propertyKey: string, descriptor: PropertyDescriptor) {
```
JKMarkowski: I know that in general this function is descriptive, but could you describe it with some comments `/** */`. Also let's describe all of these arguments.

libs/core/src/lib/button/button.component.ts
rengare: in that case, seter is needed or ngonchange although it is not recomended since every change whole css style will be rebuilded

libs/core/src/lib/utils/decorators/apply-css-class.decorator.ts
rengare: done

libs/core/src/lib/utils/abstract-fd-ngx-class.ts
rengare: done

libs/core/src/lib/button/button.component.ts
rengare: done

libs/core/src/lib/utils/abstract-fd-ngx-class.ts
rengare: import removed

libs/core/src/lib/utils/decorators/apply-css-style.decorator.ts
rengare: done

libs/core/src/lib/utils/decorators/apply-css-style.decorator.ts
rengare: done

libs/core/src/lib/button/button.component.ts
```diff
@@ -46,35 +57,58 @@ export class ButtonComponent extends AbstractFdNgxClass {
...
+     * function used by applyCssClass and applyCssStyle decorators 
+     */
+    elementRef(): ElementRef<any> {
+        return this._elementRef;
+    }
+
     /** @hidden */
-    constructor(private elementRef: ElementRef) {
-        super(elementRef);
+    private getOptionCssClass(options: string | ButtonOptions[]): string {
```
salarenko: Shouldn't be named `_getOptionCssClass`?

libs/core/src/lib/button/button.component.ts
```diff
@@ -46,35 +57,58 @@ export class ButtonComponent extends AbstractFdNgxClass {
...
+        if (Array.isArray(this.options)) {
+            return this.options.reduce((_class, option) => _class += ` fd-button--${option}`, ' ')
```
salarenko: This will produce string with additional white spaces like: `"  fd-button--1 fd-button--2 fd-button--3"`. I'd recommend using `.map()` and `.join()`

libs/core/src/lib/utils/abstract-fd-ngx-class.ts
```diff
@@ -21,6 +21,14 @@ export abstract class AbstractFdNgxClass implements OnInit, OnChanges {
     /** @hidden */
     abstract _setProperties(): void;
 
+    _setClassToElement(className: string) {
```
salarenko: Return type and shouldn't it be named `setClassToElement` (no prefix)?

libs/core/src/lib/utils/abstract-fd-ngx-class.ts
```diff
@@ -21,6 +21,14 @@ export abstract class AbstractFdNgxClass implements OnInit, OnChanges {
...
+        (this._elementRef.nativeElement as HTMLElement).classList.value = `${className} ${this.class}`;
+    }
+
+    _clearElementClass() {
```
salarenko: Return type and shouldn't it be named `clearElementClass` (no prefix)?

libs/core/src/lib/utils/decorators/apply-css-class.decorator.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Method decorator to apply css class to a component through native element
+ * decorator will store original method in variable and wrap it with custom one
+ * component has to implement CssClassBuilder interface
+ * more info abour method decorator: https://www.typescriptlang.org/docs/handbook/decorators.html#method-decorators
+ * @param target a component
+ * @param propertyKey name of the method
+ * @param descriptor method
+ */
+export function applyCssClass(target: any, propertyKey: string, descriptor: PropertyDescriptor) {
```
salarenko: Return type

libs/core/src/lib/utils/decorators/apply-css-style.decorator.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Method decorator to apply css styles to a component through native element
+ * decorator will store original method in variable and wrap it with custom one
+ * component has to implement HasElementRef or CssStyleBuilder interface
+ * more info abour method decorator: https://www.typescriptlang.org/docs/handbook/decorators.html#method-decorators
+ * @param target a component
+ * @param propertyKey name of the method
+ * @param descriptor method
+ */
+export function applyCssStyle(target: any, propertyKey: string, descriptor: PropertyDescriptor) {
```
salarenko: Return type

libs/core/src/lib/utils/interfaces/css-style-builder.interface.ts
```diff
@@ -0,0 +1,9 @@
+import { HasElementRef } from '../public_api';
+
+export interface Hash {
```
salarenko: Maybe lets move Hash interface to separate file?
I think I'ts going to be used widely in application, and it would be better to import it from '../hash' than '../css-style-builder'

Or make hash with more generic approach as:
```
export interface Hash <keyType, valueType> {
    [key: keyType]: valueType;
}
```

And use it like `Hash<string, any>`

libs/core/src/lib/button/button.component.ts
```diff
@@ -46,35 +57,58 @@ export class ButtonComponent extends AbstractFdNgxClass {
...
+        ].filter(x => x !== '').join(' ');
+    }
+
+    @applyCssStyle
+    /** CssStyleBuilder interface implementation
+     * function must return hashmap where
+     * key:string
+     * value:any
+     */
+    buildComponentCssStyle(): Hash {
```
salarenko: Isn't the component missing input interface and logic from providing inline styles for this component?

libs/core/src/lib/utils/abstract-fd-ngx-class.ts
rengare: true although other functions have the same prefix so in order to not break anything I left code style here untouched

libs/core/src/lib/button/button.component.ts
rengare: indeed, but it was out of the scope of the issue

libs/core/src/lib/utils/interfaces/css-style-builder.interface.ts
rengare: key
image

libs/core/src/lib/utils/abstract-fd-ngx-class.ts
rengare: return type added, regarding to `_` same reason as above

libs/core/src/lib/utils/interfaces/css-style-builder.interface.ts
rengare: key have to be string or number, interface has been moved to another file with generic value type

libs/core/src/lib/utils/decorators/apply-css-style.decorator.ts
rengare: return type added

libs/core/src/lib/utils/decorators/apply-css-class.decorator.ts
rengare: return type added

libs/core/src/lib/utils/abstract-fd-ngx-class.ts
rengare: return type added

libs/core/src/lib/button/button.component.ts
rengare: changes applied as suggested

libs/core/src/lib/button/button.component.ts
rengare: underscore prefix added

PR #1776, opened 2020-01-03
fix: Spacing not available in rtl view

Lokanathan-k:

closes: https://github.com/SAP/fundamental-ngx/issues/1268

when we are using the icon based select component in the rtl view there is no spacing between the icon and the placeholder so introduced a class to accomodate this change

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


JKMarkowski:
Hi @Lokanathan-k, can you put it to `.scss` file? Let's not use inline-styling anymore.

libs/core/src/lib/select/select.component.scss
```diff
@@ -108,6 +108,9 @@
         pointer-events: none;
         margin-right: auto;
     }
```
JKMarkowski: Let's just add new line between classes, then we can merge it.

libs/core/src/lib/select/select.component.scss
```diff
@@ -108,6 +108,10 @@
...
+	
+    .icon-position{
+        margin: 1px 12px
+	}
```
fkolar: missing colon. We shoudl add more restrictive lint rules.

mikerodonnell89:
The RTL toggle for the one with icons actually toggles the previous example

mikerodonnell89:
It's not a result of code changes in this PR, but I noticed the RTL toggle for the second example actually toggles the RTL for the first, due to both examples using `[name]="'ex1'"`

droshev:
@Lokanathan-k do you need more time for the PR? It is getting a bit old.
Thank you!

Lokanathan-k:
> @Lokanathan-k do you need more time for the PR? It is getting a bit old.
> Thank you!

@droshev  i dont have any code change need approval from @fkolar post that i can merge the change to the master till then i am unable to merge the code

droshev:
@fkolar do you think you can take a look?

libs/core/src/lib/select/select.component.scss
Lokanathan-k: fixed pls review @fkolar

PR #1784, opened 2020-01-07
fix: null check in decorator added

rengare:
Added a null checks for elementRef
- [/ ] the commit message follows the guideline:
- [ /] tests for the changes that have been done


InnaAtanasova:
@rengare , I believe this is not introducing Breaking Changes??? I renamed the title to fix.

Issue #1811, opened 2020-01-14
Test the latest changes before 0.13.0 release


InnaAtanasova:
We need to smoke test the latest changes we added to the library before we release 0.13.0

- bugs
- a11y
- different browsers

# Checkbox (Michal)
 * tristate - IE11

# Combobox (Jedrzej) - on PR / Partially Done
* missing icons on the input, Open State Control, rounded corners for the button (all browsers), Input groups border radius issue

# DatePicker - IE11 - Done
<img width="320" alt="Screen Shot 2020-01-15 at 9 36 26 AM" src="https://user-images.githubusercontent.com/39598672/72442773-22a58780-377b-11ea-8836-ed4a46723adc.png">

# Safari - DateTimePicker - Done
<img width="199" alt="Screen Shot 2020-01-15 at 9 44 04 AM" src="https://user-images.githubusercontent.com/39598672/72443046-a790a100-377b-11ea-9ff1-4d80dc23a79f.png">

<img width="700" alt="Screen Shot 2020-01-15 at 9 44 04 AM" src="https://user-images.githubusercontent.com/39598672/72443698-b2980100-377c-11ea-8b10-f79ceeb08449.png">

# Input (All Browsers): (Stefano)
<img width="951" alt="Screen Shot 2020-01-15 at 9 51 26 AM" src="https://user-images.githubusercontent.com/39598672/72443676-ab70f300-377c-11ea-8b9e-dcb6643cdad0.png">
<img width="978" alt="Screen Shot 2020-01-15 at 9 51 04 AM" src="https://user-images.githubusercontent.com/39598672/72443686-add34d00-377c-11ea-9466-3980638685b0.png">

# Input Edge and IE11 - Couldn't reproduce anymore / button truncate fixed on styles:
<img width="984" alt="Screen Shot 2020-01-15 at 9 53 38 AM" src="https://user-images.githubusercontent.com/39598672/72443887-086ca900-377d-11ea-8608-fda3e89cd3d0.png">

<img width="1460" alt="Screen Shot 2020-01-15 at 9 57 52 AM" src="https://user-images.githubusercontent.com/39598672/72444175-8e88ef80-377d-11ea-90b7-d727821d5d6a.png">

# MultiInput: (Mike) - done
* remove the empty popover body:
[...]

PR #1827, opened 2020-01-16
feat: form messages adoption from styles 4.0

stefanoScalzo:
#1811 
Adding the new form-input-message-group component which uses the fd-form-input-message class. 
Added input for the fd-popover for the popover container since the div might need extra classes added for whatever reasons. In this case I needed it for the width of the inputs.
**Note**: *there are styling errors in text area which will be fixed in a pr in fundamental styles such as the z-index for text areas and the margins for both inputs and textareas* 

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/form/form-input-message-group/form-input-message-group.component.ts
+++ b/libs/core/src/lib/form/form-input-message-group/form-input-message-group.component.ts
@@ -1,17 +1,51 @@
-import { Component, Input, ViewEncapsulation } from '@angular/core';
+import { ChangeDetectionStrategy, Component, EventEmitter, Input, Output, ViewEncapsulation } from '@angular/core';
+import { PopoverFillMode } from '@fundamental-ngx/core';
 
 @Component({
-  selector: 'fd-form-input-message-group',
-  templateUrl: './form-input-message-group.component.html',
-  styleUrls: ['./form-input-message-group.component.scss'],
-  encapsulation: ViewEncapsulation.None
-
+    selector: 'fd-form-input-message-group',
+    templateUrl: './form-input-message-group.component.html',
+    styleUrls: ['./form-input-message-group.component.scss'],
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class FormInputMessageGroupComponent {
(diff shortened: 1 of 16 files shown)


libs/core/src/lib/form/form-input-message-group/form-input-message-group.component.ts
```diff
@@ -0,0 +1,25 @@
...
+  type: MessageStates;
+
+  /** Whether form message should be opened */
+  @Input()
+  open: boolean = false;
+
+  @Output()
+  readonly openChange: EventEmitter<boolean> = new EventEmitter<boolean>();
+
+  constructor() { }
```
JKMarkowski: Remove constructor is unused

libs/core/src/lib/form/form-input-message-group/form-input-message-group.component.ts
```diff
@@ -0,0 +1,25 @@
...
+@Component({
+  selector: 'fd-form-input-message-group',
+  templateUrl: './form-input-message-group.component.html',
+  styleUrls: ['./form-input-message-group.component.scss'],
+  encapsulation: ViewEncapsulation.None
+
+})
+export class FormInputMessageGroupComponent {
+
+  @Input() state: string;
```
JKMarkowski: Add some comments to variables on this file.

libs/core/src/lib/form/form-input-message-group/form-input-message-group.component.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  /** Whether form message should be opened */
+  @Input()
+  open: boolean = false;
+
+  @Output()
+  readonly openChange: EventEmitter<boolean> = new EventEmitter<boolean>();
+
+  constructor() { }
+}
```
JKMarkowski: In general I think we should add some inputs, that are `popover` related,  `trigger`, `fillControlMode`, `closeOnescape`, `closeOnOutsideClick`, `id`.

mikerodonnell89:
Why does focusing the warning input open Chrome autofill for me?
Screen Shot 2020-01-17 at 4 35 35 PM

PR #1828, opened 2020-01-16
fix: multi-input 0.13.0 bugs

mikerodonnell89:

#1811 

Hide empty popover container when the displayedValues list is empty, open popover container when user types anything into the search, prevent click event from firing on disabled multi input


JKMarkowski:
Hi @mikerodonnell89, I was more thinking not about adding class, but maybe about changing `isOpen` state of multi input, what do you think about it?

mikerodonnell89:
@JKMarkowski yep that makes a lot more sense, updated the pr

PR #1832, opened 2020-01-17
fix: change checkbox state update trigger

salarenko:
#### Link to the associated issue.
#1811 

#### Brief summary of this pull request.
This PR fixes weird behavior of tristate checkbox in IE11, caused by IE11 not emitting change event on indeterminate state change. To fix this error I've changed the source triggering change of checkbox state from input `(change)` to label `(click)` and input `(keyup)` events.

### DIFF
--- a/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
+++ b/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
@@ -154,7 +154,7 @@ export class CheckboxComponent implements ControlValueAccessor {
     /** @hidden Updates checkbox state on spacebar key
      * and prevents from double check update from label-input binding */
     public checkByKey(event: KeyboardEvent): void {
-        if (event.keyCode === 32) {
+        if (this._isSpaceBarEvent(event)) {
             this.nextValue();
             event.preventDefault();
         }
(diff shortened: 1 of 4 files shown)


libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
```diff
@@ -144,6 +144,29 @@ export class CheckboxComponent implements ControlValueAccessor {
...
+     * and prevents from double check update from label-input binding */
+    public checkByClick(event: Event) {
+        this.nextValue();
+        event.preventDefault();
+    }
+
+    /** @hidden Updates checkbox state on spacebar key
+     * and prevents from double check update from label-input binding */
+    public checkByKey(event: KeyboardEvent): void {
+        if (event.keyCode === 32) {
```
rengare: lets move key `32` to constant

rengare: also move that if statement to function because it is called 2 times here and line 165 `is'x'Entered()`

salarenko: Moving conditional statement to separate function is a good idea, but I don't think there is a need to create variable for space bar key code.

PR #1857, opened 2020-01-22
fix: added class to product switch which fixes dots in li tag

rengare:
PR overwrites LI tag dots for product switch


stefanoScalzo:
There is actually no styling that needs to be added. Just make sure the markup is the same. Right now the fd-product-switch__item is at the li's child level but it should be at the li level!

PR #1872, opened 2020-01-28
fix: busy-indicator added aria-hidden attribute and a storybook for testing

droshev:
## Please provide a brief summary of this pull request.
set `aria-hidden` attribute and create a storybook for `busy-indicator`

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


libs/core/src/lib/busy-indicator/busy-indicator.component.html
```diff
@@ -1,13 +1,12 @@
...
+<div [ngClass]="{
         'fd-busy-indicator--l': size == 'l',
         'fd-busy-indicator--m': size == 'm',
         'fd-busy-indicator--s': size == 's'
     }"
-    *ngIf="loading"
-    [attr.aria-label]="loadingLabel"
->
+     *ngIf="loading"
+     aria-hidden="true"
```
mikerodonnell89: Should this be `aria-hidden="!loading"`? If this is hidden from screenreaders etc. all the time then the aria-label will always be ignored

InnaAtanasova:
Please provide a description for the Pull Request. What problem you try to solve by setting the `aria-hidden` to true?

libs/core/src/lib/busy-indicator/busy-indicator.component.html
droshev: check whether this is fine in Firefox

Issue #1839, opened 2020-01-20
Token and Tokenizer Fundamental-styles 0.5.0 adoption


InnaAtanasova:
Token component has been refactored in version 0.5.0 of Fundamental-styles and the Tokenizer has been added as a new feature to the component. Fundamental-ngx needs to adopt these changes.

PR #1878, opened 2020-01-29
fix: token and tokenizer implementation of 0.5.0

mikerodonnell89:

#1839 

Update the token and add the tokenizer component.  Also refactor the multi-input to use the tokenizer


apps/docs/src/app/core/component-docs/token/token-docs.component.ts
```diff
@@ -1,26 +1,77 @@
...
+            component: 'TokenCompactExampleComponent',
+            typescriptFileCode: compactTokenTs
+        }
+    ];
+    tokenizer: ExampleFile[] = [
+        {
+            language: 'html',
+            code: tokenizerH,
+            fileName: 'tokenizer-example-example',
+            component: 'TokenizerExampleComponent'
```
stefanoScalzo: component should only be in the typescript part

mikerodonnell89:
Going to fix a few preexisting bugs then reopen this

libs/core/src/lib/multi-input/multi-input.component.html
```diff
@@ -1,61 +1,54 @@
...
-    </fd-token>
+            </fd-popover-control>
+            <fd-popover-body [attr.aria-hidden]="!open">
+                <fd-menu class="fd-multi-input-menu-overflow"
+                         *ngIf="displayedValues && displayedValues.length"
+                         [style.maxHeight]="maxHeight">
+                    <ul fd-menu-list>
+                        <li *ngFor="let value of displayedValues; let ind  = index">
+                            <label fd-menu-item (keydown)="handleKeyDown($event, ind)">
+                                <fd-checkbox [ngModel]="selected ? selected.indexOf(value) !== -1 : false"
```
salarenko: There is a problem with this particular implementation. If you click inside the `label` inside the `li` element in popover (outside of checkbox label but still inside of the list element) checkbox is going to be selected but only visually! Checkbox value is updated on checkbox label click, I guess it'a a problem with checkbox.

libs/core/src/lib/token/tokenizer.component.html
```diff
@@ -0,0 +1,7 @@
+<div class="fd-tokenizer" (focus)="tokenizerHasFocus = true" (blur)="tokenizerHasFocus = false; null"
```
salarenko: `(blur)="tokenizerHasFocus = false; null"` I think that `null` is a leftover from some previous implementation

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -0,0 +1,23 @@
+import { Component, Input, ViewEncapsulation } from '@angular/core';
+
+@Component({
+  selector: 'fd-tokenizer',
+  templateUrl: './tokenizer.component.html',
+  styleUrls: ['./tokenizer.component.scss'],
+  encapsulation: ViewEncapsulation.None,
```
salarenko: Change detection strategy

libs/core/src/lib/multi-input/multi-input.component.scss
```diff
@@ -17,11 +19,21 @@
...
     }
 
-    .fd-multi-input-token-spacing {
-        margin: 0 4px 4px 0;
+    .fd-multi-input-tokenizer {
+        width: 100%;
+    }
+
+    .fd-multi-input-tokenizer-input {
+        border: none !important;
```
salarenko: Do we have to use `!important` :( ?

libs/core/src/lib/multi-input/multi-input.component.html
mikerodonnell89: What browser are you seeing this with?  I fixed it (with Chrome at least), do you have the latest?

mikerodonnell89:
> * Please write some tests for `Tokenizer` and a disabled case for `Token`
> * It seems that you can focus `Token` but you cannot remove it with keyboard
> * There is no visual focus on `Tokenizer` input

There is no longer a disabled case for token

libs/core/src/lib/multi-input/multi-input.component.scss
mikerodonnell89: This css class felt code-smelly to me as well but I couldn't find another way, I was hoping to solicit help from the team on these properties

mikerodonnell89:
> * It seems that you can focus `Token` but you cannot remove it with keyboard

I can tab out of a focused token... did you have another method of doing so in mind?

libs/core/src/lib/multi-input/multi-input.component.scss
JKMarkowski: Please do not use `!important` in library, always try to use stronger selectors

InnaAtanasova:
Selected state token has black outline, it's white in Fund-styles
Screen Shot 2020-02-03 at 3 35 33 PM
Screen Shot 2020-02-03 at 3 35 27 PM

InnaAtanasova:
Screen Shot 2020-02-03 at 3 39 53 PM

I think the "functionality" part of Tokenizer is missing. When user enters some value in the input field and press Enter a new Token should be created.

InnaAtanasova:
Screen Shot 2020-02-03 at 3 49 35 PM

Compact Tokenizer doesn't have the correct dimensions for the input (Input should also be compact)

mikerodonnell89:
> Selected state token has black outline, it's white in Fund-styles
> 
> Styles:
> Screen Shot 2020-02-03 at 3 35 33 PM
> 
> NGX:
> Screen Shot 2020-02-03 at 3 35 27 PM

i am only seeing the black outline on selected tokens on the deployment, not on my local server… any idea how that could happen?

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -0,0 +1,86 @@
...
+        if (event.code === 'ArrowLeft' || event.code === 'ArrowRight') {
+            let elementToFocus;
+            if (event.code === 'ArrowLeft') {
+                newIndex = fromIndex - 1;
+            } else if (event.code === 'ArrowRight') {
+                newIndex = fromIndex + 1;
+            }
+            if (newIndex >= 0 && newIndex < this.tokenList.length) {
+                elementToFocus = this.tokenList.filter((element, index) => index === newIndex)[0]
+                    .elementRef.nativeElement.querySelector('.fd-token');
```
JKMarkowski: Could you add `focus` method on `td-token` element and use elements from `contentChildren`, instead of queryList? 
Currently there is JS solution. I would in general replace it by observables with `renderer.listen` methods. You can check how it's done with in [tabs](https://github.com/SAP/fundamental-ngx/blob/26746d3fa15628b7be3aebf012bf696d762c271e/libs/core/src/lib/tabs/tab-nav/tab-nav.component.ts)
@salarenko  @rengare  @InnaAtanasova  What do you think about it?

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    /** @hidden */
+    inputKeyDown(event) {
+        if (event.key === 'ArrowLeft') {
+            this.handleTokenFocus(event, this.tokenList.length);
+        }
+    }
+
+    /** @hidden */
+    handleTokenFocus(event, fromIndex) {
```
JKMarkowski: Also It would be great to somehow split this method.

libs/core/src/lib/token/tokenizer.component.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    fixture = TestBed.createComponent(TokenizerComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+    expect(component.tokenizerHasFocus).toBeFalsy();
+    expect(component.compact).toBeFalsy();
+  });
```
JKMarkowski: I would suggest to add some tests for keyboard support

libs/core/src/lib/token/token.component.ts
```diff
@@ -18,42 +16,44 @@ import {
...
     styleUrls: ['./token.component.scss'],
-    host: {
-        class: 'fd-token',
-        '[class.fd-token__disabled]': 'disabled',
-        'role': 'button'
-    },
     encapsulation: ViewEncapsulation.None,
     changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class TokenComponent {
```
JKMarkowski: Let's add there some onKeyDown  EventEmitter<KeyboardEvent>

libs/core/src/lib/token/tokenizer.component.ts
JKMarkowski: Please add type return and types for parameters

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -0,0 +1,86 @@
...
+    @Input()
+    inputValue: string;
+
+    /** @hidden */
+    ngAfterViewInit() {
+        this.input.elementRef.nativeElement.addEventListener('keydown', (event) => this.inputKeyDown(event));
+    }
+
+    /** @hidden */
+    inputKeyDown(event) {
```
JKMarkowski: Please add type return (void) and types for parameters

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -0,0 +1,86 @@
...
+    /** Whether the tokenizer-example is compact */
+    @Input()
+    compact: boolean = false;
+
+    /** The value for the tokenizer input */
+    @Input()
+    inputValue: string;
+
+    /** @hidden */
+    ngAfterViewInit() {
```
JKMarkowski: Please add type return (void)

libs/core/src/lib/token/token.component.ts
mikerodonnell89: Would they not just use `<fd-token (keydown)="someFunction()">`?

libs/core/src/lib/form/form-control/form-control.directive.ts
```diff
@@ -34,6 +34,9 @@ export class FormControlDirective extends AbstractFdNgxClass {
     @Input()
     type: string;
 
+    /** @hidden */
+    elementRef: ElementRef;
+
```
JKMarkowski: If you need to use it somewhere else, maybe put it to `constructor(public elementRef: ElementRef) {}`

libs/core/src/lib/multi-input/multi-input.component.scss
```diff
@@ -1,4 +1,6 @@
 @import "~fundamental-styles/dist/multi-input";
+@import "~fundamental-styles/dist/tokenizer";
+@import "~fundamental-styles/dist/token";
```
JKMarkowski: Do we need it? token/tokenizer module is imported to multi-input module. These imports increases size of output file.

libs/core/src/lib/token/token.component.ts
JKMarkowski: actually you are right, it's propagated upper.

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -0,0 +1,90 @@
...
+@Component({
+  selector: 'fd-tokenizer',
+  templateUrl: './tokenizer.component.html',
+  styleUrls: ['./tokenizer.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class TokenizerComponent implements AfterViewInit {
+
+    @ContentChildren(forwardRef(() => TokenComponent))
```
JKMarkowski: Can we add `/** @hidden  */` comment here?

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -0,0 +1,90 @@
...
+    tokenList: QueryList<TokenComponent>;
+
+    @ContentChild(forwardRef(() => FormControlDirective), {static: true})
```
JKMarkowski: Can we add `/** @hidden  */` comment here?

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -0,0 +1,90 @@
...
+    ngAfterViewInit(): void {
+        if (this.input && this.input.elementRef) {
+            this.input.elementRef.nativeElement.addEventListener('keydown', (event) => {
+                this.handleKeyDown(event, this.tokenList.length);
+            });
+        }
+    }
+
+    /** @hidden */
+    handleKeyDown(event, fromIndex): void {
```
JKMarkowski: Let's add types to parameters

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -0,0 +1,90 @@
...
+        }
+        if (newIndex === this.tokenList.length && event.code === 'ArrowRight') {
+            this.input.elementRef.nativeElement.focus();
+        } else if (newIndex || newIndex === 0) {
+            this.focusTokenElement(event, newIndex);
+        }
+    }
+
+    /** @hidden */
+    focusTokenElement(event, newIndex): HTMLElement {
```
JKMarkowski: Let's add types to parameters

libs/core/src/lib/token/token.component.ts
```diff
@@ -18,42 +16,44 @@ import {
...
-            }
+        event.stopPropagation();
+        if (!this.disabled) {
+            this.onCloseClick.emit(event);
         }
     }
 
-}
-
+    constructor(elementRef: ElementRef) {
```
JKMarkowski: The same situation as above, maybe consider adding `constructor(public elementRef: ElementRef) `

JKMarkowski:
> > Selected state token has black outline, it's white in Fund-styles
> > Styles:
> > Screen Shot 2020-02-03 at 3 35 33 PM
> > NGX:
> > Screen Shot 2020-02-03 at 3 35 27 PM
> 
> i am only seeing the black outline on selected tokens on the deployment, not on my local server… any idea how that could happen?

Angular tries to optimize `.scss` code during compilation, it causes such issues.

apps/docs/src/app/core/component-docs/token/examples/tokenizer-compact-example/tokenizer-compact-example.component.html
```diff
@@ -0,0 +1,10 @@
+<form [formGroup]="tokenizerExampleForm" (ngSubmit)="onSubmit()">
+    <div fd-form-item>
+        <fd-tokenizer [compact]="true">
+            <fd-token (onCloseClick)="removeToken(token)" *ngFor="let token of tokens" [compact]="true"
+                      [readOnly]="token.readOnly">{{token.text}}</fd-token>
+            <input formControlName="inputControl" fd-form-control class="fd-input fd-tokenizer__input" [compact]="true"
```
JKMarkowski: Can you create directive `fd-tokenizer-input`, which will allow `fd-tokenizer__input` class? We don't force users to put classes in element. Also please remove `fd-input`, because it's provided by `fd-form-control`

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -0,0 +1,93 @@
...
+        if (newIndex >= 0 && newIndex < this.tokenList.length) {
+            elementToFocus = this.tokenList.filter((element, index) => index === newIndex)[0]
+                .elementRef.nativeElement.querySelector('.fd-token');
+            // element needs tabindex in order to be focused
+            elementToFocus.setAttribute('tabindex', '0');
+            elementToFocus.focus();
+            elementToFocus.addEventListener('keydown', handleFunctionReference);
+            elementToFocus.addEventListener('blur', () => {
+                elementToFocus.setAttribute('tabindex', '-1');
+                elementToFocus.removeEventListener('keydown', handleFunctionReference);
```
JKMarkowski: Could you put it to different method, named for example: `addKeyboardListener(element: type): void`

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    /** @hidden */
+    focusTokenElement(event: KeyboardEvent, newIndex: number): HTMLElement {
+        // function needs to be defined in order to be referenced later by addEventListener/removeEventListener
+        const handleFunctionReference = (e) => {
+            if (newIndex || newIndex === 0) {
+                this.handleKeyDown(e, newIndex);
+            }
+        };
+        let elementToFocus;
```
JKMarkowski: Please add type to this variable

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -0,0 +1,93 @@
...
+        if (this.input && this.input.elementRef) {
+            this.input.elementRef.nativeElement.addEventListener('keydown', (event) => {
+                this.handleKeyDown(event, this.tokenList.length);
+            });
+        }
+    }
+
+    /** @hidden */
+    handleKeyDown(event: KeyboardEvent, fromIndex: number): void {
+        let newIndex;
```
JKMarkowski: Could you add type to this variable?

InnaAtanasova:
We need to discuss the missing border but it's in styles. The PR looks good to me!

PR #1902, opened 2020-02-03
chore: storybook plugins versions updated

rengare:
previous : 5.3.8
after: 5.3.10


droshev:
@rengare this is handled by dependabot so you don't have to handle this.

PR #1938, opened 2020-02-10
chore: fix netlify build

droshev:


JKMarkowski:
@droshev  The netlify's preview is still crashing

droshev:
@JKMarkowski it is fine now

Issue #2014, opened 2020-02-19
product switch during DnD has unexpected rectangle


droshev:
#### Is this a bug, enhancement, or feature request?
bug 

#### Briefly describe your proposal.
open the product switch and start drag and dropping
![Screen Recording 2020-02-19 at 3 52 07 PM 2020-02-19 15_53_09](https://user-images.githubusercontent.com/4380815/74875323-0a9fc580-5330-11ea-8008-fa1b681cc377.gif)

PR #2019, opened 2020-02-19
fix: product switch drag and drop blue box issue

mikerodonnell89:

#2014

### DIFF
--- a/libs/core/src/lib/product-switch/product-switch-body/product-switch-body.component.scss
+++ b/libs/core/src/lib/product-switch/product-switch-body/product-switch-body.component.scss
@@ -40,6 +40,13 @@ $product-switch-dnd-line-color: #0854A0;
     &.fd-dnd-on-drag {
         background-color: white;
         z-index: 15;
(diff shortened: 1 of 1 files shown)


libs/core/src/lib/product-switch/product-switch-body/product-switch-body.component.scss
```diff
@@ -40,6 +40,13 @@ $product-switch-dnd-line-color: #0854A0;
     &.fd-dnd-on-drag {
         background-color: white;
         z-index: 15;
+        .fd-product-switch__item {
+            .fd-product-switch__title {
+                &:before, &:after {
+                    background-color: white;
```
droshev: there should be a theming parameter for that

InnaAtanasova: The whole file has hard-coded values

droshev: I see - we should address it in a separate issue

PR #2023, opened 2020-02-20
chore: migrate the lib to Angular 9

rengare:
fixes: https://github.com/SAP/fundamental-ngx/issues/2125

### Packages
There  documentation page / library / schematics updated to 9nth version of angular. 
For IE11 support I had to add some libraries
* Classlist - it was on cdn before, now we have it as a package on package.json / node_modules. It's included in polyfill.ts file.
* web-animations-js - Library added to support IE11/Safari. It's added to `polyfills.ts` 
* core-js is added to package.json (It was before as peer-dependency). Core-js polyfills are also included into `polyfills.ts` file

* Scheamtics and some testing tools are updated. 
* Typescript version is changed from 3.5 -> 3.7 due to compatibility.
* Due to some errors/warnings, `raw-loader` also had to be included into packages.

### IVY - Outdated - IVY stays disabled for now.

> * IVY is now enabled by default, so we had to disable it for libs (core/platform) it's not ready to be used with them. Officials say that it will be ready at ng10/11.
> * IVY now works with IE11, so it's enabled on docs page

### Netlify Build

> * After enabling IVY, splitted module was quite big. Size of it increased by ~50%, it was because on netlify we had unoptimized build. Now build is optimized, which costs us ~1.5/2 more minutes per build. 

*  Netlify build's time stays the same.

[Test results](https://docs.google.com/document/d/1hr46-5D1PaUbABDPvxuCUHKiXy2Sp9P8S06SfFu2fy8/edit?usp=sharing)


droshev:
#2125

tsconfig.json
```diff
@@ -15,7 +15,7 @@
         "typeRoots": ["node_modules/@types"],
```
fkolar: Is this on purpose that we are getting to 9 and not using IVY compiler ?

libs/platform/tsconfig.lib.prod.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./tsconfig.lib.json",
+  "angularCompilerOptions": {
+    "enableIvy": false
```
fkolar: Just like my other comment, when we go 9, we want to also with Ivy, unless we have some specific reason .

JKMarkowski: `IVY` is not ready to be used on library level I think

fkolar: You are right. To publish Ivy compiled code does not make sense at this time.

PR #2044, opened 2020-02-21
fix: placeholder's position fixed in rtl

### LINKED ISSUE #2041: Select placeholder in RTL is wrongly postioned
bug
Screen Shot 2020-02-20 at 6 35 12 PM
Screen Shot 2020-02-20 at 6 34 54 PM

rengare:
Fixes #2041

Rtl service added in select component to handle rtl

### Before:
image

### After:
image

### DIFF
--- a/libs/core/src/lib/select/select.component.ts
+++ b/libs/core/src/lib/select/select.component.ts
@@ -20,10 +20,11 @@ import {
 } from '@angular/core';
 import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
 import { OptionComponent } from './option/option.component';
-import { defer, merge, Observable, Subject } from 'rxjs';
+import { defer, merge, Observable, Subject, BehaviorSubject } from 'rxjs';
 import { startWith, switchMap, takeUntil } from 'rxjs/operators';
-import { PopperOptions } from 'popper.js';
+import { PopperOptions, Behavior } from 'popper.js';
 import { PopoverFillMode } from '../popover/popover-directive/popover.directive';
+import { RtlService } from 'libs/core/src/lib/utils/public_api';
 
 type SelectType = 'noborder' | 'splitborder';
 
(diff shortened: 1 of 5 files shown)


libs/platform/src/lib/components/select/select.component.ts
```diff
@@ -7,9 +7,10 @@ import {
     ChangeDetectorRef,
     ContentChild,
     TemplateRef,
-    AfterContentInit } from '@angular/core';
+    AfterContentInit
+} from '@angular/core';
 import { NG_VALUE_ACCESSOR } from '@angular/forms';
-import { OptionComponent, SelectComponent as fdSelect } from '@fundamental-ngx/core';
+import { OptionComponent, SelectComponent as fdSelect, RtlService } from '@fundamental-ngx/core';
```
mikerodonnell89: Importing these from the lib causes a circular dependency

libs/core/src/lib/select/select.component.ts
```diff
@@ -163,7 +168,7 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnInit, OnD
     /** @hidden */
     onTouched: Function = () => { };
 
-    constructor(private changeDetectorRef: ChangeDetectorRef) { }
+    constructor(private changeDetectorRef: ChangeDetectorRef, @Optional() private rtlService: RtlService) { }
```
salarenko: `_` prefix

libs/core/src/lib/select/select.component.ts
```diff
@@ -185,7 +190,12 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnInit, OnD
     }
 
     ngOnInit() {
-        // console.log(this.loading);
+        if (this.rtlService) {
+            this.rtlService.rtl
```
salarenko: Unsubscribe missing and can be simplified to observable

libs/core/src/lib/select/select.component.ts
rengare: prefix added

libs/core/src/lib/select/select.component.ts
rengare: observable used instead

Issue #2037, opened 2020-02-20
Message states wrongly position in RTL mode


droshev:
#### Is this a bug, enhancement, or feature request?
bug
#### Briefly describe your proposal.

![Screen Shot 2020-02-20 at 6 09 16 PM](https://user-images.githubusercontent.com/4380815/74988610-3b552d00-540c-11ea-8dc0-41c45620aef4.png)
![Screen Shot 2020-02-20 at 6 07 55 PM](https://user-images.githubusercontent.com/4380815/74988621-41e3a480-540c-11ea-89c0-94e154578c3c.png)

PR #2047, opened 2020-02-21
fix: Add RTL support for input messages

salarenko:
Closes: #2037 

This PR adds RTL support for input message displayed inside popover.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/form/form-input-message-group/form-input-message-group.component.ts
+++ b/libs/core/src/lib/form/form-input-message-group/form-input-message-group.component.ts
@@ -1,5 +1,16 @@
-import { ChangeDetectionStrategy, Component, EventEmitter, Input, Output, ViewEncapsulation } from '@angular/core';
-import { PopoverFillMode } from '../../popover/popover-directive/popover.directive';
+import {
+    ChangeDetectionStrategy,
+    Component,
+    EventEmitter,
+    Input,
+    Optional,
+    Output,
+    ViewEncapsulation
+} from '@angular/core';
+import { PopoverFillMode, RtlService } from '../../..';
+import { Observable, of } from 'rxjs';
+import { Placement } from 'popper.js';
+import { map } from 'rxjs/operators';
 
 @Component({
     selector: 'fd-form-input-message-group',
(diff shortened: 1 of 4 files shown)


libs/core/src/lib/form/form-input-message-group/form-input-message-group.component.ts
```diff
@@ -41,11 +53,24 @@ export class FormInputMessageGroupComponent {
...
      * Function is called every time message changes isOpen attribute
      */
     public openChanged(isOpen: boolean) {
         this.isOpenChange.emit(isOpen);
     }
 
+    /** @hidden */
+    private _createRtlObservable(): void {
+        this.placement$ = this._rtlService
+            ? this._rtlService.rtl.pipe(map(isRtl => isRtl ? 'bottom-end' : 'bottom-start'))
```
mikerodonnell89: I think this way is more concise: https://github.com/SAP/fundamental-ngx/blob/master/libs/core/src/lib/breadcrumb/breadcrumb.component.ts#L149

libs/core/src/lib/nested-list/nested-list-popover/nested-list-popover.component.ts
```diff
@@ -76,4 +63,21 @@ export class NestedListPopoverComponent {
...
+            }
+        });
+    }
+
+    /** @hidden */
+    private _createRtlObservable(): void {
+        this.placement$ = this._rtlService
+            ? this._rtlService.rtl.pipe(map(isRtl => isRtl ? 'left-start' : 'right-start'))
+            : of('right-start')
+    }
```
InnaAtanasova: Mike's comment for the file above?

libs/core/src/lib/form/form-input-message-group/form-input-message-group.component.ts
salarenko: I don't agree.
- This example needs to take care of two subscriptions instead of one. (by the way there is memory leak, because RTL service is never unsubscribed).
- This is not a valid usage of Subject [From RXJS creators](https://medium.com/@benlesh/on-the-subject-of-subjects-in-rxjs-2b08b7198b93)

If you want to I can split this example to If/else to be more readable.

rengare: lets clear things :)
- Subjects are both observer and observable source
- rxjs didn't follow the exact specification of observer pattern
- if subscribe is done in ts file, the unsubscribe method must be called
- if the async pipe is used and there is no subscribe in ts file then async pipe takes care and unsubscribe from the subscription
@mikerodonnell89 I prefer @salarenko approach because there is no need for the state to be stored in the component in fact only source of truth is rtl service. I have changed my implementation.

PR #2099, opened 2020-03-02
fix: status indicator refactoring to Info Label

Lokanathan-k:
feat:status indicator refactoring to infoLabel
fixes: #2128 

https://github.com/SAP/fundamental-styles/issues/574

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


libs/core/src/lib/info-label/info-label.component.html
```diff
@@ -0,0 +1 @@
+<span class="fd-info-label" [ngClass]=" getCustomIconClass">{{message}}</span>
```
InnaAtanasova: Please use the approach the team is using for applying the classes. You can take a look at components like Button, Bar, Link, Form-group, Tab, etc. You need to use the `buildComponentCssClass` method instead of calling `getCustomIconClass`. You can use `<ng-content></ng-content>` instead of `{{message}}`

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,55 @@
...
+    glyph: String; 
+
+    @Input()
+    message: String;
+
+    @Input()
+    color: String;
+
+    cssname: string = '';
+
```
InnaAtanasova: These Input props need TypeDocs comments ` /** ....... */`

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,55 @@
...
+            this.cssname = this.cssname + 'fd-info-label--only-icon';
+        }
+        if (this.glyph) {
+            this.cssname = this.cssname + ' sap-icon--' + this.glyph;
+        }
+        if (this.color) {
+            this.cssname = this.cssname + ' fd-info-label--accent-color-' + this.color;
+        }
+        return this.cssname;
+    }
```
InnaAtanasova: Please use the approach the team is using for applying the classes. You can take a look at components like Button, Bar, Link, Form-group, Tab, etc. You need to use the buildComponentCssClass method instead of calling getCustomIconClass.

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    @Input()
+    glyph: String; 
+
+    @Input()
+    message: String;
+
+    @Input()
+    color: String;
+
```
InnaAtanasova: Should be private with getters and setters.

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,54 @@
...
+})
+export class InfoLabelComponent implements OnInit {
+
+    /** Select type defines the label types */
+    @Input()
+    labelType: LabelType;
+
+    /** define the icon type */
+    @Input()
+    glyph: String;
```
JKMarkowski: Let's change `String` to lowercase `string`. changes are easier to detect on primitive types

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,54 @@
...
+    /**define the colour of the info label */
+    @Input()
+    color: String;
+
+    cssname: string = '';
+
+    /** @hidden */
+    constructor() {
+    }
+    ngOnInit() {
```
JKMarkowski: Add types returned to all methods, even if it's `void`

libs/core/src/lib/info-label/public_api.ts
```diff
@@ -0,0 +1 @@
+export * from './info-label.module';
```
JKMarkowski: Add InfoLabelComponent there as well

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    /**define the colour of the info label */
+    @Input()
+    color: string;
+
+    cssname: string = '';
+    clickable: boolean;
+    inverted: boolean;
+    large: boolean;
+    class: string;
```
InnaAtanasova: Missing TypeDocs for these properties

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,69 @@
...
+@Component({
+    selector: 'fd-info-label',
+    templateUrl: './info-label.component.html',
+    styleUrls: ['./info-label.component.scss'],
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class InfoLabelComponent implements OnInit, OnChanges, CssClassBuilder {
+
+    /** Select type defines the label types */
```
InnaAtanasova: The TypeDocs should mention what are these types. Whatever you put here appears in the API documentation of the library, so it's important for the user to know what are the available types.

libs/core/src/lib/info-label/info-label.component.html
```diff
@@ -0,0 +1 @@
+<span class="fd-info-label" [ngClass]="buildComponentCssClass"><ng-content></ng-content></span>
```
InnaAtanasova: I wonder why do you go with this approach instead of using Directive? This way you don't give the user the option to put the directive on any HTML element, like `<a>, <span>, <div>`, etc.

libs/core/src/lib/modal/modal.component.scss
```diff
@@ -1,4 +1,4 @@
-@import "~fundamental-styles/dist/modal";
+//@import "~fundamental-styles/dist/modal";
```
mikerodonnell89: Don't want to commit this

libs/core/src/lib/button-group/button-group.component.scss
```diff
@@ -1,2 +1,2 @@
-@import '~fundamental-styles/dist/button-group';
+//@import '~fundamental-styles/dist/button-group';
```
mikerodonnell89: Shouldn't commit this

Lokanathan-k: @mikerodonnell89 i m getting the below error on building code

12:57:27 AM: ERROR: Can't find stylesheet to import.
12:57:27 AM:   ╷
12:57:27 AM: 1 │ @import '~fundamental-styles/dist/button-group';
12:57:27 AM:   │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
12:57:27 AM:   ╵
12:57:27 AM:   libs/core/src/lib/button-group/button-group.component.scss 1:9  root stylesheet
12:57:27 AM: An unhandled exception occurred: Can't find stylesheet to import.
12:57:27 AM:   ╷
12:57:27 AM: 1 │ @import '~fundamental-styles/dist/button-group';

libs/core/src/lib/info-label/info-label.component.html
InnaAtanasova: You need to remove this file since you don't use it.

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,99 @@
...
+    changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class InfoLabelComponent implements OnInit, OnChanges, CssClassBuilder {
+  
+    /** 
+     * The LabelType represented by the info label .
+     * Can be one of the following: 'numeric' | 'only-icon' | 'icon'
+     * For default info label omit this property
+     */
+    _labelType: LabelType;
```
InnaAtanasova: Should be private

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,99 @@
...
+        this._labelType = value;
+        this.buildComponentCssClass();
+    }
+
+    get labelType(): LabelType {
+        return this._labelType;
+    }
+
+    /** define the icon type */
+    _glyph: string;
```
InnaAtanasova: Should be private

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,99 @@
...
+        this._glyph = value;
+        this.buildComponentCssClass();
+    }
+
+    get glyph(): string {
+        return this._glyph;
+    }
+
+    /**define the colour of the info label */
+    _color: string;
```
InnaAtanasova: Should be private

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,99 @@
...
+        this._color = value;
+        this.buildComponentCssClass();
+    }
+
+    get color(): string {
+        return this._color;
+    }
+
+
+    _class: string = '';
```
InnaAtanasova: Should be private and should have TypeDocs

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,99 @@
+import { Component, ViewEncapsulation, ChangeDetectionStrategy, Input, OnInit, ElementRef, OnChanges } from '@angular/core';
+import { CssClassBuilder, applyCssClass } from '../utils/public_api';
+
+type LabelType = 'numeric' | 'only-icon' | 'icon';
+
+@Component({
+    // tslint:disable-next-line:component-selector
+    selector: '[fd-info-label]',
+    template: `<ng-content></ng-content>`,
```
fkolar: Why would a component have directive (attribute) selector ?

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    get labelType(): LabelType {
+        return this._labelType;
+    }
+
+    /** define the icon type */
+    _glyph: string; 
+
+    @Input()
+    set glyph(value: string) {
```
fkolar: We shoudl be more clear when to use setters and getters and in this case it does not make much sense here. Usually you use setters (or ngOnChange) for specific properly when it can over component lifecycle. After it is initialized. I dont think any of these properties can change after its created. 

Instead of having 1 call at the end once view is initialized, you make here 5 different calls.

InnaAtanasova: https://fundamental-ngx.netlify.com/#/core/button

Our Playground is an example where you can see `glyph` property being changed. Another case would be the RTL support where you might want to change the icon ( `arrow-left` to `arrow-right`)

apps/docs/src/app/core/component-docs/info-label/info-label-docs.component.html
```diff
@@ -0,0 +1,42 @@
...
+</fd-docs-section-title>
+<description> <code>fd-info-label fd-info-label--accent-color-1 fd-info-label--icon</code> modifier can be used to render a icon with text version of the Info Label. </description>
+<component-example>
+	<fd-info-label-text-icon-example></fd-info-label-text-icon-example>
+</component-example>
+<code-example [exampleFiles]="InfoLabelTextIconExample"></code-example>
+
+<separator></separator>
+
+<fd-docs-section-title [id]="'filled'" [componentName]="'infolabel'">
```
salarenko: Used same `[id]` attribute value for multiple examples.

apps/docs/src/app/core/component-docs/info-label/info-label-docs.component.ts
```diff
@@ -0,0 +1,49 @@
...
+    selector: 'app-info-label',
+    templateUrl: './info-label-docs.component.html'
+})
+export class InfoLabelDocsComponent {
+
+    defaultInfoLabelHtmlType: ExampleFile[] = [
+        {
+            language: 'html',
+            code: InfoLableDefaultExample,
+            fileName: 'Info-label-default-example',
```
salarenko: I think it should be:
`fileName: 'Info-label-default-example',` => `fileName: 'info-label-default-example',`

apps/docs/src/app/core/component-docs/info-label/info-label-docs.module.ts
```diff
@@ -0,0 +1,44 @@
...
+import {InfoLabelHeaderComponent} from './info-label-header/info-label-header.component';
+import {InfoLabelModule} from '@fundamental-ngx/core';
+
+const routes: Routes = [
+    {
+        path: '',
+        component: InfoLabelHeaderComponent,
+        children: [
+            {path: '', component: InfoLabelDocsComponent},
+            {path: 'api', component: ApiComponent, data: {content: API_FILES.infolabel}}
```
salarenko: Should be: `API_FILES.infolabel` => `API_FILES.infoLabel`

apps/docs/src/app/core/component-docs/info-label/info-label-header/info-label-header.component.html
```diff
@@ -0,0 +1,13 @@
+<header>Info Label</header>
+<description>
+    <p>Info labels are used to indicate status. Colors, generally in combination with text, numeric and icon, are used to easily
```
salarenko: I think this description needs to be improved.

> Colors, generally in combination with text, numeric and icon, are used to easily highlight.

Highlight what?

> Black is used to indicate default or inactive status  

What "Black"?

> Diffrent color options are availabel to define the status of the object

Typo

> Info Label is a small non-interactive numeric or text based control

I feels like you should start from that

> Diffrent color options are availabel to define the status of the object

What object?

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,99 @@
...
+    /** @hidden */
+    constructor(private _elementRef: ElementRef) {
+    }
+
+    ngOnInit(): void  {
+        this.buildComponentCssClass();
+    }
+
+    ngOnChanges(): void {
+        this.buildComponentCssClass();
```
salarenko: I think there is no point of using `ngOnChanges` only to trigger `this.buildComponentCssClass();` if every `@Input` does it by itself.

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,99 @@
...
+            this._class
+        ].filter(x => x !== '').join(' ');
+    }
+
+    /** @hidden */
+    constructor(private _elementRef: ElementRef) {
+    }
+
+    ngOnInit(): void  {
+        this.buildComponentCssClass();
```
salarenko: I think there is no point of using `ngOnInit` only to trigger `this.buildComponentCssClass();` if every `@Input` does it by itself.

InnaAtanasova: Actually here he did the right thing, he was following our approach. If you take a look at components like Link, Tabs, List, etc they all have the same code as this one

JKMarkowski: I think we should keep it in `ngOnInit`, `ngOnChanges`/`@Input` sometimes are triggered before elementRef is available (before ngOnInit)

libs/core/src/lib/info-label/info-label.component.ts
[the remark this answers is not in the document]
Lokanathan-k: we have content to add in this so we cannot remove this

apps/docs/src/app/core/component-docs/info-label/info-label-docs.component.ts
Lokanathan-k: all the file names are named accordingly referring previous component or already existing component 
Screenshot 2020-03-26 at 10 21 49 AM

apps/docs/src/app/core/component-docs/info-label/info-label-docs.component.html
```diff
@@ -8,7 +8,7 @@
 <label fd-form-label>Default Info label by passing the default message </label>
 <separator></separator>
 
-<fd-docs-section-title [id]="'filled'" [componentName]="'infolabel'">
+<fd-docs-section-title [id]="'infoLabelText'" [componentName]="'infolabel'">
 	Info label with text
 </fd-docs-section-title>
 <description> <code>fd-info-label fd-info-label--accent-color-1</code> modifier can be used to render a Text version of the Info label. </description>
```
Vanessa-Cusmich, line 14: The `fd-info-label`and `fd-info-label--accent-color-1` modifiers can be used to render a text version of the Info Label.

apps/docs/src/app/core/component-docs/info-label/info-label-docs.component.html
```diff
@@ -19,7 +19,7 @@
 
 <separator></separator>
 
-<fd-docs-section-title [id]="'filled'" [componentName]="'infolabel'">
+<fd-docs-section-title [id]="'infoLabelTextIcon'" [componentName]="'infolabel'">
 	Info Label with text and icon
 </fd-docs-section-title>
 <description> <code>fd-info-label fd-info-label--accent-color-1 fd-info-label--icon</code> modifier can be used to render a icon with text version of the Info Label. </description>
```
Vanessa-Cusmich, line 25: The `fd-info-label`, `fd-info-label--accent-color-1`and `fd-info-label--icon`modifiers can be used to render an icon with a text version of the Info Label.

apps/docs/src/app/core/component-docs/info-label/info-label-docs.component.html
```diff
@@ -30,7 +30,7 @@
 
 <separator></separator>
 
-<fd-docs-section-title [id]="'filled'" [componentName]="'infolabel'">
+<fd-docs-section-title [id]="'infoLabelNumeric'" [componentName]="'infolabel'">
 	Info label with numeric and only Icon
 </fd-docs-section-title>
 <description> <code>fd-info-label--numeric fd-info-label--only-icon</code> modifier can be used to render a numeric and icon only version of the badge. </description>
```
Vanessa-Cusmich, line 36: The `fd-info-label--numeric` and `fd-info-label--only-icon` modifiers can be used to render numeric-only and icon-only versions of the badge.

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    cssname: string = '';
+
+    /** @hidden */
+    constructor() {
+    }
+    ngOnInit() {
+    }
+
+    get buildCustomCss(): string {
```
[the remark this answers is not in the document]
Lokanathan-k: we have used CssClassBuilder and overidden  @applyCssClass
    buildComponentCssClass(): string { methord to achieve the same.

libs/core/src/lib/info-label/info-label.component.spec.ts
```diff
@@ -36,23 +36,23 @@ describe('InfoLabelComponent', () => {
 
   it('Should Add label Type', () => {
     component.ngOnInit();
-    component.labelType = 'numeric';
+    component.type = 'numeric';
     component.ngOnChanges();
```
fkolar, line 40: You dont need to call ngOnChange, if you do and something does not there is could be a problem

libs/core/src/lib/info-label/info-label.component.ts
```diff
@@ -12,79 +12,42 @@ type LabelType = 'numeric' | 'only-icon' | 'icon';
...
-        this.buildComponentCssClass();
-    }
-
-    get class(): string {
-        return this._class;
-    }
+    color: string;
 
     @applyCssClass
     buildComponentCssClass(): string {
```
fkolar, line 36: Can we try to avoid calling method from template? I know we are using it on several places, but its serious perf. problem. I would change the result out of this function call and assign only property to the template.  Angular needs to do allot of work, every time you click on that thing it needs to recalc. 

In this case I am not even sure we need to recalc. this every time. You initialize this only once and you are done. If you expect changes during lifecycle of the component, then you can recalc. this manually and save it. 

This is not issue only for this component but also for others. I am doing heavy caching and refactoring on my code as well. Thinking that OnPush solves the problem we are wrong. 

Angular introduced while ago something they called purePipes. 

Just imagine we will have 60 components like this and every components will do this call.

libs/core/src/lib/info-label/info-label.component.spec.ts
rengare: In general I agree on that but this is a unit test and ngOnChanges only rebuilds css classes, ofc there is `buildCssClass` which can be used

libs/core/src/lib/info-label/info-label.component.ts
rengare: I agree on that methods should not be used in templates. Getters behave a little bit different. I used them quite often. The biggest performance problem though is when there are too many flow controls added in template (like ngif, ngfor) even when You use properties on them the performance hit is huge (70 lines of ngif added to template  kills application) , thats why on push strategy comes in handy

Lokanathan-k: As per discussion betweek deno and frank this will be taken later, skipping this for now

droshev:
@salarenko @fkolar - could you review the PR?

libs/core/src/lib/info-label/info-label.component.ts
fkolar: If everything works with with removing settings/getters and your stackblitz examples are running fine , then i am find with this. 

I dont think loki is going to refactor all this now, as this need to be solved once for all components.

Issue #2106, opened 2020-03-03
Issues found during the Defect Hunting on March 3rd


InnaAtanasova:
### Bar - Mike - version `0.15.0`
- [x] RTL for the navigation icon 

### Input Messages (to be addressed in next version `0.16.0`
- [x] Need to adopt the latest styling (which will fix the issue)
<img width="1115" alt="Screen Shot 2020-03-03 at 11 17 14 AM" src="https://user-images.githubusercontent.com/39598672/75795674-a96cee80-5d40-11ea-8252-b554421d6fac.png">
Please  see [this](https://github.com/SAP/fundamental-ngx/issues/2071) - next release

### Input Group - Mike - version `0.15.0`
- [x] The value of the Placeholder is undefined, the input field should display the number 123
<img width="1196" alt="Screen Shot 2020-03-03 at 9 41 58 AM" src="https://user-images.githubusercontent.com/39598672/75796768-2056b700-5d42-11ea-99dc-4ae1a03423f6.png">

### RTL for Links - Mike - version `0.15.0`
- [x] the navigation icon should point to the other direction
<img width="1225" alt="Screen Shot 2020-03-03 at 9 44 15 AM" src="https://user-images.githubusercontent.com/39598672/75796934-5e53db00-5d42-11ea-8758-a7c771b36ea3.png">

### Popover - version `0.16.0`
- [x] In RTL the Popover body should appear on the other side
![2020-03-03_15-49](https://user-images.githubusercontent.com/39598672/75795753-bee21880-5d40-11ea-9987-730c4cba1250.png)

Please consider [this](https://github.com/popperjs/popper-core/issues/695)

- [ ] Issues with the focus of the Popover control - version `0.16.0`
![popover_issue_2](https://user-images.githubusercontent.com/39598672/75795950-05377780-5d41-11ea-9b71-b04fbe9ce466.gif)

- [x] IE11 - issue with Popover body overflow  - version `0.15.0` - FUNDAMENTAL-STYLES ISSUE
![image (3)](https://user-images.githubusercontent.com/39598672/75796035-1e402880-5d41-11ea-86da-3f1947a52868.png) addressed in  https://github.com/SAP/fundamental-styles/issues/755

- [x] The content of the Popover body is cutting the border radius of the Popover  - version `0.16.0` addressed in https://github.com/SAP/fundamental-styles/issues/876
**Should be fixed with this [#682](https://github.com/SAP/fundamental-styles/pull/682)**
[...]

PR #2108, opened 2020-03-03
docs: bar nav arrow direction in RTL mode, example source code

mikerodonnell89:

#2106 

Fix bar navigation arrows in RTL mode and add examples

### DIFF
--- a/apps/docs/src/app/core/component-docs/bar/examples/bar-default-example.component.ts
+++ b/apps/docs/src/app/core/component-docs/bar/examples/bar-default-example.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, OnDestroy, OnInit } from '@angular/core';
 import { BehaviorSubject } from 'rxjs';
 import { RtlService } from '@fundamental-ngx/core';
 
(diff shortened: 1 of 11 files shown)


apps/docs/src/app/core/component-docs/bar/examples/bar-default-example.component.ts
```diff
@@ -0,0 +1,23 @@
...
+})
+export class BarDefaultExampleComponent implements OnInit {
+
+    navigationArrow$: BehaviorSubject<string> = new BehaviorSubject<string>('navigation-left-arrow');
+
+    constructor(private rtlService: RtlService) { }
+
+    ngOnInit(): void {
+        if (this.rtlService) {
+            this.rtlService.rtl.subscribe(value => {
```
salarenko: You need to take care of that subscription, to avoid memory leak.
This can be simplified to one Observable usage.
Same for `BarHeaderExampleComponent`, `BarPageExampleComponent` and `BarPageResponsiveExampleComponent`

rengare: also for all examples we can skip null check for rtl service because we controle when this service is injected

mikerodonnell89:
As discussed on the call this morning, we won't be using css transforms to spin the icons around as we can't say for sure what icons will be used where, so the developer should use the RTL service if they need to.  However I have used observables and skipped the null check for the RtlService

Issue #2114, opened 2020-03-04
Getting started docs should mention required styles


mikerodonnell89:
#### Is this a bug, enhancement, or feature request?

Currently the getting started docs don't mention that fonts.css and icons.css need to be imported in angular.json

PR #2115, opened 2020-03-04
docs: get installation README.md up to date

mikerodonnell89:

#2114 

Add icon.css and and fonts.css to readme.  Also change example component to checkbox, since switch no longer has any inputs or built-in label

### DIFF
--- a/libs/core/README.md
+++ b/libs/core/README.md
@@ -91,17 +91,17 @@ For an existing Angular CLI application,
     For example, to use Switchs, add the following import to your main application module.
 
     ```javascript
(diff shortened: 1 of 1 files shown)


libs/core/README.md
```diff
@@ -83,25 +91,25 @@ For an existing Angular CLI application,
     For example, to use Switchs, add the following import to your main application module.
```
stefanoScalzo: typo

stefanoScalzo: should be to use checkbox

PR #2117, opened 2020-03-04
feat(platform): Adding initial code for new platform menu.

KevinOkamoto:
N/A

The new platform menu models itself after the Angular Material Menu. Instead of being one component "button" plus "dropdown menu", the "dropdown menu" is a separate component and can be attached/opened by any element using a "trigger" directive.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

Screen Shot 2020-03-04 at 1 34 51 PM

Screen Shot 2020-03-04 at 1 35 04 PM


kavya-b:
1. When navigating through mouse, on clicking a submenu item, the parent menu closes. But when navigating through keyboard, on clicking a submenu item, the parent menu does not close. Shouldn't this behaviour be consistent?
2. Can we add an example that shows icons, separator lines and a longer menu with scrolling inside the menu; basically something like what the earlier 'complex menu' example had?

libs/platform/src/lib/components/menu/menu-item.component.scss
```diff
@@ -1,131 +1,31 @@
...
+  &.trigger:after {
+    position: absolute;
+    right: 0;
+    width: 2rem;
+    font-family: SAP-icons;
+    font-style: normal;
+    font-weight: 400;
+    font-size: 0.75rem;
+    text-align: center;
+    content: '\e066';
```
kavya-b: Fiori team says although right arrow should be used to show a submenu, we shouldn't be restricting it to this particular icon in our code.  @droshev could you please confirm if using it in this way is okay or should we provide developer the means to place his own icon instead?

libs/platform/src/lib/components/menu/menu-item.component.ts
```diff
@@ -1,159 +1,73 @@
...
+    @HostListener('mouseenter') onMouseEnter() {
+        this.hovered.next(this);
     }
-    /**
-     * implemented method for `click` from `DefaultMenuItem`
-     */
-    public click(): void {
-        this.itemClick.emit();
+
+    ngOnInit() { }
```
fkolar: you can probably remove this one as its not used..

libs/platform/src/lib/components/menu/menu-item.component.scss
KevinOkamoto: > 1. When navigating through mouse, on clicking a submenu item, the parent menu closes. But when navigating through keyboard, on clicking a submenu item, the parent menu does not close. Shouldn't this behaviour be consistent?

All parent menus should close when "mouse clicking" or  "keyboard typing enter" on a leaf node.
I can't seem to reproduce the behavior you're describing.

> 2. Can we add an example that shows icons, separator lines and a longer menu with scrolling inside the menu; basically something like what the earlier 'complex menu' example had?

I'll try and add the additional examples. Scrolling has not been implemented - I'll try to add that feature as well.

KevinOkamoto: @kavya-b I think I'm going to wait for the fundamental-styles update (0.8.x? the one with updated Menu styles) before adding examples with icons and separator lines. I'm guessing I'll have to redo the component templates when the update occurs.

Issue #2134, opened 2020-03-05
Vertical label layout breaks when used with inline help


fkolar:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.

When You switch to vertical label layout as you can see it on this picture

<img width="1349" alt="Screen Shot 2020-03-05 at 9 13 36 PM" src="https://user-images.githubusercontent.com/17149942/76021876-581e5400-5f26-11ea-9074-ceda3031eacd.png">

The class:

```css
.fd-form-label--inline-help {
    width: 100%;
    display: flex;
    justify-content: space-between;
    align-items: center;
}
```
takes over and streatches label part  to 100% and shrings the actual control

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
latest `0.15.0-rc.17`

#### If this is a bug, please provide steps for reproducing it.
Please see above screenshot

#### Please provide relevant source code if applicable.
https://sap.github.io/fundamental-ngx/#/core/input
#### Is there anything else we should know?

Also see #2133 and other related issues

droshev:
@JKMarkowski is working on it - @JKMarkowski  can you validate this in https://github.com/SAP/fundamental-ngx/pull/4061?

JKMarkowski:
@droshev this is fixed on https://github.com/SAP/fundamental-ngx/pull/4061

Issue #2133, opened 2020-03-05
fd-form-input-message-group is limited to native input elements


fkolar:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
When working with forms you dont use only `<input>` elements, therefore current implementation is limited to only for few elements but it shoud be transparent to use any input **component**  (select, combobox, etc..)

For example:

```html
<div fd-form-item>
    <label fd-form-label for="input-54">
        Warning Input
    </label>
    <fd-form-input-message-group>
      <fd-select #select [(value)]="value" placeholder="Click here!">
        <fd-option value="apple">Apple</fd-option>
        <fd-option value="pineapple">Pineapple</fd-option>
        <fd-option value="tomato">Tomato</fd-option>
        <fd-option value="strawberry">Strawberry</fd-option>
     </fd-select>
        <fd-form-message [type]="'warning'">
            Pellentesque metus lacus commodo eget justo ut rutrum varius nunc
        </fd-form-message>
    </fd-form-input-message-group>
</div>

```

Current implementation only expect a directive:

```html
   <fd-popover-control>
        <ng-content select="[fd-form-control]"></ng-content>
    </fd-popover-control>
```

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)

latest 0.15.x

#### Please provide relevant source code if applicable.
https://sap.github.io/fundamental-ngx/#/core/input

#### Is there anything else we should know?

Ideally you want to use template in here such as:

```html

   <fd-popover-control>
      <ng-container *ngTemplateOutlet="triggerItemTemplate"></ng-container>
    </fd-popover-control>
```

and ts has just this:

```ts
  @ContentChild('triggerItem', {static: false})
  triggerItemTemplate: TemplateRef<any>;

```

and you would use it as:

```html
<ng-template #withFormMessage>
  <fd-form-input-message-group>  

    <ng-template #triggerItem>
       <fd-select #select [(value)]="value" placeholder="Click here!">
          <fd-option value="apple">Apple</fd-option>
          <fd-option value="pineapple">Pineapple</fd-option>
          <fd-option value="tomato">Tomato</fd-option>
           <fd-option value="strawberry">Strawberry</fd-option>
     </fd-select>
    </ng-template>


    <fd-form-message [type]="'error'">
       My Message....
    </fd-form-message>
  </fd-form-input-message-group>
</ng-template>
```

[...]

Issue #2131, opened 2020-03-05
Introduce FormGroup Layout to simplify form assembly


fkolar:
#### Is this a bug, enhancement, or feature request?
Feature

#### Briefly describe your proposal.
Implement (migrate) Application level component based on [following spec](https://github.com/SAP/fundamental-ngx/wiki/Platform:-FormGroup-Layout-Technical-Design) and introduce formal structure how form inputs control will be treated. 

Need to use directly Fundamental Styles as new updated form structure will not allow us to extend to higher component.

droshev:
@manjunathanagaraj @fkolar any plans bringing this? if not, would you close the issue?

droshev:
done with this https://github.com/SAP/fundamental-ngx/pull/5622

PR #2139, opened 2020-03-07
feat(platform): FormGroup layout support

fkolar:
Initial implementation for the FormGroup Layout based on the [Spec](https://github.com/SAP/fundamental-ngx/wiki/Platform:-FormGroup-Layout-Technical-Design)

Also disabling one lint rule due the problem which is still present in the tslint [64](https://github.com/mgechev/codelyzer/issues/64) issue

Once it is merged we will do next updated  based on the latest FormGroup/FormField/FormControl adjustments we have made with @KevinOkamoto.

We expect that added `package.json` dependencies for the flexgrid will be removed and replaced with fundamental-styles layout implementations once it is available. 

Current implemented as based on `core` form's directives for which I have filed also several issues:
https://github.com/SAP/fundamental-ngx/issues/2134
https://github.com/SAP/fundamental-ngx/issues/2133

Also we need to move all the inputs under this `form/` dir

@DeepakSap14  let's merge this first as I have made proper structure for the form that you can follow, so you can put your code in here. 

#2131


libs/platform/src/lib/components/form/base.input.ts
```diff
@@ -0,0 +1,267 @@
...
+    id: string = this.defaultId;
+
+    @Input()
+    get name(): string {
+        return this._name ? this._name : this.id;
+    }
+
+    set name(value: string) {
+        this._name = value;
+        this.stateChanges.next('rb set name');
```
KevinOkamoto: Is the `name` property something we need to track changes on? Will the `name` of an input field change in an application or is there an operation that we need to do on setting of the input `name`?

libs/platform/src/lib/components/form/input/input.component.html
```diff
@@ -0,0 +1,16 @@
+<!--
+  Does it make sense to have some other state than error?
+-->
+<input #elemRef fd-form-control
+       [(ngModel)]="value"
+       [ngModelOptions]="{standalone: true}"
+       [attr.id]="id"
+       [attr.type]="type"
```
KevinOkamoto: Need to add `[attr.name]` property.

libs/platform/src/lib/components/form/base.input.ts
fkolar: I have this only for debugging. But there are properties that are expected to change after is initialized and properties that never change. Next push is is going to reflect this.

fkolar:
Screen Shot 2020-03-09 at 10 15 25 PM
Screen Shot 2020-03-09 at 10 15 16 PM
Screen Shot 2020-03-09 at 10 15 08 PM



```html
<fdp-form-group #fg [multiLayout]="true"  >
    <fdp-form-field #ffl1 [id]="'serialNum'" label="Serial Number" hint="Hint for SN " zone="zLeft"
                    rank="10" required="true">
        <fdp-input [formControl]="ffl1.formControl"></fdp-input>
    </fdp-form-field>

    <fdp-form-field #ffl2 [id]="'employeeId'" label="Employee ID"  hint="Hint for one " zone="zRight"
                    rank="10">
        <fdp-input [formControl]="ffl2.formControl"></fdp-input>
    </fdp-form-field>


    <ng-template #i18n let-errros>
        <span *ngIf="errros.required">
            Value is required
        </span>
        <span *ngIf="errros.maxlength">
            You reached maxium allowed length. Required is {{errros.maxlength.requiredLength}}
            current is
            {{errros.maxlength.actualLength}}
        </span>
    </ng-template>
</fdp-form-group>

```

fkolar:
For the `hintPlacement` we will come back to this once it starts to work in the core.. See the ticket I opened

fkolar:
@JKMarkowski and @droshev adding to other approvers to pass PR check.

Issue #2053, opened 2020-02-21
Select: Standard Select drop down button


estelaxu3:
#### Bug
In standard select loading mode, the drop down button is not deactivated. 
https://sap.github.io/fundamental-ngx/#/core/select#standard
<img width="964" alt="Screen Shot 2020-02-21 at 1 55 09 PM" src="https://user-images.githubusercontent.com/54589269/75063103-2cc15100-54b2-11ea-9d0c-869f6b6d66d4.png">

JKMarkowski:
I belive, it's gonna be fixed in https://github.com/SAP/fundamental-ngx/issues/2006

PR #2150, opened 2020-03-09
feat: Update Select component to match new fundamental styles

### LINKED ISSUE #2006: Select component migration
enhancement
migrate `select` component reflecting fundamental-styles version `0.8.0` including the mobile

salarenko:
Closes #2006, #2286, #2053, #2431, #2464

This PR:
- adapts new styles and structure of Select component enforced by Fiori3 specification
- improves keyboard navigation
- changes interactive behavior of select control
- introduces `isKey` utility to unify key recognition through different browser platforms
- updates documentation
- removes unsupported by Fiori3 features like:
    - `type SelectType = 'noborder' | 'splitborder';`
    - possibility of defining whole new template using `@Input() triggerTemplate`, narrowing it only to select control content.
- removes Select.isOpen `@Input`
- removes Option.viewValue `@Input`
- removes Select.loading `@Input`

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


InnaAtanasova:
@salarenko Is it `Switch` component or `Select` ?

fkolar:
The Select implementation is becoming really complex. One thing that I see and I have commented it in slack:

I see some new code, and I am sure @droshev  already told us about the mobile full screen behavior.
```
<fd-select placeholder="Select an option"
                   [(value)]="selectedValue"
                   [mobile]="true"
                   (isOpenChange)="toggleDialog(dialogTemplate, $event)">
            <ng-template #optionsTemplate>
                <fd-select-mobile-list>
                    <fd-option *ngFor="let option of options" [value]="option">
                        {{option}}
                    </fd-option>
                </fd-select-mobile-list>
            </ng-template>
        </fd-select>
```

The construction seem to be little strange to support  full screen mobile behavior. Let’s think also about the develop experience.  Does this mean that whoever is going to use select and know that it will run on mobile needs to do this above fd-select-mobile-list  ?

You have two things:
you have 
- [mobile]=true
- then this new wrapping  component.

If you are using component you probably want to do only `[mobile]` binding but I guess by default it should be adaptive to any device.  We should do this inside the component .

There is too many combination.  I think it should be possible if you explicitly say its for mobile, to handle this internally do developer can still with `fd-option`.

libs/core/src/lib/select/select.component.ts
```diff
@@ -263,211 +308,98 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnInit, OnD
...
 
-    /**
-     * Function that tests whether the tested option is currently selected.
-     * @param option Option to test against the selected option.
-     */
-    private _isOptionActive(option: OptionComponent): boolean {
-        return option && this._selected && option === this._selected;
+    /** @hidden */
+    private _setSelectedOption(): void {
+        this.selected = this.options.find(option => option.selected)
```
fkolar, line 382: missing semicolon

libs/core/src/lib/select/select.component.ts
```diff
@@ -263,211 +308,98 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnInit, OnD
...
-            this._selected = option;
-            this._updateValue(fireEvents);
+        if (controlChange) {
+            this._updateValue(option.value);
             this.close();
-            return option;
         }
-        this.close();
-        return;
+        this._changeDetectorRef.detectChanges();
```
fkolar, line 322: You want to be careful when using detectChanges vs MarkForCheck. In this situation you are better with markForCheck. Think of situation where on the same level as this component you will for example some error logic to be displayed , so you want mark this  component and parent to be check in the next cycle so both are updated. 


Such excessive use of DetectChanges just make angular busy more than its needed. When you have a page with 50 similar controls and each of them runs detectChanges it will be pretty noisy, instead of marking component once and Angular can nicely update all marked to be checked. And you do use `OnPush`

libs/core/src/lib/select/option/option.component.ts
```diff
@@ -51,35 +50,63 @@ export class OptionComponent implements OnInit {
 
     /** Emitted when the selected state changes. */
     @Output()
-    readonly selectedChange: EventEmitter<OptionComponent>
-        = new EventEmitter<OptionComponent>();
+    readonly selectedChange: EventEmitter<OptionComponent> = new EventEmitter<OptionComponent>();
```
fkolar, line 53: Who is going to use this ?  This binding does not have much value  when not connected with the parent `select`

salarenko: Yes, sorry this is a leftover from previous implementation. I left it for eventual backwards compatibility, but since this PR brings breaking changes anyway I'm going to remove it.

salarenko:
> The Select implementation is becoming really complex. One thing that I see and I have commented it in slack: 
> ...
Hey @fkolar, thanks for your comments,

we had a discussion about our approach to components and their mobile views and we decided rather to show the user how to create mobile view out of existing components than create mobile mode under the hood of the component.

We decided on his approach mostly because we didn't wanted to bring `DialogModule` with every `ControlModule` and rather give the user an option whether he wants to support mobile mode in his application.

Another thing is `DialogComponent` is highly customizable component, and we wanted to escape hiding its API under the layer of another component.

`SelectComponent` and its mobile mode is on of the "edge cases", because some part of the control needs to be displayed outside of the Dialog. After rethinking this example I think I might be able to escape from using this additional wrapper `fd-select-mobile-list`, but I will still need to render partial views defined in `SelectControl` in `DialogModule`.

In most cases I believe our composition is going to look like this [CalendarComponent in mobile mode](https://fundamental-ngx.netlify.app/#/core/calendar#calendar-mobile).

fkolar:
> > The Select implementation is becoming really complex. One thing that I see and I have commented it in slack:
> > ...
> 
> Hey @fkolar, thanks for your comments,
> 
> we had a discussion about our approach to components and their mobile views and we decided rather to show the user how to create mobile view out of existing components than create mobile mode under the hood of the component.
> 
> We decided on his approach mostly because we didn't wanted to bring `DialogModule` with every `ControlModule` and rather give the user an option whether he wants to support mobile mode in his application.
> 
> Another thing is `DialogComponent` is highly customizable component, and we wanted to escape hiding its API under the layer of another component.
> 
> `SelectComponent` and its mobile mode is on of the "edge cases", because some part of the control needs to be displayed outside of the Dialog. After rethinking this example I think I might be able to escape from using this additional wrapper `fd-select-mobile-list`, but I will still need to render partial views defined in `SelectControl` in `DialogModule`.
> 
> In most cases I believe our composition is going to look like this [CalendarComponent in mobile mode](https://fundamental-ngx.netlify.app/#/core/calendar#calendar-mobile).

Thanks for your response. We need to figure this out how to get around this under platform then.  This component & ComboBox are for me candidates to write these components from the scratch. 

Btw. do you still need binding `mobile=true` when you can detect this from the content?

salarenko:
> > > The Select implementation is becoming really complex. One thing that I see and I have commented it in slack:
> > > ...
> > 
> > 
> > Hey @fkolar, thanks for your comments,
> > we had a discussion about our approach to components and their mobile views and we decided rather to show the user how to create mobile view out of existing components than create mobile mode under the hood of the component.
> > We decided on his approach mostly because we didn't wanted to bring `DialogModule` with every `ControlModule` and rather give the user an option whether he wants to support mobile mode in his application.
> > Another thing is `DialogComponent` is highly customizable component, and we wanted to escape hiding its API under the layer of another component.
> > `SelectComponent` and its mobile mode is on of the "edge cases", because some part of the control needs to be displayed outside of the Dialog. After rethinking this example I think I might be able to escape from using this additional wrapper `fd-select-mobile-list`, but I will still need to render partial views defined in `SelectControl` in `DialogModule`.
> > In most cases I believe our composition is going to look like this [CalendarComponent in mobile mode](https://fundamental-ngx.netlify.app/#/core/calendar#calendar-mobile).
> 
> Thanks for your response. We need to figure this out how to get around this under platform then. This component & ComboBox are for me candidates to write these components from the scratch.
> 
> Btw. do you still need binding `mobile=true` when you can detect this from the content?

Yes, `[mobile]="true"` is needed, because sometimes component needs to react differently to certain actions in mobile mode. Select for example abandons usage of Popover.

InnaAtanasova:
I'm also not sure about the `combobox` role. WAI_ARIA specifies that  the `A composite widget containing a single-line textbox and another element such as listbox or grid, that can dynamically pop up to help the user set the value of the textbox.`. In our case we cannot write inside the input field. We use it to trigger the menu and to show the user the selected value.

InnaAtanasova:
Also, a PR marked with feat won't be approved without updated Breaking Changes

apps/docs/src/app/core/component-docs/select/select-docs.component.html
```diff
@@ -1,107 +1,130 @@
...
 
-
-<fd-docs-section-title [id]="'Types'" [componentName]="'select'">
-    Select Border Types
+<fd-docs-section-title [id]="'semantic'" [componentName]="'select'">
+    Select Semantic States
 </fd-docs-section-title>
-<description>An example of the select component with diffrent types of Border type and ability to add icon.
-</description>
+<description></description>
```
Vanessa-Cusmich, line 18: The states of the Select control can be customized by assigning different border colors for each state.

@Michal: The description was removed here but I think it would be good to add one because all the other examples do :)

mikerodonnell89:
Clicked item has different active state here:
Screen Shot 2020-04-23 at 1 15 02 PM

vs styles:
Screen Shot 2020-04-23 at 1 14 50 PM

apps/docs/src/app/core/component-docs/select/select-header/select-header.component.html
```diff
@@ -1,11 +1,7 @@
 <header>Select</header>
 <description>
-    The select component was created to emulate the native select element as closely as possible while also providing
+    The Select component is composed of the SelectComponent and the OptionComponent. It was created to emulate the native Select element as closely as possible while also providing
     more extended styling options.
```
Vanessa-Cusmich, line 4: I moved the description of the component to the header (instead of the mode title) because you reference it in several descriptions. Can you confirm if it still makes sense?  Thanks :)

apps/docs/src/app/core/component-docs/select/select-docs.component.html
```diff
@@ -1,107 +1,130 @@
-<fd-docs-section-title [id]="'standard'" [componentName]="'select'">
-    Standard Select
+<fd-docs-section-title [id]="'mode'" [componentName]="'select'">
+    Select mMdes
```
Vanessa-Cusmich, line 2: Should this be 'Select Modes'?

mikerodonnell89:
First list item is focused after clicking the select - is this intentional?
Untitled

mikerodonnell89:
Programmatic open/close it broken
Untitled

mikerodonnell89:
Do we need to close the popover when an option is added or removed?  I think it would be good for the user to actually see this happen

libs/core/src/lib/utils/functions/is-key.ts
```diff
@@ -0,0 +1,27 @@
...
+    [
+        ['ArrowRight', {aliases: ['ArrowRight', 'Right'], keyCode: 39}],
+        ['ArrowDown', {aliases: ['ArrowDown', 'Down'], keyCode: 40}],
+        ['ArrowLeft', {aliases: ['ArrowLeft', 'Left'], keyCode: 37}],
+        ['ArrowUp', {aliases: ['ArrowUp', 'Up'], keyCode: 38}],
+        [' ', {aliases: ['Space', 'Spacebar', ' '], keyCode: 32}],
+        ['Escape', {aliases: ['Escape', 'Esc'], keyCode: 27}],
+        ['Delete', {aliases: ['Delete', 'Del'], keyCode: 46}],
+        ['Enter', {aliases: ['Enter'], keyCode: 13}],
+        ['Tab', {aliases: ['Tab'], keyCode: 9}],
```
mikerodonnell89, line 11: Not being able to check for shift+tab could be an issue in the future

libs/core/src/lib/select/select.component.ts
```diff
@@ -74,9 +86,17 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnInit, OnD
     @Input()
     isOpen: boolean = false;
```
JKMarkowski, line 87: Is there way, to control open state only programatically, so clicking inside won't trigger this flag?

libs/core/src/lib/select/option/option.component.ts
```diff
@@ -1,41 +1,50 @@
...
-    fdMenuItemClass: boolean = true;
+        .fd-list__item[aria-disabled="true"],
+        .fd-list__item.is-disabled,
+        .fd-list__item:disabled {
+            opacity: 0.4;
+            cursor: not-allowed;
+        }
+        .fd-list__item:disabled:focus {
+            outline: none;
+        }
```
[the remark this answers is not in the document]
salarenko: Fundamental Styles List component doesn't support list item disabled state styling (at least I couldn't find it in `v0.8.1`). I added the styles temporarily to NGX until we bring this in Fundamental Styles.

libs/core/src/lib/select/option/option.component.ts
```diff
@@ -49,38 +58,61 @@ export class OptionComponent implements OnInit {
...
-    /** Emitted when the selected state changes. */
-    @Output()
-    readonly selectedChange: EventEmitter<OptionComponent>
-        = new EventEmitter<OptionComponent>();
+    /** @hidden */
+    @HostBinding('class.is-selected')
+    selected: boolean = false;
+
+    /** @hidden */
+    private _subscriptions: Subscription = new Subscription();
```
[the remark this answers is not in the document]
salarenko: This is one Subscription instance which i use as an aggregate for other subscription cancellation functions using `add()` and call all of them with `unsubscribe()` when component is destroyed.

fkolar:
> I'm also not sure about the `combobox` role. WAI_ARIA specifies that the `A composite widget containing a single-line textbox and another element such as listbox or grid, that can dynamically pop up to help the user set the value of the textbox.`. In our case we cannot write inside the input field. We use it to trigger the menu and to show the user the selected value.

For Aria the best source is to look at material implementation (autocomplete), they have all the required aria attributes, and this should be good source to double check for all of our components.

libs/core/src/lib/utils/functions/is-key.ts
JKMarkowski: For `shift/crtl/alt` + any key, there is flag in object keybaord event object
image

libs/core/src/lib/utils/functions/is-key.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @param event - KeyboardEvent
+ * @param key   - event.key name matching W3C specification
+ * */
+export function isKey(event: KeyboardEvent, key: string) {
+    if (event && keyMap.get(key)) {
+        return keyMap.get(key).aliases.some(alias => alias === event.key)
+            || keyMap.get(key).keyCode === event.keyCode;
+    }
+
+    throw new Error(`Invalid function arguments. Check if "event" is a KeyboardEvent or "key" is defined in keyMap`);
```
JKMarkowski, line 25: Can you throw it only on devMode? I'm not sure if end-user should see this kind of errors

libs/core/src/lib/select/select.component.ts
salarenko: No there is not. I was also wondering what is the correct behavior for such scenario, but in my opinion isOpen should always reflect select open state.

salarenko:
@InnaAtanasova 
You are right about aria `combobox`, I changed it to `listbox` and added also `aria-labelledby` and `aria-activedescendant`.

@mikerodonnell89 @InnaAtanasova 
Select popover is closing in Add/Remove options example because of `closeOnOutsideClick="true"` I switched it to `false` for this example.

@mikerodonnell89 
Could you precise how to reproduce this behavior? For me its:
image

I also tried to play around with "Programmatic open/close is broken" but I cannot break it. Is there some certain way to do it? Did you found some errors in console?

I'm not sure of we should extend `isKey` with combination keys functionality. The purpose of this util is unification of key codes for different browsers.

@JKMarkowski 
`isKey` now throws errors only in devMode

stefanoScalzo:
I dont think that Select in loading mode should not be clickable

stefanoScalzo:
Should we imliment some type of close when we click outside of the options for mobile view, they currently have no way to go back without selecting an option

stefanoScalzo:
For removing options, should we give the user the ability to choose the min , same for adding and max

stefanoScalzo:
The programatic control is broken because when you click change value it selects randomly and not one after the other

salarenko:
@stefanoScalzo 
> I dont think that Select in loading mode should not be clickable

That's a good thing to discuss. To be honest I just went through the Fiori3 specs and it says nothing about Loading Mode so we probably event shouldn't introduce it in NGX Core.

> Should we imliment some type of close when we click outside of the options for mobile view, they currently have no way to go back without selecting an option

This is only an example, user can modify Dialog in any way he wants to. You can use `Esc` to close the Dialog. Anyway I just added close button to the example.

> For removing options, should we give the user the ability to choose the min , same for adding and max

I don't think there is a need of creating upper and lower limits for number of option in select.

> The programatic control is broken because when you click change value it selects randomly and not one after the other

I don't agree with that. Change Value is just changing value. Why should it change to the next value in the options list?

apps/docs/src/app/core/component-docs/select/select-docs.component.html
```diff
@@ -1,9 +1,8 @@
 <fd-docs-section-title [id]="'mode'" [componentName]="'select'">
-    Select mMdes
+    Select Modes
 </fd-docs-section-title>
 <description>
-    The Select control supports a wide variety of modes to display your list items. <br/>
-    
+    Following examples presents different modes of the Select control.
```
Vanessa-Cusmich, line 5: The following examples present the different modes of the Select control.

apps/docs/src/app/core/component-docs/select/select-docs.component.html
```diff
@@ -1,107 +1,131 @@
...
+<fd-docs-section-title [id]="'custom-select-template'" [componentName]="'select'">
+    Custom Control Content
 </fd-docs-section-title>
-
 <description>
-    Controlling the select through programmatic methods is possible by calling the select component instance directly.
-    Note that this can be done through ViewChild as well.
+    The content of the Select control can be customized using a custom template which needs to be referenced in the
+    <code>[controlTemplate]</code> input.<br/>
+    Select provides the following context values to the template:
```
Vanessa-Cusmich, line 33: Should we move this sentence to line 32 before the break? It looks weird to me.
Screen Shot 2020-04-27 at 9 18 16 AM

apps/docs/src/app/core/component-docs/select/select-docs.component.html
```diff
@@ -1,107 +1,130 @@
...
+
+        <li>
+            To get the full Fiori3 experience, set Select in mobile mode using <code>[mobile]="true"</code>. Also remember to instantiate the
+            Dialog inside Select using <code>DialogConfig.container</code> and <code>Select.dialogContainerElementRef</code>.
+        </li>
+
+        <li>
+            The Options template needs to be declared inside the <code>fd-select</code> tags to properly configure element
+            injectors. Use <code>Select.selectOptionsListTemplate</code> to render options in the desired place.
+        </li>
```
Vanessa-Cusmich, line 96: Should we add breaks between the bullet points to improve readability?

Screen Shot 2020-04-27 at 9 29 34 AM

stefanoScalzo:
> @stefanoScalzo
> 
> > I dont think that Select in loading mode should not be clickable
> 
> That's a good thing to discuss. To be honest I just went through the Fiori3 specs and it says nothing about Loading Mode so we probably event shouldn't introduce it in NGX Core.
>  Perfect ill put it in the chat
> > Should we imliment some type of close when we click outside of the options for mobile view, they currently have no way to go back without selecting an option
> 
> This is only an example, user can modify Dialog in any way he wants to. You can use `Esc` to close the Dialog. Anyway I just added close button to the example.
> Perfect
> > For removing options, should we give the user the ability to choose the min , same for adding and max
> 
> I don't think there is a need of creating upper and lower limits for number of option in select.
> What if the user only wants the minimum to go down to 2 and not 1
> > The programatic control is broken because when you click change value it selects randomly and not one after the other
> 
> I don't agree with that. Change Value is just changing value. Why should it change to the next value in the options list?

Well, I think the user should put a function for the next choice. Not that it is done completely random

libs/core/src/lib/select/select.component.ts
```diff
@@ -142,98 +170,154 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnInit, OnD
...
+            }
+            event.preventDefault();
+        }
+    }
+
+    /** @hidden */
+    @HostListener('window:resize')
+    resizeScrollHandler(): void {
+        this.calculatedMaxHeight = window.innerHeight * 0.45;
+    }
```
JKMarkowski, line 240: Could you add this method also to `ngAfterViewInit`? calcualtedMaxHeight is only, when someone resize window

salarenko:
@JKMarkowski @rengare @InnaAtanasova @mikerodonnell89 @stefanoScalzo @fkolar 

Could you guys take a looks at changes brought with [this recent commit](https://github.com/SAP/fundamental-ngx/pull/2150/commits/84555c9ecc4bb68ce5578182c0fbced6eb5239ee)?

Changes introduced:
- Reorganize `isKey` utility to `KeyUtility` class with static methods:
    - `isKey` - cross browser key unification
    - `isKeyType(alpa | num)` - cross browser alphabetical / numerical key recognition
- Add to `KeyUtility.isKey` possibility to pass collection of keys to check e.g. `KeyUtility.isKey(event, ['Space', 'Enter'])`
- Add selecting options using `ArrowUp` | `ArrowDown` when select is closed.
list-keyboard-select
- Add selecting options using alphanumeric keys
    - If searched phrase has been typed "quickly" Select search and select element by phrase
    - If subset of options matching query is > 1, Select should:
        - Interact with first option if no other option in subset is currently active
        - Interact with next option in the subset after currently active option in the subset
    - Interaction options:
        - Selection if the Select is closed
list-alpha-select
        - Focus if the select is opened
list-alpha-select-opened

I recommend testing alphanumerical selection on [this example](https://deploy-preview-2150--fundamental-ngx.netlify.app/#/core/select#select-max-height)

libs/platform/src/lib/components/select/select.component.html
```diff
@@ -1,29 +1,28 @@
...
-        [viewValue]="item.label"
-        [value]="item"
-    >
-        <ng-template [ngIf]="!optionValueTemplate"
-            ><fd-icon [glyph]="item.icon" size="m" class="fd-template-icon icon-margin"></fd-icon
-            >{{ item.label ? item.label : item }}</ng-template
-        >
+
+    <fd-option *ngFor="let item of list ; index as index"
+               [disabled]="item.disabled ? true : false"
```
rengare, line 13: ` [disabled]="item.disabled" `

libs/core/src/lib/select/select.component.ts
```diff
@@ -142,98 +171,160 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnInit, OnD
...
-            return options.changes.pipe(
-                startWith(options),
-                switchMap(() => merge(...options.map((option) => option.selectedChange)))
-            );
-        }
-    }) as Observable<OptionComponent>;
+    selected: OptionComponent;
+
+    /** @hidden */
+    optionsArray: OptionComponent[];
```
rengare, line 197: `options: OptionComponent[];`

optionsArray name is redundant because you can have many options and an array is a type

libs/core/src/lib/select/select.component.ts
```diff
@@ -131,8 +154,14 @@ export class SelectComponent implements OnChanges, AfterContentInit, OnInit, OnD
...
     unselectMissingOption: boolean = true;
+
     /** If user wants to disable clicking when the content has not yet loaded and apply the three dots. */
-    @Input() loading: boolean = false;
+    @Input()
+    loading: boolean = false;
+
+    /** Binds to control aria-labelledby attribute */
+    @Input()
+    labelledby: string;
```
rengare, line 164: `labelledBy`

libs/platform/src/lib/components/select/select.component.html
salarenko: This is platform

libs/core/src/lib/button-group/button-group.component.scss
```diff
@@ -1,3 +1,3 @@
 // @import '~fundamental-styles/dist/button-group';
 @import '~fundamental-styles/dist/button';
-@import '~fundamental-styles/dist/icon';
+@import "~fundamental-styles/dist/icon";
```
rengare, line 3: kinda strange that it changed to double quote 
image
(btw button-group styles are commented out but I'll handle that)

rengare, line 3: `@import '~fundamental-styles/dist/icon';`

salarenko:
@rengare I addressed you comments.
I also:
- Removed `Select.[isOpen]`
- Removed `Option.[viewValue]`
- Added `Select.[typeaheadDebounceInterval]`
- Changed `selectViewValue` from getter to Select property updated only when needed

libs/platform/src/lib/components/select/select.component.ts
```diff
@@ -77,13 +78,12 @@ export class SelectPlatformComponent extends fdSelect implements OnInit, AfterCo
      */
     private _value: any;
 
-    constructor(private cd: ChangeDetectorRef, private rtl: RtlService) {
-        super(cd, rtl);
-        this.fdDropdownClass = false;
+    constructor(private cd: ChangeDetectorRef, elementRef: ElementRef) {
+        super(elementRef, new SelectProxy(), cd);
```
fkolar, line 82: We need change this but this is just a comment. - we need to get rid of this inheritance.

libs/core/src/lib/select/select-proxy.service.ts
```diff
@@ -0,0 +1,9 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject, Subject } from 'rxjs';
+import { OptionStatusChange } from './select.component';
+
+@Injectable()
+export class SelectProxy {
```
fkolar, line 6: What is the added value moving all this to extra service ?

libs/core/src/lib/select/select.module.ts
```diff
@@ -16,10 +15,9 @@ import { BusyIndicatorModule } from '../busy-indicator/busy-indicator.module';
         BusyIndicatorModule,
         CommonModule,
         PopoverModule,
-        MenuModule,
         ButtonModule,
         IconModule,
-        LoadingSpinnerModule
+        ListModule
```
fkolar, line 20: Isn't this too big overhead to import module with all its directives ?

libs/core/src/lib/select/select-proxy.service.ts
salarenko: Previous implementation included cyclic dependency:
- Select to Option with `ContentChildren(OptionComponent)`
- Option to Select with element injector `constructor(SelectComponent)`

My first try was implementing a communication between Select an Option based on:
- Select to Option with `ContentChildren(OptionComponent)`
- Select subscribes to `Option.onOptionSelect()` emitter
but once again I couldn't escape change detection error and the code have got more complicated because of access to Select->Option in late change detection phase.

I decided to create communication proxy between Select and Option to escape cyclic dependency.

libs/core/src/lib/select/select.module.ts
salarenko: Well it is. The only alternatives I see are:
- creating a `List->Select` dedicated module
- implementing list features in Select and abandon `ListModule` import
but both of them seems to introduce additional maintenance. This is a thing to discuss with the team if we want to and how we want to handle it.

Issue #2160, opened 2020-03-11
Breadcrumb containerElement input not working


stefanoScalzo:
#### Is this a bug, enhancement, or feature request?
bug
#### Briefly describe your proposal.
The containerElement input is not working shown when we do not use it, the direct parent does not assume control over elipses
#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
latest
#### If this is a bug, please provide steps for reproducing it.
put breadcrumb in a div without the input and it does not create the popover box
![Screen Shot 2020-03-11 at 5 13 50 PM](https://user-images.githubusercontent.com/50607147/76464599-06f5e080-63bc-11ea-91aa-04cb0adcaed0.png)

#### Please provide relevant source code if applicable.
https://stackblitz.com/run?embed=1&file=src/app/fd-breadcrumb-responsive-example.component.html

PR #2185, opened 2020-03-16
fix: fixed-width breadcrumb containers not collapsing overflowing breadcrumbs

mikerodonnell89:
Also explain that the responsive behavior will not work if the container has a fixed width.

#2160 

Before: 
Screen Shot 2020-03-19 at 11 52 17 AM

After:
Screen Shot 2020-03-19 at 11 52 51 AM


rengare:
@mikerodonnell89 please attach before and after screenshots

PR #2187, opened 2020-03-17
fix: bring icons and fonts

InnaAtanasova:
This PR is bringing the fonts and the icons into the feature branch.

### DIFF
--- a/libs/platform/src/lib/components/button/button.component.ts
+++ b/libs/platform/src/lib/components/button/button.component.ts
@@ -37,6 +37,9 @@ export class ButtonComponent {
     @Input()
     type: ButtonType;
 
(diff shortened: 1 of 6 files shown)


libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -37,6 +37,9 @@ export class ButtonComponent {
     @Input()
     type: ButtonType;
 
+    @Input()
+    fdType: ButtonType;
```
stefanoScalzo: dont think this should be included in this pr

rengare: it should be there

Issue #2136, opened 2020-03-06
Not able to modify the type of the input in Input Groups


InnaAtanasova:
#### Is this a bug, enhancement, or feature request?
Bug
#### Briefly describe your proposal.
The current implementation of Input Groups does not allow the user to specify the type of the input. Only `text` type is possible. 
*Expected Behaviour:* the user should be able to specify the type
*Implementation in NGX:*
<img width="807" alt="Screen Shot 2020-03-06 at 10 12 45 AM" src="https://user-images.githubusercontent.com/39598672/76096109-7f5f3a80-5f93-11ea-82bf-b61dbe310f99.png">
*User Case:*
`<fd-input-group [placement]="'after'" [addOnText]="'%'" [placeholder]="'Percentage'" type="number" formControlName="percentage"></fd-input-group>`

PR #2190, opened 2020-03-18
fix: (input group) Add type of input group and example

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/2136
There is added way to add `type` of input inside input-group component. 

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


JKMarkowski:
IT was mostly caused by `RTLService` not provided in stackblitz module. I added it to providers array. Also switch had wrong selector.

stefanoScalzo:
I think you should make it more clear as to what the different types of input groups are in the documentation. You say search but what does that do exactly. should be more clear to the user

PR #2203, opened 2020-03-19
fix: add icons styles to dependant components

InnaAtanasova:
Bring `@import "~fundamental-styles/dist/icon";` to all components that rely on icons, such as Buttons, Input-Groups, etc.


JKMarkowski:
Hi @InnaAtanasova, Could you change importing scss files, to just import `IconModule` into certain modules? Your solution will make our lib bigger, because css code will be repeated in most components.

Issue #2135, opened 2020-03-06
Default placeholder for DatePicker should change base on date format


vin-chen:
#### 
Now the placeholder for DatePicker is by defaut use 'mm/dd/yyyy', even if the user set `format` for it, it won't change. This would confuse the user.
eg. placeholder: `mm/dd/yyyy`, but actual format is `yyyy/dd/mm`

https://github.com/SAP/fundamental-ngx/blob/8877b24ea7173fde8c4ec4af062534c13991c933/libs/core/src/lib/date-picker/date-picker.component.ts#L76

mikerodonnell89:
Due to how custom formatting is handled via the DateFormatParser and DateTimeFormatParser abstract classes, where the application developer must implement their own format and parse functions, I believe setting the placeholder is best left to the application developer as it can not be safely assumed by the component.  I think we should leave the current placeholder as-is because it matches the current default format, however I will make this more clear in the documentation

PR #2206, opened 2020-03-20
docs: improve datepicker/datetimepicker placeholder formatting info

mikerodonnell89:

#2135 

Elaborate on setting placeholders for datepickers and datetimepickers with custom formatting

### DIFF
--- a/apps/docs/src/app/core/component-docs/datetime-picker/datetime-picker-docs.component.html
+++ b/apps/docs/src/app/core/component-docs/datetime-picker/datetime-picker-docs.component.html
@@ -58,7 +58,10 @@
 <fd-docs-section-title [id]="'datetime-picker-formatting'" [componentName]="'datetime-picker'">
     Formatting
 </fd-docs-section-title>
-<description>Providing a custom format for the dates is possible through providing a service.</description>
+<description>
+    Providing a custom format for the dates is possible through providing a service.
+    Note that when providing a custom format, the <code>[placeholder]</code> input should be set to match the new format as well.
+</description>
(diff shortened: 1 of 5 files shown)


droshev:
@mikerodonnell89 if we touch `libs/core/src/lib/date-picker/date-picker.component.ts` then the PR should be a fix

apps/docs/src/app/core/component-docs/datetime-picker/datetime-picker-docs.component.html
```diff
@@ -58,7 +58,10 @@
 <fd-docs-section-title [id]="'datetime-picker-formatting'" [componentName]="'datetime-picker'">
     Formatting
 </fd-docs-section-title>
-<description>Providing a custom format for the dates is possible through providing a service.</description>
+<description>
+    Providing a custom format for the dates is possible through providing a service.
+    Note that when providing a custom format, the <code>[placeholder]</code> input should be set to match the new format as well.
```
salarenko: > Note that when providing a custom format, the <code>[placeholder]</code> input should be set to match the new format as well.

Is is true with implementation changes introduced in this PR? Now when you set `[format]` input placeholder is automatically set to custom format.

mikerodonnell89: Ah I accidentally committed that.  It was from my first attempt at fixing this problem.  However the [format] input is actually not used for this purpose and due to DateTimeFormatParser/DateFormatParser it is not possible to assume the placeholder

Issue #2222, opened 2020-03-24
fix: (Platform) Menu Button Standalone component


DeepakSap14:
#### Is this a bug, enhancement, or feature request?
Enhancement

#### Briefly describe your proposal.
A new standalone component of menu button component

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Anguar v9

PR #2223, opened 2020-03-24
fix: (platform) Menu button As a component

DeepakSap14:
https://github.com/SAP/fundamental-ngx/issues/2222

A new standalone component of menu button component

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/apps/docs/src/app/platform/component-docs/platform-menu-button/platform-menu-button-examples/platform-menu-button-compact-example.component.html
+++ b/apps/docs/src/app/platform/component-docs/platform-menu-button/platform-menu-button-examples/platform-menu-button-compact-example.component.html
@@ -1,0 +1,149 @@
+<fdp-menu-button
+    [id]="'menu-button-0'"
+    [displaySize]="'compact'"
+    [icon]="'world'"
+    [disabled]="false"
+    (click)="userDefined($event)"
+    [fdpMenuTriggerFor]="basicMenu"
+>
+    Default Button
+</fdp-menu-button>
+
+<span> &nbsp;&nbsp;&nbsp;</span>
+<fdp-menu-button
+    [id]="'menu-button-2'"
+    [displaySize]="'compact'"
+    [icon]="'world'"
+    [disabled]="false"
+    [type]="'emphasized'"
+    (click)="userDefined($event)"
+    [fdpMenuTriggerFor]="basicMenu"
+>
+    Emphasized Button
+</fdp-menu-button>
+
+<span> &nbsp;&nbsp;&nbsp;</span>
+<fdp-menu-button
+    [id]="'menu-button-2'"
+    [displaySize]="'compact'"
+    [icon]="'world'"
+    [disabled]="false"
+    [type]="'ghost'"
+    (click)="userDefined($event)"
+    [fdpMenuTriggerFor]="basicMenu"
+>
+    Ghost Button
+</fdp-menu-button>
+
+<span> &nbsp;&nbsp;&nbsp;</span>
+<fdp-menu-button
+    [id]="'menu-button-1'"
+    [displaySize]="'compact'"
+    [icon]="'world'"
+    [disabled]="false"
+    [type]="'positive'"
+    (click)="userDefined($event)"
+    [fdpMenuTriggerFor]="basicMenu"
+>
+    Positive Button
+</fdp-menu-button>
+
+<span> &nbsp;&nbsp;&nbsp;</span>
+<fdp-menu-button
+    [id]="'menu-button-3'"
+    [displaySize]="'compact'"
+    [icon]="'world'"
+    [disabled]="false"
+    [type]="'negative'"
+    (click)="userDefined($event)"
+    [fdpMenuTriggerFor]="cascadeMenu"
+>
+    Negative Button
+</fdp-menu-button>
+
+<span> &nbsp;&nbsp;&nbsp;</span>
+<fdp-menu-button
+    [id]="'menu-button-4'"
+    [displaySize]="'compact'"
+    [icon]="'world'"
+    [disabled]="false"
+    [type]="'attention'"
+    (click)="userDefined($event)"
+    [fdpMenuTriggerFor]="cascadeMenu"
+>
+    Attention Button
+</fdp-menu-button>
+
+<span> &nbsp;</span>
+<fdp-menu-button
+    [id]="'menu-button-5'"
+    [displaySize]="'compact'"
+    [icon]="'world'"
+    [disabled]="false"
+    [type]="'transparent'"
+    (click)="userDefined($event)"
+    [fdpMenuTriggerFor]="cascadeMenu"
+>
+    Transparent Button
+</fdp-menu-button>
+
+<span> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</span>
+
+<fdp-menu #basicMenu id="basic-menu">
+    <fdp-menu-item (itemSelect)="onItemSelect('First Item')">First Item</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Second Item')">Second Item</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Third Item')">Third Item</fdp-menu-item>
+</fdp-menu>
+
+<fdp-menu #cascadeMenu id="cascade-menu">
+    <fdp-menu-item [fdpMenuTriggerFor]="fruitMenu">Fruits</fdp-menu-item>
+    <fdp-menu-item [fdpMenuTriggerFor]="vegetableMenu">Vegetables</fdp-menu-item>
+</fdp-menu>
+
+<fdp-menu #fruitMenu id="fruit-menu">
+    <fdp-menu-item [fdpMenuTriggerFor]="appleMenu">Apple</fdp-menu-item>
+    <fdp-menu-item [fdpMenuTriggerFor]="bananaMenu">Banana</fdp-menu-item>
+    <fdp-menu-item [fdpMenuTriggerFor]="orangeMenu">Orange</fdp-menu-item>
+</fdp-menu>
+
+<fdp-menu #appleMenu id="apple-menu">
+    <fdp-menu-item (itemSelect)="onItemSelect('Braeburn Apple')">Braeburn Apple</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Fuji Apple')">Fuji Apple</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Gala Apple')">Gala Apple</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Honeycrisp Apple')">Honeycrisp Apple</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Red Delicious Apple')">Red Delicious Apple</fdp-menu-item>
+</fdp-menu>
+
+<fdp-menu #bananaMenu id="banana-menu">
+    <fdp-menu-item (itemSelect)="onItemSelect('Cavendish Banana')">Cavendish Banana</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Lady Finger Banana')">Lady Finger Banana</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Red Banana')">Red Banana</fdp-menu-item>
+</fdp-menu>
+
+<fdp-menu #orangeMenu id="orange-menu">
+    <fdp-menu-item (itemSelect)="onItemSelect('Blood Orange')">Blood Orange</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Navel Orange')">Navel Orange</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Valencia Orange')">Valencia Orange</fdp-menu-item>
+</fdp-menu>
+
+<fdp-menu #vegetableMenu id="vegetable-menu">
+    <fdp-menu-item [fdpMenuTriggerFor]="carrotMenu">Carrot</fdp-menu-item>
+    <fdp-menu-item [fdpMenuTriggerFor]="lettuceMenu">Lettuce</fdp-menu-item>
+</fdp-menu>
+
+<fdp-menu #carrotMenu id="carrot-menu">
+    <fdp-menu-item (itemSelect)="onItemSelect('Nantes Carrot')">Nantes Carrot</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Imperator Carrot')">Imperator Carrot</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Chantenay Carrot')">Chantenay Carrot</fdp-menu-item>
+</fdp-menu>
+
+<fdp-menu #lettuceMenu id="lettuce-menu">
+    <fdp-menu-item (itemSelect)="onItemSelect('Arugula')">Arugula</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Belgian Endive')">Belgian Endive</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Butter Lettuce')">Butter Lettuce</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Iceberg Lettuce')">Iceberg Lettuce</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Red Leaf Lettuce')">Red Leaf Lettuce</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Romaine Lettuce')">Romaine Lettuce</fdp-menu-item>
+</fdp-menu>
+
+<div style="margin-top: 1rem;"><label>Item Selected:</label> {{ item }}</div>
--- a/apps/docs/src/app/platform/component-docs/platform-menu-button/platform-menu-button-examples/platform-menu-button-compact-example.component.html
+++ b/apps/docs/src/app/platform/component-docs/platform-menu-button/platform-menu-button-examples/platform-menu-button-compact-example.component.html
@@ -1,6 +1,6 @@
 <fdp-menu-button
     [id]="'menu-button-0'"
-    [displaySize]="'compact'"
+    [contentSize]="'compact'"
     [icon]="'world'"
     [disabled]="false"
     (click)="userDefined($event)"
@@ -12,7 +12,7 @@
 <span> &nbsp;&nbsp;&nbsp;</span>
 <fdp-menu-button
     [id]="'menu-button-2'"
-    [displaySize]="'compact'"
+    [contentSize]="'compact'"
     [icon]="'world'"
     [disabled]="false"
     [type]="'emphasized'"
@@ -25,7 +25,7 @@
 <span> &nbsp;&nbsp;&nbsp;</span>
 <fdp-menu-button
     [id]="'menu-button-2'"
-    [displaySize]="'compact'"
+    [contentSize]="'compact'"
     [icon]="'world'"
     [disabled]="false"
     [type]="'ghost'"
@@ -38,7 +38,7 @@
 <span> &nbsp;&nbsp;&nbsp;</span>
 <fdp-menu-button
     [id]="'menu-button-1'"
-    [displaySize]="'compact'"
+    [contentSize]="'compact'"
     [icon]="'world'"
     [disabled]="false"
     [type]="'positive'"
@@ -51,7 +51,7 @@
 <span> &nbsp;&nbsp;&nbsp;</span>
 <fdp-menu-button
     [id]="'menu-button-3'"
-    [displaySize]="'compact'"
+    [contentSize]="'compact'"
     [icon]="'world'"
     [disabled]="false"
     [type]="'negative'"
@@ -64,7 +64,7 @@
 <span> &nbsp;&nbsp;&nbsp;</span>
 <fdp-menu-button
     [id]="'menu-button-4'"
-    [displaySize]="'compact'"
+    [contentSize]="'compact'"
     [icon]="'world'"
     [disabled]="false"
     [type]="'attention'"
@@ -77,7 +77,7 @@
 <span> &nbsp;</span>
 <fdp-menu-button
     [id]="'menu-button-5'"
-    [displaySize]="'compact'"
+    [contentSize]="'compact'"
     [icon]="'world'"
     [disabled]="false"
     [type]="'transparent'"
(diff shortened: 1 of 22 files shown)


DeepakSap14:
> The code looks fine. However when I checked out your branch on my local, I noticed that your branch is running on Angular 8. I believe the current master branch is on Angular 9 now. Have you done a rebase against the latest master branch?

I have rebased master on my branch.
I have run 'build-all' as well on my system. seems to be fine.
my Angular version:
"Angular CLI: 9.0.7
Node: 10.16.3
OS: darwin x64

Angular: 9.0.6"

InnaAtanasova:
Please follow the Contribution Guidelines for naming your PR. If there are Breaking Changes please update the documentation, if there are no BC you can check the checkbox in the template. You need all checkboxes to be checked in order for the Mergeable validation to pass.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
```diff
@@ -0,0 +1,75 @@
...
+})
+export class MenuButtonComponent implements OnInit {
+    /** Id for menu-button. */
+    @Input()
+    Id: string;
+
+    /** Option to make to menu-button compact. */
+    @Input()
+    size: InputSize = 'cozy';
+
```
InnaAtanasova: It's not a comment, just a  question. Why do you go with this approach of having `size` and `iscompact`? You can have an Input property called `compact` which can be boolean and set to false by default. If you need the button to be in compact mode you can set it to true. Having size property can be confusing to the user.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
```diff
@@ -0,0 +1,75 @@
...
+    /** @hidden */
+    menuOptions: string[] = new Array();
+
+    _iscompact: boolean = false;
+
+    constructor() {}
+
+    /** @hidden */
+    ngOnInit() {
+        this._iscompact = this.size === 'compact' ? true : false;
```
InnaAtanasova: I don't think ngOnInit is the best place for this

libs/platform/src/lib/components/menu-button/menu-button.component.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    /** Event sent when menu-button is clicked */
+    @Output()
+    click: EventEmitter<MouseEvent | KeyboardEvent | TouchEvent> = new EventEmitter();
+
+    /** add 'menu' as button option */
+    /** @hidden */
+    menuOptions: string[] = new Array();
+
+    _iscompact: boolean = false;
```
InnaAtanasova: In this case you need to declare this as `private` and use `\** @hidden *\`

libs/platform/src/lib/components/menu-button/menu-button.component.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    constructor() {}
```
InnaAtanasova: Should be hidden

libs/platform/src/lib/components/menu-button/menu-button.component.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+    it('should create', () => {
+        expect(component).toBeTruthy();
+    });
+
+    it('should have a disabled menu button', () => {
+        spyOn(component, 'setDisabledState');
+        component.setDisabledState(true);
+        expect(component.setDisabledState).toHaveBeenCalled();
+    });
+
```
InnaAtanasova: I would add tests for compact mode and for type (check if type is applied)

libs/platform/src/lib/components/menu-button/menu-button.component.scss
```diff
@@ -0,0 +1,7 @@
+.fdp-menu-btn-truncate__txt {
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    display: inline-block;
+    vertical-align: bottom;
+}
```
InnaAtanasova: I might be wrong but the truncation behavior is coming from Fundamental-styles. If it's not it's probably better to add it there instead of having it here

libs/platform/src/lib/components/menu-button/menu-button.component.ts
DeepakSap14: we are following this in Platform as per Frank comments in other components.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
DeepakSap14: size value will be available in Oninit. you want me to move it into constructor if this approach is followed.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
DeepakSap14: private variable will not be visible in template. hence can't be used in template.

libs/platform/src/lib/components/menu-button/menu-button.component.spec.ts
DeepakSap14: will do.

libs/platform/src/lib/components/menu-button/menu-button.component.scss
DeepakSap14: will make a separate PR for this. will raise PR in this week.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
InnaAtanasova: Then why do you prefix it with `_` ? Usually when we see `_varName` we assume it's private

libs/platform/src/lib/components/menu-button/menu-button.component.ts
InnaAtanasova: Ok. Thank you for answering my question

libs/platform/src/lib/components/menu-button/menu-button.component.ts
InnaAtanasova: I am just wondering how this property is updated after the component has been initialized. Let's say you have a Playground with a checkbox and you change the value of `_iscompact` based on the value of the checkbox. How is the change detection triggered?

libs/platform/src/lib/components/menu-button/menu-button.component.ts
InnaAtanasova: Also needs TypeDocs for the API

libs/platform/src/lib/components/menu-button/menu-button.component.scss
InnaAtanasova: Thank you!

libs/platform/src/lib/components/menu-button/menu-button.component.spec.ts
```diff
@@ -0,0 +1,37 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { MenuButtonComponent } from './menu-button.component';
+import { ButtonModule } from '@fundamental-ngx/core';
+
```
fkolar: I dont see any tests.

libs/platform/src/lib/components/menu-button/menu-button.component.html
```diff
@@ -0,0 +1,13 @@
+<button
+    fd-button
+    [options]="menuOptions"
+    [compact]="_iscompact"
```
fkolar: when you prefix something as `_` then it means its private and when something is private you should not really use in the template

libs/platform/src/lib/components/menu-button/menu-button.component.ts
fkolar: > Ok. Thank you for answering my question

@InnaAtanasova  I dont really see any explaination ;-(,  @DeepakSap14 you should try to explain as `per Frank` does not mean much and does not tell a thing to anyone ;-(

fkolar: We have introduced `size: Compact | cozy` for input Widgets and it is defined in the BaseInput. The reason for this was since we know that every INPUT is going to expand 100% to its parent container so most likely we are not going to use native `size` attribute and this cozy | compact is kind of size so we used this so every component have unified naming for this. From expirience we expect it will not end only with these two compact and cozy, they might be other.  I am still thinking about some better name.. 

This was for Inputs. As for the menu button, can you define the actual width? 

For the non-input and I see there is a `width` in this case it might be confusing size, width .

libs/platform/src/lib/components/menu-button/menu-button.component.ts
```diff
@@ -0,0 +1,75 @@
...
+    options: ButtonOptions;
+
+    /** Event sent when menu-button is clicked */
+    @Output()
+    click: EventEmitter<MouseEvent | KeyboardEvent | TouchEvent> = new EventEmitter();
+
+    /** add 'menu' as button option */
+    /** @hidden */
+    menuOptions: string[] = new Array();
+
```
fkolar: this does not make much sense to have a some getter for isCOmpact, there is no added value. you can just do size === 'compact' in the template

libs/platform/src/lib/components/menu-button/menu-button.component.ts
fkolar: this is extra code that does not have to be there.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
```diff
@@ -0,0 +1,75 @@
+import { Component, ChangeDetectionStrategy, OnInit, Input, Output, EventEmitter } from '@angular/core';
+import { InputSize } from './../form/form-control';
+import { ButtonType } from '../button/button.component';
+import { ButtonOptions } from '../button/button.component';
+
+@Component({
+    selector: 'fdp-menu-button',
+    templateUrl: './menu-button.component.html',
+    styleUrls: ['./menu-button.component.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
```
fkolar: Again, since we are using onPush change detection strategy, you want to take under control change detection. There are cases, that might not update as you are not handling it. 

You want to probably inject ChangeDectionRef. Also what about RTL?  Do we need to insert RTL service or it is handled by core template ?

libs/platform/src/lib/components/menu-button/menu-button.component.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    _iscompact: boolean = false;
+
+    constructor() {}
+
+    /** @hidden */
+    ngOnInit() {
+        this._iscompact = this.size === 'compact' ? true : false;
+        this.menuOptions.push('menu'); // for dropdown icon
+        if (this.options) {
```
fkolar: Why do you need two properties to do the same ?

libs/platform/src/lib/components/menu-button/menu-button.component.ts
```diff
@@ -0,0 +1,75 @@
...
+    @Input()
+    Id: string;
+
+    /** Option to make to menu-button compact. */
+    @Input()
+    size: InputSize = 'cozy';
+
+    /** Option to truncate content of the menu-button based on width. */
+    @Input()
+    width: string;
```
fkolar: Here I agree `width` and `size` is confusing how do you know as user which one is which?

libs/platform/src/lib/components/menu-button/menu-button.component.spec.ts
DeepakSap14: will be writing test cases.

libs/platform/src/lib/components/menu-button/menu-button.component.html
DeepakSap14: will remove and use size=='compact' in template.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
DeepakSap14: will remove and use size=='compact' in template.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
DeepakSap14: will be removed. as size=='compact' in template will be used.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
DeepakSap14: can you show me a example on how to use RTL service, i  did not find in any component developed till now. i will remove ChangeDetectionStrategy.OnPush, hope this solves the problem and i don't have to include 'ChangeDectionRef'.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
DeepakSap14: bcz options is string and menuOptions is array. do you want me to handle it a variable?

libs/platform/src/lib/components/menu-button/menu-button.component.ts
DeepakSap14: what variable should i put instead of size. please suggest?

libs/platform/src/lib/components/menu-button/menu-button.component.ts
DeepakSap14: i don't see RTL handled in core template as well

libs/platform/src/lib/components/menu-button/menu-button.component.ts
InnaAtanasova: Thank you for the detailed explanation, @fkolar It makes sense. From "we are following this in Platform as per Frank comments in other components" I understood that this is an approach you agreed on for the Platform and it has been discussed with you.

libs/platform/src/lib/components/menu-button/menu-button.component.scss
KevinOkamoto: I'm seeing the following misalignment of the menu in relation to the button:

Screen Shot 2020-03-25 at 3 53 28 PM

You need to add the following CSS or else the menu will not be positioned properly at the bottom of the menu button:
```
:host {
    display: inline-block;
}
```

DeepakSap14: Thanks, had missed it. included above css.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
DeepakSap14: I have put displaySize instead of size now.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
DeepakSap14: RTL is working. see screenshot.
RTL
LTR

libs/platform/src/lib/components/menu-button/menu-button.component.spec.ts
DeepakSap14: included test cases for type and disabled menu buttons.

DeepakSap14:
comments has been addressed, please review and approve.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
fkolar: it should not matter ! 
Have you look at the button implementation both on the core and platform side ?

Where did you see its array?

DeepakSap14:
@fkolar , Regarding menuOptions Array.
for button to have more than one icon, they have to pass icon choices as below.
<fdp-button [options]="['emphasized', 'menu']">Emphasized Button</fdp-button>
which i don't want in Menu button.

You can see in core button that they have a check of Array.
private _getOptionCssClass(options: string | ButtonOptions[]): string {
        if (Array.isArray(this.options)) {
            return this.options.map(option => `fd-button--${option}`).join(' ');
        }
        return `fd-button--${options}`
    }
here for multiple option icon to have, it needs to be Array.
Thus i am doing so.

libs/platform/src/lib/components/menu-button/menu-button.component.html
```diff
@@ -0,0 +1,13 @@
+<button
+    fd-button
+    [options]="menuOptions"
+    [compact]="displaySize == 'compact' ? true : false"
```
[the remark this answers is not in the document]
DeepakSap14: Now using displaySize === 'compact' only, it will work.

libs/platform/src/lib/components/menu-button/menu-button.component.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { Component, Input } from '@angular/core';
+import { MenuButtonComponent } from './menu-button.component';
+import { ButtonModule } from '@fundamental-ngx/core';
+
+describe('MenuButtonComponent', () => {
+    let component: MenuButtonComponent;
+    let fixture: ComponentFixture<MenuButtonComponent>;
+
+    beforeEach(async(() => {
```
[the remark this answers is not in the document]
DeepakSap14: I will be writing katalon tests for these functionalities. it will be duplicate effort.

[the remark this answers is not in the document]
DeepakSap14: will try quickly to create some tests.

DeepakSap14:
pushed changes as per fiori change for button. please review.

DeepakSap14:
@InnaAtanasova @JKMarkowski 
I have pushed changes which was breaking because of button changes. It is working fine now.
please review.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
```diff
@@ -0,0 +1,56 @@
+import { Component, Input, Output, EventEmitter } from '@angular/core';
+import { ButtonType } from '@fundamental-ngx/core';
+import { InputSize } from './../form/form-control';
+
+@Component({
+    selector: 'fdp-menu-button',
+    templateUrl: './menu-button.component.html',
+    styleUrls: ['./menu-button.component.scss'],
+})
+export class MenuButtonComponent {
```
fkolar, line 10: We have now base component that should have common properties already that this component can inherit so you would be able to remove Id (property start with lowercase btw.), displaySize, disabled

libs/platform/src/lib/components/menu-button/menu-button.component.html
```diff
@@ -0,0 +1,13 @@
+<button
+    fd-button
+    class="fd-button--menu"
+    [compact]="displaySize === 'compact'"
```
fkolar, line 4: Please see my comment for BaseComponent

libs/platform/src/lib/components/menu-button/menu-button.component.ts
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     *  Handles menu-button click
+     */
+    public buttonclick($event: any) {
+        event.stopPropagation();
+        this.click.emit();
+    }
+
+    /** @hidden */
+    setDisabledState(isDisabled: boolean): void {
```
fkolar, line 53: who is calling this method?

rengare:
@DeepakSap14 fix conflicts please ;)

DeepakSap14:
I have dependency on https://github.com/SAP/fundamental-ngx/pull/2369, once that PR is merged, i will make necessary changes in this PR.

DeepakSap14:
rebased with Base component.

apps/docs/src/app/platform/component-docs/platform-menu-button/platform-menu-button-examples/platform-menu-button-compact-examples.component.ts
```diff
@@ -0,0 +1,14 @@
...
+
+@Component({
+    selector: 'fdp-platform-menu-button-compact-example',
+    templateUrl: './platform-menu-button-compact-example.component.html',
+})
+export class PlatformMenuButtonCompactExampleComponent {
+    basicMenuData: any[] = [];
+    item = '';
+
+    onItemSelect(item: string) {
```
salarenko, line 11: Return value

apps/docs/src/app/platform/component-docs/platform-menu-button/platform-menu-button-examples/platform-menu-button-compact-examples.component.ts
```diff
@@ -0,0 +1,14 @@
+import { Component } from '@angular/core';
+
+@Component({
+    selector: 'fdp-platform-menu-button-compact-example',
+    templateUrl: './platform-menu-button-compact-example.component.html',
+})
+export class PlatformMenuButtonCompactExampleComponent {
+    basicMenuData: any[] = [];
```
salarenko, line 8: Decide if you use explicite typing and use one convention

apps/docs/src/app/platform/component-docs/platform-menu-button/platform-menu-button-examples/platform-menu-button-compact-examples.component.ts
DeepakSap14: resolved

apps/docs/src/app/platform/component-docs/platform-menu-button/platform-menu-button-examples/platform-menu-button-compact-examples.component.ts
DeepakSap14: resolved

stefanoScalzo:
I think instead of contentDensity we should have compact with a boolean

apps/docs/src/app/platform/component-docs/platform-menu-button/platform-menu-button-examples/platform-menu-button-compact-example.component.html
```diff
@@ -0,0 +1,143 @@
...
+<fdp-menu #lettuceMenu id="lettuce-menu">
+    <fdp-menu-item (itemSelect)="onItemSelect('Arugula')">Arugula</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Belgian Endive')">Belgian Endive</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Butter Lettuce')">Butter Lettuce</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Iceberg Lettuce')">Iceberg Lettuce</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Red Leaf Lettuce')">Red Leaf Lettuce</fdp-menu-item>
+    <fdp-menu-item (itemSelect)="onItemSelect('Romaine Lettuce')">Romaine Lettuce</fdp-menu-item>
+</fdp-menu>
+
+<div style="margin-top: 1rem;"><label>Item Selected:</label> {{ item }}</div>
```
stefanoScalzo, line 143: better to create a class

DeepakSap14:
> I think instead of contentDensity we should have compact with a boolean

In Platform library implementation, we have decided to have contentDensity. as in future there may be more options than "compact" and "cozy".

DeepakSap14:
@fkolar please review this.

DeepakSap14:
@salarenko @stefanoScalzo 
please check this PR. i have pushed changes requested.
if you find it all right, please approve it.

fkolar:
@DeepakSap14 can you do something with this PR? Either close it or rebase..

DeepakSap14:
All reviews already implemented.
Rebased with master. so merging it to master.

PR #2228, opened 2020-03-27
fix: replacing setters to ngonchanges

rengare:
Replaced setters in favour of ngOnChanges for building component CSS classes.

[Code-Contribution-Guideline](https://github.com/SAP/fundamental-ngx/wiki/Code-Contribution-Guideline#building-component-css-classes) has been updated

### DIFF
--- a/libs/core/src/lib/bar/bar.component.ts
+++ b/libs/core/src/lib/bar/bar.component.ts
@@ -1,4 +1,12 @@
-import { ChangeDetectionStrategy, Component, ElementRef, Input, OnChanges, OnInit, ViewEncapsulation } from '@angular/core';
+import {
+    ChangeDetectionStrategy,
+    Component,
+    ElementRef,
+    Input,
+    OnChanges,
+    OnInit,
+    ViewEncapsulation,
+} from '@angular/core';
 import { applyCssClass, CssClassBuilder } from '../utils/public_api';
 
 export type SizeType = '' | 's' | 'm_l' | 'xl';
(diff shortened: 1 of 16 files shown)


libs/core/src/lib/bar/bar.component.ts
```diff
@@ -17,95 +17,66 @@ export type BarDesignType = 'header' | 'subheader' | 'header-with-subheader' | '
...
-
-    /** @hidden */
-    elementRef(): ElementRef<any> {
+    public elementRef(): ElementRef<any> {
         return this._elementRef;
     }
 
     /** @hidden */
-    ngOnChanges(): void {
+    public ngOnChanges(): void {
```
InnaAtanasova: Aren't these public by default?

rengare: public access modifier removed

droshev:
@rengare Travis and netlify are failing.

mikerodonnell89:
The Tokenizer pattern requires us to use both fd-form-control and fd-tokenizer-input directives on the same input element.  I tried having both fd-form-control and fd-tokenizer-input directives implement CssClassBuilder, and when I do this, only the classes from one directive will be applied to the element, so I end up with `class="fd-tokenizer__input` when I want `class="fd-tokenizer__input fd-input fd-input--compact`.

rengare:
> The Tokenizer pattern requires us to use both fd-form-control and fd-tokenizer-input directives on the same input element. I tried having both fd-form-control and fd-tokenizer-input directives implement CssClassBuilder, and when I do this, only the classes from one directive will be applied to the element, so I end up with `class="fd-tokenizer__input` when I want `class="fd-tokenizer__input fd-input fd-input--compact`.

I'll check that

libs/core/src/lib/tabs/tab-utils/tab-directives.ts
```diff
@@ -81,23 +82,20 @@ export class TabIconDirective implements CssClassBuilder, OnInit {
...
-    ) {}
+    ) { }
 
     /** @hidden
      * Function runs when component is initialized
      * function should build component css class
      * function should build css style
      */
-    ngOnInit(): void {
+    public ngOnChanges(): void {
```
InnaAtanasova, line 98: Don't hate me :D You need to remove the `public` :)

libs/core/src/lib/tabs/tab-utils/tab-directives.ts
```diff
@@ -106,18 +104,18 @@ export class TabIconDirective implements CssClassBuilder, OnInit {
      * function must return single string
      * function is responsible for order which css classes are applied
      */
-    buildComponentCssClass(): string {
+    public buildComponentCssClass(): string {
```
InnaAtanasova, line 107: `public`

libs/core/src/lib/tabs/tab-utils/tab-directives.ts
```diff
@@ -106,18 +104,18 @@ export class TabIconDirective implements CssClassBuilder, OnInit {
...
+            this.icon ? `sap-icon--${this.icon}` : '',
+            this.class
         ].filter(x => x !== '').join(' ');
     }
 
     /** HasElementRef interface implementation
      * function used by applyCssClass and applyCssStyle decorators
      */
-    elementRef(): ElementRef<any> {
+    public elementRef(): ElementRef<any> {
```
InnaAtanasova, line 118: `public`

InnaAtanasova:
> Can we also use this approach and remove the AbstractFdNgxClass?

I think this is outside of the scope of this PR. This PR was supposed to get rid of the private Input pros and setters/getters we recently introduced. We agreed that once we start refactoring a component we will migrate the `AbstractFdNgxClass` to this approach. At some point we can create an issue for this and go over all components and change the ones that are not migrated.

InnaAtanasova:
@mikerodonnell89 , could you take a look at the PR again and add a new review? Thanks

libs/core/src/lib/tabs/tab-utils/tab-directives.ts
rengare: :) good point, I'll do that

libs/core/src/lib/tabs/tab-utils/tab-directives.ts
rengare: removed

libs/core/src/lib/tabs/tab-utils/tab-directives.ts
rengare: removed

libs/core/src/lib/tabs/tab-utils/tab-directives.ts
rengare: removed

libs/core/src/lib/tabs/tab-utils/tab-directives.ts
```diff
@@ -126,70 +125,61 @@ export class TabIconDirective implements CssClassBuilder, OnInit {
     // tslint:disable-next-line:directive-selector
     selector: '[fd-tab-tag]',
     host: {
-        'class': 'fd-tabs__tag'
-    }
+        class: 'fd-tabs__tag',
+    },
```
JKMarkowski, line 129: IT's really detail, but I see that in `tab-directives` there are a lot of hanging commas. Could you remove them ?

mikerodonnell89:
This plus https://github.com/SAP/fundamental-ngx/pull/2249 looks great

libs/core/src/lib/tabs/tab-utils/tab-directives.ts
rengare: removed

PR #2279, opened 2020-04-07
fix: side navigation selected state on parent/child

JKMarkowski:
Example of issue is well shown there
https://stackblitz.com/edit/angular-side-navigation
Before:
The parent element can't have selected state, unless child doesn't have it.

After:
sidenav2

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


InnaAtanasova:
1. The Utility Section (bottom section) has wrong paddings in the "Side navigation with Programmatically State Change." example
!!! The same title has a `.` at the end that should be removed. Please check with @Vanessa-Cusmich  but I believe each word in the title should start with a capital letter.

Screen Shot 2020-04-08 at 11 21 20 AM

2. The focus in the "Side Navigation Condensed Mode Generated from Configuration Object" example is a bit off. I checked styles and the "overflow" problem with selected item is also reproducible there so we can create a ticket there but the rest of the issues are on ngx side, also the focus color is different from that in styles.
Screen Shot 2020-04-08 at 11 22 47 AM

Screen Shot 2020-04-08 at 11 22 52 AM
Screen Shot 2020-04-08 at 11 23 00 AM
Screen Shot 2020-04-08 at 11 23 09 AM

Issue #2292, opened 2020-04-08
fix: (Platform) Checkbox Component created in Platform


DeepakSap14:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
Created Checkbox component in platform, integrated with form control of platform.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 9

#### If this is a bug, please provide steps for reproducing it.
No Bug

#### Is there anything else we should know?
Created Checkbox component in platform

PR #2293, opened 2020-04-08
fix: (platform) Introduce Checkbox as a Component

DeepakSap14:
https://github.com/SAP/fundamental-ngx/issues/2292

Created Checkbox component in platform, which works with platform FormFieldControl.
 It works fine with Reactive form as well as template driven form.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

Technical specification:
https://github.com/SAP/fundamental-ngx/wiki/Platform:-Checkbox-Component-Technical-Design


fkolar:
Build failed

DeepakSap14:
> Build failed

Build is passed, but deploy is failing. there may be some issue with netify. will check with Deno and team.
I have also checked it in local with "npm run build-all", it was passing.

DeepakSap14:
Deployment failing is happening for other PR as well. once reviewed will check on that issue.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /** Stores current checkbox value. */
+    public checkboxValue: any;
+
+    /** Stores current checkbox state. */
+    private checkboxState: 'checked' | 'unchecked' | 'indeterminate' | 'force-checked';
+
+    /** Values returned by control. */
+    public values: FdCheckboxValues = { trueValue: true, falseValue: false, thirdStateValue: null };
```
fkolar, line 68: I am not sure about these values ..

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,140 @@
...
+import { Status } from './../form-control';
+
+let uniqueId = 0;
+
+@Component({
+    selector: 'fdp-checkbox',
+    templateUrl: './checkbox.component.html',
+})
+export class CheckboxComponent extends BaseInput {
+    /** Sets text of control label. */
```
fkolar, line 14: How are you going to handle complex objects just image you will have:

Object payment terms

terms: PaymentTerms = [
   {id: 1, name: "2 weeks", description: "xxxx"},
   {id: 2, name: "3 weeks", description: "xxxx"},
   {id: 2, name: "4 weeks", description: "xxxx"}
]

` <fdp-checkbox [list]=terms....`

I guess for primitive string it already works. 


and when users check, I want to receive back to the application all the terms that you selected

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-checkbox-styling.component.ts
```diff
@@ -0,0 +1,30 @@
+import { FormGroup, FormControl, Validators } from '@angular/forms';
+import { Component, DoCheck } from '@angular/core';
+
+@Component({
+    selector: 'fdp-checkbox-style',
+    templateUrl: 'platform-checkbox-styling.component.html',
+})
+export class PlatformChekboxStyleComponent implements DoCheck {
+    values = { trueValue: 'Yes', falseValue: 'No', thirdStateValue: 'I dont have an opinion' };
```
fkolar, line 9: I dont think on application level anybody will work with CV like this. please see my comment for the component ts file

DeepakSap14:
Have pushed new changes based on reviews.
This is platform checkbox implementation only (not checkbox group), there will be a separate PR for checkbox group.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-checkbox-styling.component.html
```diff
@@ -0,0 +1,35 @@
...
+            [id]="'checkbox-10'"
+            [name]="'checkbox-10'"
+            [label]="'Mandatory checkbox'"
+            [binary]="true"
+            [size]="'compact'"
+            formControlName="example1"
+        ></fdp-checkbox>
+    </fieldset>
+</form>
+<div>Value: {{ customForm.get('example1').value }}</div>
```
rengare, line 15: Please avoid using methods inside templates because it leads to performance issues. (methods being called every angular check cycle)

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-checkbox-styling.component.ts
```diff
@@ -0,0 +1,27 @@
...
+    templateUrl: 'platform-checkbox-styling.component.html',
+})
+export class PlatformChekboxStyleComponent implements DoCheck {
+    customForm = new FormGroup({
+        example1: new FormControl({ value: '', disabled: false }),
+        example2: new FormControl({ value: '', disabled: false }, Validators.required),
+    });
+
+    ngDoCheck() {
+        if (this.customForm.get('example1').value !== true) {
```
rengare, line 15: Store reference to object in varible to avoid unecessary method calls, like:
```let exampleRef = this.customForm.get('example1');

exampleRef.value !== true
...
```

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    @Input()
+    tristateSelectable: boolean = true;
+
+    /** value for checkbox, when it is not binary */
+    @Input()
+    get value(): any {
+        return super.getValue();
+    }
+    set value(value: any) {
+        if (value) {
```
fkolar, line 40: check parent implementation what it is doing, when setting a value, I need to emit event to parent formField so form field can call markForCheck. 

there is `stateChanges`  - please check it out. 

Also what is the added value set your value here instead of setValue from parent?

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,104 @@
...
+            this.ngControl.valueAccessor = this;
+        }
+
+        // @hidden have to set default initial values as base class has check and throws error
+        this.id = `fdp-checkbox-id-${uniqueId++}`;
+        this.name = `fdp-checkbox-id-${uniqueId++}`;
+    }
+
+    /** ControlvalueAccessor */
+    writeValue(value: any) {
```
fkolar, line 79: check what parent writeValue is doing and why

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+    set value(value: any) {
+        if (value) {
+            this._value = value;
+        }
+    }
+
+    /** set status value , will be coming from checkbox group*/
+    set stateStatus(newStatus: Status) {
+        this._status = newStatus;
```
fkolar, line 47: the same when you set somethign and you do call markforcheck you need to inform parent. formField. also we should try to limit markForCheck as this is called by parent

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    writeValue(value: any) {
+        this.modelValue = value;
+        this.checkboxValue = value;
+        this._changeDetector.markForCheck();
+    }
+
+    /** child checkbox is initialized here */
+    ngAfterViewInit() {
+        if (!this.binary) {
+            // updating checkbox values property for this custom checkbox
```
fkolar, line 88: you are inheriting from parent and parent is doing something. Dont disable it just like that. you should be find calling parent method

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,104 @@
...
+        @Optional() @Self() public ngForm: NgForm
+    ) {
+        super(_changeDetector, ngControl, ngForm);
+
+        if (this.ngControl) {
+            this.ngControl.valueAccessor = this;
+        }
+
+        // @hidden have to set default initial values as base class has check and throws error
+        this.id = `fdp-checkbox-id-${uniqueId++}`;
```
fkolar, line 74: let's remove this one, this is why we have it in parent. and the goal is to let app developer set it. and if not then have some default value, but we should not do this. this is really a fallback in case you dont sent it. I am seeing this more and more and we should remove this code from our compoennts

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,104 @@
...
+        this._cd.markForCheck();
+    }
+
+    /** returns form status */
+    get stateStatus(): Status {
+        return this._status;
+    }
+
+    /** @hidden captures controller value */
+    private modelValue: any;
```
fkolar, line 57: why do you need this extra property, why cant you set it directly? to value ?

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    /** tracking checkbox current value */
+    checkboxValue: any;
```
fkolar, line 60: another one, this is too complicated.  The checkbox is capable of setting value directly.

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
```diff
@@ -0,0 +1,14 @@
+<fd-checkbox
```
fkolar, line 1: Also I see only checkbox here but how this is going to work with FormField and FormGroup?  It needs to be part of it, so in formField you set label for the field ..

DeepakSap14:
> What are the breaking changes?

In platform we need checkbox somethink like "https://primefaces.org/primeng/showcase/#/checkbox".
This will be easy to integrate with MetaUi Implementation. current implementation in checkbox expects to have values {"truevalue: '', falseValue: '', tristatevalue:''} passed. but with this implementation, application developer have to just pass value='checkboxvalueonselect', and on select it will update controller. on uncheck, it will be removed from controller.

droshev:
> > What are the breaking changes?
> 
> @droshev
> In platform we need checkbox something like "https://primefaces.org/primeng/showcase/#/checkbox".
> This will be easy to integrate with MetaUi Implementation. current implementation in checkbox expects to have values {"truevalue: '', falseValue: '', tristatevalue:''} passed. but with this implementation, application developer have to just pass value='checkboxvalueonselect', and on select it will update controller. on uncheck, it will be removed from controller.

My comment was related to the changes you did with this PR. Do you introduce a breaking changes with this PR? If not then you need to update the title of the PR.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: calling super. ngAfterViewInit() now.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: This is needed to notify corecheckbox for current checkbox value. based on this value checkbox state changes.

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
```diff
@@ -0,0 +1,18 @@
+<div fd-form-item>
+    <fd-checkbox
+        [inputId]="id"
+        [name]="name"
+        [disabled]="disabled"
+        [label]="label"
```
fkolar, line 6: We need to consult this with somebody, if adding aria tags to this level will really work, as fd-checkbox has yet another layer under this

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
```diff
@@ -0,0 +1,18 @@
...
+        [attr.aria-label]="ariaLabel || null"
+        [attr.aria-labelledby]="ariaLabelledby"
+        [tristate]="tristate"
+        [tristateSelectable]="tristateSelectable"
+        [state]="stateStatus"
```
fkolar, line 11: For the state please follow similar path like other component, you dont need extra setter/getter.  This where we have in base input method that checks error and just mark component if in error state or not. 


If we want to support in general all statuses (even i dont see usecase why we would show info status, or success - maybe to mark some questionaries "yes you did good", "all is wrong", "we are not sure"), we better move state to the parent class. 

state needs to be driven by FormAPI.

DeepakSap14: i was thinking checkbox group may require some way to update status of individual checkbox. so while implementing, this setStatus will be used. have removed this. Now Directly getting state from formApi status.

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
DeepakSap14: yes these aria tag won't be transmitted to native "input type=checkbox". because core-checkbox is missing this handling. i can check with fundamentals-team if aria-tag can be included. for now if you want i can remove.

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
fkolar: I see. Can you find out from @manjunathanagaraj  , when we have such usecase where will have checkboxgroup if we will mark individual checkbox? I am not sure, then you need to provide a message. 

If we are talkg about isBinary, yes/no, then it make sense but again it should follow regular form/error handling logic like we have it. What do you think ?

DeepakSap14: Now I have kept it where state is read from form Api Status.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-checkbox-styling.component.html
```diff
@@ -0,0 +1,32 @@
...
+        <h2>Mandatory Checkbox with Reactive form and Error Status Style</h2>
+        <span>Checkbox is Mandatory to check or it should show Error on submit</span>
+        <fdp-checkbox
+            [id]="'checkbox-020'"
+            [name]="'checkbox-020'"
+            [contentDensity]="'compact'"
+            [isBinary]="true"
+            formControlName="example2"
+        >
+            Do you Agrre with <a href="#"> Terms and Conditions</a>?
```
stefanoScalzo, line 25: typo

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-checkbox-example.component.html
```diff
@@ -0,0 +1,301 @@
...
+    </fieldset>
+
+    <fieldset fd-fieldset>
+        <span>Example 10: Binary Checkbox without default value</span>
+        <fdp-checkbox
+            [id]="'checkbox-13'"
+            [name]="'checkbox-13'"
+            [label]="'Checkbox13'"
+            [isBinary]="true"
+            [contentDensity]="'compact'"
```
stefanoScalzo, line 170: this should have the same approach as disabled and all other compacts with a boolean

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
stefanoScalzo: I think this is fine @mikerodonnell89 what do you think?

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-checkbox-example.component.html
DeepakSap14: In Platform library implementation, we have decided to have contentDensity. as in future there may be more options than "compact" and "cozy".

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-checkbox-styling.component.html
DeepakSap14: resolved

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
DeepakSap14: @stefanoScalzo @mikerodonnell89 , any input on this?

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,196 @@
...
+    tristateSelectable: boolean = true;
+
+    /** value for checkbox, when it is not binary */
+    @Input()
+    get value(): any {
+        return super.getValue();
+    }
+    set value(selectValue: any) {
+        this._value = selectValue;
+        this._cd.markForCheck();
```
fkolar, line 62: What allot of issue this is causing, so let;s fix those issues? 
What at least we need to do is compare if the _value is different than selectValue, only then set the value and fire markForCheck

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,196 @@
...
+        // calling super.setValue() causing lots of issues. so setting value here.
+        // rest functionality imported from baseinput
+    }
+
+    /** due to an issue in Angular bind the formControl instance instead of using formControlName.
+     *  https://github.com/angular/angular/issues/17685
+     */
+    @Input()
+    formControl: FormControl;
```
fkolar, line 71: Why do we need this ? You dont want to have this as Input. 
First formControl internal field should be already in baseInput
Second we dont need to explicitly add this as INPUT. This is provided by Angular

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
fkolar: Ok. Let's pass this. I think we can use single _value

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,196 @@
...
+    ) {
+        super(_changeDetector, ngControl, ngForm);
+    }
+
+    /** ControlvalueAccessor */
+    writeValue(value: any): void {
+        if (value !== null) {
+            // value of checkbox will have type string, formcontrol is expected as array
+            if (typeof value !== 'string') {
+                this.checkboxValue = value;
```
fkolar, line 96: You dont want to have all this logic inside writeValue, it should be broken down to smaller pieces,

DeepakSap14: Done

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: Done

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: using setValue now. issues handled.
Done

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,196 @@
...
+    constructor(
+        protected _changeDetector: ChangeDetectorRef,
+        @Optional() @Self() public ngControl: NgControl,
+        @Optional() @Self() public ngForm: NgForm
+    ) {
+        super(_changeDetector, ngControl, ngForm);
+    }
+
+    /** ControlvalueAccessor */
+    writeValue(value: any): void {
```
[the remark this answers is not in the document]
DeepakSap14: Done.

stefanoScalzo:
Build is failing

DeepakSap14:
> Build is failing

Failing on core checkbox test. this change is not related to this PR.

libs/platform/src/lib/components/form/radio-group/radio-group.component.spec.ts
```diff
@@ -151,7 +151,7 @@ class TestComponentList {
     radioButtons: RadioButtonComponent;
 }
 
-describe('GroupRadioButtonComponent', () => {
+xdescribe('GroupRadioButtonComponent', () => {
```
KevinOkamoto, line 154: Is there a reason for excluding this test suite with `xdescribe`?

DeepakSap14: This test case behaving flacky. it is failing in local some times.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,226 @@
...
+        if (value) {
```
JKMarkowski, line 106: What if someone changes value fo `false` from outside? It won't be triggered

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
JKMarkowski: We should have some input with arias on `fundamental-ngx`

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
JKMarkowski: Check if you can change value to `false` from outside (programatically from docs)

libs/platform/src/lib/components/form/radio-group/radio-group.component.spec.ts
KevinOkamoto: Does this "flaky" test suite block the PR from being merged? Since this file is not part of Checkbox Component, it's a little odd having this code change in the PR.

Commenting out, or rather skipping a unit test suite which will be merged into master is bad practice. We should probably create an issue and try and fix the unit test(s) in another PR.

As for now, if this test suite is blocking this PR from being "mergeable" then I guess it's OK to skip it, but definitely create an issue so we don't forget it. If it's not blocking this PR then I would revert it back.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,226 @@
...
+        if (!this.isBinary) {
+            // updating checkbox values property for this custom checkbox
+            this.corecheckbox.values.trueValue = this.value;
+            this.corecheckbox.values.falseValue = undefined;
+
+            // set core checkbox control value based on platform checkbox control value
+            if (this.value && this.multiSelectModel && this.multiSelectModel.includes(this.value)) {
+                this.checkboxValue = this.value;
+            } else {
+                if (this.tristate && this.multiSelectModel.includes(null)) {
```
KevinOkamoto, line 123: I'm unsure about this if statement, it implies that we have a tristate checkbox(es) as part of a multi select group. To me "tristate" and "multi-select" are mutually exclusive - when you are checking an item from a multi select group, that item is either "in" or "out" of the select group, it can't be in a third state. This if statement kind of illustrates that, since if you have multiple tristate checkboxes in a multi select group, they may all have the same "null" value.

Can anyone provide an example where we need a tristate checkbox in a multi select?

DeepakSap14:
pushing changes after multiple reviews today. it will resolve branch conflict as well.

droshev:
> > Build is failing
> 
> Failing on core checkbox test. this change is not related to this PR.

@DeepakSap14 we cannot allow merging PRs if there are tests failing even if you think your PR is not related to those tests.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,295 @@
...
+     */
+    private multiSelectModel: any;
+
+    /**
+     * @hidden checkbox state, used when checkbox is used without form.
+     */
+    private _checked: boolean = false;
+
+    /** @hidden value of checkbox */
+    private _checkboxValue: any;
```
fkolar, line 119: we have discussed also this extra property to save the value. do we need it? Why we cannot directly save this to value we have already there. 

To save the selection we have this multiselectModel

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,295 @@
...
+    /* @hidden
+     * reference of child checkbox implementation
+     */
+    @ViewChild(CoreCheckboxComponent)
+    private corecheckbox: CoreCheckboxComponent;
+
+    /* @hidden
+     * stores  formControl values
+     */
+    private multiSelectModel: any;
```
fkolar, line 111: just call it `model`

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,295 @@
...
+            this.multiSelectModel = [this.checkboxCurrentValue];
+        }
+    }
+
+    /**
+     * @hidden
+     * @param value , Array or boolean and string/null for tristate
+     * setting core checkbox control value
+     */
+    private _setCoreCheckboxControl(value: any): void {
```
fkolar, line 265: Not sure what is this method doing. The name of method is confusing..

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
```diff
@@ -0,0 +1,16 @@
+<div fd-form-item>
+    <fd-checkbox
+        [inputId]="id"
```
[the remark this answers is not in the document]
DeepakSap14: Solution:
	option 1: can introduce accessibility in core checkbox. and then platform checkbox 
                         accessibility will work.
	option 2: remove from platform checkbox, until it is present in core checkbox.
which option we should apply for now??

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,295 @@
...
+
+    /**
+     * Users can specify the `aria-labelledby` attribute which will be forwarded to the input element
+     */
+    @Input('aria-labelledby')
+    ariaLabelledby: string | null = null;
+
+    /** indeterminate state */
+    @Input()
+    tristate: boolean = false;
```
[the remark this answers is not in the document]
DeepakSap14: Solution:
	option 1: have got confirmation from design team on use of tri-state checkbox. forwarding the same email again to you.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: solution:
	option 1: saving control value to this._value(in baseInput) by using super.writeValue(value); this._value is different from checkbox value(value when checked), so i have used an extra variable.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: solution:
	1.renamed to model

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: solution: 
	1. this function is initialising checkbox with value passed from controller.
	2. renamed to _initialiseCheckboxWithControl

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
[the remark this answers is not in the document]
fkolar: When you say Option 2, Where do we have it in platform I dont see it. 

We need to have in there at least 
```
[attr.aria-checked]="..."
           [attr.aria-label]="ariaLabel || null"
           [attr.aria-labelledby]="ariaLabelledby"
```
Let's do it in platform and raise issue against core, once its in core we will update this component

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: have used a different variable to store checkbox value.
setValue() and getValue() is used and sets control value on checkbox change and while writeValue().

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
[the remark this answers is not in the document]
DeepakSap14: this has been removed from here. until implemented in core checkbox.

DeepakSap14:
@fkolar  All the changes has been made. please review this PR and approve if everything is fine.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-header/platform-checkbox-header.component.html
```diff
@@ -0,0 +1,30 @@
+<header>Checkbox</header>
+<description>
+    Platform Checkbox is wrapper implemention on
```
fkolar, line 3: I was under impression that we have new docs that we try to polish in gitbook but I can see old  text here.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-multiselect-checkbox.component.html
```diff
@@ -0,0 +1,167 @@
+<h3>Template Driven Form</h3>
+<p>Checkbox can be pre-selected if value of checkbox matches any value in ngModel directive.</p>
+<fdp-form-group [multiLayout]="true">
+    <fdp-form-field [id]="'sirius'" zone="zLeft" rank="1">
+        <fdp-checkbox [name]="'sirius'" [label]="'Sirius'" [checkboxValue]="'sirius'" [(ngModel)]="field1"></fdp-checkbox>
```
fkolar, line 5: The idea behind formGroup/formField is that you should not repeat attributes which are set or should be set on the formField level. 

name = the same as ID or name if you set on formField Level

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-multiselect-checkbox.component.html
```diff
@@ -0,0 +1,167 @@
...
+    <fdp-form-field [id]="'vega'" zone="zRight" rank="1">
+        <fdp-checkbox [name]="'vega'" [label]="'Vega'" [checkboxValue]="'vega'" [(ngModel)]="field2"></fdp-checkbox>
+    </fdp-form-field>
+</fdp-form-group>
+<div style="padding-left: 2rem;">field1: {{ field1 }}</div>
+<div style="padding-left: 2rem;">field2: {{ field2 }}</div>
+<br />
+<br />
+
+<p>Multiple checkboxes connected with same ngModel directive. Multiple checkbox can be selected.</p>
```
fkolar, line 17: this example is a bit miss-leading. Usually when try have CheckboxGroup, those checkboxes should have identical name. and most likely we will have this covered with CheckboxGroup. This will be platform Solution. To group multiple checkboxes we should have reference here to different component.

libs/platform/src/lib/components/form/checkbox/checkbox.module.ts
```diff
@@ -0,0 +1,12 @@
+import { FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { CheckboxModule, FormModule as FdFormModule } from '@fundamental-ngx/core';
+import { CheckboxComponent } from './checkbox.component';
+
+@NgModule({
+    declarations: [CheckboxComponent],
+    imports: [CommonModule, FormsModule, ReactiveFormsModule, FdFormModule, CheckboxModule],
```
fkolar, line 9: We should avoid using FormModule from core. This is big problem. Even using checkbox FormModule is importing all the other modules which we dont need. Its the same as you would import our big PlatformModule that has everything.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-multiselect-checkbox.component.html
DeepakSap14: reference here to different component :-- this i did not understand. please elaborate.

libs/platform/src/lib/components/form/checkbox/checkbox.module.ts
DeepakSap14: removed unused ReactiveFormModule and core FormModule.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-multiselect-checkbox.component.html
DeepakSap14: moved common attribute to fdp-form-field. changed id and name value non-matching.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-header/platform-checkbox-header.component.html
DeepakSap14: have reviewed documentation with Manju and updated it now.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-binary-checkbox-no-form.component.html
```diff
@@ -0,0 +1,16 @@
...
+<fdp-checkbox [name]="'checkbox1'" [isBinary]="true" [label]="'checkbox1'" [value]="'Yes'" [(checked)]="checkbox1">
+</fdp-checkbox>
+<div>value: {{ checkbox1 }}</div>
+
+<fdp-checkbox [name]="'checkbox2'" [isBinary]="true" [label]="'checkbox2'" [value]="'Yes'" [(checked)]="checkbox2">
+</fdp-checkbox>
+<div>value: {{ checkbox2 }}</div>
+<br />
+<br />
+<h4>Binary Checkbox without value</h4>
```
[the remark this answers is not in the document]
DeepakSap14: have made all suggested changes.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-multiselect-checkbox.component.html
```diff
@@ -0,0 +1,177 @@
+<h3>Template Driven form</h3>
```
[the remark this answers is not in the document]
DeepakSap14: have made all suggested changes.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: have changed the logic for tristate after discussion.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: removed writevalue if(value), because it will block control value=false.

libs/platform/src/lib/components/form/radio-group/radio-group.component.spec.ts
DeepakSap14: have removed these unwanted changes.

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
DeepakSap14: taking state value as form status.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-multiselect-checkbox.component.html
DeepakSap14: have removed multiple checkbox example. will be covered in Checkbox group.
Have mentioned this in documentation.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: implementation is changed now. now only checked value is expected from user.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: implementation is changed now. now only checked value is expected from user. this case will be handled in checkbox group, but checkbox will be able to set these properties now.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-checkbox-styling.component.ts
DeepakSap14: implementation is changed now. now only checked value is expected from user.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-checkbox-styling.component.html
DeepakSap14: replaced with customForm.getRawValue() | json.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-checkbox-styling.component.ts
DeepakSap14: using reference now to avoid multiple calls.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: This has been removed. default id is getting set from baseInput.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: now status is getting fetched from FormFieldControl.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: now using set value method to call setValue() of baseInput.

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
DeepakSap14: have included <div fd-form-item> in template.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,280 @@
...
+@Component({
+    selector: 'fdp-checkbox',
+    templateUrl: './checkbox.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    encapsulation: ViewEncapsulation.None,
+    providers: [{ provide: FormFieldControl, useExisting: CheckboxComponent, multi: true }]
+})
+export class CheckboxComponent extends BaseInput implements AfterViewInit {
+    /** set to true if binary checkbox */
+    @Input()
```
fkolar, line 34: I am more or less ok with this version I would work on more on the Selection Model and remove all the code and replace it with Selection Model that exists in the Material/cdk. it will make things much cleaner

stefanoScalzo:
After clicking checkbox quickly the page generates errors out

DeepakSap14:
> After clicking checkbox quickly the page generates errors out

This issue is coming from core checkbox. Will raise issue for fix in core checkbox.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-checkbox-error-handling.component.html
```diff
@@ -0,0 +1,45 @@
...
+    </fdp-form-field>
+
+    <ng-template #i18n let-errors>
+        <span *ngIf="errors.required">
+            Value is required
+        </span>
+        <span *ngIf="errors.invalid">
+            Value is invalid
+        </span>
+    </ng-template>
```
JKMarkowski, line 41: Example with errors inside message works weird, could you change trigger events to [focus ,blur] inside of click?

JKMarkowski:
> > After clicking checkbox quickly the page generates errors out
> 
> This issue is coming from core checkbox. Will raise issue for fix in core checkbox.

This issue was reported some time ago. It's not even ngx problem, but chrome&styles issue.

apps/docs/src/app/platform/component-docs/platform-forms/checkbox/platform-checkbox-examples/platform-checkbox-error-handling.component.html
DeepakSap14: have to work separately on form group and form field to resolve trigger issue.

DeepakSap14:
Approvals are done. It looks Good for me. Merging it.

Issue #2277, opened 2020-04-06
Platform: Menu styles are broken after upgrade to fundamental-styles 0.8.x


KevinOkamoto:
#### Is this a bug, enhancement, or feature request?

Bug.

#### Briefly describe your proposal.

After the upgrade to Fundamental-Styles 0.8, the menu lists are not displaying properly.

<img width="650" alt="Screen Shot 2020-04-06 at 10 22 20 AM" src="https://user-images.githubusercontent.com/48103491/78586636-9098a280-77f0-11ea-828f-9171df6fa1aa.png">
<img width="604" alt="Screen Shot 2020-04-06 at 10 22 32 AM" src="https://user-images.githubusercontent.com/48103491/78586656-968e8380-77f0-11ea-811b-d8bb9e38657a.png">

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 9
Fundamental NGX 0.16.0

#### If this is a bug, please provide steps for reproducing it.

* Goto documentation page for Platform:Menu
* Click on a button to open menu

KevinOkamoto:
Fixed by #2294

PR #2294, opened 2020-04-08
feat: (platform) Fixing Menu/SearchField styles after fundamental-style upgrade

KevinOkamoto:
#2277 

The architecture of the Menu CSS styles in Fundamental-Styles 0.8.x has changed from previous versions, and as a result, has broken the display of the platform Menu and SearchField components.

This PR makes fixes to the Menu and SearchField components to account for the Menu style changes.

**Menu BEFORE:**
Screen Shot 2020-04-08 at 9 08 28 AM

**Menu AFTER:**
Screen Shot 2020-04-08 at 9 05 36 AM

**SearchField BEFORE:**
Screen Shot 2020-04-08 at 9 08 15 AM

**SearchField AFTER:**
Screen Shot 2020-04-08 at 9 06 05 AM

### DIFF
--- a/libs/platform/src/lib/components/menu/menu-item.component.ts
+++ b/libs/platform/src/lib/components/menu/menu-item.component.ts
@@ -27,7 +27,7 @@ export class MenuItemComponent implements OnDestroy, FocusableOption {
     // Track when menu item is hovered over
     public hovered: Subject<MenuItemComponent> = new Subject<MenuItemComponent>();
 
-    constructor(private elementRef: ElementRef) {}
+    constructor(private elementRef: ElementRef) { }
 
     // Add Fundamental-Styles class for menu item
     @HostBinding('class.fd-menu__item') menuItemClass = true;
@@ -40,7 +40,7 @@ export class MenuItemComponent implements OnDestroy, FocusableOption {
         return this.cascadeDirection === 'left';
     }
 
-    @HostBinding('attr.role') role = 'listitem';
+    @HostBinding('attr.role') role = 'menuitem';
     @HostBinding('attr.tabindex') tabindex = '-1';
 
     // Handle selection of item via keyboard 'Enter' or mouseclick
--- a/libs/platform/src/lib/components/search-field/search-field.component.html
+++ b/libs/platform/src/lib/components/search-field/search-field.component.html
@@ -1,106 +1,63 @@
-<div
-    class="search-field"
-    [ngClass]="{
+<div class="search-field" [ngClass]="{
         'with-categories': showCategoryDropdown,
         'is-loading': isLoading,
         'hide-category-label': hideCategoryLabel
-    }"
->
-    <div
-        class="search-field--input-group fd-input-group"
-        [attr.aria-controls]="menuId"
-        [attr.aria-expanded]="showDropdown ? 'true' : 'false'"
-        aria-haspopup="true"
-        #inputGroup
-    >
-        <span
-            class="fd-input-group__addon fd-input-group__addon--button"
-            [ngClass]="{ 'fd-input-group__addon--compact': compact }"
-            *ngIf="showCategoryDropdown"
-        >
-            <button
-                class="search-field--category-button fd-input-group__button fd-button fd-button--menu"
-                [ngClass]="{ 'fd-button--compact': compact }"
-                [fdpMenuTriggerFor]="categoryMenu"
-            >
-                <span class="search-field--category-label" *ngIf="!hideCategoryLabel">{{
+    }">
+    <div #inputGroup class="fdp-search-field__input-group fd-input-group" [attr.aria-controls]="menuId"
+        [attr.aria-expanded]="showDropdown ? 'true' : 'false'" aria-haspopup="true">
+        <span class="fd-input-group__addon fd-input-group__addon--button"
+            [ngClass]="{ 'fd-input-group__addon--compact': contentDensity === 'compact' }" *ngIf="showCategoryDropdown">
+            <button class="fdp-search-field__category-button fd-input-group__button fd-button fd-button--menu"
+                [ngClass]="{ 'fd-button--compact': contentDensity === 'compact' }" [fdpMenuTriggerFor]="categoryMenu">
+                <span class="fdp-search-field__category-label" *ngIf="!hideCategoryLabel">{{
                     currentCategory ? currentCategory.label : categoryLabel
                 }}</span>
             </button>
         </span>
-        <input
-            #inputField
-            type="text"
-            placeholder="{{ placeholder }}"
-            class="search-field--input fd-input fd-input-group__input"
-            [attr.id]="inputId"
-            [attr.disabled]="disabled ? '' : null"
-            [ngClass]="{ 'fd-input--compact': compact }"
-            (keydown)="onKeydown($event)"
-            (keydown.enter)="onSearchSubmit()"
-            [(ngModel)]="inputText"
-            (ngModelChange)="onValueChange($event)"
-        />
-        <span
-            class="fd-input-group__addon fd-input-group__addon--button"
-            [ngClass]="{ 'fd-input-group__addon--compact': compact }"
-            *ngIf="inputText && inputText.length > 0"
-        >
+        <input #inputField type="search" [attr.placeholder]="placeholder"
+            class="fdp-search-field__input fd-input fd-input-group__input" [attr.id]="inputId"
+            [attr.disabled]="disabled ? '' : null" [ngClass]="{ 'fd-input--compact': contentDensity === 'compact' }"
+            (keydown)="onKeydown($event)" (keydown.enter)="onSearchSubmit()" [(ngModel)]="inputText"
+            (ngModelChange)="onValueChange($event)" />
+        <span class="fd-input-group__addon fd-input-group__addon--button"
+            [ngClass]="{ 'fd-input-group__addon--compact': contentDensity === 'compact' }"
+            *ngIf="inputText && inputText.length > 0">
             <button
-                class="search-field--clear fd-input-group__button fd-button fd-button--transparent sap-icon--decline"
-                [ngClass]="{ 'fd-button--compact': compact }"
-                (click)="clearTextInput()"
-            ></button>
+                class="fdp-search-field__clear fd-input-group__button fd-button fd-button--transparent sap-icon--decline"
+                [ngClass]="{ 'fd-button--compact': contentDensity === 'compact' }" (click)="clearTextInput()"></button>
         </span>
-        <span
-            class="fd-input-group__addon fd-input-group__addon--button"
-            [ngClass]="{ 'fd-input-group__addon--compact': compact }"
-            *ngIf="!isLoading"
-        >
+        <span class="fd-input-group__addon fd-input-group__addon--button"
+            [ngClass]="{ 'fd-input-group__addon--compact': contentDensity === 'compact' }" *ngIf="!isLoading">
             <button
-                class="search-field--submit fd-input-group__button fd-button fd-button--transparent sap-icon--search"
-                [attr.id]="submitId"
-                [attr.disabled]="disabled ? '' : null"
-                [ngClass]="{ 'fd-button--compact': compact }"
-                (click)="onSearchSubmit()"
-            ></button>
+                class="fdp-search-field__submit fd-input-group__button fd-button fd-button--transparent sap-icon--search"
+                [attr.id]="submitId" [attr.disabled]="disabled ? '' : null"
+                [ngClass]="{ 'fd-button--compact': contentDensity === 'compact' }" (click)="onSearchSubmit()"></button>
         </span>
-        <span
-            class="fd-input-group__addon fd-input-group__addon--button"
-            [ngClass]="{ 'fd-input-group__addon--compact': compact }"
-            *ngIf="isLoading"
-        >
+        <span class="fd-input-group__addon fd-input-group__addon--button"
+            [ngClass]="{ 'fd-input-group__addon--compact': contentDensity === 'compact' }" *ngIf="isLoading">
             <button
-                class="search-field--loading fd-input-group__button fd-button fd-button--transparent sap-icon--synchronize"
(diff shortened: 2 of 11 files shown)


libs/platform/src/lib/components/search-field/search-field.component.html
```diff
@@ -40,12 +40,13 @@
...
+        <nav class="fd-menu" tabindex="-1" role="menu" [ngClass]="{ 'fd-menu--compact': compact}" [attr.id]="menuId" (keydown)="onKeydown($event)">
             <ul class="fd-menu__list">
-                <li fdpSearchFieldSuggestion class="search-field--suggestion-item fd-menu__item"
+                <li fdpSearchFieldSuggestion class="search-field--suggestion-item fd-menu__item fd-menu__link"
                     *ngFor="let value of dropdownValues$ | async | suggestionMatches:inputText"
-                    (keydown.enter)="onItemClick(value)" (click)="onItemClick(value)"
-                    [innerHTML]="value | highlight:inputText:true"></li>
+                    (keydown.enter)="onItemClick(value)" (click)="onItemClick(value)">
+                    <span class="fd-menu__title" [innerHTML]="value | highlight:inputText:true"></span>
+                </li>
```
InnaAtanasova, line 49: Is it intentional that you are not following the Fundamental-styles structure? 
`fd-menu__item` and `fd-menu__link` are not applied on the same element. 

```
<li class="fd-menu__item">
      <a class="fd-menu__link" href="#">
           <span class="fd-menu__title">Option 1</span>
      </a>
</li>
```

KevinOkamoto: This was done because of the way the CDK FocusManager works with menu items. When doing keyboard navigation the CDK FocusManager puts the focus on the "host" item, i.e. the element with "fd-menu__item" and cannot invoke focus on child elements. As a result, if the usual Fundamental-styles structure is used the element with "fd-menu__link" never receives focus and then the default "blue" highlight appears instead of the "dotted outline" as per Fiori specs.

I looked at the Fundamental-Styles code, before going this route, and it looked like there two CSS rules `.fd-menu__item` and `.fd-menu__link` are not dependent/nested, and none of the CSS attributes seemed to conflict. However if this is not the case, please let me know.

KevinOkamoto:
@salarenko, yes this is PR is meant to be a fix for the display issues, but would you mind expanding on your comment 

> none of this components follow Fiori3 and Fundamental Styles structure, a11y and behavior

I'm aware that this doesn't follow Fundamental Styles structure, but I'm not clear on how this violates Fiori 3, a11y and behavior. If there are deviations from the expected design, I'd like to understand those and close them if possible.

If there are large requirements gaps, perhaps we can open a new issue.

KevinOkamoto:
@kavya-b yes, adding icons and separators are missing features which need to be added later. This PR was mostly to fix the broken styles.

fkolar:
> If this is a quick-fix then it looks good, but otherwise none of this components follow Fiori3 and Fundamental Styles structure, a11y and behavior.

@salarenko We should look into this, can you expand your comment? That none of these components follows F3?  Do you mean visually, interacation-wise? Not sure how to interpret your comment.

InnaAtanasova:
Looks good to me. I will wait for @fkolar to approve first since the PR can be merged with only my approval and we agreed to have approvals from both teams :)

InnaAtanasova:
@salarenko could you comment on what @fkolar wrote? Thanks

salarenko:
@fkolar @InnaAtanasova Sorry gyus I didn't see the notification.

- For behavior problems I've created this ticket #2371
- For aria attributes
    - Menu - I recommend checking out Fundamental Styles (updated recently). Current Platform implementation seems to be using List aria roles for Menu items.
    - Search - missing `type="search"` and using aria attributes for Menu, but I believe proper implementation should be based on List attributes and roles.
- I think that @InnaAtanasova already gave the example of the difference in HTML structure between Fundamental Styles Menu and Platform Menu. Since Menu is relatively new in Fundamental Styles (2/3 weeks old) Platform probably didn't adapt new changes yet.

KevinOkamoto:
@salarenko 

> * For behavior problems I've created this ticket #2371

Thanks for creating the ticket.

> * For aria attributes
>   
>   * Menu - I recommend checking out Fundamental Styles (updated recently). Current Platform implementation seems to be using List aria roles for Menu items.
>   * Search - missing `type="search"` and using aria attributes for Menu, but I believe proper implementation should be based on List attributes and roles.

I've updated the Menu and SearchField component templates with roles `menu` and `menuitem`.
I have also changed the input type to `search` for the Search Field. In doing so, I needed to add CSS to hide the default clear "x" icon button for IE and Chrome browsers, as the input clear button is not a feature of all browsers and doesn't comply with Fiori 3 design.

> * I think that @InnaAtanasova already gave the example of the difference in HTML structure between Fundamental Styles Menu and Platform Menu. Since Menu is relatively new in Fundamental Styles (2/3 weeks old) Platform probably didn't adapt new changes yet.

Looking at the current core Menu code it looks as if the `.fd-menu__link` and `.fd-menu__title` elements are not used. Are they considered optional elements? Copying core Menu, I've removed those elements from the platform Menu and SearchField, as this HTML structure works better with the CDK library used for the platform Menu dropdown.

@InnaAtanasova , @fkolar please review these changes as they have altered the original PR - thanks!

libs/platform/src/lib/components/search-field/search-field.component.html
```diff
@@ -1,106 +1,62 @@
...
-            <ul class="fd-menu__list">
-                <li
-                    fdpSearchFieldSuggestion
-                    class="search-field--suggestion-item fd-menu__item"
-                    *ngFor="let value of dropdownValues$ | async | suggestionMatches: inputText"
-                    (keydown.enter)="onItemClick(value)"
-                    (click)="onItemClick(value)"
-                    [innerHTML]="value | highlight: inputText:true"
-                ></li>
+    <div class="search-field--dropdown fd-popover__body fd-popover__body--no-arrow" [attr.dir]='dir'
```
InnaAtanasova, line 44: I have mixed feelings about the `search-field--dropdown` naming. It's not prefixed by `fd` or `fdp` so the enduser will have no idea this class is coming from fundamentals. It's not following the BEM methodology. `--` is used for modifier classes like `--compact`. If it's an element it should be with `__`

KevinOkamoto: I will update the class names to `fdp-search-field__*`.

KevinOkamoto:
@JKMarkowski 

> I don't know why you use classes there, instead of core's directives/components.
> `fd-button--*` class can be changed to `fd-button` directive with input properties. Same for Input group

The reason I couldn't use the `fd-button` directives of the `fd-input-group` component was because the Fiori 3 Search Field design calls for two buttons: one for search and one for cancel.

I needed the "input-group" button style, which the `fd-button` directive does not provide on its own. The `fd-input-group` only allows for one button to be added, so I had to go the fundamental-styles route to get the look and functionality I desired.

libs/platform/src/lib/components/search-field/search-field.component.html
KevinOkamoto: @InnaAtanasova OK, I've updated the style names.

stefanoScalzo:
> @salarenko
> 
> > * For behavior problems I've created this ticket #2371
> 
> Thanks for creating the ticket.
> 
> > * For aria attributes
> >   
> >   * Menu - I recommend checking out Fundamental Styles (updated recently). Current Platform implementation seems to be using List aria roles for Menu items.
> >   * Search - missing `type="search"` and using aria attributes for Menu, but I believe proper implementation should be based on List attributes and roles.
> 
> I've updated the Menu and SearchField component templates with roles `menu` and `menuitem`.
> I have also changed the input type to `search` for the Search Field. In doing so, I needed to add CSS to hide the default clear "x" icon button for IE and Chrome browsers, as the input clear button is not a feature of all browsers and doesn't comply with Fiori 3 design.
> 
> > * I think that @InnaAtanasova already gave the example of the difference in HTML structure between Fundamental Styles Menu and Platform Menu. Since Menu is relatively new in Fundamental Styles (2/3 weeks old) Platform probably didn't adapt new changes yet.
> 
> Looking at the current core Menu code it looks as if the `.fd-menu__link` and `.fd-menu__title` elements are not used. Are they considered optional elements? Copying core Menu, I've removed those elements from the platform Menu and SearchField, as this HTML structure works better with the CDK library used for the platform Menu dropdown.
> 
> @InnaAtanasova , @fkolar please review these changes as they have altered the original PR - thanks!

You can remove the x for IE, there has been a fix in fundamental-styles removing this https://github.com/SAP/fundamental-styles/pull/963

stefanoScalzo:
Not sure if something is wrong with netlify, but these do not look like what we have in fundamental styles. Also please look at the behavior of fiori 3. @mikerodonnell89 I don't think the search is matching the actions you had showed me on your pr
Screen Shot 2020-05-11 at 9 13 44 AM
Screen Shot 2020-05-11 at 9 14 29 AM

stefanoScalzo:
We should also probably be targetting menu and search field in different PRs

KevinOkamoto:
@stefanoScalzo What browser are you using and what URL you are going to to view these changes. This is what I see in Chrome when visiting https://deploy-preview-2294--fundamental-ngx.netlify.app/#/platform/search-field

Screen Shot 2020-05-11 at 7 03 24 AM

KevinOkamoto:
> We should also probably be targetting menu and search field in different PRs

The initial issue was to fix SearchField, however SearchField is dependent on Menu. I decided to tackle them both at the same time since the changes to Menu were relatively simple.

KevinOkamoto:
> You can remove the x for IE, there has been a fix in fundamental-styles removing this [SAP/fundamental-styles#963](https://github.com/SAP/fundamental-styles/pull/963)

@stefanoScalzo Has Fundamental-NGX been updated to the latest version of fundamental-styles with this change? If not I'd rather wait till that happens before making a change.

libs/platform/src/lib/components/search-field/search-field.component.html
```diff
@@ -1,106 +1,62 @@
...
-            (keydown.enter)="onSearchSubmit()"
-            [(ngModel)]="inputText"
-            (ngModelChange)="onValueChange($event)"
-        />
-        <span
-            class="fd-input-group__addon fd-input-group__addon--button"
-            [ngClass]="{ 'fd-input-group__addon--compact': compact }"
-            *ngIf="inputText && inputText.length > 0"
-        >
+        <input #inputField type="search" placeholder="{{ placeholder }}"
```
fkolar, line 17: Let's try to use `[attr.placeholder]` syntax over interpolation, unless there is a specific reason for it

libs/platform/src/lib/components/search-field/search-field.component.html
```diff
@@ -1,106 +1,62 @@
...
-        >
-            <button
-                class="search-field--category-button fd-input-group__button fd-button fd-button--menu"
-                [ngClass]="{ 'fd-button--compact': compact }"
-                [fdpMenuTriggerFor]="categoryMenu"
-            >
-                <span class="search-field--category-label" *ngIf="!hideCategoryLabel">{{
+    }">
+    <div class="fdp-search-field__input-group fd-input-group" [attr.aria-controls]="menuId"
+        [attr.aria-expanded]="showDropdown ? 'true' : 'false'" aria-haspopup="true" #inputGroup>
```
fkolar, line 7: Can we put ` #inputGroup` at the beginning of the tag like other lines ?

fkolar:
Besides latest comments I have these few:

- Can we update Compact:true|false to content density? 
- You might want to inherit this from base component..

libs/platform/src/lib/components/search-field/search-field.component.html
KevinOkamoto: I've changed it to `[attr.placeholder]`.

libs/platform/src/lib/components/search-field/search-field.component.html
KevinOkamoto: This has been fixed.

libs/platform/src/lib/components/menu/menu-item.component.ts
```diff
@@ -40,7 +40,7 @@ export class MenuItemComponent implements OnDestroy, FocusableOption {
         return this.cascadeDirection === 'left';
     }
```
fkolar, line 41: We should have our own strategy using hostAttributes vs HostBindings, so we can unify all this

libs/platform/src/lib/components/search-field/search-field.component.ts
```diff
@@ -122,10 +123,10 @@ export class SearchFieldComponent implements OnInit, OnDestroy {
      * Set size of search input component.
      */
     @Input()
-    get size(): 'cozy' | 'compact' {
+    get contentDensity(): 'cozy' | 'compact' {
```
fkolar, line 126: U dont need this, this is already defined in parent class. Parent class also defines a type for `'cozy' | 'compact'`

fkolar:
Also we probably want to have module for search field and is this documentation correct? 

https://sap.github.io/fundamental-ngx/#/platform/search-field

Do you import search and not module?

libs/platform/src/lib/components/search-field/search-field.component.ts
KevinOkamoto: OK. I've updated the code and removed the `get` and `set` of `contentDensity`. So the code is using the parent input. I had to update the template as well to not use the computed `compact` boolean and instead changed it to `contentDensity === 'compact'`.

libs/platform/src/lib/components/menu/menu-item.component.ts
KevinOkamoto: Not sure if this is necessary for this PR. We should address this separately.

fkolar: not at all just a comment.. 

> Not sure if this is necessary for this PR. We should address this separately.

Issue #2304, opened 2020-04-10
Defect hunting 10.04 for 0.16.0v


JKMarkowski:
Issues from defect hunting on 10.04.

mikerodonnell89:
Tokenizer RTL mode - broken padding, need to figure out which side to align "more" text on, maybe provide option for the dev to decide?
![Screen Shot 2020-04-10 at 8 39 35 AM](https://user-images.githubusercontent.com/2471874/78999389-bc28cf00-7b07-11ea-9a3b-0992e6bda9d2.png)

mikerodonnell89:
Last menu item has an extra bit of space before the bottom of the popover
![Screen Shot 2020-04-10 at 9 00 48 AM](https://user-images.githubusercontent.com/2471874/79000437-debbe780-7b09-11ea-9eb0-c9df38c60e52.png)

mikerodonnell89:
Should probably add some padding here
![Screen Shot 2020-04-10 at 9 05 07 AM](https://user-images.githubusercontent.com/2471874/79000729-673a8800-7b0a-11ea-973b-197557d4e77e.png)

mikerodonnell89:
Calendar - I think we should only display the week counter specifically on the week counter example.  It is a strange feature that I think will usually do more to confuse than anything.
![Screen Shot 2020-04-10 at 9 14 04 AM](https://user-images.githubusercontent.com/2471874/79001434-b634ed00-7b0b-11ea-936f-d45f85684b75.png)

JKMarkowski:
Calendar Keyboard support should consider RTL mode as well.

salarenko:
- No arrows navigation and spacebar as select support in calendar | IE11
- Dialog is not centered because of `flexbox` + `position: absolute` | IE11
- Dialog Body has rounded top corners hen header is visible | IE11
- Dialog cannot be dismissed by Esc key | IE11
- Dropdown links have visited native styling | IE11
![image](https://user-images.githubusercontent.com/17496353/79002205-3e3ce600-7b50-11ea-8cc7-c5243b807141.png)
- Input number spinner broken buttons | IE11
![image](https://user-images.githubusercontent.com/17496353/79002369-9b389c00-7b50-11ea-850f-40ac4b1bf7fa.png)
- Should these look like that? | Cross browser
![image](https://user-images.githubusercontent.com/17496353/79002500-e8b50900-7b50-11ea-859a-e86556c30e5a.png)
[...]

PR #2305, opened 2020-04-10
docs: popover placement menus

mikerodonnell89:

https://github.com/SAP/fundamental-ngx/issues/2304

For some reason, when importing layout.css, the "fd-col--" styles don't get imported.  So hardcoding here for now.  Wondering if anyone else has a better idea of how to fix in styles


droshev:
@mikerodonnell89 this fix is only in the documentation and it won't be part of the release

InnaAtanasova:
@mikerodonnell89 I renamed the PR to docs since the changes are concerning only the documentation part

PR #2327, opened 2020-04-12
chore: bump packages' versions in package.json

droshev:
bump packages' versions in package.json


droshev:
> There are conflicts

updated

PR #2335, opened 2020-04-13
fix: rtl tokenizer

mikerodonnell89:

#2304 

Fixes tokenizer keyboard support and resizing issues in RTL mode.

Depends on https://github.com/SAP/fundamental-styles/pull/861

### DIFF
--- a/libs/core/src/lib/token/tokenizer.component.ts
+++ b/libs/core/src/lib/token/tokenizer.component.ts
@@ -7,13 +7,16 @@ import {
     ContentChild,
     ContentChildren, ElementRef,
     forwardRef, HostListener,
-    Input,
+    Input, Optional,
     QueryList,
     ViewChild,
     ViewEncapsulation
 } from '@angular/core';
 import { FormControlDirective } from '../form/form-control/form-control.directive';
 import { TokenComponent } from './token.component';
+import { RtlService } from '../..';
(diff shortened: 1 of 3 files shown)


libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -169,14 +170,15 @@ export class TokenizerComponent implements AfterViewInit, AfterContentInit {
     /** @hidden */
     handleKeyDown(event: KeyboardEvent, fromIndex: number): void {
         let newIndex: number;
-        if (event.code === 'ArrowLeft') {
+        const rtl = this._rtlService.rtl.getValue();
```
salarenko, line 173: Won't you get `property of undefined` error if there will be no `RtlService` provided?

mikerodonnell89:
> One thing I notice is that the spacing between the token and the text is different from LTR layout:
> Screen Shot 2020-04-13 at 10 50 20 PM
> Screen Shot 2020-04-13 at 10 50 04 PM
> The rest looks great.

this will be fixed with https://github.com/SAP/fundamental-styles/pull/861

mikerodonnell89:
> This is not the scope of this PR, but there are two unguarded subscriptions in `TokenizerComponent`.

Actually I guess potentially a lot more since each token gets its own subscription

Issue #2111, opened 2020-03-04
fix(platform): Button has to be updated with fundamental-styles 0.7.0 changes


sKudum:
Is this a bug, enhancement, or feature request?
Enhancement

Briefly describe your proposal.
fundamental -styles -0.7.0 has below-mentioned changes, the same has to reflect in platform as well.

1).New button types added: fd-button--attention and fd-button--ghost
2).fd-button--light renamed to fd-button--transparent
3).fd-button--standard has been removed and replaced by the default (fd-button)

droshev:
@sKudum is this still valid? If not, can we close it?

sKudum:
Closing it, as it is already addressed  and merged in PR #2365

PR #2365, opened 2020-04-15
fix: (platform) button styles updated to 0.8 styles

sKudum:
#2111
1). Light is renamed to transparent
2).half type is removed as Fiori does not have type half.
3). [Options] attribute removed only [type] is present.
4). Ghost style type is added.
5). Documentation examples updated.
6). Test scenarios are tested for all the updated changes.

Documentation checklist:
- [x] Documentation Examples

### DIFF
--- a/libs/platform/src/lib/components/button/button.component.ts
+++ b/libs/platform/src/lib/components/button/button.component.ts
@@ -59,8 +59,8 @@ export class ButtonComponent extends BaseComponent {
     /**
      *  Handles button click
      */
-    public onBtnClick(event: any) {
-        this.buttonClicked.emit(event);
+    public onBtnClick($event: any) {
+        this.buttonClicked.emit($event);
     }
 
     /** @hidden */
--- a/libs/platform/src/lib/components/button/button.component.ts
+++ b/libs/platform/src/lib/components/button/button.component.ts
@@ -7,9 +7,16 @@ import {
     ElementRef
 } from '@angular/core';
 
-export type ButtonType = 'standard' | 'positive' | 'medium' | 'negative' | 'toolbar' | 'main';
-export type ButtonOptions = 'light' | 'emphasized' | '';
-
+export type ButtonType =
+    | ''
+    | 'standard'
+    | 'positive'
+    | 'negative'
+    | 'attention'
+    | 'ghost'
+    | 'transparent'
+    | 'emphasized'
+    | 'menu';
 @Component({
     selector: 'fdp-button',
     templateUrl: './button.component.html',
@@ -32,17 +39,13 @@ export class ButtonComponent {
     @Input()
     disabled: boolean;
 
-    /** The type of the button. Types include 'standard', 'positive', 'medium', and 'negative'.
-     * Leave empty for default (Action button).'*/
+    /** The type of the button. Types include 'standard','positive' and 'negative' etc.
+     * Leave empty for default (standard button).'*/
     @Input()
     type: ButtonType;
 
     @Input()
     fdType: ButtonType;
-
-    /** Button options.  Options include 'emphasized' and 'light'. Leave empty for default.' */
-    @Input()
-    options: ButtonOptions | ButtonOptions[];
 
     /** Event sent when button is clicked */
     @Output()
--- a/libs/platform/src/lib/components/button/button.component.ts
+++ b/libs/platform/src/lib/components/button/button.component.ts
@@ -4,8 +4,10 @@ import {
     Output,
     EventEmitter,
     ViewChild,
-    ElementRef
+    ElementRef,
+    ChangeDetectorRef
 } from '@angular/core';
+import { BaseComponent } from '../base';
 
 export type ButtonType =
     | ''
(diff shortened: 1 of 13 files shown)


libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -32,18 +39,14 @@ export class ButtonComponent {
...
 
-    /** The type of the button. Types include 'standard', 'positive', 'medium', and 'negative'.
-     * Leave empty for default (Action button).'*/
+    /** The type of the button. Types include 'standard','positive' and 'negative' etc.
+     * Leave empty for default (standard button).'*/
     @Input()
     type: ButtonType;
 
     @Input()
     fdType: ButtonType;
```
DeepakSap14, line 48: fdType can be removed, as only type is used in template.

sKudum: Done

rengare:
@sKudum I saw the documentation looks different for buttons in the platform. What do you think about sync it with core's?

sKudum:
> @sKudum I saw the documentation looks different for buttons on the platform. What do you think about sync it with core's?

Hi @rengare , Yes you are right.

**Reason**:
   In platform        
         1.               Button
         2.               Menu button 
         3.               Segmented button
         4.               Split menu button 

are given independent components unlike core. 

Once the respective components are ready then in documentation respective component example and information will be added.

InnaAtanasova:
@sKudum I updated the PR title but in future please make sure it follows the guidelines. I know it's only a PR title but it can really cause problems :)

libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -32,18 +39,11 @@ export class ButtonComponent {
     @Input()
     disabled: boolean;
 
-    /** The type of the button. Types include 'standard', 'positive', 'medium', and 'negative'.
-     * Leave empty for default (Action button).'*/
+    /** The type of the button. Types include 'standard','positive' and 'negative' etc.
```
InnaAtanasova, line 42: You should not use etc. All types need to be listed here since it's going to the API that the user is reading

sKudum:
> @sKudum I updated the PR title but in the future please make sure it follows the guidelines. I know it's only a PR title but it can really cause problems :)

Got your concern @Inna, even though I am **<type>: <subject>** formate I need to place platform or core only in the subject.Made a note will do the same.

libs/platform/src/lib/components/button/button.component.ts
sKudum: Removed **etc** and listed all type names.

libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -7,9 +7,16 @@ import {
     ElementRef
 } from '@angular/core';
 
-export type ButtonType = 'standard' | 'positive' | 'medium' | 'negative' | 'toolbar' | 'main';
-export type ButtonOptions = 'light' | 'emphasized' | '';
-
+export type ButtonType =
+    | ''
```
fkolar, line 11: Why do we need to redefine ButtonType, why we cannot use it? I see some new option  but what are those, do you have functionality or styling for it?

libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -7,9 +7,16 @@ import {
...
+    | 'standard'
+    | 'positive'
+    | 'negative'
+    | 'attention'
+    | 'ghost'
+    | 'transparent'
+    | 'emphasized'
+    | 'menu';
 @Component({
```
fkolar, line 20: Also we are creating BaseComponent connect to Loki and Deepak, so you can rename some of your bindings and extend it from base component.

libs/platform/src/lib/components/button/button.component.ts
sKudum: Styles are defined for these types.
and the changes are 

1).Light is renamed to Transparent.
2).Half type is removed.
3).emphasized will be part of type.
4).button options won't be supported, so as per 0.7 style update added emphasized, ghost, transparent to button types.

libs/platform/src/lib/components/button/button.component.ts
sKudum: Sure will do that, but Base component work is in progress and it is not available in master yet. 

Do you want me to wait until it is merged because PR won't be kept open for a long time it will get closed automatically after 20days. 

Or merge this now and update the code and raise one more PR once the base component is available in master by this I can proceed with app-framework documentation changes.

fkolar:
> > @sKudum I saw the documentation looks different for buttons on the platform. What do you think about sync it with core's?
> 
> Hi @rengare , Yes you are right.
> 
> **Reason**:
> In platform
> 1. Button
> 2. Menu button
> 3. Segmented button
> 4. Split menu button
> 
> are given independent components unlike core.
> 
> Once the respective components are ready then in documentation respective component example and information will be added.

@sKudum Do you think we can group them together? if I am app developer and looking at the docs I want to have these button all together what do you think ?

libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -4,45 +4,46 @@ import {
...
     selector: 'fdp-button',
     templateUrl: './button.component.html',
     styleUrls: ['./button.component.scss']
 })
-export class ButtonComponent {
+export class ButtonComponent extends BaseComponent {
+
     /** Option to make to button compact. */
     @Input()
     compact: boolean;
```
fkolar, line 31: This is one of the reason why we did base class to remove compact.

libs/platform/src/lib/components/button/button.component.ts
fkolar: Are you saying that core does not have this definition ? It does not have to perfect and they dont need to have everything, but let;s use one Type. . 

I assume for this you need to have support in styles as well right ?

sKudum:
> > > @sKudum I saw the documentation looks different for buttons on the platform. What do you think about sync it with core's?
> > 
> > 
> > Hi @rengare , Yes you are right.
> > **Reason**:
> > In platform
> > 
> > 1. Button
> > 2. Menu button
> > 3. Segmented button
> > 4. Split menu button
> > 
> > are given independent components unlike core.
> > Once the respective components are ready then in documentation respective component example and information will be added.
> 
> @sKudum Do you think we can group them together? if I am app developer and looking at the docs I want to have these button all together what do you think ?
> 
> They can have separate pagse, but cant we group them somehow together under some section?

Yes @Frank, that is the plan once the components are ready then they will be arranged something like below.

```
-->button
        -->menu button
        -->split menu button
        --->Segmented button
```

We can all examples in a single page as well but as we are giving them as separate components (unlike core). it will be confusing for users.

PR #2366, opened 2020-04-15
fix: remove circular dependency for Nested List 

JKMarkowski:
Those changes removed circular dependency warning/error and all of the issues, that comes with it.
Tested with The lib es6/ng9 lib.
We could achieve that, thanks to services and interfaces, which allowed to avoid circular imports.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/nested-list/nested-item/nested-item.directive.spec.ts
+++ b/libs/core/src/lib/nested-list/nested-item/nested-item.directive.spec.ts
@@ -76,7 +76,7 @@ describe('NestedItemDirective', () => {
         nestedItemListDirective = component.nestedItemListDirective;
         emptyItemDirective = component.emptyItemDirective;
         nestedItemPopoverDirective = component.nestedItemPopoverDirective;
-        itemService = (<any>nestedItemPopoverDirective).itemService;
+        itemService = (<any>nestedItemPopoverDirective)._itemService;
         fixture.detectChanges();
     });
 
--- a/libs/core/src/lib/nested-list/nested-item/nested-item.directive.spec.ts
+++ b/libs/core/src/lib/nested-list/nested-item/nested-item.directive.spec.ts
@@ -5,6 +5,7 @@ import { NestedListModule } from '../nested-list.module';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { NestedListKeyboardService } from '../nested-list-keyboard.service';
 import { NestedListStateService } from '../nested-list-state.service';
+import { NestedItemService } from './nested-item.service';
 
 @Component({
     template: `
@@ -57,6 +58,7 @@ describe('NestedItemDirective', () => {
     let nestedItemListDirective: NestedItemDirective;
     let emptyItemDirective: NestedItemDirective;
     let fixture: ComponentFixture<TestNestedContainerComponent>;
+    let itemService: NestedItemService;
 
     beforeEach(async(() => {
         TestBed.configureTestingModule({
@@ -74,6 +76,7 @@ describe('NestedItemDirective', () => {
         nestedItemListDirective = component.nestedItemListDirective;
         emptyItemDirective = component.emptyItemDirective;
         nestedItemPopoverDirective = component.nestedItemPopoverDirective;
+        itemService = (<any>nestedItemPopoverDirective).itemService;
         fixture.detectChanges();
     });
 
@@ -84,7 +87,7 @@ describe('NestedItemDirective', () => {
 
     it('Item with popover should react to open change from popover', () => {
         spyOn(nestedItemPopoverDirective, 'triggerOpen');
-        nestedItemPopoverDirective.popoverItem.handleOpenChange(true);
+        itemService.popover.handleOpenChange(true);
         expect(nestedItemPopoverDirective.triggerOpen).toHaveBeenCalled();
     });
 
@@ -102,7 +105,7 @@ describe('NestedItemDirective', () => {
     });
 
     it('Item with popover, should pass reference to popover child', () => {
-        expect(nestedItemPopoverDirective.popoverItem.parentItemElement).toBe(nestedItemPopoverDirective);
+        expect(itemService.popover.parentItemElement).toBe(nestedItemPopoverDirective);
     });
 
     it('Item with list should have children', () => {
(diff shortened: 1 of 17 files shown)


libs/core/src/lib/nested-list/nested-item/nested-item.directive.ts
```diff
@@ -4,20 +4,20 @@ import {
...
+    Input, OnDestroy,
     Output
 } from '@angular/core';
 import { NestedLinkDirective } from '../nested-link/nested-link.directive';
 import { NestedListKeyboardService } from '../nested-list-keyboard.service';
-import { NestedListPopoverComponent } from '../nested-list-popover/nested-list-popover.component';
 import { NestedItemInterface } from './nested-item.interface';
-import { NestedListDirective } from '../nested-list/nested-list.directive';
-import { PreparedNestedListComponent } from '../prepared-nested-list/prepared-nested-list.component';
+import { NestedItemService } from './nested-item.service';
```
InnaAtanasova, line 15: Missing typedocs

libs/core/src/lib/nested-list/nested-item/nested-item.directive.ts
```diff
@@ -110,9 +91,9 @@ export class NestedItemDirective implements AfterContentInit, NestedItemInterfac
             );
         }
 
-        /** Pass this element to popover child item, to allow control `expanded` value */
-        if (this.popoverItem) {
-            this.popoverItem.parentItemElement = this;
+        // /** Pass this element to popover child item, to allow control `expanded` value */
```
salarenko, line 94: Is that commented out comment?

libs/core/src/lib/nested-list/nested-list-keyboard.service.ts
```diff
@@ -56,7 +55,7 @@ export class NestedListKeyboardService {
     private getAllListItems(list: NestedListInterface): NestedItemInterface[] {
 
         const _items: NestedItemInterface[] = [];
-        if (list && list.nestedItems) {
+        if (list && list.nestedItems && list.nestedItems.toArray()) {
```
salarenko, line 58: I `nestedItems` is a `QueryList`. You can use `QueryList.length` to improve performance a bit.

libs/core/src/lib/nested-list/nested-list-keyboard.service.ts
```diff
@@ -56,7 +55,7 @@ export class NestedListKeyboardService {
...
             list.nestedItems.toArray().forEach(item => {
```
salarenko, line 59: No need to use `toArray()`. `QueryList` supports `forEach`

libs/core/src/lib/nested-list/nested-item/nested-item.directive.spec.ts
```diff
@@ -57,6 +58,7 @@ describe('NestedItemDirective', () => {
     let nestedItemListDirective: NestedItemDirective;
     let emptyItemDirective: NestedItemDirective;
     let fixture: ComponentFixture<TestNestedContainerComponent>;
+    let itemService: NestedItemService;
```
InnaAtanasova, line 61: Shouldn't itemService be a const?

JKMarkowski: It's overwritten below. There is only declaration

libs/core/src/lib/nested-list/prepared-nested-list/prepared-nested-list.component.ts
```diff
@@ -40,28 +47,28 @@ export class PreparedNestedListComponent implements AfterViewInit {
...
+        private _changeDetRef: ChangeDetectorRef,
+        @Optional() private _nestedItemService: NestedItemService
     ) {}
 
     /** @hidden */
     ngAfterViewInit(): void {
-        this.changeDetRef.markForCheck();
-        this.changeDetRef.detectChanges();
+        this._changeDetRef.markForCheck();
+        this._changeDetRef.detectChanges();
```
salarenko, line 67: Does it make sense to call `this._changeDetRef.markForCheck();` right before forcing change detection?

libs/core/src/lib/nested-list/nested-item/nested-item.directive.ts
JKMarkowski: Yes :D

libs/core/src/lib/nested-list/nested-item/nested-item.directive.ts
```diff
@@ -32,42 +32,24 @@ export class NestedItemDirective implements AfterContentInit, NestedItemInterfac
...
-    nestedListItem: NestedListDirective;
-
-    /** @hidden */
-    @ContentChild(forwardRef(() => PreparedNestedListComponent))
-    preparedListComponent: PreparedNestedListComponent;
-
     /** Check if the item element has any child */
     public get hasChildren(): boolean {
-        return !!(this.nestedListItem || this.popoverItem || this.nestedListFromPreparedComponent);
+        return !!(this._itemService && this._itemService.list);
```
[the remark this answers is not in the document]
JKMarkowski: Typescript doesn't really like it, `types not compatible`. For statements `if(this._itemService && this._itemService.list)` it would be ok.

mikerodonnell89:
Tested this in a separate project and works great, thanks @JKMarkowski

libs/core/src/lib/nested-list/nested-item/nested-item.directive.ts
```diff
@@ -32,42 +31,24 @@ export class NestedItemDirective implements AfterContentInit, NestedItemInterfac
...
-            /** Get elements from child list */
-            return this.nestedListItem.nestedItems.toArray();
-
-        } else if (this.nestedListFromPreparedComponent && this.nestedListFromPreparedComponent.nestedItems) {
-            /** Get elements from child prepared list  component */
-            return this.nestedListFromPreparedComponent.nestedItems.toArray();
-
+        if (this._itemService && this._itemService.list) {
+            return this._itemService.list.nestedItems.toArray()
         } else {
```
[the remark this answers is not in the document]
JKMarkowski: It would break service recursive logic. I wanted to always return valid array

Issue #2368, opened 2020-04-16
enhancement: (Platform) Create common base component for Platform libraries


DeepakSap14:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
In Platform we have multiple components with some common properties eg. id, name, placeholder etc. these can be moved to a common class and can be extended by all components being developed or to be developed. This PR includes base class and changes to existing components to extend this base class.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 9.1.1

#### If this is a bug, please provide steps for reproducing it.
No Bug

#### Is there anything else we should know?
Radio Button Group, combo-box, and input components modified for extending this class.
This change only affects platform libraries and their corresponding documentation.

PR #2369, opened 2020-04-16
fix: (platform) Create common base class for Platform components

DeepakSap14:
https://github.com/SAP/fundamental-ngx/issues/2368

In Platform we have multiple components with some common properties eg. id, name, placeholder etc. these can be moved to a common class and can be extended by all components being developed or to be developed. This PR includes base class and changes to existing components to extend this base class.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/platform/src/lib/components/base.ts
+++ b/libs/platform/src/lib/components/base.ts
@@ -1,0 +1,38 @@
+import { Input, ChangeDetectorRef } from '@angular/core';
+import { InputSize } from './form/form-control';
+
+let randomId = 0;
+
+/**
+ * This class contains common properties used across components. 
+ * this can be extended to reduce the code duplication across components.
+ * @hidden for form related Base , see BaseInput.
+ */
+export class BaseComponent {
+    protected defaultId: string = `fdp-id-${randomId++}`;
+    protected _disabled: boolean;
+
+    @Input()
+    id: string = this.defaultId;
+
+    @Input()
+    name: string;
+
+    @Input()
+    placeholder: string;
+
+    @Input()
+    contentDensity: InputSize = 'cozy';
+
+    @Input()
+    get disabled(): boolean {
+        return this._disabled;
+    }
+
+    set disabled(disabled: boolean) {
+        this._disabled = disabled;
+    }
+
+    constructor(protected _cd: ChangeDetectorRef) {}
+
+}
--- a/libs/platform/src/lib/components/base.ts
+++ b/libs/platform/src/lib/components/base.ts
@@ -1,29 +1,34 @@
 import { Input, ChangeDetectorRef } from '@angular/core';
-import { InputSize } from './form/form-control';
+import { ContentDensity } from './form/form-control';
 
 let randomId = 0;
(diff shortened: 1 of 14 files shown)


libs/platform/src/lib/components/base.ts
```diff
@@ -0,0 +1,38 @@
...
+    protected _disabled: boolean;
+
+    @Input()
+    id: string = this.defaultId;
+
+    @Input()
+    name: string;
+
+    @Input()
+    placeholder: string;
```
fkolar, line 22: Placeholder makes sense  only for inputs, unless you can tell me o the situatoins

libs/platform/src/lib/components/base.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    @Input()
+    contentDensity: InputSize = 'cozy';
```
fkolar, line 25: The reason why we renamed bindings is to also fix naming for InputSize that does not make sense, Rename inputsize to ContentDensity

libs/platform/src/lib/components/base.ts
```diff
@@ -0,0 +1,38 @@
...
+ * this can be extended to reduce the code duplication across components.
+ * @hidden for form related Base , see BaseInput.
+ */
+export class BaseComponent {
+    protected defaultId: string = `fdp-id-${randomId++}`;
+    protected _disabled: boolean;
+
+    @Input()
+    id: string = this.defaultId;
+
```
fkolar, line 17: also add in there Inputs(), called width

libs/platform/src/lib/components/base.ts
```diff
@@ -0,0 +1,38 @@
...
+import { InputSize } from './form/form-control';
+
+let randomId = 0;
+
+/**
+ * This class contains common properties used across components. 
+ * this can be extended to reduce the code duplication across components.
+ * @hidden for form related Base , see BaseInput.
+ */
+export class BaseComponent {
```
fkolar, line 11: make this class abstract

libs/platform/src/lib/components/form/form-control.ts
```diff
@@ -30,7 +30,7 @@ export abstract class FormFieldControl<T> {
     /**
      *  Components works in two sizes compact or cozy
      */
-    size: InputSize;
+    contentDensity: InputSize;
```
fkolar, line 33: see my other comment

DeepakSap14: have pushed changes with reviews implemented. please check.

fkolar:
> I'm not sure `BaseComponent` should contain the `width` and `disabled` properties. These properties pertain to inline elements and input elements. I don't think they'll be used for components like ActionBar or List.

We should see if there are other cases for this: 
- Alert
- button
- etc..

If we cannot see other usecase, even I think container could have width and could be disabled, then let;s remove it, we can add it back anything

DeepakSap14:
Merging it after master rebase.

PR #2375, opened 2020-04-17
chore: remove tslib from devDependencies and add repo and homepage link

droshev:
remove tslib from devDependencies and add repo and homepage link from package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,11 @@
   "license": "Apache-2.0",
   "main": "dist/libs/core/index.js",
   "description": "SAP Fundamental Styles, implemented in Angular",
+  "homepage": "https://sap.github.io/fundamental-ngx",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/SAP/fundamental-ngx"
+  },
   "scripts": {
     "build": "ng build",
     "build-all": "npm run build-artifacts && npm run build-docs -- --prod --base-href=\"/fundamental-ngx/\"",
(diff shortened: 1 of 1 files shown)


fkolar:
@droshev I think you need this tslib inside dependencies, I woudl not remove it. 

As part of Ng9 is not automatically added to dependencies but they recomend to add it there, if you are are using `https://github.com/SAP/fundamental-ngx/blob/master/tsconfig.json` importHelpers and we do use them. 


Did somebody from approvers actually checked this PR?  What is the reason for removing it?

droshev:
> @droshev I think you need this tslib inside dependencies, I woudl not remove it.
> 
> As part of Ng9 is not automatically added to dependencies but they recomend to add it there, if you are are using `https://github.com/SAP/fundamental-ngx/blob/master/tsconfig.json` importHelpers and we do use them.
> 
> Did approvers actually checked this PR? What is the reason for removing it?
> 
> https://github.com/microsoft/tslib/blob/master/README.md
> 
> https://angular.io/guide/updating-to-version-9#changes-and-deprecations-in-version-9

@fkolar  we have it as dependency and devDependency. Do we need it in both places

fkolar:
> > @droshev I think you need this tslib inside dependencies, I woudl not remove it.
> > As part of Ng9 is not automatically added to dependencies but they recomend to add it there, if you are are using `https://github.com/SAP/fundamental-ngx/blob/master/tsconfig.json` importHelpers and we do use them.
> > Did approvers actually checked this PR? What is the reason for removing it?
> > https://github.com/microsoft/tslib/blob/master/README.md
> > https://angular.io/guide/updating-to-version-9#changes-and-deprecations-in-version-9
> 
> @fkolar we have it as dependency and devDependency. Do we need it in both places

Ok. I just saw it removing it completly. Yes it should be under dependencies. I remember when I was doing refactoring on whole repo, I did similar change.. , then let's merge it..

Issue #1954, opened 2020-02-12
storybook tests for Fiori 3 components


droshev:
#### Is this a bug, enhancement, or feature request?
enhancement
#### Briefly describe your proposal.

create storybooks for the following components
- [x] breadcrumb
- [x] input
- [x] inline help
- [x] input group
- [x] input states
- [x] textarea
- [ ] form select
- [x] link
- [x] notifications
- [x] product switch
- [x] side navigation

stefanoScalzo:
@droshev form select as in select component or checkbox/radio ?

PR #2400, opened 2020-04-23
test: add textarea to storybook

stefanoScalzo:
https://github.com/SAP/fundamental-ngx/issues/1954
This or adds textarea to storybook
The disabled for textarea is broken, an issue as been created and will be targeted in another pr

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/src/stories/textarea/fd-textarea-stories.ts
+++ b/src/stories/textarea/fd-textarea-stories.ts
@@ -1,0 +1,44 @@
+import { moduleMetadata } from '@storybook/angular';
+import { withKnobs, boolean, select, text } from '@storybook/addon-knobs';
+import { withA11y } from '@storybook/addon-a11y';
+import { ReactiveFormsModule } from '@angular/forms';
+
+import { FormControlDirective, FormModule } from 'libs/core/src/lib/form/public_api';
+
+export default {
+    title: 'Fd text-area',
+    component: FormControlDirective,
+    moduleMetadata: moduleMetadata,
+    decorators: [
+        withKnobs,
+        withA11y,
+        moduleMetadata({
+            imports: [FormModule, ReactiveFormsModule],
+            declarations: []
+        })
+    ]
+};
+
+export const TextArea = () => ({
+    template:
+        `
+        <div fd-form-item>
+        <label fd-form-label for="input-1">Default Text Area</label>
+        <textarea fd-form-control type="text" id="input-1" placeholder="Field placeholder text"
+        [compact]="compactVar"
+        [state]="stateVar"
+        >
+        </textarea>
+        </div>
+  `,
+    props: {
+        stateVar: select('State', {
+            information: 'information',
+            success: 'success',
+            error: 'error',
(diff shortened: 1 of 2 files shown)


src/stories/textarea/fd-textarea-stories.ts
```diff
@@ -0,0 +1,44 @@
...
+        })
+    ]
+};
+
+export const TextArea = () => ({
+    template:
+        `
+        <div fd-form-item>
+        <label fd-form-label for="input-1">Default Text Area</label>
+        <textarea fd-form-control type="text" id="input-1" placeholder="Field placeholder text"
```
InnaAtanasova, line 27: Based on this PR https://github.com/SAP/fundamental-ngx/pull/2401/files the values here should be dynamic (knobs)

stefanoScalzo: > the filename is incorrect. it should be `fd-textarea.stories.ts`.
> I am not sure how you ran this.
> 
Weird it was working mustve clicked it by accident before pushing
> * [ ]  can you add a knob for placeholder, or at least set one

Sure! thought we were only testing the API on the website but will add that now!
> * [ ]  It would be good to have an example with form message or maybe create another story for textarea with form message

Each storybook PR was meant to do one single thing. The storybook with form-messages will be in another PR!
> * [ ]  option to disable the textarea
Added

mikerodonnell89:
Not sure why but the placeholder disappears after toggling the disabled knob. Also there is an accessibility warning

JKMarkowski:
Let's remove initial value, there are some whitespaces
image

stefanoScalzo:
> Not sure why but the placeholder disappears after toggling the disabled knob. Also there is an accessibility warning

accessibility warning fixed...disabled is actually a bug we have with textarea

stefanoScalzo:
> Let's remove initial value, there are some whitespaces
> image

This was caused by the line break which is fixed

src/stories/textarea/fd-textarea.stories.ts
```diff
@@ -0,0 +1,45 @@
...
+export const TextArea = () => ({
+    template:
+        `
+        <div fd-form-item>
+        <label fd-form-label for="input-1">Default Text Area</label>
+        <textarea fd-form-control id="id" [attr.aria-label]="ariaLabel" [placeholder]="placeholderVar" [disabled]="disabledVar" [compact]="compactVar" [state]="stateVar"></textarea>
+        </div>
+  `,
+    props: {
+        stateVar: select('State', {
```
rengare, line 31: avoid using `Var` suffix, it is a good practice to give variable more meaningfull name (but without suffix like Var, List, Array)
ie. 
- state
- buttonOptions
- discountPerProductIncludingClientDiscount

stefanoScalzo: fixed

droshev:
@stefanoScalzo the pipelines are failing. Once they are green i can approve it

PR #2401, opened 2020-04-23
test: add text knobs for bar

stefanoScalzo:
https://github.com/SAP/fundamental-ngx/issues/1954
This or will add text knobs for left,middle and right section of the bar component.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/src/stories/bar/fd-bar.stories.ts
+++ b/src/stories/bar/fd-bar.stories.ts
@@ -95,7 +95,7 @@ export const Bar = () => ({
         fullwidthVar: boolean('Full width element', true)
 =======
         fullwidthVar: boolean('Full width element', true),
-        leftText: text('Left Section Text', 'Section'),
+        leftText: text('Left Section Text', 'Left Section with very very very very very very very very very very very very very very very very very very  very very very very very very long text'),
         middleText: text('Middle Section Text', 'Section'),
         rightText: text('RightSection Text', 'Section'),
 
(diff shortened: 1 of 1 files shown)


InnaAtanasova:
What's the value of this change? This is just text. We are not testing any functionality with this.

stefanoScalzo:
> What's the value of this change? This is just text. We are not testing any functionality with this.

Anything that we should be able to change such as text or anything else related to that component needs to have a knob

stefanoScalzo:
> I don't see any value

It could be testing te width of the text, ellipses whatever else. 
In the breadcrumb example, I had to make sure all the texts had knobs so I am doing the same thing here

stefanoScalzo:
> I see value of it, as Stefano said it allows to test truncation/overflow of text. I would keep these changes

@droshev 
Screen Shot 2020-04-28 at 9 40 57 AM

The text is inside a bar-element and belongs to the bar component. This will test the full width spec of the fd-bar-element

salarenko:
I agree with @JKMarkowski and @stefanoScalzo. If there is a certain way how text should behave in Bar, then we should have a possibility to test it.

src/stories/bar/fd-bar.stories.ts
```diff
@@ -91,6 +91,15 @@ export const Bar = () => ({
...
+<<<<<<< HEAD
         fullwidthVar: boolean('Full width element', true)
+=======
+        fullwidthVar: boolean('Full width element', true),
+        leftText: text('Left Section Text', 'Left Section with very very very very very very very very very very very very very very very very very very  very very very very very very long text'),
+        middleText: text('Middle Section Text', 'Section'),
+        rightText: text('RightSection Text', 'Section'),
+
+        
+>>>>>>> test:add text knobs for bar
```
InnaAtanasova, line 103: Leftovers from rebase

PR #2403, opened 2020-04-23
fix: add aria label to notification header for a11y

stefanoScalzo:

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/notification/notification-header/notification-header.component.html
+++ b/libs/core/src/lib/notification/notification-header/notification-header.component.html
@@ -2,7 +2,7 @@
 <ng-content select="[fd-notification-title]"></ng-content>
 <button
     fd-button
-    aria-label="close-notification"
+    [attr.aria-label]="closeNotificationLabel"
     [glyph]="'decline'"
     [fdType]="'transparent'"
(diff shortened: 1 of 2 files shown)


libs/core/src/lib/notification/notification-header/notification-header.component.html
```diff
@@ -2,6 +2,7 @@
 <ng-content select="[fd-notification-title]"></ng-content>
 <button
     fd-button
+    aria-label="close-notification"
```
salarenko, line 5: Should we use kebab-case here? This is normal text value that should be read by device (at least that's how I understand it).

JKMarkowski: Shouldn't we have a way to add it in different language?

stefanoScalzo: fixed, the user will now decide like they do in shellbar :)

JKMarkowski: Thanks Stefano, we can merge it :D

Issue #2364, opened 2020-04-15
ViewDestroyedError: Attempt to use a destroyed view: detectChanges


zaheral:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
There is an open bug issue for angular that effects also the fundamental-ngx components: https://github.com/angular/angular/issues/27803

fundamental-ngx components are using the ControlValueAccessor like in the [Checkbox](https://github.com/SAP/fundamental-ngx/blob/d7c06dcd63af0013de4fb99e5c3680736d073af7/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts#L29). The API's of this interface (e.g. [writeValue](https://github.com/SAP/fundamental-ngx/blob/d7c06dcd63af0013de4fb99e5c3680736d073af7/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts#L108)) are called if some API's of the bined formControl is called (e.g. patchValue). Inside the writeValue API, you are calling:  "this._changeDetectorRef.detectChanges();", which causing following error if the component is destroyed:
````
ERROR Error: ViewDestroyedError: Attempt to use a destroyed view: detectChanges
    at viewDestroyedError (core.js:19641)
    at Object.debugUpdateDirectives [as updateDirectives] (core.js:30040)
    at checkAndUpdateView (core.js:29439)
    at callWithDebugContext (core.js:30309)
    at Object.debugCheckAndUpdateView [as checkAndUpdateView] (core.js:30011)
    at ViewRef_.push.../../node_modules/@angular/core/fesm5/core.js.ViewRef_.detectChanges (core.js:20686)
````
After the error, the component is also not working fine. The checkbox doesn't reacts anymore on the clicks.

This happens, if the component will be destroyed, but the formControl will still in use. See following example code:
````
<div fd-form-item [formGroup]="myformGroup">

  <div *ngIf="( viewMode$ | async ) === 'EDIT'"> 
      <fd-checkbox [formControlName]="someFormControl"></fd-checkbox>
  </div>

  <div *ngIf="( viewMode$ | async ) === 'DISPLAY'"> 
     <!--   Just show the value, without combobox  -->
  </div>

</div>
````

[...]

PR #2427, opened 2020-04-27
fix: Checkbox check destroyed state on detect changes

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/2364
There is check on every change detection, to not trigger it, when component is destroyed

### DIFF
--- a/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
+++ b/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
@@ -179,7 +179,6 @@ export class CheckboxComponent implements ControlValueAccessor {
         } else if (this.tristate && this._compare(this.checkboxValue, this.values.thirdStateValue)) {
             this.checkboxState = 'indeterminate';
         }
-        this._detectChanges();
     }
 
     /** @hidden */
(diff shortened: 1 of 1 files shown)


libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
```diff
@@ -208,4 +208,11 @@ export class CheckboxComponent implements ControlValueAccessor {
     private _isSpaceBarEvent(event: KeyboardEvent): boolean {
         return event.keyCode === 32;
     }
+
+    /** Method to trigger change detection in component */
+    private _detectChanges(): void {
+        if (!this._changeDetectorRef['destroyed']) {
```
stefanoScalzo, line 214: I think if it doesnt exist we should detach it. This just hides the real cause. The error is a symptom not the cause.

libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
```diff
@@ -208,4 +208,11 @@ export class CheckboxComponent implements ControlValueAccessor {
...
+            this._changeDetectorRef.detectChanges();
+        }
+    }
```
InnaAtanasova, line 217:
```
if (!(this.changeDetector as ViewRef).destroyed) {
  this._changeDetector.detectChanges()
}
```
If you want to make use of the types but your solution is great so I will approve

PR #2447, opened 2020-04-30
fix: combobox click autocomplete

### LINKED ISSUE #2285: Combobox open/close behavior improvements
The open/close behavior on our combobox differs between our implementation and ui5's.  Ours should match theirs.  Here is an explanation of the differences, thanks to @salarenko for compiling this list

https://sapui5.hana.ondemand.com/#/entity/sap.m.ComboBox/sample/sap.m.sample.ComboBox

**NGX Core Combobox vs SAPUI5 Combobox behaviour:**

| Action | NGX Core  | SAPUI5 |
| ------------- | ------------- | ------------- |
| Focus input with `Tab` and start typing | No action | Displays list of elements matching provided input. |
| Click on Combobox `input` | Displays list of all options | No action |
| Click on Combobox `input` and start typing | Filters elements lists, makrs list element only when its 1:1 with typed in query, element cannot be selected without manual `Arrow Down` navigation and `Enter` selection. | Filters elements lists, marks most accurate result on the list which can be selected by pressing `Enter` key. |
| Select an element from the list and click on `Input expand button` | Shows results filtered by selected element. | Show all possible options with currently selected element focused. |
| Click on Combobox `input` and start typing | No element has visible focus. | Input is focused. |
| Expand options list | No visuals for Combobox input group changed | Expand button marked as active |

mikerodonnell89:

fixes #2285 

This PR modifies our current combobox functionality to more closely match that of ui5


libs/core/src/lib/combobox/combobox.component.spec.ts
```diff
@@ -7,7 +7,7 @@ import { ListModule } from '../list/list.module';
 import { PipeModule } from '../utils/pipes/pipe.module';
 import { InputGroupModule } from '../input-group/input-group.module';
 
-describe('ComboboxComponent', () => {
+fdescribe('ComboboxComponent', () => {
```
InnaAtanasova, line 10: forgot to get rid of the focus

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -368,10 +407,12 @@ export class ComboboxComponent implements ControlValueAccessor, OnInit, OnChange
             this.open = isOpen;
             this.openChange.emit(this.open);
             this.onTouched();
-            if (isOpen) {
+            if (isOpen && this.displayedValues && this.displayedValues.length) {
```
InnaAtanasova, line 410: I see this check in 3 places. Wouldn't it be better if we move the logic to a function?

InnaAtanasova:
Screen Shot 2020-05-01 at 4 48 24 PM

When the options list is expanded the button should be active (gets the dark blue background).

JKMarkowski:
@mikerodonnell89 When the option is focused, shouldn't we allow user to write?

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -353,13 +383,22 @@ export class ComboboxComponent implements ControlValueAccessor, OnInit, OnChange
         if (this.popoverComponent) {
             this.popoverComponent.updatePopover();
         }
+        if (this.displayedValues && this.displayedValues.length === 0 && this.inputGroup.inputGroupButtonElement) {
+            this.inputGroup.inputGroupButtonElement.elementRef().nativeElement.classList.remove('is-expanded');
+        } else if (this.inputGroup.inputGroupButtonElement) {
+            this.inputGroup.inputGroupButtonElement.elementRef().nativeElement.classList.add('is-expanded');
+        }
```
JKMarkowski, line 390: If you need to get reference to button, or bind something, I encourage to use `fd-input-group-addon` - [example](https://fundamental-ngx.netlify.app/#/core/inputGroup#complex). You would be able to put `<button [ngClass]={'is-expanded": isExpanded} [glyph]="..."></button>`

mikerodonnell89:
@InnaAtanasova None of our inputs have focus outlines, is that something we want to introduce? It will also need to be done on the styles side

libs/core/src/lib/combobox/combobox.component.html
```diff
@@ -72,7 +73,7 @@
                     fd-list-item
                     tabindex="0"
                     class="fd-combobox-list-item"
-                    [selected]="displayFn(term) === inputText"
+                    [selected]="displayFn(term) === searchInputElement.value"
```
rengare, line 76: Regarding to our discussion about performance, we should avoid using methods in templates,
in this case, I think BehaviourSubject would be better

mikerodonnell89: Maybe not be the best place for that, because the idea behind having displayFn be an Input() here is to allow application developers to provide their own functions that handle displaying the string.  If we used a behavior subject here, rather than just accepting a function that returns a string, we would be adding another layer of complexity we're going to have to explain ourselves about

JKMarkowski:
combobox1
We should handle key down/up there

JKMarkowski:
I know it's quite hard to achieve, but maybe let's try to add some listener for alphabetical 
combobox2
Even when you are focusing some options,

JKMarkowski:
It's not part of this PR, but we should have some event thrown, when value on input is changed. For object mode it would be handy

salarenko:
I think we should rethink Combobox implementation and API

mikerodonnell89:
@salarenko 
> After selecting list element popover closes -> opens -> closes

Are you sure about this? If I set a console.log at the top of isOpenChangeHandle, it gets logged once.

mikerodonnell89:
@salarenko 
> In some cases default filtering doesn't refresh properly

I don't understand what this one means

mikerodonnell89:
@salarenko 

>  Combobox programmatic control not refreshing Combobox state

Do we still want this now that there is no open input?  I guess we could document the openChangeHandle method

mikerodonnell89:
> * What is the point of that getter?

IIRC this is standard practice when implementing ControlValueAccessor

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -286,7 +285,34 @@ export class ComboboxComponent implements ControlValueAccessor, OnInit, OnChange
             event.key !== 'Enter'
         ) {
             this.isOpenChangeHandle(true);
+            if (this.open && this._hasDisplayedValues() &&
+                    (!this.oldInputText || this.oldInputText !== this.inputText)) {
```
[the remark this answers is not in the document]
mikerodonnell89: If the user performs some keyboard event that we don't catch that doesn't result in changing the value in the input

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -191,16 +191,29 @@ export class ComboboxComponent implements ControlValueAccessor, OnInit, OnChange
...
     displayedValues: any[] = [];
 
     /** @hidden */
     inputTextValue: string;
 
+    /** @hidden */
+    oldInputText: string;
+
+    /** @hidden */
+    isExpanded: boolean = false;
```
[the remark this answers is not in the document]
mikerodonnell89: You're right, not sure what I was thinking here, removing

stefanoScalzo:
Build is failing

mikerodonnell89:
> We definitely need to escape that

What does this mean?

salarenko:
@mikerodonnell89
> We definitely need to escape
- We need to escape forcing user to use `@Input() inShellbar` by himself. This should be moved into Shellbar styles or if that's not possible, done under the hood.

> After selecting list element popover closes -> opens -> closes
- Sorry I was wrong, this error is because of that: "After selecting list element, all list elements appear (list has been previously filtered) and then popover is closed."


> In some cases default filtering doesn't refresh properly
- When you are typing highlighting doesn't catch all searched phrase.
On GIF below you can see that when I'm typing each time 'apple', Combobox reacts in different ways:
    - sometimes catching whole phrase and displaying all items with complete match item focused
    - sometimes its catching only 'appl' and its still in filtering mode

combobox-incomplete-marker

JKMarkowski:
I found some weird behaviour on combobox
combobox123

mikerodonnell89:
> > We definitely need to escape
> 
> * We need to escape forcing user to use `@Input() inShellbar` by himself. This should be moved into Shellbar styles or if that's not possible, done under the hood.

This is currently done under the hood, shellbar pattern does it automatically. User doesn't have to do anything

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -379,14 +398,52 @@ export class ComboboxComponent implements ControlValueAccessor, OnInit, OnChange
...
     /** Method that reset filtering for displayed values. It overrides displayed values by all possible dropdown values */
     public resetDisplayedValues(): void {
         this.displayedValues = this.dropdownValues;
     }
 
+    /** @hidden */
+    private _setupKeyboardService(): void {
+        this._menuKeyboardService.itemClicked
+            .pipe(takeUntil(this.onDestroy$))
+            .subscribe((index) => this.onMenuClickHandler(index));
```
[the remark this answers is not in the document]
mikerodonnell89: By unguarded do you mean, not unsubscribed? That is what takeUntil is for.  Correct me if I'm wrong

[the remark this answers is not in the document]
salarenko: Ah sorry, I didn't notice this `takeUntil`

salarenko:
> > > We definitely need to escape
> > 
> > 
> > 
> > * We need to escape forcing user to use `@Input() inShellbar` by himself. This should be moved into Shellbar styles or if that's not possible, done under the hood.
> 
> This is currently done under the hood, shellbar pattern does it automatically. User doesn't have to do anything

So what is the point of keeping this as a part of component API?

mikerodonnell89:
> "Custom Item Template" and "Combobox with Two Columns" examples are broken.
> Please test your changes and check is reported issues have been solved. I still see issues I've reported last time.

@salarenko forgot to test those after a change in the last commit, fixed.  However I just went through all the issues you reported last time and can't find anything else still broken?

JKMarkowski:
I noticed, when I use keyboard only, after selecting value by enter, the combobox is opened again 
combobox10

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -24,12 +24,14 @@ import { ListItemDirective } from '../list/list-item.directive';
...
 import { MenuKeyboardService } from '../menu/menu-keyboard.service';
-import { Subject } from 'rxjs';
+import { BehaviorSubject, Subject } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 import focusTrap, { FocusTrap } from 'focus-trap';
 import { FormStates } from '../form/form-control/form-states';
 import { PopoverComponent } from '../popover/popover.component';
 import { GroupFunction } from '../utils/pipes/list-group.pipe';
+import { InputGroupComponent } from '../input-group/input-group.component';
+import { KeyUtil } from '../..';
```
JKMarkowski, line 34: Let's add exact import, pathes like cause weird errors

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -128,7 +125,7 @@ export class ComboboxComponent implements ControlValueAccessor, OnInit, OnChange
 
     /** Max height of the popover. Any overflowing elements will be accessible through scrolling. */
     @Input()
-    maxHeight: string = '200px';
+    maxHeight: string = '50vh';
```
JKMarkowski, line 129: Could you add also `inputTextChange` output? It would throw `inputText` everytime it's changed. Sometimes developers refresh api calls on input change, which is not currently possible on `objectMode`.

JKMarkowski:
combobox11
What about arrowDown/arrowUp support like on Ui5?

libs/core/src/lib/combobox/combobox.component.ts
mikerodonnell89: Wouldn't they just use ngModelChange?

JKMarkowski: Nope, on `[objectMode]` ngModelChange is not triggered on text input change, it's triggered only, when the option is choosen.

InnaAtanasova:
Updating the value only from the input field does not change the value of the index. 

To reproduce it: Type Apple in the input field and select it. The selected index is 0. Type Pineapple and press enter (don't click on the menu item, just use the input field and the keyboard). The search term is updated, selected is updated, but the index is still the same. Same behaviour for Custom Item Template.

Screen Shot 2020-06-04 at 2 26 02 PM
Screen Shot 2020-06-04 at 2 25 46 PM

mikerodonnell89:
@JKMarkowski I just pushed a commit fixing some small issues:

- When autocomplete is on, partially typing a word and pressing enter should move cursor to last char and lose selection.
- console error from template example
- Autocomplete directive's function to check for control key should check for mac's Meta key too

I noticed you removed the logic of comparing the old input value and new input value before doing an autocomplete.  I actually do think we need this for a few reasons, but one example I'm coming across is this.  

- focus input and (on mac) cmd+A to select all.  Be sure to release the cmd (or ctrl) key before releasing the A key.  Because the cmd key is no longer pressed, the keyup event received by the handleKeyboardEvent function only sees this as an 'A' keyup, and typeahead is still called, moving the cursor to the last char and removing the selection.

Lastly, when autocomplete is on, if you start to type a word and it autocompletes and then you blur the input, the text in the input reverts back to simply what you had typed.  In UI5 it selects whatever had been autocompleted, and I think that is how we should do it, but let me know if you made this change intentionally

mikerodonnell89:
@JKMarkowski I just pushed a commit fixing some small issues:

- When autocomplete is on, partially typing a word and pressing enter should move cursor to last char and lose selection.
- console error from template example
- Autocomplete directive's function to check for control key should check for mac's Meta key too

I noticed you removed the logic of comparing the old input value and new input value before doing an autocomplete.  I actually do think we need this for a few reasons, but one example I'm coming across is this.  

- focus input and (on mac) cmd+A to select all.  Be sure to release the cmd (or ctrl) key before releasing the A key.  Because the cmd key is no longer pressed, the keyup event received by the handleKeyboardEvent function only sees this as an 'A' keyup, and typeahead is still called, moving the cursor to the last char and removing the selection.

Lastly, when autocomplete is on, if you start to type a word and it autocompletes and then you blur the input, the text in the input reverts back to simply what you had typed.  In UI5 it selects whatever had been autocompleted, and I think that is how we should do it, but let me know if you made this change intentionally

JKMarkowski:
Hi @mikerodonnell89, thank you for review
This cmd/crtl + a issue has been fixed. 
But about the blur/escape/backspace I think we shouldn't force users to have options from combobox. This is the way how I allow them to write anything different than we provide on options

mikerodonnell89:
> Hi @mikerodonnell89, thank you for review
> This cmd/crtl + a issue has been fixed.
> But about the blur/escape/backspace I think we shouldn't force users to have options from combobox. This is the way how I allow them to write anything different than we provide on options

I agree that we shouldn't force users to enter one of the options, but that isn't what I'm suggesting.  Take a look at ui5 combobox - https://sapui5.hana.ondemand.com/#/entity/sap.m.ComboBox/sample/sap.m.sample.ComboBox 

If you type "Alg" then it autocompletes to Algeria, and if you blur the input, Algeria is selected.  However in this PR if you type "App" it autocompletes to Apple, then when you blur, it goes back to App.  I think it should keep Apple on blur.  The user can then backspace to Appl and when they blur then, it will not go back to Apple.

mikerodonnell89:
Pushed a commit adding Meta key as an alias to Ctrl, fixes a few mac keyboard bugs here

JKMarkowski:
> > Hi @mikerodonnell89, thank you for review
> > This cmd/crtl + a issue has been fixed.
> > But about the blur/escape/backspace I think we shouldn't force users to have options from combobox. This is the way how I allow them to write anything different than we provide on options
> 
> I agree that we shouldn't force users to enter one of the options, but that isn't what I'm suggesting. Take a look at ui5 combobox - https://sapui5.hana.ondemand.com/#/entity/sap.m.ComboBox/sample/sap.m.sample.ComboBox
> 
> If you type "Alg" then it autocompletes to Algeria, and if you blur the input, Algeria is selected. However in this PR if you type "App" it autocompletes to Apple, then when you blur, it goes back to App. I think it should keep Apple on blur. The user can then backspace to Appl and when they blur then, it will not go back to Apple.

@mikerodonnell89 Let's raise it on meeting, I got no clue how it should behave

JKMarkowski:
Blurevent is fixed and now works as on ui5

apps/docs/src/app/core/component-docs/combobox/examples/combobox-dynamic-example.component.ts
```diff
@@ -12,9 +12,13 @@ export class ComboboxDynamicExampleComponent {
...
     dropdownValues = ['Apple', 'Banana', 'Kiwi', 'Strawberry', 'Tomato', 'Pineapple'];
 
-    selectItem(event: { item: any; index: number }) {
-        this.selected = event.item;
-        this.selectedIndex = event.index;
+    selectItem(event: string) {
+        this.dropdownValues.forEach((value, i) => {
+            if (value === event) {
+                this.selected = value;
+                this.selectedIndex = i;
```
rengare, line 19: add `return;` to end function earlier when the match is found

apps/docs/src/app/core/component-docs/combobox/examples/combobox-template-example.component.ts
```diff
@@ -25,7 +25,11 @@ export class ComboboxTemplateExampleComponent {
 
     selected: any;
 
+    searchTerm: string;
+
     displayFunction(item: { name: string; icon: string }): string {
-        return item.name;
+        if (item && item.name) {
+            return item.name;
+        }
```
rengare, line 33: when if statement is false, then there is no return from function

apps/docs/src/app/core/component-docs/combobox/examples/combobox-dynamic-example.component.ts
JKMarkowski: It's changed to findIndex method

PR #2484, opened 2020-05-07
fix: Add popover update, when created

JKMarkowski:
Issue is well reproduced [there](https://stackblitz.com/edit/angular-z4pvxw?file=src/app/popover-placement-example.component.html).
For some reason, popover is not updated, when it overflows the browser window.
There is forced update, on every open.

We should consider migrating to `@angular/cdk/overflow`

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


stefanoScalzo:
Approving but agree we should go over cdk

PR #2491, opened 2020-05-09
docs: enable theme selection on the doc site

droshev:
enable theme selection on the doc site


mikerodonnell89:
Yea that is a strange place for that.  I think maybe a gear icon in the top right where the github logo is

estelaxu3:
Two things I noticed on this page: 
1. The Github version text color should change accordingly in High Contrast White
2. The browser tab icon's background changes when switching the theme, I don't think it is necessary since the logo represents entire library including Styles. If we really want that, maybe we should switch back to white back ground in High Contrast White theme so it is consistent. 
Screen Shot 2020-05-12 at 10 09 53 AM

apps/docs/src/app/documentation/core-helpers/sections-toolbar/sections-toolbar.component.ts
```diff
@@ -52,4 +68,9 @@ export class SectionsToolbarComponent implements OnInit {
     onResize() {
         this.windowSize();
     }
+
+    selectTheme() {
+        const cssLink: any = document.getElementById('themes');
+        cssLink.href = 'assets/' + this.selectedTheme + '.css';
```
rengare, line 74: > cssLink.href = `assets/${this.selectedTheme}.css`

salarenko:
blinking-theme-select

estelaxu3:
The popover is has white top and button space. 
Screen Shot 2020-05-14 at 8 56 44 AM

estelaxu3:
looks good to me.

InnaAtanasova:
Screen Shot 2020-05-14 at 5 38 50 PM

Shouldn't this button be like the other buttons in the Shellbar? (the hover/active state)

PR #2511, opened 2020-05-13
fix: Create dynamic mobile view for Select component

salarenko:
This PR introduces mobile mode for Select component based on `DynamicComponentService`.
This PR also brings initial values for Dialogs Header/Body/Footer and initializes them when their are set.
For some reason if Header/Body/Footer is conditional, the order of template initialization changed, initializing before DialogCOmponent could pas Config objects to his childs.


stefanoScalzo:
Really like this behind the scenes approach. If platform can reproduce using this i like it. Good Job!

libs/platform/src/lib/components/select/select.component.ts
```diff
@@ -11,7 +11,7 @@ import {
     ElementRef
 } from '@angular/core';
 import { NG_VALUE_ACCESSOR } from '@angular/forms';
-import { SelectComponent as fdSelect, SelectProxy } from '@fundamental-ngx/core';
+import { DynamicComponentService, SelectComponent as fdSelect, SelectProxy } from '@fundamental-ngx/core';
```
fkolar, line 14: I just assume this is only for compatibility reason so it does compile as we inherit select. we need to change this part..

mikerodonnell89:
Maybe outside the scope of this PR but dialog close button should not be compact because this is intended for use in mobile mode

mikerodonnell89:
Also probably out of the scope of this PR, but I would think that there should be no border radius on the fullscreen modals to avoid this look:
Screen Shot 2020-05-18 at 12 00 09 PM

otherwise this PR looks great, clean solution

libs/core/src/lib/select/select.component.spec.ts
```diff
@@ -35,7 +35,7 @@ describe('SelectComponent', () => {
     beforeEach(async(() => {
         TestBed.configureTestingModule({
             declarations: [TestWrapperComponent],
-            imports: [SelectModule]
+            imports: [SelectModule],
```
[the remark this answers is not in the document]
salarenko: I will check it, comma removed.

libs/core/src/lib/dialog/dialog-footer/dialog-footer.component.ts
```diff
@@ -22,7 +22,9 @@ export class DialogFooterComponent implements AfterContentInit {
     /** @hidden */
     @ContentChildren(TemplateDirective) customTemplates: QueryList<TemplateDirective>;
 
-    constructor(@Optional() @Inject(DIALOG_CONFIG) public dialogConfig: DialogConfig) {}
+    constructor(@Optional() @Inject(DIALOG_CONFIG) public dialogConfig: DialogConfig) {
```
InnaAtanasova, line 25: do we need the `public` here?

InnaAtanasova:
Screen Shot 2020-05-28 at 4 49 11 PM

I don't think the state message should have any kind of interaction. In the current implementation you can click on it and it gets a border

libs/core/src/lib/dialog/dialog-footer/dialog-footer.component.ts
salarenko: Yes, this is a shortcut for creating object prop, otherwise this will be a normal constructor parameter.

salarenko:
> I don't think the state message should have any kind of interaction. In the current implementation you can click on it and it gets a border
@InnaAtanasova You are right, this styling comes from user agent. If that behavior is not desired I think we should reset `focus` from messages on `Fundamental Styles` side.

PR #2512, opened 2020-05-13
fix: add tabindex to shellbar button and user menu

### LINKED ISSUE #2448: shellbar userMenu not included in tab order
This is an **accessibility** bug

Currently, when you include a usermenu under fd-shellbar-actions in fd-shellbar, it does not have a tabindex. i.e. it can not be reached via keyboard.
However, operability via keyboard is a requirement for accessible applications.

Angular 9, fundamental-ngx 0.17.1

Include a shellbar with a usermenu and try to access the user menu via tabbing through the elements via keyboard. You can't.

It can be reproduced with your examples

The user placeholder/image should have a tabindex of zero and should react to a return (to access the menu).

mikerodonnell89:

fixes #2448 

Add tabindex to shellbar buttons.  Visual fix will come with this PR on styles: https://github.com/SAP/fundamental-styles/pull/1027


mikerodonnell89:
Buttons are indexed by default, just needs the styles PR

mikerodonnell89:
Actually we do need this but only for the user menu. We can remove tabindex="0" from shellbar action

PR #2520, opened 2020-05-15
test: Add docker to unit tests

JKMarkowski:
There is included new command for testing in docker


mikerodonnell89:
So wasn't one of the intentions of this PR to remove the variables that are causing tests to run differently on each of our machines? I am getting the same test failures.  I don't understand why these tests (breadcrumb/tokenizer) are failing for me now because they definitely worked when I wrote them

Issue #2553, opened 2020-05-22
Fix(Platform) : Rtl Bug for action bar


manu-kr:
#### Is this a bug, enhancement, or feature request?
Bug
#### Briefly describe your proposal.
Click on RTL simulate button, 
Actual: Back button is not changing its direction.
Expected: Should change to right navigation icon.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 9

#### If this is a bug, please provide steps for reproducing it.

Click on RTL simulate button, 
Actual: Back button is not changing its direction.
Expected: Should change to right navigation icon.

PR #2554, opened 2020-05-22
Fix(Platform): RTL Action bar back button.

manu-kr:

https://github.com/SAP/fundamental-ngx/issues/2553
This PR contains fix for rtl issue of action bar back button.

Documentation checklist:
- [x] Documentation Examples

### DIFF
--- a/apps/docs/src/app/core/component-docs/action-bar/action-bar-docs.module.ts
+++ b/apps/docs/src/app/core/component-docs/action-bar/action-bar-docs.module.ts
@@ -15,6 +15,7 @@ import { API_FILES } from '../../api-files';
 import { SharedDocumentationModule } from '../../../documentation/shared-documentation.module';
 import { ActionBarModule } from '@fundamental-ngx/core';
 
+
 const routes: Routes = [
     {
         path: '',
@@ -37,6 +38,7 @@ const routes: Routes = [
         ActionBarNoBackExampleComponent,
         ActionBarContextualMenuExampleComponent,
         ActionBarLongStringTitleTruncationExampleComponent
+
     ]
 })
-export class ActionBarDocsModule {}
+export class ActionBarDocsModule { }
--- a/apps/docs/src/app/platform/component-docs/platform-action-bar/platform-action-bar-examples/platform-action-bar-positive-and-negative-action-example.component.html
+++ b/apps/docs/src/app/platform/component-docs/platform-action-bar/platform-action-bar-examples/platform-action-bar-positive-and-negative-action-example.component.html
@@ -1,5 +1,6 @@
 <fdp-action-bar
     id="actionbar_with_menu"
+    [contentDensity]="'compact'"
     [title]="'Page Action bar component'"
     [showBackButton]="true"
     [description]="'Action bar description from the component.'"
--- a/libs/platform/src/lib/components/action-bar/action-bar.component.html
+++ b/libs/platform/src/lib/components/action-bar/action-bar.component.html
@@ -2,13 +2,13 @@
     <div fd-action-bar-header>
         <div fd-action-bar-back *ngIf="showBackButton">
             <button
-                class="sap-icon--navigation-left-arrow"
                 fd-button
                 [compact]="compact"
                 [fdType]="'transparent'"
                 [attr.aria-label]="backButtonLabel"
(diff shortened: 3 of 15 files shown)


libs/platform/src/lib/components/action-bar/action-bar.component.ts
```diff
@@ -28,7 +31,7 @@ export class ActionBarComponent extends BaseComponent implements OnInit {
     /**
      * Compact mode
      */
-    compact = true;
+    compact = false;
```
fkolar, line 34: Is this some left over. You dont need this one, and you are not going to save anything using `contentDensity === 'compact'` in the template vs saving it to extra property.

libs/platform/src/lib/components/action-bar/action-bar.component.html
```diff
@@ -2,13 +2,13 @@
...
         <div fd-action-bar-back *ngIf="showBackButton">
             <button
-                class="sap-icon--navigation-left-arrow"
                 fd-button
                 [compact]="compact"
                 [fdType]="'transparent'"
                 [attr.aria-label]="backButtonLabel"
                 data-tag="actionbar__back-button"
                 [attr.title]="backButtonLabel"
+                [glyph]="navigationArrow$ | async"
```
fkolar, line 11: What about case where navigation arrow is null so you there is no icon? you want to have this case too.

apps/docs/src/app/platform/component-docs/platform-action-bar/platform-action-bar-examples/platform-action-bar-positive-and-negative-action-example.component.html
```diff
@@ -1,5 +1,6 @@
 <fdp-action-bar
     id="actionbar_with_menu"
+    [contentDensity]="'compact'"
```
fkolar, line 3: you have everywhere contentDensity=compact, you might want to have some examples with cozi, or have some dropdown to pick

manu-kr: added example.

libs/platform/src/lib/components/action-bar/action-bar.component.ts
manu-kr: added it in template removed 'compact' property.

libs/platform/src/lib/components/action-bar/action-bar.component.html
manu-kr: Without back button case is already impleanted with showbackbutton flag in action if this flag is false it won't load back button element.

libs/platform/src/lib/components/action-bar/action-bar.component.ts
```diff
@@ -48,13 +46,17 @@ export class ActionBarComponent extends BaseComponent implements OnInit {
...
         super(_cd);
     }
 
-    ngOnInit() {
-        if (this.contentDensity === 'cozy') {
-            this.compact = false;
-        }
+    ngOnInit(): void {
+
+        this.navigationArrow$ = this._rtlService.rtl.pipe(
```
fkolar, line 57: I am thinking if its not more effective to use angular/cdk the https://material.angular.io/cdk/bidi/api, I am using it in select, but this is just a question. You can definitely stick with this RTL service. I just dont like the observable approach. You either load this page within RTL lang or dont.

apps/docs/src/app/core/component-docs/action-bar/action-bar-docs.module.ts
```diff
@@ -15,6 +15,7 @@ import { API_FILES } from '../../api-files';
 import { SharedDocumentationModule } from '../../../documentation/shared-documentation.module';
 import { ActionBarModule } from '@fundamental-ngx/core';
```
fkolar, line 16: What about documentation page, do you have the sample in github what text do we have there? I think since you have it opened we can update this part too. What do you think?

manu-kr: Shall i take this documentation as a separate Task?

fkolar:
I am not much sure about the english wordings, but i guess we can address this with some technical writter

PR #2575, opened 2020-05-28
fix: only display "show all" text when options have been filtered

### LINKED ISSUE #2574: Multi-input: "Show All" text should only display when list is filtered

Bug

If there is no text in the input and the full list of options is displayed, the "Show All" text should not be displayed

mikerodonnell89:

fixes #2574 

only display "show all" text when options have been filtered with text in the input


InnaAtanasova:
Screen Shot 2020-05-28 at 4 26 04 PM

How would the button look like in mobile?

libs/core/src/lib/multi-input/multi-input.component.spec.ts
```diff
@@ -241,4 +241,21 @@ describe('MultiInputComponent', () => {
         expect(component.selectedChange.emit).toHaveBeenCalled();
         expect(component.selected).toEqual([component.dropdownValues[0]]);
     });
+
+    it('should handle showAll button', async () => {
```
salarenko, line 245: What this test do is checking if calling `showAllClicked` method clears `searchTerm`.
Well this test is actually needed, because I cannot find any tests for `Show All Button`, but what new functionality actually does is Showing/Hiding `Show All` button depending on filter state.

libs/core/src/lib/multi-input/multi-input.component.spec.ts
```diff
@@ -241,4 +241,21 @@ describe('MultiInputComponent', () => {
...
+        await fixture.whenStable();
+        const event: any = {
```
salarenko, line 247: Go typesafe, use `new MouseEvent()`

libs/core/src/lib/multi-input/multi-input.component.spec.ts
```diff
@@ -241,4 +241,21 @@ describe('MultiInputComponent', () => {
...
+            stopPropagation: () => {}
+        };
+        component.searchTerm = 'term';
+        spyOn(event, 'preventDefault');
+        spyOn(event, 'stopPropagation');
+        spyOn(component, 'handleSearchTermChange');
+        component.showAllClicked(event);
+        expect(event.preventDefault).toHaveBeenCalled();
+        expect(event.stopPropagation).toHaveBeenCalled();
+        expect(component.searchTerm).toBe('');
```
JKMarkowski, line 259: You can also check if `displayedValues.length < dropdownValues.length` is false

mikerodonnell89:
Not sure what to do about mobile mode, but I'm not going to deal with that in this pr

mikerodonnell89:
I've asked the designer about what to do for mobile

mikerodonnell89:
Addressed PR comments and added "___ more" text to thumb-swipe-able cozy tokenizer/multi-input @JKMarkowski

apps/docs/src/app/core/component-docs/multi-input/multi-input-docs.component.html
```diff
@@ -10,7 +10,7 @@
     <fd-multi-input-example></fd-multi-input-example>
 </component-example>
 <code-example [exampleFiles]="multiInputBasic"></code-example>
-
+<!--
```
JKMarkowski, line 13: Whole docs page is commented

stefanoScalzo:
> Let's remove the comments from code and check sapui5 behaviour

comments agreed, but ui5s behavior is buggy and doens't make sense

mikerodonnell89:
> Let's remove the comments from code and check sapui5 behaviour

@JKMarkowski This PR was originally only intended to address the "show all" issue and we already tacked on the cozy mode "more" text thing.  For the sake of keeping PRs small and not having them open for months on end, let's address any missing ui5 features in other issues

PR #2612, opened 2020-06-04
feat: refactor old Panel component to Layout Panel and introduce Fiori 3 Panel

### LINKED ISSUE #2599: Fiori 3 Panel implementation in Fundamental-NGX
The Panel component has been refactored in Fundamental-styles. The "old" panel is now called Layout Panel and a new Panel that follows Fiori 3 guidelines has been introduced.

InnaAtanasova:
fixes #2599 

The Panel component has been refactored in Fundamental-styles. The "old" panel is now called Layout Panel and a new Panel that follows Fiori 3 guidelines has been introduced.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


JKMarkowski:
Could you add the way to provide some `aria-controls` to button inside `panel-expand` component?
Also could you put into examples a11y things like `role="region" aria-labelledby="jhqDKYdf" aria-hidden="true" id="fghqwe321"` to `<div fd-panel-content>`

libs/core/src/lib/layout-panel/layout-panel.component.ts
```diff
@@ -0,0 +1,37 @@
...
+    selector: 'fd-layout-panel',
+    templateUrl: './layout-panel.component.html',
+    host: {
+        '[class.fd-has-display-block]': 'true'
+    },
+    encapsulation: ViewEncapsulation.None,
+    styleUrls: ['./layout-panel.component.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class LayoutPanelComponent extends AbstractFdNgxClass {
```
JKMarkowski, line 17: Let's not use `AbstractFdNgxClass` ( I know i was there before, but please replace it by Gracjan's approach with `CssClassBuilder`)

libs/core/src/lib/panel/panel-header/panel-header.component.ts
```diff
@@ -1,26 +1,57 @@
...
-    fdPanelHeaderClass = true;
+    readonly fdPanelHeaderClass: boolean = true;
+
+    /** 
+     * An event emitted to the parent PaneComponent 
+     * with a boolean value captured from the child PanelExpandComponent 
+     * that controls the expanded state of the Panel Content.
+     */
+    @Output()
+    expandedValue: EventEmitter<boolean> = new EventEmitter<boolean>();
```
JKMarkowski, line 44: Could you change this event's naming to `expandedChange`

JKMarkowski: Also let's have `expanded` flag instead of `isExpanded`. Compiled with `expandedChange` user will be able to use `[(expanded)]` binding, to change it programatically and react on changes.

libs/core/src/lib/layout-panel/layout-panel.component.ts
InnaAtanasova: I have just renamed this component :) I haven't refactored it

libs/core/src/lib/panel/panel-header/panel-header.component.ts
InnaAtanasova: I will. That's a good idea. Thank you!

libs/core/src/lib/panel/panel-header/panel-header.component.ts
```diff
@@ -1,26 +1,66 @@
...
+    @ContentChild(PanelExpandComponent) panelExpand: PanelExpandComponent;
+
+    /** @hidden */
+    ngAfterViewInit(): void {
+        if (this.panelExpand) {
+            this.panelExpand.expandedChange.subscribe(
+                (value: boolean) => this.expandedChange.emit(value)
+            );
+        }
+    }
```
JKMarkowski, line 65: Could you put it to `ngAfterContentInit`, it's not big deal, but we pick content reference there. 
Also let's make some guard for subscription,

libs/core/src/lib/panel/panel.component.ts
```diff
@@ -1,37 +1,106 @@
...
+        this.buildComponentCssClass();
+        this._subscription = this._panelService.expanded$.subscribe((value) => {
+            this.expanded = value;
+            this._cdRef.detectChanges();
+        });
+    }
+
+    /** @hidden */
+    ngOnChanges(changes: SimpleChanges): void {
+        this.buildComponentCssClass();
```
salarenko, line 76: This should be called only on `[class]` change

libs/core/src/lib/panel/panel.component.ts
```diff
@@ -1,37 +1,106 @@
...
+        this._panelService.updateExpanded(this.expanded);
```
salarenko, line 77: Double `this._panelService.updateExpanded(this.expanded);` calls

Issue #2598, opened 2020-06-02
Cyclic dependency in components MobileModes


salarenko:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
Current MobileMode approach implemented in Select, Multi-Input, Menu brings cyclic dependency between mobile and desktop version of the component.
We can create injection tokens and interfaces to escape cyclic dependency.

salarenko:
Issue has been resolved with
#2619 
#2696
#2707

PR #2619, opened 2020-06-05
feat(select): Complete refactoring based on material common behavior

fkolar:
This PR provides complete refactored code of Select Component, based on the materials. This also showcase how to isolate some reusable logic into mixins in order to differentiate between our internal code and Material and try to reuse as much as we can from well know established library. 

Also one of the thing that can be seen and its pretty common is to make sure you do proper Attribution inside your code, instead of mixings outside code like I have seen without proper credit to the library author. 

This PR also is fully based on CDK overlays and removing dependency to popover. 

Refactored and updated test as there were a few old test that were pretty much redudant and either tested obvious things and redundant case whcih was tested in different way somewhere else. 

As part of this PR I completely wiped out Platform select as it needs to be done. 

**BREAKING CHANGE:**
This PR also introduces **BREAKING CHANGE**  as binding `[closeOnOutsideClick]`  was removed. CDK handles all this natively. This binding was specific to Popover functionality

 This PR also fixes these two issues #2598 

@droshev should I as part of this push also update this Braking changes wiki? or this is when when we release

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


fkolar:
Let's wait once all the checks and builds finished, and you can start looking at this, or at least dont comment if any CI fails. I am monitoring it.

libs/core/src/lib/select/option/option.component.ts
```diff
@@ -1,22 +1,61 @@
...
+/**
+ * Event object emitted by OptionComponent when selected or deselected. *
+ */
+export class FdOptionSelectionChange {
+    constructor(
+        /** Reference to the option that emitted the event. */
+        public source: OptionComponent,
+        /** Whether the change in the option's value was a result of a user action. */
+        public isUserInput = false) {
+    }
```
JKMarkowski, line 32: What do you think about replacing `public` to `readonly`?

libs/core/src/lib/select/option/option.component.ts
```diff
@@ -29,9 +68,13 @@ let optionUniqueId: number = 0;
         'class': 'fd-list__item',
         '[attr.aria-disabled]': 'disabled',
         '[attr.aria-selected]': 'selected',
+        '[class.is-hover]': 'active',
```
JKMarkowski, line 71: Shouldn't we add `is-focus` instead of hover?

libs/core/src/lib/select/option/option.component.ts
```diff
@@ -47,7 +90,7 @@ let optionUniqueId: number = 0;
     `]
 
 })
-export class OptionComponent implements OnInit, OnDestroy {
+export class OptionComponent implements FocusableOption, OnDestroy, AfterViewChecked {
```
JKMarkowski, line 93: Let's follow [contribution guidelines](https://github.com/SAP/fundamental-ngx/wiki/Code-Contribution-Guideline) on order of declarations

libs/core/src/lib/select/option/option.component.ts
```diff
@@ -61,81 +104,150 @@ export class OptionComponent implements OnInit, OnDestroy {
...
+    }
+
+
+    getLabel(): string {
+        return this.viewValue;
+    }
+
+
+    /** Ensures the option is selected when activated from the keyboard. */
+    _handleKeydown(event: KeyboardEvent): void {
```
JKMarkowski, line 211: What about the `_` prefix? We decided to reserve it only for private properties/methods

libs/core/src/lib/select/select.component.ts
```diff
@@ -53,284 +87,288 @@ export interface OptionStatusChange {
...
-    }
 
     constructor(
-        private _elementRef: ElementRef,
-        private _selectProxy: SelectProxy,
-        private _changeDetectorRef: ChangeDetectorRef,
+        public _viewportRuler: ViewportRuler,
+        public  _elementRef: ElementRef,
+        public  _ngZone: NgZone,
+        public  _changeDetectorRef: ChangeDetectorRef,
```
JKMarkowski, line 305: Do you add underscore only to implement common interface?

libs/core/src/lib/select/commons/mat-select-commons.ts
```diff
@@ -0,0 +1,917 @@
...
+                    this._keyManager.setFirstItemActive();
+                } else {
+                    this._keyManager.setActiveItem(this.selected);
+                }
+            }
+        }
+
+        /** Handles keyboard events while the select is closed. */
+        private _handleClosedKeydown(event: KeyboardEvent): void {
+            const keyCode = event.keyCode;
```
salarenko, line 852: Should not rely on `keyCode`

libs/core/src/lib/select/commons/mat-select-commons.ts
```diff
@@ -0,0 +1,917 @@
...
+                event.preventDefault();
+                this.close();
+                // Don't do anything in this case if the user is typing,
+                // because the typing sequence can include the space key.
+            } else if (!isTyping && (keyCode === ENTER || keyCode === SPACE) && manager.activeItem &&
+                !hasModifierKey(event)) {
+                event.preventDefault();
+                manager.activeItem._selectViaInteraction();
+            } else {
+                const previouslyFocusedIndex = manager.activeItemIndex;
```
salarenko, line 908: Leftover variable?

libs/core/src/lib/select/commons/mat-select-commons.ts
```diff
@@ -0,0 +1,917 @@
...
+            const options = this.options;
+
+            if (options) {
+                return options.changes.pipe(
+                    startWith(options),
+                    switchMap(() => merge(...options.map(option => option.selectionChange)))
+                );
+            }
+
+            return this._ngZone.onStable
```
salarenko: Could you explain what is this doing and why its needed?

fkolar:
> Source files were not formatted according to `.prettierrc`

I will check and update, but we dont have pretty detailed this definition file: 

```
{
    "printWidth": 120,
    "singleQuote": true,
    "useTabs": false,
    "tabWidth": 4,
    "semi": true,
    "bracketSpacing": true,
    "trailingComma": "none"
}
```

What we should do so to create CodeStyles files for each IDE (VS Code, WebStorm), its not only about file, its more how classes should be wrapped or comments, how import should be organized, different spacing for different sitation, blank lines, ... etc.

fkolar:
@rengare can you point me to some files which are not complient with prettierc? 
If I do format accourding to prettier I do have some changes files, but this is just each IDE has some also default rules for this utility. But those 7 rules we I should be ok except one where i see different bracket spacing.

libs/core/src/lib/select/option/option.component.ts
fkolar: @JKMarkowski This is something that should be comming from styles. You dont want to re-define classes in your component. I would like is-`active`

libs/core/src/lib/select/option/option.component.ts
fkolar: @JKMarkowski I must have missed this WIKI page, we should probably put this into PR template, we refer only to MD files. But you are right I see in the COntributing.md

libs/core/src/lib/select/option/option.component.ts
fkolar: @JKMarkowski  This just means that this method is not public and its more private API but since you want this to be called from outside, or template, its hybrid. `_` you tell its private API but you make it public as you want to access it. 

This is common practice. also in other opensource project, Angular.io, material.

fkolar:
@JKMarkowski Strange, I just saw your comments, but haven't received any emails

fkolar:
@rengare I pushed updated 1 commit, just to showcase, how different it could be. If you checkout e.g. HTML - select-nested-options.component.ts, how default prettier settings adds semi-colon for the fruit definition - it can be tricky. 




> IE11
> image
> 
> I don't know what to think about this PR. the code here looks great, basing on material solution is also good, but I don't think that defining `Mat` prefixed classes is ok.
> It's step forward in replacing popper, which actually we should propagate to all of the components. This solution is so complex only for select (much more complexity, without flexibility), In future I would share this kind of functionality(keyboard support/overflow cdk) to other menu/list/dropdown components. I would be great, to have it built-in in our new popover.
> 
> There are some minor things I raised, you can take a look at them.

@JKMarkowski The take away from this paragrams is to rename mat-xxx to xxx?  I have received some feedback like this already. more voices, I will just rename this one.

libs/core/src/lib/select/option/option.component.ts
fkolar: make sense we are not going to change it

libs/core/src/lib/select/option/option.component.ts
fkolar: I will mark this as resolved as I work only with avaialbe classes to make it active. We should have this active style in the styles, then we can replace it.

libs/core/src/lib/select/option/option.component.ts
fkolar: But In this specifc case it makes sense as we dont reference templates

libs/core/src/lib/select/select.component.ts
fkolar: This is the trouble with mixins. Again:
  - You want to make it accesible within the class
 - But you want other aware this is private API/Attribute dont use it. 

@JKMarkowski do you recomend other way to go around it? This is related only to those properties inside Mixins Interfaces.

libs/core/src/lib/select/select.component.ts
```diff
@@ -1,50 +1,84 @@
...
+
+export class MatSelectCommons {
+    constructor(
+        public _viewportRuler: ViewportRuler,
+        public  _elementRef: ElementRef,
+        public  _ngZone: NgZone,
+        public  _changeDetectorRef: ChangeDetectorRef,
+        public  _dir: Directionality,
+        public ngControl: NgControl,
+        public _liveAnnouncer: LiveAnnouncer
```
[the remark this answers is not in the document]
fkolar: As already explained above to @JKMarkowski and we should have a way dealing with this as its common in material or Angular itself. 

prefixing _xxxx even its public, it has several reasons:
  - Its private method you dont want others ot use it but its either used called from template
  - Its part of interface that you need to be part of it  - in this case mixins, and here you needs to be modifier public, but on the other hand, you are telling dont use them on I am part of private API. 

@rengare We shoudl have solution for this. It not really one way or the other.

libs/core/src/lib/select/commons/mat-select-commons.ts
```diff
@@ -0,0 +1,917 @@
...
+    _elementRef: ElementRef;
+    _ngZone: NgZone;
+    _changeDetectorRef: ChangeDetectorRef;
+    _dir: Directionality;
+    ngControl: NgControl
+    _liveAnnouncer: LiveAnnouncer;
+}
+
+
+export function mixinMatSelectCommons<T extends Constructor<HasMatCommonBehavior>>(base: T)
```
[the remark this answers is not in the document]
fkolar: This is more openned questions, can you expand? This code contains code from material, its not really part of the function its inthe unnamed class, this function is just wrapping it so you can plug it in as mixins. 

@rengare Would use say that any other component we implement e.g. compobox has much code inside it  (allot of responsiblity) , does it mean we need to change it?

libs/core/src/lib/select/select-mobile/select-mobile/select-mobile.component.html
```diff
@@ -4,7 +4,7 @@
         <fd-dialog-header>
             <h1 *ngIf="this.mobileConfig.title" fd-dialog-title>{{this.mobileConfig.title}}</h1>
             <button *ngIf="this.mobileConfig.hasCloseButton"
-                    fd-dialog-close-button
+                    fd-dialog-close-button tabindex="0"
```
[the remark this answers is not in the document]
fkolar: Great comments. I just hope I will not get lost as there so many bullet points ..

**Read-only select is selectable.**
   -  updated (waiting for push to resolve other comments, )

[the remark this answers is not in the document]
fkolar: **When select is closed, should control value be changed**:
 - Can you reference your source - not recomended? When you work with native select this is common behavior and this something user expects. Its easy to disable this, but are @salarenko  we sure about it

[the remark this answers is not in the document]
fkolar: **Select uses hover styling for marking options list element marked as focused/navigated to when using ArrowUp, ArrowDown. This also comes from Material design but all a11y specifications for select say about "focusing" list elements. In our case select input seems to be the focused element all the time.**  
- @salarenko I am not accesiblity expert but in Fiori 3 it should not behave like this? In original implementation we apply styles on active item when you mouse over it, but with keyboard navigation we did not do it. In here we were not consistent. 
 - I have my opinion about this one as I feel we had inconsistency before and this just aligns both mouse hover and active items from keyboard navigation which is the same thing

@droshev @rengare how we go about this? What do you think ?

libs/core/src/lib/select/option/option.component.ts
fkolar: Addressed

libs/core/src/lib/select/option/option.component.ts
fkolar: Addressed

libs/core/src/lib/select/option/option.component.ts
fkolar: All updated to much our guidelines,

tslint.json
```diff
@@ -49,7 +49,7 @@
         "prefer-const": true,
         "prefer-for-of": false,
         "quotemark": [true, "single"],
-        "radix": true,
+        "radix": false,
```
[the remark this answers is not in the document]
fkolar: updated and using longer version now

libs/core/src/lib/select/commons/mat-select-commons.ts
fkolar: Yes. thanks.

libs/core/src/lib/select/select.component.html
```diff
@@ -1,85 +1,99 @@
...
+         [attr.id]="controlId"
+         [attr.aria-active]="panelOpen"
+         [attr.aria-expanded]="panelOpen"
+         [attr.aria-disabled]="disabled"
+         [attr.aria-readonly]="readonly"
+         [attr.aria-controls]="controlId + '-list-box'"
+         [attr.aria-haspopup]="!(this.readonly || this.disabled)"
+         [attr.aria-label]="ariaLabel"
+         [attr.aria-labelledby]="ariaLabelledBy"
+         [attr.aria-activedescendant]="_getAriaActiveDescendant()"
```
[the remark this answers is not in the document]
fkolar: Leaving this for now as I check our WCAG guy that it should be there, but we can change this one anytime

libs/core/src/lib/select/select.component.scss
```diff
@@ -1,13 +1,14 @@
...
-    & > fd-popover {
-        width: 100%;
-    }
 
-    & ul[fd-list] {
-        max-width: 100%;
-        overflow: auto;
-    }
+
+ul[fd-list] {
```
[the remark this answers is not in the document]
fkolar: Updated to specific styles for the overlay part. Cant use fd-select as part with overlay. they are two classes here

libs/core/src/lib/select/select.component.ts
```diff
@@ -53,284 +87,288 @@ export interface OptionStatusChange {
...
-    state: SelectControlState = null;
+    get disabled(): boolean {
+        return this._disabled;
+    }
 
-    /** Whether the select component should be displayed in mobile mode. */
-    @Input()
-    mobile: boolean = false;
+    set disabled(newValue: boolean) {
+        if (newValue !== this._disabled) {
```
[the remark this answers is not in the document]
fkolar: removed setter/getter

libs/core/src/lib/select/select.component.html
```diff
@@ -1,85 +1,99 @@
...
+    >
+
+        <div class="fd-select__content">
+            <ng-container
+                [ngTemplateOutlet]="controlTemplate ? controlTemplate : defaultSelectContent"
+                [ngTemplateOutletContext]="{$implicit: triggerValue, selected: selected}">
+            </ng-container>
+
+            <ng-template #defaultSelectContent>
+                {{triggerValue || '\u00A0'}}
```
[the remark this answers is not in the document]
fkolar: added to the triggerValue

libs/core/src/lib/select/commons/mat-select-commons.ts
salarenko: We are slowly replacing our `keyCode` usages with `KeyUtil.isKey` utility, but I think we already talked about this during last call.

stefanoScalzo:
The first option is staying on hover state even when hovering other items

stefanoScalzo:
Can you please remove the refactors from the examples and all other files from this pr please

libs/core/src/lib/select/commons/mat-select-commons.ts
fkolar: let me double check.. i have replaced all keycodes and updated KeyUtils

fkolar:
> Can you please remove the refactors from the examples and all other files from this pr please

@stefanoScalzo  You need to be more specifics. I have no idea what are you refering to ?

mikerodonnell89:
Hover and focus styles should be different: this can be tested by hovering an item and then using the keyboard to focus a different item in the list.  In this PR, you get two list items that appear hovered: 

Screen Shot 2020-06-15 at 1 35 22 PM

However if you take a look at the current select on master, the hovered item has the darker background and the focused item gets an outline.

Screen Shot 2020-06-15 at 1 40 59 PM

fkolar:
> Hover and focus styles should be different: this can be tested by hovering an item and then using the keyboard to focus a different item in the list. In this PR, you get two list items that appear hovered:
> 
> Screen Shot 2020-06-15 at 1 35 22 PM
> 
> However if you take a look at the current select on master, the hovered item has the darker background and the focused item gets an outline.
> 
> Screen Shot 2020-06-15 at 1 40 59 PM

@mikerodonnell89 This is already commented above. But in fundamentals we dont have any styling for it and its missing. thsi is why we should go again and once the class exists in styles we can update this one. 

Basically, when you mouse over or you do Arrow UP and Arrow Down, now its hte same style..  I needed a style for active selected item.

mikerodonnell89:
@fkolar it is the same class name but mouse should have the additional `:hover` element state selector and keyboard should use `:focus`

fkolar:
> @fkolar it is the same class name but mouse should have the additional `:hover` element state selector and keyboard should use `:focus`

@mikerodonnell89 I found there is a class .is-focus for the the keyboard navigation. Chaning it to this one, but i must say the other class was better. or we need to have different style for. `:focus` or active item. it is bearly visible. If I would have strong glasses and using ARROWup & ARROWdown you cant even it see it.

libs/core/src/lib/select/option/option.component.ts
```diff
@@ -61,81 +110,149 @@ export class OptionComponent implements OnInit, OnDestroy {
     @Input()
     disabled: boolean = false;
 
+    @Output()
```
salarenko, line 113: This needs a comment for documentation.

libs/core/src/lib/select/option/option.component.ts
salarenko: Did you guys agreed to stick to the guidelines (`_` prefix only for private properties) or we allow to use it for internal public properties which should not be used by the user?
Asking because `_handleKeydown` still has prefix.

libs/core/src/lib/select/public_api.ts
```diff
@@ -1,5 +1,5 @@
 export * from './select.module';
 export * from './select.component';
-export * from './select-proxy.service';
+export * from './option/option.component';
```
salarenko, line 3: Doubled `export * from './option/option.component';` export

libs/core/src/lib/select/select-mobile/select-mobile/select-mobile.component.ts
```diff
@@ -1,41 +1,67 @@
...
 
+    private _activeItemIndex: number;
+
     constructor(
         private _elementRef: ElementRef,
         private _dialogService: DialogService,
-        private _selectComponent: SelectComponent
-    ) { }
+        private _injector: Injector,
+        @Optional() @Inject(FD_OPTION_PARENT_COMPONENT) private _parent: FdOptionParentComponent
```
salarenko, line 51: - Why is this `@Optional`? I don't think SelectMobile view should be ever used without parent dependency. Also implementation of SelectMobile component doesn't expect `_parent` to be missing.
- It looks like `private _injector: Injector` is also not needed.

libs/core/src/lib/select/select-mobile/select-mobile/select-mobile.component.ts
```diff
@@ -52,41 +78,33 @@ export class SelectMobileComponent implements OnInit, AfterViewInit, OnDestroy {
...
 
-    /** @hidden */
-    cancel(): void {
-        this._selectComponent.setSelectedOption({
-            option: this._selectedOnOpen,
-            controlChange: !!this._selectedOnOpen
-        });
-        this._selectComponent.close();
+    get isDismissable(): boolean {
+        return !!!this.mobileConfig.approveButtonText;
```
salarenko, line 85: Triple negation? 🤔

libs/core/src/lib/select/select.component.ts
```diff
@@ -1,336 +1,347 @@
...
 
-    /** @hidden */
-    calculatedMaxHeight: number;
-
-    /** Current selected option component reference. */
-    selected: OptionComponent;
-
-    /** Text value displayed in select control */
-    selectViewValue: string;
+    ////////////////////////////////////////////
```
salarenko, line 265: I don't think we need that separator comment

libs/core/src/lib/select/select.component.html
```diff
@@ -1,95 +1,113 @@
...
-                    [glyph]="glyph">
-            </button>
+            <ng-template #defaultSelectContent>
+                {{ triggerValue }}
+            </ng-template>
         </div>
+        <button
+            fd-button
+            tabindex="-1"
+            class="fd-button--transparent fd-select__button"
```
JKMarkowski, line 80: Shouldn't we have button input to deal with `--transparent`?

libs/core/src/lib/select/option/option.component.ts
fkolar: @JKMarkowski @salarenko this one seems like leftover, I replaced all the public that are shareble and those which does kept private. not this one.

apps/docs/src/app/core/component-docs/select/examples/select-adding-example/select-adding-example.component.html
```diff
@@ -3,7 +3,7 @@
     <button fd-button (click)="removeOption()">Remove Option</button>
 </div>
 
-<fd-select placeholder="Select an option" [(value)]="selectedValue" [closeOnOutsideClick]="false">
-    <fd-option *ngFor="let option of options" [value]="option">{{option}}</fd-option>
+<fd-select placeholder="Select an option" [(value)]="selectedValue">
```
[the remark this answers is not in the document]
fkolar: @JKMarkowski This is one of the thing why this is braking change. this close on click was feature of popover as I saw. and Select handles this naturally. Then it would not be really a select. I wanted to do also little cleanup, fresh start.

fkolar:
@salarenko , @JKMarkowski getting close guys let's give it last check and let merge it.

libs/core/src/lib/select/option/option.component.ts
fkolar: @salarenko This was the specific thing that I documented in the documentation as difference between previous version current version..

libs/core/src/lib/select/select-mobile/select-mobile/select-mobile.component.ts
fkolar: > Should we base `isDismissable` on the fact that approve button text is set?
@JKMarkowski 
The was the previous behavior was based on this, but I will rename it to Ccancalable not to confuse you from original naming

fkolar: Even cancalable sounds wierds. let;s call it hasApproveButton

libs/core/src/lib/select/select-mobile/select-mobile/select-mobile.component.ts
```diff
@@ -52,41 +78,33 @@ export class SelectMobileComponent implements OnInit, AfterViewInit, OnDestroy {
...
 
     /** @hidden */
-    approve(): void {
-        if (this._selectComponent.selected) {
-            this._selectComponent.setSelectedOption({
-                option: this._selectComponent.selected,
-                controlChange: true
-            }, true);
+    cancel(): void {
+        if (this._parent.keyManager.activeItem && !this.isDismissable) {
```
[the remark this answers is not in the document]
fkolar: @JKMarkowski its more !(!!XX)  - you want to use !! 

But here I did completly different solution. We dont need isCancelable or isDismissable now

libs/core/src/lib/select/select.component.html
fkolar: @JKMarkowski  I will resolve this one as I cant say much to it. This something you guys can address later if you decide if the button trigger needs different styling

libs/core/src/lib/select/public_api.ts
fkolar: Too many updates with this change

PR #2649, opened 2020-06-09
test: (platform) Adding tests for FormGroup/FormField and fixing tests for ActionBar and RadioGroup

KevinOkamoto:

N/A

The FormGroup and FormField component code currently have no unit tests. I wanted to create an initial unit test spec with a test component set-up and a few basic tests.

While creating the unit tests for FormGroup and FormField, I found some broken tests in the platform code - namely the ActionBar and RadioGroup components.

_ActionBar_

The ActionBar component tests were failing because some of the query selectors were returning undefined. The selectors were based off of Fundamental-Style CSS class names. However, it relied on the assumption that these class names would not change within the core component templates.

I changed the query selectors to be based of custom `data-*` attributes which are defined in the ActionBar component template. In this way the query definition and selection are contained entirely in the platform code and will not be affected by changes in the core code.

_RadioGroup_

The RadioGroup tests were complaining:
```
WARN: 'Can't bind to 'list' since it isn't a known property of 'fdp-radio-group'.'
```
RadioGroup inherits from the CollectionBaseInput class. As per this issue:

https://github.com/angular/angular/issues/32428

CollectionBaseInput needs to be annotated with `@Directive()` in order to serve as a component base class.

_fd-typedoc Helper Function_

Because the `@Directive()` annotation was added to the CollectionBaseInput without a `selector`, the documentation helper function `parseSelector` was failing because it was expecting a string value as input. I had to add a string guard to the `parseSelector` function in order for the documentation to build.

### DIFF
--- a/libs/platform/src/lib/components/form/form.spec.ts
+++ b/libs/platform/src/lib/components/form/form.spec.ts
@@ -1,0 +1,286 @@
+import { Component, ViewChild, ElementRef } from '@angular/core';
+import { async, TestBed, ComponentFixture } from '@angular/core/testing';
+
+import { FdpFormGroupModule } from './form-group/fdp-form.module';
+import { FdpInputModule } from './input/fdp-input.module';
+import { ReactiveFormsModule, FormGroup } from '@angular/forms';
+import { FormFieldComponent } from './form-group/form-field/form-field.component';
+import { FormGroupComponent } from '@fundamental-ngx/core';
+import { By } from '@angular/platform-browser';
+
+interface TestUser {
+    firstName: string;
+    lastName: string;
+    favoriteColor: string;
+    street?: string;
+    city?: string;
+    state?: string;
+}
+
+@Component({
+    template: `
+    <form [formGroup]="userFormGroup" (ngSubmit)="onSubmit($event)">
+        <fdp-form-group #userForm
+            [object]="user"
+            [formGroup]="userFormGroup"
+            [hintPlacement]="'right'"
+            [i18Strings]="i18n">
+            <fdp-form-field #firstName
+                id="firstName"
+                label="First Name"
+                hint="Enter your first name."
+                zone="zTop"
+                required=true>
+                <fdp-input [formControl]="firstName.formControl"></fdp-input>
+            </fdp-form-field>
+            <fdp-form-field #lastName
+                id="lastName"
+                label="Last Name"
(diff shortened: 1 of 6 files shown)


libs/platform/src/lib/components/form/form.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+
+    public result: any = null;
+
+    onSubmit() {
+        this.result = this.userFormGroup.value;
+    }
+
+}
+
+describe('Nested Form Groups', () => {
```
fkolar, line 234: You can better structure test so you can have describe on the main level 
```
 describe ("Form Group"
  
    describe ("core behavior"
    
    );

    describe ("layout"    
    );
  
   

    describe ("Form Fields"
        describe ("Form controls"
        )

        describe ("so on.. "
        )
     );
)
```
It is always easier to follow.

libs/platform/src/lib/components/form/radio-group/radio-group.component.spec.ts
```diff
@@ -183,7 +183,7 @@ describe('GroupRadioButtonComponent', () => {
         await wait(fixture);
         fixture.detectChanges();
 
-        const fdradiobuttonsElem = fixture.debugElement.queryAll(By.css('fd-radio-button'));
+        const fdradiobuttonsElem = fixture.debugElement.queryAll(By.css('fdp-radio-button'));
```
fkolar, line 186: I dont think you need debug element here.

KevinOkamoto: In this PR, I'm just trying to fix the tests so they pass. I don't want to change the RadioGroup tests at this time. My main goal is to add unit test for FormGroup and FormField, fixing ActionBar and RadioGroup were just necessary steps to get everything passing.

libs/platform/src/lib/components/form/form.spec.ts
KevinOkamoto: Because FormGroup and FormFields are so integrated, I though it best to test them together  using use-case scenarios, rather than around the specific features of each. There are use cases like layout and error handling which touch on features of both components. However this is just the start of the testing of FormGroup/FormField, a lot more test need to be added, so perhaps we can discuss the structure of these test at a later time. Right now I just want to add a few tests to get started.

PR #2650, opened 2020-06-09
feat: fiori 3 table sort filter

### LINKED ISSUE #2438: Fiori 3 Table - sorting/filtering via column header
Each column should be sortable and filterable.  Application developer should be able to pass custom sortFn/filterFn as inputs to the table to hand their own logic. Different columns should allow different sorting and filtering functions.

3 days

mikerodonnell89:
fixes #2438 Note this is not a pr against master


libs/core/src/lib/table/directives/table-inner.directive.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { TableInnerDirective } from './table-inner.directive';
+
+describe('TableInnerDirective', () => {
+    it('should create an instance', () => {
+        const directive = new TableInnerDirective();
+        expect(directive).toBeTruthy();
+    });
+});
```
InnaAtanasova, line 8: maybe test if `fd-table__inner` class is assigned to the host

libs/core/src/lib/table/directives/table-popover.directive.ts
```diff
@@ -0,0 +1,13 @@
+import { Directive, HostBinding } from '@angular/core';
+
+@Directive({
+    selector: '[fdTablePopover], [fd-table-popover]',
+    host: {
+        class: ''
```
InnaAtanasova, line 6: maybe remove this since it's not used

libs/core/src/lib/table/directives/table-popover.directive.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { TablePopoverDirective } from './table-popover.directive';
+
+describe('TablePopoverDirective', () => {
+    it('should create an instance', () => {
+        const directive = new TablePopoverDirective();
+        expect(directive).toBeTruthy();
+    });
```
InnaAtanasova, line 7: same as above.It's good to test if the proper class is assigned

JKMarkowski:
image

I know that it's mostly styles issue, can you add some flag to `list` item, to add class, that will remove active state changes, by overwriting background color?

apps/docs/src/app/core/component-docs/table/examples/table-column-sorting-example.component.html
```diff
@@ -1,15 +1,44 @@
...
             <th fd-table-cell>Column 2</th>
             <th fd-table-cell>Column 3</th>
-            <th fd-table-cell fdColumnSortable [sortDir]="dateSortDir" (click)="sortDate()">Date</th>
+            <th fd-table-cell>Date</th>
             <th fd-table-cell>Type</th>
         </tr>
     </thead>
     <tbody fd-table-body>
-        <tr *ngFor="let row of tableRows" fd-table-row>
+        <tr *ngFor="let row of displayedRows" fd-table-row>
```
JKMarkowski, line 41: I know that it's only the example, but isn't easier to put pipe like `displayedRows | sortingPipe : 'argumentName'`?

mikerodonnell89:
> Screen Shot 2020-06-11 at 1 43 02 PM
> 
> It's prob styles issue but I think when the cursor is over this area it should be a pointer

I agree that it should be handled in styles

rengare:
@mikerodonnell89 take a look at @JKMarkowski comment

mikerodonnell89:
@JKMarkowski @rengare fixed

apps/docs/src/app/core/component-docs/table/table-docs.component.ts
```diff
@@ -117,6 +119,18 @@ export class TableDocsComponent {
             component: 'TableColumnSortExampleComponent',
             code: tableColumnSortTsSrc,
             fileName: 'table-column-sort-example'
```
stefanoScalzo, line 121: this should be table-column-sorting-example

mikerodonnell89: Just seeing this, I'll fix it on the main table branch

PR #2656, opened 2020-06-12
fix: Failing tests on checkbox and switch

JKMarkowski:
On travis, tests for checkbox and switch components are randomly failing. It happens quite often. Those hcanges should fix it.


fkolar:
@JKMarkowski is this the same issue? 

https://github.com/SAP/fundamental-ngx/issues/2685

JKMarkowski:
It was same issue, we have to make something with those tests, because they fail randomly

Issue #2674, opened 2020-06-15
0.19.0 15.06 Defect Hunting


JKMarkowski:
- [x] Landscape calendar mobile mode - Inna [PR](https://github.com/SAP/fundamental-ngx/pull/2678)
![image](https://user-images.githubusercontent.com/26483208/84655014-7d046900-af10-11ea-8f13-3adaf9b902d6.png)

- [x] Checkbox IE11 tristate - https://github.com/SAP/fundamental-ngx/pull/2677
![checkboxgif](https://user-images.githubusercontent.com/26483208/84655289-02881900-af11-11ea-99d9-fd98a3d12928.gif)

- [x] Docs page header IE11 - this is a styling issue, needs to be fixed in Fund-styles - Inna (No PR for this since I pushed in Master)
![image](https://user-images.githubusercontent.com/26483208/84655713-bb4e5800-af11-11ea-956f-58856d1bb89c.png)

- [x] Input required label IE11 - this is a styling issue, needs to be fixed in Fund-styles - Inna
![image](https://user-images.githubusercontent.com/26483208/84655811-efc21400-af11-11ea-92a9-44137836af25.png)

- [x] Multi Input Mobile Mode - input inside bar - Jedrzej https://github.com/SAP/fundamental-ngx/pull/2695
![image](https://user-images.githubusercontent.com/26483208/84655967-33b51900-af12-11ea-86ac-c2369eca2f37.png)

![image](https://user-images.githubusercontent.com/17496353/84743771-afff3900-afb2-11ea-9032-99de89495e24.png)

- [x] Side Navigation, selected when clicked on expand icon  Jedrzej https://github.com/SAP/fundamental-ngx/pull/2676
![selectsidenav](https://user-images.githubusercontent.com/26483208/84656688-5f84ce80-af13-11ea-9491-fdfa43279f57.gif)

- [x] Cannot iterate down with `Tab` over busy indicator wrappers (IE11) #2697
![image](https://user-images.githubusercontent.com/17496353/84741008-abd11c80-afae-11ea-9ae9-af306582ec2c.png)

- [ ] Calendar not supporting new mobile approach #2551 (Cross Browser)

- [x] Complex Dialog list elements are missing `is-selected/is-active` styling (Cross Browser) - Jedrzej
![image](https://user-images.githubusercontent.com/17496353/84742881-6d892c80-afb1-11ea-936a-57c24adb2d07.png)
[...]

PR #2676, opened 2020-06-15
fix: Stop propagating clicks from nested icon

JKMarkowski:
relates to: https://github.com/SAP/fundamental-ngx/issues/2674
Click event now is not propagated upper, so it doesn't provide selected flag change.


mikerodonnell89:
Not sure if this should be considered a separate issue - but if I select the main link 3, then open the children, then select a child, i can have 2 items selected at once
Screen Shot 2020-06-15 at 10 58 13 AM

Issue #2571, opened 2020-05-28
feat: introduce new Textarea component in ngx/platform


kavya-b:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
Currently, we do not have a separate component for Textarea in ngx/core and use it directly with form component in the documentation pages. However, Fiori3 has requirements of textarea-counter, autogrow and some specific error message state handling scenarios(typing vs copy-paste has different behaviors) which warrants having textarea in its own component. This component will also be used alongwith Platform's Forms components.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 9, Fundamental Lib: 0.8.0

#### If this is a bug, please provide steps for reproducing it.
NA

#### Please provide relevant source code if applicable.
NA

#### Is there anything else we should know?
NA

kavya-b:
Introduced in #2691

PR #2691, opened 2020-06-17
fix: (platform) Introduce textarea as a component

kavya-b:
#2571 

Introducing textarea as a component in ngx/platform, which works with Platform Forms in a reactive and template approach, and also has the added features of autogrowing and displaying a counter message.

Documentation checklist:
- [x] Documentation Examples

### DIFF
--- a/apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-autogrow-example.component.html
+++ b/apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-autogrow-example.component.html
@@ -1,0 +1,151 @@
+<main class="fd-page">
+    <div *ngIf="true" class="fd-page__content fd-has-background-color-neutral-2">
+        <fdp-form-group #fg4 [multiLayout]="true" [formGroup]="form">
+            <fdp-form-field
+                #ff1
+                [id]="'growingDisabledTextarea'"
+                label="Textarea with growing=false and growingMaxLines=7"
+                [placeholder]="
+                    'Start entering something. Growing is not enabled, but growingMaxLines of 7 is given, so it is restricted to a height of 7 lines.'
+                "
+                hint="This is tooltip help"
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+                style="width: auto;"
+            >
+                <fdp-textarea
+                    [name]="'growingDisabledTextarea'"
+                    [formControl]="ff1.formControl"
+                    [growingMaxLines]="7"
+                    [cols]="10"
+                    [wrapType]="'hard'"
+                >
+                </fdp-textarea>
+            </fdp-form-field>
+            <fdp-form-field
+                #ff11
+                [id]="'growingDisabledHeightTextarea'"
+                label="Textarea with growing=false and height='80px'"
+                [placeholder]="
+                    'Start entering something.  Growing is not enabled, but height is set, so it is restricted to a height of 80px'
+                "
+                hint="This is tooltip help"
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+                [validators]="textareaValidator"
+            >
+                <fdp-textarea
+                    [formControl]="ff11.formControl"
+                    [name]="'growingDisabledHeightTextarea'"
+                    [growing]="false"
+                    [showExceededText]="true"
+                    [maxLength]="6"
+                    [cols]="10"
+                    [height]="'80px'"
+                    [wrapType]="'hard'"
+                >
+                </fdp-textarea>
+            </fdp-form-field>
+
+            <fdp-form-field
+                #ff2
+                [id]="'growingMaxLinesTextarea'"
+                label="Textarea with growing=true and growingMaxLines=5"
+                [placeholder]="
+                    'Start entering something. It will grow until growingMaxLines number of lines are visible.'
+                "
+                hint="This is tooltip help"
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+            >
+                <fdp-textarea
+                    [formControl]="ff2.formControl"
+                    [name]="'growingMaxLinesTextarea'"
+                    [growing]="true"
+                    [growingMaxLines]="5"
+                    [cols]="10"
+                    [wrapType]="'hard'"
+                >
+                </fdp-textarea>
+            </fdp-form-field>
+
+            <fdp-form-field
+                #ff3
+                [id]="'growingHeightTextarea'"
+                label="Textarea with growing=true and height='80px'"
+                [placeholder]="'Start entering something. It will grow until height is reached.'"
+                hint="This is tooltip help"
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+                [validators]="textareaValidator"
+            >
+                <fdp-textarea
+                    [formControl]="ff3.formControl"
+                    [name]="'growingHeightTextarea'"
+                    [growing]="true"
+                    [showExceededText]="true"
+                    [maxLength]="6"
+                    [cols]="10"
+                    [height]="'80px'"
+                    [wrapType]="'hard'"
+                    [value]="tValue"
+                >
+                </fdp-textarea>
+            </fdp-form-field>
+
+            <fdp-form-field
+                #ff4
+                [id]="'growingTextarea'"
+                label="Textarea with growingMaxLines and height both set"
+                [placeholder]="'Start entering something. Preference will be given to grow by height.'"
+                hint="This is tooltip help"
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+            >
+                <fdp-textarea
+                    [formControl]="ff4.formControl"
+                    [name]="'growingTextarea'"
+                    [growing]="true"
+                    [growingMaxLines]="3"
+                    [cols]="10"
+                    [height]="'80px'"
+                    [wrapType]="'hard'"
+                >
+                </fdp-textarea>
+            </fdp-form-field>
+
+            <fdp-form-field
+                #ff5
+                [id]="'growingOptionsDisabledTextarea'"
+                label="Textarea with growing=true neither height nor growingMaxLines set"
+                [placeholder]="'Start entering something. It will keep growing.'"
+                hint="This is tooltip help"
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+            >
+                <fdp-textarea
+                    [name]="'growingOptionsDisabledTextarea'"
+                    [formControl]="ff5.formControl"
+                    [growing]="true"
+                    [cols]="10"
+                    [wrapType]="'hard'"
+                >
+                </fdp-textarea>
+            </fdp-form-field>
+            <ng-template #i18n let-errros>
+                <span *ngIf="errros.required">
+                    Value is required
+                </span>
+                <span role="alert" aria-atomic="true" *ngIf="errros.maxlength">
+                    Please get your character count under limit.
+                </span>
+            </ng-template>
+        </fdp-form-group>
+    </div>
+</main>
--- a/apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-basic-example.component.html
+++ b/apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-basic-example.component.html
@@ -1,0 +1,82 @@
+<main class="fd-page">
+    <div *ngIf="true" class="fd-page__content fd-has-background-color-neutral-2">
+        <fdp-form-group #fg1 [multiLayout]="true" [formGroup]="form" [object]="data">
+            <fdp-form-field
+                #ff1
+                [id]="'basicTextarea'"
+                label="Basic Textarea with Platform Forms"
+                [placeholder]="'Start entering detailed description'"
+                hint="This is tooltip help"
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+            >
+                <fdp-textarea [name]="'basicTextarea'" [formControl]="ff1.formControl"></fdp-textarea>
+            </fdp-form-field>
+
+            <fdp-form-field
+                #ff2
+                [id]="'readonlyDescription'"
+                label="Readonly Textarea"
+                [placeholder]="'Start entering detailed description'"
+                hint="This is tooltip help"
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+            >
+                <fdp-textarea [name]="'readonlyDescription'" [formControl]="ff2.formControl" [readonly]="true">
+                </fdp-textarea>
+            </fdp-form-field>
+
+            <fdp-form-field
+                #ff3
+                [id]="'disabledDescription'"
+                label="Disabled Textarea"
+                [placeholder]="'Start entering detailed description'"
+                hint="This is tooltip help"
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+            >
+                <fdp-textarea [name]="'disabledDescription'" [disabled]="true" [value]="'this is a disabled textarea'">
+                </fdp-textarea>
+            </fdp-form-field>
+
+            <fdp-form-field
+                #ff4
+                [id]="'compactTextarea'"
+                label="Compact Textarea"
+                [placeholder]="'Start entering detailed description'"
+                hint="This is tooltip help"
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+            >
+                <fdp-textarea [name]="'compactTextarea'" [formControl]="ff4.formControl" [contentDensity]="'compact'">
+                </fdp-textarea>
+            </fdp-form-field>
+        </fdp-form-group>
+
+        <div fd-form-item>
+            <label fd-form-label for="textarea-1">Textarea without Platform Forms</label>
+            <fdp-textarea
+                name="textarea-1"
+                placeholder="write something here"
+                [maxLength]="10"
+                [showExceededText]="true"
+                [growing]="true"
+                [growingMaxLines]="5"
+                [cols]="10"
+                [wrapType]="'soft'"
+                #textarea
+            >
+            </fdp-textarea>
+            <fd-form-message
+                class="custom-form-message"
+                *ngIf="textarea.value && textarea.value.length > 10"
+                [type]="'warning'"
+                >This is an example warning when used without forms.
+            </fd-form-message>
+        </div>
+    </div>
+</main>
--- a/apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-counter-example.component.html
+++ b/apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-counter-example.component.html
@@ -1,0 +1,68 @@
+<main class="fd-page">
+    <div *ngIf="true" class="fd-page__content fd-has-background-color-neutral-2">
+        <fdp-form-group #fg2 [multiLayout]="true" [formGroup]="form">
+            <fdp-form-field
+                #ff2
+                [id]="'detailedDescription'"
+                label="Detailed description"
+                [placeholder]="'Start entering detailed description'"
+                hint="This is tooltip help"
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+                [required]="true"
+                [validators]="textareaValidator"
+            >
+                <fdp-textarea
+                    [formControl]="ff2.formControl"
+                    [name]="'detailedDescription'"
+                    [maxLength]="10"
+                    [stateType]="'error'"
+                    [showExceededText]="true"
+                    value="Lorem ipsum, dolor sit amet"
+                >
+                </fdp-textarea>
+            </fdp-form-field>
+
+            <fdp-form-field
+                #ff3
+                [id]="'noCounterMessageInteraction'"
+                label="When showExceededText is false and value is set"
+                [placeholder]="
+                    'Start entering. Counter message will not be shown, but if an initial value is set whose length > maxLength, excess text is removed on Backpress'
+                "
+                hint="This is tooltip help"
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+                [required]="true"
+                [validators]="textareaNoCounterMessageValidator"
+            >
+                <fdp-textarea
+                    [formControl]="ff3.formControl"
+                    [name]="'noCounterMessageInteraction'"
+                    [maxLength]="10"
+                    [showExceededText]="false"
+                    value="Some value that crosses max length"
+                >
+                </fdp-textarea>
+            </fdp-form-field>
+
+            <fdp-form-field [id]="'submit-button'" rank="20" zone="zLeft">
+                <fdp-button [type]="'submit'" (click)="onSubmit()">Submit</fdp-button>
+            </fdp-form-field>
+
+            <ng-template #i18n let-errros>
+                <span *ngIf="errros.required">
+                    Value is required
+                </span>
+                <span role="alert" aria-atomic="true" *ngIf="errros.maxlength">
+                    Please get your character count under limit.
+                </span>
+            </ng-template>
+        </fdp-form-group>
+    </div>
+</main>
+
+<pre>Form result: {{ form.value | json }}</pre>
+<pre>Form Status: {{ form.status | json }}</pre>
--- a/libs/platform/src/lib/components/form/text-area/text-area.component.html
+++ b/libs/platform/src/lib/components/form/text-area/text-area.component.html
@@ -1,0 +1,36 @@
+<textarea
+    #textareaElement
+    fd-form-control
+    [compact]="contentDensity == 'compact'"
+    [ngClass]="[growing ? 'fdp-textarea-custom-resize' : 'fdp-textarea-normal-size']"
+    [attr.id]="id"
+    [attr.placeholder]="placeholder"
+    [disabled]="disabled"
+    [attr.aria-disabled]="disabled"
+    [readonly]="readonly"
+    [attr.aria-readonly]="readonly"
+    [rows]="growing ? 2 : height ? 2 : growingMaxLines"
+    [cols]="cols"
+    [wrap]="wrapType"
+    [maxlength]="!showExceededText ? maxLength : null"
+    [(ngModel)]="value"
+    [state]="getUpdatedState()"
+    (paste)="handlePasteInteraction()"
+    (blur)="_onFocusChanged(false)"
+    (focus)="_onFocusChanged(true)"
+></textarea>
+
+<!-- ICU recommends full text in format -->
+<div class="fd-textarea-counter" *ngIf="showExceededText" role="alert" aria-atomic="true">
+    <!-- 
+        The above will work only when i18n marker is used. Without i18n, the code works, but tests fail as the spec is apparently 
+        internally dependent on i18n marker.
+        Instead, we use I18nPluralPipe or I18nSelectPipe to achieve the same result but without using i18n markers or having failing tests.
+        Will revert to the above code once we bring in i18n and localize lib to ngx.
+    -->
+    <span>
+        {{ exceededCharCount }}
+        {{ exceededCharCount | i18nPlural: { '=1': 'character', other: 'characters' } }}
+        {{ counterExcessOrRemaining | i18nSelect: { excess: 'over the limit', remaining: 'remaining' } }}</span
+    >
+</div>
--- a/libs/platform/src/lib/components/form/text-area/text-area.component.scss
+++ b/libs/platform/src/lib/components/form/text-area/text-area.component.scss
@@ -1,0 +1,21 @@
+.fdp-textarea-custom-resize {
+    width: auto;
+    height: auto;
+    min-width: 0rem;
+    min-height: 0rem;
+}
+
+.fdp-textarea-normal-size {
+    max-height: inherit;
+    width: auto;
+    min-width: 0rem;
+}
+
+.fd-textarea-counter {
+    display: inline;
+    float: right;
+
+    [dir='rtl'] & {
+        float: left;
+    }
+}
--- a/libs/platform/src/lib/components/form/text-area/text-area.component.spec.ts
+++ b/libs/platform/src/lib/components/form/text-area/text-area.component.spec.ts
@@ -1,0 +1,318 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { TextAreaComponent } from './text-area.component';
+import { FormGroup, ReactiveFormsModule, ValidatorFn, Validators, FormControl } from '@angular/forms';
+import { Component, ViewChild, ElementRef } from '@angular/core';
+import { FormFieldComponent, FdpFormGroupModule } from '@fundamental-ngx/platform';
+import { PlatformTextAreaModule } from './text-area.module';
+import { By } from '@angular/platform-browser';
+import { createKeyboardEvent } from '../../../testing/event-objects';
+import { DELETE } from '@angular/cdk/keycodes';
+
+@Component({
+    selector: 'fdp-test-textarea',
+    template: `
+        <form [formGroup]="form" (ngSubmit)="onSubmit($event)">
+            <fdp-form-group #fg1 [multiLayout]="true" [formGroup]="form">
+                <fdp-form-field
+                    #basicTextareaField
+                    [id]="'basicTextarea'"
+                    label="Basic Textarea with Platform Forms"
+                    [placeholder]="'Start entering detailed description'"
+                    hint="This is tooltip help"
+                    zone="zLeft"
+                    rank="10"
+                    hintPlacement="'left'"
+                    [validators]="textareaValidator"
+                >
+                    <fdp-textarea
+                        formControlName="basicTextarea"
+                        name="'basicTextarea'"
+                        [growingMaxLines]="3"
+                        [growing]="true"
+                        [contentDensity]="'compact'"
+                        [maxLength]="10"
+                        [cols]="10"
+                        [stateType]="'error'"
+                        [showExceededText]="true"
+                        [height]="'80px'"
+                        [wrapType]="'hard'"
+                    ></fdp-textarea>
+                </fdp-form-field>
+                <ng-template #i18n let-errors>
+                    <span *ngIf="errors && errors.required" class="error">This field is required.</span>
+                    <span *ngIf="errors && errors.maxlength">
+                        Please get your character count under limit.
+                    </span>
+                </ng-template>
+            </fdp-form-group>
+            <button type="submit" #submitButton>Submit</button>
+        </form>
+    `
+})
+class BasicTextareaTestWrapperComponent {
+    @ViewChild(TextAreaComponent)
+    textareaComponent: TextAreaComponent;
+    @ViewChild('basicTextareaField') basicTextareaField: FormFieldComponent;
+
+    @ViewChild('submitButton') submitButton: ElementRef<HTMLElement>;
+
+    public form: FormGroup = new FormGroup({
+        basicTextarea: new FormControl('this is a random note')
+    });
+
+    textareaValidator: ValidatorFn[] = [Validators.maxLength(10), Validators.required];
+
+    public result: any = null;
+
+    onSubmit() {
+        this.result = this.form.value;
+    }
+}
+
+describe('Basic Textarea', () => {
+    let fixture: ComponentFixture<BasicTextareaTestWrapperComponent>;
+    let host: BasicTextareaTestWrapperComponent;
+
+    beforeEach(async(() => {
+        TestBed.configureTestingModule({
+            imports: [ReactiveFormsModule, FdpFormGroupModule, PlatformTextAreaModule],
+            declarations: [BasicTextareaTestWrapperComponent]
+        }).compileComponents();
+    }));
+
+    beforeEach(() => {
+        fixture = TestBed.createComponent(BasicTextareaTestWrapperComponent);
+        host = fixture.componentInstance;
+
+        fixture.detectChanges();
+    });
+
+    async function wait(componentFixture: ComponentFixture<any>) {
+        componentFixture.detectChanges();
+        await componentFixture.whenStable();
+    }
+
+    it('should create', () => {
+        expect(host).toBeTruthy();
+    });
+
+    it('should have compact class applied', () => {
+        const textareaElement = fixture.debugElement.query(By.css('textarea'));
+        expect(textareaElement.nativeElement.classList.contains('fd-textarea--compact')).toBeTruthy();
+    });
+
+    it('should have a label, hint, placeholder and default value', async () => {
+        await wait(fixture);
+
+        const textareaLabel = host.basicTextareaField.label;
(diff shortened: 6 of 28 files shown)


libs/platform/src/lib/components/form/textarea/textarea.component.ts
```diff
@@ -0,0 +1,354 @@
...
+    encapsulation: ViewEncapsulation.None,
+    providers: [
+        {
+            provide: FormFieldControl,
+            useExisting: TextareaComponent,
+            multi: true
+        }
+    ]
+})
+export class TextareaComponent extends BaseInput {
```
fkolar, line 60: Let;s name it TextAreaXXXXX

libs/platform/src/lib/components/form/textarea/textarea.component.ts
```diff
@@ -0,0 +1,354 @@
...
+    @Input()
+    height: string;
+
+    @Input()
+    growingMaxLines: number; // rows
```
fkolar, line 65: Missing Comments for @Inputs. Not sure what is growing

libs/platform/src/lib/components/form/textarea/textarea.component.scss
```diff
@@ -0,0 +1,22 @@
+.customResize {
```
fkolar, line 1: We dont name classes like this

libs/platform/src/lib/components/form/textarea/textarea.component.ts
```diff
@@ -0,0 +1,354 @@
...
+
+    @Input()
+    cols: number;
+
+    @Input()
+    wrap: WrapType = 'soft';
```
fkolar, line 71: easier to name it wrapType

libs/platform/src/lib/components/form/textarea/textarea.component.ts
```diff
@@ -0,0 +1,354 @@
...
+    readonly: boolean;
+
+    @Input()
+    maxLength: number;
+
+    @Input()
+    showExceededText: boolean = false;
+
+    @Input()
+    growing: boolean = false;
```
fkolar, line 83: Let's rename it to autoSize

libs/platform/src/lib/components/form/textarea/textarea.component.ts
```diff
@@ -0,0 +1,354 @@
...
+
+    @Input()
+    textExceededMessage: string;
+
+    /**
+     *  The state of the form control - applies css classes.
+     *  Can be `success`, `error`, `warning`, `information` or blank for default.
+     */
+    @Input()
+    stateType: FormStates = 'warning';
```
fkolar, line 93: this shoudl be in the base component

libs/platform/src/lib/components/form/textarea/textarea.component.ts
```diff
@@ -0,0 +1,354 @@
...
+    @ViewChild('textareaElement')
+    textareaElement: ElementRef;
+
+    /** @hidden */
+    hasTextExceeded: boolean = false;
+
+    /** excess character count */
+    exceededCharCount: number = this.maxLength ? this.maxLength : 0;
+
+    counterExcessOrRemaining = 'remaining';
```
fkolar, line 123: I think angular 9 supports i18n inside TS code

libs/platform/src/lib/components/form/textarea/textarea.component.ts
```diff
@@ -0,0 +1,354 @@
...
+
+    isValueCustomSet = false;
```
fkolar, line 125: What is this ?

libs/platform/src/lib/components/form/textarea/textarea.component.ts
```diff
@@ -0,0 +1,354 @@
...
+    constructor(
+        protected _cd: ChangeDetectorRef,
+        @Optional() @Self() public ngControl: NgControl,
+        @Optional() @Self() public ngForm: NgForm
+    ) {
+        super(_cd, ngControl, ngForm);
+        if (this.ngControl) {
+            this.ngControl.valueAccessor = this;
+        }
+        this.id = `textarea-${uniqueId++}`;
```
fkolar, line 156: remove it from contructor - just intialize just like other inputs.

libs/platform/src/lib/components/form/textarea/textarea.component.ts
kavya-b: added comments

libs/platform/src/lib/components/form/textarea/textarea.component.scss
kavya-b: renamed classes

libs/platform/src/lib/components/form/textarea/textarea.component.ts
kavya-b: renamed to wrapType

libs/platform/src/lib/components/form/textarea/textarea.component.ts
kavya-b: fiori design and interaction page for textarea talks about a 'growing' property. Kept the same name for consistency.

libs/platform/src/lib/components/form/textarea/textarea.component.ts
kavya-b: fiori design and interaction page for textarea allows application user to specify what state the textarea can fall into once it crosses the max length, for eg, `warning`, `error` etc. 
By default, the stateType is `warning`, but if set to `error`, then beyond the max length character, the state type changes to `error`. 
The base component currently supports only `error` state, and I did not want to change that code for this specific usecase.

libs/platform/src/lib/components/form/textarea/textarea.component.ts
kavya-b: Since the counter message is dynamic, it can be `10 characters remaining`, or `10 characters over the limit`, or even `1 character remaining` (character vs characters). 
ICU expressions handle this kind of dynamically changing text for i18n by it plural and select expressions. The Select ICU expression expects a variable(in this case `counterExcessOrRemaining`)  which can take some values (in this case `remaining` or `excess`) and we can give the text to be displayed in {} as follows:
```html
{counterExcessOrRemaining, select, excess {over the limit} remaining {remaining}} 
```
the extracted .xlf files will only replace the text in the {} with the respective language.

libs/platform/src/lib/components/form/textarea/textarea.component.ts
kavya-b: This is a flag to check if there is an initial value being set in the html through fdp-form-field so that the height of the textarea can contain this value, if it multi-line, according to the height limits specified by `height` or `growingMaxLines` attributes.

libs/platform/src/lib/components/form/textarea/textarea.component.ts
kavya-b: removed

libs/platform/src/lib/components/form/textarea/textarea.component.ts
```diff
@@ -0,0 +1,393 @@
...
+    ngAfterViewInit(): void {
+        // set max height
+        this._setMaxHeight();
+
+        // don't set any error state if we are not showing counter message
+        if (!this.showExceededText) {
+            this.stateType = null;
+        }
+
+        this._cd.detectChanges();
```
KevinOkamoto, line 210: Should replace this line with super call: `super.ngAfterViewInit()`.

libs/platform/src/lib/components/form/textarea/textarea.component.ts
kavya-b: have removed FormStates and used Status instead, but I still need to take this as Input() because of the design requirement as mentioned in the previous comment.

kavya-b:
@sKudum 
> 1). Can we remove duplicated content in examples?

Done.
> 2). For counter(static text) i18n support=>I think we need to discuss the semantic approach that Frank was mentioned in our previous discussions.

This needs more discussion but I think it is out of the scope for this PR.
> 3). For keycode comparisons instead of hard code, can we use @angular/cdk/Keycodes?

Have used angular keycodes now.
> 4).Readme check you can remove, this PR does not have any changes in read me file

Done

libs/platform/src/lib/components/form/textarea/textarea.component.ts
kavya-b: changed lib classes to TextAreaXXX. Should I also change the documentation class names?

libs/platform/src/lib/components/form/textarea/textarea.component.ts
```diff
@@ -0,0 +1,393 @@
...
+        if (!this.wrapType || VALID_WRAP_TYPES.indexOf(this.wrapType) === -1) {
+            throw new Error(`Textarea wrap type $ {this.wrapType} is not supported`);
+        }
+        // if not custom set, set counter to max length value, else it calculates remaining/exceeded characters.
+        if (!this.value) {
+            this.exceededCharCount = this.maxLength ? this.maxLength : 0;
+        } else {
+            this.isValueCustomSet = true;
+        }
+    }
```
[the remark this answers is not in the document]
kavya-b: done

libs/platform/src/lib/components/form/textarea/textarea.component.ts
kavya-b: done

fkolar:
When you show examples with the formgroup., you should not use name on the controller level. You are adding Id, that enough. If you want to specify a name then you do it on FormField Level

angular.json
```diff
@@ -126,14 +132,8 @@
...
                     "options": {
-                        "tsConfig": [
-                            "apps/docs/tsconfig.app.json",
-                            "apps/docs/tsconfig.spec.json"
-                        ],
-                        "exclude": [
-                            "**/node_modules/**",
-                            "!apps/docs/**"
-                        ]
+                        "tsConfig": ["apps/docs/tsconfig.app.json", "apps/docs/tsconfig.spec.json"],
```
fkolar, line 135: Why are you changing formating of these section that you havent touched ?

apps/docs/src/app/core/component-docs/select/examples/select-adding-example/select-adding-example.component.html
```diff
@@ -3,7 +3,7 @@
     <button fd-button (click)="removeOption()">Remove Option</button>
 </div>
```
fkolar, line 4: Is this correct PR? Whey are you dealing with select? It just means that you havent properly rebased master adn you are sumbitting this as change.  this should not be here all related select

kavya-b: I was rebasing from master at the exact time of your comment, not sure where this file came from, but i don't see it anymore. Please check

angular.json
kavya-b: I am not explicitly modifying formatting, I have prettier installed, every time I save the file, it changes this file's formatting :(

kavya-b:
> When you show examples with the formgroup., you should not use name on the controller level. You are adding Id, that enough. If you want to specify a name then you do it on FormField Level

base-input explicitly checks for both id and name availability:
image

angular.json
fkolar: ok if this prettier it should be ok

libs/platform/src/lib/components/form/text-area/text-area.component.ts
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    /** @hidden */
+    ngAfterViewChecked() {
+        // when value is custom set(initial value is present), the heights don't get
+        // set to show the full text in the ngAfterViewInit immediately. therefore, we call autoGrowTextArea in
+        // ngAfterViewChecked and detect changes again.
+        if (this.value && this.isValueCustomSet) {
+            this.autoGrowTextArea(); // if applicable, grow the textarea when value is custom set
+            this._cd.detectChanges();
```
fkolar, line 221: Why are you calling detect changes . Can you call simply stateChanges.emit() to inform the parent (formField). Maybe we dont need to call autoGrow everytime and only when it changed.

libs/platform/src/lib/components/form/text-area/text-area.component.ts
```diff
@@ -0,0 +1,395 @@
...
+        }
+    }
+
+    /** write value for ControlValueAccessor */
+    writeValue(value: any): void {
+        if (value) {
+            super.writeValue(value);
+            this.updateCounterInteractions();
+            this._cd.markForCheck();
+        }
```
fkolar, line 231: Again here you want to call stateChanges again like writeValue does.

libs/platform/src/lib/components/form/text-area/text-area.component.ts
```diff
@@ -0,0 +1,395 @@
...
+    /* else handle autogrow case
+    */
+    @HostListener('keyup', ['$event'])
+    handleBackPress(event: KeyboardEvent) {
+        // if not showing exceeded text message/interactions or growing, and custom value set
+        if (this.growing) {
+            this.autoGrowTextArea();
+            this.isValueCustomSet = false; // set it to false first time it comes here
+        }
+        if (!this.showExceededText && (event.keyCode === DELETE || event.keyCode === BACKSPACE)) {
```
fkolar, line 329: Use key utils. We agreed that we will try to use keyUtils as much as we can and stop using keyCOde which is deprecated. Please see select for the usage (select-commons.ts)

libs/platform/src/lib/components/form/text-area/text-area.component.ts
```diff
@@ -0,0 +1,395 @@
...
+        }
+    }
+
+    /**
+     * get the updated state when character count breaches `maxLength`
+     */
+    getUpdatedState(): string {
+        if (
+            ((this.textareaElement && this.textareaElement.nativeElement.value.length) ||
+                (this.value && this.value.length)) > this.maxLength
```
fkolar, line 348: I am not sure if I understand this fragment 
`(this.value && this.value.lenght)` > this.maxLenght. 

The blog where you expect a value is boolean expression

libs/platform/src/lib/components/form/text-area/text-area.module.ts
```diff
@@ -0,0 +1,12 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { FormModule as FdFormModule } from '@fundamental-ngx/core';
+import { FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { TextAreaComponent } from './text-area.component';
+
+@NgModule({
+    imports: [CommonModule, FormsModule, ReactiveFormsModule, FdFormModule],
```
fkolar, line 8: We should really fix core and stop using FormModule. Its pretty expensive to import everything. Its the same as we would be importing our PLatform Module.

kavya-b: Core does not have any directive for Textarea as such. Everything related to `state` and `compact` comes through the `fd-form-control` directive. We can expose the `fd-form-control` directive in a separate module, but I think it is out of the scope of this PR.

libs/platform/src/lib/components/form/text-area/text-area.component.ts
kavya-b: This block checks whether the underlying element ref `textareaElement` or the `value` field, if it has some content, if the length is greater than the `maxLength` value.
The reason we are looking at both `textareaElement` and the `value` fields is because if value is initially set through `fdp-form-field`, then `textareaElement` will be undefined, therefore we are checking for `value` as well.

libs/platform/src/lib/components/form/text-area/text-area.component.ts
kavya-b: calling stateChanges now

libs/platform/src/lib/components/form/text-area/text-area.component.ts
kavya-b: calling stateChanges now.
autogrow method will only be called once `isValueCustomSet` is set. Once user edits the textarea or changes this value, `isValueCustomSet` is set to false and this block will not be run again.

libs/platform/src/lib/components/form/text-area/text-area.component.ts
kavya-b: using KeyUtils now

apps/docs/src/app/platform/component-docs/locale/messages.xlf
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
fkolar, line 1: Why do you have XLS files here? I think we said you will remove them ? First we need to have end2end process with schematics that cna add these files to the project. As you know this would not work right ?


SourceFile path have relative path to the file, but on the application level you showed me that path starts with ../nodemodules/..

libs/platform/src/lib/components/form/text-area/text-area.component.html
```diff
@@ -0,0 +1,31 @@
...
+    [maxlength]="!showExceededText ? maxLength : null"
+    [(ngModel)]="value"
+    [state]="getUpdatedState()"
+    (paste)="handlePasteInteraction()"
+    (blur)="_onFocusChanged(false)"
+    (focus)="_onFocusChanged(true)"
+></textarea>
+
+<!-- ICU recommends full text in format -->
+<div class="fd-textarea-counter" *ngIf="showExceededText" role="alert" aria-atomic="true">
```
fkolar, line 24: Let's first figure out my comment above about i18n

package.json
```diff
@@ -80,6 +80,7 @@
     "@angular/compiler": "~9.1.11",
     "@angular/core": "~9.1.11",
     "@angular/forms": "~9.1.11",
+    "@angular/localize": "^9.1.11",
```
fkolar, line 83: See my i18n comment

libs/platform/src/lib/components/form/text-area/text-area.component.ts
fkolar: Can you restructure this is not good for readability one thing is to test if value is defined and thing is to compare length don't mix those two

(this.value && this.value.lenght > xxxx) instead of (this.value && this.value.length)) > this.maxLength

apps/docs/src/app/platform/component-docs/locale/messages.xlf
kavya-b: regarding i18n, when I checked about the path issue you were mentioning
image
the one that starts with `../../libs` is the static string from our library, but the one starting with `src/app/` are the i18n markers placed by other components in the documentation pages. 
Until we bring in the schematics, when an application using ngx runs `ng xi18n`, it will not consider this `messages.xlf` at all, it will generate its own `messages.xlf` where the path will anyway resolve to `../../nodemodules`. It will work fine for them.
We are keeping i18n and` messages.xlf `here, because as discussed, we are not working on app-framework(demo app) for now, so we would be bringing i18n also to ngx for Puru to test the whole loop.

This is why I kept those files and the ICU expression logic for counter message.

libs/platform/src/lib/components/form/text-area/text-area.component.ts
kavya-b: Okay I see the readability issue, the second bracket was just an extra wrapping. I will move this condition to a private method and refer to it like `if (getContentLength() > maxLength)` instead.

kavya-b: have restructured it now

libs/platform/src/lib/components/form/text-area/text-area.module.ts
fkolar: Yes this is just a comment not something to change. The goal is to only import modules we need. FormModule imports and declares everything.

libs/platform/src/lib/components/form/text-area/text-area.component.html
kavya-b: Please check my reply above on i18n https://github.com/SAP/fundamental-ngx/pull/2691#discussion_r445471699. 
If it doesn't make much sense, I can remove it from this PR. I would also probably have to remove the `Support for i18n of message counter` as a platform enhancement description in this case.

package.json
kavya-b: @fkolar have removed localize library, .xlf files and modified the ICU expression in lib file. There is no point adding a whole new library just for one component's changes.

angular.json
```diff
@@ -23,18 +23,13 @@
                         "main": "apps/docs/src/main.ts",
                         "polyfills": "apps/docs/src/polyfills.ts",
                         "tsConfig": "apps/docs/tsconfig.app.json",
-                        "assets": [
-                            "apps/docs/src/favicon.ico",
-                            "apps/docs/src/assets"
-                        ],
+                        "assets": ["apps/docs/src/favicon.ico", "apps/docs/src/assets"],
```
DeepakSap14, line 26: you can remove this formatting. As No changes is made in this file.

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-basic-example.component.html
```diff
@@ -0,0 +1,95 @@
...
+                hintPlacement="'left'"
+            >
+                <fdp-textarea [name]="'readonlyDescription'" [formControl]="ff2.formControl" [readonly]="true">
+                </fdp-textarea>
+            </fdp-form-field>
+
+            <!-- <fdp-form-field
+                #ff3
+                [id]="'disabledDescription'"
+                label="Disabled Textarea"
```
DeepakSap14, line 34: unused code can be removed.

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-basic-example.component.html
```diff
@@ -0,0 +1,95 @@
...
+                zone="zLeft"
+                rank="10"
+                hintPlacement="'left'"
+                
+            >
+                <fdp-textarea formControlName="ff3" > </fdp-textarea>
+            </fdp-form-field> -->
+            <fdp-form-field
+                #ff3
+                [id]="'disabledDescription'"
```
stefanoScalzo, line 46: hover should not have grab cursor

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-basic-example.component.html
```diff
@@ -0,0 +1,95 @@
...
+                rank="10"
+                hintPlacement="'left'"
+            >
+                <fdp-textarea [name]="'compactTextarea'" [formControl]="ff4.formControl" [contentDensity]="'compact'">
+                </fdp-textarea>
+            </fdp-form-field>
+        </fdp-form-group>
+
+        <div fd-form-item>
+            <label fd-form-label for="textarea-1">Textarea without Platform Forms</label>
```
stefanoScalzo, line 74: I think it should be clarified what a 'platofrm form' is

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-counter-example.component.html
```diff
@@ -0,0 +1,77 @@
...
+                rank="10"
+                hintPlacement="'left'"
+                [required]="true"
+                [validators]="textareaValidator"
+            >
+                <fdp-textarea
+                    [formControl]="ff2.formControl"
+                    [name]="'detailedDescription'"
+                    [growingMaxLines]="3"
+                    [growing]="true"
```
stefanoScalzo, line 20: growing should take into account when user clicks enter as a line

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-counter-example.component.html
```diff
@@ -0,0 +1,77 @@
...
+                    [maxLength]="10"
+                    [cols]="10"
+                    [showExceededText]="false"
+                    [wrapType]="'hard'"
+                    value="Some value that crosses max length"
+                >
+                </fdp-textarea>
+            </fdp-form-field>
+
+            <fdp-form-field [id]="'submit-button'" rank="20" zone="zLeft">
```
stefanoScalzo, line 60: If you are showing only growing option, do not include character count. It will confuse the user

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-autogrow-example.component.html
```diff
@@ -0,0 +1,151 @@
+<main class="fd-page">
+    <div *ngIf="true" class="fd-page__content fd-has-background-color-neutral-2">
+        <fdp-form-group #fg4 [multiLayout]="true" [formGroup]="form">
+            <fdp-form-field
```
stefanoScalzo, line 4: fdp-form-help is not needed, we are showing the text area, not this component

libs/platform/src/lib/components/form/text-area/text-area.component.scss
```diff
@@ -0,0 +1,37 @@
...
+```
+Platform forms cannot use the above structure because our `fdp-input-message-group` is part of the `fdp-form-field` template
+that contains `fdp-textarea` and the form message. Also, our `fdp-textarea` needs to be tightly bound to the counter message, hence
+`fd-textarea-counter` class will have to be used within `fdp-textarea`, that is within `fdp-input-message-group` ultimately.
+Since this is a different structure from fund-styles, there is a gap visible for error messages when counter message is displayed.
+The following styles fixes that gap issue.
+*/
+.fd-textarea-counter {
+    display: inline;
+    float: right;
```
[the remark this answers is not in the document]
kavya-b: Unfortunately not. I tried with inline-block and text-align right in the counter's child element, and also with flex and space-between, but none work in the expected way. I understand that generally inline-block display and float right are never used together since only one of them works when 2 sibling elements are needed side-by-side, and the above mentioned solutions would all work when 2 css classes are in the same context of the parent elements. But here, the .fd-textarea-counter is in a completely different context than the form-message context(in fact, used one inside the other), and these solutions don't apply. Hence i had to use them this way.

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-basic-example.component.html
kavya-b: Basically when we use components in the context of `fdp-form-field` and `fdp-form-group`, we say we use it with Platform Forms. 
@fkolar are we talking about Platform Forms in the homepage of Platform documentation pages?

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-counter-example.component.html
kavya-b: I don't quite understand how to click enter, is it also there in the UI5 textarea sample?

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-counter-example.component.html
kavya-b: I will remove the growing options from this example. 
In the autogrow example, I have used a combination just to show the behaviour when both counter and growing options are enabled. Hope that is okay.

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-autogrow-example.component.html
kavya-b: Do you mean `fdp-form-field`? We are showing textarea in the form-field's context most of the times because we want the end user to understand where things like placeholder, label, validation arrays etc need to go and how they work with `fdp-textarea` or any child component. When just provided with `fdp-textarea`, we are not showing how it would work with `fdp-form-field` as such, or how error messages are thrown. 
I see even in core textarea, we are showing every example in the context of `fd-form-item` or `fd-form-input-message-group` etc. 
I think this is required.

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-basic-example.component.html
kavya-b: This is an optional way of disabling through formControlName and by passing `disabled:true` in the FormControl initialization in .ts. Since both achieve the same purpose, I kept one commented, and this can be used for user's reference if they don't want to use `[disabled]="true"` method.

angular.json
kavya-b: removed formatting

libs/platform/src/lib/components/form/text-area/text-area.component.ts
```diff
@@ -0,0 +1,405 @@
...
+                this.updateCounterInteractions();
+            });
+        }
+    }
+
+    /** handle auto growing of textarea */
+    autoGrowTextArea() {
+        if (this.growing) {
+            this.textareaElement.nativeElement.style.height = 'inherit';
+            // console.log('maxheight' + this.textareaElement.nativeElement.style.maxHeight);
```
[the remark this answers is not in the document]
kavya-b: removed

libs/platform/src/lib/components/form/text-area/text-area.component.ts
```diff
@@ -0,0 +1,402 @@
...
+        } else {
+            // for other navigators
+            navigator['clipboard'].readText().then((clipText) => {
+                this.updateCounterInteractions();
+            });
+        }
+    }
+
+    /** handle auto growing of textarea */
+    autoGrowTextArea() {
```
JKMarkowski, line 281: Why don't you bring cdk auto resize?

libs/platform/src/lib/components/form/text-area/text-area.component.ts
```diff
@@ -0,0 +1,402 @@
...
+        } else {
+            this.counterExcessOrRemaining = this.remainingText;
+            this.exceededCharCount = this.maxLength - this._textAreaCharCount;
+        }
+        // reset flag
+        this._isPasted = false;
+    }
+
+    /** handle exceeded maxlength case when text is pasted */
+    handlePasteInteraction(): void {
```
JKMarkowski, line 265: textareacopypaste
It's not working well.

kavya-b: there is a requirement that if pasting results in exceeded characters, we need to highlight the exceeded text : 
> When the user pastes copied text, characters that are over the limit are selected automatically. The user can delete any excess characters by pressing Delete or Back on the keyboard (or virtual keyboard for phones and tablets).

In UI5 sample also we can see same behaviour:

fkolar:
@JKMarkowski I did not wanted to enforce the CDK at the moment but its good idea as enhancements t to this. Not only CDK to handle autoresize but also CDK to do clipboard management so we can avoid window[xxx].



@kavya-b can you as part of the PR create an issue to capture this? Let’s get this in and plan the enhancement to this

fkolar:
https://timdeschryver.dev/blog/use-the-new-angular-clipboard-cdk-to-interact-with-the-clipboard

libs/platform/src/lib/components/form/text-area/text-area.component.ts
kavya-b: cdk's auto resizing works only on rows, and expects min rows and max rows to be given. Fiori requirement does not allow user to specify min rows. Fiori requirement also specifies growing by height, in addition to growing by rows(or maxLines), and when both are specified, height should be given preference.
Since some very specific requirements are present, we didn't use cdk features.

kavya-b:
> @JKMarkowski I did not wanted to enforce the CDK at the moment but its good idea as enhancements t to this. Not only CDK to handle autoresize but also CDK to do clipboard management so we can avoid window[xxx].
> 
> @kavya-b can you as part of the PR create an issue to capture this? Let’s get this in and plan the enhancement to this

Hi @fkolar , for textarea: cdk's auto resizing works only on rows, and expects min rows and max rows to be given. Fiori requirement does not allow user to specify min rows. Fiori requirement also specifies growing by height, in addition to growing by rows(or maxLines), and when both are specified, height should be given preference.
Since some very specific requirements are present, we didn't use cdk features.

fkolar:
> > @JKMarkowski I did not wanted to enforce the CDK at the moment but its good idea as enhancements t to this. Not only CDK to handle autoresize but also CDK to do clipboard management so we can avoid window[xxx].
> > @kavya-b can you as part of the PR create an issue to capture this? Let’s get this in and plan the enhancement to this
> 
> Hi @fkolar , for textarea: cdk's auto resizing works only on rows, and expects min rows and max rows to be given. Fiori requirement does not allow user to specify min rows. Fiori requirement also specifies growing by height, in addition to growing by rows(or maxLines), and when both are specified, height should be given preference.
> Since some very specific requirements are present, we didn't use cdk features.

I believe we can make use of cdk, if its by row or height it should not matter, you can work with both no. you always resize by row, you add or remove rows, you dont resize by px. But even in CDK you can specify heigh to set the limit.

kavya-b:
> > > @JKMarkowski I did not wanted to enforce the CDK at the moment but its good idea as enhancements t to this. Not only CDK to handle autoresize but also CDK to do clipboard management so we can avoid window[xxx].
> > > @kavya-b can you as part of the PR create an issue to capture this? Let’s get this in and plan the enhancement to this
> > 
> > 
> > Hi @fkolar , for textarea: cdk's auto resizing works only on rows, and expects min rows and max rows to be given. Fiori requirement does not allow user to specify min rows. Fiori requirement also specifies growing by height, in addition to growing by rows(or maxLines), and when both are specified, height should be given preference.
> > Since some very specific requirements are present, we didn't use cdk features.
> 
> I believe we can make use of cdk, if its by row or height it should not matter, you can work with both no. you always resize by row, you add or remove rows, you dont resize by px. But even in CDK you can specify heigh to set the limit.

@fkolar I can try this, but can it be done on a separate PR after this one as it seems to be a relatively bigger change?

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-basic-example.component.html
kavya-b: fixed

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-basic-example.component.html
droshev: @kavya-b any commented code won't work or won't be displayed. You can either uncomment and explain it or remove it

libs/platform/src/lib/components/form/text-area/text-area.component.html
```diff
@@ -0,0 +1,42 @@
...
+    <!-- <span>
```
droshev, line 25: remove the commented code

libs/platform/src/lib/components/form/text-area/text-area.component.scss
```diff
@@ -0,0 +1,37 @@
...
+    min-height: 0rem;
+}
+
+.fdp-textarea-normal-size {
+    max-height: inherit;
+    width: auto;
+    min-width: 0rem;
+}
+
+/* textarea counter in fund styles is used in the structure
```
droshev, line 14: @kavya-b remove commented code

libs/platform/src/lib/components/form/text-area/text-area.component.ts
```diff
@@ -0,0 +1,402 @@
+/**
+ * @license
```
droshev, line 2: can you remove this?

libs/platform/src/lib/components/form/text-area/text-area.component.ts
```diff
@@ -0,0 +1,402 @@
...
+     *  Can be `error` or `warning`.
+     */
+    @Input()
+    stateType: Status = 'warning';
+
+    /**
+     * return: The textarea's value
+     */
+    @Input() get value(): any {
+        // if (super.getValue() == null || !super.getValue()) {
```
droshev, line 126: @kavya-b remove the commented code

libs/platform/src/lib/components/form/text-area/text-area.component.ts
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    /** @hidden get the total height including borders and scroll height */
+    private _getTextareaTotalHeight(): number {
+        // Get the computed styles for the element
+        const computed = window.getComputedStyle(this.textareaElement.nativeElement);
+        // Calculate the height
+        const height =
+            parseInt(computed.getPropertyValue('border-top-width'), 10) +
+            //  + parseInt(computed.getPropertyValue('padding-top'), 10)
```
droshev, line 375: @kavya-b remove the commented code

libs/platform/src/lib/components/form/text-area/text-area.component.spec.ts
```diff
@@ -0,0 +1,336 @@
...
+
+        const textareaCounterElement = fixture.debugElement.query(By.css('.fd-textarea-counter'));
+        const result = textareaCounterElement.nativeElement.textContent.toString().trim();
+        expect(result).toBe('2 characters remaining');
+    });
+
+    // clipboard actions have stricter security and only work on secure https urls.
+    // https://developers.google.com/web/updates/2018/03/clipboardapi
+    // Therefore, ignoring this test.
+    xit('should handle paste interaction', async () => {
```
droshev, line 206: @kavya-b if you don't want to have this test then can you just remove it?

kavya-b: removed

libs/platform/src/lib/components/form/text-area/text-area.component.ts
kavya-b: removed

libs/platform/src/lib/components/form/text-area/text-area.component.ts
kavya-b: removed

libs/platform/src/lib/components/form/text-area/text-area.component.ts
kavya-b: removed

libs/platform/src/lib/components/form/text-area/text-area.component.scss
kavya-b: removed

libs/platform/src/lib/components/form/text-area/text-area.component.html
kavya-b: removed

apps/docs/src/app/platform/component-docs/platform-forms/platform-textarea/platform-textarea-examples/platform-textarea-basic-example.component.html
kavya-b: ok.. removed

PR #2696, opened 2020-06-17
fix: remove circular dependency on mobile modes

JKMarkowski:
relates: https://github.com/SAP/fundamental-ngx/issues/2674
There are included interfaces of components that have mobile mode. Also the way how we provide themselves is changed, to break import of each others there is InjectionToken

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/select/select.interface.ts
+++ b/libs/core/src/lib/select/select.interface.ts
@@ -1,0 +1,19 @@
+import { EventEmitter, InjectionToken } from '@angular/core';
+import { MobileModeConfig } from '../utils/interfaces/mobile-mode-config';
+import { OptionComponent } from './option/option.component';
+import { OptionStatusChange } from './select.component';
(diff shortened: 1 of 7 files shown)


libs/core/src/lib/select/select.interface.ts
```diff
@@ -11,8 +11,8 @@ export const SELECT_COMPONENT = new InjectionToken<string[]>('SelectInterface');
  */
 export interface SelectInterface {
     mobileConfig: MobileModeConfig;
```
fkolar, line 13: You dont want to push this code into select, as I have another PR in Progress

JKMarkowski: Hi Frank, we want to get rid of circular dependency for 0.19.0 version, which will be released soon. After it's done we can overwrite those changes by your changes.

fkolar: hmmm. ok.. 

My goal was push my changes latest today..

Issue #2481, opened 2020-05-06
docs: Add support for nested list in documentation side-nav


sKudum:
#### Is this a bug, enhancement, or feature request?
Enhancement
#### Briefly describe your proposal.

<img width="277" alt="Screen Shot 2020-05-06 at 2 39 17 PM" src="https://user-images.githubusercontent.com/53487468/81160365-9926f900-8fa7-11ea-83d7-3b2fd2e4349b.png">

Is not supporting nested list items display:

In the below formate:
<img width="211" alt="Screen Shot 2020-05-06 at 2 59 01 PM" src="https://user-images.githubusercontent.com/53487468/81162027-2c612e00-8faa-11ea-81c6-0e84bd7c7d8d.png">

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use the current version.)
9

droshev:
@sKudum you can find the API here https://sap.github.io/fundamental-ngx/#/core/sideNavigation/api

sKudum:
> @sKudum you can find the API here https://sap.github.io/fundamental-ngx/#/core/sideNavigation/api

we need to replace sectiontoolbar in documentation with sideNivagtion API which in turn will support nested items.

PR #2709, opened 2020-06-18
docs: use side navigation instead of sections toolbar

chrisdavisp:
docs issue request #2481


chrisdavisp:
Added support for SubItems

PR #2702, opened 2020-06-17
fix: add Panel component improvements

InnaAtanasova:
#### Please provide a brief summary of this pull request.
The PR for the new Panel component was merged before some comments by @salarenko were added. This PR addresses the following comments:

Screen Shot 2020-06-17 at 2 58 14 PM
Screen Shot 2020-06-17 at 3 07 51 PM

PR #2710, opened 2020-06-18
Fix panel expanded binding

salarenko:

Current Panel implementation does not support `expanded` two-way binding and includes duplicate `expanded` state settings inside PanelComponent logic, for example within `PanelExpandComponent`:

 => `PanelExpand.click()`
 => `set expanded state`
 => `updateExpanded`
 => `updateExpandedListener`
 => `set expanded state`

The problem was there was no way to distinguish if the expanded event is coming from outside or inside of the components and it has been resolved with adding `isExpandTriggerClick` parameter to internal expand event.

I think it is a little bit over-complicated but I cannot find other quick solution.


mikerodonnell89:
Two-way binding for "expanded" appears to be working fine on master, am I missing something?

salarenko:
> Two-way binding for "expanded" appears to be working fine on master, am I missing something?

Did you check this one?
https://github.com/SAP/fundamental-ngx/pull/2702#pullrequestreview-433146465

PR #2714, opened 2020-06-18
fix: Change addon place on multi input, remove doubled padding from dropdown list on multi input

JKMarkowski:
Related to: https://github.com/SAP/fundamental-ngx/issues/2674
There was addon button rendered in tokenizer, instead of input group.
Also the checkbox should now if it's used in list component, to apply paddings and change focus outline.
Before:
image

After:
image

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
+++ b/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
@@ -66,6 +66,10 @@ export class CheckboxComponent implements ControlValueAccessor {
     @Input()
     tristateSelectable: boolean = true;
 
+    /** Whenever the checkbox is used inside list, it applies another class to label */
+    @Input()
+    checkboxInList: boolean = false;
+
     /** Sets values returned by control. */
     @Input('values')
     set _values(checkboxValues: FdCheckboxValues) {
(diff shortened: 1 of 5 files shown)


libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
```diff
@@ -66,6 +66,10 @@ export class CheckboxComponent implements ControlValueAccessor {
     @Input()
     tristateSelectable: boolean = true;
 
+    /** Whenever the checkbox is used inside list, it applies another class to label */
+    @Input()
+    checkboxInList: boolean = false;
```
mikerodonnell89, line 71: Instead of using an input, can we just use `ContentChildr(CheckboxComponent)` from ListItemDirective and automatically add the class from there?

rengare: Indeed it would be better

JKMarkowski: Hi @mikerodonnell89  Actually, it's better solution, checkbox doesn't need to know if it's used inside list.

libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
```diff
@@ -34,6 +34,10 @@ export class CheckboxComponent implements ControlValueAccessor {
     @ViewChild('inputLabel', { static: false })
     inputLabel: ElementRef;
 
+    /** @hidden */
+    @ViewChild('labelElement', { static: false })
```
salarenko, line 38: This is `static: false` by default

PR #2719, opened 2020-06-19
fix: simplify Panel's implementation

InnaAtanasova:
This is a simplified version of the Panel component with less markup and no service

BREAKING CHANGES: Removed directives: `fd-panel-expand`, `fd-panel-header`, `fd-panel` selector. Before `[fd-panel]`, After:  `fd-panel`


libs/core/src/lib/panel/panel.component.ts
```diff
@@ -52,42 +47,37 @@ export class PanelComponent implements CssClassBuilder, OnChanges, OnInit, OnDes
     @HostBinding('attr.id')
     id: string = 'fd-panel-' + panelUniqueId++;
 
+    /** Id of the expand button */
+    @Input()
+    expandId: string = 'fd-panel-expand-' + panelExpandUniqueId++;
```
salarenko, line 52: Shouldn't this be applied as `[id]` parameter to expand button?

libs/core/src/lib/panel/panel.component.ts
```diff
@@ -106,14 +96,10 @@ export class PanelComponent implements CssClassBuilder, OnChanges, OnInit, OnDes
...
-            .subscribe(value => {
-                this.expanded = value.isExpanded;
-                this.expandedChange.emit(this.expanded);
-                this._cdRef.detectChanges();
-            });
+    /** Methods that toggles the Panel Content */
+    toggleExpand(): void {
+        this.expanded = !this.expanded;
+        this.expandedChange.emit(this.expanded);
+        this._cdRef.detectChanges();
```
salarenko, line 103: There is no need to manually force detect changes, it will be triggered with `click` event.

InnaAtanasova: Good point, removed it

libs/core/src/lib/panel/panel.component.ts
InnaAtanasova: Yes, I used it for aria-controls but it's good to have add the id too. Thanks

salarenko:
I think we are good here.
@InnaAtanasova, @droshev?

Issue #2538, opened 2020-05-17
Platform update button to match the core button API


droshev:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
The current platform button doesn't match the latest updates of the button from core platform.
The extensibility part is outdated: https://sap.github.io/fundamental-ngx/#/platform/button#Button-extensibility

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)

`@fundamental-ngx/platform` version `0.18.0`

sKudum:
Hi @Deno,

sKudum:
Hi @droshev ,
 As mentioned in [#2365 ](https://github.com/SAP/fundamental-ngx/pull/2365). 
![Uploading Screen Shot 2020-05-18 at 10.17.37 AM.png…]()
and it is dependent on issue:[2481](https://github.com/SAP/fundamental-ngx/issues/2481)

droshev:
The fdp button should support this:
```
If you'd like to add multiple options, you could use [options]="['emphasized', 'option1', 'option2']".
This would result in 3 classes being added to the button: fd-button--emphasized, fd-button--option1, and fd-button--option2.
```
it is against Fiori 3

sKudum:
> The fdp button should support this:
> 
> ```
> If you'd like to add multiple options, you could use [options]="['emphasized', 'option1', 'option2']".
> This would result in 3 classes being added to the button: fd-button--emphasized, fd-button--option1, and fd-button--option2.
> ```
> 
> it is against Fiori 3
> The issue is just targeting the button component and not the other button variations


Hi @Deno, 

Fiori3 has types and platform has  [type]="['emphasized', 'option1', 'option2']". it won't be having options going further in core as well.

This is considered by following the

1). fiori3 guideline.( these does not mention about options, it just list down the types.)

<img width="885" alt="Screen Shot 2020-05-18 at 4 42 29 PM" src="https://user-images.githubusercontent.com/53487468/82206917-adafad80-9926-11ea-83b5-59b7b52ce738.png">

2).Below PR changes done in Core. Which mentions fdOption will be deprecated.

<img width="826" alt="Screen Shot 2020-05-18 at 4 48 51 PM" src="https://user-images.githubusercontent.com/53487468/82207529-97562180-9927-11ea-8479-721eeb349bd4.png">

droshev:
More button issues:
[...]

PR #2720, opened 2020-06-19
feat:(platform) button is updated with core changes

sKudum:
#2538 

1). It has documentation details updated.
2). Aria Selected, area disabled added as properties for accessibility.
3). Test specs are updated with the same.
4). Class added for width truncation because in the latest (0.9.0 +)version truncation is not working as per Fiori3.same issue is observed in core.

BREAKING CHANGES:
Renamed properties:
Before :compact       After: contentDensity
Before: type               After:buttonType

Documentation checklist:
- [x] Documentation Examples

### DIFF
--- a/apps/docs/src/app/platform/component-docs/platform-action-bar/platform-action-bar-examples/platform-action-bar-with-long-title-example.component.html
+++ b/apps/docs/src/app/platform/component-docs/platform-action-bar/platform-action-bar-examples/platform-action-bar-with-long-title-example.component.html
@@ -1,18 +1,13 @@
-<fdp-action-bar
-    [contentDensity]="'compact'"
-    [title]="
+<fdp-action-bar [contentDensity]="'compact'" [title]="
         'A Very Long Page Title Which Should Be Truncated at Some Point with a Small Enough Page Width and a Long Enough Text String'
-    "
-    [showBackButton]="true"
-    [description]="
+    " [showBackButton]="true" [description]="
         'This Action Bar description should be composed with enough text to span the entire width of the Action Bar plus a good amount more such that it will force the Action Bar to truncate the text to a more managable size but not so much that it looses its meaning.'
-    "
->
+    ">
     <fdp-action-button-group>
-        <fdp-button [compact]="true" [type]="'emphasized'">
+        <fdp-button [contentDensity]="'compact'" [type]="'emphasized'">
             Save
         </fdp-button>
-        <fdp-button [compact]="true" [type]="'light'">
+        <fdp-button [contentDensity]="'compact'" [type]="'light'">
             Cancel
         </fdp-button>
     </fdp-action-button-group>
--- a/apps/docs/src/app/platform/component-docs/platform-action-bar/platform-action-bar-examples/platform-action-bar-with-long-title-example.component.html
+++ b/apps/docs/src/app/platform/component-docs/platform-action-bar/platform-action-bar-examples/platform-action-bar-with-long-title-example.component.html
@@ -4,10 +4,10 @@
         'This Action Bar description should be composed with enough text to span the entire width of the Action Bar plus a good amount more such that it will force the Action Bar to truncate the text to a more managable size but not so much that it looses its meaning.'
     ">
     <fdp-action-button-group>
-        <fdp-button [contentDensity]="'compact'" [type]="'emphasized'">
+        <fdp-button [contentDensity]="'compact'" [buttonType]="'emphasized'">
             Save
         </fdp-button>
-        <fdp-button [contentDensity]="'compact'" [type]="'light'">
+        <fdp-button [contentDensity]="'compact'" [buttonType]="'light'">
             Cancel
         </fdp-button>
     </fdp-action-button-group>
--- a/apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
+++ b/apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
@@ -1,8 +1,7 @@
 <fd-docs-section-title [id]="'Button-types'" [componentName]="'button'"> Button Types </fd-docs-section-title>
 <description>
-    The default button does not require modifier. Use
-    <code class="code-snippet">[type]="'standard'", [type]="'positive'", [type]="'medium'"</code>, or
-    <code class="code-snippet">[type]="'negative'"</code> to modify the type of the button.
+    The button component follows fundamental-styles for button guidelines. The default button type is "Default button",
+    no modifier is required. Use the <code class="code-snippet">[type]=""</code> property to set the button type.
 </description>
 <component-example>
     <fdp-button-types-example></fdp-button-types-example>
@@ -11,7 +10,9 @@
 <separator></separator>
 
 <fd-docs-section-title [id]="'Button-sizes'" [componentName]="'button'"> Button Sizes </fd-docs-section-title>
-<description> Button support two sized: compact and cozy  </description>
+<description> Button support two sized: compact and cozy. Use <code
+        class="code-snippet">[contentDensity]="'compact'"</code>
+    to set the values. </description>
 <component-example>
     <fdp-button-sizes-example></fdp-button-sizes-example>
 </component-example>
@@ -35,23 +36,44 @@
     Button State
 </fd-docs-section-title>
 <description>
-    <code>[disabled]="true"</code> sets the button disabled
+    'disabled' and 'selected' states can be achieved through 'disabled' / 'ariaDisabled' and 'ariaSelected'
+    attributes.
 </description>
 <component-example>
     <fdp-button-state-example></fdp-button-state-example>
 </component-example>
 <code-example [exampleFiles]="buttonHtmlState"></code-example>
 
-<!-- <separator></separator> -->
+<separator></separator>
 
-<!-- <fd-docs-section-title [id]="'Button-truncate'" [componentName]="'button'">
+<fd-docs-section-title [id]="'Button-truncate'" [componentName]="'button'">
     Button with long text truncated
 </fd-docs-section-title>
 <description>
-    When user specifices with width of the button then text will start truncating to accomdate it self in the specified
+    When user specifices with width of the button then text will start truncating to accomdate it self in the
+    specified
     width as shown below
 </description>
 <component-example>
     <fdp-button-truncate-example></fdp-button-truncate-example>
 </component-example>
-<code-example [exampleFiles]="buttonHtmlTruncate"></code-example> -->
+<code-example [exampleFiles]="buttonHtmlTruncate"></code-example>
+
+<separator></separator>
+
+<description>The inputs for <code>button</code> will be translated into classes attributed to fundamental-styles
+    guidelines.
+    <br />For example: <br />if <code>positive</code> is passed for <code>type</code>, the class
+    <code>fd-button--positive</code> will be added to the element. <br />if <code>compact</code> is passed for
+    <code>size</code>, the class <code>fd-button--compact</code> will be added to the element.
+</description>
+<separator></separator>
+
+<playground [schema]="schema" [schemaInitialValues]="fdpData" (onFormChanges)="onSchemaValues($event)">
+    <fdp-button [type]="fdpData.properties.type" [contentDensity]="fdpData.properties.contentDensity"
+        [ariaSelected]="fdpData.properties.ariaSelected" [ariaDisabled]="fdpData.properties.ariaDisabled"
+        [disabled]="fdpData.properties.disabled" [attr.title]="fdpData.properties.label"
+        [width]="fdpData.properties.width" [glyph]="fdpData.properties.icon">
+        {{ fdpData.properties.label }}
+    </fdp-button>
+</playground>
--- a/apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
+++ b/apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
@@ -1,7 +1,8 @@
 <fd-docs-section-title [id]="'Button-types'" [componentName]="'button'"> Button Types </fd-docs-section-title>
 <description>
-    The button component follows fundamental-styles for button guidelines. The default button type is "Default button",
-    no modifier is required. Use the <code class="code-snippet">[type]=""</code> property to set the button type.
+    The default button does not require modifier. Use <code class="code-snippet">[buttonType]="'emphasized'",
+    [buttonType]="'ghost'", [buttonType]="'positive'",[buttonType]="'attention'", [buttonType]="'transparent'" or
+    [buttonType]="'negative'" </code>to modify the type of the button.
 </description>
 <component-example>
     <fdp-button-types-example></fdp-button-types-example>
@@ -50,7 +51,7 @@
     Button with long text truncated
 </fd-docs-section-title>
 <description>
-    When user specifices with width of the button then text will start truncating to accomdate it self in the
+    When user specifices width of the button then text will start truncating to accomadate it self in the
     specified
     width as shown below
 </description>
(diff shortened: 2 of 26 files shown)


libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -16,14 +16,14 @@ export type ButtonType =
     templateUrl: './button.component.html',
     styleUrls: ['./button.component.scss']
 })
-export class ButtonComponent extends BaseComponent {
-    /** Option to make to button compact. */
-    @Input()
-    compact: boolean;
```
droshev, line 22: since you are removing this from the API the PR should start with `feat:`

droshev:
The documentation is not generated properly - the name of the component is all lower cases
Screen Shot 2020-06-20 at 2 19 41 PM

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
```diff
@@ -11,7 +10,8 @@
 <separator></separator>
 
 <fd-docs-section-title [id]="'Button-sizes'" [componentName]="'button'"> Button Sizes </fd-docs-section-title>
-<description> Button support two sized: compact and cozy  </description>
+<description> Button support two sized: compact and cozy. Use <code class="code-snippet">[contentDensity]="''"</code>
```
droshev, line 13: what is the value here?

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
```diff
@@ -1,8 +1,7 @@
 <fd-docs-section-title [id]="'Button-types'" [componentName]="'button'"> Button Types </fd-docs-section-title>
 <description>
-    The default button does not require modifier. Use
-    <code class="code-snippet">[type]="'standard'", [type]="'positive'", [type]="'medium'"</code>, or
-    <code class="code-snippet">[type]="'negative'"</code> to modify the type of the button.
+    The button component follows fundamental-styles for button guidelines. The default button type is "Default button",
```
droshev, line 3: I would not copy the documentation from the core package

apps/docs/src/app/platform/component-docs/platform-button/platform-button-examples/platform-button-sizes-example.component.html
```diff
@@ -1,2 +1,2 @@
 <fdp-button>Normal Button</fdp-button>
-<fdp-button [compact]="true">Compact Button</fdp-button>
+<fdp-button [contentDensity]="'compact'">Compact Button</fdp-button>
```
droshev, line 2: Can we consistent with core and use the same naming?

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
```diff
@@ -35,23 +35,35 @@
...
 </description>
 <component-example>
     <fdp-button-truncate-example></fdp-button-truncate-example>
 </component-example>
-<code-example [exampleFiles]="buttonHtmlTruncate"></code-example> -->
+<code-example [exampleFiles]="buttonHtmlTruncate"></code-example>
+
+<separator></separator>
+
+<description>The inputs for <code>button</code> will be translated into classes attributed to fundamental-styles
```
droshev, line 63: not needed of this paragraph as it is covered before

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.ts
```diff
@@ -27,12 +27,19 @@ export class PlatformButtonDocsComponent implements OnInit {
                         type: 'string',
                         enum: ['', 'standard', 'positive', 'transparent', 'negative', 'emphasized', 'ghost']
                     },
-                    compact: {
-                        type: 'boolean'
+                    contentDensity: {
```
droshev, line 30: let's align with core

libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -32,18 +32,24 @@ export class ButtonComponent extends BaseComponent {
     @Input()
     type: ButtonType;
```
droshev, line 33: this need to be reviewed as the native button has `type` which is not exposed in the platform button. I would not mix the fundamental button type with the native.

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.ts
sKudum: Hi @fkolar , base.ts has content Density for platform components. Please let me know your suggestion on this. So that I can update other comments accordingly.

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
sKudum: > The documentation is not generated properly - the name of the component is all lower cases
> Screen Shot 2020-06-20 at 2 19 41 PM

Updated
Screen Shot 2020-06-22 at 12 05 26 PM

sKudum: Screen Shot 2020-06-22 at 12 18 45 PM
Updated

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
sKudum: Removed

libs/platform/src/lib/components/button/button.component.ts
KevinOkamoto: We should rename the property `buttonType`.

KevinOkamoto: As for adding the other attributes, I don’t think we should. The “platform” is an opinionated framework of components, we try to shield the end-user from complexity and provide a system of components that just work with each other.

The “core” library is designed for flexibility. It tries to add Fiori 3 UX design to components without getting in the way of the end-user. It allows the end-user to have more control over the configuration of components.

I think this is a key difference between the two libraries.

In my opinion, we should try be minimalistic in our approach of exposing optional native attributes. Until we see a need from our end-users or within our library, we should refrain from exposing additional attributes unless we think them essential.

libs/platform/src/lib/components/button/button.component.ts
fkolar: @sKudum As deno say, already existing code which is out there and we do this kind of chaing its always breaking change.

fkolar:
@sKudum This is PR is not ready:

1) We aggred to have the API doc inside gitbook, let;s syncup in there, Formating is important tool .

2)

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.ts
fkolar: Even I have updated the title to go back to WIP mode so we can sync on the docs. we choose the name very carefully. 

This had its own evolution, where we started with size which conflicted with native attributes and then we try to find real meaning what compact and cozy really is. Developer can pass clear values `compact / cozy`.  If there will be in the future 3 option, like fluid , then have a type is safe approach. 

In the platform there are several differences on namings trying to put structure across all component.  

If we are able to find better name we are open to this. the` compact` just does not sound right for platform level. It needs to be a type with 2 possible values.

libs/platform/src/lib/components/button/button.component.ts
sKudum: Updated

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
sKudum: updated with [contentDensity]="''compact"

libs/platform/src/lib/components/button/button.component.ts
sKudum: updated to buttonType.

Other attributes like formaction, former type,form,autofocus,form method,formnovalidate,form target will be added in feature PRs . Need to explore on them more.

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.ts
sKudum: With respect to platform component consistency will be having it as content density.

apps/docs/src/app/platform/component-docs/platform-button/platform-button-examples/platform-button-sizes-example.component.html
sKudum: With respect to platform component consistency will be having it as content density.

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
sKudum: Updated [contentDensity]="'compact"

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.html
sKudum: removed

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.ts
sKudum: With respect to platform component consistency will be having it as content density.

apps/docs/src/app/platform/component-docs/platform-button/platform-button-examples/platform-button-sizes-example.component.html
sKudum: With respect to platform component consistency will be having it as content density.

sKudum:
> @sKudum This is PR is not ready:
> 
> 1. We aggred to have the API doc inside gitbook, let's sync up in there, Formating is important tool .

I updated the doc part after reviewing with Kevin and Manju, please do let me know your comments on this. I shared gitbook link in slack. (not sure whether we can share it here)

sKudum:
> Rename `type` property to `buttonType`.
Done

sKudum:
> Can you rebase your branch off master?

Done, 

can we use BREAKING CHANGE in commit-msg now? 
if versioning issue because of it is resolved.

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.ts
```diff
@@ -51,13 +58,14 @@ export class PlatformButtonDocsComponent implements OnInit {
...
             label: 'click here',
-            type: 'default',
-            option: 'default',
+            buttonType: 'default',
             width: '100px',
             size: 'default',
             icon: '',
-            compact: false,
-            disabled: false
+            contentDensity: 'false',
```
manu-kr, line 65: Why it is boolean value as it is type string you can initialised to empty string.

sKudum: It is added for examples, the default value will be cozy

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.ts
droshev: How about we align with the core where we decided to use input `compact`?

libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -16,34 +16,52 @@ export type ButtonType =
...
+
+    /** aria-disabled for acccesiblity of the element */
+    @Input()
+    ariaDisabled: boolean;
+
+    /** Specifies a name for the button */
+    @Input()
+    name: string;
+
+    /** Specifies the type of button */
```
droshev, line 52: I think it would be nice to describe that those inputs are setting attributes to the native HTML button

sKudum: Done

apps/docs/src/app/platform/component-docs/platform-button/platform-button-docs.component.ts
sKudum: I need confirmation from Frank to this change.

fkolar: We try to call the things right name even `contentDensity` might not be the best one, and I would like something with size we are just not comportable with just boolean compact. 

This way we could go to other names, where in combobox there is property dropdownValues which to me a its bit confusing and we call this differently as well. And there are many differences. 

We are definitelly open to suggestions maybe sizeFactor, but density is comon term. E.g. material has 3 values: 
https://material.io/design/layout/applying-density.html#components

libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -16,34 +16,57 @@ export type ButtonType =
...
+export class ButtonComponent extends BaseComponent implements AfterViewInit {
 
-    /** The icon to include in the button */
+    /** The icon to include in the button. See the icon page for the list of icons.
+     * Setter is used to control when css class have to be rebuilded.
+     * Default value is set to ''.
+     */
     @Input()
-    glyph: string;
+    public glyph: string;
```
stefanoScalzo, line 26: no need to add public, it is public by default

libs/platform/src/lib/components/button/button.module.ts
```diff
@@ -8,4 +8,4 @@ import { ButtonComponent } from './button.component';
     exports: [ButtonComponent],
     declarations: [ButtonComponent]
 })
-export class PlatformButtonModule {}
+export class PlatformButtonModule { }
```
stefanoScalzo, line 11: no need to include this file as it does not change anything

apps/docs/src/app/platform/component-docs/platform-action-bar/platform-action-bar-examples/platform-action-bar-with-long-title-example.component.html
```diff
@@ -1,18 +1,13 @@
-<fdp-action-bar
-    [contentDensity]="'compact'"
-    [title]="
+<fdp-action-bar [contentDensity]="'compact'" [title]="
         'A Very Long Page Title Which Should Be Truncated at Some Point with a Small Enough Page Width and a Long Enough Text String'
-    "
-    [showBackButton]="true"
-    [description]="
+    " [showBackButton]="true" [description]="
         'This Action Bar description should be composed with enough text to span the entire width of the Action Bar plus a good amount more such that it will force the Action Bar to truncate the text to a more managable size but not so much that it looses its meaning.'
```
kavya-b, line 4: more than 120 chars in a line, doesn't this throw a lint error?

libs/platform/src/lib/components/button/button.component.html
```diff
@@ -1,14 +1,6 @@
...
-    [attr.aria-label]="title"
-    [compact]="compact"
-    [glyph]="glyph"
-    [disabled]="disabled"
-    (click)="onBtnClick($event)"
-    [fdType]="type"
-    [ngStyle]="{ width: width }"
->
+<button #fdButton fd-button [attr.id]="id" [attr.title]="title" [attr.name]="name" [attr.aria-label]="title"
+    [attr.type]="type" [attr.value]="value" [attr.aria-selected]="ariaSelected" [attr.aria-disabled]="ariaDisabled"
```
kavya-b, line 2: is there an issue with the formatter? why is it wrapping it up?

sKudum: Done by Prettifier.

apps/docs/src/app/platform/component-docs/platform-action-bar/platform-action-bar-examples/platform-action-bar-with-long-title-example.component.html
sKudum: no lint errors for it thrown.

libs/platform/src/lib/components/button/button.component.ts
sKudum: done

libs/platform/src/lib/components/button/button.module.ts
sKudum: Prettifier suggested having a space inside {}. By which it shows in diffs.

libs/platform/src/lib/components/button/button.component.html
kavya-b: when i use the same piece of code in my vscode, prettier retains the older formatting. I guess you can still save without formatting, it looks more readable in the previous formatting. Right click -> Command Palette -> Save without formatting

Issue #2539, opened 2020-05-18
Platform Menu issues


droshev:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.

- styling looks broken

![Screen Shot 2020-05-17 at 8 56 36 PM](https://user-images.githubusercontent.com/4380815/82164893-eac16480-9880-11ea-807e-2b9b2f754bd8.png)
- the menu doesn't get open if another menu is open. Steps to reproduce:
  - open one menu -> it will open it
  - click on another menu -> the second menu doesn't get open but it will close the first one only

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
`@fundamental-ngx/platform` version `0.18.0`

KevinOkamoto:
Styling has been fixed by PR #2294

KevinOkamoto:
> Styling has been fixed by PR #2294

Looks like this is not fixed in the latest master. I'll need to take a look at it again.

Issue #2371, opened 2020-04-17
(Platform) Menu behavior bugs


salarenko:
#### Is this a bug, enhancement, or feature request?
This is a bug.

#### Briefly describe your proposal.
Menu component suffer for multiple behavior bugs and accessibility features.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
v0.16.0-rc.64

#### If this is a bug, please provide steps for reproducing it.
- Need two clicks to open another menu.
![platform-menu-1](https://user-images.githubusercontent.com/17496353/79556581-940a1480-80a1-11ea-87b1-e9e4ab289cac.gif)

-  Breaking sub-lists
![image](https://user-images.githubusercontent.com/17496353/79555933-a6378300-80a0-11ea-872c-258247e97c7a.png)

- Focus not following hover behavior
![image](https://user-images.githubusercontent.com/17496353/79555969-b2234500-80a0-11ea-8c61-3a904f970a71.png)

- Misplaced popover for RTL example
![image](https://user-images.githubusercontent.com/17496353/79556218-0fb79180-80a1-11ea-93b8-bbfd959bef7c.png)

- No focus trap

droshev:
@manjunathanagaraj some of the issues are still reproducible

droshev:
@manjunathanagaraj

manjunathanagaraj:
These are fixed.

PR #2722, opened 2020-06-19
fix: (platform) Bug fix for multiple menu triggers with same menu

KevinOkamoto:
#2371 
#2539 

For a scenario where a single menu is referenced by multiple trigger via the `menuTriggerFor` directive, there is a bug when clicking from one button to the next.

Clicking on the first button will open the menu. Clicking on the second button will then close the menu, without opening it from the second trigger element. This happens because the `closeMenu` method invoked from an "outside" click is resolved after the open menu call from the second button click, so the menu does not open the second time.

The solution was to add a `setTimeout` delay on the open menu call to insure it happens after the close menu is resolved.

Because I added a `setTimeout` I needed to add `fakeAsync` to all unit tests which make use of a menu trigger.


libs/platform/src/lib/components/menu/menu-trigger.directive.ts
```diff
@@ -124,7 +124,15 @@ export class MenuTriggerDirective implements OnDestroy, AfterContentInit {
...
     toggleMenu() {
-        this._isMenuOpen ? this.destroyMenu() : this.openMenu();
+        /**
+         * Need to add delay here to ensure that any "closeMenu" operation which
+         * has been invoked within an "outsideClickSubscription" gets resolved
+         * before "openMenu" is called. This can happen when there are multiple
+         * triggers for the same menu.
+         */
+        setTimeout(() => {
+            this._isMenuOpen ? this.destroyMenu() : this.openMenu();
```
fkolar, line 134: Can you use tick() instead ?

KevinOkamoto: I believe `tick()` is a method from Angular's testing module:
https://angular.io/api/core/testing/tick

I don't think there exist a `tick` function in native Javascript, unless you know of some other instance.

droshev:
@KevinOkamoto the styling is still broken.
Screen Shot 2020-06-26 at 6 18 56 PM

KevinOkamoto:
Menu styles have been fixed.

libs/platform/src/lib/components/menu/menu-trigger.directive.ts
fkolar: you are right. I though I was commenting test.

stefanoScalzo:
When menu is open, tabbing does not work

stefanoScalzo:
The cursor should change when hovering over the image (not sure if that is in the scope of this pr though)

KevinOkamoto:
@stefanoScalzo Tabbing issue has been fixed.

InnaAtanasova:
There are some minor styling issues that I noticed:
1. When an item is selected from the dropdown, the buttons gets a solid blue line:
Screen Shot 2020-07-31 at 3 30 41 PM
Screen Shot 2020-07-31 at 3 30 49 PM

2. In Cascading menu the menu items have rounded border corners:
Screen Shot 2020-07-31 at 3 31 29 PM
Screen Shot 2020-07-31 at 3 32 55 PM

3. In the example `Menu Within Scrolling Container` the Menu button looks a bit cut apart from having the solid blue border:
Screen Shot 2020-07-31 at 3 33 58 PM

libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -66,6 +66,13 @@ export class ButtonComponent extends BaseComponent implements AfterViewInit {
     @Output()
     buttonClicked = new EventEmitter();
```
InnaAtanasova, line 67: missing type

libs/platform/src/lib/components/menu/menu-item.component.ts
```diff
@@ -31,6 +31,8 @@ export class MenuItemComponent implements OnDestroy, FocusableOption {
 
     // Add Fundamental-Styles class for menu item
```
InnaAtanasova, line 32: These shouldn't be normal comments but typedocs

libs/platform/src/lib/components/menu/menu-item.component.ts
```diff
@@ -31,6 +31,8 @@ export class MenuItemComponent implements OnDestroy, FocusableOption {
...
     @HostBinding('class.fd-menu__item') menuItemClass = true;
+    @HostBinding('class.fd-menu__link') menuLinkClass = true;
+
```
InnaAtanasova, line 35: All hostbindings should have typedocs

libs/platform/src/lib/components/menu/menu-trigger.directive.ts
```diff
@@ -54,7 +54,7 @@ export class MenuTriggerDirective implements OnDestroy, AfterContentInit {
         private _viewContainerRef: ViewContainerRef,
         @Optional() @Self() private _menuItem: MenuItemComponent,
         @Optional() private _parentMenu: MenuComponent
-    ) {}
+    ) { }
 
     ngAfterContentInit() {
```
InnaAtanasova, line 59: should return void

libs/platform/src/lib/components/menu/menu-trigger.directive.ts
```diff
@@ -124,7 +124,15 @@ export class MenuTriggerDirective implements OnDestroy, AfterContentInit {
     }
 
     toggleMenu() {
```
InnaAtanasova, line 126: missing return type

libs/platform/src/lib/components/menu/menu-trigger.directive.ts
```diff
@@ -124,7 +124,15 @@ export class MenuTriggerDirective implements OnDestroy, AfterContentInit {
...
+         * has been invoked within an "outsideClickSubscription" gets resolved
+         * before "openMenu" is called. This can happen when there are multiple
+         * triggers for the same menu.
+         */
+        setTimeout(() => {
+            this._isMenuOpen ? this.destroyMenu() : this.openMenu();
+        }, 0);
     }
 
     openMenu() {
```
InnaAtanasova, line 138: missing return type

libs/platform/src/lib/components/button/button.component.ts
KevinOkamoto: Added missing type.

libs/platform/src/lib/components/menu/menu-item.component.ts
KevinOkamoto: Added typedocs.

libs/platform/src/lib/components/menu/menu-trigger.directive.ts
KevinOkamoto: Added return type.

libs/platform/src/lib/components/menu/menu-trigger.directive.ts
KevinOkamoto: Added return type.

libs/platform/src/lib/components/menu/menu-item.component.ts
KevinOkamoto: Added typedocs.

libs/platform/src/lib/components/menu/menu-trigger.directive.ts
KevinOkamoto: Added return type.

KevinOkamoto:
@InnaAtanasova I was able to fix the "blue" outline on focus of the `ftp-button` component host element.

InnaAtanasova:
Screen Shot 2020-08-04 at 4 18 25 PM

@KevinOkamoto What about this issue?

KevinOkamoto:
@InnaAtanasova The round corner issue has been fixed.

PR #2737, opened 2020-06-23
fix: pagination layout and dot hover issues

### LINKED ISSUE #2736: Pagination dots are underlined when hovered
Pagination dots are underlined when hovered.  Nothing should happen

EDIT: spacing between numbers is also off from styles.  Need to revisit template entirely

Screen Shot 2020-06-23 at 1 21 24 PM

mikerodonnell89:

fixes #2736 

Angular was stripping whitespace from the pagination template, which inline-block needs to display the pagination numbers correctly.  Also removes the pagination link class from the ellipses.

Before:
Screen Shot 2020-06-23 at 1 21 24 PM

After:
Screen Shot 2020-06-23 at 3 55 54 PM


mikerodonnell89:
> this PR touches too many files, can you check it @mikerodonnell89 ?

fixed

PR #2744, opened 2020-06-23
feat(select): Complete refactoring based on material common behavior

jbadan:
This pr is a copy of https://github.com/SAP/fundamental-ngx/pull/2619 which had to be removed from git history due to using breaking change in the commit message.


jbadan:
closing as we cannot have breaking change in a commit message https://github.com/SAP/fundamental-ngx/pull/2744/commits/f03545e490cd49a43a6a6de6f3f41d25eabd9c04

Issue #2387, opened 2020-04-20
[WIP] fix: (platform) Checkbox Group as a component


DeepakSap14:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
Creation of checkbox group component, which will create checkboxes based on list of values/objects or checkbox given. will bind them together as a form field. and should work with form controls.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 9

#### Is there anything else we should know?
This will be created in platform, and will be using checkbox platform to create checkboxes.

PR #2774, opened 2020-06-30
fix: (platform) Introduce checkbox group as component

DeepakSap14:
#2387 

This PR introduces checkbox Group as component in platform.
Platform Checkbox group can be created with list of values or by passing platform checkboxes.
It reduces complexity to create checkbox group in comparison to native way of creating groups.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


kavya-b:
1. keyboard navigation: on pressing enter, it should select/unselect the checkbox. Currently this is not happening
2. the error message displays randomly, and sometimes doesn't display at all.

DeepakSap14:
> 1. keyboard navigation: on pressing enter, it should select/unselect the checkbox. Currently this is not happening
> 2. the error message displays randomly, and sometimes doesn't display at all.
>    

1. Checkbox does not have requirement to check or un-check on Enter Button. In Accessibility for checkbox according to standards it has check or uncheck on space bar hit or clicking the checkbox.
2. Display of error message, have to investigate separately. This will be issue because of Form-field/form-group.

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.spec.ts
```diff
@@ -0,0 +1,451 @@
...
+    ];
+
+    form1 = new FormGroup({});
+    formData = { phones: ['Samsung', 'OnePlus'], visited: ['India', 'USA'], hobbies: ['coding', 'gardening'] };
+
+    @ViewChildren(CheckboxComponent)
+    checkboxGroups: QueryList<CheckboxGroupComponent>;
+}
+
+fdescribe('CheckboxGroup component Reactive Form Test', () => {
```
KevinOkamoto, line 78: change `fdescribe` to `describe` or else all other test will be skipped.

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.spec.ts
```diff
@@ -0,0 +1,451 @@
...
+    fruits = '';
+    languages = [
+        new LanguageKnown('Java', 'java', false),
+        new LanguageKnown('Javascript', 'javascript', true),
+        new LanguageKnown('Python', 'python', false),
+        new LanguageKnown('GoLang', 'go', true)
+    ];
+}
+
+fdescribe('Checkbox Group Component Template driven Form Tests', () => {
```
KevinOkamoto, line 283: Change `fdescribe` to `describe`

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -115,6 +139,7 @@ export class CheckboxComponent extends BaseInput implements AfterViewInit {
         private _ngZone: NgZone
     ) {
         super(_changeDetector, ngControl, ngForm);
+        this.name = `fdp-checkbox-${nextUniqueId++}`;
```
KevinOkamoto, line 142: Why are we setting `name` here? Isn't it an `@Input` field inherited from BaseComponent?

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.spec.ts
DeepakSap14: Done.
Thanks, had forgot to remove after debug.

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.spec.ts
DeepakSap14: Done.
Thanks, had forgot to remove after debug.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
DeepakSap14: in declarative checkbox-group, fdp-checkbox is passed under fdp-checkbox-group. so fdp-checkbox  is trying to create component, but id and name will not be available, and it fails to create because of baseInput check.
id and name are passed for fdp-checkbox-group not for fdp-checkbox.
So have to give some dummy value for id and name for fdp-checkbox. so component creation happens.
upon creation of fdp-checkbox-group, these get over written.

DeepakSap14:
> > 1. keyboard navigation: on pressing enter, it should select/unselect the checkbox. Currently this is not happening
> > 2. the error message displays randomly, and sometimes doesn't display at all.
> 
> 1. Checkbox does not have requirement to check or un-check on Enter Button. In Accessibility for checkbox according to standards it has check or uncheck on space bar hit or clicking the checkbox.
> 2. Display of error message, have to investigate. This will be issue because of Form-field/form-group.
>    will try to fix this issue in this PR.

for issue regarding, 
Display of error message: this need fix in core checkbox. Click on label is not stopped from propagation. so error message opens and close simultaneously. will raise different PR for fixing this.

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
KevinOkamoto: Hmm ... I feel like this would be a common problem with all of our components. I guess this is OK for now, but we should take a look at solving this problem at the BaseComponent or BaseInput level.

kavya-b:
1. Just a doubt, when checkboxes are in error state, the error state is getting applied to the disabled checkbox as well.
I am wondering if this is correct.. disabled won't have any action on it, so there is no point showing a state change either right?
2. I am seeing hover on disabled checkbox in a cbg randomly shows up the grab cursor. This issue is happening due to `cursor: pointer` getting applied to `fd-popover__control` class and it already has a fix of `cursor: auto` to override this behaviour, but is getting applied randomly. This might be out of context for this PR but should probably be looked into as it could affect other platform components used in a form context.

DeepakSap14:
> 1. Just a doubt, when checkboxes are in error state, the error state is getting applied to the disabled checkbox as well.
>    I am wondering if this is correct.. disabled won't have any action on it, so there is no point showing a state change either right?
> 2. I am seeing hover on disabled checkbox in a cbg randomly shows up the grab cursor. This issue is happening due to `cursor: pointer` getting applied to `fd-popover__control` class and it already has a fix of `cursor: auto` to override this behaviour, but is getting applied randomly. This might be out of context for this PR but should probably be looked into as it could affect other platform components used in a form context.

Issue 1: i personally think that, if checkboxes are as a group then should have same style. if disabled checkbox color does not change then it will look odd that enabled checkboxes and disabled checkboxes are are of different color. There is no hard coded rule for this. but o think, current implementation looks Good.

Issue 2 is fixed by giving importance to fd-popover__control class cursor property.

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.ts
```diff
@@ -0,0 +1,86 @@
...
+        @Optional() @Self() public ngForm: NgForm
+    ) {
+        super(_changeDetector, ngControl, ngForm);
+    }
+
+    writeValue(value: any): void {
+        if (value) {
+            super.writeValue(value);
+        }
+        this._changeDetector.detectChanges();
```
fkolar, line 75: Why are you calling detect changes when value could be NULL or undefined ? Why are you not using the stateChanges that is triggered in parent? Is this too late ?

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * raises event when Checkbox group value changes.
+     * @param event: contains checkbox and event in a PlatformCheckboxChange class object.
+     */
+    public groupValueChanges(event: PlatformCheckboxChange): void {
+        this.onTouched();
```
fkolar, line 83: Why did you decided to call OnTouched here on the onChange ?

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.spec.ts
```diff
@@ -0,0 +1,451 @@
...
+                    <fdp-checkbox [value]="'apple'" [label]="'Apple'" [disabled]="true"></fdp-checkbox>
+                    <fdp-checkbox [value]="'banana'" [label]="'Banana'"></fdp-checkbox>
+                    <fdp-checkbox [value]="'guava'" [label]="'Guava'" [disabled]="true"></fdp-checkbox>
+                    <fdp-checkbox [value]="'papaya'" [label]="'Papaya'"></fdp-checkbox>
+                </fdp-checkbox-group>
+            </fdp-form-field>
+        </fdp-form-group>
+    `
+})
+class TestReactiveCheckboxGroupComponnet {
```
fkolar, line 60: Also create test to use displayKey and Lookup Key

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.spec.ts
```diff
@@ -0,0 +1,451 @@
...
+
+    form1 = new FormGroup({});
+    formData = { phones: ['Samsung', 'OnePlus'], visited: ['India', 'USA'], hobbies: ['coding', 'gardening'] };
+
+    @ViewChildren(CheckboxComponent)
+    checkboxGroups: QueryList<CheckboxGroupComponent>;
+}
+
+describe('CheckboxGroup component Reactive Form Test', () => {
+    let host: TestReactiveCheckboxGroupComponnet;
```
fkolar, line 79: You want to also do more reactive tests where you initialize formGroup with values, I see you have object which is good. but FormGroup is needed as well.

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
```diff
@@ -6,7 +6,7 @@
         [label]="label"
         [tristate]="tristate"
         [tristateSelectable]="tristateSelectable"
-        [state]="status"
+        [state]="getUpdatedState()"
```
fkolar, line 9: We dont call methods in the template

fkolar:
I might left some comment instead of requestChange comment so please check if there are no additional comments from me

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.ts
DeepakSap14: error will only trigger when control is touched. By just changing the group value, touched property is not getting set.
so whenever change event is triggered, i am touching the control. so when the value is not there for required field, it will set error.

Is there any other way to do this?

libs/platform/src/lib/components/form/checkbox/checkbox.component.ts
```diff
@@ -3,21 +3,30 @@ import { Input, NgZone, Optional, Output, Self, ViewChild, ViewEncapsulation } f
...
 export class PlatformCheckboxChange {
     /** The source Checkbox of the event. */
     source: CheckboxComponent;
-    /** The new `checked` value of the checkbox. */
-    checked: boolean;
+    /**
+     * The new `checked` value of the checkbox.
+     * possible value: true/false and array of checkbox values.
+     */
+    checked: any;
```
stefanoScalzo, line 16: I think checked should stay as boolean and user will pass something like
[checked] = "whateverValue === condition"

JKMarkowski: It's checkbox value, so it can be any other type as well

apps/docs/src/app/platform/component-docs/platform-forms/platform-checkbox-group/platform-checkbox-group-docs.component.html
```diff
@@ -0,0 +1,41 @@
+<fd-docs-section-title [id]="'def'" [componentName]="'PlatformCheckboxGroupListComponent'">
```
salarenko, line 1: You are using the same `[id]` attribute for every documentation section.

DeepakSap14: fixed.

libs/platform/src/lib/components/form/checkbox/checkbox.component.html
DeepakSap14: fixed.
using stateType variable now.

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.spec.ts
DeepakSap14: included tests and examples for formGroup with values.

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.spec.ts
DeepakSap14: included tests and examples using displayKey and Lookup Key

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.ts
DeepakSap14: now using stateChanges.next()

stefanoScalzo:
I would have one message per checkbox in the validation example

fkolar:
> I would have one message per checkbox in the validation example

What do you mean? Maybe you missinterpretted usage for this component ?  This is a checkbox Group as one field and you cant have multiple messages. This is never the case

DeepakSap14:
> I would have one message per checkbox in the validation example

checkbox group is under one form-field. and currently one form-field has only one error message. so providing error message per checkbox is not possible currently.

DeepakSap14:
rebased with master.

stefanoScalzo:
As discussed with the team, if checkbox group is not in any fiori 3 examples, message designers on how to handle error messages

DeepakSap14:
Email Message from designers approves this implementation. Requirement for tooltip on checkbox will require change in core checkbox implementation.
Merging this PR.

Issue #2777, opened 2020-06-30
Defect Hunting - June 30, 2020


droshev:
List all the issues here:
Stefano takes Safari
Mike takes Chrome
Inna takes FF
Jedrzej takes IE11
Michal takes Edge

- [x] stackblitz for action bar - https://sap.github.io/fundamental-ngx/#/core/action-bar#actionbar-standard not working #2784
- [x] dialog with popover overflow issue - the popover doesn't go beyond the borders of the dialog even if it is bigger
![Screen Shot 2020-06-30 at 10 28 32 AM](https://user-images.githubusercontent.com/4380815/86138464-82d77e00-babc-11ea-916b-5949280a93bd.png)

Inna: Ii tested the current master and it looks like it's already fixed. Here's a screenshot with a similar example:
<img width="1454" alt="Screen Shot 2020-06-30 at 4 16 15 PM" src="https://user-images.githubusercontent.com/39598672/86173116-7a4b6b80-baed-11ea-8e25-27f3c609d6a1.png">

- [x] select has white corners - Inna - https://github.com/SAP/fundamental-ngx/pull/2779
![Screen Shot 2020-06-30 at 10 53 34 AM](https://user-images.githubusercontent.com/50607147/86141660-502f8480-bac0-11ea-8091-ff97b3dad570.png)

- [x] select has a double focus - fixed in #2796 
![Screen Shot 2020-06-30 at 10 53 45 AM](https://user-images.githubusercontent.com/50607147/86141647-4b6ad080-bac0-11ea-8b7e-bdf62fdf8c32.png)

JKMarkowski:
Checkbox - when user clicks fast on tristate enabled
![image](https://user-images.githubusercontent.com/26483208/86137140-38bac080-baed-11ea-9619-a9c85ec853c9.png)

Docs page logo is centered
![image](https://user-images.githubusercontent.com/26483208/86137170-407a6500-baed-11ea-9590-16f573a8bddf.png)

mikerodonnell89:
When cozy tokenizer receives focus, outline is cut off on right side
![Screen Shot 2020-06-30 at 8 14 52 AM](https://user-images.githubusercontent.com/2471874/86138234-739c0480-baab-11ea-8ca2-ec6658c49d45.png)

Need to remove status indicator

Split button needs top/bottom margin on examples

Buttons for example code copy and stackblitz need margin

Docs resize issue where at 992 pixels the content gets darkened but the side nav is normal

[...]

PR #2779, opened 2020-06-30
fix: remove rounded corners for the first element of Select component

InnaAtanasova:
Part of https://github.com/SAP/fundamental-ngx/issues/2777

The first element of the Popover body had rounded corners. They have been removed.


salarenko:
Can we for now scope this styling to select component? This will go global and I'm not sure how it will affect other components using `popover__body`.

PR #2787, opened 2020-07-01
fix: multi-input keyboard support, bug fixes, performance enhancements

### LINKED ISSUE #2739: Missing multi-input/tokenizer keyboard features
Some multi-input and tokenizer keyboard features available in UI5 are not available in our library.

- [ ] cmd+A to select all tokens (and backspace/delete to delete them.  After delete, previous token focused.  If no previous tokens, cursor goes to input)
- [ ] Arrow key to focus token works but space bar should select them
- [ ] Pressing enter when a token is focused removes focus from the token and moves the cursor to the input.  Maybe not intentional?

mikerodonnell89:
Fixes #2739 as well as some of the issues from defect hunting including the performance issue


libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -176,15 +184,23 @@ export class TokenizerComponent implements AfterViewChecked, AfterContentInit, O
 
     /** @hidden */
     handleTokenClickSubscriptions(): void {
+        this._unsubscribeClicks();
         this.tokenList.forEach((token, index) => {
-            token.onTokenClick.subscribe((event) => {
-                this.focusTokenElement(event, index);
-            });
+            this.tokenListClickSubscriptions.push(token.onTokenClick.subscribe(() => {
+                console.log('tokenIndex: ' + index);
```
JKMarkowski, line 190: Let's remove console log from lib

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -229,16 +245,39 @@ export class TokenizerComponent implements AfterViewChecked, AfterContentInit, O
...
 
+    /** @hidden */
+    @HostListener('document: click', ['$event'])
+    clickout(event) {
+        if (this.elementRef().nativeElement.contains(event.target) === false) {
+            this.tokenList.forEach(token => {
+                token.selected = false;
+            });
+        }
+    }
```
JKMarkowski, line 256: I remember having some document listeners on popover. It was very bad for performance, for more than 100 popovers. I hope no one will put more than 10 tokenizers in 1 page.

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -229,16 +245,39 @@ export class TokenizerComponent implements AfterViewChecked, AfterContentInit, O
...
     /** @hidden */
     handleKeyDown(event: KeyboardEvent, fromIndex: number): void {
         let newIndex: number;
         const rtl = this._rtlService && this._rtlService.rtl ? this._rtlService.rtl.getValue() : false;
-        if ((event.code === 'ArrowLeft' && !rtl) || (event.code === 'ArrowRight' && rtl)) {
+        if (KeyUtil.isKey(event, 'Space')) {
+            const token = this.tokenList.filter((element, index) => index === fromIndex);
+            this.tokenList.forEach(shadowedToken => {if (shadowedToken !== token[0]) {shadowedToken.selected = false}});
+            token[0].selected = !token[0].selected;
+            event.preventDefault();
```
JKMarkowski, line 266: Could you just unselect all tokens from tokenList and then switch current one? I'm not sure about keeping selected non-visible tokens.

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -229,16 +245,39 @@ export class TokenizerComponent implements AfterViewChecked, AfterContentInit, O
...
         } else if ((event.code === 'ArrowRight' && !rtl) || (event.code === 'ArrowLeft' && rtl)) {
             this._handleArrowRight(fromIndex);
             newIndex = fromIndex + 1;
+        } else if (event.code === 'KeyA' && this.input.elementRef().nativeElement.value === '') {
+            if (event.ctrlKey || event.metaKey) {
+                this.tokenList.forEach(token => {
+                    token.selected = true;
+                });
+            }
         }
```
JKMarkowski, line 281: Let's add stopPropagation or prevent default here

libs/core/src/lib/token/tokenizer.component.ts
mikerodonnell89: Not sure I totally understand what you're saying... do you mean we don't need the "if" at line 264?  If I am trying to unselect token[0] and it is selected...  If I unselect every token in the tokenList.forEach, then token[0] will be unselected, and then it will be re-selected after the next line

JKMarkowski: Actually you are right, it won't be possible to unselect. NVM.

stefanoScalzo:
When the user selects and clicks delete it should delete the item

stefanoScalzo:
There seems to be a bug when trying to select previous items and going all the way to the first item when there is x more

mikerodonnell89:
> There seems to be a bug when trying to select previous items and going all the way to the first item when there is x more

This is a non-issue, cozy mode is for mobile screens, test it there

Issue #2651, opened 2020-06-10
<fd-tab> can't be disabled


shaysap:
<fd-tab> can't be disabled. This functionality is very basic and required for many use cases.

shaysap:
Hey, I've upgraded to version 0.19.0 but still can't disable the tabs. Is there any way this can be done?

droshev:
@shaysap could you provide us with a business use case so we can follow up with the designers.
You can contact me directly in slack

droshev:
@shaysap feel free to reopen or create a new issue once you bring your business case

PR #2788, opened 2020-07-02
fix: add disabled state to tabs

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/2651
There is added disabled state for tabs. It's not shown in examples, but property is opened to be changed by passing `[disabled]="true"`

image
tabs

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/tabs/tab-list.component.ts
+++ b/libs/core/src/lib/tabs/tab-list.component.ts
@@ -117,8 +117,8 @@ export class TabListComponent implements AfterViewInit, OnChanges, OnDestroy {
     }
 
     /** @hidden */
-    tabHeaderClickHandler(tabIndex: number, disabled?: boolean): void {
-        if (this.selectedIndex !== tabIndex && !disabled) {
+    tabHeaderClickHandler(tabIndex: number): void {
+        if (tabIndex !== this.selectedIndex) {
             this.selectTab(tabIndex);
         }
     }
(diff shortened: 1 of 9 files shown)


JKMarkowski:
@InnaAtanasova I can't reproduce it, could you check again? Maybe it was some netlify cache

libs/core/src/lib/tabs/tab-list.component.ts
```diff
@@ -116,8 +117,8 @@ export class TabListComponent implements AfterViewInit, OnChanges, OnDestroy {
     }
 
     /** @hidden */
-    tabHeaderClickHandler(tabIndex: number): void {
-        if (this.selectedIndex !== tabIndex) {
+    tabHeaderClickHandler(tabIndex: number, disabled?: boolean): void {
```
salarenko, line 120: Wouldn't it be better to be consistent here and use `this._canBeSelected(tabIndex)` as well?

JKMarkowski: I think we can't have example of disabled in docs, because it's not designed by fiori3

InnaAtanasova:
Screen Shot 2020-07-06 at 11 44 50 AM

It's still reproducible, just apply `is-disabled` class to the `li` element, then click on the first tab, press the `tab` key to go to the next Tab, it gets the focus and it shouldn't, when you press Enter it selects the disabled Tab

PR #2796, opened 2020-07-02
fix: select double outline

mikerodonnell89:
fixes this issue from defect hunting
Screen Shot 2020-07-02 at 2 08 43 PM

### DIFF
--- a/libs/core/src/lib/select/select.component.html
+++ b/libs/core/src/lib/select/select.component.html
@@ -44,7 +44,7 @@
         class="fd-select__control"
         role="listbox"
         #selectControl
-        [ngClass]="state ? 'is-' + state : ''"
+        [ngClass]="(state ? 'is-' + state : '') + (panelOpen ? 'fd-select--hide-outline' : '')"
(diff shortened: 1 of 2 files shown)


libs/core/src/lib/select/select.component.html
```diff
@@ -44,7 +44,7 @@
         class="fd-select__control"
         role="listbox"
         #selectControl
-        [ngClass]="state ? 'is-' + state : ''"
+        [ngClass]="(state ? 'is-' + state : '') + (panelOpen ? 'fd-select--hide-outline' : '')"
```
InnaAtanasova, line 47: I think you need space here ' fd-select--hide-outline' otherwise will append to `is-state` 
You can probably use it this way:
[ngClass]="state ? 'is-' + state : ''"
[class.fd-select--hide-outline]="panelOpen"

salarenko:
I see we decided to bring back previous version of Select which doesn't have this problem. I think we can close this issue and PR for now.

Issue #2818, opened 2020-07-07
Remove hardcoded Fiori 3 table styles code


mikerodonnell89:
Assuming https://github.com/SAP/fundamental-ngx/pull/2819 is merged before the newest Styles version comes out, we'll need to remove the hardcoded code in table.component.scss

PR #2819, opened 2020-07-07
feat: implement fiori 3 table

### LINKED ISSUE #2610: Fiori 3 Table - initial port
We need to start bringing over the Fiori 3 table with some basic updates and additions before tackling the previously posted issues.  For this ticket we'll handle:

Updating table checkboxes
Support border removal
Add table footer directive
Add active/hover directives
Add status indicator directives

1.5 days

mikerodonnell89:

fixes #2610 

Breaking Changes: `fdColumnSortable` deprecated, along with associated input `[sortDir]`

Addresses all of the "priority 1" features we had previously discussed.  You may notice the new table SCSS hardcoded in this PR - this was intentional and will be pulled out when the new major styles version is released (see #2818)

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


apps/docs/src/app/core/component-docs/table/examples/table-pagination-example.component.html
```diff
@@ -0,0 +1,34 @@
...
+    </tbody>
+    <br/>
+    <tfoot fd-table-footer>
+        <button fd-button [fdMenu]="true" [fdMenuTrigger]="itemsPerPageMenu" style="margin-right: 20px">Items Per Page</button>
+        <fd-pagination
+            [totalItems]="totalItems"
+            (pageChangeStart)="newPageClicked($event)"
+            [itemsPerPage]="itemsPerPage"
+            [currentPage]="currentPage"
+        ></fd-pagination>
```
JKMarkowski, line 26: Pagination is still weird
image

apps/docs/src/app/core/component-docs/table/table-docs.component.ts
```diff
@@ -55,6 +74,136 @@ export class TableDocsComponent {
...
+
+    tableColumnSortExample: ExampleFile[] = [
+        {
+            language: 'html',
+            code: tableColumnSortHtmlSrc,
+            fileName: 'table-column-sorting-example'
+        },
+        {
+            language: 'typescript',
+            component: 'TableColumnSortExampleComponent',
```
JKMarkowski, line 157: Change component name to `TableColumnSortingExampleComponent`, Current one doesn't work on stackblitz

libs/core/src/lib/table/directives/table-cell-checkbox.directive.ts
```diff
@@ -0,0 +1,10 @@
+import { Directive, HostBinding, Input } from '@angular/core';
+
+@Directive({
+    selector: '[fdTableCellCheckbox], [fd-table-cell-checkbox]'
+})
+export class TableCellCheckboxDirective {
+    /** @hidden */
+    @HostBinding('class.fd-table__cell--checkbox')
+    fdTableCellClass: boolean = true;
+}
```
JKMarkowski, line 10: I think it will be better to keep this as a property `@Input() checkbox: boolean = false` on `fd-table-cell`. So on examples it would look like that
`<th fd-table-cell [checkbox]="true">`

libs/core/src/lib/table/directives/table-status-indicator.directive.ts
```diff
@@ -0,0 +1,17 @@
...
+        '[class.fd-table__cell--status-indicator--valid]': 'status === "valid"',
+        '[class.fd-table__cell--status-indicator--warning]': 'status === "warning"',
+        '[class.fd-table__cell--status-indicator--information]': 'status === "information"',
+        '[class.fd-table__cell--status-indicator--error]': 'status === "error"',
+    }
+})
+export class TableStatusIndicatorDirective {
+    /** The type of indicator. Options are 'valid', 'warning', 'information' and 'error'. */
+    @Input()
+    status: string;
```
JKMarkowski, line 16: Maybe let's add some `valid` | `warning` | `information` | `error` type here, or use existing declared type from forms

libs/core/src/lib/table/table.component.scss
```diff
@@ -1 +1,524 @@
...
+@charset "UTF-8";
+/*!
+.fd-table(--compact, --condensed, --no-horizontal-borders, --no-vertical-borders)
+    thead(.fd-table__header(--no-horizontal-borders, --no-vertical-borders))
+    tbody(.fd-table__body(--no-horizontal-borders, --no-vertical-borders))
+    tfoot(.fd-table__footer)
+        tr(.fd-table__row)([aria-selected])
+          td(.fd-table__cell(--status-indicator, --checkbox, --fit-content)) + semantic
+*/
+.fd-table {
```
JKMarkowski, line 24: Could you take latest styles there? pop-in table is fixed on current styles master.

apps/docs/src/app/core/component-docs/table/examples/table-pagination-example.component.html
mikerodonnell89: I agree this needs fixing but I'd consider it out of the scope of this PR

mikerodonnell89: mentioned in this defect hunting https://github.com/SAP/fundamental-ngx/issues/2777 i'll do this today

apps/docs/src/app/documentation/core-helpers/stackblitz/stackblitz.service.ts
```diff
@@ -131,6 +131,8 @@ import { Component } from '@angular/core';
     private getFileBasis(file: ExampleFile): string {
         if (file.service) {
             return file.fileName + '.service';
+        } else if (file.pipe) {
```
rengare, line 134: This could be introduced in another PR because other components may use it faster

mikerodonnell89:
> IE 11
> image

I believe this is a problem with modals/buttons, not table.  I will use shorter text in this example but this issue should be addressed separately from this PR

InnaAtanasova:
Some RTL issues:
- missing borders
- navigation arrow
Screen Shot 2020-07-09 at 2 38 21 PM
Screen Shot 2020-07-09 at 2 38 03 PM
- spacing of the button and the text
Screen Shot 2020-07-09 at 2 38 41 PM
Screen Shot 2020-07-09 at 2 38 48 PM

InnaAtanasova:
Shouldn't we include D&D or it will be part of another PR?

libs/core/src/lib/table/directives/table-row.directive.ts
```diff
@@ -1,10 +1,116 @@
...
+    }
+
+    /** @hidden */
+    ngOnDestroy(): void {
+        this.propagateKeysSubscription.unsubscribe();
+    }
+
+    /** @hidden */
+    private _resetCells(keys: string[]): void {
+        if (this.cells && keys && keys.length > 0) {
```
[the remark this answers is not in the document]
mikerodonnell89: I do not think we need to introduce a new function for this one line

libs/core/src/lib/table/directives/table-popover.directive.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+        TestBed.configureTestingModule({
+            declarations: [TestComponent],
+            imports: [TableModule]
+        }).compileComponents();
+    }));
+
+    beforeEach(() => {
+        fixture = TestBed.createComponent(TestComponent);
+        component = fixture.componentInstance;
+        fixture.detectChanges();
```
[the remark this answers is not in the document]
mikerodonnell89: why do we need this?

libs/core/src/lib/table/directives/table-cell.directive.ts
```diff
@@ -7,4 +7,47 @@ export class TableCellDirective {
...
+    @Input()
+    checkbox: boolean = false;
+
+    /** Key of cell element, it's used to identify this cell with certain column */
+    @Input()
+    key: string;
+
+    constructor (
+        public elementRef: ElementRef
+    ) {}
```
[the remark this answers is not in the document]
mikerodonnell89: It is used by table-row directive

mikerodonnell89:
> Some RTL issues:
> 
> * missing borders
> * navigation arrow

Fixed nav arrow, however the borders is an issue with styles

mikerodonnell89:
> Shouldn't we include D&D or it will be part of another PR?

Different PR, this falls under "priority 2"

stefanoScalzo:
I know its only an example but maybe add Validation.required to the new item in  Table with Toolbar

stefanoScalzo:
For table customization when you close without saving it still saves the previously entered values when u re open the options

stefanoScalzo:
Not sure if weve gone over this but for mobile mode shouldnt the hover be on the whole item?

mikerodonnell89:
> I know its only an example but maybe add Validation.required to the new item in Table with Toolbar

I don't think we need to bloat the examples with details like this

mikerodonnell89:
> I know its only an example but maybe add Validation.required to the new item in Table with Toolbar

> For table customization when you close without saving it still saves the previously entered values when u re open the options

Not a big deal to have stuff like this in the examples in my opinion. If I were reading these docs I'd rather have the examples be as succinct as possible, rather than tons of code checking this type of thing that's irrelevant to what I'm trying to learn

mikerodonnell89:
> Not sure if weve gone over this but for mobile mode shouldnt the select be on the whole item?

You mean for the click?  We did talk about this at some point, can't remember why we decided against it, @JKMarkowski can explain it

libs/core/src/lib/table/directives/table-popover.directive.spec.ts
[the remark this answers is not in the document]
rengare: this way you are sure that component finished rendering stage

libs/core/src/lib/table/directives/table-status-indicator.directive.ts
```diff
@@ -0,0 +1,19 @@
...
+    selector: '[fdTableStatusIndicator], [fd-table-status-indicator]',
+    host: {
+        class: 'fd-table__cell--status-indicator',
+        '[class.fd-table__cell--status-indicator--valid]': 'status === "valid"',
+        '[class.fd-table__cell--status-indicator--warning]': 'status === "warning"',
+        '[class.fd-table__cell--status-indicator--information]': 'status === "information"',
+        '[class.fd-table__cell--status-indicator--error]': 'status === "error"',
+    }
+})
+export class TableStatusIndicatorDirective {
```
salarenko, line 15: Is there an example showing usage of status indicator?

mikerodonnell89:
> * What is the reason for having `TableWrapperComponent`?

It will be needed for the drag-and-drop functionality to be able to add classes to the table.  Angular cdk table directive needs to be applied to table element (not fd-table).  See the commented out angular CDK example

apps/docs/src/app/core/component-docs/table/examples/table-checkboxes-example.component.ts
```diff
@@ -5,74 +5,131 @@ import { Component } from '@angular/core';
     templateUrl: './table-checkboxes-example.component.html'
 })
 export class TableCheckboxesExampleComponent {
-    selectMasterModel = false;
+    selectMasterModel: boolean = false;
```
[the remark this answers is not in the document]
mikerodonnell89: No idea, I'll change it to match the naming from checkbox docs

apps/docs/src/app/core/component-docs/table/examples/table-column-sorting-example.component.html
```diff
@@ -0,0 +1,51 @@
+<table fd-table>
+    <thead fd-table-header>
+        <tr fd-table-row>
+            <th fd-table-cell>
+                <fd-popover fd-table-popover [(isOpen)]="open">
+                    <fd-popover-control>
```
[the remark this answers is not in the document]
mikerodonnell89: This matches the sort/filter from the designers, but ultimately it's up to the application developer what they do

apps/docs/src/app/core/component-docs/table/examples/table-column-sorting-example.component.html
```diff
@@ -0,0 +1,51 @@
+<table fd-table>
+    <thead fd-table-header>
+        <tr fd-table-row>
+            <th fd-table-cell>
+                <fd-popover fd-table-popover [(isOpen)]="open">
```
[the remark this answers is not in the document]
mikerodonnell89: I don't think that makes sense here because you would have to use a button with [fdMenu] inputs etc.  And we don't want a button, we want the whole header cell to handle the click

libs/core/src/lib/table/directives/table-status-indicator.directive.ts
mikerodonnell89: Yes
Screen Shot 2020-07-10 at 10 38 05 AM

libs/core/src/lib/table/directives/table-row.directive.ts
```diff
@@ -1,10 +1,117 @@
...
+    HostBinding,
+    Input,
+    OnDestroy,
+    QueryList
+} from '@angular/core';
+import { TableCellDirective } from './table-cell.directive';
+import { TableService } from '../table.service';
+import { Subscription } from 'rxjs';
+
+export const Hidden_Class_Name = 'fd-table-hidden';
```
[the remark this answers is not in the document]
mikerodonnell89: I didn't write this, but perhaps for testing?

libs/core/src/lib/table/directives/table-body.directive.ts
```diff
@@ -7,4 +7,14 @@ export class TableBodyDirective {
     /** @hidden */
     @HostBinding('class.fd-table__body')
     fdTableBodyClass: boolean = true;
+
+    /** Whether or not to show the table body's horizontal borders */
+    @HostBinding('class.fd-table__body--no-horizontal-borders')
+    @Input()
+    noBorderX: boolean = false;
```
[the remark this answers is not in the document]
mikerodonnell89: I agree, but in this case I think the change needs to start from styles because the class that's being added is a negative.  So let's address this separately after that if we really think it's needed

Issue #2830, opened 2020-07-08
Avatar - new component


InnaAtanasova:
Avatar was introduced in Fundamental-styles a while ago. It's a component that will replace the Image and Identifier component. It's also needed for List with Byline
Dev estimation: 4 days

PR #2852, opened 2020-07-13
fix: introduce Avatar component

InnaAtanasova:
https://github.com/SAP/fundamental-ngx/issues/2830

This PR introduces Avatar component that was released in Fundamental-styles a while ago.
Also adds Deprecation alerts for Identifier and Image components

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


libs/core/src/lib/avatar/avatar.component.scss
```diff
@@ -0,0 +1,47 @@
+@import '~fundamental-styles/dist/avatar';
+
+// Temporary style fix until the latest version of Fundamental-styles is released
+fd-avatar[role='button'] {
+    cursor: pointer;
+}
+
+.fd-avatar--accent-color-1 {
```
JKMarkowski, line 8: Don't we have them on styles?

libs/core/src/lib/avatar/avatar.component.ts
```diff
@@ -0,0 +1,127 @@
...
+            this.transparent ? 'fd-avatar--transparent' : '',
+            this.placeholder ? 'fd-avatar--placeholder' : '',
+            this.tile ? 'fd-avatar--tile' : '',
+            this.class
+        ]
+            .filter((x) => x !== '')
+            .join(' ');
+    }
+
+    elementRef(): ElementRef<any> {
```
JKMarkowski, line 124: Could you add hidden comment, hiding it from documentation?

libs/core/src/lib/avatar/avatar.component.scss
InnaAtanasova: We do, but it's not released yet. The current styles version (in ngx) uses `has-color-`

libs/core/src/lib/avatar/avatar.component.ts
InnaAtanasova: Yes, thanks! I missed this one :D

Issue #2440, opened 2020-04-28
Fiori 3 table - drag and drop


mikerodonnell89:
Update table wrapper to work with new Fiori 3 table, and put back CDK example

PR #2855, opened 2020-07-13
fix: update table wrapper component to apply classes to thead/tbody/tfoot

mikerodonnell89:

#2440 

update table wrapper component to apply classes to thead/tbody/tfoot


JKMarkowski:
That's the problem in drag and drop. Could you try to catch element, when it's being dragged and add/remove some styling? I think having width evenly divided would be better than current.

tablednd

mikerodonnell89:
@JKMarkowski fixed

Issue #2198, opened 2020-03-18
Combobox - implement mobile mode w/ dialog


mikerodonnell89:
Mobile comboboxes should use fd-dialog, as per the last example here: https://sap.github.io/fundamental-styles/patterns/combobox-input.html

PR #2862, opened 2020-07-15
fix: Combobox Mobile

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/2198
There is introduced mobile mode for combobox. It follows strategy used in select and multi input.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


mikerodonnell89:
Doesn't make sense to have the input group addon dropdown icon here if it doesn't do anything, in my opinion

mikerodonnell89:
I also think the original input should remain visible after being clicked to open the mobile dialog

mikerodonnell89:
There are a handful of references to "multi-input" in the new code.  Wondering if you meant to use "combobox" there?  Possibly related note - i think, after merging this, we should open an issue for seeing what we can abstract between these two components.

JKMarkowski:
@mikerodonnell89 
> Doesn't make sense to have the input group addon dropdown icon here if it doesn't do anything, in my opinion

I agree with you, but designs keep addon even on mobiles

> I also think the original input should remain visible after being clicked to open the mobile dialog

There is no way to have templates of 2 same inputs at time (some id conflicts). Dialog with combobox is supposed to be on full screen, so user won't notice missing input.

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -483,21 +537,38 @@ export class ComboboxComponent implements ControlValueAccessor, OnInit, OnChange
         }
     }
 
-    private _handleClickActions(term): void {
+
+    /** @hidden*/
+    private _popoverOpenHandle(): void {
+        if (this._hasDisplayedValues) {
```
salarenko, line 543: Focus trap is based on whether component has `displayedValues`?

JKMarkowski: In combobox, the popover body is shown only, when there are displayed values. That's why I base focus trap on fact if it's open + if there are some values shown.

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -206,6 +222,14 @@ export class ComboboxComponent implements ControlValueAccessor, OnInit, OnChange
     @ContentChildren(ListMessageDirective)
     listMessages: QueryList<ListMessageDirective>;
 
+    /** @hidden */
+    @ViewChild('control', { read: TemplateRef })
```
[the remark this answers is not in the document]
JKMarkowski: Don't know why, but I was sure it's `ElementRef` by default.

stefanoScalzo:
I dont think approve should be the auto focus when it is first opened

stefanoScalzo:
There is a weird action that when the input group addon is clicked it removes the text

apps/docs/src/app/core/component-docs/combobox/combobox-docs.component.html
```diff
@@ -41,6 +41,26 @@
...
 <separator></separator>
 
+<fd-docs-section-title [id]="'combobox-mobile'" [componentName]="'combobox'">
+    Combobox Mobile Mode
+</fd-docs-section-title>
+<description>
+    There is way to use mobile mode on combobox component. To achieve it user needs to:
+    <ul>
+        <li>Set <code>[mobile]</code> property to true</li>
+        <li>Provide <code>DIALOG_CONFIG</code></li>
```
stefanoScalzo, line 51: what if when user provides dialog config it automatically sets mobile to true and mobile is a hidden property. Since he is forced to provide both might as well only force user to provide one

apps/docs/src/app/core/component-docs/combobox/examples/combobox-example.component.ts
```diff
@@ -11,5 +11,11 @@ export class ComboboxExampleComponent {
     searchTermFour: string = '';
     searchTermFive: string = '';
 
-    dropdownValues = ['Apple', 'Pineapple', 'Banana', 'Kiwi', 'Strawberry'];
+    dropdownValues = [
```
stefanoScalzo, line 14: remove this file, it is only a refactor

JKMarkowski: I think this format is better

apps/docs/src/app/core/component-docs/combobox/combobox-docs.component.html
JKMarkowski: Dialog config doesn't have impact on fact if mobile is set on combobox.

libs/core/src/lib/combobox/combobox.component.ts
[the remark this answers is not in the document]
salarenko: It depends on the type of element you are targeting with `@ViewChild`

stefanoScalzo:
Maybe we should contact designers for the drop-down icon button, because having it there with no functionality looks like a bug, everything else looks good to me

PR #2870, opened 2020-07-16
fix: bug where initial hidden token count was wrong on cozy multi-input

### LINKED ISSUE #2566: "Show All (x)" Feature in Multi Input Control
Dear Fundamentals Team,

we are using the component "Multi Input" quite intensively.
https://sap.github.io/fundamental-ngx/#/core/multi-input

But we haven't figured out yet what the additional feature of "Show All (x)" is supposed to do. We do not see any documentation. Playing around with the provided examples on the page mentioned above ends nowhere.

We kindly ask to point us to proper documentation of that feature or to briefly explain here what it is doing.

Many thanks,
Kai
image

### LINKED ISSUE #2853: Multi-Input: "___ more" text count incorrect on initial load (cozy only)

The initial count for the "more" text on the multi input in cozy mode is incorrect.  As you can see in the screenshot, 7 items are selected, but the text says there are only 3 more in addition to the 3 visible in the multi-input.  If you resize the screen, then back to this size, the count will be correct.

Screen Shot 2020-07-13 at 2 42 16 PM

mikerodonnell89:
fixes #2853 where the "___ more" count was incorrect on first load for cozy multi-inputs

fixes #2566 improved behavior (showing/hiding) of "more" text


libs/core/src/lib/token/token.component.html
```diff
@@ -1,4 +1,4 @@
-<span
+<span #spanElement
```
InnaAtanasova, line 1: #spanElement is not the best naming for this considering the fact there are another 2 spans :)
The rest looks gook!

salarenko: Why did you decided to base on `span` reference? `ElementRef` was not enough?

mikerodonnell89: @salarenko I don't understand, what is your suggested alternative here?

salarenko: I'm just asking why did you switched from this solution?
image

mikerodonnell89: Because the value returned from that included the margin which we don't want for this calculation

droshev:
@mikerodonnell89 there are few issues with multi input.
here is the first:
multi-input-xmore
We see `1 more` even though if you scroll you see all items there

droshev:
@mikerodonnell89 there is weird inconsistency with the items that are shown when you open the menu.
1. go to https://deploy-preview-2870--fundamental-ngx.netlify.app/#/core/multi-input#simpleMultiInput and click on 1 `more` and you will see all items including the non-selected `Orange`.
2. close it and open it again by clicking on `1 more` and when the menu open you won't see `Orange`

mikerodonnell89:
Fixed the issue described in Deno's second comment but we discussed separately that the situation explained in the first comment is fine as-is.  This situation becomes complicated as it's unclear whether or not a partially-visible token should be applied to the "more" count.  Will probably have to talk to designers about this.

JKMarkowski:
It works great, but there is one thing missing, can multi input become non-scrollable with `__more`, when it's not active? So basically same as 
image

PR #2873, opened 2020-07-16
docs: demonstrate variant switching with table

### LINKED ISSUE #2872: Demonstrate variant switching with table docs
Table priority 2 requires variant switching.  This should not require any changes to the table and will be handled in the documentation

mikerodonnell89:

fixes #2872 

Demonstrates variant switching on table

### DIFF
--- a/apps/docs/src/app/core/component-docs/table/examples/table-toolbar-example.component.ts
+++ b/apps/docs/src/app/core/component-docs/table/examples/table-toolbar-example.component.ts
@@ -1,22 +1,31 @@
-import { Component, OnInit, TemplateRef } from '@angular/core';
-import { DialogService } from '@fundamental-ngx/core';
-import { FormBuilder, FormControl, FormGroup, Validators } from '@angular/forms';
+import { Component, OnInit, TemplateRef, ViewChild } from '@angular/core';
+import { DialogService, MenuComponent } from '@fundamental-ngx/core';
+import { FormBuilder, FormControl, FormGroup } from '@angular/forms';
 
 @Component({
     selector: 'fd-table-toolbar-example',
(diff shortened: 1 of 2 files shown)


apps/docs/src/app/core/component-docs/table/examples/table-toolbar-example.component.ts
```diff
@@ -87,4 +108,19 @@ export class TableToolbarExampleComponent implements OnInit {
         );
     }
 
+    variantChange(value: string): void {
+        this.variantMenu.close();
+        this.selectedVariant = value;
+        this.searchTerm = '';
+        if (value === 'Produce') {
```
salarenko, line 115: I think you can use here `getSelectedVariantRows()`

apps/docs/src/app/core/component-docs/table/examples/table-toolbar-example.component.ts
```diff
@@ -54,16 +75,16 @@ export class TableToolbarExampleComponent implements OnInit {
 
     searchInputChanged(event: string): void {
         if (event) {
-            this.displayedRows = this.tableRows.filter(row => {
+            this.displayedRows = this.getSelectedVariantRows().filter(row => {
                 return JSON.stringify(row).toLowerCase().indexOf(event.toLowerCase()) !== -1;
```
salarenko, line 79: You are sorting here also based on object properties names, which causes filtering bugs (try to filter by '1')

salarenko: It would be also cool to have such base filtering login inside the table, but thats probably the topic for future discussion

Issue #2867, opened 2020-07-16
fix: (platform) Introduce Switch in Platform


Betrozov:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
Create the Switch component that is a UI element which enables users to turn things "on" and "off".

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 9.1.11

#### If this is a bug, please provide steps for reproducing it.
No, Its not a bug.

#### Please provide relevant source code if applicable.
NA

PR #2918, opened 2020-07-28
fix: (platform) add switch component

Betrozov:
fixes https://github.com/SAP/fundamental-ngx/issues/2867

Created the Platform Switch component

Documentation checklist:
- [X] Documentation Examples


apps/docs/src/app/platform/component-docs/platform-forms/switch/switch-header/switch-header.component.html
```diff
@@ -0,0 +1,8 @@
+<header>Switch</header>
+<description>
+    <p>Switch description</p>
+</description>
+<import module="SwitchModule"></import>
```
fkolar, line 5: Missing actual description what we usually add to API docs:

Checkout other components in the platform where we not try to say the same what you can read in core but at minimum how we are enhancing this.

libs/platform/src/lib/components/form/switch/switch/switch.config.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ContentDensity } from '../../../form/form-control';
+
+/**
+ * Default options for platform Switch
+ */
+@Injectable({ providedIn: 'root' })
+export class SwitchConfig {
+    /**
+     * Content Density of element. 'cozy' | 'compact'
+     */
```
fkolar, line 12: We could have more generic config as content density can be applied to all the components like this. I remember Dima was showing me `PlatformConfig`

valorkin: It wasn't yet approved, so let's stick with component level configs, and later we could came up with platform/core/ or even all level configs and services

fkolar: sounds good if its one component but if we need to keep config for each new component like this, I would go in the next iteration to more generic approach and refactor this.

valorkin: components specific configs are must have anyway, over time they will have even more properties

JKMarkowski:
Hi @Betrozov,
I see that whole switch component is implemented from scratch, wasn't it supposed to use implementation from core?

fkolar:
There is missing disabled case in fundamentals- styles

libs/platform/src/lib/components/form/switch/switch/switch.component.ts
```diff
@@ -0,0 +1,129 @@
...
+    templateUrl: './switch.component.html',
+    styleUrls: ['./switch.component.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    encapsulation: ViewEncapsulation.None,
+    providers: [{ provide: FormFieldControl, useExisting: SwitchComponent, multi: true }]
+})
+export class SwitchComponent extends BaseInput {
+    /** aria-label attribute of the inner input element. */
+    @Input()
+    ariaLabel = 'Switch input';
```
JKMarkowski, line 38: I know it's detail, but could you add type there?

libs/platform/src/lib/components/form/switch/switch/switch.component.ts
```diff
@@ -0,0 +1,129 @@
...
+     */
+    @Input()
+    set contentDensity(contentDensity: ContentDensity) {
+        this._contentDensity = contentDensity;
+        this.isCompact = contentDensity === 'compact';
+    }
+
+    /** Whether the switch is semantic */
+    @Input()
+    semantic = false;
```
JKMarkowski, line 55: Let's add missing type

libs/platform/src/lib/components/form/switch/switch/index.ts
```diff
@@ -0,0 +1 @@
+export * from './switch.component';
```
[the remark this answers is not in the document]
Betrozov: it is for decreasing the path when we import this component.

libs/platform/src/lib/components/form/switch/switch/switch.component.ts
fkolar: In this you dont really need to. as you assign a string, this is default behavior for Lint Rules. We dont have general guidelines for this and this is some thing to be noted as well and decide on the direction. I would leave this as it is as this is also proper form.

libs/platform/src/lib/components/form/switch/switch/index.ts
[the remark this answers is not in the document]
fkolar: Its common practice to use barrel files.

apps/docs/src/app/platform/platform-documentation.routes.ts
```diff
@@ -4,91 +4,97 @@ import { PlatformHomeComponent } from './component-docs/platform-home/platform-h
...
-            {
-                path: 'textarea',
-                loadChildren: () =>
-                    import('./component-docs/platform-forms/platform-textarea/platform-textarea-docs.module').then(
-                        (m) => m.PlatformTextareaDocsModule
-                    )
-            }
-        ]
-    }
+  {
```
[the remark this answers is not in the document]
Betrozov: fixed

Betrozov:
> The styling for disabled looks not proper
> in platform
> Screen Shot 2020-07-28 at 7 33 32 AM
> 
> in core
> Screen Shot 2020-07-28 at 7 31 57 AM

fixed

apps/docs/src/app/platform/component-docs/platform-forms/switch/switch-header/switch-header.component.html
Betrozov: fixed

Betrozov:
> I added some comments regarding code.
> There is also different focus ring on switch in platform.
> Core:
> image
> 
> Platform:
> image

fixed

I074914:
Behaviour of switch toggle is not working as expected. When we click on the "Default Semantic Compact Switch" and "Default Semantic Compact Switch" on first mouse click it go highlighted. and second click it will work. 
Expected: On first mouse click toggle should work.

I074914:
> > Behaviour of switch toggle is not working as expected. When we click on the "Default Semantic Compact Switch" and "Default Semantic Compact Switch" on first click it is highlighted and second click it will work. But the issue is on first click switch toggle is not working.
> > Expected: On first mouse click toggle should work.
> 
> Fixed

Tested it is working fine. Thanks

manjunathanagaraj:
Looks fine to me. Accessbility testing will be verified by Purushotham.

PR #2924, opened 2020-07-28
fix: multi cozy token count bug

### LINKED ISSUE #2566: "Show All (x)" Feature in Multi Input Control

mikerodonnell89:
fixes #2566 

Addresses visual issues that arose due to changes in list component: see the bottom of the discussion on the linked issue


mikerodonnell89:
> Clicking left ?
> clicking_left

This is the cozy token - for mobile devices - you don't have any arrow keys

libs/core/src/lib/list/list.component.scss
```diff
@@ -376,16 +376,13 @@
 }
 .fd-list--dropdown .fd-list__item {
     height: auto;
-    padding: .75rem 1rem
+    padding: .75rem 1rem;
```
salarenko, line 379: These style changes are an update from fundamental-styles?

mikerodonnell89: > I don't think this is acceptable that MultiInput behaves differently based on cozy/compact mode. I think we should take care of it in separate ticket #2948.

@salarenko this is directly from the interaction specification, I've had a few conversations about it with the designers and this is what they want

JKMarkowski:
could you change color of link element inside `fd-list__item:active` in this multi input to inherit? 
image

PR #2937, opened 2020-07-30
fix: Update class builder

salarenko:
Previous `@applyCssClass` implementation was overriding native element styling with styles coming from specified component. This behavior was problematic because native element classes might come from other sources than the original component, so these classes would be wiped out by `@applyCssClass`.

The proposition is replacing only these classes that have been originally added by the component, leaving the rest of native elements classes untouched.

Proposed solution is a bit more complex but I don't think it will have significant impact if we operate on small arrays and re-run onPushChanges only.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


libs/core/src/lib/button/button.component.ts
```diff
@@ -38,10 +38,9 @@ export type ButtonType =
     template: ` <ng-content></ng-content> `,
     styleUrls: ['./button.component.scss'],
     encapsulation: ViewEncapsulation.None,
-    changeDetection: ChangeDetectionStrategy.OnPush
+    changeDetection: ChangeDetectionStrategy.OnPush,
```
InnaAtanasova, line 41: I don't think you need the `,` here

InnaAtanasova:
Screen Shot 2020-07-31 at 2 16 38 PM

Screen Shot 2020-07-31 at 2 10 06 PM


When a custom class is applied on a directive we get this class double. For components it works ok.

salarenko:
@InnaAtanasova I updated my solution

InnaAtanasova:
Screen Shot 2020-08-03 at 5 05 54 PM
I tested in StackBlitz and the double class is still there

Issue #2940, opened 2020-07-30
fix: (core) configure angular.json for linting


Betrozov:
#### Is this a bug, enhancement, or feature request?
enhancement
#### Briefly describe your proposal.
At the moment Tslint works only for docs.
Need to configure `angular.json` lint to work with `core` and `platform`
#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 10
#### If this is a bug, please provide steps for reproducing it.
NA
#### Please provide relevant source code if applicable.
NA
#### Is there anything else we should know?
NA

salarenko:
Closed by #2941

PR #2941, opened 2020-07-30
fix: (core) update angular.json for tslint

Betrozov:
https://github.com/SAP/fundamental-ngx/issues/2940

Angular linter works for docs, core and platform

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -137,7 +137,6 @@
                             "apps/docs/tsconfig.spec.json"
                         ],
                         "exclude": [
-                            "**/node_modules/**",
                             "!apps/docs/**"
                         ]
                     }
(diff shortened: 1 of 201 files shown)


angular.json
```diff
@@ -184,8 +183,7 @@
                     "options": {
                         "tsConfig": "apps/docs-e2e/tsconfig.e2e.json",
                         "exclude": [
-                            "**/node_modules/**",
```
salarenko, line 187: Why you removed node_modules matcher from excludes? I don't think we want to run linter over packages.

Betrozov: `!apps/docs-e2e/**` as I see, this string means that we exclude everything except `apps/docs-e2e`

salarenko: Makes sense 😋

PR #2946, opened 2020-07-30
feat: introduce generic tile

mikerodonnell89:
fixes #2819 

Slide and line tile will be addressed in separate PRs


stefanoScalzo:
focus is missing on click of tile

stefanoScalzo:
Can you have an example of kip tile where the status is different

libs/core/src/lib/badge/badge.component.ts
```diff
@@ -0,0 +1,9 @@
+import { Component, ViewEncapsulation } from '@angular/core';
+
+@Component({
+    selector: 'fd-badge',
+    template: `<div class="fd-badge"><ng-content></ng-content></div>`,
```
JKMarkowski, line 5: Maybe let's make it same as button, inline component with `[fd-badge]` selector.

libs/core/src/lib/tile/directives/numeric-content.directives.ts
```diff
@@ -0,0 +1,280 @@
+import { Directive, ElementRef, HostBinding, Input, OnChanges, OnInit } from '@angular/core';
+import { applyCssClass } from '../../utils/decorators/apply-css-class.decorator';
+
+type NumericContentState = 'negative' | 'critical' | 'positive' | 'informative' | '' | 'neutral';
+type NumericContentSize = 's' | 'm' | 'l' | '';
```
JKMarkowski, line 5: Is there a need to add empty `''` strings into types? 
In case of no value I would suggest to keep null value, instead of empty string

libs/core/src/lib/tile/tile.component.html
```diff
@@ -1,3 +1,12 @@
-<ng-content select="[fd-tile-media]"></ng-content>
-<ng-content select="[fd-tile-content]"></ng-content>
-<ng-content></ng-content>
+<div class="fd-tile" #tileDiv>
```
JKMarkowski, line 1: What about putting there `tabindex="0"`? Maybe you can add some flag to tile component, that will control tabindex of this element

libs/core/src/lib/tile/tile.component.ts
```diff
@@ -1,61 +1,70 @@
...
+
 @Component({
+    // tslint:disable-next-line:component-selector
     selector: 'fd-tile',
-    host: {
-        '[attr.role]': '(this.isButton === true ? \'button\' : \'\')'
-    },
     templateUrl: './tile.component.html',
     styleUrls: ['./tile.component.scss'],
     encapsulation: ViewEncapsulation.None
```
JKMarkowski, line 12: Let's add on push strategy there

libs/core/src/lib/badge/badge.component.ts
mikerodonnell89: So would it be a div or a span? What is the advantage here? Makes sense for an actionable element like button but i don't think so for this one

libs/core/src/lib/tile/tile.component.html
```diff
@@ -1,3 +1,12 @@
-<ng-content select="[fd-tile-media]"></ng-content>
-<ng-content select="[fd-tile-content]"></ng-content>
-<ng-content></ng-content>
+<div class="fd-tile" #tileDiv tabindex="0">
+    <div *ngIf="action && !(type === 'line')" class="fd-tile__overlay"></div>
```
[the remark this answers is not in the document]
mikerodonnell89: haha not sure what i was thinking there

Issue #2962, opened 2020-08-02
Product switch popover arrow not present


droshev:
Product switch popover arrow not present

PR #2963, opened 2020-08-02
fix: fix product switch arrow

droshev:
fixes: #2962 

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


libs/core/src/lib/product-switch/product-switch/product-switch.component.html
```diff
@@ -8,6 +8,7 @@
         [disabled]="disabled"
         [triggers]="triggers"
         [focusTrapped]="false"
+        [noArrow]="noArrow"
```
stefanoScalzo, line 11: it needs to take a property in the .ts file that is noArrow

mikerodonnell89: This should be [noArrow]="false"

apps/docs/src/app/core/component-docs/product-switch/examples/product-switch-dnd-example.component.html
```diff
@@ -1,6 +1,6 @@
 <fd-shellbar>
     <fd-shellbar-actions>
-        <fd-product-switch>
+        <fd-product-switch [noArrow]="false">
```
mikerodonnell89, line 3: this change should be undone

libs/core/src/lib/product-switch/product-switch/product-switch.component.html
droshev: @mikerodonnell89 don't we want to keep it as a parameter, instead of always hardcoding it?

Issue #2939, opened 2020-07-30
0.21.0 Defect Hunting - 30.07.2020


JKMarkowski:
- [x] Checkbox on IE11 - Difference between fast/slow clicking - https://github.com/SAP/fundamental-ngx/pull/2949
![checkboxx](https://user-images.githubusercontent.com/26483208/88909370-4cc42000-d25b-11ea-93a6-63bbe26fbbc4.gif)

- [x] Combobox crashes on IE11 - 1) try to pick something from first combobox, 2) Open this combobox, 3) click on other combobox - try to blur - https://github.com/SAP/fundamental-ngx/pull/2966
![comboboxie11](https://user-images.githubusercontent.com/26483208/88909931-03c09b80-d25c-11ea-8d87-cc5c801308ed.gif)

- [x] Example of warning input - https://fundamental-ngx.netlify.app/#/core/input#input-states is not triggered - all browsers - https://github.com/SAP/fundamental-ngx/pull/2943
![image](https://user-images.githubusercontent.com/26483208/88923448-0af2a400-d272-11ea-9a83-67c5d52c4a1e.png)

 - [x] Multi input - all browsers - https://github.com/SAP/fundamental-ngx/pull/2942
![image](https://user-images.githubusercontent.com/26483208/88925784-4e9add00-d275-11ea-9c7a-1e0117d779f2.png)

- [x] Multi input keyboard support on IE11 - https://github.com/SAP/fundamental-ngx/pull/2942
![multiinput](https://user-images.githubusercontent.com/26483208/88927054-27dda600-d277-11ea-8106-20a5d48e87d4.gif)

- [x] Combobox limited width for container
![image](https://user-images.githubusercontent.com/26483208/89169158-274a5580-d57e-11ea-91f9-da9a108251f3.png)

salarenko:
- [ ] Combobox - Chrome
![image](https://user-images.githubusercontent.com/17496353/89163392-ce76bf00-d575-11ea-9a30-1c18758db0d2.png)

- [ ] Combobox does not update its view value when used with Reactive Forms - Chrome
![combobox-reactive-forms](https://user-images.githubusercontent.com/17496353/89165391-c79d7b80-d578-11ea-85a0-bc7eca2027cc.gif)

- [ ] Datepicker/Datetimepicker/Timepicker opening on input click - Chrome
[...]

PR #2973, opened 2020-08-03
fix: (core) Add Checkbox label class input | Simplify Multi Input

salarenko:
Part of: #2939

Checkbox used inside list needs to have additional class applied to display properly. Previously class was applied directly to the `nativeElement` in some specific moments of component life-cycle but it didn't work well with filter quiery based rendering. The easiest way is creating `Checkbox[labelClass]` to pass class directly to the label. I think this solution would be sooner or later requested by the users.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
+++ b/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
@@ -72,6 +72,10 @@ export class CheckboxComponent implements ControlValueAccessor {
     @Input()
     tristateSelectable = true;
 
+    /** Assigns given class to checkbox label element */
+    @Input()
+    labelClass: string;
+
     /** Sets values returned by control. */
     @Input('values')
     set _values(checkboxValues: FdCheckboxValues) {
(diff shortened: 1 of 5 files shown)


libs/core/src/lib/multi-input/multi-input.component.ts
```diff
@@ -190,10 +190,6 @@ export class MultiInputComponent implements
     @ViewChild(PopoverComponent)
```
JKMarkowski, line 190: I believe we can remove import as well there

libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
```diff
@@ -72,6 +72,10 @@ export class CheckboxComponent implements ControlValueAccessor {
     @Input()
     tristateSelectable = true;
 
+    /** Assigns given class to checkbox label element */
+    @Input()
+    labelClass: string;
```
mikerodonnell89, line 77: I think this should be hidden

salarenko: I think this is very useful input. Users often apply their classes to form label/values to customize their look. In this case label element is barely accessible.

PR #2975, opened 2020-08-03
fix: Date-picker disabled state

salarenko:
Part of: #2939

DatePicker component required adding separate `[disabled]` attributes on input ad button elements to truly support disabled state.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


libs/core/src/lib/date-picker/date-picker.component.html
```diff
@@ -9,27 +9,29 @@
...
         <fd-input-group [compact]="compact" [state]="state" [disabled]="disabled">
             <input
+                type="text"
                 class="fd-input"
                 fd-input-group-input
-                type="text"
+                [compact]="compact"
+                [disabled]="disabled"
+                [placeholder]="placeholder"
+                [class.is-error]="isInvalidDateInput && useValidation"
```
valorkin, line 18: this is a nice change, I like it :D

salarenko: Look guys! I'm not crazy! Other people also find this grouping useful 😆

valorkin: One thing though, this will be executed on each change detection cycle, because angular doesn't have concept of computed properties.
What I really like is removing of ngClass/ngStyle usage, which are slow and bring the whole module with it

Issue #2976, opened 2020-08-03
Core maintenance


salarenko:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
Core library needs to update its status on its deprecated elements:

- [x] Remove Dropdown
- [x] Remove Identifier
- [ ] Remove checkbox crash notice for v0.22.0
- [ ] Remove radio button crash notice for v0.22.0
- [x] Remove Input Group Number PR #2993
- [x] Remove Loading Spinner PR #2988
- [x] Remove Select Native
- [ ] Remove usages of `fd-segmented-button` directive and the directive itself
- [ ] Get rid of Shellbar related classes from InputGroup

katekozlowska:
work will be continued under https://github.com/SAP/fundamental-ngx/issues/3079

PR #2993, opened 2020-08-06
fix: remove input group number

droshev:
mentioned in #2976
BREAKING CHANGE: remove input group number


droshev:
> Could you by the way fix StackBlitz files config?
> One of the files is using `input-group-number-example` file name, and this file has been removed.

renamed the field

Issue #2776, opened 2020-06-30
Core :List component enhancement as per Fiori3 standards


sKudum:
#### Is this a bug, enhancement, or feature request?
enhancement
#### Briefly describe your proposal.
List and list item has to include Fiori3 features which are listed below.

Attachment List

Buttons

Deletion

Growing List-on button press

Header

Interactive States- Styles has to be updated as. with 0.9

 Navigation Indication

 No Data

Selection

Wrapping Behavior

These components and features are exposed as components in the platform.
We can have them as directive and be part of HTML ul and li in the core. Which provides felixiblity to the user.

**List item Variants:**

Standard List

Standard List with Byline

Action list. Item

Display list item

Object list item

Input list item

Grid list

Grid list item

Feed List item

https://fundamental-styles.netlify.app/?path=/docs/components-list-byline--standard
https://fundamental-styles.netlify.app/?path=/docs/components-list-standard--standard

PR #2996, opened 2020-08-06
feat: refactor DND directive

JKMarkowski:
related to: https://github.com/SAP/fundamental-ngx/issues/2776
Drag and drop directive for now is used only inside product switch. Before there was a requirement, to keep `cdkDrag` directive under `fd-dnd-container`. So the final structure had to be 
`ul fd-dnd-list` > `li fd-dnd-container` > `div cdkDrag`.
Now `fd-dnd-container` is changed to `fd-dnd-item` and has cdk drag as a service. It allows optimized usage of dnd.
`ul fd-dnd-list` > `li fd-dnd-item`
It changed final markup of product-switch, but it's not visible from end-user perspective

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
+++ b/libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
@@ -1,5 +1,5 @@
 import { AfterContentInit, Directive, ElementRef, EventEmitter, Input, OnDestroy, Output } from '@angular/core';
-import { CdkDragMove, DragDrop, DragRef } from '@angular/cdk/drag-drop';
+import { DragDrop, DragRef } from '@angular/cdk/drag-drop';
 import { ElementChord, LinkPosition } from '../dnd-list/dnd-list.directive';
 import { Subscription } from 'rxjs';
 
--- a/libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
+++ b/libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
@@ -63,7 +63,9 @@ export class DndItemDirective implements AfterContentInit, OnDestroy {
     /** @hidden */
     private _subscriptions = new Subscription();
 
-    /** @hidden */
+    /** @hidden
+     * Element that indicates position of dragged element. It's ghost copy of dragged element.
+     */
     private _placeholderElement: HTMLElement;
 
     /** @hidden */
--- a/libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
+++ b/libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
@@ -1,6 +1,7 @@
-import { AfterContentInit, Directive, ElementRef, EventEmitter, Input, Output } from '@angular/core';
+import { AfterContentInit, Directive, ElementRef, EventEmitter, Input, OnDestroy, Output } from '@angular/core';
 import { CdkDragMove, DragDrop, DragRef } from '@angular/cdk/drag-drop';
 import { ElementChord, LinkPosition } from '../dnd-list/dnd-list.directive';
+import { Subscription } from 'rxjs';
 
 @Directive({
     // tslint:disable-next-line:directive-selector
@@ -12,7 +13,7 @@ import { ElementChord, LinkPosition } from '../dnd-list/dnd-list.directive';
         DragDrop
     ]
 })
-export class DndItemDirective implements AfterContentInit {
+export class DndItemDirective implements AfterContentInit, OnDestroy {
     /** Event thrown when the element is moved by 1px */
     @Output()
     readonly moved = new EventEmitter<CdkDragMove>();
@@ -34,13 +35,15 @@ export class DndItemDirective implements AfterContentInit {
     dndDisabled = false;
 
     /** Class added to element, when it's dragged. */
-    readonly CLASS_WHEN_ELEMENT_DRAGGED: string = 'fd-dnd-on-drag';
+    @Input()
+    classWhenElementDragged = 'fd-dnd-on-drag';
 
     /** @hidden
      * Drag reference, object created from DND CDK Service
      */
     private _dragRef: DragRef;
 
+    private _subscriptions = new Subscription();
     private _placeholderElement: HTMLElement;
     private _lineElement: HTMLElement;
     private _replaceIndicator: HTMLElement;
@@ -48,14 +51,14 @@ export class DndItemDirective implements AfterContentInit {
     constructor(public element: ElementRef, private _dragDrop: DragDrop) {}
 
     /** @hidden */
-    public getElementChord(isBefore: boolean, listMode: boolean): ElementChord {
+    getElementCoordinates(isBefore: boolean, gridMode: boolean): ElementChord {
         /** Takes distance from the beginning of window page */
         const rect = <DOMRect>this.element.nativeElement.getBoundingClientRect();
 
         const position: LinkPosition = isBefore ? 'before' : 'after';
 
         /** Depending on the position, gets the left or right side of element */
-        const x = rect.left + (isBefore || listMode ? 0 : this.element.nativeElement.offsetWidth);
+        const x = rect.left + (isBefore || gridMode ? this.element.nativeElement.offsetWidth : 0);
 
         /** Vertically distance is counted by distance from top of the side + half of the element height */
         return {
@@ -67,21 +70,26 @@ export class DndItemDirective implements AfterContentInit {
     }
 
     /** @hidden */
-    public ngAfterContentInit(): void {
+    ngAfterContentInit(): void {
         if (!this.dndDisabled) {
             this._setCDKDrag();
         }
     }
 
     /** @hidden */
-    public onCdkMove(cdkMovedEvent: CdkDragMove): void {
+    ngOnDestroy(): void {
+        this._subscriptions.unsubscribe();
+    }
+
+    /** @hidden */
+    onCdkMove(cdkMovedEvent: CdkDragMove): void {
         this.moved.emit(cdkMovedEvent);
     }
 
     /** @hidden */
-    public onCdkDragReleased(): void {
+    onCdkDragReleased(): void {
         /** Remove class which is added, when element is dragged */
-        this.element.nativeElement.classList.remove(this.CLASS_WHEN_ELEMENT_DRAGGED);
+        this.element.nativeElement.classList.remove(this.classWhenElementDragged);
         this.released.emit();
 
         /** Resets the position of element. */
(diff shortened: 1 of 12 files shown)


libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
```diff
@@ -0,0 +1,216 @@
...
+    /** Event thrown when the element is released */
+    @Output()
+    readonly released = new EventEmitter<void>();
+
+    /** Event thrown when the element is started to be dragged */
+    @Output()
+    readonly started = new EventEmitter<void>();
+
+    /** Whether this element should stick in one place, without changing position */
+    @Input()
```
salarenko, line 29: Do we need both `stickInPlace` and `dndDisabled`? They seem to define the same behavior. Maybe we could replace it with one `draggable` input.

libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
```diff
@@ -0,0 +1,216 @@
...
+    stickInPlace = false;
+
+    /** Defines if element is disabled from drag and drop */
+    @Input()
+    dndDisabled = false;
+
+    /** Class added to element, when it's dragged. */
+    readonly CLASS_WHEN_ELEMENT_DRAGGED: string = 'fd-dnd-on-drag';
```
salarenko, line 37: Should be implicite typing

libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
```diff
@@ -0,0 +1,216 @@
...
+        if (this._replaceIndicator && this._replaceIndicator.parentNode) {
+            // IE11 workaround
+            this._replaceIndicator.parentNode.removeChild(this._replaceIndicator);
+            this._replaceIndicator = null;
+        }
+    }
+
+    /** @hidden */
+    public createReplaceIndicator(): void {
+        this._replaceIndicator = document.createElement('DIV');
```
salarenko, line 133: Why upper case `DIV`?

libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
```diff
@@ -0,0 +1,216 @@
...
+            x: Math.abs(element.getBoundingClientRect().left - parentLeft),
+            y: Math.abs(element.getBoundingClientRect().top - parentTop)
+        }
+
+    }
+
+    /** @hidden */
+    private _setCDKDrag(): void {
+        this._dragRef = this._dragDrop.createDrag(this.element);
+        this._dragRef.moved.subscribe((event: any) => {
```
salarenko, line 206: Never unsubscribed.

libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
salarenko, line 37: Maybe we should create an `@Inpit` for `CLASS_WHEN_ELEMENT_DRAGGED` class?

libs/core/src/lib/utils/drag-and-drop/dnd-list/dnd-list.directive.ts
```diff
@@ -29,40 +28,44 @@ export interface ElementChord {
 })
 export class DndListDirective implements AfterContentInit {
     /** @hidden */
-    @ContentChildren(DndContainerDirective)
-    dndContainerItems: QueryList<DndContainerDirective>;
+    @ContentChildren(DndItemDirective)
+    dndContainerItems: QueryList<DndItemDirective>;
```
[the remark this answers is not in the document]
JKMarkowski: Leftover, dndItems will be better

libs/core/src/lib/utils/drag-and-drop/dnd-list/dnd-list.directive.ts
```diff
@@ -29,40 +28,44 @@ export interface ElementChord {
...
 
     /** Defines if the distance between elements should be counted only by vertical distance */
     @Input()
     listMode = false;
 
+    /** When enabled, replace indicator will appear on whole element, instead of horizontal/vertical line */
+    @Input()
+    replaceMode = false;
```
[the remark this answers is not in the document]
JKMarkowski: It can be misleading, that's leftover after only product switch usage, which by default is grid.
I will change it to gridMode with proper comments

libs/core/src/lib/utils/drag-and-drop/dnd-list/dnd-list.directive.ts
```diff
@@ -29,40 +28,44 @@ export interface ElementChord {
...
+
     /** Array of items, that will be sorted */
     @Input()
     public items: Array<any>;
 
     /** Event that is thrown, when the item is dropped */
     @Output()
-    readonly itemsChange: EventEmitter<Array<any>> = new EventEmitter<any>();
+    readonly itemsChange = new EventEmitter<any>();
```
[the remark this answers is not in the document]
JKMarkowski: I would keep `itemsChange` and add other event with:
```   
 replacedItemIndex: number;
 draggedItemIndex: number;
```

libs/core/src/lib/utils/drag-and-drop/dnd-list/dnd-list.directive.ts
```diff
@@ -91,28 +94,38 @@ export class DndListDirective implements AfterContentInit {
...
+            } else {
+                this._generateLine(this._closestLinkIndex, this._closestLinkPosition);
+            }
         }
     }
 
     /** Method called, when element is started to be dragged */
     dragStart(ind: number): void {
-        this.draggedItemIndex = ind;
+        this._draggedItemIndex = ind;
```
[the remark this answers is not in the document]
JKMarkowski: Good Idea

libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
JKMarkowski: Yes, It's needed cause for example first item with `stickInPlace` can't have dragged any element before it's position. I can see case, where someone blocks element from being dragged, but makes placing elements before, or after this element possible.

libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
JKMarkowski: I don't know, it is treated the same as a lowercase

libs/core/src/lib/utils/drag-and-drop/dnd-list/dnd-list.directive.ts
```diff
@@ -91,28 +94,38 @@ export class DndListDirective implements AfterContentInit {
...
         const draggedItemElement = this.dndContainerItems.toArray()[ind].element;
         /** Counting all of the elements's chords */
-        this.elementChords = this.dndContainerItems
+        this._elementChords = this.dndContainerItems
```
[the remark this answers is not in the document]
JKMarkowski: What is advantage of it? I would consider it, on trying to add dnd as a service.

libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
salarenko: - Can we make it clear in the documentation? That part saying that `stickInPlace` prevents the item to be dragged and moved by the other elements.

- I think we should also make it possible to turn off and on the dragging. Right now its only set-up on `ngAfterContentInit` so there is no way to change that behavior after its set at the beginning.

- What do you think about renaming it into `draggable`? `dndDisabled` is something that I'd rather expect to see on the container, disabling and enabling draggable behavior in for all of the elements in the container (maybe we should also add this feature?).

libs/core/src/lib/utils/drag-and-drop/dnd-list/dnd-list.directive.ts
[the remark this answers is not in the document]
salarenko: Optimization, right now you are mapping `toArray` 4 times just t get something by index. You can reduce it to only one mapping.

libs/core/src/lib/utils/drag-and-drop/dnd-list/dnd-list.directive.ts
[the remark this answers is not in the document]
salarenko: `/** Defines if the distance between elements should be counted only by vertical distance */`
I think from the user perspective its more important that in `gridMode` you can drag and drop items in two dimensions. Current description is not very helpful 😛

libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    /** @hidden
+     * Drag reference, object created from DND CDK Service
+     */
+    private _dragRef: DragRef;
+
+    private _subscriptions = new Subscription();
+    private _placeholderElement: HTMLElement;
+    private _lineElement: HTMLElement;
+    private _replaceIndicator: HTMLElement;
```
InnaAtanasova, line 54: shouldn't these be hidden?

libs/core/src/lib/utils/drag-and-drop/dnd-list/dnd-list.directive.ts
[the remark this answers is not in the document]
JKMarkowski: I will add some reference then.

libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
JKMarkowski: The disabled state for whole list is added. Also naming has been changed. I updated comments

libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
JKMarkowski: Done

libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
```diff
@@ -0,0 +1,255 @@
...
+    private _dragRef: DragRef;
+
+    /** @hidden */
+    private _draggable = true;
+
+    /** @hidden */
+    private _subscriptions = new Subscription();
+
+    /** @hidden */
+    private _placeholderElement: HTMLElement;
```
stefanoScalzo, line 67: I think we should name anything with placeholder differently, it is confusing

stefanoScalzo:
Should specify in docs that home is not draggable and what to do to achieve this

libs/core/src/lib/utils/drag-and-drop/dnd-item/dnd-item.directive.ts
JKMarkowski: The naming is same as on angular CDK. I think placeholderElement is meaningful. I can add some additional comment there, that it's added as a ghost element, that indicates position of dragged element.

JKMarkowski:
> Should specify in docs that home is not draggable and what to do to achieve this

Everything is written in product switch docs. DND is not standalone diractive yet. I will add some external page for DND in feature.

Issue #3018, opened 2020-08-12
Core: Toolbar Enhancements


olshansky:
#### Is this a bug, enhancement, or feature request?
enhancement
#### Briefly describe your proposal.
[Prioritization](https://experience.sap.com/fiori-design-web/toolbar-overview/#prioritization) & [Grouping](https://experience.sap.com/fiori-design-web/toolbar-overview/#grouping)
##### Prioritization
Individual child elements can be "prioritized" for the overflow behavior using the `fdOverflowPriority` attribute directive. The values for prioritization are as follows:
- `always` - This element will always be positioned in the overflow menu.
- `never` - This element will never be added to the overflow. It will always be visible in the toolbar.
- `low` - The element has low priority and will be consider first for moving to overflow menu.
- `high` - This element has high priority and will be considered last for moving to overflow menu.
- `disappear` - This element is not important and will be removed from the toolbar, and will not be added to the overflow menu.

##### Grouping
Child elements can be grouped so that they can "overflow" together. Setting the attribute directive `fdOverflowGroup` to the same number will allow child elements to be grouped. Child elements which are grouped, will enter (or leave) the overflow menu together. Child elements grouped in this manner cannot have the `fdOverflowPriority` values of `always` or `never`.

#### Is there anything else we should know?
- [ ] Should be able to add Button, Menu Button, Segmented Button, Select, Toggle Button, Checkbox Button, Input, Search Field, Combobox, Date Time, Label, Link, and Toolbar Separator.
- [ ] Should always place a child element designated `fdOverflowPriority="always"` in the overflow menu.
- [ ] Should never place a child element designated `fdOverflowPriority="never"` in the overflow menu.
- [ ] Should place a child element designated as `fdOverflowPriority="low"` in the overflow menu before a child element designated as `fdOverflowPriority="high"` as the toolbar width is reduced.
- [ ] Should remove child element designated as `fdOverflowPriority="disappear"` from toolbar and not place it in the overflow menu, as the toolbar width is reduced.
- [ ] Should place child elements with the same `fdOverflowGroup` value in the overflow menu together.

Issue #2890, opened 2020-07-21
Toolbar Prioritization


droshev:
#### Is this a bug, enhancement, or feature request?
enhancement

https://experience.sap.com/fiori-design-web/toolbar-overview/#prioritization

PR #3019, opened 2020-08-12
fix: (core) toolbar enhancements

olshansky:
Closes: https://github.com/SAP/fundamental-ngx/issues/2890
Closes: https://github.com/SAP/fundamental-ngx/issues/3018

Documentation checklist:
- [x] Documentation Examples

### DIFF
--- a/apps/docs/src/app/core/component-docs/toolbar/examples/toolbar-example.component.ts
+++ b/apps/docs/src/app/core/component-docs/toolbar/examples/toolbar-example.component.ts
@@ -31,6 +31,18 @@ export class ToolbarSeparatorExampleComponent {}
 export class ToolbarOverflowExampleComponent {}
 
 @Component({
+    selector: 'fd-toolbar-overflow-priority-example',
+    templateUrl: './toolbar-overflow-priority-example.component.html'
+})
+export class ToolbarOverflowPriorityExampleComponent {}
+
+@Component({
+    selector: 'fd-toolbar-overflow-grouping-example',
+    templateUrl: './toolbar-overflow-grouping-example.component.html'
+})
+export class ToolbarOverflowGroupingExampleComponent {}
+
+@Component({
     selector: 'fd-toolbar-size-example',
     templateUrl: './toolbar-size-example.component.html'
 })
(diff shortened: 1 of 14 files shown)


JKMarkowski:
Hi @olshansky Code looks great for me. Just one thing that bothers me 
toolbarpriority
As you can see high/low priority are treated same. I guess this is also the reason why overflow button goes outside on lower resolutions

olshansky:
@JKMarkowski thanks, fixed

stefanoScalzo:
Screen Shot 2020-08-17 at 2 29 28 PM
i dont think its normal that they are all disappearing with this much space left

mikerodonnell89:
In the overflow priority example, "high" gets moved to overflow before "button first".  I think any item without a specified priority should be moved before high priority items

olshansky:
@stefanoScalzo I believe that it's correct according to the specifications:
> Child elements can be grouped so that they can "overflow" together.

olshansky:
@mikerodonnell89 note from fiori design guidelines: 
> The priority of each item is high by default. If two items have equal priority, the item on the right side overflows first.

`button first` has the same priority as the button `high`, but the button `high` on the right side of `button first`. So, it's correct behavior

olshansky:
As for a11y

> All child elements should be keyboard accessible
> - User should be able to "tab" navigate through all child elements.
> - User should be able to click "enter" to invoke action on the child elements.
>
> The overflow menu should be keyboard accessible.
> - User should be able to click "enter" to open the overflow menu.
> - User should be able to navigate the overflow menu using "up" and "down" arrow keys.
_____

> - User should be able to click "enter" to invoke action on the child elements.

I think it's related to the implementation of each child component, because of components could have different behavior on occurred action _(click, space or enter keypress)_

> - User should be able to navigate the overflow menu using "up" and "down" arrow keys.

It's a popover component and could be implemented in a separate PR.
_____
I could make a separate issue and describe these notes if necessary. Please, ping me if it necessary to do

KevinOkamoto:
A minor issue - is there a way to get rid of the button flicker as the toolbar is resized? Can we add a check to see if the set of visible buttons change in the `_collapseItems()` function, or something similar?

olshansky:
@KevinOkamoto with checking it's not so easy as I expected. But I got rid of flickering similar as much as possible now to UI5 examples

salarenko (approved):

Issue #3039, opened 2020-08-18
Platform components Switch and Textarea build fail in consuming applications


kavya-b:
#### Is this a bug, enhancement, or feature request?
bug
#### Briefly describe your proposal.
Running `ng build` on an application consuming ngx/platform lib gives the following error:
![image](https://user-images.githubusercontent.com/53509521/90505778-d5562200-e170-11ea-8ec8-5aef9a8cfce3.png)

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Ang 10, ngx 0.21.0
#### If this is a bug, please provide steps for reproducing it.
1. Download ngx platform lib packages between versions 0.20.1-rc.66 to the latest rc 0.21.1-rc.10
2. Run `ng build` in consuming application.
3. Observe error as in screenshot.

PR #3040, opened 2020-08-18
fix: (platform) add types to contentDensity to fix build error

kavya-b:
#3039 
missing return type was resulting in some random type added on running `ng build`
image

This can be observed even on running `ng build` in ngx repo, and looking for the affected files in the dist lib, but no error is thrown on terminal. 

Documentation checklist:
- [x] Documentation Examples


kavya-b:
> It's weird that this error message is caused by typing, it rather looks like invalid import. However it builds successfully.

Yeah, even I thought it was weird, though the build ran successfully in consuming application after this change, and this was the only difference I found between these files and the other successfully 'build-passing' components like Panel and Step Input that used config files for contentDensity. :)
Shall I go ahead and merge this then?

PR #3052, opened 2020-08-19
fix: append datepicker and datetime picker popovers to body

### LINKED ISSUE #3051: Datepicker popover cut off by dialog
Datepicker popovers in dialogs are currently cut off.  We should append the popover to the body by default

Screen Shot 2020-08-18 at 5 49 50 PM

mikerodonnell89:
fixes #3051

### DIFF
--- a/libs/core/src/lib/popover/popover.component.ts
+++ b/libs/core/src/lib/popover/popover.component.ts
@@ -61,7 +61,7 @@ export class PopoverComponent {
 
     /** The element to which the popover should be appended. */
     @Input()
-    appendTo: HTMLElement | 'body' = 'body';
+    appendTo: HTMLElement | 'body';
 
     /** The trigger events that will open/close the popover.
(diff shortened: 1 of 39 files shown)


libs/core/src/lib/popover/popover.component.ts
```diff
@@ -61,7 +61,7 @@ export class PopoverComponent {
 
     /** The element to which the popover should be appended. */
     @Input()
-    appendTo: HTMLElement | 'body';
+    appendTo: HTMLElement | 'body' = 'body';
```
JKMarkowski, line 64: wasn't it supposed to be only for date/datetime pickers?

mikerodonnell89: Yes, i thought i had rolled this part back

salarenko (approved):
Partially covering the problem, but ok for now. We finally need to migrate from popper v1 and then lib-globally cover this problem.

JKMarkowski (approved):

PR #3062, opened 2020-08-20
fix: add flags to combobox and multiinput to hide addon button

### LINKED ISSUE #3004: Not able to hide button icon from fd-combobox
Bug

Please add a flag to hide the button from combobox

v.13

Add fd-combobox by default button section is shown with the glyph. Please let the user have a control to hide the combobox button

mikerodonnell89:
fixes #3004 

adds the input [displayAddonButton] (default value true) to combobox and multiinput


apps/docs/src/app/core/component-docs/combobox/examples/combobox-example.component.html
```diff
@@ -61,3 +61,16 @@
 
 <span>Search Term: {{ searchTermFive }}</span>
 <br/><br/>
+
+Hide Addon Button
+<fd-combobox
+    [(ngModel)]="searchTermSix"
+    [dropdownValues]="dropdownValues"
```
salarenko, line 68: Please use static string bindings when possible.

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -190,6 +190,10 @@ export class ComboboxComponent
     @Input()
     mobileConfig: MobileModeConfig;
 
+    /** Whether or not to display the addon button. */
+    @Input()
+    displayAddonButton = true;
```
salarenko, line 195: I think in this context our input should be named `showDropdownButton`

salarenko (changes requested):

JKMarkowski (approved):
LGTM

Issue #3028, opened 2020-08-14
split button issues


droshev:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.
The spec for the split menu button says [link](https://experience.sap.com/fiori-design-web/button/#split-menu-button-behaviors):

**Split Menu Button Behaviors**
The split menu button can have two different behaviors:

- The button always triggers the default action set by the app developer. If no default action has been defined, the first item in the menu list becomes the default.
- The button triggers the last action chosen by the user. Initially, it also triggers the default action. However, when the user selects a different action, this user action becomes the default, and the button text label changes accordingly. The button has a fixed size and the text truncates if the menu item exceeds the available width (as with the combo box).

PR #3063, opened 2020-08-20
fix: add first option as title and remove mainActionTitle

stefanoScalzo:
BREAKING CHANGES: remove mainActionTitle 

fixes: #3028 
This PR will make the first item the title and then change to the selected item. mainActionTitle is removed as input and replace it behind the scenes. If the user wants to put a custom name, he can use the template example.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/apps/docs/src/app/core/component-docs/split-button/split-button-docs.component.html
+++ b/apps/docs/src/app/core/component-docs/split-button/split-button-docs.component.html
@@ -12,7 +12,7 @@
     <code class="code-snippet">[fdType]="'attention'"</code>
     <code class="code-snippet">[fdType]="'transparent'"</code>. Leave
     <code class="code-snippet">[fdType]=""</code> empty to use Action button<br />
-    To add a custom title use the <code class="code-snippet">mainActionTitle</code> input directive.
+    To add a custom title use the <code class="code-snippet">mainActionTitle</code> input.
 </description>
 <component-example>
     <fd-split-button-types-example></fd-split-button-types-example>
--- a/apps/docs/src/app/core/component-docs/split-button/split-button-docs.component.html
+++ b/apps/docs/src/app/core/component-docs/split-button/split-button-docs.component.html
@@ -11,7 +11,8 @@
     <code class="code-snippet">[fdType]="'negative'"</code>
     <code class="code-snippet">[fdType]="'attention'"</code>
     <code class="code-snippet">[fdType]="'transparent'"</code>. Leave
-    <code class="code-snippet">[fdType]=""</code> empty to use Action button
+    <code class="code-snippet">[fdType]=""</code> empty to use Action button<br />
+    To add a custom title use the <code class="code-snippet">mainActionTitle</code> input directive.
 </description>
 <component-example>
     <fd-split-button-types-example></fd-split-button-types-example>
@@ -38,7 +39,8 @@
     Open/Close Programmatically
 </fd-docs-section-title>
 <description>
-    It is possible to control the open state of the popover using <code>MenuComponent.open()</code> and <code>MenuComponent.close()</code>
+    It is possible to control the open state of the popover using <code>MenuComponent.open()</code> and
+    <code>MenuComponent.close()</code>
     methods.
 </description>
 <component-example>
--- a/libs/core/src/lib/menu/menu.component.ts
+++ b/libs/core/src/lib/menu/menu.component.ts
@@ -129,6 +129,10 @@ export class MenuComponent implements MenuInterface, AfterContentInit, AfterView
     @Output()
     readonly activePath: EventEmitter<MenuItemComponent[]> = new EventEmitter<MenuItemComponent[]>();
 
+    /** Emits array of active menu items */
+    @Output()
+    readonly selected: EventEmitter<string> = new EventEmitter<string>();
+
     /** @hidden Emits event when the menu is opened/closed */
     @Output()
     isOpenChange: EventEmitter<boolean> = new EventEmitter<boolean>();
@@ -142,11 +146,11 @@ export class MenuComponent implements MenuInterface, AfterContentInit, AfterView
     menuWithPopover: TemplateRef<any>;
 
     /** @hidden Reference  the container where component views are instantiated */
-    @ViewChild('viewContainer', {read: ViewContainerRef})
+    @ViewChild('viewContainer', { read: ViewContainerRef })
     viewContainer: ViewContainerRef;
 
     /** @hidden Reference to all menu Items */
-    @ContentChildren(MenuItemComponent, {descendants: true})
+    @ContentChildren(MenuItemComponent, { descendants: true })
     menuItems: QueryList<MenuItemComponent>;
 
     /** Whether use menu in mobile mode */
@@ -168,13 +172,13 @@ export class MenuComponent implements MenuInterface, AfterContentInit, AfterView
     private _mobileModeComponentRef: ComponentRef<MenuMobileComponent>;
 
     constructor(public elementRef: ElementRef,
-                @Optional() @Inject(DIALOG_CONFIG) public dialogConfig: DialogConfig,
-                private _rendered: Renderer2,
-                private _menuService: MenuService,
-                private _changeDetectorRef: ChangeDetectorRef,
-                private _componentFactoryResolver: ComponentFactoryResolver,
-                @Optional() private _rtlService: RtlService,
-                @Optional() private _dynamicComponentService: DynamicComponentService) {
+        @Optional() @Inject(DIALOG_CONFIG) public dialogConfig: DialogConfig,
+        private _rendered: Renderer2,
+        private _menuService: MenuService,
+        private _changeDetectorRef: ChangeDetectorRef,
+        private _componentFactoryResolver: ComponentFactoryResolver,
+        @Optional() private _rtlService: RtlService,
+        @Optional() private _dynamicComponentService: DynamicComponentService) {
     }
 
     /** @hidden */
@@ -187,6 +191,7 @@ export class MenuComponent implements MenuInterface, AfterContentInit, AfterView
     ngAfterViewInit(): void {
         this._listenOnMenuMode();
         this._menuService.setMenuMode(this.mobile);
+        this._menuService.sendSelected(this.menuItems.first);
     }
 
     /** @hidden */
@@ -283,8 +288,9 @@ export class MenuComponent implements MenuInterface, AfterContentInit, AfterView
                 MenuMobileComponent,
                 { container: this.elementRef.nativeElement },
                 {
-                    injector: Injector.create({providers: [{ provide: MENU_COMPONENT, useValue: this }]}),
-                    services: [this._menuService, this._rtlService] }
+                    injector: Injector.create({ providers: [{ provide: MENU_COMPONENT, useValue: this }] }),
+                    services: [this._menuService, this._rtlService]
+                }
             )
     }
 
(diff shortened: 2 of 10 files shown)


libs/core/src/lib/menu/menu.component.ts
```diff
@@ -129,6 +129,10 @@ export class MenuComponent implements MenuInterface, AfterContentInit, AfterView
     @Output()
     readonly activePath: EventEmitter<MenuItemComponent[]> = new EventEmitter<MenuItemComponent[]>();
 
+    /** Emits array of active menu items */
```
JKMarkowski, line 132: Shouldn't it be `Emits event, when item is selected`?

libs/core/src/lib/split-button/split-button.component.ts
```diff
@@ -55,8 +58,15 @@ export class SplitButtonComponent {
...
-    mainActionTitle: string;
+    private _mainActionTitle: string;
+
+    set mainActionTitle(title: string) {
+        this._mainActionTitle = title;
+    }
+
+    get mainActionTitle(): string {
+        return this._mainActionTitle;
+    }
```
JKMarkowski, line 69: I think we should keep mainActionTitle, just in case user wants to define his own. If it's not set, then it should be changed. 
Maybe you can put some flag eg `@Input() changeTitleOnSelection: boolean`, which will define if selection of item should change action/title

JKMarkowski (commented):
Hi @stefanoScalzo. Good job, take a look at my  comments there

libs/core/src/lib/menu/menu.component.ts
stefanoScalzo: Not in basic menu component but in the split button component, which listens for the menu item which is selected. Because this feature is for split button not menu

libs/core/src/lib/split-button/split-button.component.ts
stefanoScalzo: Well if he uses template example he can set the title. Look at the split button template example.

stefanoScalzo: Actually what i did is i made by default it add the option one, if the user adds a mainItemTitle it will use that instead

apps/docs/src/app/core/component-docs/split-button/split-button-docs.component.html
```diff
@@ -11,7 +11,8 @@
     <code class="code-snippet">[fdType]="'negative'"</code>
     <code class="code-snippet">[fdType]="'attention'"</code>
     <code class="code-snippet">[fdType]="'transparent'"</code>. Leave
-    <code class="code-snippet">[fdType]=""</code> empty to use Action button
+    <code class="code-snippet">[fdType]=""</code> empty to use Action button<br />
+    To add a custom title use the <code class="code-snippet">mainActionTitle</code> input directive.
```
mikerodonnell89, line 15: "Input directive" is confusing, it's just an input

libs/core/src/lib/menu/services/menu.service.ts
```diff
@@ -132,6 +133,18 @@ export class MenuService {
         this._removeActiveSibling(menuItem);
         this.activeNodePath.push(menuNode);
         menuNode.item.setSelected(true);
+        this.sendSelected(menuNode.item)
+    }
+
+    /** @hidden Sends current selected menu item*/
+    sendSelected(menuItem: MenuItemComponent): void {
```
mikerodonnell89, line 140: I think you mean "set" not "send"?

libs/core/src/lib/menu/menu.component.ts
```diff
@@ -187,6 +191,7 @@ export class MenuComponent implements MenuInterface, AfterContentInit, AfterView
     ngAfterViewInit(): void {
         this._listenOnMenuMode();
         this._menuService.setMenuMode(this.mobile);
+        this._menuService.sendSelected(this.menuItems.first);
```
mikerodonnell89, line 194: I think there will be cases when we may want to select a different menu item on first load

libs/core/src/lib/menu/services/menu.service.ts
mikerodonnell89: Also I think this could be used programmatically for many reasons so it should not be hidden from the docs (related to my comment about maybe not always selected the first item)

libs/core/src/lib/menu/menu.component.ts
stefanoScalzo: Default behavior is to have the first element as the title

libs/core/src/lib/menu/services/menu.service.ts
stefanoScalzo: I meant send as it will be sending/emitting it to whatever the parent component is

apps/docs/src/app/core/component-docs/split-button/split-button-docs.component.html
stefanoScalzo: changed

libs/core/src/lib/menu/menu.component.ts
salarenko: Do you realize that you are changing the behavior of all menus, not only the ones inside `SplitButton`?
There is no reason for any menu to emit its first item as `selected` just after the component has been created.

libs/core/src/lib/menu/menu.component.ts
```diff
@@ -129,6 +129,10 @@ export class MenuComponent implements MenuInterface, AfterContentInit, AfterView
     @Output()
     readonly activePath: EventEmitter<MenuItemComponent[]> = new EventEmitter<MenuItemComponent[]>();
 
+    /** Emits array of active menu items */
+    @Output()
+    readonly selected: EventEmitter<string> = new EventEmitter<string>();
```
salarenko, line 134: Should emit `MenuItem`

libs/core/src/lib/split-button/split-button.component.ts
```diff
@@ -79,9 +81,22 @@ export class SplitButtonComponent {
...
+    constructor(private _changeDetectorRef: ChangeDetectorRef) { }
+
     /** @hidden Emits event when main button is clicked */
     onMainButtonClick(event: MouseEvent): void {
         this.primaryButtonClicked.emit(event);
         event.stopPropagation();
     }
+
+    ngAfterContentChecked(): void {
+        if (!this.mainActionTitle) {
```
salarenko, line 93: - Should be moved to separate method
- Never unsubscribed

salarenko (changes requested):
This solution does not introduce desired functionality. The thing that has been done here is copying first menu option name into buttons content, but whats missing is the whole functionality of an action brought into the button.

libs/core/src/lib/split-button/split-button.component.html
```diff
@@ -1,23 +1,18 @@
 <div class="fd-button-split fd-has-margin-right-small" role="group" aria-label="button-split" [fdMenuTrigger]="menu">
-    <button fd-button
-            [fdType]="fdType"
-            [compact]="compact"
-            [disabled]="disabled"
-            (click)="onMainButtonClick($event)">
-
+    <button fd-button [fdType]="fdType" [compact]="compact" [disabled]="disabled" (click)="onMainButtonClick($event)">
```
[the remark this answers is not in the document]
stefanoScalzo: fixed

libs/core/src/lib/split-button/split-button.component.ts
```diff
@@ -55,8 +58,7 @@ export class SplitButtonComponent {
     disabled: boolean;
 
     /** The Title for main  action button */
-    @Input()
-    mainActionTitle: string;
+    @Input() mainActionTitle: string;
```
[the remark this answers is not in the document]
stefanoScalzo: fixed

libs/core/src/lib/menu/menu.component.ts
stefanoScalzo: fixed

libs/core/src/lib/split-button/split-button.component.ts
stefanoScalzo: fixed

mikerodonnell89:
First button in the example ("Action menu") doesn't work

libs/core/src/lib/menu/menu.component.ts
mikerodonnell89: This doesn't emit an array as the comment suggests

libs/core/src/lib/split-button/split-button.component.ts
```diff
@@ -79,9 +85,63 @@ export class SplitButtonComponent {
...
+    /** @hidden */
+    ngOnDestroy(): void {
+        this._onDestroy$.next();
+        this._onDestroy$.complete();
+    }
+
+    /** @hidden */
+    private mainButtonClicked(): void {
+        if (this.mainActionTitle) {
+            this.primaryButtonClicked.emit(event);
```
mikerodonnell89, line 124: where is "event" coming from here?

libs/core/src/lib/split-button/split-button.component.html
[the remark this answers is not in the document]
salarenko: It's missing indent
image

libs/core/src/lib/menu/menu.component.ts
salarenko: You should also avoid explicit typing when property is declared and initialized at the same time.

libs/core/src/lib/split-button/split-button.component.ts
```diff
@@ -79,9 +85,63 @@ export class SplitButtonComponent {
...
+    constructor(private _changeDetectorRef: ChangeDetectorRef) { }
+
     /** @hidden Emits event when main button is clicked */
     onMainButtonClick(event: MouseEvent): void {
-        this.primaryButtonClicked.emit(event);
+        this.mainButtonClicked();
         event.stopPropagation();
     }
+
+    ngAfterContentChecked(): void {
```
salarenko, line 110: First declare life cycle methods then other public methods.
I recommend following our unofficial but popular convention of naming methods responsible for setting up observers/listeners like: `selectedSubscribed` -> `listenOnSelectedAction`

salarenko (changes requested):

libs/core/src/lib/split-button/split-button.component.ts
stefanoScalzo: it was from onMainButtonClicked i did this for it to be more modular but i changed it back

libs/core/src/lib/menu/menu.component.ts
```diff
@@ -168,13 +172,13 @@ export class MenuComponent implements MenuInterface, AfterContentInit, AfterView
...
 
     constructor(public elementRef: ElementRef,
-                @Optional() @Inject(DIALOG_CONFIG) public dialogConfig: DialogConfig,
-                private _rendered: Renderer2,
-                private _menuService: MenuService,
-                private _changeDetectorRef: ChangeDetectorRef,
-                private _componentFactoryResolver: ComponentFactoryResolver,
-                @Optional() private _rtlService: RtlService,
-                @Optional() private _dynamicComponentService: DynamicComponentService) {
+        @Optional() @Inject(DIALOG_CONFIG) public dialogConfig: DialogConfig,
```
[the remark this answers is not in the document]
stefanoScalzo: Can't due to constructor

libs/core/src/lib/split-button/split-button.component.html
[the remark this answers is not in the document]
stefanoScalzo: The editor either puts them all on one line or no indent. Even with no formatter

libs/core/src/lib/menu/menu.component.ts
stefanoScalzo: fixed

mikerodonnell89:
Screen Shot 2020-09-09 at 12 40 15 PM

Crashing with errors right now

droshev:
@stefanoScalzo take a look on Mike's PR #3239

PR #3076, opened 2020-08-25
fix: remove hardcoded segmented button styling

droshev:
remove hardcoded segmented button styling


JKMarkowski (approved):
LGTM. Now it looks good, without additional styling, hopefully nothing will break again

PR #3110, opened 2020-08-27
fix: datetime picker not rendering correctly

### LINKED ISSUE #3109: Datetime picker body not displaying properly and click issue
Due to #3052 and some custom styling, datetime picker is not displaying properly. Datepicker and time picker are being stacked vertically. Also, clicking anything in the popover other than a date will close it (clicking a date is fine because of stopPropagation)

Screen Shot 2020-08-27 at 4 17 11 PM

mikerodonnell89:
fixes #3109

### DIFF
--- a/libs/core/src/lib/datetime-picker/datetime-picker.component.scss
+++ b/libs/core/src/lib/datetime-picker/datetime-picker.component.scss
@@ -1,30 +1,25 @@
-.fd-datetime-host {
-    display: inline-block;
-    width: 230px;
+.fd-datetime {
+    display: block;
 
-    .fd-datetime {
+    &__container {
+        display: flex;
+        align-items: center;
+        margin: 0 16px;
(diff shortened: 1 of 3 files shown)


libs/core/src/lib/datetime-picker/datetime-picker.component.scss
```diff
@@ -1,30 +1,25 @@
...
+        margin: 42px 28px;
+        align-self: stretch;
+    }
 
-        fd-popover {
-            display: block;
-        }
+    fd-popover {
+        display: block;
+    }
```
stefanoScalzo, line 20: Remove file due to formatting

mikerodonnell89: this is not a formatting change. removal of the fd-datetime-host class

PR #3136, opened 2020-08-28
fix: add submit/cancel buttons for datetime picker

### LINKED ISSUE #2950: bring buttons into date time picker
We need to bring the buttons to confirm or cancel the date/time
Screen Shot 2020-07-31 at 9 44 41 AM
This way we can update the date/time once the user click OK button

mikerodonnell89:
fixes #2950 

Add submit and cancel buttons for datetime picker.

This does not have breaking changes for the developer but results in a slightly different UX which developers may need to consider

### DIFF
--- a/libs/core/src/lib/datetime-picker/datetime-picker.component.html
+++ b/libs/core/src/lib/datetime-picker/datetime-picker.component.html
@@ -72,11 +72,11 @@
                     (focusArrowLeft)="focusArrowLeft()"
                 ></fd-time>
             </div>
-            <div fd-popover-body-footer>
-                <div fd-bar [barDesign]="'footer'" [cosy]="!compact">
+            <div fd-popover-body-footer *ngIf="showFooter">
+                <div fd-bar barDesign="footer" [cosy]="!compact">
                     <div fd-bar-right>
                         <fd-bar-element>
-                            <button fd-button [fdType]="'emphasized'" [compact]="compact" (click)="submit()">Submit</button>
+                            <button fd-button fdType="emphasized" [compact]="compact" (click)="submit()">Submit</button>
                         </fd-bar-element>
                         <fd-bar-element>
                             <button fd-button [compact]="compact" (click)="cancel()">Cancel</button>
--- a/libs/core/src/lib/datetime-picker/datetime-picker.component.html
+++ b/libs/core/src/lib/datetime-picker/datetime-picker.component.html
@@ -15,14 +15,13 @@
                     type="text"
                     class="fd-input"
                     fd-input-group-input
-                    [attr.aria-label]="datetimeInputLabel"
                     [(ngModel)]="inputFieldDate"
-                    (ngModelChange)="handleInputChange($event)"
+                    [attr.aria-label]="datetimeInputLabel"
                     [placeholder]="placeholder"
-                    (click)="openPopover()"
                     [compact]="compact"
                     [ngClass]="{ 'is-error': isInvalidDateInput && useValidation }"
                     [disabled]="disabled"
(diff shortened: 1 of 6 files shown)


libs/core/src/lib/datetime-picker/datetime-picker.component.html
```diff
@@ -73,6 +73,18 @@
                     (focusArrowLeft)="focusArrowLeft()"
                 ></fd-time>
             </div>
+            <div fd-popover-body-footer>
+                <div fd-bar [barDesign]="'footer'" [cosy]="true">
```
JKMarkowski, line 77: It would be better to depend `cosy` on `!compact` flag from datetime picker instead of hardcoding it

libs/core/src/lib/datetime-picker/datetime-picker.component.html
```diff
@@ -73,6 +73,18 @@
...
+                    <div fd-bar-right>
+                        <fd-bar-element>
+                            <button fd-button [fdType]="'emphasized'" [compact]="true" (click)="submit()">Submit</button>
```
JKMarkowski, line 80: Same here for buttons, let's change it to `[compact]="compact"`

libs/core/src/lib/datetime-picker/datetime-picker.component.html
```diff
@@ -46,8 +46,8 @@
                     [activeView]="activeView"
                     (activeViewChange)="handleCalendarActiveViewChange($event)"
                     [disableFunction]="disableFunction ? disableFunction : null"
-                    [selectedDate]="selectedDate"
                     (selectedDateChange)="handleDateChange($event)"
+                    [selectedDate]="tempDate"
                     (isValidDateChange)="isInvalidDateInputHandler($event)"
                     [escapeFocusFunction]="null"
                     [compact]="compact"
```
JKMarkowski, line 53: Could you reorder properties there by inputs/outputs?

JKMarkowski (approved):
I tested it and I works well for me. 
I added some minor comments regarding code, but in my opinion after applying them it can be merged.
I also don't like opening of popover, when input is clicked. Let's have a talk about it.

katekozlowska (approved):

katekozlowska:
Looks good to me, work fine.

InnaAtanasova (approved):
Good improvement

libs/core/src/lib/datetime-picker/datetime-picker.component.html
```diff
@@ -57,22 +53,37 @@
...
-                    (ngModelChange)="handleTimeChange($event)"
                     [displaySeconds]="displaySeconds"
                     [displayMinutes]="displayMinutes"
                     [displayHours]="displayHours"
+                    (ngModelChange)="handleTimeChange($event)"
                     (focusArrowLeft)="focusArrowLeft()"
                 ></fd-time>
             </div>
+            <div fd-popover-body-footer>
+                <div fd-bar [barDesign]="'footer'" [cosy]="!compact">
```
salarenko, line 76: Use static string binding if possible

salarenko: Should the Datetime Picker have the footer by default? Maybe we should make this optional?

salarenko (changes requested):

libs/core/src/lib/datetime-picker/datetime-picker.component.html
mikerodonnell89: No specs for this but I suppose making it optional couldn't hurt

salarenko (approved):

PR #3162, opened 2020-08-30
chore: deprecate Localization Editor component

droshev:
deprecate Localization Editor component


JKMarkowski (approved):

JohnHavok:
I just realized while going to the component documentation that it is deprecated. It's a component that is planned to be used in our application and is planned in the current milestone.
There will be no replacement component?
What would be the recommended good practice in your opinion in this case?

Issue #2896, opened 2020-07-22
Accessibility inconsistency: Product Switch Keyboard Focus


estelaxu3:
According to Fiori Keyboard Interaction Guideline: https://wiki.wdf.sap.corp/wiki/display/Acc/Fiori+Keyboard+Interaction?preview=/1842806926/1867396960/Fiori%20Keyboard%20Interaction%20(untagged).pdf
<img width="855" alt="Screen Shot 2020-07-22 at 1 47 38 PM" src="https://user-images.githubusercontent.com/54589269/88227094-ecf9c380-cc21-11ea-850e-cbe8102be2d3.png">

> [UP], [LEFT] should move focus backwards. [DOWN], [RIGHT] should move focus forwards

In the current interaction, only [TAB] works. 
<img width="664" alt="Screen Shot 2020-07-22 at 1 47 01 PM" src="https://user-images.githubusercontent.com/54589269/88227032-d6536c80-cc21-11ea-9182-899e41481202.png">

PR #3187, opened 2020-09-02
fix: programmatically click product menu item on space/enter and add keyboard arrow navigation

### LINKED ISSUE #2899: Accessibility inconsistency: Product Switch Keyboard Interaction
According to Fiori Keyboard Interaction Guideline: https://wiki.wdf.sap.corp/wiki/display/Acc/Fiori+Keyboard+Interaction?preview=/1842806926/1867396960/Fiori%20Keyboard%20Interaction%20(untagged).pdf
[Enter-Key] and [Space-Key] should select current focussed element, trigger action.

In current implementation, [Enter-Key] and [Space-Key] does nothing.

mikerodonnell89:
fixes #2899, #2896 

wrong issue number on the branch name

### DIFF
--- a/libs/core/src/lib/product-switch/product-switch-body/product-switch-body.component.ts
+++ b/libs/core/src/lib/product-switch/product-switch-body/product-switch-body.component.ts
@@ -50,7 +50,7 @@ export class ProductSwitchBodyComponent implements OnInit {
 
     /** @hidden */
     ngOnInit(): void {
-        this.checkSize();
+        this._checkSize();
     }
 
     /**
@@ -72,21 +72,22 @@ export class ProductSwitchBodyComponent implements OnInit {
     /** @hidden */
     @HostListener('window:resize', [])
     onResize(): void {
-        this.checkSize();
+        this._checkSize();
     }
 
     /** @hidden */
     keyDownHandle(event: KeyboardEvent): void {
-        const target = (<HTMLElement>event.target);
+        const target = <HTMLElement>event.target;
+        const i = Array.from(target.parentElement.children).indexOf(target);
         if (!KeyUtil.isKey(event, 'Tab')) {
             event.preventDefault();
         }
         if (KeyUtil.isKey(event, ['Enter', ' '])) {
             target.click();
         } else if (!this.isListMode()) {
-            this.handleNoListKeydown(event);
+            this._handleNoListKeydown(event, target, i);
         } else if (this.isListMode() && KeyUtil.isKey(event, ['ArrowDown', 'ArrowUp'])) {
-            this.handleListArrowUpDown(event);
+            this._handleListArrowUpDown(event, target);
         }
     }
 
@@ -101,7 +102,7 @@ export class ProductSwitchBodyComponent implements OnInit {
     }
 
     /** @hidden */
-    private checkSize(): void {
+    private _checkSize(): void {
         if (this.isSmallMode()) {
             this.listMode = window.innerWidth < 588;
         } else {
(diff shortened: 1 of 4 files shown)


libs/core/src/lib/product-switch/product-switch-body/product-switch-body.component.ts
```diff
@@ -92,4 +108,67 @@ export class ProductSwitchBodyComponent implements OnInit {
             this.listMode = window.innerWidth < 776;
         }
     }
+
+    /** @hidden */
+    private handleNoListKeydown(event: KeyboardEvent): void {
```
InnaAtanasova, line 113: should have an `_` since it's private

libs/core/src/lib/product-switch/product-switch-body/product-switch-body.component.ts
```diff
@@ -92,4 +108,67 @@ export class ProductSwitchBodyComponent implements OnInit {
...
+        const target = (<HTMLElement>event.target);
```
InnaAtanasova, line 114: do you need the ( )?

libs/core/src/lib/product-switch/product-switch-body/product-switch-body.component.ts
```diff
@@ -92,4 +108,67 @@ export class ProductSwitchBodyComponent implements OnInit {
...
+            if (target.nextElementSibling && target.nextElementSibling.nextElementSibling &&
+                target.nextElementSibling.nextElementSibling.nextElementSibling &&
+                target.nextElementSibling.nextElementSibling.nextElementSibling.nextElementSibling) {
+                (<HTMLElement>target.nextElementSibling.nextElementSibling.nextElementSibling.nextElementSibling).focus();
+            }
+        }
+        if (KeyUtil.isKey(event, 'ArrowUp')) {
+            if (target.previousElementSibling && target.previousElementSibling.previousElementSibling &&
+                target.previousElementSibling.previousElementSibling.previousElementSibling &&
+                target.previousElementSibling.previousElementSibling.previousElementSibling.previousElementSibling) {
```
InnaAtanasova, line 141: Can we simplify these if statements?

libs/core/src/lib/product-switch/product-switch-body/product-switch-body.component.ts
```diff
@@ -92,4 +108,67 @@ export class ProductSwitchBodyComponent implements OnInit {
...
+                    .previousElementSibling).focus();
+            }
+        }
+    }
+
+    /** @hidden */
+    private handleNoListLessThanSeven(event: KeyboardEvent): void {
+        const target = (<HTMLElement>event.target);
+        if (KeyUtil.isKey(event, 'ArrowDown')) {
+            if (target.nextElementSibling && target.nextElementSibling.nextElementSibling &&
```
salarenko, line 152: What have you done 😆 
Please go recursive on that, or maybe there is a way to get all of the nodes of the parent and then easily get the desired node from the collection.

salarenko (changes requested):

libs/core/src/lib/product-switch/product-switch-body/product-switch-body.component.ts
mikerodonnell89: Re: this vs. recursive, I think this is about as straightforward as it gets but ok

salarenko (approved):

JKMarkowski (approved):
Works well for me. Good Job!

PR #3193, opened 2020-09-03
fix: multi input should attach to body

### LINKED ISSUE #3000: Multi select menu goes below the modal footer
bug

The multi select menu goes below the modal footer (or header).
image

Fundamental-ngx: 0.14.0
Fundamental-style: 0.5.2

```html
[..]
<fd-modal-body>
  <div fd-form-item>
    <label class="fd-has-display-block">...</label>
    <select fd-form-control [compareWith]="compare" [(ngModel)]="selectedCurrency">...</select>
  </div>
  <div fd-form-item>
    <label class="fd-has-display-block">...</label>
    <fd-multi-input [dropdownValues]="statuses" [(ngModel)]="selected" [displayFn]="displayNameFn"></fd-multi-input>
  </div>
</fd-modal-body>
[...]
```

Also the checkbox label is offset upwards and aligns perfectly when the checkbox is clicked.
You can see a margin on the left side of the checkbox (I think I saw a merge request on that).
Edit: The scroll, once in the middle, goes up by itself, impossible to go down.

mikerodonnell89:
fixes #3000

### DIFF
--- a/libs/core/src/lib/multi-input/multi-input.component.html
+++ b/libs/core/src/lib/multi-input/multi-input.component.html
@@ -9,6 +9,7 @@
             [disabled]="disabled"
             [closeOnOutsideClick]="true"
             [fillControlMode]="fillControlMode"
+            [appendTo]="'body'"
             class="fd-multi-input-popover-custom"
         >
             <fd-popover-control>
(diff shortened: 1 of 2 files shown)


libs/core/src/lib/multi-input/multi-input.component.html
```diff
@@ -9,6 +9,7 @@
             [disabled]="disabled"
             [closeOnOutsideClick]="true"
             [fillControlMode]="fillControlMode"
+            [appendTo]="'body'"
```
salarenko, line 12: Switch to static string binding

salarenko (changes requested):

salarenko (approved):

JKMarkowski (approved):

Issue #2772, opened 2020-06-30
Components having menu (ex: popover, menu button) do not work under shadow dom.


rajeev-learner:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
There is an Angular application which has a Home component, in which shadow dom is enabled. Now, when I try to use popover component inside the Home component, the popup menu does not open up upon mouse click. It starts working as expected as soon as I turn off ViewEncapsulation.ShadowDom on Home component.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
"@fundamental-ngx/core": "^0.17.1",
"@angular/core": "~9.1.3",

#### If this is a bug, please provide steps for reproducing it.
1. enable encapsulation: ViewEncapsulation.ShadowDom for a component.
2. use <fd-combobox> in component's template
3. load the ui and observe the combo control.
Issue: popup menu does not open upon mouse click.

droshev:
we can reproduce it [here](https://stackblitz.com/edit/angular-zxxnky?file=src/app/popover-example.component.ts)

JKMarkowski:
Adding `[closeOnOutsideClick]="false"` to popover is workaround there. [this](https://github.com/SAP/fundamental-ngx/blob/master/libs/core/src/lib/popover/popover-directive/popover.directive.ts#L393) method doesn't work properly with shadow dom.

PR #3196, opened 2020-09-04
fix: Bring support for shadow dom on Popover

JKMarkowski:
fixes: https://github.com/SAP/fundamental-ngx/issues/2772

In shadow DOM, event target returns a root container, not certain element. So there is another way to read it.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/popover/popover-directive/popover.directive.ts
+++ b/libs/core/src/lib/popover/popover-directive/popover.directive.ts
@@ -402,17 +402,34 @@ export class PopoverDirective implements OnInit, OnDestroy, OnChanges {
         }
     }
 
+    /** @hidden */
     private _shouldClose(event: MouseEvent): boolean {
         return (
             this.containerRef &&
             this.isOpen &&
             this.closeOnOutsideClick &&
-            event.target !== this.triggerRef.nativeElement &&
-            !this.triggerRef.nativeElement.contains(event.target) &&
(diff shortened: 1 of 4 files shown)


libs/core/src/lib/popover/popover-directive/popover.directive.ts
```diff
@@ -402,17 +402,34 @@ export class PopoverDirective implements OnInit, OnDestroy, OnChanges {
...
+        const containerElement = this.containerRef.location.nativeElement;
+        return containerElement.contains(event.target) ||
+            containerElement.contains(event.composedPath()[0])
+        ;
+    }
+
+    /** @hidden */
+    private _triggerContainsTarget(event: Event): boolean {
+        const triggerElement = this.triggerRef.nativeElement;
+        return triggerElement.contains(event.target) ||
```
salarenko, line 427: I believe we can always use `containerElement.contains(event.composedPath()[0])` because first element of the path should be the target. No need to check twice.

salarenko (changes requested):

libs/core/src/lib/popover/popover-directive/popover.directive.ts
JKMarkowski: Good Catch, thanks

salarenko (approved):

Issue #3085, opened 2020-08-25
Defect Hunting Fundamental NGX 0.22


droshev:
Test the library and the doc website.
List the issues here:
- [x] **Core**-Combobox: on page load console displays below error. [#3191](https://github.com/SAP/fundamental-ngx/pull/3191)
 
```
     main.js:24086 ERROR Error: Multiple components match node with tagname fd-combobox
    at throwMultipleComponentError (main.js:25668)
```

- [x] **Core**-table  : on page load console displays below error.[#3191](https://github.com/SAP/fundamental-ngx/pull/3191)
`main.js:24086 ERROR Error: The pipe ‘filter’ could not be found!`

- [x] **Core**-file-uploader : on page load console displays logs. - https://github.com/SAP/fundamental-ngx/pull/3207

- [x] **Core**-Select-native  : on page load console displays below error. [#3191](https://github.com/SAP/fundamental-ngx/pull/3191)

```
main.js:24086 ERROR Error: Uncaught (in promise): Error: Cannot match any routes. URL Segment: ‘core/select-native’
Error: Cannot match any routes. URL Segment: ‘core/select-native’
```

- [x] **Core**-Input group : on page load console displays below error.[#3191](https://github.com/SAP/fundamental-ngx/pull/3191)
```
 Error: Multiple components match node with tagname fd-input-group
    at throwMultipleComponentError (main.js:25668)
    at findDirectiveDefMatches (main.js:29137)
    at resolveDirectives (main.js:28883)
    at elementStartFirstCreatePass (main.js:36219)
```

sKudum:
- [ ] Combobox(Core) : on page load console displays below error.
        ```
main.js:24086 ERROR Error: Multiple components match node with tagname fd-combobox
    at throwMultipleComponentError (main.js:25668)
```

- [ ] file-uploader(Core) : on page load console displays logs.

- [ ] Input group(Core) : on page load console displays below error.
```
 Error: Multiple components match node with tagname fd-input-group
    at throwMultipleComponentError (main.js:25668)
    at findDirectiveDefMatches (main.js:29137)
    at resolveDirectives (main.js:28883)
    at elementStartFirstCreatePass (main.js:36219) 
```

- [ ] Select-native (Core) : on page load console displays below error.
```
main.js:24086 ERROR Error: Uncaught (in promise): Error: Cannot match any routes. URL Segment: ‘core/select-native’
Error: Cannot match any routes. URL Segment: ‘core/select-native’
[...]

PR #3207, opened 2020-09-07
fix: fix files validation on file uploader, add unique ID

JKMarkowski:
part of https://github.com/SAP/fundamental-ngx/issues/3085
There is added unique ID for every single file uploader, if ID is not set, to avoid repeated inputs.
There is added same validation for file DND directive and FileUploader Component. The API for file uploader component is same, the only breaking change is in file dnd
Breaking Changes - `file-uploader-dragndrop.directive.ts` removed `invalidFileDrop` event, `fileChanged` event now throws all of dropped files wrapped in proper interface 
```
{
    validFiles?: File[];
    invalidFiles?: File[];
}
```


libs/core/src/lib/file-uploader/file-uploader.component.ts
```diff
@@ -179,15 +180,24 @@ export class FileUploaderComponent implements ControlValueAccessor {
...
 
-        this.validFiles = event.filter(file => checkSize(file.size));
-        this.invalidFiles = event.filter(file => !checkSize(file.size));
+        let allowedExtensions = null;
+        if (this.accept) {
+            allowedExtensions = this.accept.toLocaleLowerCase().replace(/[\s.]/g, '').split(',')
+        }
 
+        this.validFiles = event.filter(file => checkSize(file.size) && this._checkExtension(file, allowedExtensions));
+        this.invalidFiles = event.filter(file => !checkSize(file.size) || !this._checkExtension(file, allowedExtensions));
```
DeepakSap14, line 200: should we not change style, if invalid file is getting uploaded.
like we do in other form component, we set state = 'error' or 'warning'.

JKMarkowski: For sure state shouldn't be changed under the hood

JKMarkowski: I tried to add some example with states, but `readonly` state of input has it's own styling and it overwrites the semantic states(warning, success etc.). We should contact designer how it should look like in this case

libs/core/src/lib/file-uploader/file-uploader.component.ts
```diff
@@ -245,4 +255,11 @@ export class FileUploaderComponent implements ControlValueAccessor {
         this.inputRef.nativeElement.value = '';
         this.onChange([]);
     }
+
+    /** @hidden */
+    private _checkExtension(file: File, allowedExtensions?: string[]): boolean {
```
salarenko, line 260: In what situation is `allowedExtensions` optional?

salarenko (commented):
LGTM, one question:

mikerodonnell89 (changes requested):
For the invalid upload example, when I drag a png (should be valid), I get both a valid alert and an invalid alert.

Also for min/max file size, if i drag a file that is <1kb, i get the invalid alert twice

libs/core/src/lib/file-uploader/file-uploader.component.ts
stefanoScalzo: Im assuming the case where all extensions are allowed

libs/core/src/lib/file-uploader/file-uploader.component.ts
```diff
@@ -179,15 +180,24 @@ export class FileUploaderComponent implements ControlValueAccessor {
         }
         const maxSize = this.parseFileSize(this.maxFileSize);
         const minSize = this.parseFileSize(this.minFileSize);
+
         const checkSize = (fileSize: number): boolean => {
-            if (this.maxFileSize && fileSize > maxSize) { return false; }
-            if (this.minFileSize && fileSize < minSize) { return false; }
+            if (this.maxFileSize && fileSize > maxSize) {
```
stefanoScalzo, line 185: if(this.maxFileSize && fileSize > maxSize || this.minFileSize && fileSize < minSize) {
return false;
}

JKMarkowski: It has been moved to service, I sticked to 2 ifs, because it's easier to read

JKMarkowski:
@mikerodonnell89 This componet has been refactored, I moved everything to separated service, now file DND directive and fileUploader component got same validation. It works well atm.

apps/docs/src/app/core/component-docs/file-uploader/examples/file-uploader-max-example/file-uploader-max-example.component.ts
```diff
@@ -9,14 +9,12 @@ export class FileUploaderMaxExampleComponent {
     files: File[];
     invalidFiles: File[];
 
-    handleFileSection(): void {
-
-        alert('Files selected successfully!!!');
-
+    handleFileSection(files: File[]): void {
+        alert(files.length + ' Files selected successfully!!!');
```
InnaAtanasova, line 13: I don't think this should be hardcoded. What if there's i18n and you need to display this in another language?

apps/docs/src/app/core/component-docs/file-uploader/examples/file-uploader-min-and-max-size-example/file-uploader-min-and-max-size-example.component.ts
```diff
@@ -9,13 +9,13 @@ export class FileUploaderMinAndMaxSizeExampleComponent {
...
+    handleFileSection(files: File[]): void {
+        alert(files.length + ' Files selected successfully!!!');
     }
 
-    handleInvalidFiles($event): void {
-        alert(' Invalid file selection ');
-        this.invalidFiles = $event;
+    handleInvalidFiles(files: File[]): void {
+        alert(files.length + ' Invalid files selected ');
+        this.invalidFiles = files;
```
InnaAtanasova, line 18: same as my other comment about i18n. At least the application developer should be able to customize how they want to handle these 2 functions.

libs/core/src/lib/file-uploader/file-uploader.component.ts
JKMarkowski: It shouldn't be there. Anyway it has been moved to another service

salarenko (changes requested):
1. This example:
- Is inconsistent (button text and actual `accept` parameter)
- Does not accept `.png` files
image
Example `.png` file: [image-2.zip](https://github.com/SAP/fundamental-ngx/files/5245268/image-2.zip)

2. I've already asked about it in some other PR, but this is still very confusing for me, why do we have:
- `FileUploaderDragndropDirective.disabled`
- `FileUploaderDragndropDirective.dragndrop`
if both are responsible for disabling drag and drop functionality?

3. `FileUploaderDragndropDirective.disabled` still emits on `dragEntered` and `dragLeave` events.

4. Does it make sense to perform any file validation in `FileUploaderDragndropDirective`?
It's good to split the responsibility between specific components. This directive should rather be responsible for emitting drag/drop related events and any other component specific logic should be handled outside.

5. What is this 😄? How can the user set the files using Template/Reactive forms if this is not handled?
image

6. We have separate examples for:
- Max and Min file size
- Max file size
- Min file size
I think we can reduce them to Max Min example.

apps/docs/src/app/core/component-docs/file-uploader/examples/file-uploader-min-and-max-size-example/file-uploader-min-and-max-size-example.component.ts
InnaAtanasova: You are not missing anything :) I know these are in the examples, but don't you think the application developers would like to handle the success/failure cases in a similar way? Like provide an API for this?

salarenko:
@JKMarkowski 
Maybe lets move all of the issues/concerns with file-uploader file to new issue and tackle it as a separate task.
I think first point from my list "1. This example..." is in scope of this PR and should be fixed here.

JKMarkowski:
@salarenko 
@InnaAtanasova 
Some Minor changes in docs have been introduced, let's create another task with issues listed, I wouldn't put those as a scope of this PR

mikerodonnell89 (commented):
I don't think we need 3 examples for min/max file size.  But this looks good

salarenko (approved):

InnaAtanasova (approved):

mikerodonnell89 (approved):

droshev (approved):

PR #3213, opened 2020-09-07
fix: Bring valid selection into multi input component

JKMarkowski:
part of https://github.com/SAP/fundamental-ngx/issues/3085
Now multi input uses same structure as list with selection. Temporary there are added some styles, to keep fiori3 design. 
Also there is fixed focusing/selecting items by keyboard

### DIFF
--- a/libs/core/src/lib/list/list-item/list-item.component.ts
+++ b/libs/core/src/lib/list/list-item/list-item.component.ts
@@ -68,6 +68,10 @@ export class ListItemComponent implements KeyboardSupportItemInterface, AfterCon
     @HostBinding('class.fd-list__item--link')
     link = false;
 
+    /** Defines if list item is used inside MultiInputComponent */
+    @Input()
+    inMultiInput = false;
+
     /** @hidden */
     @ContentChild(CheckboxComponent)
     checkbox: CheckboxComponent;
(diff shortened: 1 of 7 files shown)


libs/core/src/lib/list/list-item/list-item.component.ts
```diff
@@ -68,6 +68,10 @@ export class ListItemComponent implements KeyboardSupportItemInterface, AfterCon
     @HostBinding('class.fd-list__item--link')
     link = false;
 
+    /** Defines if list item is used inside MultiInputComponent */
```
mikerodonnell89, line 71: At the very least this should be hidden from the app developer, but I think we do not need to use an input at all, the multi-input could get list item content children and set the property there.  Or this list item could ascend a few parents and figure out if it is in a multi input.

mikerodonnell89 (changes requested):
Looks good other than this one thing

libs/core/src/lib/list/list-item/list-item.component.ts
JKMarkowski: @mikerodonnell89 You are right, it should be controlled from parent. Anyway it has been changed, so now it uses selection list for multi input with some minor styles adjustments.

salarenko (approved):

mikerodonnell89 (changes requested):
I think after opening the popover, user should be able to tab in to the list.  Currently I have to click a list item to focus it before I can use the arrow keys to nav through the list

mikerodonnell89 (approved):

InnaAtanasova (approved):

PR #3227, opened 2020-09-08
fix: add flag for closing dialog on route change

### LINKED ISSUE #1877: dialog remains open when browser back button is clicked
feature

dialog remains open when browser back (and forward) buttons are clicked 
We can offer both options
closeOnNavigation: boolean

mikerodonnell89:
fixes #1877 

adds [closeOnNavigation] boolean input, when set to true, will automatically close the dialog when a router navigation event starts.  True by default so may be considered a breaking change rather than a bug fix if any developers intentionally utilized the previous behavior

### DIFF
--- a/libs/core/src/lib/dialog/dialog.component.ts
+++ b/libs/core/src/lib/dialog/dialog.component.ts
@@ -24,6 +24,7 @@ import { fromEvent, Subscription } from 'rxjs';
 import { applyCssClass } from '../utils/decorators/apply-css-class.decorator';
 import { CssClassBuilder } from '../utils/interfaces/css-class-builder.interface';
 import { debounceTime } from 'rxjs/operators';
+import { NavigationStart, Router } from '@angular/router';
 
 @Component({
     selector: 'fd-dialog',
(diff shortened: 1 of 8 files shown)


libs/core/src/lib/dialog/dialog.component.ts
```diff
@@ -110,8 +115,17 @@ export class DialogComponent implements OnInit, AfterViewInit, OnDestroy, CssCla
         @Optional() @Inject(DIALOG_CONFIG) public dialogConfig: DialogConfig,
         @Optional() @Inject(DIALOG_REF) private _dialogRef: DialogRef,
         private _elementRef: ElementRef,
-        private _changeDetectorRef: ChangeDetectorRef
-    ) {}
+        private _changeDetectorRef: ChangeDetectorRef,
+        private _router: Router
+    ) {
+        this._subscriptions.add(
```
salarenko, line 121: - Do not implement component logic in constructor, use life-cycle hooks
- Move to separate method

salarenko (changes requested):
This approach is not compatible with current Dialog implementation (there is no way to set `closeOnNavigation` for object based Dialog). I believe `closeOnNavigation` should be a part of `DialogConfig`.

salarenko (commented):
Looks good, but do we need an example for that behavior?

If yes, then please change the form of it...
image

salarenko (approved):

PR #3239, opened 2020-09-09
fix: split button title take 2

### LINKED ISSUE #3028: split button issues

mikerodonnell89:

fixes #3028 

similar to #3063, but this also includes the option of using an input to select the split button's selected menu item, as well as a flag to set a fixed width for the split button regardless of the length of the text in the input


libs/core/src/lib/menu/menu-item/menu-item.component.ts
```diff
@@ -66,7 +66,7 @@ export class MenuItemComponent implements DefaultMenuItem, OnChanges, AfterConte
     submenu: BaseSubmenu;
 
     @Output()
-    selected: EventEmitter<void> = new EventEmitter<void>();
+    onSelect: EventEmitter<void> = new EventEmitter<void>();
```
JKMarkowski, line 69: Should it be considered as a breaking change?

libs/core/src/lib/split-button/split-button.component.ts
```diff
@@ -79,9 +92,71 @@ export class SplitButtonComponent {
...
+    private _setupMenuItemSubscriptions(): void {
+        this.menu.menuItems.map((menuItem: MenuItemComponent) => {
+            this._menuItemSubscriptions.add(
+                menuItem.onSelect.subscribe(() => {
+                    this.mainActionTitle = menuItem.menuItemTitle.title;
+                    this._cdRef.detectChanges();
+                })
+            );
+        });
+    }
```
JKMarkowski, line 161: You would need to listen for changes
dynamic split button
Because dynamically addded options are not supported

libs/core/src/lib/menu/menu-item/menu-item.component.ts
```diff
@@ -66,7 +66,7 @@ export class MenuItemComponent implements DefaultMenuItem, OnChanges, AfterConte
     submenu: BaseSubmenu;
 
     @Output()
```
JKMarkowski, line 68: Let's maybe add some comment for Output

JKMarkowski (changes requested):
I like more this approach, there are few things about I noticed. Also width is fixed, so it can cause this
split button

libs/core/src/lib/menu/menu-item/menu-item.component.ts
mikerodonnell89: This was never emitted anywhere so I don't think so

mikerodonnell89:
> I like more this approach, there are few things about I noticed. Also width is fixed, so it can cause this
> split button

This is by design, can be turned off with the fixedWidth input

libs/core/src/lib/split-button/split-button.component.ts
```diff
@@ -67,6 +77,24 @@ export class SplitButtonComponent {
...
+
+    /** Whether or not the main action title should remain set as the button's action after another option is selected. */
+    @Input()
+    keepMainAction = false;
+
+    /**
+     * The object that contains the mainActionTitle and the callback function that should be executed when the button is clicked.
+     */
+    @Input()
+    mainAction: MainAction;
```
stefanoScalzo, line 96: it should also contain a keepMainAction property so that its solved all in one

libs/core/src/lib/split-button/split-button.component.ts
```diff
@@ -67,6 +77,24 @@ export class SplitButtonComponent {
...
+    fixedWidth = true;
+
+    /** Whether or not the main action title should remain set as the button's action after another option is selected. */
+    @Input()
+    keepMainAction = false;
+
+    /**
+     * The object that contains the mainActionTitle and the callback function that should be executed when the button is clicked.
+     */
+    @Input()
```
stefanoScalzo, line 95: Since itll take a callback we can also remove the primaryButtonClicked output

stefanoScalzo (approved):

stefanoScalzo (commented):
see comments above

libs/core/src/lib/split-button/split-button.component.ts
mikerodonnell89: I initially thought the same, but figured it is possible there could be cases when the dev needs to know if the primary button is clicked even though it's been changed to a different option.  Doesn't hurt to leave it, I think

Betrozov:
please look at Travis

apps/docs/src/app/core/component-docs/split-button/examples/split-button-types-example.component.html
```diff
@@ -13,7 +13,7 @@
     </fd-menu>
 </fd-split-button>
 
-<fd-split-button [fdType]="'standard'" [mainActionTitle]="'Standard Button'">
+<fd-split-button [fdType]="'standard'">
```
Betrozov, line 16: unnecessary bindings

apps/docs/src/app/core/component-docs/split-button/examples/split-button-types-example.component.html
```diff
@@ -28,7 +28,7 @@
     </fd-menu>
 </fd-split-button>
 
-<fd-split-button [fdType]="'emphasized'" [mainActionTitle]="'Emphasized Button'">
+<fd-split-button [fdType]="'emphasized'">
```
Betrozov, line 31: unnecessary bindings

apps/docs/src/app/core/component-docs/split-button/examples/split-button-types-example.component.html
```diff
@@ -43,7 +43,7 @@
     </fd-menu>
 </fd-split-button>
 
-<fd-split-button [fdType]="'ghost'" [mainActionTitle]="'Ghost Button'">
+<fd-split-button [fdType]="'ghost'">
```
Betrozov, line 46: unnecessary bindings

apps/docs/src/app/core/component-docs/split-button/examples/split-button-types-example.component.html
```diff
@@ -58,7 +58,7 @@
     </fd-menu>
 </fd-split-button>
 
-<fd-split-button [fdType]="'positive'" [mainActionTitle]="'Positive Button'">
+<fd-split-button [fdType]="'positive'">
```
Betrozov, line 61: unnecessary bindings

apps/docs/src/app/core/component-docs/split-button/examples/split-button-types-example.component.html
```diff
@@ -73,7 +73,7 @@
     </fd-menu>
 </fd-split-button>
 
-<fd-split-button [fdType]="'negative'" [mainActionTitle]="'Negative Button'">
+<fd-split-button [fdType]="'negative'">
```
Betrozov, line 76: unnecessary bindings

apps/docs/src/app/core/component-docs/split-button/examples/split-button-types-example.component.html
```diff
@@ -88,7 +88,7 @@
     </fd-menu>
 </fd-split-button>
 
-<fd-split-button [fdType]="'attention'" [mainActionTitle]="'Attention Button'">
+<fd-split-button [fdType]="'attention'">
```
Betrozov, line 91: unnecessary bindings

apps/docs/src/app/core/component-docs/split-button/examples/split-button-types-example.component.html
```diff
@@ -103,7 +103,7 @@
     </fd-menu>
 </fd-split-button>
 
-<fd-split-button [fdType]="'transparent'" [mainActionTitle]="'Transparent Button'">
+<fd-split-button [fdType]="'transparent'">
```
Betrozov, line 106: unnecessary bindings

Betrozov (changes requested):

apps/docs/src/app/core/component-docs/split-button/examples/split-button-options-example.component.html
```diff
@@ -30,7 +30,7 @@
 
 <fd-split-button
     [fdType]="'emphasized'"
```
Betrozov, line 32: unnecessary bindings

apps/docs/src/app/core/component-docs/split-button/examples/split-button-options-example.component.html
```diff
@@ -48,7 +48,7 @@
 
 <fd-split-button
     [fdType]="'transparent'"
```
Betrozov, line 50: unnecessary bindings

Betrozov (changes requested):

libs/core/src/lib/split-button/split-button.component.ts
```diff
@@ -79,9 +104,113 @@ export class SplitButtonComponent {
...
+    }
+
+    /** @hidden */
+    private _setupMenuItemSubscriptions(): void {
+        this.menu.menuItems.forEach(menuItem => {
+            this._menuItemSubscriptions.add(
+                menuItem.onSelect.subscribe(() => {
+                    if (!this.mainAction || !this.mainAction.keepMainAction) {
+                        this.selected = menuItem;
+                        if (this._init) {
```
JKMarkowski, line 186: Consider `first()` operator, to react on first event

JKMarkowski (approved):
Code works well for me! I added 1 comment regarding rxjs, if you find it's not proper for this use case, you can keep `_init` variable.

Betrozov (approved):

Issue #3420, opened 2020-09-28
Defect Hunting Fundamental NGX 0.22 [2]


salarenko:
Defect Hunting before v0.22.0 release.

JKMarkowski:
- [ ] Notifications - Need to bring old styling

- [ ] Header Github icon not centered
![image](https://user-images.githubusercontent.com/26483208/94420607-ade97080-0184-11eb-838b-b358d654f4f6.png)

- [ ] Action Bar - Need to bring changes from [this pr](https://github.com/SAP/fundamental-styles/pull/1338/files)
![image](https://user-images.githubusercontent.com/26483208/94420684-c8bbe500-0184-11eb-9297-caa0a33bab04.png)

- [ ] Overflow breadcrump - Color + cursor is not pointer
![image](https://user-images.githubusercontent.com/26483208/94420830-00c32800-0185-11eb-838b-42c769777ecf.png)

katekozlowska:
## CHROME 


 - [ ] Tokens are not centered
![deploy-preview-3400--fundamental-ngx netlify app_](https://user-images.githubusercontent.com/7014311/94432441-117c9980-0197-11eb-9115-0c8a9fff9f7d.png)

salarenko:
- [ ] Floating Bar Footer example is broken - IE11
![image](https://user-images.githubusercontent.com/17496353/94422919-19810d00-0188-11eb-9448-5bbb10067504.png)

- [ ] Combobox Custom Item Template should be updated to use list item markup to provide proper styling.

- [ ] `composedPath` is not supported in IE11 (Popover)

- [ ] `attachShadow` is not supported in IE11, breaking whole documentation page (Popover)

- [ ] File Uploader placeholder is cleared out when file input is focused - IE11
![file-uploader](https://user-images.githubusercontent.com/17496353/94429734-12abc780-0193-11eb-91c4-c47b08074b50.gif)

- [ ] Notification has not been migrated to Fundamental Styles v0.12.0
![image](https://user-images.githubusercontent.com/17496353/94430727-aaf67c00-0194-11eb-83b8-c1e8e4cbb38f.png)

- [ ] Basic Shellbar button - IE11
![image](https://user-images.githubusercontent.com/17496353/94431822-19880980-0196-11eb-871a-d0ae94af892a.png)

- [ ] Shellbar Menu button icon - IE11
[...]

PR #3434, opened 2020-09-28
fix: refactor Object Status to use an input property instead of content projection

InnaAtanasova:
Part of [#3420](https://github.com/SAP/fundamental-ngx/issues/3420)

Replace the content projection with an input property in order to apply the `<span *ngIf="label" class="fd-object-status__text">{{ label }}</span>` conditionally.

Before:
```
<span fd-object-status status="negative" inverted="true">Negative</span>
```
After:
```
<span fd-object-status status="negative" inverted="true" label="Negative"></span>
```

BREAKING CHANGE:
Replace the content projection of the Object Status with an input property called `label`.

### DIFF
--- a/apps/docs/src/app/core/component-docs/object-status/examples/object-status-clickable-and-icon-example.component.html
+++ b/apps/docs/src/app/core/component-docs/object-status/examples/object-status-clickable-and-icon-example.component.html
@@ -1,10 +1,10 @@
-<a fd-object-status [status]="'negative'" [glyph]="'status-negative'" [clickable]="true">Negative</a>
-<a fd-object-status [status]="'critical'" [glyph]="'status-critical'" [clickable]="true">Critical</a>
-<a fd-object-status [status]="'positive'" [glyph]="'status-positive'" [clickable]="true">Positive</a>
-<a fd-object-status [status]="'informative'" [glyph]="'hint'" [clickable]="true">Informative</a>
-<a fd-object-status [glyph]="'to-be-reviewed'" [clickable]="true">Default</a>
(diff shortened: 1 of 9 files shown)


JKMarkowski (approved):
LGTM. I'm just wondering if the `<ng-content>` should be deprecated

apps/docs/src/app/core/component-docs/object-status/examples/object-status-clickable-and-icon-example.component.html
```diff
@@ -1,10 +1,10 @@
-<a fd-object-status [status]="'negative'" [glyph]="'status-negative'" [clickable]="true">Negative</a>
-<a fd-object-status [status]="'critical'" [glyph]="'status-critical'" [clickable]="true">Critical</a>
-<a fd-object-status [status]="'positive'" [glyph]="'status-positive'" [clickable]="true">Positive</a>
-<a fd-object-status [status]="'informative'" [glyph]="'hint'" [clickable]="true">Informative</a>
-<a fd-object-status [glyph]="'to-be-reviewed'" [clickable]="true">Default</a>
+<a fd-object-status status="negative" glyph="status-negative" clickable="true" label="Negative"></a>
+<a fd-object-status status="critical" glyph="status-critical" clickable="true" label="Critical"></a>
+<a fd-object-status status="positive" glyph="status-positive" clickable="true" label="Positive"></a>
+<a fd-object-status status="informative" glyph="hint" clickable="true" label="Informative"></a>
+<a fd-object-status glyph="to-be-reviewed" clickable="true" label="Default"></a>
```
JKMarkowski, line 5: Boolean values can't have static binding (false would be treated as true won't be applied in this case)

PR #3468, opened 2020-09-29
fix: add fix for ShellBar user menu button using Avatar

InnaAtanasova:
Part of https://github.com/SAP/fundamental-ngx/issues/3420

Add glyph to Avatar inside user menu button.

### DIFF
--- a/libs/core/src/lib/avatar/avatar.component.html
+++ b/libs/core/src/lib/avatar/avatar.component.html
@@ -1,12 +1,11 @@
 {{ abbreviate }}
 
-<i *ngIf="zoomGlyph"
-   role="presentation"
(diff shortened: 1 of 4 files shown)


JKMarkowski (approved):

salarenko (changes requested):
This is not the fix! There is no reason to import Avatar styles into Shellbar, Avatar provides the styles by itself.

Here is the problem:
image

1. Avatar has icon class on its main element
2. Avatar is placed inside the button
3. Button selector is targeting the avatar as an icon and overriding its styles with icon styles

salarenko (approved):

salarenko (approved):

PR #3506, opened 2020-09-30
fix: Firefox double triggered checkbox

JKMarkowski:
part of https://github.com/SAP/fundamental-ngx/issues/3420
The space event is not stopped on firefox browser, it causes double trigger on checkbox inside list. 


libs/core/src/lib/utils/functions/browser-detection.ts
```diff
@@ -0,0 +1,15 @@
...
+        const trident = ua.indexOf('Trident/'); // IE 11
+
+        return msie > 0 || trident > 0;
+    }
+
+    /** @hidden */
+    static isFirefox(): boolean {
+        return navigator.userAgent.indexOf('Firefox') !== -1
+    }
+}
```
dimamarksman, line 15: Since we started using `angular/cdk` for some internal staff maybe it would be better to use [angular/cdk/platform](https://material.angular.io/cdk/platform/api) for such purposes?

JKMarkowski: @dimamarksman I removed this file and added `cdk/platform` browser detection. Thanks for pointing it

InnaAtanasova (approved):

Issue #3370, opened 2020-09-23
Adopt 0.12.0 of Fundamental-styles


InnaAtanasova:
Bring version 0.12.0 of Fundamental-styles in the project, address the Breaking Changes and defect hunt for new bugs that might be introduced with the migration.

Components to migrate:
 - [x] Avatar
 - [x] Generic Tile
 - [x] Info Label
 - [x] Buttons [WIP]
 - [x] List
 - [x] Menu
 - [x] Message Page - Not in NGX
 - [x] Nested List - J
 - [x] Notifications - Bring 0.11.0 styles
 - [x] Numeric Content
 - [x] Object Status
 - [x] Product Switch
 - [x] Switch
 - [x] Table
 - [x] Tabs

PR #3514, opened 2020-09-30
fix: (platform) updating button with styles 0.12.0

sKudum:
 https://github.com/SAP/fundamental-ngx/issues/3370

The `[label]` property should be used now instead of adding text as a content. The content still works fine, unless it's used with icons.

BREAKING CHANGE:

The value for the text button is now passed as an input property, not as content projection.

Before: `<fdp-button>Text</fdp-button>`
After: `<fdp-button label="Text"></fdp-button>`

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/platform/src/lib/components/button/button.component.html
+++ b/libs/platform/src/lib/components/button/button.component.html
@@ -1,6 +1,7 @@
 <button #fdButton fd-button [attr.id]="id" [attr.title]="title" [attr.name]="name" [attr.aria-label]="title"
-    [type]="type" [attr.value]="value" [attr.aria-selected]="ariaSelected" [attr.aria-disabled]="ariaDisabled"
-    [compact]="contentDensity === 'compact'" [glyph]="glyph" [disabled]="disabled" (click)="onBtnClick($event)"
-    [fdType]="buttonType" [ngStyle]="{ width: width }">
+    [glyphPosition]="glyphPosition" [label]="label" [type]="type" [attr.value]="value"
+    [attr.aria-selected]="ariaSelected" [attr.aria-disabled]="ariaDisabled" [compact]="contentDensity === 'compact'"
+    [glyph]="glyph" [disabled]="disabled" (click)="onBtnClick($event)" [fdType]="buttonType"
+    [ngStyle]="{ width: width }">
     <ng-content></ng-content>
 </button>
--- a/libs/platform/src/lib/components/button/button.component.html
+++ b/libs/platform/src/lib/components/button/button.component.html
@@ -3,5 +3,4 @@
     [attr.aria-selected]="ariaSelected" [attr.aria-disabled]="ariaDisabled" [compact]="contentDensity === 'compact'"
     [glyph]="glyph" [disabled]="disabled" (click)="onBtnClick($event)" [fdType]="buttonType"
     [ngStyle]="{ width: width }">
-    <ng-content></ng-content>
 </button>
--- a/libs/platform/src/lib/components/button/button.component.ts
+++ b/libs/platform/src/lib/components/button/button.component.ts
@@ -11,12 +11,24 @@ export type ButtonType =
     | 'transparent'
     | 'emphasized'
     | 'menu';
+
+export type GlyphPosition = 'before' | 'after';
 @Component({
     selector: 'fdp-button',
     templateUrl: './button.component.html',
     styleUrls: ['./button.component.scss']
 })
 export class ButtonComponent extends BaseComponent implements AfterViewInit {
+
+    /** Position of glyph related to text */
+    @Input()
+    public glyphPosition: GlyphPosition = 'before';
+
+    /**
+    * Text rendered inside button component
+    */
+    @Input()
+    label: string;
 
     /** The icon to include in the button. See the icon page for the list of icons.
      * Setter is used to control when css class have to be rebuilded.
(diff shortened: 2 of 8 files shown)


InnaAtanasova:
I don't know if this is in the scope of this PR (Core team usually targets this in 1 PR) but here you only address the Button changes in the Button documentation. For example, the buttons in Search field are broken:
Screen Shot 2020-09-30 at 11 10 11 AM
Screen Shot 2020-09-30 at 11 10 03 AM


In other components documentation where Buttons are used, the markup is not updated. Here for example, the value of the button is passed through content projection, not as an input prop:
Screen Shot 2020-09-30 at 11 11 52 AM

sKudum:
> I don't know if this is in the scope of this PR (Core team usually targets this in 1 PR) but here you only address the Button changes in the Button documentation. For example, the buttons in Search field are broken:
> Screen Shot 2020-09-30 at 11 10 11 AM
> Screen Shot 2020-09-30 at 11 10 03 AM
> 
> In other components documentation where Buttons are used, the markup is not updated. Here for example, the value of the button is passed through content projection, not as an input prop:
> Screen Shot 2020-09-30 at 11 11 52 AM

Hi @InnaAtanasova,  Respective component owners will be taking care of any breaking changes.
In this PR only button, we are addressing. @manjunathanagaraj has informed the team already about it.

InnaAtanasova:
> > I don't know if this is in the scope of this PR (Core team usually targets this in 1 PR) but here you only address the Button changes in the Button documentation. For example, the buttons in Search field are broken:
> > Screen Shot 2020-09-30 at 11 10 11 AM
> > Screen Shot 2020-09-30 at 11 10 03 AM
> > In other components documentation where Buttons are used, the markup is not updated. Here for example, the value of the button is passed through content projection, not as an input prop:
> > Screen Shot 2020-09-30 at 11 11 52 AM
> 
> Hi @InnaAtanasova, Respective component owners will be taking care of any breaking changes.
> In this PR only button, we are addressing. @manjunathanagaraj has informed the team already about it.

Sounds good. I removed the second part of my comment as I saw both ways are possible, but I still think the docs should be unified. It's up to your team to decide.

InnaAtanasova:
The rest looks good. If your team approve the PR I will add my approval right after. Thanks

libs/platform/src/lib/components/button/button.component.html
```diff
@@ -1,6 +1,7 @@
 <button #fdButton fd-button [attr.id]="id" [attr.title]="title" [attr.name]="name" [attr.aria-label]="title"
-    [type]="type" [attr.value]="value" [attr.aria-selected]="ariaSelected" [attr.aria-disabled]="ariaDisabled"
-    [compact]="contentDensity === 'compact'" [glyph]="glyph" [disabled]="disabled" (click)="onBtnClick($event)"
-    [fdType]="buttonType" [ngStyle]="{ width: width }">
+    [glyphPosition]="glyphPosition" [label]="label" [type]="type" [attr.value]="value"
+    [attr.aria-selected]="ariaSelected" [attr.aria-disabled]="ariaDisabled" [compact]="contentDensity === 'compact'"
+    [glyph]="glyph" [disabled]="disabled" (click)="onBtnClick($event)" [fdType]="buttonType"
+    [ngStyle]="{ width: width }">
     <ng-content></ng-content>
```
KevinOkamoto, line 6: Do we still need `ng-content`?

libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -11,13 +11,25 @@ export type ButtonType =
     | 'transparent'
     | 'emphasized'
     | 'menu';
+
+export type GlyphPosition = 'before' | 'after';
```
KevinOkamoto, line 15: We should not create our own type here. Just use the one created in core.

https://github.com/SAP/fundamental-ngx/blob/3fff22203122155c4308c093f7956734a23bfa52/libs/core/src/lib/button/button.component.ts#L13

KevinOkamoto (changes requested):
See comments.

libs/platform/src/lib/components/button/button.component.html
sKudum: removed

libs/platform/src/lib/components/button/button.component.ts
sKudum: removed

libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -11,13 +11,24 @@ export type ButtonType =
...
 @Component({
     selector: 'fdp-button',
     templateUrl: './button.component.html',
     styleUrls: ['./button.component.scss']
 })
 export class ButtonComponent extends BaseComponent implements AfterViewInit {
 
+    /** Position of glyph related to text */
+    @Input()
+    public glyphPosition = 'before';
```
KevinOkamoto, line 24: Add `GlyphPosition` type here but import it from core.
```
public glyphPosition: GlyphPosition = 'before';
```

KevinOkamoto (changes requested):

libs/platform/src/lib/components/button/button.component.ts
sKudum: updated

Issue #3279, opened 2020-09-14
Carousel component implementation in ngx-core


manjunathanagaraj:

manjunathanagaraj:
Awaiting tech spec presentation from fundamental team.

manjunathanagaraj:
Tech spec presented, approval pending. Work in progress. Moving to sprint 47.

DeepakSap14:
PR #3551 is merged. This can be closed.

manjunathanagaraj:
PR 3551 review in progress. Will be moved to sprint 48 to take it to closure.

PR #3551, opened 2020-10-05
feat: (Core) introduce new Carousel component

DeepakSap14:
https://github.com/SAP/fundamental-ngx/issues/3279

Introducing carousel component in core.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

Known issues:
1. Looping is not working for Multiple active item.
2. Mobile swipe functionality not implemented.


libs/core/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,309 @@
...
+@Component({
+    selector: 'fd-carousel',
+    templateUrl: './carousel.component.html',
+    styleUrls: ['./carousel.component.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    encapsulation: ViewEncapsulation.None
+})
+export class CarouselComponent implements OnInit, AfterContentInit, OnDestroy {
+    /** @hidden */
+    @ContentChildren(CarouselItemComponent)
```
sKudum, line 38: we need to follow angular coding guidelines for properties alignment. reference angular.io,  we can update where ever it is applicable

libs/core/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,309 @@
...
+    items: QueryList<CarouselItemComponent>;
+
+    /** Id for the Carousel. */
+    @Input()
+    @HostBinding('attr.id')
```
sKudum, line 43: It is recommended not to use @hostbinding()  along with @input()

libs/core/src/lib/carousel/carousel-item/carousel-item.component.ts
```diff
@@ -0,0 +1,77 @@
...
+let carouselItemUniqueId = 0;
+
+@Component({
+    selector: 'fd-carousel-item',
+    templateUrl: './carousel-item.component.html',
+    styleUrls: ['./carousel-item.component.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    encapsulation: ViewEncapsulation.None
+})
+export class CarouselItemComponent {
```
sKudum, line 19: Base component extension won't help here?

libs/core/src/lib/carousel/carousel-item/carousel-item.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  beforeEach(() => {
+    fixture = TestBed.createComponent(CarouselItemComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
```
sKudum, line 25: having few test scenarios will help.

libs/core/src/lib/carousel/carousel.module.ts
```diff
@@ -0,0 +1,11 @@
...
+import { CommonModule } from '@angular/common';
+import { CarouselComponent } from './carousel.component';
+import { CarouselItemComponent } from './carousel-item/carousel-item.component';
+
+@NgModule({
+    declarations: [CarouselComponent, CarouselItemComponent],
+    imports: [CommonModule],
+    exports: [CarouselComponent, CarouselItemComponent]
+})
+export class CarouselNewModule {}
```
sKudum, line 11: Only core we have a carousel why should we have "New"

sKudum (changes requested):

mikerodonnell89:
The images in the preview take a very long time to load if they're not cached, perhaps each example can share a service that will load the images so they'll only need to be downloaded once. Also I think that service itself is just very slow so maybe use a different service, placeimg seems to be faster

mikerodonnell89:
Thumb swiping for the mobile example (hidden navigation buttons) does not work

mikerodonnell89:
Playing with the width/height inputs, the carousel items overflow beyond the boundaries, is that intentional?

mikerodonnell89:
I think `<div style="display: flex; justify-content: center; margin-top: 0px;">` can be built in to the component rather than requiring the dev to add it themselves

droshev (changes requested):
have you implemented the sliding effect:
```
When the user navigates from the current item to another item, the current item is moved out of the content area, and the next or previous item slides in (depending on the direction of navigation).

On touch devices, users navigate with swipe gestures (swipe right or swipe left).
```
also the animation is missing when you change the component in

droshev:
> I think `<div style="display: flex; justify-content: center; margin-top: 0px;">` can be built in to the component rather than requiring the dev to add it themselves

I agree with @mikerodonnell89 and also it doesn't follow the markup from styles

droshev:
> ork for coz

the sliding should work on any device including the browser

apps/docs/src/app/core/component-docs/carousel/examples/carousel-content-navigation.component.html
```diff
@@ -0,0 +1,39 @@
+<fd-carousel [navigatorInPageIndicator]="false" [pageIndicatorContainerPlacement]="'top'">
```
kavya-b, line 1: remove `"'top'"` [] binding here and at other places

libs/core/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,315 @@
...
+@Component({
+    selector: 'fd-carousel',
+    templateUrl: './carousel.component.html',
+    styleUrls: ['./carousel.component.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    encapsulation: ViewEncapsulation.None
+})
+export class CarouselComponent implements OnInit, AfterContentInit, OnDestroy {
+    /** @hidden */
+    @ContentChildren(CarouselItemComponent, { descendants: true })
```
kavya-b, line 38: I think `@ContentChildren` is supposed to come after `Input()` and `Output()`, right?

libs/core/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,315 @@
...
+} from '@angular/core';
+import { Subject } from 'rxjs';
+import { takeUntil } from 'rxjs/operators';
+import { RtlService } from '../utils/services/rtl.service';
+import { CarouselItemComponent } from './carousel-item/carousel-item.component';
+
+const ICON_PAGE_INDICATOR_LIMIT = 8;
+let carouselUniqueId = 0;
+
+class CarouselActiveItem {
```
kavya-b, line 25: would this need to be exported?

apps/docs/src/app/core/component-docs/carousel/examples/carousel-no-page-indicator.component.html
```diff
@@ -0,0 +1,35 @@
+<fd-carousel [showPageIndicator]="false" [pageIndicatorContainerPlacement]="'top'">
```
Betrozov, line 1: unnecessary binding

Betrozov (approved):

JKMarkowski:
I agree with @droshev, this should be used with some functionality of `carousel.directive` from `utils`. There is already implemented swipe support

JKMarkowski:
@DeepakSap14 I preapred some service, by copy/paste some properties from `carousel.directive`. Check it there: https://github.com/SAP/fundamental-ngx/blob/fix/create-carousel-service/libs/core/src/lib/utils/services/carousel.service.ts

There you can see how it can be used:
https://github.com/SAP/fundamental-ngx/blob/fix/create-carousel-service/libs/core/src/lib/utils/directives/carousel/carousel.directive.ts

KevinOkamoto (approved):
Looks good.

libs/core/src/lib/carousel/carousel.component.ts
DeepakSap14: I see its getting used in core components. Do you have any other approach?

libs/core/src/lib/carousel/carousel.component.ts
DeepakSap14: fixed. followed angular guidelines.

libs/core/src/lib/carousel/carousel-item/carousel-item.component.ts
DeepakSap14: Base component is in platform. carousel is in core. so will not be able to use base component.

libs/core/src/lib/carousel/carousel-item/carousel-item.component.spec.ts
DeepakSap14: have included tests in carousel component for carousel items as well. removed this file.

libs/core/src/lib/carousel/carousel.module.ts
DeepakSap14: using CarouselModule now.

apps/docs/src/app/core/component-docs/carousel/examples/carousel-content-navigation.component.html
DeepakSap14: fixed

libs/core/src/lib/carousel/carousel.component.ts
DeepakSap14: fixed

libs/core/src/lib/carousel/carousel.component.ts
DeepakSap14: fixed. export added

apps/docs/src/app/core/component-docs/carousel/examples/carousel-no-page-indicator.component.html
DeepakSap14: fixed

DeepakSap14:
Now i have included Sliding effect for browsers and swiping effect for touch devices as well.
i have included example where cards can be used as carousel item as well.
please review now.

DeepakSap14:
> The images in the preview take a very long time to load if they're not cached, perhaps each example can share a service that will load the images so they'll only need to be downloaded once. Also I think that service itself is just very slow so maybe use a different service, placeimg seems to be faster

Creating service to fetch these images gives Cross-origin error.
I have changed the image provider to "placeimg", hope this will be good enough.

libs/core/src/lib/utils/services/carousel.service.ts
```diff
@@ -0,0 +1,297 @@
...
+        this._elementRef.nativeElement.style.transitionDuration = '0s';
+        this._prePanTransitionPx = this._currentTransitionPx;
+        this.dragStateChange.emit(true);
+    }
+
+    /** @hidden */
+    private _hammerSetup(): void {
+        this._hammer = new Hammer(this._elementRef.nativeElement);
+
+        this._hammer.get('pan').set({ direction: Hammer.DIRECTION_ALL });
```
mykolavorobiov, line 276: Could please apply direction depends on your config? for example, `this.config.vertical` should use only `Hammer.DIRECTION_VERTICAL`

libs/core/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,404 @@
...
+                this.leftButtonDisabled = true;
+                this.rightButtonDisabled = true;
+            }
+            if (this.currentActiveSlidesStartIndex === 0) {
+                this.leftButtonDisabled = true;
+            } else if (this.currentActiveSlidesStartIndex === this.slides.length - 1) {
+                this.rightButtonDisabled = true;
+            }
+
+            if (this.currentActiveSlidesStartIndex > 0) {
```
mykolavorobiov, line 341: Could you please move `this.leftButtonDisabled = false;` and `this.rightButtonDisabled = false;` to above construction with `...else...`, and remove unnecessary conditions

`if (this.currentActiveSlidesStartIndex === 0) {`
`                this.leftButtonDisabled = true;`
`            } else if (this.currentActiveSlidesStartIndex === this.slides.length - 1) {`
`                this.rightButtonDisabled = true;`
`            } else {`
`this.leftButtonDisabled = false;`
`this.rightButtonDisabled = false;`
`}`

libs/core/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,404 @@
...
+                this.currentActiveSlidesStartIndex + this.visibleSlidesCount >= this.slides.length
+            ) {
+                this.rightButtonDisabled = true;
+            }
+
+            if (this.slides.length === 1) {
+                this.leftButtonDisabled = true;
+                this.rightButtonDisabled = true;
+            }
+            if (this.currentActiveSlidesStartIndex === 0) {
```
mykolavorobiov, line 335: Looks like the same as the above conditions.

mykolavorobiov (changes requested):
Looks good, please take a look few minor comments

JKMarkowski (changes requested):
Hi @DeepakSap14, good job, most of examples work great, but

carousel

As you can see slide doesn't work good, indicator doesn't stop on last element. 
Also TimeComponent doesn't work with those changes
image

libs/core/src/lib/utils/services/carousel.service.ts
DeepakSap14: @JKMarkowski , do you want me to make this change. this code was present from earlier. so please confirm.

DeepakSap14: changed to Hammer.DIRECTION_VERTICAL inside this.config.vertical condition.
fixed.

libs/core/src/lib/carousel/carousel.component.ts
DeepakSap14: fixed.

libs/core/src/lib/carousel/carousel.component.ts
DeepakSap14: fixed.

DeepakSap14:
> Hi @DeepakSap14, good job, most of examples work great, but
> 
> carousel
> 
> As you can see slide doesn't work good, indicator doesn't stop on last element.
> Also TimeComponent doesn't work with those changes
> image

1. Fixed, extra swiping happening in carousel service in case of multiple item display.
2. Time component is using "carousel directive" and internally it is using "fd-carousel" class.
     so namespace conflict is happening for "fd-carousel", when page is switched from "carousel" to "time" component.
    I tried directly importing "fundamental-styles/carousel.scss" in time-column.component.scss but this also does not resolve 
    issue.
   **As a temporary fix**: i have disabled "fd-carousel" class load on "fdCarouselDirective" by appending extra characted to class name.
  This does not introduce any issue in time component examples, all examples works good in time component.
  If needed we have to raise a separate issue and provide fix in "time component".

DeepakSap14:
@JKMarkowski please review this.

JKMarkowski:
Hi @DeepakSap14 I need to ask you how the API there. I see multi item doesn't have any `[swipeEnabled]` and the swipe works fine (on desktop). For mobile swipe works everywhere and it doesn't matter if `[swipeEnabled]` is set to true

JKMarkowski:
carouselloop
After enabling swipe on loop carousel, it's a little bit bugged

droshev:
> carouselloop
> After enabling swipe on loop carousel, it's a little bit bugged

Can we enable this with mouse down + swipe left/right?

DeepakSap14:
> > carouselloop
> > After enabling swipe on loop carousel, it's a little bit bugged
> 
> Can we enable this with mouse down + swipe left/right?

swipe event is coming from hammerjs. handling this with key will not be possible.
we have left and right arrow navigation alrerady.

DeepakSap14:
@JKMarkowski fixed bug for looped carousel. please check now.

DeepakSap14:
> Hi @DeepakSap14, Great Job with it! I just need to ask you how the API there. I see [multi item example](https://deploy-preview-3551--fundamental-ngx.netlify.app/#/core/carousel#carousel-multi-item-active) doesn't have any `[swipeEnabled]` and the swipe works fine (on desktop). For mobile swipe works fine everywhere.

By default i have made swipeEnabled="true", so it will be work in touch devices without any change.

DeepakSap14:
@JKMarkowski @droshev Please review and approve.
It has passed expected due date. If some minor issue, i will raise a different issue to track.

sKudum (approved):
My comments are addressed

libs/core/src/lib/fundamental-ngx-core.module.ts
```diff
@@ -142,6 +143,6 @@ import { MessageToastService } from './message-toast/message-toast-service/messa
         CardModule,
         MessageToastModule
     ],
-    providers: [AlertService, DialogService, NotificationService, MessageToastService]
+    providers: [AlertService, CarouselService, DialogService, NotificationService, MessageToastService]
```
dimamarksman, line 146: We don't need to include `CarouselDirective` since it's defined as provider on the CarouselDirective level.
Please correct me if I'm wrong.

dimamarksman (approved):
Looks good!
Just a few questions

JKMarkowski:
IE11
ie11carousel

JKMarkowski:
Multiple items bug
carouselgfi

libs/core/src/lib/carousel/carousel.component.ts
JKMarkowski: IMO IT's correct

libs/core/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,412 @@
...
+    @Input()
+    vertical = false;
+
+    /** Number of items to be visible at a time */
+    @Input()
+    visibleSlidesCount = 1;
+
+    /** An event that is emitted after a slide transition has happened */
+    @Output()
+    readonly slideChange: EventEmitter<CarouselActiveSlides> = new EventEmitter<CarouselActiveSlides>();
```
JKMarkowski, line 174: Could you put API properties in valid places?

JKMarkowski:
Hello @DeepakSap14 Rest LGTM. It's almost ready to be merged, but those 2 bugs need to be addressed

libs/core/src/lib/carousel/carousel.component.ts
DeepakSap14: fixed

DeepakSap14:
> Multiple items bug
> carouselgfi

This is fixed now.

DeepakSap14:
> IE11
> ie11carousel

currently do not have IE11 access. will check and fix in some time.
this should not be happening. May be image did not load completely. please verify once.

libs/core/src/lib/fundamental-ngx-core.module.ts
DeepakSap14: fixed now. Does not need service here.

apps/docs/src/app/core/component-docs/carousel/examples/carousel-looped-navigation.component.ts
```diff
@@ -0,0 +1,9 @@
+import { ChangeDetectionStrategy, Component, ViewEncapsulation } from '@angular/core';
+
+@Component({
+    selector: 'fd-carousel-looped-navigation',
+    templateUrl: './carousel-looped-navigation.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    encapsulation: ViewEncapsulation.None
```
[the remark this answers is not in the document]
DeepakSap14: fixed

apps/docs/src/app/core/component-docs/carousel/carousel-docs.component.html
```diff
@@ -0,0 +1,96 @@
...
+<fd-docs-section-title id="carousel-multi-item-active" componentName="carousel">
+    Carousel with vertical direction
+</fd-docs-section-title>
+<description>
+    carousel movement is possible in vertical direction as well.
+</description>
+<component-example>
+    <fd-carousel-vertical-direction></fd-carousel-vertical-direction>
+</component-example>
+<code-example [exampleFiles]="carouselVertical"></code-example>
```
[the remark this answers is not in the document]
DeepakSap14: had wrong initial height for carousel. fixed now.

DeepakSap14:
> IE11
> ie11carousel

Looking into this.

DeepakSap14:
> > IE11
> > ie11carousel
> 
> Looking into this.

This is fixed now using max-width: 100%, width:auto, height:auto.

DeepakSap14:
Branch conflict will be resolved once platform table fixes use of "KeyUtil.isKey". on latest master rebase build error coming from platform table component.

JKMarkowski (approved):
LGTM. It's ready for merge

droshev (changes requested):
Great work! Few things:
- it looks like there is a double border on the bottom part:
Screen Shot 2020-10-21 at 9 26 36 PM
- there is a weird effect when you try to drag images:
carousel
- can we achieve this sliding with the mouse from the carousel with multiple items in the carousel with 1 active item?
sliding-carousel

InnaAtanasova:
Hey @DeepakSap14 ,
I found some minor issues testing the component on a mobile device: 
- [ ] if you slide slowly sometimes two images are visible at the same time. I understand it's a performance issue but if it's not addressed now could you create an issue for targeting this in the future. Here's a screenshot of the issue:
Image from iOS (3)

- [ ] Landscape view, you can't make the last card appear in the view. The right button is disabled and the last "dot" is marked as active but you basically can't display the last item. See screenshot:
Image from iOS (2)

- [ ] this is similar to the issue above. In portrait mode you can't display cards 7th and 8th:
Image from iOS (1)

- [ ] Message page is not centered on a mobile device
Image from iOS

DeepakSap14:
> Great work! Few things:
> 
> * it looks like there is a double border on the bottom part:
> 
> Screen Shot 2020-10-21 at 9 26 36 PM
> 
> * there is a weird effect when you try to drag images:
>   carousel
> * can we achieve this sliding with the mouse from the carousel with multiple items in the carousel with 1 active item?
>   sliding-carousel

1. carousel styles does not define focus. So each browser was behaving differently by default. Now for documentation i have included, fd-carousel:focus styles. It is consistent now.
2. fixed by pointer-event: none style.
3. fixed. swipe working for single active item as well.

DeepakSap14:
> Hey @DeepakSap14 ,
> I found some minor issues testing the component on a mobile device:
> 
> * [ ]  if you slide slowly sometimes two images are visible at the same time. I understand it's a performance issue but if it's not addressed now could you create an issue for targeting this in the future. Here's a screenshot of the issue:
>   Image from iOS (3)
> * [ ]  Landscape view, you can't make the last card appear in the view. The right button is disabled and the last "dot" is marked as active but you basically can't display the last item. See screenshot:
>   Image from iOS (2)
> * [ ]  this is similar to the issue above. In portrait mode you can't display cards 7th and 8th:
>   Image from iOS (1)
> * [ ]  Message page is not centered on a mobile device
>   Image from iOS

@InnaAtanasova 
1. I did not get this issue. When animation effect is sliding. then in this current active item slides out and new item slides in. In this process, until sliding is completed. part of both item will be visible. see this example: http://css3.bradshawenterprises.com/sliding/

2. fixed hidden card issue in Mobile view portrait and landscape.
3. fixed hidden card issue in Mobile view portrait and landscape.
4. Error message is center aligned in every screen now.

DeepakSap14:
@InnaAtanasova @droshev This can be reviewed now.
I will hold this for merge as you had suggested.

InnaAtanasova (approved):
@DeepakSap14 I verified that my comments have been addressed. Thanks for doing this. Also, great job with the component.

droshev (changes requested):
carousel
looks much better!
One thing i noticed when you Tab over the carousel with multiple items the content gets shifted but the little dots indicating the current page don't

manavs:
> carousel
> looks much better!
> One thing i noticed when you Tab over the carousel with multiple items the content gets shifted but the little dots indicating the current page don't.

May be I am viewing this incorrectly, but looks like when tabbing through to the card which is current is always the middle one. Not sure whether that it part of the requirement spec. This happens when the new set of cards come under view

DeepakSap14:
> carousel
> looks much better!
> One thing i noticed when you Tab over the carousel with multiple items the content gets shifted but the little dots indicating the current page don't.

@droshev 
taborder issue has been fixed now.
please review.

droshev (commented):
@DeepakSap14 Great work! LGTM 🚀

droshev (approved):

PR #3601, opened 2020-10-12
fix: Plugin configuration changes

valorkin:

### DIFF
--- a/apps/app-shell-app/src/app/app.component.ts
+++ b/apps/app-shell-app/src/app/app.component.ts
@@ -78,7 +78,7 @@ export class AppComponent implements OnInit {
         this.appShell.shellBar.setTitle('One Procurement');
         this.appShell.shellBar.setSubTitle('Solution for all');
 
-        this.appShell.messageBus.sendTo('error:event',
+        this.appShell.messageBus.publish('error:event',
             new TextMessage('There was a error during Appinitialization'));
     }
 }
--- a/libs/app-shell/src/lib/api/events/messaging.service.spec.ts
+++ b/libs/app-shell/src/lib/api/events/messaging.service.spec.ts
@@ -92,10 +92,10 @@ describe('MessagingService', () => {
                 });
                 const m = new TextMessage('system:events');
                 m.text = 'app registered';
-                service.sendTo('system:events', m);
+                service.publish('system:events', m);
 
                 let expectedMessage;
-                service.onMessage('system:events', (msg: Message) => {
+                service.subscribe('system:events', (msg: Message) => {
                     expectedMessage = (msg as TextMessage).text;
                 });
                 expect(expectedMessage).toBeUndefined();
@@ -139,13 +139,13 @@ describe('MessagingService', () => {
                     eventType: EventType.DEFAULT, shared: true, prefix: 'system:'
                 });
                 let expectedMessage = 'none';
-                service.onMessage('system:events', (msg: Message) => {
+                service.subscribe('system:events', (msg: Message) => {
                     expectedMessage = (msg as TextMessage).text;
                 });
 
                 const m = new TextMessage('system:events');
                 m.text = 'app registered';
-                service.sendTo('system:events', m);
+                service.publish('system:events', m);
 
                 expect(expectedMessage).toBe(m.text);
             }
--- a/libs/app-shell/src/lib/api/events/topics.service.ts
+++ b/libs/app-shell/src/lib/api/events/topics.service.ts
@@ -5,23 +5,19 @@ import { Injectable } from '@angular/core';
 /**
  * Its important to define topic first, which can set some parameters up front
  */
-@Injectable()
+@Injectable({ providedIn: 'root' })
 export class MessagingTopics {
-    public topicsDef: Array<Topic> = [];
-
-    constructor() {
-
-    }
+    public topicsDef = new Map<string, Topic>();
 
     defineTopic(topic: Topic): void {
-        const length = this.topicsDef.filter((t) => t.name === topic.name).length;
-        if (length === 0) {
-            this.topicsDef.push(topic);
+        if (this.topicsDef.has(topic.name)) {
+            return;
         }
+        this.topicsDef.set(topic.name, topic);
     }
 
     hasTopic(name: string): boolean {
-        return this.topicsDef.filter((t) => t.name === name).length > 0;
+        return this.topicsDef.has(name);
     }
 
     /**
@@ -29,8 +25,11 @@ export class MessagingTopics {
      *
      */
     getTopic(name: string): Topic {
-        const topics = this.topicsDef.filter((t) => t.name === name);
-        return topics[0];
+        if (this.topicsDef.has(name)) {
+            return this.topicsDef.get(name);
+        }
+
+        return this.topicsDef.entries().next().value;
     }
 }
 
--- a/libs/app-shell/src/lib/api/extensions/lookup/plugin-descriptor.model.ts
+++ b/libs/app-shell/src/lib/api/extensions/lookup/plugin-descriptor.model.ts
@@ -1,31 +1,50 @@
-export interface PluginDescriptor {
+export interface AngularIvyComponentDescriptor {
+    /** Uniq name of module in plugin */
     name: string;
-    provider: string
-    version: string;
-    sinceVersion: string;
-    untilVersion: string;
-    remoteEntry: string;
-    remoteName: string;
-    remoteRoute: string;
+    /** Type of a plugin module */
+    type: 'angular-ivy-component';
+    /** Exposed Angular Module */
     exposedModule: string;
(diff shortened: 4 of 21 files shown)


libs/app-shell/src/lib/api/events/messaging.service.spec.ts
```diff
@@ -139,13 +139,13 @@ describe('MessagingService', () => {
                     eventType: EventType.DEFAULT, shared: true, prefix: 'system:'
                 });
                 let expectedMessage = 'none';
-                service.onMessage('system:events', (msg: Message) => {
+                service.subscribe('system:events', (msg: Message) => {
```
fkolar, line 142: I had Subscribe before, and I changed to onMessage as I though its more intuitive. If you think subscribe feels better we will leave this

libs/app-shell/src/lib/api/events/messaging.service.spec.ts
```diff
@@ -139,13 +139,13 @@ describe('MessagingService', () => {
...
                     expectedMessage = (msg as TextMessage).text;
                 });
 
                 const m = new TextMessage('system:events');
                 m.text = 'app registered';
-                service.sendTo('system:events', m);
+                service.send('system:events', m);
```
fkolar, line 148: When you renamed subcription to subscribe, should not we call thsi publish ?

libs/app-shell/src/lib/components/plugin-launcher/plugin-launcher.component.ts
```diff
@@ -61,22 +52,24 @@ export class PluginLauncherComponent implements OnChanges {
...
     }
 
 
-    async doCreateComponent(descriptor: Partial<PluginDescriptor>): Promise<void> {
-        const component = await loadRemoteModule(descriptor)
-            .then(m => m[descriptor.componentName]);
-        const factory = this.cfr.resolveComponentFactory(component);
+    async doCreateComponent(descriptor: Partial<PluginDescriptor>, moduleName: string): Promise<void> {
+        const _module = descriptor.modules.find(module => module.name === moduleName);
+        if (_module.type === 'angular-ivy-component') {
```
fkolar, line 61: How do we have webcomponents ? When I used plugin launcher to execute all , and I had little IF ELSE that works with webcomponents

fkolar (changes requested):
More clarifications than request changes. Cleanup looks good

libs/app-shell/src/lib/api/extensions/lookup/plugin-descriptor.model.ts
```diff
@@ -1,31 +1,50 @@
...
-    displayName: string;
+    /** Component name we are planning to inject */
     componentName: string;
-
-    type: Scope;
-    category: string;
-    hasRoutes: boolean;
-    changeNotes: string;
+    /** most likely this property will be removed, but as for this this is a way to register route for a component*/
+    route?: string;
```
fkolar, line 11: I think you mixed hasRoute original property, but this Route is still needed , this is our current routing solution

libs/app-shell/src/lib/api/extensions/lookup/plugin-descriptor.model.ts
```diff
@@ -1,31 +1,50 @@
...
 }
 
+export interface IframePageDescriptor {
```
fkolar, line 14: Can you have one common interface PLuginDescription that you have below whcih has common properties and let the others extends it ?

libs/app-shell/src/lib/api/extensions/lookup/plugin-descriptor.model.ts
```diff
@@ -1,31 +1,50 @@
...
-/**
- * The idea behind the scope is that there are components that are instantiated at the time we use Page launcher.
- * But there could be a situation, where we need to register a PLUGIN at the bootstrap at the lauch time, there we
- * have scope Application.
- * Not implemented yet default value should be as Page
- */
-export enum Scope {
-    Application = 1,
-    Page
+export interface AngularElementComponentDescriptor {
```
fkolar, line 23: This should not be an Angular Element but more CustomElementComponentDesriptor, this is our current demo, what I had. we dont have demo for AngularElements

fkolar (changes requested):

libs/app-shell/src/lib/api/events/topics.service.ts
```diff
@@ -5,32 +5,31 @@ import { Injectable } from '@angular/core';
 /**
  * Its important to define topic first, which can set some parameters up front
  */
-@Injectable()
+@Injectable({ providedIn: 'root' })
 export class MessagingTopics {
```
fkolar, line 9: We dont want to expose this service yet outside adn we are not even exporting it to application. Should this still be providedin ROOT ? 


You are not able to create new topic or subsribe to topic if its not defined. and this is responsibility of internal api. I dont want application to create random topics.

fkolar (changes requested):
I dont have anything else. overall it looks good.

apps/app-shell-app/src/app/app.component.ts
```diff
@@ -78,7 +78,7 @@ export class AppComponent implements OnInit {
         this.appShell.shellBar.setTitle('One Procurement');
         this.appShell.shellBar.setSubTitle('Solution for all');
 
-        this.appShell.messageBus.sendTo('error:event',
+        this.appShell.messageBus.send('error:event',
```
fkolar, line 81: There is app-shell-app under `apps` when you changed signature please update the app as well. I should not compile if you do ng build app-shell-app

fkolar (changes requested):

libs/app-shell/src/lib/api/events/messaging.service.spec.ts
valorkin: Angular devs, subscribe/unsubscribe/emit are native words now ;)

libs/app-shell/src/lib/api/events/messaging.service.spec.ts
valorkin: I think it should be emit. Send is a kafka word, but 'send a message' sounds better then 'emit message'. Publish could be better. Because it will be clear pub/sub

libs/app-shell/src/lib/components/plugin-launcher/plugin-launcher.component.ts
valorkin: I will be adding it tomorrow. Iframe, web components...

libs/app-shell/src/lib/api/extensions/lookup/plugin-descriptor.model.ts
valorkin: Yes it's, but it's rather optional. Right?

libs/app-shell/src/lib/api/events/messaging.service.spec.ts
fkolar: emit is ugly its too functional and not human friendly ;-) We have publishers/ subcribers, so subcribe and publish

libs/app-shell/src/lib/components/plugin-launcher/plugin-launcher.component.ts
fkolar: ok

libs/app-shell/src/lib/api/extensions/lookup/plugin-descriptor.model.ts
fkolar: yes tru tru

libs/app-shell/src/lib/api/extensions/lookup/plugin-descriptor.model.ts
valorkin: This is an interfaces for modules. One container has several exported modules, one uri could host several iframes to inject, etc 
But share same name and version

libs/app-shell/src/lib/api/extensions/lookup/plugin-descriptor.model.ts
valorkin: It's not final, I will check what we actually have in bundles

libs/app-shell/src/lib/api/events/topics.service.ts
valorkin: Yes, root. We don't have expose it, but we don't want to have duplicates

libs/app-shell/src/lib/api/extensions/lookup/plugin-descriptor.model.ts
fkolar: Yes, they definitelly shoudl share some common properties.

fkolar (approved):

fkolar (approved):

PR #3626, opened 2020-10-13
feat(app-shell): updated launcher to include custom elements and routing fix

valorkin:


libs/app-shell/src/lib/api/error/default-error-handler.service.ts
```diff
@@ -48,9 +48,10 @@ export class DefaultErrorHandlerService implements ErrorHandler, OnDestroy {
                 @Inject(ERROR_FORMATTER) private formatter: ErrorFormatter,
                 @Inject(ERROR_NOTIFIERS) @Optional() private notifiers: ErrorNotifier[]) {
 
-        this.topics.defineTopic({
-            prefix: 'error:', eventType: EventType.DURABLE,
-            name: TOPIC_ERROR_EVENT, shared: true
+        this.topics.set({
```
fkolar, line 51: We should leave `defineTopic`. we are not setting topic. 

The purpose of this is that you want to `create topic definition` which is later used to create subscribers / publishers. so the defineTopic is the correct one. 

If topic definition does not exists then you cant really create some random topics.

libs/app-shell/src/lib/api/events/topics.service.ts
```diff
@@ -9,22 +9,22 @@ import { Injectable } from '@angular/core';
 export class MessagingTopics {
     public topicsDef = new Map<string, Topic>();
 
-    defineTopic(topic: Topic): void {
+    set(topic: Topic): void {
```
fkolar, line 12: Let's stick with define. We dont need call it defineTopic, but define should be there. 

Not even create, you are not creating this topic yet. you just creating definition of it.

fkolar (approved):

PR #3630, opened 2020-10-15
chore: Remove redundant build command from netlify

JKMarkowski:
It seems like one of commands is not necessary and takes time from netlify's build

### Previous PR netlify build:
7m 17s
image

### This PR netlify build:
6m 11s
image


salarenko (approved):
Revert changes done for testing purposes and LGTM

PR #3640, opened 2020-10-16
fix: (Core) refactor Layout Grid with latest changes from fundamental-styles

mikerodonnell89:
Layout grid has received a significant refactor with many breaking changes.  Note most code in this PR was done by @salarenko and I just wrote tests/docs

Fundamental Styles issues found during implementation: 
- [Offset in RTL](https://github.com/SAP/fundamental-styles/issues/1801) 

BREAKING CHANGE:
The layout grid is no longer a series of components but rather directives. Here are a few other breaking changes:
* `gapSize` input replaced with `fdLayoutGridColOffset`
* `nogap` replaced with `noGap`
* responsiveness is achieved by setting individual sizes for each of the following: `fdLayoutGridCol fdLayoutGridColMd fdLayoutGridColLg fdLayoutGridColXl`
* individual rows can be set with the `fdLayoutGridRow` directive.


stefanoScalzo:
We need examples of the following

- [ ] Responsive columns and offsets

- [ ] Nesting

- [ ] Auto adjusting

salarenko:
I've introduces some changes:
- Moved responsive columns and responsive offsets to separate examples for better readability.
- Moved the logic used to apply `.fd-col` class to `LayoutGridColDirective` only.
- Commented out the API responsible for auto-adjustment behavior which is not available yet.

@stefanoScalzo 
- Auto adjusting is not available in currently used Fundamental Styles version
- The other examples have been added to the documentation

@mikerodonnell89 
- The documentation is a bit poor. It lacks in information how grid actually behaves, how grid elements should be used, what are the breaking points and many more etc. Could you improve it a bit.
- Did you document the breaking changes? I don't see them described in the PR.

stefanoScalzo:
We still need to improve the documentation,
We need to define what the numbers correspond to
What about having a config being passed instead of having a bunch of inputs?
Nesting example: "Add a row" not "Add an row"

salarenko:
@stefanoScalzo I also thought about the "config" approach, but in this case I find the directives much easier to use and read. Anyway I'm open for discussion, we can talk this through during one of the meetings.

JKMarkowski (commented):
There is 1 minor thing. I will continue review after changing directives to inputs, as we discussed

salarenko:
@stefanoScalzo @JKMarkowski 
I've introduced changes according to our latest agreements:
- Abandon separate directives approach and introduce one col directive with inputs
- Implement `CssClassBuilder`

libs/core/src/lib/layout-grid/directives/layout-grid-col.directive.ts
```diff
@@ -0,0 +1,88 @@
+import { Directive, ElementRef, Input, OnChanges, OnInit } from '@angular/core';
+import { CSS_CLASS_NAME, GRID_COLUMNS_NUMBER } from '../constants';
+import { applyCssClass } from '../../utils/decorators/apply-css-class.decorator';
+import { CssClassBuilder } from '../../utils/interfaces/css-class-builder.interface';
+
+@Directive({
+    selector: '[fdLayoutGridCol]'
```
JKMarkowski, line 7: Could you add also `fd-layout-grid-col` selector, just to keep same convention as on other directives?

libs/core/src/lib/layout-grid/directives/layout-grid-row.directive.ts
```diff
@@ -0,0 +1,9 @@
+import { Directive } from '@angular/core';
+
+@Directive({
+    selector: '[fdLayoutGridRow]',
```
JKMarkowski, line 4: Could you add also `fd-layout-grid-row` selector, just to keep same convention as on other directives?

JKMarkowski (approved):
Great Job! I left 2 minor comments, but the rest LGTM

libs/core/src/lib/layout-grid/directives/layout-grid-row.directive.ts
salarenko: Done

libs/core/src/lib/layout-grid/directives/layout-grid-col.directive.ts
salarenko: Done

stefanoScalzo:
The row example should come after before the column example
Nesting columns should be before responsive columns
Explain how to use responsive offset
State breakpoints for responsiveness

mikerodonnell89:
> The row example should come after before the column example
> Nesting columns should be before responsive columns

Why?

> Explain how to use responsive offset

The responsiveness and offsets work together without any additional effort

InnaAtanasova (approved):

PR #3678, opened 2020-10-22
feat: list with byline migration

### LINKED ISSUE #3189: List with Byline migration
There is a need to migrate list with byline to NGX. Previously only standard list was added

mikerodonnell89:
fixes #3189 

Adds the byline list, used by adding the `[byline]` input to the `fd-list`.  Requires various sub-directives, see docs


salarenko (changes requested):
- Maybe we could move the documentation for standard list and byline list into separate sub-pages?
```
List
|__ Standard List
|__ Byline List
```

- Similar for the source files. Right now all list building blocks are mixed, for example all over 20 directives sitting in one directory. Maybe we could split it and add index barrels for easier maintenance?

- There is a relatively new, but appreciated an used approach for declaring and using Fundamental Styles classes in Fundamental NGX (const CSS_CLASS_NAME). Could you update your implementation to also use it?

mikerodonnell89:
> * Maybe we could move the documentation for standard list and byline list into separate sub-pages?
> 
> ```
> List
> |__ Standard List
> |__ Byline List
> ```

Well each component is already a nested list item in the side nav, so we can't go any deeper without a big refactor.  But I could move byline list to its own page, do we think that's a good idea?  It's what we did for button/segmented button so i'd be fine with it

mikerodonnell89:
> * Similar for the source files. Right now all list building blocks are mixed, for example all over 20 directives sitting in one directory. Maybe we could split it and add index barrels for easier maintenance?

these are all being exported with the barrel as public_api.ts

libs/core/src/lib/list/directives/list-byline.directive.ts
```diff
@@ -0,0 +1,14 @@
+import { Directive, Input } from '@angular/core';
+
+@Directive({
+    selector: '[fdListByline], [fd-list-byline]',
+    host: {
+        class: 'fd-list__byline',
+        '[class.fd-list__byline--2-col]': 'twoCol === true',
```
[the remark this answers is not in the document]
mikerodonnell89: So should I not use the class name constants file? I'm not a big fan of that whole practice, seems unnecessary when find and replace exists

apps/docs/src/app/core/component-docs/list-byline/examples/list-byline-standard-example/list-byline-standard-example.component.html
```diff
@@ -0,0 +1,66 @@
...
+    </li>
+    <li fd-list-item>
+        <div fd-list-content>
+            <div fd-list-title>Text-Only List Item</div>
+            <div fd-list-byline>Byline Text (description)</div>
+        </div>
+    </li>
+</ul>
+
+<h4>Compact</h4>
```
JKMarkowski, line 35: Do you need compact mode? I see there is no difference between compact and cozy modes

apps/docs/src/app/core/component-docs/list-byline/list-byline-docs.module.ts
```diff
@@ -0,0 +1,62 @@
...
+        SharedDocumentationPageModule,
+        ListModule,
+        LinkModule,
+        CheckboxModule,
+        RadioModule,
+        InfiniteScrollModule,
+        DragAndDropModule,
+        BusyIndicatorModule,
+        ToolbarModule,
+        InputGroupModule,
```
JKMarkowski, line 48: Can you remove unused imports?

JKMarkowski (commented):
Hi @mikerodonnell89  Good Job! I left some minor comments, let me know when these are addressed, so I will rereview

apps/docs/src/app/core/component-docs/list-byline/examples/list-byline-standard-example/list-byline-standard-example.component.html
InnaAtanasova: The difference is the font-size of the title `fd-list__title`. In mobile it's 1rem, in compact is 0.875rem

JKMarkowski (approved):

apps/docs/src/app/core/component-docs/list/list-docs.module.ts
```diff
@@ -20,7 +20,7 @@ import {
     DragAndDropModule,
     BusyIndicatorModule,
     ToolbarModule,
-    InputGroupModule
+    InputGroupModule, FundamentalNgxCoreModule
```
salarenko, line 23: I don't think we should import whole lib here `FundamentalNgxCoreModule`

apps/docs/src/app/core/component-docs/list-byline/list-byline-docs.module.ts
```diff
@@ -0,0 +1,53 @@
...
+@NgModule({
+    imports: [
+        RouterModule.forChild(routes),
+        SharedDocumentationPageModule,
+        ListModule,
+        LinkModule,
+        CheckboxModule,
+        RadioModule,
+        SharedDocumentationModule,
+        FundamentalNgxCoreModule
```
salarenko, line 41: I don't think we should import whole lib here FundamentalNgxCoreModule.

salarenko (changes requested):
LGTM, but:

salarenko (approved):

PR #3707, opened 2020-10-29
feat: implement title component

### LINKED ISSUE #3703: Implement Title directive
Implement the Title directive for NGX

https://fundamental-styles.netlify.app/?path=/docs/components-title--levels

Screen Shot 2020-10-28 at 5 10 13 PM

mikerodonnell89:
fixes #3703 

Surprise PR as I found this is something we need (in addition to some other things) in order to implement the wizard summary page.

Screen Shot 2020-10-29 at 11 25 55 AM

### DIFF
--- a/libs/core/src/lib/title/title.component.ts
+++ b/libs/core/src/lib/title/title.component.ts
@@ -1,0 +1,28 @@
+import { ChangeDetectionStrategy, Component, ElementRef, Input, OnInit, ViewEncapsulation } from '@angular/core';
+
+@Component({
+    // tslint:disable-next-line
+    selector: 'h1[fd-title], h2[fd-title], h3[fd-title], h4[fd-title], h5[fd-title], h6[fd-title]',
+    exportAs: 'fd-title',
+    template: `<ng-content></ng-content>`,
+    host: {
+        class: 'fd-title',
+        '[class.fd-title--wrap]': 'wrap'
+    },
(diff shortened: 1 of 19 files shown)


libs/core/src/lib/title/title.component.ts
```diff
@@ -0,0 +1,28 @@
...
+    /** Whether or not the title should wrap (truncates by default) */
+    @Input()
+    wrap = false;
+
+    constructor(private _elRef: ElementRef) {}
+
+    /** @hidden */
+    ngOnInit(): void {
+        const headerSize = this._elRef.nativeElement.tagName.charAt(1);
+        this._elRef.nativeElement.classList.add('fd-title--h' + headerSize);
```
InnaAtanasova, line 26: I think in this case the user should pass the value for the header. The way it's now (`Visual` example) the compiled code is:
Screen Shot 2020-10-30 at 12 22 25 PM
The html tag matches the `fd-title--hX` class. 
In styles it's different:
Screen Shot 2020-10-30 at 12 22 37 PM

The idea is that semantically it would make sense to use `h1`, for example, but the user might want to change the default font-size and have something like:
`<h1 class="fd-title fd-title--h6">Title Level 1</h1>`

InnaAtanasova (changes requested):

InnaAtanasova (approved):

PR #3713, opened 2020-10-30
test: fix unit tests errors due to imports

JKMarkowski:
There are imports changed, to get rid of mysterious errors on unit tests


InnaAtanasova (approved):
I tested and the tests are running on my side. There's still a warning about the focus-trap tho.

mikerodonnell89:
This is not working for me, same error about focus trap

PR #3759, opened 2020-11-05
chore: (Core) add hammerjs dependency to the library

JKMarkowski:
There is added hammerjs dependency on library + sync of version


InnaAtanasova (changes requested):
Please fix the title of the PR, the rest is good

InnaAtanasova (approved):

Issue #3720, opened 2020-11-02
Bug: Cursor on disabled "Switch" should not change state on hover over or should indicate "Not-Allowed".


pierrom:
#### Briefly describe your proposal.
Cursor on disabled switch should not change state on hover over or should show "Not-Allowed" cursor (see 2nd screenshot).

#### If this is a bug, please provide steps for reproducing it.
Go to: https://sap.github.io/fundamental-ngx/#/core/switch
Scroll down to: Switch used within Reactive Form or any section that has disabled switch
Hover: Any disabled switch
See error below, screen shot attached
![image](https://user-images.githubusercontent.com/71797052/97821080-f0133f80-1c65-11eb-9b58-ed3b479cf662.png)
![image](https://user-images.githubusercontent.com/71797052/97821087-f7d2e400-1c65-11eb-93d2-44e8260c1850.png)

PR #3769, opened 2020-11-06
fix: (Core) cursor on disabled Switch should not be pointer

olshansky:
fixes https://github.com/SAP/fundamental-ngx/issues/3720
switch-disabled

Documentation checklist:
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/switch/switch.component.html
+++ b/libs/core/src/lib/switch/switch.component.html
@@ -1,4 +1,4 @@
-<label class="fd-switch__label" (click)="focus()">
+<label class="fd-switch__label" (click)="focus()" [ngStyle]="{ cursor: disabled ? 'auto' : 'pointer'}">
     <span
         class="fd-switch"
         [ngClass]="{


libs/core/src/lib/switch/switch.component.html
```diff
@@ -1,4 +1,4 @@
-<label class="fd-switch__label" (click)="focus()">
+<label class="fd-switch__label" (click)="focus()" [ngStyle]="{ cursor: disabled ? 'not-allowed' : 'pointer'}">
```
JKMarkowski, line 1: I think this cursor should be auto

salarenko (changes requested):
We should just patch the styles here:
```
.fd-switch.is-disabled {
    cursor: not-allowed;
}
```

olshansky:
> We should just patch the styles here:
> 
> ```
> .fd-switch.is-disabled {
>     cursor: not-allowed;
> }
> ```

@salarenko unfortunately, it won't work
switch-disabled

libs/core/src/lib/switch/switch.component.html
olshansky: @JKMarkowski fixed

salarenko (approved):

droshev (approved):

PR #3803, opened 2020-11-10
feat(appshell): support module injection

valorkin:


libs/app-shell/src/lib/components/plugin-launcher/plugin-launcher.component.ts
```diff
@@ -98,31 +101,30 @@ export class PluginLauncherComponent implements OnChanges, AfterViewChecked {
...
+            const element = document.createElement(_module);
             this._render.appendChild(this._elementRef.nativeElement, element);
             return;
         }
 
-        if (_module.type === 'angular-ivy-component' && this.ngContentView) {
-            this.ngContentView.clear();
-            const factory = this.cfr.resolveComponentFactory(_component);
-            this.ngContentView.createComponent(factory, null, this._injector);
+        if (_pluginModule.type === 'angular-ivy-component') {
```
fkolar, line 125: Does this changes also completly functionality that we discuss in the other thread? 
That we dont support exposing components anymore and we go fully to Modules only?

fkolar (changes requested):
Looks good to me, except I need one explanation for my bellow comment.

libs/app-shell/src/lib/components/plugin-launcher/plugin-launcher.component.ts
valorkin: it may work without moduleFactory, just with component

fkolar (commented):
Good. We need this `compileAsync` but at the same time we dont want to break existing functionality, when application teams are using using it both with components and modules. . 

When you say its `it may work` do you need more time to test it ? Are there any breaking changes on this one ? 
I am ok with both options:
 - We can merge and release, so you can have time to test it with our demo app to see if existing exposed components works, and then comunicate the release
 - You can double check and spend more time on it, if there are any additional changes needed which we need to comunicate before we merge, and then we merge. 
     - e..g: updated plugins.json, etc.. 


Let me know please

valorkin:
Don't release yet, I will finish it and test locally.
The only thing I DON'T like right now is a fallbacks, I will do a commit to show what I mean so we could discuss

libs/app-shell/src/lib/components/plugin-launcher/plugin-launcher.component.ts
```diff
@@ -123,8 +126,18 @@ export class PluginLauncherComponent implements OnChanges, AfterViewChecked {
         }
 
         if (_pluginModule.type === 'angular-ivy-component') {
-            this._ngModule = await this.compiler.compileModuleAsync(_module);
-            this._ngComponent = _remoteModule[(_pluginModule as AngularIvyComponentDescriptor).component];
+            const ngPluginModule = _pluginModule as AngularIvyComponentDescriptor;
+            // if provided module name and component name we bootstrap both
+            if (ngPluginModule.name && ngPluginModule.component) {
```
valorkin, line 131: @fkolar this fallback is based on convention rather than data structure, but should work anyway.
I will test it with demo and will give a note when it's ready to merge

fkolar: great !

fkolar (approved):

Issue #3171, opened 2020-08-31
Upload Collection - implement the component in ngx-platform


manjunathanagaraj:

manjunathanagaraj:
This was estimated for 2 sprints. Therefore moving to 47.

manjunathanagaraj:
still in progress. Moving to sprint 48.

manjunathanagaraj:
Internal PR is raised. Moving to sprint 50.

PR #3888, opened 2020-11-18
feat: (Platform) Introduce Upload Collection

Betrozov:
https://github.com/SAP/fundamental-ngx/issues/3171
Introduce new Platform Upload Collection component

Documentation checklist:
- [X] Documentation Examples


libs/platform/src/lib/components/upload-collection/dialogs/move-to/move-to.component.ts
```diff
@@ -0,0 +1,125 @@
...
+@Component({
+    templateUrl: './move-to.component.html',
+    styles: [`
+        .fd-list__item--alternative-background {
+            background: #f2f2f2;
+        }
+
+        .fd-list__icon--alternative-background {
+            color: #74777A !important;
+        }
```
KevinOkamoto, line 21: Do we need to hardcode the colors here? Why do these need to be changed from the default colors?

libs/platform/src/lib/components/upload-collection/dialogs/move-to/move-to.component.ts
```diff
@@ -0,0 +1,125 @@
...
+            responsivePadding: true,
+            backdropClickCloseable: false,
+            height: '70%',
+            backdropClass: 'fdp-upload-collection-dialig-no-bg',
+            data: {
+                currentFolder: this._currentFolder
+            }
+        } as DialogConfig);
+
+        dialogRef.afterClosed.subscribe((folderName) => {
```
KevinOkamoto, line 85: Create a reference to this subscription and `unsubscribe` in OnDestroy lifecycle hook for this component.

KevinOkamoto (changes requested):
See comments for requested changes, but overall impressive work.

KevinOkamoto:
I noticed that when you remove files/folders consecutively, the message strips stack. Like this:

FIleUploaderMessageStrips

Is this the expected behavior? I was expecting that the new message strip would replace the old one. However this is probably something which can be debated.

Betrozov:
> I noticed that when you remove files/folders consecutively, the message strips stack. Like this:
> 
> FIleUploaderMessageStrips
> 
> Is this the expected behavior? I was expecting that the new message strip would replace the old one. However this is probably something which can be debated.

@KevinOkamoto 
yes, it is expected behavior.
but would be better if the Message Stripe has property like "duration" to display a message
and yes we can make only one Message Stripe and the new one will be a rewrite previous

apps/docs/src/app/platform/component-docs/upload-collection/examples/upload-collection-example.component.html
```diff
@@ -0,0 +1,11 @@
+<fdp-upload-collection
+    [dataSource]="dataSource"
+    maxFileSize="1mb"
+    maxFilenameLength="10"
+    [mimeTypes]="mimeTypes"
+    [fileTypes]="fileTypes"
+    [itemsPerPage]="itemsPerPage"
+    (typeMissmatch)="typeMissmatch($event)"
```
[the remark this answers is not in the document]
Betrozov: fixed

apps/docs/src/app/platform/component-docs/upload-collection/examples/upload-collection-example.component.ts
```diff
@@ -0,0 +1,52 @@
+import { Component } from '@angular/core';
+
+import {
+    ItemsPerPage,
+    FilenameLengthExceedEvent,
+    FileSizeExceedEvent,
+    TypeMissmatchEvent,
```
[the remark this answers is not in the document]
Betrozov: fixed

libs/platform/src/lib/components/upload-collection/dialogs/move-to/move-to.component.ts
Betrozov: fixed

libs/platform/src/lib/components/upload-collection/dialogs/move-to/move-to.component.ts
Betrozov: fixed

libs/platform/src/lib/components/upload-collection/upload-collection/upload-collection.component.ts
```diff
@@ -0,0 +1,974 @@
...
+         */
+        this._dsSubscription = initDataSource.open().subscribe((data) => {
+            this._visibleList = data;
+
+            this._totalItems = initDataSource.dataProvider.totalItems;
+
+            this._countUnvisibleItems();
+        });
+
+        // initial data fetch
```
mykolavorobiov, line 946: Please move this call `_match` outside, after call `_openDataStream`.

libs/platform/src/lib/components/upload-collection/upload-collection/upload-collection.component.ts
```diff
@@ -0,0 +1,974 @@
...
+    _dragDropCancel(): void {
+        this._showDragDropArea = false;
+    }
+
+    /** @hidden */
+    _breadcrumbGoTo(event: MouseEvent, index?: number): void {
+        event.preventDefault();
+        event.stopPropagation();
+
+        if (typeof index === 'undefined') {
```
mykolavorobiov, line 584: I think that it would be better to use `isNaN` instead `undefined`

libs/platform/src/lib/components/upload-collection/dialogs/move-to/move-to.component.ts
```diff
@@ -0,0 +1,124 @@
+import { Component, Inject, OnInit } from '@angular/core';
+
+import { DIALOG_REF, DialogRef, DialogService, DialogConfig } from '@fundamental-ngx/core';
+import { take } from 'rxjs/operators';
+
+import { UploadCollectionItem, UploadCollectionFolder } from '../../models/upload-collection.models';
+import { NewFolderComponent } from '../new-folder';
+
+function isObject<T>(item: T): boolean {
+    return typeof item === 'object' && !Array.isArray(item) && item !== null;
```
mykolavorobiov, line 10: What's happen with `undefined`?
What do you think about this?
`return !!item && typeof item === 'object';`
or
`return !!item && item.contructor === ({}).constructor;`

libs/platform/src/lib/components/upload-collection/dialogs/move-to/move-to.component.ts
```diff
@@ -0,0 +1,124 @@
...
+                currentFolder: this._currentFolder
+            }
+        } as DialogConfig);
+
+        dialogRef.afterClosed.pipe(take(1)).subscribe((folderName) => {
+            const folder: UploadCollectionFolder = {
+                documentId: null,
+                type: 'folder',
+                name: folderName,
+                uploadedBy: {
```
mykolavorobiov, line 89: I think that it would be better allow to control this params

mykolavorobiov (changes requested):

libs/platform/src/lib/components/upload-collection/upload-collection/upload-collection.component.ts
```diff
@@ -0,0 +1,974 @@
...
+    _getItemExtension(item: UploadCollectionItem): string {
+        if (item.type === 'folder') {
+            return '';
+        }
+
+        return `.${item.name.split('.').pop()}`;
+    }
+
+    /** @hidden */
+    _getIconByType(file: UploadCollectionFile): string {
```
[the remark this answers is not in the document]
Betrozov: [here](https://github.com/SAP/fundamental-ngx/pull/3888/files#diff-8636bc66f0ad6c574222c0f6578f8ba3b9d18e53af01072844832a7c6caf78c5R155)

libs/platform/src/lib/components/upload-collection/upload-collection/upload-collection.component.ts
```diff
@@ -0,0 +1,974 @@
...
+    /** @hidden */
+    @ViewChild('fileInput')
+    private readonly _uploadFilesRef: ElementRef;
+
+    /** @hidden */
+    @ViewChild('updateVersionInput')
+    private readonly _updateVersionRef: ElementRef;
+
+    /** @hidden */
+    private readonly _imageFileTypes: string[] = [
```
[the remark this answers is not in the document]
Betrozov: I'll rename it on `_imageExtensions` because it contains only image extensions

libs/platform/src/lib/components/upload-collection/upload-collection/upload-collection.component.ts
```diff
@@ -0,0 +1,974 @@
...
+
+            if (currentItem.type !== 'folder') {
+                continue;
+            }
+
+            if (currentItem.documentId === parentFolderId) {
+                const index = currentItem.files.findIndex((item) => item.documentId === updatedItem.documentId);
+                if (index !== -1) {
+                    currentItem.files[i] = updatedItem;
+                    this._match();
```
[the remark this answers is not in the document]
Betrozov: it will be called only once, when the condition is triggered then `for` will break

libs/platform/src/lib/components/upload-collection/dialogs/move-to/move-to.component.ts
Betrozov: if we call `isObject(undefined)` resault will be `false` as expected
the function `isObject` must define only object (`{}`), not an array and other

libs/platform/src/lib/components/upload-collection/dialogs/move-to/move-to.component.ts
Betrozov: it will be updated immediately after we get a response from the server
it is needs only when you create a new folder
also, this behavior was taken from this [sample](https://sapui5.hana.ondemand.com/#/entity/sap.m.UploadCollection/sample/sap.m.sample.UploadCollection)

KevinOkamoto:
Build seems to be failing. I'll review again once build passes.

mykolavorobiov (approved):

KevinOkamoto (approved):

JKMarkowski:
Could you rebase to newest master? I see that cdk styles are missing in popover components
image

Betrozov:
> Could you rebase to newest master? I see that cdk styles are missing in popover components
> image

this [PR](https://github.com/SAP/fundamental-ngx/pull/3985) will fix this issue
FYI: @JKMarkowski

libs/platform/src/lib/components/upload-collection/dialogs/move-to/move-to.component.ts
```diff
@@ -0,0 +1,125 @@
...
+function isObject<T>(item: T): boolean {
+    return typeof item === 'object' && !Array.isArray(item) && item !== null;
+}
+
+@Component({
+    templateUrl: './move-to.component.html',
+    styles: [
+        `
+            .fd-list__item--title-background {
+                background: #f2f2f2;
```
InnaAtanasova, line 18: This is not themable. If you change the theme, the hard-coded color will stay the same.

libs/platform/src/lib/components/upload-collection/dialogs/new-folder/new-folder.component.ts
```diff
@@ -0,0 +1,32 @@
...
+            this._setSelectionRange(el, 0, this.newFolderName.length);
+        }
+    }
+
+    private _setSelectionRange(el: HTMLInputElement, selectionStart: number, selectionEnd: number): void {
+        const direction = el.selectionDirection;
+
+        el.setSelectionRange(selectionStart, selectionEnd, direction);
+    }
+}
```
InnaAtanasova, line 32: This entire file is missing typedocs

libs/platform/src/lib/components/upload-collection/dialogs/move-to/move-to.component.ts
```diff
@@ -0,0 +1,125 @@
...
+            if (nestedValue.files.some((item) => item.documentId === documentId)) {
+                foundObj = nestedValue;
+            }
+
+            return nestedValue;
+        });
+
+        return foundObj;
+    }
+}
```
InnaAtanasova, line 125: The entire file is missing typedocs

libs/core/src/lib/file-uploader/file-uploader.service.ts
```diff
@@ -9,6 +14,36 @@ export interface FileUploadOutput {
 
 @Injectable({ providedIn: 'root' })
 export class FileUploaderService {
+    static _parseFileSize(fileSize: string): number {
```
JKMarkowski, line 17: Could you move this method to utils? I see it's used in upload collection, maybe it will be handy in other component as well

libs/platform/src/lib/components/upload-collection/dialogs/move-to/move-to.component.ts
JKMarkowski: Maybe let's put this method somewhere into utils?

libs/platform/src/lib/components/upload-collection/pipes/convert-bytes.pipe.ts
```diff
@@ -0,0 +1,27 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({ name: 'convertBytes' })
```
JKMarkowski, line 3: It would be great to have it as a util's pipe as well

libs/platform/src/lib/components/upload-collection/upload-collection/upload-collection.component.scss
```diff
@@ -0,0 +1,203 @@
+@mixin fd-reset {
+    font-size: var(--sapFontSize);
+    line-height: var(--sapContent_LineHeight);
+    color: var(--sapTextColor);
+    font-family: var(--sapFontFamily);
```
JKMarkowski, line 5: Great job using variables there, Could you add the declaration with value, before each? So
from:
`font-size: var(--sapFontSize);` 
To:
```
font-size: 0.75rem;
font-size: var(--sapFontSize);
```
Otherwise IE11 won't know what's going on

libs/platform/src/lib/domain/index.ts
```diff
@@ -1,5 +1,8 @@
+import { from } from 'rxjs';
```
JKMarkowski, line 1: Is this import necessary?

JKMarkowski (dismissed):
Hi @Betrozov, It works well for me, Great Job, this component looks huge.I just added some comments regarding the code.

libs/platform/src/lib/components/upload-collection/directives/upload-collection-dragndrop.directive.ts
```diff
@@ -0,0 +1,154 @@
...
+    @Input()
+    mimeTypes: string[] = [];
+
+    /** Whether selecting of new files is disabled. */
+    @Input()
+    disabled = false;
+
+    /** Whether drag and drop is enabled. Disables this directive. */
+    @Input()
+    dragndrop = true;
```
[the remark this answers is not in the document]
Betrozov: everywhere we use this convention for `dragndrop`

libs/platform/src/lib/components/upload-collection/upload-collection/upload-collection.component.scss
Betrozov: Hi @JKMarkowski, thank you for the comments.

what about if it is a color variable?
from which theme should I use color?

InnaAtanasova (approved):
Great job! Thanks for addressing my comments.

JKMarkowski (approved):
LGTM

Issue #3786, opened 2020-11-09
Fundamental-ngx 0.25.0 release (+Defect Hunting)


InnaAtanasova:
All the activities related to the release of 0.25.0 version of Fundamental-ngx.
Please list of critical issues as comments below.

InnaAtanasova:
- [ ] Action Sheet - clicking on the Action Sheet trigger focusses the first element. In the documentation we should provide an example where clicking on any of the list item triggers an event (shows alert, message box, etc to the user) and closes the menu
- [ ] Inline Help - we need to remove Inline Help before the release as part of 0.13.0 adoption. If we keep it, we need to fix this:
<img width="1240" alt="Screen Shot 2020-11-17 at 12 39 16 PM" src="https://user-images.githubusercontent.com/39598672/99426383-01995000-28d2-11eb-98ac-7db1b59f718c.png">
<img width="1275" alt="Screen Shot 2020-11-17 at 12 39 10 PM" src="https://user-images.githubusercontent.com/39598672/99426390-03631380-28d2-11eb-9513-e1b8c322cb61.png">

- [ ] Popover
<img width="1105" alt="Screen Shot 2020-11-17 at 12 41 29 PM" src="https://user-images.githubusercontent.com/39598672/99426564-41603780-28d2-11eb-8ff5-dc07646aaa77.png">
<img width="1063" alt="Screen Shot 2020-11-17 at 12 41 18 PM" src="https://user-images.githubusercontent.com/39598672/99426571-42916480-28d2-11eb-99dc-f1e6653e9ebf.png">

- [ ] Select Semantic States - the size of the messages is not correct. The size fits the content until it reaches its maximum and then the texts goes to second line
<img width="992" alt="Screen Shot 2020-11-17 at 12 43 01 PM" src="https://user-images.githubusercontent.com/39598672/99426782-969c4900-28d2-11eb-8df9-9b78669cf88e.png">
<img width="1083" alt="Screen Shot 2020-11-17 at 12 42 54 PM" src="https://user-images.githubusercontent.com/39598672/99426788-98660c80-28d2-11eb-9222-218e9f104e0b.png">

- [ ] Toolbar - Button in the overflow menu takes the width of the menu
<img width="874" alt="Screen Shot 2020-11-17 at 12 51 59 PM" src="https://user-images.githubusercontent.com/39598672/99427987-14148900-28d4-11eb-8f3e-8ee544daa1b1.png">

- [ ] ShellBar issues
[...]

PR #3899, opened 2020-11-19
fix: (Core) Add icons to Action Tile buttons

JKMarkowski:
part of: #3786

There are added icons rendering on action tile buttons.

### DIFF
--- a/libs/core/src/lib/tile/directives/tile.directives.ts
+++ b/libs/core/src/lib/tile/directives/tile.directives.ts
@@ -299,6 +299,7 @@ export class TileActionCloseDirective implements OnInit, OnChanges, CssClassBuil
     /** @hidden */
     ngOnInit(): void {
         this.buildComponentCssClass();
+        this._addCloseIcon();
     }
 
     @applyCssClass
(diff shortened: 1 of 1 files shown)


libs/core/src/lib/tile/directives/tile.directives.ts
```diff
@@ -314,6 +315,14 @@ export class TileActionCloseDirective implements OnInit, OnChanges, CssClassBuil
     elementRef(): ElementRef<any> {
         return this._elementRef;
     }
+
+    /** @hidden */
+    private _addCloseIcon(): void {
+        const element = document.createElement('i');
```
Betrozov, line 321: @JKMarkowski What if we create a function that will be generated an icon element and will be used in two places?
something like this
```typescript
private _addCloseIcon(): void {
    const element = generateIcon('decline');
    this._elementRef.nativeElement.appendChild(element);
}

function generateIcon(iconName: string): HTMLElement {
    const element = document.createElement('i');
    element.setAttribute('role', 'presentation');
    element.classList.add(`sap-icon--${iconName}`);
    
    return element;
}
```

Betrozov (approved):

InnaAtanasova (approved):

Issue #3874, opened 2020-11-17
Implement Flexible column layout


droshev:
#### Is this a bug, enhancement, or feature request?
feature request

#### Briefly describe your proposal.
https://experience.sap.com/fiori-design-web/flexible-column-layout/

8d

Preserving the state of the columns is out of scope

PR #3934, opened 2020-11-26
feat: (Core) introduce Flexible Column Layout component

InnaAtanasova:
closes https://github.com/SAP/fundamental-ngx/issues/3874

Introduces new Flexible Column Layout component. 
Adds "karma-viewport" so I can test the layouts in desktop and tablet.

**Note 1**: Please visit the [Fiori3 Guidelines](https://experience.sap.com/fiori-design-web/flexible-column-layout/#possible-layouts) to check the behaviour of the component, its layouts and responsive design. Also check the description provided in the documentation. 

**Note 2**: the css code will be removed once Fundamental-styles is released. 

Documentation checklist:
- [NA] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/apps/docs/src/app/core/component-docs/flexible-column-layout/examples/default/flexible-column-layout-example.component.ts
+++ b/apps/docs/src/app/core/component-docs/flexible-column-layout/examples/default/flexible-column-layout-example.component.ts
@@ -19,8 +19,6 @@ export class FlexibleColumnLayoutExampleComponent {
      */
     fullscreen = false;
 
-    constructor() {}
-
     /**
      * documentation related property
      * sets the initial layout of the component to 'OneColumnStartFullScreen'
@@ -40,7 +38,6 @@ export class FlexibleColumnLayoutExampleComponent {
     changeLayout(newValue: string): void {
         this.localLayout = newValue;
     }
-
 
     /**
      * documentation related function
--- a/apps/docs/src/app/core/component-docs/flexible-column-layout/examples/default/flexible-column-layout-example.component.ts
+++ b/apps/docs/src/app/core/component-docs/flexible-column-layout/examples/default/flexible-column-layout-example.component.ts
@@ -1,0 +1,63 @@
+import { Component, ElementRef, ViewChild } from '@angular/core';
+
+@Component({
+    selector: 'fd-flexible-column-layout-example',
+    templateUrl: './flexible-column-layout-example.component.html',
+    styleUrls: ['flexible-column-layout-example.component.scss']
+})
+export class FlexibleColumnLayoutExampleComponent {
+    /**
+     * documentation related property
+     * provides access to the HTML element with "overlay" reference
+     */
+    @ViewChild('overlay')
+    overlay: ElementRef<HTMLElement>;
+
+    /**
+     * documentation related property
+     * specifies if the doc example is rendered in fullscreen or not
+     */
+    fullscreen = false;
+
+    constructor() {}
+
+    /**
+     * documentation related property
+     * sets the initial layout of the component to 'OneColumnStartFullScreen'
+     * sets a new layout for the component 
+     */
+    localLayout = 'OneColumnStartFullScreen';
+
+    /**
+     * this function is reacting to events (button clicks) and 
+     * updates the local property which sets a new layout for the component.
+     * Available values for the layouts include:
+     * 'OneColumnStartFullScreen' | 'OneColumnMidFullScreen' | 'OneColumnEndFullScreen' |
+     * 'TwoColumnsStartExpanded' | 'TwoColumnsMidExpanded' | 'TwoColumnsEndExpanded' | 
+     * 'ThreeColumnsMidExpanded' | 'ThreeColumnsEndExpanded' | 'ThreeColumnsStartMinimized' | 
+     * 'ThreeColumnsEndMinimized';
+     */
+    changeLayout(newValue: string): void {
+        this.localLayout = newValue;
+    }
+
+
+    /**
+     * documentation related function
+     * opens the example in full screen
+     */
+    enterFullscreenExample(): void {
+        this.fullscreen = true;
+        this.overlay.nativeElement.style.width = '100%';
+    }
+    
+    /**
+     * documentation related function
+     * exits the full screen mode of the example
+     */
+    exitFullscreenExample(event: Event): void {
+        event.stopPropagation();
+        this.fullscreen = false;
+        this.overlay.nativeElement.style.width = '0%';
+    }
+}
--- a/libs/core/src/lib/flexible-column-layout/constants.ts
+++ b/libs/core/src/lib/flexible-column-layout/constants.ts
@@ -1,0 +1,23 @@
+  
+export const ONE_COLUMN_START_FULL_SCREEN = 'OneColumnStartFullScreen';
+export const ONE_COLUMN_MID_FULL_SCREEN = 'OneColumnMidFullScreen';
+export const ONE_COLUMN_END_FULL_SCREEN = 'OneColumnEndFullScreen';
+export const TWO_COLUMNS_START_EXPANDED = 'TwoColumnsStartExpanded';
+export const TWO_COLUMNS_MID_EXPANDED = 'TwoColumnsMidExpanded';
+export const TWO_COLUMNS_END_EXPANDED = 'TwoColumnsEndExpanded';
+export const THREE_COLUMNS_MID_EXPANDED = 'ThreeColumnsMidExpanded';
+export const THREE_COLUMNS_END_EXPANDED = 'ThreeColumnsEndExpanded';
+export const THREE_COLUMNS_START_MINIMIZED = 'ThreeColumnsStartMinimized';
+export const THREE_COLUMNS_END_MINIMIZED = 'ThreeColumnsEndMinimized';
+
+export const SM_SCREEN_SIZE = 'sm';
+export const MD_SCREEN_SIZE = 'md';
+export const LG_SCREEN_SIZE = 'lg';
+export const XL_SCREEN_SIZE = 'xl';
+
+export type ScreenSize = 'sm' | 'md' | 'lg' | 'xl';
+
+export type Layout = 'OneColumnStartFullScreen' | 
+'OneColumnMidFullScreen' | 'OneColumnEndFullScreen' | 'TwoColumnsStartExpanded' | 'TwoColumnsMidExpanded' | 'TwoColumnsEndExpanded' | 'ThreeColumnsMidExpanded' | 'ThreeColumnsEndExpanded' | 'ThreeColumnsStartMinimized' | 'ThreeColumnsEndMinimized';
+
+export type ColumnSeparatorValue = 'right' | 'left' | null;
--- a/libs/core/src/lib/flexible-column-layout/constants.ts
+++ b/libs/core/src/lib/flexible-column-layout/constants.ts
@@ -13,11 +13,12 @@ export const THREE_COLUMNS_END_MINIMIZED = 'ThreeColumnsEndMinimized';
 export const SM_SCREEN_SIZE = 'sm';
 export const MD_SCREEN_SIZE = 'md';
 export const LG_SCREEN_SIZE = 'lg';
-export const XL_SCREEN_SIZE = 'xl';
 
-export type ScreenSize = 'sm' | 'md' | 'lg' | 'xl';
+export type ScreenSize = 'sm' | 'md' | 'lg';
 
-export type Layout = 'OneColumnStartFullScreen' | 
+export type FlexibleColumnLayout = 'OneColumnStartFullScreen' | 
 'OneColumnMidFullScreen' | 'OneColumnEndFullScreen' | 'TwoColumnsStartExpanded' | 'TwoColumnsMidExpanded' | 'TwoColumnsEndExpanded' | 'ThreeColumnsMidExpanded' | 'ThreeColumnsEndExpanded' | 'ThreeColumnsStartMinimized' | 'ThreeColumnsEndMinimized';
 
 export type ColumnSeparatorValue = 'right' | 'left' | null;
+
+export interface FlexibleColumnSettings { start: number; mid: number; end: number; }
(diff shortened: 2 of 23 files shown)


apps/docs/src/app/core/component-docs/flexible-column-layout/examples/default/flexible-column-layout-example.component.ts
```diff
@@ -0,0 +1,63 @@
...
+    @ViewChild('overlay')
+    overlay: ElementRef<HTMLElement>;
+
+    /**
+     * documentation related property
+     * specifies if the doc example is rendered in fullscreen or not
+     */
+    fullscreen = false;
+
+    constructor() {}
```
Betrozov, line 22: please remove empty block

libs/core/src/lib/flexible-column-layout/constants.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const SM_SCREEN_SIZE = 'sm';
+export const MD_SCREEN_SIZE = 'md';
+export const LG_SCREEN_SIZE = 'lg';
+export const XL_SCREEN_SIZE = 'xl';
+
+export type ScreenSize = 'sm' | 'md' | 'lg' | 'xl';
+
+export type Layout = 'OneColumnStartFullScreen' | 
+'OneColumnMidFullScreen' | 'OneColumnEndFullScreen' | 'TwoColumnsStartExpanded' | 'TwoColumnsMidExpanded' | 'TwoColumnsEndExpanded' | 'ThreeColumnsMidExpanded' | 'ThreeColumnsEndExpanded' | 'ThreeColumnsStartMinimized' | 'ThreeColumnsEndMinimized';
```
Betrozov, line 21: What if we use `enum` instead of `type Layout` and `const`s that created above?

libs/core/src/lib/flexible-column-layout/constants.ts
```diff
@@ -0,0 +1,23 @@
...
+export const THREE_COLUMNS_END_EXPANDED = 'ThreeColumnsEndExpanded';
+export const THREE_COLUMNS_START_MINIMIZED = 'ThreeColumnsStartMinimized';
+export const THREE_COLUMNS_END_MINIMIZED = 'ThreeColumnsEndMinimized';
+
+export const SM_SCREEN_SIZE = 'sm';
+export const MD_SCREEN_SIZE = 'md';
+export const LG_SCREEN_SIZE = 'lg';
+export const XL_SCREEN_SIZE = 'xl';
+
+export type ScreenSize = 'sm' | 'md' | 'lg' | 'xl';
```
Betrozov, line 18: the same

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.html
```diff
@@ -0,0 +1,50 @@
+<div 
+    class="fd-flexible-column-layout"
+    [ngClass]="{
+        'fd-flexible-column-layout--translucent': backgroundDesign==='translucent', 
+        'fd-flexible-column-layout--transparent': backgroundDesign==='transparent'
```
Betrozov, line 5: please use like this `[class.fd-flexible-column-layout--translucent]="backgroundDesign==='translucent'"`
it will be faster

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.ts
```diff
@@ -0,0 +1,421 @@
...
+      }
+
+      if (this._screenSize === MD_SCREEN_SIZE) {
+          this._updateCurrentLayout(THREE_COLUMNS_START_MINIMIZED);
+      }
+      break;
+      }
+
+      case THREE_COLUMNS_END_EXPANDED : {
+      if (this._screenSize === SM_SCREEN_SIZE) {
```
Betrozov, line 294: please reformat code

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.ts
```diff
@@ -0,0 +1,421 @@
...
+      case ONE_COLUMN_MID_FULL_SCREEN : return null;
+      case ONE_COLUMN_END_FULL_SCREEN : return null;
+      case TWO_COLUMNS_START_EXPANDED : return 'left';
+      case TWO_COLUMNS_MID_EXPANDED : return 'right';
+      case TWO_COLUMNS_END_EXPANDED : return null;
+      case THREE_COLUMNS_MID_EXPANDED : return 'right';
+      case THREE_COLUMNS_END_EXPANDED : return null;
+      case THREE_COLUMNS_END_MINIMIZED : return 'right';
+      case THREE_COLUMNS_START_MINIMIZED : return 'right';
+    }
```
Betrozov, line 341:
```suggestion
    switch (this.layout) {
      case TWO_COLUMNS_START_EXPANDED : return 'left';
      case TWO_COLUMNS_MID_EXPANDED :
      case THREE_COLUMNS_MID_EXPANDED :
      case THREE_COLUMNS_END_MINIMIZED :
      case THREE_COLUMNS_START_MINIMIZED : return 'right';
      default: return null;
    }
```

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.ts
```diff
@@ -0,0 +1,421 @@
...
+  }
+
+  /** 
+   * @hidden 
+   * determines if the right separator should be visible
+   * and the value that will specify the direction of the arrow
+   */
+  private _getRightColumnSeparatorValue(): ColumnSeparatorValue {
+    switch (this.layout) {
```
Betrozov, line 350: the same

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.ts
```diff
@@ -0,0 +1,421 @@
...
+   * updates the layout
+   * emits an event
+   * makes a call to the helper function that will update the column layout and the separators
+   */
+  private _updateCurrentLayout(newLayout: Layout): void {
+    this.layout = newLayout;
+    this._updateColumnLayoutParameters();
+    
+    // setTimeout fixes "ExpressionChangedAfterItHasBeenCheckedError"
+    setTimeout(() => {
```
Betrozov, line 387: `setTimeout` => `timer`?

Betrozov (changes requested):

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.ts
```diff
@@ -0,0 +1,421 @@
...
+   */
+  @Output()
+  layoutChange: EventEmitter<Layout> = new EventEmitter<Layout>();
+
+  /**
+   * user defined break point for MD screens
+   * the default (Fiori 3) value is 599
+   */
+  @Input()
+  mdBreakPoint = 599;
```
JKMarkowski, line 92: IMO it should follow breakpoints from documentation, so md starting from `961px`. 
The main reason why I want it to be brought is that every column have `320px` min-width. To display two columns in md in proper way, you need to have at least `960px`, to keep 67%/33% ratio

JKMarkowski: image

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.ts
```diff
@@ -0,0 +1,421 @@
...
+   */
+  @Input()
+  lgBreakPoint = 1023;
+
+  /**
+   * user defined break point for XL screens
+   * the default (Fiori 3) value is 1439
+   */
+  @Input()
+  xlBreakPoint = 1439;
```
JKMarkowski, line 106: Do you need `xlBreakPoint` there? I can see that on docs and in code it shares behaviour with `lgBreakPoint`, so IMO only lg shoudl be kept

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.ts
```diff
@@ -0,0 +1,421 @@
...
+    ThreeColumnsEndExpanded: { start: 25, mid: 25, end: 50 },
+    ThreeColumnsStartMinimized: { start: 0, mid: 67, end: 33 },
+    ThreeColumnsEndMinimized: { start: 33, mid: 67, end: 0 }
+  };
+
+  /**
+   * the column layout representing the distribution of the width
+   * between the first (start), the middle and the last(end) column
+   */
+  _columnLayout: { start: number; mid: number; end: number; } = this._layouts[this.layout];
```
JKMarkowski, line 164: I think for this object should be created some interface, eg.
`export interface FlexibleColumnSettings { start: number; mid: number; end: number; }`

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.ts
```diff
@@ -0,0 +1,421 @@
...
+  private _screenSize: ScreenSize = LG_SCREEN_SIZE;
+
+  /** @hidden */
+  private _subscriptions = new Subscription();
+
+  /** 
+   * @hidden 
+   * mappting of the layout name and the column layout in %
+   */
+  private _layouts = {
```
JKMarkowski, line 147: With `FlexibleColumnSettings` interface you can add type there: `_layouts: {[key: string]: FlexibleColumnSettings} = {`

libs/core/src/lib/flexible-column-layout/constants.ts
JKMarkowski: I would stick to `type`, but with changed named to `FlexibleColumnLayout`

JKMarkowski (changes requested):
Great job! It's bulletproof with already prepared `Layout` settings. I added some minor comments, let me know when they are addressed, so I will re-review

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.ts
InnaAtanasova: Don't understand the comment. `setTimeout` is used without a problem in the library.

Betrozov: I'm about to use `rxjs.timer`

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.ts
InnaAtanasova: I was using the break points specified in Fiori 3. Was not aware this component is using different break points. Changed to use the ones from the wiki for Flexible Column Layout

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.ts
InnaAtanasova: I kept the `setTimeout` as this is the current approach of the library. Don't mind to change it to timer if everyone is ok with it.

libs/core/src/lib/flexible-column-layout/flexible-column-layout.component.ts
InnaAtanasova: removed. The idea was to have it in case the user passed a different break point for XL

Betrozov (approved):

stefanoScalzo (approved):
LGTM

mikerodonnell89:
Clicking the expand button between column 2 and 3 hides the expand button between columns 1 and 2 and all columns remain visible, but clicking the expand column between 1 and 2 hides column 3 entirely? Is this the desired functionality?

flexlayout

mikerodonnell89:
Mobile mode - cannot exit column 3

flexlayoutmobile

mikerodonnell89 (changes requested):
Couple bugs, see GIFs

InnaAtanasova:
> Chrome/osx - Clicking the expand button between column 2 and 3 hides the expand button between columns 1 and 2 and all columns remain visible, but clicking the expand column between 1 and 2 hides column 3 entirely? Is this the desired functionality?
> 
> flexlayout

This is the "Three columns, end expanded" design, the first separator is hidden as the first 2 columns can no longer be expanded. 
Screen Shot 2020-11-30 at 3 42 38 PM

InnaAtanasova:
> Mobile mode - cannot exit column 3
> 
> flexlayoutmobile

You can't exit it because in the documentation example I specify that when I exit full screen I want to render the 'ThreeColumnsMidExpanded' layout. 
```
<button class="fd-button" *ngIf="localLayout === 'OneColumnEndFullScreen'"
              (click)="changeLayout('ThreeColumnsMidExpanded')">Exit Full Screen Column 3</button>
```

but the layout "detects" the mobile mode and renders again the last column (as per UX guidelines), which is column 3.

mikerodonnell89 (approved):

package.json
```diff
@@ -103,6 +103,7 @@
     "highlight.js": "9.18.1",
     "intl": "1.2.5",
     "jasminewd2": "2.2.0",
+    "karma-viewport": "1.0.7",
```
JKMarkowski, line 106: Just qq, is it needed?

InnaAtanasova: Hey @JKMarkowski , as per Fiori 3 guidelines for Responsiveness, the layout should automatically change depending on the screen size. For example, a 'THREE_COLUMNS_MID_EXPANDED' should render 3 columns, start open, mid expanded, end open in a large screen. This layout is changed to 'THREE_COLUMNS_START_MINIMIZED' in MD screen. I need a way to control the screen size in order to test these 2 cases. One, where the screen is large and one, where I pass this layout but the screen is tablet size. "karma-viewport" allows me to do this by setting the size of the viewport while testing. In the test file you will see `viewport.set(1023, 900);`. That's where I specify in which viewport I am testing the layout. 
If you think I should remove these test cases OR you know a better way to do it, please let me know and I will remove "karma-viewport" package.

JKMarkowski: Ok @InnaAtanasova , Thank You for clarification :D We should keep it then. Probably we'll need to test such a ase on other components as well

JKMarkowski (approved):

PR #3962, opened 2020-11-30
fix: (Core) throw error on theme switcher via url, when router is not imported

JKMarkowski:
There is a need to handle lack of router module import, when theme switching via url is enabled


libs/core/src/lib/utils/consts/error-messages.ts
```diff
@@ -1 +1,3 @@
 export const MOBILE_CONFIG_ERROR = `Missing configuration object for mobile version of the component. Consult the documentation on how to provide MobileModeConfiguration object for this component.`;
+
+export const THEME_SWITCHER_ROUTER_MISSING_ERROR = `You need to import RouterModule, to enable setting theme via url.`
```
InnaAtanasova, line 3: Discussed with Jessie and it should be `You need to import RouterModule to enable theme settings via the url.`

InnaAtanasova (changes requested):

droshev:
@JKMarkowski do you think it is worth mentioning on the doc example too?

salarenko (approved):

katekozlowska (approved):
LGTM

InnaAtanasova (approved):

Issue #3146, opened 2020-08-30
Accessibility Bug: When the ‘button’ in the shell bar is accessed, reading is made as ‘Popover button Button’.


puru-hk:
Fundamental core component: "Product switch"

Description: When the ‘button’ in the shell bar is accessed, reading is made as ‘Popover button Button’.

Expected:  The extra ‘button’ announcement should be removed when the button from the shell bar region is accessed.

Attached screenshot:

![image](https://user-images.githubusercontent.com/32538291/91651554-c72ace80-eaab-11ea-9059-5dac7311b2a0.png)

Issue #3145, opened 2020-08-30
Accessibility Bug: There is no text alternative in the form of tooltip provided for the highlighted button in the shell bar region


puru-hk:
Fundamental core component: "Product switch". 

Description:  There is no text alternative in the form of tooltip provided for the highlighted button in the shell bar region

Expected: The text alternative should be provided for every non-text content element which fulfills the purpose of the content.

Attached screenshot:

![image](https://user-images.githubusercontent.com/32538291/91651531-7f0bac00-eaab-11ea-808e-3facdb89d9b9.png)

PR #3989, opened 2020-12-02
fix: (Core) added title to Product Switch button

olshansky:
and made the default button label text a more meaningful

Closes: https://github.com/SAP/fundamental-ngx/issues/3145
Closes: https://github.com/SAP/fundamental-ngx/issues/3146
- Added a `title` to the product switch button
- Default label text of product switch button changed to a more meaningful 

Documentation checklist:
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/product-switch/product-switch/product-switch.component.html
+++ b/libs/core/src/lib/product-switch/product-switch/product-switch.component.html
@@ -14,9 +14,10 @@
             <button
                 fd-button
                 [attr.aria-label]="ariaLabel"
+                [attr.title]="ariaLabel"
                 class="fd-shellbar__button fd-product-switch__control"
-                [glyph]="'grid'"
-                [fdType]="'transparent'"
+                glyph="grid"
+                fdType="transparent"
             ></button>
         </fd-popover-control>
         <fd-popover-body>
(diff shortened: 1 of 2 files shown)


libs/core/src/lib/product-switch/product-switch/product-switch.component.html
```diff
@@ -14,9 +14,10 @@
             <button
                 fd-button
                 [attr.aria-label]="ariaLabel"
+                [attr.title]="ariaLabel"
                 class="fd-shellbar__button fd-product-switch__control"
-                [glyph]="'grid'"
-                [fdType]="'transparent'"
+                glyph="grid"
```
fkolar, line 19: Maybe we should also add this to guideline too. So people are not mix expression that angular needs to evaluate this simple form. 

And also if. there will be always this, maybe we should think to use `@Attribute`  I see this all over the places .. good. !

fkolar (approved):

libs/core/src/lib/product-switch/product-switch/product-switch.component.html
olshansky: Yeah, would be great to update the most of guidelines, because not all of it in an actual state

libs/core/src/lib/product-switch/product-switch/product-switch.component.ts
```diff
@@ -1,21 +1,19 @@
-import { ChangeDetectionStrategy, Component, ViewEncapsulation, Input, HostBinding } from '@angular/core';
+import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
 import {Placement} from 'popper.js';
 import { BasePopoverClass } from '../../popover/base/base-popover.class';
 
 @Component({
     selector: 'fd-product-switch',
     templateUrl: './product-switch.component.html',
```
InnaAtanasova, line 7: Why are you removing the ViewEncapsulation? Our components have encapsulation set to `None`.

InnaAtanasova (approved):

Betrozov (approved):

Issue #3970, opened 2020-12-01
Datetime Picker: Button name is different in spec it is "OK" but in the component it is "Submit"


Lokanathan-k:
#### Is this a bug, enhancement, or feature request?
bug
#### Briefly describe your proposal.

button named wrongly 

Specs:
![image](https://user-images.githubusercontent.com/53534379/100715017-9d4fb380-33dc-11eb-978b-52e474b40d0e.png)

implementation:
![image](https://user-images.githubusercontent.com/53534379/100715133-c07a6300-33dc-11eb-959e-a4b828009d54.png)

kavya-b:
fixed with #3991

PR #3991, opened 2020-12-02
fix: (Core) replace datetime picker button label from 'Submit' to 'OK'

kavya-b:
#3970 
This PR replaced the 'Submit' button in datetime picker to the recommended 'OK' button as per visual spec.

Documentation checklist:
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/datetime-picker/datetime-picker.component.html
+++ b/libs/core/src/lib/datetime-picker/datetime-picker.component.html
@@ -99,7 +99,7 @@
                             <button
                                 fd-button
                                 fdType="emphasized"
-                                label="Submit"
+                                label="OK"
                                 [compact]="compact"
                                 (click)="submit()"
                             ></button>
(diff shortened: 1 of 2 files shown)


libs/core/src/lib/datetime-picker/datetime-picker.component.html
```diff
@@ -99,7 +99,7 @@
                             <button
                                 fd-button
                                 fdType="emphasized"
-                                label="Submit"
+                                label="OK"
```
dimamarksman, line 102: Would be great to add `i18` markers for OK and CANCEL buttons

dimamarksman (changes requested):
Could you please add i18 markers for these buttons

libs/core/src/lib/datetime-picker/datetime-picker.component.html
kavya-b: Not sure if this has changed but I don't think we usually place i18n markers for core components.

stefanoScalzo (approved):

libs/core/src/lib/datetime-picker/datetime-picker.component.html
dimamarksman: Yeah, that can be done by using introducing date-time-i18n service, but not in scope of this pr

dimamarksman (approved):

Issue #3038, opened 2020-08-18
Add new options to UI element 'Select' and set the new value directly as 'selected' does not work


rodema:
#### Is this a bug, enhancement, or feature request?
This is a Bug.

#### Briefly describe your proposal.
Adding new values (options) to UI element 'Select' and directly set this new value as "selcted value" via code, this does not work. 
The UI field which shows the selected value in the collapsed drop down list box stays empty.
If you expand the drop down list box, the new value is included in the value list and marked as selected

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Fundamental-For-Angular version: v 0.21.0

#### If this is a bug, please provide steps for reproducing it.
I have created a stackblitz example where options are added and selected values are set with a delay via timeout function.
You can find the example here:
https://stackblitz.com/edit/angular-hczejb?file=src/app/select-mode-example.component.ts

This is a description what happens (adding options and setting selected value) in which second.
Start with option : ['Apple', 'Pineapple', 'Tomato', 'Strawberry'];

2s : Apple is selected. OK
3s : Banana is added as an option. OK
4s : Banana is selected. OK
5s : Chips is added as option and selected at the same time. not OK. 
     Selected field of drop down list box stays empty 
7s : Chips is selected. not OK.
     Selected field of drop down list box stays empty 
9s : Tomato is selected. OK  
11s: Chips is selected. now OK.	 
     The only way to select Chips was to select a different value (Tomato) inbetween
13s: Pretzel is selected which is currently not in the option list. 
     The "old" selected value (Chips) is still be displayed what is OK because Pretzel is not in the option list
14s: Pretzel is added as option. not OK.
     Selected field of drop down list box is cleared (Chips is wiped out)
15s: Pretzel is selected. not OK.
     Selected field of drop down list box stays empty 	 

#### Please provide relevant source code if applicable.

https://stackblitz.com/edit/angular-hczejb?file=src/app/select-mode-example.component.ts

PR #4006, opened 2020-12-03
fix: (Core) adding and setting new options dynamically in 'Select'

katekozlowska:
closes https://github.com/SAP/fundamental-ngx/issues/3038

In a situation of race condition caused by setTimeout user could not add an option and select it immediately (when this option was added the selected.viewValueText was not refreshed).


Betrozov (approved):

dimamarksman:
It would be great to cover this certain case by unit test as well in order to prevent this issue in the future

salarenko (approved):

Issue #4036, opened 2020-12-08
Carousel RTL: Location of Arrow Buttons (over content) are not correct and not aligned with Visual Core. 


pierrom:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.
Carousel RTL: Location of Arrow Buttons (over content, within the picture) are not correct and not aligned with Visual Core. 
Arrow buttons should physically switch location. Currently when RTL is enabled **outward pointing** button simply points **inward**. 
![image](https://user-images.githubusercontent.com/71797052/101437403-fc259900-38c4-11eb-9103-6b1a58ff5e9f.png)

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
v0.25.0

#### If this is a bug, please provide steps for reproducing it.
1. Go to https://fundamental-ngx.netlify.app/#/core/carousel
2. Scroll down to: Carousel with navigation inside content area
3. Click RTL switch and see error

Issue #4035, opened 2020-12-08
Carousel RTL: Location of Arrow Buttons are not correct and not aligned with Visual Core.


pierrom:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.
Carousel RTL: Location of Arrow Buttons are not correct and should switch position just like the dots (page indicator)
![image](https://user-images.githubusercontent.com/71797052/101436675-7ead5900-38c3-11eb-9f6b-1fc62f218d9b.png)

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
v0.25.0

#### If this is a bug, please provide steps for reproducing it.
1. Go to: https://fundamental-ngx.netlify.app/#/core/carousel
2. Click RTL switch on any carousel component
3. See error

PR #4080, opened 2020-12-10
fix: (Core) Carousel direction defects

olshansky:
Closes: https://github.com/SAP/fundamental-ngx/issues/4035
Closes: https://github.com/SAP/fundamental-ngx/issues/4036

Documentation checklist:
- [n/a] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/carousel/carousel.component.scss
+++ b/libs/core/src/lib/carousel/carousel.component.scss
@@ -18,3 +18,30 @@ fd-carousel:focus {
 [dir=rtl] .fd-carousel__button.fd-button > [class*=sap-icon] {
     transform: rotate(0deg);
 }
+
+.fd-carousel__content > .fd-carousel__button.fd-button {
+    &.is-focus,
+    &:focus {
+        top: 50%;
+    }
+
+    [dir=ltr] &.fd-carousel__button--left.is-focus,
+    [dir=ltr] &.fd-carousel__button--left:focus {
+        right: auto;
+    }
+
+    [dir=ltr] &.fd-carousel__button--right.is-focus,
+    [dir=ltr] &.fd-carousel__button--right:focus {
+        left: auto;
+    }
+
+    [dir=rtr] &.fd-carousel__button--left.is-focus,
+    [dir=rtr] &.fd-carousel__button--left:focus {
+        left: auto;
+    }
+
+    [dir=rtr] &.fd-carousel__button--right.is-focus,
+    [dir=rtr] &.fd-carousel__button--right:focus {
+        right: auto;
+    }
+}
--- a/libs/core/src/lib/carousel/carousel.component.ts
+++ b/libs/core/src/lib/carousel/carousel.component.ts
@@ -308,14 +308,14 @@ export class CarouselComponent implements OnInit, AfterContentInit, AfterViewIni
     @HostListener('keydown.arrowright', ['$event'])
     onKeydownArrowRight(event: KeyboardEvent): void {
         event.preventDefault();
-        this.next();
+        this.dir === 'rtl' ? this.previous() : this.next();
     }
 
     /** @hidden */
     @HostListener('keydown.arrowleft', ['$event'])
     onKeydownArrowLeft(event: KeyboardEvent): void {
         event.preventDefault();
-        this.previous();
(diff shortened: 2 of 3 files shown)


libs/core/src/lib/carousel/carousel.component.scss
```diff
@@ -18,3 +18,9 @@ fd-carousel:focus {
 [dir=rtl] .fd-carousel__button.fd-button > [class*=sap-icon] {
     transform: rotate(0deg);
 }
+
+.fd-carousel__button.fd-button.is-focus,
+.fd-carousel__button.fd-button:focus {
+    top: 50%;
+    left: initial;
+}
```
KevinOkamoto, line 26: I think this CSS rule is causing the carousel button to move on focus. This is what I'm seeing when I keyboard navigate to the carousel button:
Screen Shot 2020-12-10 at 8 06 44 AM

KevinOkamoto (changes requested):
See comment.

DeepakSap14:
Buttons are still buggy on chrome browser.

olshansky:
> Buttons are still buggy on chrome browser when placed inside page-indicator.

@DeepakSap14 could you add more details and gifs with examples of issues? Because in my browser all work fine (Chrome 86.0.4240.198)

DeepakSap14:
> > Buttons are still buggy on chrome browser when placed inside page-indicator.
> 
> @DeepakSap14 could you add more details and gifs with examples of issues? Because in my browser all work fine (Chrome 86.0.4240.198)

Yes, it is not reproducible in chrome as well.
May be last time, i went to old deploy link, before new deployment link was available.

DeepakSap14 (approved):

Betrozov (approved):

InnaAtanasova (changes requested):
There's an RTL service implemented in Core. Can you use it?

olshansky:
> There's an RTL service implemented in Core. Can you use it?

@InnaAtanasova could you clarify what do you mean? 
Carousel component already uses RTL service ([injected here](https://github.com/SAP/fundamental-ngx/blob/d215f587485965e588e9cb473f96e14d7974e905/libs/core/src/lib/carousel/carousel.component.ts#L221) and uses below)

libs/core/src/lib/carousel/carousel.component.ts
```diff
@@ -308,14 +308,14 @@ export class CarouselComponent implements OnInit, AfterContentInit, AfterViewIni
     @HostListener('keydown.arrowright', ['$event'])
     onKeydownArrowRight(event: KeyboardEvent): void {
         event.preventDefault();
-        this.next();
+        this.dir === 'rtl' ? this.previous() : this.next();
```
stefanoScalzo, line 311: This should be using the rtl service from core. We have a service that unifies all components to be RTL

JKMarkowski (commented):
@olshansky 
Looks good to me, what about creating some `isRTL()` method, that will return boolean based on value from RtlService (it is behaviour subject storing value)?

libs/core/src/lib/carousel/carousel.component.ts
olshansky: @stefanoScalzo @InnaAtanasova as I understand you pointed this service – https://github.com/SAP/fundamental-ngx/blob/master/libs/core/src/lib/utils/services/carousel.service.ts
`this.previous()` and `this.next()` methods using this service ([in next](https://github.com/SAP/fundamental-ngx/blob/5de5d4c890d2e53e7f74b4f4682a1cc32e56c1bd/libs/core/src/lib/carousel/carousel.component.ts#L356), [in previous](https://github.com/SAP/fundamental-ngx/blob/5de5d4c890d2e53e7f74b4f4682a1cc32e56c1bd/libs/core/src/lib/carousel/carousel.component.ts#L329)), but it works not in a proper way for RTL direction, and my changes are the fastest solution for this issue

Should I try to fix carousel.service.ts?

olshansky:
> @olshansky
> Looks good to me, what about creating some `isRTL()` method, that will return boolean based on value from RtlService (it is behaviour subject storing value)?

@JKMarkowski added

InnaAtanasova (approved):

KevinOkamoto (approved):
LGTM

PR #4117, opened 2020-12-16
feat: (Platform) Introduce Value Help Dialog

mykolavorobiov:

[#4073](https://github.com/SAP/fundamental-ngx/issues/4073)

The value help dialog is a UI pattern that helps the user find and select single and multiple values. 

Documentation checklist:
- [x] Documentation Examples


dimamarksman:
Noticed an error in examples
1. Open mobile value help dialog example
2. Go to the define conditions tab
3. See error in the console

image

dimamarksman:
`Selection type: once` example: 
1. After selecting an item the dialog is not closed
2. After selecting an item and clicking `ok` the dialod is closed and selected item is displayed as `undefined`
image

dimamarksman:
In mobile mode when user goes to the `advanced` search options tab and click `cancel` there the table is not displayed anymore.
Steps:
1. Open first exampple (mobile mode)
2. Go to the  `select from list` tab.
3. Click `advanced search` (it moves to the `advanced search tab`).
4. Click `cancel` in the bottom of modal.
See result: we are still on the `advanced search tab` but without advanced search fields.
Expected result: should navigate back to the `Select from list` tab

libs/platform/src/lib/components/value-help-dialog/value-help-dialog/value-help-dialog.component.ts
```diff
@@ -0,0 +1,643 @@
...
+    } else {
+      if (this.activeDialog) {
+        this.activeDialog.close();
+      }
+    }
+  }
+
+  /** @hidden */
+  _trackByFilterFn(_index: number, item: VhdFilter): number | string | undefined {
+    return item && item.key ? item.key : undefined;
```
dimamarksman, line 497: Just wondering if there is a special meaning in returning `undefined`?

libs/platform/src/lib/components/value-help-dialog/value-help-dialog/value-help-dialog.component.ts
```diff
@@ -0,0 +1,643 @@
...
+   */
+  removeExcluded(index: number): void {
+    this.value.excluded = this.excludedItems.filter((_, i) => i !== index);
+    this._changeDetectorRef.markForCheck();
+  }
+
+  /**
+   * Switch tab by type
+   */
+  switchTab(type?: VhdTab): void {
```
dimamarksman, line 454: Let's use `(type: VhdTab = null)` to keep it consistent with initial `null` value

libs/platform/src/lib/components/value-help-dialog/value-help-dialog/value-help-dialog.component.scss
```diff
@@ -0,0 +1,104 @@
...
+    }
+    &__tokens-wrapper {
+        margin: 0 0 0 2.75rem;
+    }
+    &__tokens {
+        display: flex;
+        align-items: center;
+        border: 1px solid #89919a;
+        padding-right: 0.5rem;
+        border-radius: var(--sapElement_BorderCornerRadius, .25rem);
```
dimamarksman, line 75: It's needed to keep backup value beside `var` to make it working in IE11

libs/platform/src/lib/components/value-help-dialog/value-help-dialog/value-help-dialog.component.html
```diff
@@ -0,0 +1,341 @@
...
+                        [compact]="true"
+                        (click)="dismiss()"></button>
+                </div>
+            </ng-template>
+        </fd-dialog-footer>
+    </fd-dialog>
+</ng-template>
+
+<ng-template #selectedPanelTitle>
+    <h6 fd-title><ng-container i18n="@@platformI18nValueHelpDialog.Selected.IncludedLabel">Selected items</ng-container> ({{ selectedAndIncluded }})</h6>
```
dimamarksman, line 160: I would recommend to keep i18n marker wrapping whole place including variable as well, it will gave a translator a better context

libs/platform/src/lib/components/value-help-dialog/models/vhd-data.source.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+}
+
+export class ObservableValueHelpDialogDataSource<T> extends ValueHelpDialogDataSource<T> {
+    constructor(data: Observable<T[]>) {
+        super(new BaseDataProvider(data));
+    }
+}
+
+export class VhdDataProvider extends DataProvider<any> {
```
dimamarksman, line 44: Why we don't use `T` instead of any?

dimamarksman (commented):
Looks good! 
That's a big component, got mostly questions

dimamarksman:
The new exclude rule template is based on the last include rule options that is not applicabple in some cases:

image

sKudum:
Hi. @mykolavorobiov,

Few points.

1).In checklist breaking changes is marked true, value help is new component, where are we introducing breaking change?
2).Expect 1 other *.spec.ts files are empty event tho the respective components are have functions covered. Do we have any other plans for them?
3). On multi select , only checkboox select is working but not row select, Are we not going to provide this option to user.
4).In RTL, content in dialog remains in LTR.
5).Keyboard activities are not getting detected. Are we not going to support Keyboard?

dimamarksman:
@mykolavorobiov preview build is failed

mykolavorobiov:
Hey @sKudum 
1) Removed, had any breaking changes.
2) I will improve specs after approved component's logic. Don't want to do double fix in main logic and specs too.
3) Took this table from https://fundamental-ngx.netlify.app/#/core/table#with-checkboxes and https://fundamental-ngx.netlify.app/#/core/table#table-pop-in examples.
4) RTL switcher is not a great solution for this example. You can see same behaviour with https://fundamental-ngx.netlify.app/#/core/dialog. If you want to see correct RTL behaviour, please set up rtl directionon th body tag.

IlyaSurmay (commented):
a few minor comments

libs/platform/src/lib/components/value-help-dialog/components/define-tab/define-tab.component.html
```diff
@@ -0,0 +1,170 @@
...
+                        colGrow
+                        i18n-placeholder="@@platformI18nValueHelpDialog.Define.Condition.StrategyPlaceholder"
+                        placeholder="Select a strategy"
+                        [name]="uid + '-' + type + '-strategy-' + item.id"
+                        [(ngModel)]="item.strategy"
+                        (ngModelChange)="_filterChanged()">
+                        <fd-option *ngFor="let strategy of strategies" [value]="strategy.key">
+                            <ng-container *ngIf="type === 'exclude'" i18n="@@platformI18nValueHelpDialog.Define.Condition.StrategyPrefixExclude">not</ng-container>
+                            {{ strategy.label }}
+                        </fd-option>
```
dimamarksman, line 79: I'm afraid that's not going to work well with other languages.
`strategy.label` is not translatable and we can not translate "`not`" separately from other part of phrase, that should be one translation unit

dimamarksman:
Noticed that selected item from table appears in the outside model before an user confirms his selection

image

Steps:
1. Open dialog in multiselect mode
2. select first option
3. select second option
4. unselect first option
See result that second option becoms selected but user did not confirm his selection by clicking `OK` button

dimamarksman (approved):

stefanoScalzo:
When searching for 17, shouldn't it filter things that begin with a 1 then a 7 and not have 1 and 7 present?

stefanoScalzo:
If he clicks "enter" while searching, shouldn't it trigger a filter?

stefanoScalzo:
Searching for multiple things at once doesn't filter properly. example filtering both city and the name only filters the city

mykolavorobiov:
@stefanoScalzo 
> When searching for 17, shouldn't it filter things that begin with a 1 then a 7 and not have 1 and 7 present?
I think no, if we enter `17` it should be as one word, not a separate symbols.
Screenshot with results from original feature.
image

stefanoScalzo:
> @stefanoScalzo
> 
> > When searching for 17, shouldn't it filter things that begin with a 1 then a 7 and not have 1 and 7 present?
> > I think no, if we enter `17` it should be as one word, not a separate symbols.
> > Screenshot with results from original feature.
> > image

True I saw something different, but searching multiple things in conjunction doesn't work

mykolavorobiov:
@stefanoScalzo 
> Searching for multiple things at once doesn't filter properly. example filtering both city and the name only filters the city
Updated. Now all filled filters area required.




> > @stefanoScalzo
> > > When searching for 17, shouldn't it filter things that begin with a 1 then a 7 and not have 1 and 7 present?
> > > I think no, if we enter `17` it should be as one word, not a separate symbols.
> > > Screenshot with results from original feature.
> > > image
> 
> True I saw something different, but searching multiple things in conjunction doesn't work

I will update searching by multiple filters shortly.

mykolavorobiov:
@stefanoScalzo I pushed small changes, please check one more time.

mykolavorobiov:
Create new PR https://github.com/SAP/fundamental-ngx/pull/4184 for Travis

Issue #3887, opened 2020-11-18
Message box semantic icons


salarenko:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
Implement semantic icons into message box component. After introducing semantic icons, remember to give the user a way to optionally hide them.

#### Is there anything else we should know?
Currently message box is missing semantic icons because of [this fundamental styles bug](https://github.com/SAP/fundamental-styles/issues/1896). Once the issue is resolved semantic icons should be implemented.

2h

PR #4116, opened 2020-12-15
fix: (Core|Platform) bring v.0.14.0 of fundamental-styles

InnaAtanasova:
#### Please provide a brief summary of this pull request.
This PR is bringing version 0.14.0 of Fundamental-styles. 
Changes I made:
- removed custom styling for Flexible column layout. Now it's importing the styling from Fundamental-styles
- removed `fd-dynamic-page__collapsible-header-container--collapsed` class which was removed in Fundamental-styles
- updated Switch component to use the latest class naming
- re-structured Switch component to reflect the latest changes from Fundamental-styles, added a span with optional text for active/inactive state

BREAKING CHANGE:
- input property `optionalText` for Switch has been removed

#### Please check whether the PR fulfills the following requirements

- [x] the commit message follows the guideline:
https://github.com/SAP/fundamental-ngx/blob/master/CONTRIBUTING.md
- [x] tests for the changes that have been done
- [x] all items on the PR Review Checklist are addressed :
https://github.com/SAP/fundamental-ngx/wiki/PR-Review-Checklist

Documentation checklist:
- [na] update `README.md`
- [x] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)
- [x] Documentation Examples
- [x] Stackblitz works for all examples

PR #4108, opened 2020-12-14
fix: (Core) remove DIALOG_CONFIG and DIALOG_REF and make DialogConfig and DialogRef injectable

InnaAtanasova:
#### Please provide a link to the associated issue.
Closes https://github.com/SAP/fundamental-ngx/issues/3903

#### Please provide a brief summary of this pull request.

BREAKING CHANGE:
removed DIALOG_CONFIG and DIALOG_REF and made DialogConfig and DialogRef classes injectable

**Before:**
```
constructor(
    @Inject(DIALOG_CONFIG) public dialogConfig: DialogConfig,
    @Inject(DIALOG_REF) private _dialogRef: DialogRef
)
```

**Now:**
```
constructor(
    public dialogConfig: DialogConfig,
    private _dialogRef: DialogRef
)
```

#### Please check whether the PR fulfills the following requirements

- [x] the commit message follows the guideline:
https://github.com/SAP/fundamental-ngx/blob/master/CONTRIBUTING.md
- [x] tests for the changes that have been done
- [x] all items on the PR Review Checklist are addressed :
https://github.com/SAP/fundamental-ngx/wiki/PR-Review-Checklist

Documentation checklist:
- [na] update `README.md`
- [x] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)
- [x] Documentation Examples
- [x] Stackblitz works for all examples

PR #4128, opened 2020-12-16
feat: (Core) add semantic icons to Message box and remove the close button

InnaAtanasova:
Closes https://github.com/SAP/fundamental-ngx/issues/3887

- introduced semantic icons into Message box component
- removed the close button as per Fiori specs
- the Switch component is broken but it comes from v.0.14.0 of Fundamental-styles and is fixed in [this PR](https://github.com/SAP/fundamental-ngx/pull/4116) 

BREAKING CHANGE:
`MessageBoxCloseIconComponent` component is removed

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


InnaAtanasova:
**NOTE:** tests are failing for FlexibleColumnLayout but they are fixed in 2 other PRs (just need someone to review them :) )
https://github.com/SAP/fundamental-ngx/pull/4108/files and https://github.com/SAP/fundamental-ngx/pull/4116

As mentioned in the template, Switch component is broken but it's due to v.0.14.0 which is fixed in https://github.com/SAP/fundamental-ngx/pull/4116

IlyaSurmay (approved):

libs/core/src/lib/message-box/message-box-semantic-icon/message-box-semantic-icon.component.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Message box element representing the semantic icon in the message box header.
+ *
+ * ```html
+ * <fd-message-box-semantic-icon></fd-message-box-semantic-icon>
+ * ```
+ */
+@Component({
+    selector: 'fd-message-box-semantic-icon',
+    template: `<i [ngClass]="'sap-icon--' + this._semanticIcon" role="presentation"></i>`,
```
salarenko, line 17: no `this.` required

salarenko (changes requested):
1. RIP close button. Do we have to remove this feature? It looks like it also should't be supported in the Dialog component, but we keep it there. It's weird that the design excluded such common element.
2. Should we force user to put the icon element in the header by himself? Maybe we could put it always in the default message-box header template and show the user how to compose it by himself in the " Complex template" example?

libs/core/src/lib/message-box/message-box-semantic-icon/message-box-semantic-icon.component.ts
InnaAtanasova: 1. It's not up to us to decide if a button should be show or not. There's no button in the designs and in the UX guidelines so if we want a component to be following Fiori3 we should not display such button. 
2. will check this and let you know :)

libs/core/src/lib/message-box/message-box-semantic-icon/message-box-semantic-icon.component.ts
```diff
@@ -0,0 +1,64 @@
...
+        this.messageBoxConfig = this.messageBoxConfig || {};
+    }
+
+    /** @hidden */
+    get _showSemanticIcon(): boolean {
+        return this.messageBoxConfig.showSemanticIcon;
+    }
+
+    /** @hidden */
+    get _semanticIcon(): string {
```
salarenko, line 39: Minor suggestion, but I think in this form it's more readable:
```
get _getIcon(): string {
    return this.glyph || this.messageBoxConfig.customSemanticIcon || this._semanticIcon;
}

get _semanticIcon(): string {
    switch (this.messageBoxConfig.type) {
       case 'error':
           return 'message-error';
       case 'success':
           return 'message-success';
       case 'warning':
           return 'message-warning';
       case 'information':
           return 'message-information';
       case 'confirmation':
           return 'question-mark';
       default:
           return '';
       }
}
```
Original form is also fine, but it looks like we don't need default value in the switch statement.

libs/core/src/lib/message-box/message-box-header/message-box-header.component.html
```diff
@@ -8,13 +8,10 @@
...
     <ng-template #defaultTemplate>
         <div fd-bar-left>
-            <fd-bar-element [fullWidth]="true">
-                <ng-content select="[fd-title]"></ng-content>
-            </fd-bar-element>
-        </div>
-        <div fd-bar-right>
             <fd-bar-element>
-                <ng-content select="fd-message-box-close-icon"></ng-content>
+                <ng-content select="fd-message-box-semantic-icon"></ng-content>
```
salarenko, line 12: What's the point of having this content slot?

libs/core/src/lib/message-box/message-box-header/message-box-header.component.ts
```diff
@@ -41,6 +41,14 @@ export class MessageBoxHeaderComponent extends DialogHeaderBase implements After
         this.messageBoxConfig = this.messageBoxConfig || {};
     }
 
+    /** @hidden */
+    get _showSemanticIcon(): boolean {
+        if (this.messageBoxConfig.type) {
```
salarenko, line 46: Can we simplify this to:
```
return this.messageBoxConfig.type && this.messageBoxConfig.showSemanticIcon || this.messageBoxConfig.customSemanticIcon
```
And get rid of `hideSemanticIcon`?

salarenko (changes requested):

salarenko (approved):

JKMarkowski (approved):

PR #4252, opened 2021-01-05
chore: fix Schematics for NG versions above 10.1

JKMarkowski:
There were some old methods, that are not supported by angular cli above and equal 10.1 version.
Now some of them are imported from `@schematics/angular` library.

To test it:
- You need testing app with ng@10.1 or above
- Fetch this branch
- run `npm run build-pack-library`
- copy `.tgz` file from `dist/libs/core/` to testing app directory
- Add this version of lib by writing `ng add @fundamental-ngx/core@{{path to tgz file}}`

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/schematics/ng-add/index.ts
+++ b/libs/core/schematics/ng-add/index.ts
@@ -2,13 +2,13 @@ import { Rule, SchematicContext, Tree, chain, SchematicsException } from '@angul
 import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
 import { getPackageVersionFromPackageJson, hasPackage } from '../utils/package-utils';
 import { addPackageJsonDependency, NodeDependency, NodeDependencyType } from '@schematics/angular/utility/dependencies';
-import { addImportToRootModule, getProjectFromWorkspace, hasModuleImport } from '../utils/ng-module-utils';
-import { getAppModulePath } from '@schematics/angular/utility/ng-ast-utils';
-import { getWorkspace } from '@schematics/angular/utility/config';
+import { hasModuleImport } from '../utils/ng-module-utils';
 import { WorkspaceSchema } from '@schematics/angular/utility/workspace-models';
 
 import { cdkVersion } from './versions';
 import { defaultFontStyle } from './styles';
+import { addModuleImportToModule, getProjectFromWorkspace, getProjectMainFile, getAppModulePath } from '@angular/cdk/schematics';
+import { getWorkspace } from '@schematics/angular/utility/config';
 
 const browserAnimationsModuleName = 'BrowserAnimationsModule';
 const noopAnimationsModuleName = 'NoopAnimationsModule';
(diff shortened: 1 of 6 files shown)


InnaAtanasova (approved):

libs/core/schematics/ng-add/index.ts
```diff
@@ -2,13 +2,13 @@ import { Rule, SchematicContext, Tree, chain, SchematicsException } from '@angul
...
 import { addPackageJsonDependency, NodeDependency, NodeDependencyType } from '@schematics/angular/utility/dependencies';
-import { addImportToRootModule, getProjectFromWorkspace, hasModuleImport } from '../utils/ng-module-utils';
-import { getAppModulePath } from '@schematics/angular/utility/ng-ast-utils';
-import { getWorkspace } from '@schematics/angular/utility/config';
+import { hasModuleImport } from '../utils/ng-module-utils';
 import { WorkspaceSchema } from '@schematics/angular/utility/workspace-models';
 
 import { cdkVersion } from './versions';
 import { defaultFontStyle } from './styles';
+import { addModuleImportToModule, getProjectFromWorkspace, getProjectMainFile, getAppModulePath } from '@angular/cdk/schematics';
```
fkolar, line 10: How do you solve this CDK dependency. ? CDK needed be installed even before npm install kicks in. If you dont have CDK installed , it works?

JKMarkowski: @fkolar  Good point, it throws error; `An unhandled exception occurred: Cannot find module '@angular/cdk/schematics'`. Anyway in `fundamental-ngx/core` package.json, CDK lib is added to peerDependency list, so it's mandatory. Should I take some methods straight from cdk then, instead of importing?

fkolar: The cdk needs to be there before peerDependency ;-) But I am testing something here locally on some other project (blank empty), where I tried to do what you did, but somehow it reads CDK from somewhere. and it works. still not sure how its possible. 

Maybe cuz I am doing npm link? hmm.

fkolar:
Just found interesting thing. as I kept meesign up wiht schematics and try to edit directly some node_module files to trace some problem such as ast-utils.js, I did nto reflected changes in host (target) project I did. But It used my other changes that I made in the source repo. 

So it looks like when you compile schematics it also compiles together nessesary libraries and functions. is this possible ?

JKMarkowski:
@fkolar 
> Just found interesting thing. as I kept meesign up wiht schematics and try to edit directly some node_module files to trace some problem such as ast-utils.js, I did nto reflected changes in host (target) project I did. But It used my other changes that I made in the source repo.
> 
> So it looks like when you compile schematics it also compiles together nessesary libraries and functions. is this possible ? Or it explains why CDK works as well but when you dont have local reference it fails

So for me
- I created test app, without CDK/NGX
- Tried to add this package by `ng add @fundamental...`
- There was an error, that module from `@angular/cdk/schematics` couldn't be found, so it just relies on import and doesn't compile imported code, before deploy

fkolar:
Then it looks like we need to rely on @schematics/angular functions instead of cdk

fkolar:
@JKMarkowski 
I found this project that is also using CDK schematics. it has cdk in the dependencies of published package.json

https://github.com/NG-ZORRO/ng-zorro-antd

JKMarkowski:
> @JKMarkowski
> I found this project that is also using CDK schematics. it has cdk in the dependencies of published package.json
> 
> https://github.com/NG-ZORRO/ng-zorro-antd
> 
> Before they start doing anything they first install depenencies I think

I see we are already adding CDK in schematics, but as you said, the order of methods in ngAdd chain has to be changed. I will test it

JKMarkowski:
So now `@angular/cdk` is added as dependency, it will be installed with `ngx/core` if needed. Schematics are working correct now

PR #4288, opened 2021-01-08
fix: (Core) multi-input and combobox filtering should use startsWith

### LINKED ISSUE #3532: Combobox - filtering options
Enhancement

Currently our combobox search function is using the "contains" method - searches the full label for a match. According to the Fiori 3 specs the default filtering method should be "starts with per term", which matches the beginning of each word in an item's text.Both options should be available for the user. 

To Do:
- implement the functionality
- document it with examples

1day

mikerodonnell89:
fixes #3532 

Updates multi-input and combobox so these components use `startsWith` filtering by default, rather than the previous `includes`.  Also updates docs


Betrozov (approved):

InnaAtanasova (changes requested):
As per Fiori3 the default option should be `startsWith` but we need to keep both options and provide an option to the user to choose which one to use.

apps/docs/src/app/core/component-docs/multi-input/examples/multi-input-filter-example/multi-input-filter-example.component.ts
```diff
@@ -12,6 +12,6 @@ export class MultiInputFilterExampleComponent {
             return content;
         }
         const search = searchTerm.toLocaleLowerCase();
-        return content.filter((item) => item.toLocaleLowerCase().startsWith(search));
+        return content.filter((item) => item.toLocaleLowerCase().includesWith(search));
```
JKMarkowski, line 15: I guess it should be `includes`

JKMarkowski:
@InnaAtanasova 

> As per Fiori3 the default option should be `startsWith` but we need to keep both options and provide an option to the user to choose which one to use.

I think this is what this PR brings, by default it's `startsWith`, but user can change is easily by writing his own compare method.

InnaAtanasova:
> @InnaAtanasova
> 
> > As per Fiori3 the default option should be `startsWith` but we need to keep both options and provide an option to the user to choose which one to use.
> 
> I think this is what this PR brings, by default it's `startsWith`, but user can change is easily by writing his own compare method.

If I remember correctly from the discussion we agreed to provide both options where the second option is given by a parameter. All other possible options can be implemented by the user.

InnaAtanasova (approved):

JKMarkowski (approved):

Issue #4230, opened 2021-01-04
Bug: Core Input Group can edit input in disabled state


kavya-b:
#### Is this a bug, enhancement, or feature request?
bug
#### Briefly describe your proposal.
In disabled state, one can tab into the input and edit it.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
latest v0.26.0-rc.28
#### If this is a bug, please provide steps for reproducing it.
1. Go to https://fundamental-ngx.netlify.app/#/core/inputGroup
2. Go to example `Input Group Search Component Within Angular Reactive Forms` and tab into the input
3. Observe that one can write into it.
<img width="435" alt="Screenshot 2021-01-04 at 4 45 37 PM" src="https://user-images.githubusercontent.com/53509521/103530088-d8964400-4eac-11eb-979c-de59d95efca7.png">

PR #4292, opened 2021-01-08
fix: (Core) InputGroup editing text in disabled state

olshansky:
Closes: https://github.com/SAP/fundamental-ngx/issues/4230

Documentation checklist:
- [n/a] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/input-group/input-group.component.html
+++ b/libs/core/src/lib/input-group/input-group.component.html
@@ -35,6 +35,7 @@
                 [attr.aria-expanded]="isExpanded"
                 [attr.aria-label]="glyphAriaLabel"
                 [attr.aria-controls]="input"
+                [disabled]="disabled"
                 (mousedown)="preventFocus($event)"
                 (click)="buttonClicked($event)"
             >
@@ -63,6 +64,7 @@
         fd-input-group-input
         [compact]="compact"
         [type]="type"
+        [disabled]="disabled"
         *ngIf="!inputElement && !inputTemplate"
         [attr.placeholder]="placeholder"
     />


Betrozov (approved):

kavya-b (approved):

InnaAtanasova (approved):

InnaAtanasova:
Screen Shot 2021-01-08 at 10 54 54 AM

It's not described in the issue this PR is fixing but disabled Input Group Search has 2 other problems:
1. You can still focus it
2. The focus is around the whole element AND on the button. IMO it should not focus the button, only the input and when you tab again the focus should go on the button
Do you know if there's an issue about this?

olshansky:
> It's not described in the issue this PR is fixing but disabled Input Group Search has 2 other problems:
> 
> 1. You can still focus it
> 2. The focus is around the whole element AND on the button. IMO it should not focus the button, only the input and when you tab again the focus should go on the button
>    Do you know if there's an issue about this?

@InnaAtanasova I believe you are right. Fixed

JKMarkowski:
InputGroup/Button are still focusable - also button can be clicked inside. 

Peek 2021-01-11 14-14

olshansky:
> InputGroup/Button are still focusable - also button can be clicked inside. - I see that it's not scope of the issue it fixes, but can you fix it there?

@JKMarkowski button should be just not clickable or not focusable? And should the input be focusable?

JKMarkowski:
> button should be just not clickable or not focusable? And should the input be focusable?

@olshansky I think button shouldn't be focusable/clickable + input shouldn't be focusable as well - like on basic [disabled input](https://fundamental-ngx.netlify.app/#/core/input#reactive-form-input) . @InnaAtanasova What do you think about it?

InnaAtanasova:
> > button should be just not clickable or not focusable? And should the input be focusable?
> 
> @olshansky I think button shouldn't be focusable/clickable + input shouldn't be focusable as well - like on basic [disabled input](https://fundamental-ngx.netlify.app/#/core/input#reactive-form-input) . @InnaAtanasova What do you think about it?

In my opinion neither the input nor the button should be clickable or focusable in disabled state

olshansky:
> > > button should be just not clickable or not focusable? And should the input be focusable?
> > 
> > 
> > @olshansky I think button shouldn't be focusable/clickable + input shouldn't be focusable as well - like on basic [disabled input](https://fundamental-ngx.netlify.app/#/core/input#reactive-form-input) . @InnaAtanasova What do you think about it?
> 
> In my opinion neither the input nor the button should be clickable or focusable in disabled state

But what if the disabled state is the sort of "middle" state? Here is a pretty familiar use-case https://ux.stackexchange.com/questions/103239/should-disabled-elements-be-focusable-for-accessibility-purposes

InnaAtanasova:
> > > > button should be just not clickable or not focusable? And should the input be focusable?
> > > 
> > > 
> > > @olshansky I think button shouldn't be focusable/clickable + input shouldn't be focusable as well - like on basic [disabled input](https://fundamental-ngx.netlify.app/#/core/input#reactive-form-input) . @InnaAtanasova What do you think about it?
> > 
> > 
> > In my opinion neither the input nor the button should be clickable or focusable in disabled state
> 
> @InnaAtanasova @JKMarkowski But what if the disabled state is the sort of "middle" state? Here is a pretty familiar use-case https://ux.stackexchange.com/questions/103239/should-disabled-elements-be-focusable-for-accessibility-purposes

@olshansky the discussion you sent makes sense. I am not an a11y expert and can't comment more on this issue. We wrote and asked for an external opinion and will let you know when we get an answer. In Fundamental-NGX when we have a disabled component/element we don't focus it by TAB, that's how all components are implemented so far.

olshansky:
> @olshansky the discussion you sent makes sense. I am not an a11y expert and can't comment more on this issue. We wrote and asked for an external opinion and will let you know when we get an answer. In Fundamental-NGX when we have a disabled component/element we don't focus it by TAB, that's how all components are implemented so far.

Ok, thank you as usual 😊

JKMarkowski (approved):

PR #4321, opened 2021-01-12
fix: (Core|Platform) refactor select component with (material approach) cdk overlay, key manager, selection model

### LINKED ISSUE #4254: fix(core): select component modifications to implement material approach
fix

Create select component in core and update any breaking changes as per Franks suggested material approach.

10

sKudum:

closes #4254

Refactored select component in core with material approach.
parallel this PR fixes build issues in platform due to this changes.

BREAKING CHANGE:

Not required 
**extendedBodyTemplate**- can be deprecated- as it is handled internally by select.component, controlTemplate!=null or undefined is considered as extendedBodyTemplate as true.

**Before:**
```
<fd-select [(value)]="selectedValue"  [extendedBodyTemplate]="true" [controlTemplate]="customSelectTemplate">
```

**Now:**
```
<fd-select [(value)]="selectedValue" [controlTemplate]="customSelectTemplate">
```

Documentation checklist:
- [n/a ] update `README.md`
- [x] Documentation Examples


fkolar:
I thought that I last discussion was that we will review existing core implementation and it is already using CDKOverlay but its still missing other features. SO we decided to pull over this old UI ?

sKudum:
> I thought that I last discussion was that we will review existing core implementation and it is already using CDKOverlay but its still missing other features. SO we decided to pull over this old UI ?

Hi @fkolar, 
Yes considering it below steps are  followed.
1. compared existing core implementations and the suggested approach and did changes where ever its required.
2. While doing changes I made sure non of the existing features to break. expect closeOnOutsideClick, all are achiveable with material approach.
3. Material has multi select and invalid option features, for which we don't have styles in fiori due to which I was not able to give examples in documentations.

libs/core/src/lib/select/commons/select-commons.ts
```diff
@@ -0,0 +1,994 @@
...
+/**
+ * Common behaviors from Material Select Component to build Fundamental-ngx Select on top of it. We need this solid
+ * foundation to deliver robust solution and only focus on additional Fiori 3 enhancement.
+ *
+ * Even this seems to be implemented as one big mixins with single anonymous class all this is going to be broking down
+ * into several pieces in order to maximize reusability. Such as a possible candidates would be to separate
+ * Keyboard Support, Overlay support and option panel positioning and maybe select behavior itself.
+ *
+ * @Hidden
+ */
```
InnaAtanasova, line 43: ?

libs/core/src/lib/select/commons/select-commons.ts
```diff
@@ -0,0 +1,994 @@
...
+
+/**
+ * The select panel will only "fit" inside the viewport if it is positioned at
+ * this value or more away from the viewport boundary.
+ */
+export const SELECT_PANEL_VIEWPORT_PADDING = 5;
+
+/** The height of the select items in `em` units. */
```
InnaAtanasova, line 51: Fundamental library uses `rem` units

libs/core/src/lib/select/commons/select-commons.ts
```diff
@@ -0,0 +1,994 @@
...
+const SELECT_HEADER_IDENTIFIER = '.fd-list__group-header';
+
+export type Constructor<T> = new (...args: any[]) => T;
+
+/**
+ * Common Interface to define shareable properties and method
+ *
+ * @hidden
+ */
+export interface MatSelectCommonBehavior {
```
InnaAtanasova, line 73: Can we avoid naming interfaces and classes starting with `MAT`?

libs/core/src/lib/select/commons/select-commons.ts
```diff
@@ -0,0 +1,994 @@
...
+         * @hidden
+         */
+        closeOnOutsideClick: boolean;
+
+        /**
+         * Handles selection interaction logic
+         * @hidden
+         */
+        selectionModel: SelectionModel<OptionComponent>;
+
```
InnaAtanasova, line 187: hidden props and methods should be prefixed with `_`

libs/core/src/lib/select/select-mobile/select-mobile.component.spec.ts
```diff
@@ -29,14 +30,15 @@ class TestWrapperComponent {
     constructor(@Inject(MOBILE_CONFIG_TEST_TOKEN) public mobileConfig: MobileModeConfig) { }
 }
 
-describe('SelectComponent in mobile mode', () => {
+fdescribe('SelectComponent in mobile mode', () => {
```
InnaAtanasova, line 33: Can you remove the focus?

InnaAtanasova (commented):
Please check all methods and props that are either private or hidden, they should be prefixed with `_`

libs/core/src/lib/select/commons/select-commons.ts
fkolar: many of the method that are private needs to be public. We need to also update the API docs. In development you always have cases where: 

private _xxxx -> you dont want anybody to use it from outside
_xxxxx -> Private API , but other classes access it internaly from outside (it can change)

libs/core/src/lib/select/commons/select-commons.ts
fkolar: This shoudl be ok, as we also added MIT licence to the TOP as this file and its content was taken over from Material Design

libs/platform/src/lib/components/form/select/select.component.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { async, ComponentFixture, TestBed, fakeAsync, tick } from '@angular/core/testing';
 
-import { SelectPlatformComponent } from './select.component';
+import { PlatformSelectComponent } from './select.component';
```
fkolar, line 3: We should remove and comment out platform implemenation and submit this as separate PR. It should not be part of this

libs/core/src/lib/select/select.component.html
```diff
@@ -1,90 +1,103 @@
...
-        [disabled]="disabled"
-        [appendTo]="appendTo"
-        [options]="popperOptions"
-        [closeOnEscapeKey]="false"
-        [fillControlMode]="'at-least'"
-        [maxWidth]="600"
-        [closeOnOutsideClick]="closeOnOutsideClick"
-        (isOpenChange)="popoverOpenChangeHandle($event)">
+<!--
+    With CDK overlays the origin trigger and connected overlay needs to be together within one level
```
fkolar, line 4: Please consider using Popover that is overlay inside. I am not sure what performance impact we will get and we need to test it.

fkolar (changes requested):

libs/core/src/lib/select/commons/select-commons.ts
JKMarkowski, line 73: I see values of bringing this into library, there are a lot of repeated functionalities, such as focus support / rewritten overlay(Popover) methods .

But there is integration with selection model and list key manager that IMO should be completely separated, wrapped inside reusable standalone services, so we can get rid of this big class.

libs/core/src/lib/select/select.module.ts
```diff
@@ -1,23 +1,22 @@
...
     imports: [
         BusyIndicatorModule,
         CommonModule,
-        PopoverModule,
         ButtonModule,
         IconModule,
-        ListModule
-    ]
+        ListModule,
+        OverlayModule]
```
JKMarkowski, line 20: Formatting

libs/core/src/lib/select/select.component.ts
```diff
@@ -57,525 +96,362 @@ export interface OptionStatusChange {
...
+    @Input()
+    mobileConfig: MobileModeConfig = { hasCloseButton: true };
 
     /** Custom template used to build control body. */
     @Input()
     controlTemplate: TemplateRef<any>;
 
-    /** The element to which the popover should be appended. */
     @Input()
-    appendTo: ElementRef;
```
JKMarkowski, line 157: I guess this can be considered as a breaking change, becuase this Input has been removed

libs/core/src/lib/select/option/option.component.spec.ts
```diff
@@ -1,95 +0,0 @@
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
```
JKMarkowski, line 1: Are you entirely sure that this nothing to test?

JKMarkowski (commented):
I added some comments regarding code. Everything works great there, good job with bringing selection model. 

But I will go with completely other approach, to separate SelectionModel and KeyManager into standalone services, to make them reusable in other similar components.

libs/core/src/lib/select/commons/select-commons.ts
droshev: @sKudum you need to follow https://github.com/SAP/fundamental-ngx/blob/main/.reuse/dep5 in case we want to use this file

sKudum: Sure will do that

sKudum:
Hi @fkolar, @InnaAtanasova , @Deno , @JKMarkowski 

     I have addressed your comments, please have a look.
It has now:

1). _ , modifiers added where ever required.
2). Mat licence added in suggested formate.
3). test scripts added for options, focus are removed.
4). Popover component is used instead of cdkoverlay.
5). appendTo, closeonoutsideclick is reverted back as we are using popover component.
6). _listenOnSelectOpenChange is used.



Hi @InnaAtanasova , @fkolar 

 As mentioned in description of PR,  this PR is targeted for core changes and because of this change where ever it is causing build failures  those files are also fixed, which includes platform files too.

libs/core/src/lib/select/commons/select-commons.ts
fkolar: This implementation was commented already many times, we can always improve things, but now its how it is in material select - whcih reminds me @sKudum have you updated to latest code changes that are in version 10 for material select ?

libs/core/src/lib/select/select-mobile/select-mobile.component.ts
```diff
@@ -9,44 +9,71 @@ import {
...
-        this._listenOnSelectOpenChange();
+        // Listens for Select option HIDE/SHOW events in order to show or hide this dialog
+        // When in non dismissible mode we save currently selected item index to rollback original selection
+        //
+        this._subscriptions.add(
+            this._parent.openedChange.subscribe((isOpen) => {
+                this.dialogRef.hide(!isOpen);
+                this._activeItemIndex = this._parent.keyManager.activeItemIndex;
+            })
+        );
```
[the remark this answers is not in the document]
fkolar: This more we cahgne things and shuffle around the longer it takes . This PR has already over 300s commends ;-)

libs/core/src/lib/select/iselect-common-behavior.ts
```diff
@@ -0,0 +1,78 @@
+import { EventEmitter, ElementRef, QueryList } from '@angular/core';
```
fkolar, line 1: Maybe I missed some comment, why we call this file iselect?

libs/platform/src/lib/components/form/select/select.component.ts
```diff
@@ -2,25 +2,35 @@ import {
...
 @Component({
     selector: 'fdp-select',
     templateUrl: './select.component.html',
     styleUrls: ['./select.component.scss'],
     providers: [
         {
-            provide: NG_VALUE_ACCESSOR,
-            useExisting: forwardRef(() => SelectPlatformComponent),
+            provide: FormFieldControl,
+            useExisting: SelectPlatformComponent,
```
fkolar, line 33: We dont prefix component with platform , this is only for Module. it shoudl be just SelectComponent

fkolar (changes requested):

libs/core/src/lib/select/commons/select-commons.ts
```diff
@@ -0,0 +1,927 @@
...
+        get focused(): boolean {
+            return this._focused || this.panelOpen;
+        }
+
+        get selected(): OptionComponent {
+            return this._selectionModel.selected[0];
+        }
+
+        get positions(): ConnectedPosition[] {
```
dimamarksman, line 256: Do we really need a getter here?

dimamarksman (approved):
lgtm

libs/platform/src/lib/components/form/select/select.component.ts
sKudum: Sure @Frank, (this is taken care) all platform related changes will be pushed in another PR. as name change was leading to push more files from platform. I kept it as it is.

libs/core/src/lib/select/commons/select-commons.ts
sKudum: Hi @fkolar, I have compared latest select implementation with the changes we did in previous PR. 

 It is observed, that most of the changes done in past few months in material select are with respect to styles, directions and accessibility which are already taken care in popover component as per Fiori standards for us and tried  not to break already existing features in select given.

Considering above scenarios, In this PR I have used names for classes and properties as suggested in previous PR for mat fields and for styles, overlay, directions, positioning using popover component. For selectionModel, keyManager used material approach.

Please do let me know in case am missing anything in this process.

libs/core/src/lib/select/commons/select-commons.ts
```diff
@@ -0,0 +1,820 @@
+/**
+* The MIT License
+* Format: Copyright (c) 2010-2020 Google LLC. http://angular.io/license
+* Upstream-Name: Angular material
+* Upstream-Contact: fundamental@sap.com
```
mikerodonnell89, line 5: I think this should be google, not us

mikerodonnell89 (changes requested):
See comments

libs/core/src/lib/select/commons/select-commons.ts
sKudum: Hi @mikerodonnell89 , "google". hmm...there is no contact details shared in MIT licence from angular team.Shall I keep it blank??

sKudum:
As per the discussions taken by Frank, Deno, Dima  that "We won't be having material approach" closing this PR.

Issue #4337, opened 2021-01-14
Tabs used inside DynamicPageContent


JKMarkowski:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
So user add `fd-tabs` component niside `DynamicPageContent` there is added inline styling like
![image](https://user-images.githubusercontent.com/26483208/104594957-33c8f300-5672-11eb-8773-87801381a9b3.png)

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Newest

#### If this is a bug, please provide steps for reproducing it.
Stakcblitz: https://stackblitz.com/edit/angular-jbpbea?file=src/index.html

kavya-b:
19/01 to 20/01- worked on changing fixed positioning design, and making dynamic page work with flexible column layout. Re-requested review after incorporating changes requested. 
21/01- Removed some unused code and fixed a failing test.

PR #4343, opened 2021-01-15
fix: (Platform) remove dependency on position: fixed for Dynamic Page

kavya-b:
#4337 
- This PR fixes issue with having `position: fixed` in Dynamic Page that was causing other side-effects like appearing over the entire page and not taking up its parent's space. We are now using `position: static`. Also added an example for using with Flexible Column Layout.

Before:
Screen Shot 2021-01-15 at 9 37 51 AM

After:
image

- This PR also fixes issue where dynamic page tab styles were getting applied to other `fd-tabs` used within the page. It fixes the issue by applying more specificity to the dynamic page tab styles application.

Before:
Screenshot 2021-01-15 at 9 22 47 AM
After:
Screenshot 2021-01-15 at 9 24 23 AM

Within tabbed dynamic page example, styles will only apply to dynamic page tabs, not user-provided tabs in content:

Documentation checklist:
- [n/a] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/platform/src/lib/components/dynamic-page/dynamic-page.component.scss
+++ b/libs/platform/src/lib/components/dynamic-page/dynamic-page.component.scss
@@ -4,16 +4,8 @@
     height: 3rem;
 }
 
-.header-sticker, .tab-sticker {
-    position: absolute;
-    width: 100%;
-    top: 0;
-}
-
-.footer-sticker, .content-sticker{
-    position: absolute;
-    bottom: 0;
-    width: 100%;
+.content-sticker {
+    position: static;
 }
 
 .fd-dynamic-page__header--not-collapsible {
--- a/libs/platform/src/lib/components/dynamic-page/dynamic-page.component.scss
+++ b/libs/platform/src/lib/components/dynamic-page/dynamic-page.component.scss
@@ -4,11 +4,16 @@
     height: 3rem;
 }
 
-.tab-sticker {
-    left: 0;
-    right: 0;
-    position: relative;
+.header-sticker, .tab-sticker {
+    position: absolute;
+    width: 100%;
     top: 0;
+}
+
+.footer-sticker, .content-sticker{
+    position: absolute;
+    bottom: 0;
+    width: 100%;
 }
 
 .fd-dynamic-page__header--not-collapsible {
--- a/libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
+++ b/libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
@@ -1,4 +1,3 @@
-import { CdkScrollable, ScrollDispatcher } from '@angular/cdk/overlay';
 import {
     AfterContentInit,
     AfterViewInit,
@@ -10,19 +9,22 @@ import {
     ElementRef,
     HostBinding,
     Input,
-    NgZone,
     OnDestroy,
     QueryList,
     Renderer2,
     ViewChild,
+    ViewChildren,
     ViewEncapsulation
 } from '@angular/core';
-import { Subscription } from 'rxjs';
-import { startWith } from 'rxjs/operators';
-
+import { fromEvent, Subscription } from 'rxjs';
+import { debounceTime, startWith, throttleTime } from 'rxjs/operators';
 import { BaseComponent } from '../base';
-import { DynamicPageBackgroundType, CLASS_NAME, DynamicPageResponsiveSize } from './constants';
-import { DynamicPageContentComponent, DynamicPageTabChangeEvent } from './dynamic-page-content/dynamic-page-content.component';
+import { CLASS_NAME, DynamicPageBackgroundType, DynamicPageResponsiveSize } from './constants';
+import {
+    DynamicPageContentComponent,
+    DynamicPageTabChangeEvent
+} from './dynamic-page-content/dynamic-page-content.component';
+import { DynamicPageTabbedContentComponent } from './dynamic-page-content/dynamic-page-tabbed-content.component';
 import { DynamicPageHeaderComponent } from './dynamic-page-header/header/dynamic-page-header.component';
 import { DynamicPageTitleComponent } from './dynamic-page-header/title/dynamic-page-title.component';
 import { DynamicPageService } from './dynamic-page.service';
@@ -76,6 +78,10 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
     @ContentChildren(DynamicPageContentComponent, { descendants: true })
     tabbedContent: QueryList<DynamicPageContentComponent>;
 
+    /** reference to content component to filter tabs */
+    @ViewChildren(DynamicPageTabbedContentComponent)
+    tabContents: QueryList<DynamicPageTabbedContentComponent>;
+
     /**
      * @hidden
      * reference to header container
@@ -115,12 +121,6 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
      */
     private _collapseValSubscription: Subscription = new Subscription();
 
-    /**
-     * @hidden
-     * subscription for when content is scrolled
-     */
-    private _scrollSubscription: Subscription = new Subscription();
-
     /** @hidden */
     public headerCollapsible = true;
 
@@ -129,24 +129,20 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
         protected _cd: ChangeDetectorRef,
         private _elementRef: ElementRef<HTMLElement>,
         private _renderer: Renderer2,
-        private _dynamicPageService: DynamicPageService,
-        private _scrollDispatcher: ScrollDispatcher,
-        private _zone: NgZone
+        private _dynamicPageService: DynamicPageService
     ) {
         super(_cd);
         if (this._collapseValSubscription) {
             this._collapseValSubscription.unsubscribe();
         }
         this._collapseValSubscription = this._dynamicPageService.$collapseValue.subscribe((val) => {
-            this._setTabsPosition();
-            this._setContainerPosition();
-            this._setTabContainerPosition();
+            this.setContainerPositions();
         });
     }
 
     /** @hidden */
     ngAfterContentInit(): void {
-        this._listenToChildrenQueryListChanges();
+        this._listenToTabbedContentQueryListChanges();
         if (this.background) {
             this.titleComponent.background = this.background;
             this.headerComponent.background = this.background;
@@ -164,9 +160,7 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
     ngAfterViewInit(): void {
         this._setTabStyles();
         this._setToolbarStyles();
-        this._setTabsPosition();
-        this._setTabContainerPosition();
-        this._setContainerPosition();
+        this.setContainerPositions();
 
         this._subscriptions.add(
             this.tabbedContent.changes.subscribe(() => {
@@ -175,7 +169,8 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
             })
         );
         if (this.headerComponent?.collapsible) {
-            this.snapOnScroll();
+            this._listenOnScroll();
+            this._listenOnResize();
         }
         this._cd.detectChanges();
     }
@@ -183,28 +178,23 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
     /**@hidden */
     ngOnDestroy(): void {
         this._collapseValSubscription.unsubscribe();
-        this._scrollSubscription.unsubscribe();
         this._subscriptions.unsubscribe();
+    }
+
+    /**
+     * Set the positions of the tabs and content with respect to the window
+     */
+    setContainerPositions(): void {
+        this._setTabsPosition();
+        this._setContainerPosition();
+        this._setTabContainerPosition();
     }
 
     /**
      * Snap the header to expand or collapse based on scrolling. Uses CDKScrollable.
      */
     snapOnScroll(): void {
-        if (this._scrollSubscription) {
-            this._scrollSubscription.unsubscribe();
-        }
-        this._scrollSubscription = this._scrollDispatcher.scrolled(10).subscribe((cdk: CdkScrollable) => {
-            this._zone.run(() => {
-                const scrollPosition = cdk?.measureScrollOffset('top');
-                this.header.nativeElement.style.position = 'fixed';
-                if (scrollPosition > 0) {
-                    this._dynamicPageService.collapseHeader();
-                } else {
-                    this._dynamicPageService.expandHeader();
-                }
-            });
-        });
+        this._listenOnScroll();
     }
 
     /**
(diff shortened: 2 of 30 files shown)


DeepakSap14 (approved):

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
```diff
@@ -239,15 +239,17 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
...
+            if (tabList) {
+                tabList.style.top = this.header.nativeElement.offsetHeight + 'px';
+                tabList.style.position = 'fixed';
+                tabList.style.left = '0';
+                tabList.style.right = '0';
+                this.tabs.forEach((element) => {
+                    element.contentTop = tabList.offsetHeight + 'px';
+                });
+            }
         }
```
Betrozov, line 253: decrease nesting

```suggestion
        if (!this.contentContainer) {
            return;
        }

        const tabList: HTMLElement = this.contentContainer.nativeElement.querySelector('.fd-tabs');
        if (!tabList) {
            return;
        }
        
        tabList.style.top = this.header.nativeElement.offsetHeight + 'px';
        tabList.style.position = 'fixed';
        tabList.style.left = '0';
        tabList.style.right = '0';
        this.tabs.forEach((element) => {
            element.contentTop = tabList.offsetHeight + 'px';
        });
```

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
```diff
@@ -266,22 +268,24 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
...
                     );
+                    if (pinCollapseShadowElement) {
+                        this._addClassNameToCustomElement(
+                            pinCollapseShadowElement,
+                            CLASS_NAME.dynamicPageCollapsibleHeaderPinCollapseNoShadow
+                        );
+                    }
                 }
             }
         }
```
Betrozov, line 291: the same
decrease nesting
```suggestion
        if (!this.contentContainer) {
            return;
        }

        const tabList = this.contentContainer.nativeElement.querySelector('.fd-tabs');
        if (!tabList) {
            return;
        }

        this._addClassNameToCustomElement(tabList, CLASS_NAME.dynamicPageTabs);
        this._addClassNameToCustomElement(tabList, CLASS_NAME.dynamicPageTabsAddShadow);
        if (this.size) {
            this._setTabsSize(this.size, tabList);
        }

        if (!this.headerComponent?.collapsible) {
            return;
        }

        const pinCollapseShadowElement = this._elementRef.nativeElement.querySelector(
            '.fd-dynamic-page__collapsible-header-visibility-container'
        );
        if (!pinCollapseShadowElement) {
            return;
        }

        this._addClassNameToCustomElement(
            pinCollapseShadowElement,
            CLASS_NAME.dynamicPageCollapsibleHeaderPinCollapseNoShadow
        );
```

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
kavya-b: fixed as suggested, thanks for pointing this out!

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
kavya-b: fixed as suggested, except the last if `if (!pinCollapseShadowElement) {
            return;
        }`  was just kept like before since it wasn't further nested anyway =>
```
if (pinCollapseShadowElement) 
{ 
         this._addClassNameToCustomElement(
            pinCollapseShadowElement,
            CLASS_NAME.dynamicPageCollapsibleHeaderPinCollapseNoShadow
        ); 
}

**Review by Betrozov — APPROVED** [review]

**Review by InnaAtanasova — CHANGES_REQUESTED** [review]
Hi @kavya-b , unfortunately the issue is not fixed. The issue was detected by an application using Flexible Column Layout together with Platform Dynamic Page. 
The way you can reproduce the issue is:
- Go to `flexible-column-layout-example.component.html` and as a content for `#midColumn` template add the Dynamic Page example with Tabs
- Add the necessary imports for Button, Breadcumb, Bar, Toolbar, etc
- When you test the example now you will have this:
1. Open the example:
Screen Shot 2021-01-15 at 9 37 18 AM

2. Click the button to show the second column of the Flexible column layout:
Screen Shot 2021-01-15 at 9 37 40 AM

3. The result is:
Screen Shot 2021-01-15 at 9 37 51 AM

Because the position of the header is fixed, and you specify left:0, etc the header is now taking the whole width of the page, but not the width of the container. In this case the header should fit in the second column space.
Let me know if you need more info

**Comment by kavya-b** [comment]
> Hi @kavya-b , unfortunately the issue is not fixed. The issue was detected by an application using Flexible Column Layout together with Platform Dynamic Page.
> The way you can reproduce the issue is:
> 
> * Go to `flexible-column-layout-example.component.html` and as a content for `#midColumn` template add the Dynamic Page example with Tabs
> * Add the necessary imports for Button, Breadcumb, Bar, Toolbar, etc
> * When you test the example now you will have this:
> 
> 1. Open the example:
> 
> Screen Shot 2021-01-15 at 9 37 18 AM
> 
> 1. Click the button to show the second column of the Flexible column layout:
> 
> Screen Shot 2021-01-15 at 9 37 40 AM
> 
> 1. The result is:
> 
> Screen Shot 2021-01-15 at 9 37 51 AM
> 
> Because the position of the header is fixed, and you specify left:0, etc the header is now taking the whole width of the page, but not the width of the container. In this case the header should fit in the second column space.
> Let me know if you need more info

@InnaAtanasova  I see.. seems to be a problem with `position: fixed` being used at multiple places to achieve the sticking header on scroll. This issue with flexible column layout might need more time and effort to work on, and will probably not be a small fix that can be part of this PR. Should I raise a separate issue for flex column issue and consider this PR as only addressing the tab problem within dynamic page context; or should I close this PR and address both issues at a later point? What do you suggest?

**Comment by JKMarkowski** [comment]
> @InnaAtanasova I see.. seems to be a problem with `position: fixed` being used at multiple places to achieve the sticking header on scroll. This issue with flexible column layout might need more time and effort to work on, and will probably not be a small fix that can be part of this PR. Should I raise a separate issue for flex column issue and consider this PR as only addressing the tab problem within dynamic page context; or should I close this PR and address both issues at a later point? What do you suggest?

Hi @kavya-b This issue with `fixed` actually caused some other errors. 
Can't header become absolute, or even relative, without all of these css properties (top/right/left/position) set by JS? 
Content can be scrollable and header can stay at same position without even using `fixed`.

This can bring some unexpected behaviour for users - since we're pretty ready for release, I suggest 
Adding necessary changes to this PR that will be applied only after setting some option (ex. `@Input() fixed` enabled by default).  Or merge this PR, which partially fixes bug with tabs and address fixed problem in another PR/issue

**Comment by InnaAtanasova** [comment]
> > @InnaAtanasova I see.. seems to be a problem with `position: fixed` being used at multiple places to achieve the sticking header on scroll. This issue with flexible column layout might need more time and effort to work on, and will probably not be a small fix that can be part of this PR. Should I raise a separate issue for flex column issue and consider this PR as only addressing the tab problem within dynamic page context; or should I close this PR and address both issues at a later point? What do you suggest?
> 
> Hi @kavya-b This issue with `fixed` actually caused some other errors.
> Can't header become absolute, or even relative, without all of these css properties (top/right/left/position) set by JS?
> Content can be scrollable and header can stay at same position without even using `fixed`.
> 
> This can bring some unexpected behaviour for users - since we're pretty ready for release, I suggest
> Adding necessary changes to this PR that will be applied only after setting some option (ex. `@Input() fixed` enabled by default). Or merge this PR, which partially fixes bug with tabs and address fixed problem in another PR/issue
> 
> @InnaAtanasova what do you think?

I am not in favour of merging this PR as it's not really fixing the issue. If we merge it this means that the component can be used in only one specific case where it's used as a page layout and takes the whole width of the window. From what I see most of the use cases are when the component is part of a wrapper, not limited to Flexible column layout. The position should not be fixed. @JKMarkowski solution could work and is a better approach that the current using the styles.

**Comment by kavya-b** [comment]
@InnaAtanasova @JKMarkowski 
After adding a `fixed` input that is false by default and conditionally applying the styles, I was able to achieve this, but it has a few issues:

https://user-images.githubusercontent.com/53509521/104990657-0b127600-5a43-11eb-8518-c9fc0dda411b.mp4

- scrolling expands/collapses the header but the header does not stick anymore since `position: fixed` is no longer applied. @JKMarkowski could you give some pointers in the direction of resolving this issue? This may not be relevant, but when first developing the component, I used `cdkScrollable`s on the dynamic page content but it was too complicated in usage, and had some side-effects like double scrollbars(one on content, one on page scroll), and if multiple `cdkScrollable`s were used in the page, our scrolling logic would be applied to all `cdkScrollable`s. Since the documentation on `cdkScrollable` is not very comprehensive and we had to tweak things without knowing what other side-effects might appear later on, we decided(in PR [3515](https://github.com/SAP/fundamental-ngx/pull/3515), #[1](https://github.com/SAP/fundamental-ngx/pull/3515#issuecomment-705745937) #[2](https://github.com/SAP/fundamental-ngx/pull/3515#discussion_r500147097) #[3](https://github.com/SAP/fundamental-ngx/pull/3515#discussion_r500175345) #[4](https://github.com/SAP/fundamental-ngx/pull/3515#discussion_r500222952) )to listen to only page scrolls for which I had to use `position: fixed` (`position: sticky` would have been simpler but has no IE support) and the inline styles. 
- floating footer bar still comes over the entire page: this is because `position: absolute` is applied in the bar styles. However, if changed to `relative` it is contained within 2nd column, but is not floating any more and is only visible if you scroll to the end of the page.
- also, it seems flexible column layout also only provides one scroll at the page level, whereas it should probably be one scroll for each column.

**Comment by kavya-b** [comment]
Update: I am trying something using `position: absolute` which(if everything works right) might remove the need for having fixed in styles or as an input attribute. 


Achieving it for tabs with scrolling is a bit more challenging; I will continue working on it.

**Comment by kavya-b** [comment]
@InnaAtanasova @JKMarkowski have removed `position: fixed` and `cdkScrollable` in library now, achieving the same with `position: absolute` and scroll listeners. Also added an example for usage with Flexible Column Layout. 
Please take a look.

File: libs/platform/src/lib/components/dynamic-page/dynamic-page.component.scss [hunk 3]
```diff
@@ -4,13 +4,18 @@
...
 
-.tab-sticker {
-    left: 0;
-    right: 0;
-    position: relative;
+.header-sticker, .tab-sticker {
+    position: absolute;
+    width: 100%;
     top: 0;
 }
```
JKMarkowski, line 11: Could you keep it static instead of absolute?

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
```diff
@@ -229,24 +219,69 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
      */
     private _setContainerPosition(): void {
         if (this.contentComponent) {
-            this.contentComponent.getElementRef().nativeElement.style.top = this.header.nativeElement.offsetHeight + 'px';
-            this.contentComponent.getElementRef().nativeElement.style.position = 'relative';
+            const contentComponentElement = this.contentComponent.getElementRef().nativeElement;
+
+            contentComponentElement.style.top = this.header.nativeElement.offsetHeight + 'px';
+            this._addClassNameToCustomElement(contentComponentElement, 'content-sticker');
```
JKMarkowski, line 225: There is no need to add `top` anywhere. You can easily do it without applying `top` by applying static/relative positions

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
```diff
@@ -229,24 +219,69 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
...
+        if (!this.contentContainer) {
+            return;
+        }
+        const tabList: HTMLElement = this.contentContainer.nativeElement.querySelector('.fd-tabs');
+        if (!tabList) {
+            return;
+        }
+
+        this._addClassNameToCustomElement(tabList, 'tab-sticker');
+        tabList.style.top = this.header.nativeElement.offsetHeight + 'px';
```
JKMarkowski, line 274: Same, with static/relative you won't need to add it

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
```diff
@@ -229,24 +219,69 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
...
+        if (tabContent) {
+            tabContent.forEach((contentItem) => {
+                const element: HTMLElement = contentItem
+                    .getElementRef()
+                    .nativeElement.querySelector('.fd-dynamic-page__content');
+                if (element) {
+                    element.style.top = this.header.nativeElement.offsetHeight + tabList.offsetHeight + 'px';
+                }
             });
         }
```
JKMarkowski, line 286: That's also the thing that you don't need

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
```diff
@@ -266,24 +301,39 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
...
+        const pinCollapseShadowElement = this._elementRef.nativeElement.querySelector(
+            '.fd-dynamic-page__collapsible-header-visibility-container'
+        );
+
+        if (pinCollapseShadowElement) {
+            this._addClassNameToCustomElement(
+                pinCollapseShadowElement,
+                CLASS_NAME.dynamicPageCollapsibleHeaderPinCollapseNoShadow
+            );
         }
```
JKMarkowski, line 337: Let's recheck if this is still needed, when applied static/relative position

JKMarkowski (changes requested):
I added some comments. IMO current dynamic page is too complicated. 
Dynamic page looks like that:
image
Most of the things are not needed just to have it in this way. 
The only thing you will need to calculate for this proposition is height of content.

So you can apply on init just a height of content:
```
100vh - distance from top - offset(which is added by user)
```

Let me know what do you think about this approach, if it does make any sense

kavya-b:
> Hi @kavya-b, I added some comments. IMO current dynamic page is too complicated.
> Dynamic page looks like that:
> image
> Most of the things are not needed just to have it in this way.
> The only thing you will need to calculate for this proposition is height of content.
> 
> So you can apply on init, or header expand just a height of content:
> 
> ```
> 100vh - distance from top - offset(which is added by user)
> ```
> 
> Let me know what do you think about this approach, if it does make any sense.
> 
> In general right now it works much better than before. Great job!

I see.. did not know about this approach of using viewheight for calculating content height. So we are keeping header and footer as is(with static/relative) and content position will be static/relative but with a height of the calculated value; is my understanding correct? What would the `offset` do? Would it be provided as an `@Input` by the user?

I still somehow feel that setting `top` might come into picture when using tabbed content though 😅  Anyway, I will try this approach.

JKMarkowski:
> I see.. did not know about this approach of using viewheight for calculating content height. So we are keeping header and footer as is(with static/relative) and content position will be static/relative but with a height of the calculated value; is my understanding correct? What would the `offset` do? Would it be provided as an `@Input` by the user?
> 
> I still somehow feel that setting `top` might come into picture when using tabbed content though  Anyway, I will try this approach.

I was thinking about offset just in case user wants to put something below the dynamic page. It can be provided by the user with `@Input()`. Just to make it more flexible. 
For tabbed content this should be fine.  

The only issue can be meet with [stacked content](https://fundamental-ngx.netlify.app/#/core/tabs#stacked-content), but user should handle it by himself in this case

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.scss
kavya-b: changed to static as suggested

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
kavya-b: removed usage of top. Thanks for showing me this approach, Jedrzej!

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
kavya-b: removed usage of top

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
kavya-b: removed usage of top

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
kavya-b: this is still needed as it is unrelated to positioning. It adds some necessary modifier classes to the tabs.

JKMarkowski (approved):
Looks much better. Great Job!

InnaAtanasova (approved):
Awesome work!

Issue #3375, opened 2020-09-23
Search Field implementation in Core


InnaAtanasova:
Implement the component in Core

3days

PR #4354, opened 2021-01-16
feat: (Core) add Search Field functionality to Combobox component

InnaAtanasova:
Closes https://github.com/SAP/fundamental-ngx/issues/3375

This PR introduces Search Field enhancement to Combobox component. When the `isSearch` property is set to true, the glyph of the Input Group changes to 'search' icon and also an additional clear button is shown when the user starts typing in the input field. The Clear button can be hidden by setting the `showClearButton` to false. By default it's set to true. 

Documentation checklist:
- [na] update `README.md`
- [na] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/combobox/combobox.component.html
+++ b/libs/core/src/lib/combobox/combobox.component.html
@@ -3,6 +3,7 @@
 <ng-template #desktopTemplate>
     <fd-popover
         additionalBodyClass="fd-popover-custom-list"
+        [class.fd-combobox-full-width]="isSearch"
         [isOpen]="open && displayedValues && displayedValues.length"
         (isOpenChange)="isOpenChangeHandle($event)"
         [fillControlMode]="fillControlMode"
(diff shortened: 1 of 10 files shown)


libs/core/src/lib/combobox/combobox.component.html
```diff
@@ -3,6 +3,7 @@
 <ng-template #desktopTemplate>
     <fd-popover
         additionalBodyClass="fd-popover-custom-list"
+        [class.fd-combobox-full-width]="isSearch"
```
mikerodonnell89, line 6: Shouldn't this be up to the developer? If we are making this decision for them, not sure why it'd only apply to search comboboxes and not all of them

InnaAtanasova: That's a good question. I also thought about it during the implementation. Why I did it this way? Combobox has restrictions when it comes to max and min width, which do not apply on Search Field. Multi Input, for example, also takes the 100% of the container. Making Search Field "full-width" we give the option to the user to apply whatever width they want on the container inside which they gonna place the Search field and this way also remove the css rules that apply on the Compobox when in comes to the width. Does this make sense?

mikerodonnell89: Yes! PR looks good

mikerodonnell89 (approved):

PR #4479, opened 2021-01-26
fix: (Platform) Dynamic Content Page overflow

JKMarkowski:
This is temporary fix for scrollbar, which shouldn't be shown all the time


olshansky (approved):

InnaAtanasova (approved):

droshev (approved):
@JKMarkowski can you bring it to the version eureka is using?

Issue #4344, opened 2021-01-15
Platform table phase 3.5 (page scrolling)


dimamarksman:
Implement platform table page scrolling feature

dimamarksman:
Saturday:
* was working on table page scrolling issue

Today:
 * continue working on table page scrolling issue (table layout changes)

dimamarksman:
Saturday:
 * was working on table page scrolling
Today:
 * continue working on table page scrolling (table layout changes)

dimamarksman:
Today:
 * proceed working on table layout changes (need to rework how we handle group rows and render it in a separate layout)

dimamarksman:
Yesterday
 - worked on table scrolling functionality.

Today
 - about to raise table scrolling PR
 - going to create table component demo

dimamarksman:
[PR](https://github.com/SAP/fundamental-ngx/pull/4497) has been raised.

dimamarksman:
29.01.2021
 * Fixed scroll dispatcher service so horizontal scroll doesn't effect vertical scroll area and vice versa

PR #4497, opened 2021-01-27
feat: (Platform) Table Page Scrolling

dimamarksman:
Closes https://github.com/SAP/fundamental-ngx/issues/4344

This PR introduces table page scrolling feature.

The table layout has been changed significantly in order to get vertical and horizontal scrollbars.  

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


KevinOkamoto (approved):
Looks good.

manjunathanagaraj:
@dimamarksman , I dont see an example for horizontal scrolling.

dimamarksman:
> @dimamarksman ,
> 
> 1. I dont see an example for horizontal scrolling.
> 2. How does the developer control on number of rows to be displayed? I think this as well should appear in documentation, right?

In order to restrict table body height and get the vertical scroll we need to use `bodyHeight` option
image


Horizontal scrollbar appears automaticaly, there is option to enable it. This scrollbar you can see on column freezing example
image

libs/platform/src/lib/components/table/table.component.scss
```diff
@@ -1,42 +1,206 @@
...
-    .fd-table__row--group {
+    .fdp-table__body {
+        position: relative;
+
+        overflow: hidden;
+        overflow-y: scroll;
+
+        .fd-table--fixed {
+            overflow: hidden;
+            overflow-x: scroll;
```
JKMarkowski, line 151: Is there any reason to not having `overflow-x/y  auto`?

dimamarksman: Yeah,
The trick is the parent has only vertical scroll and the child has only horizontal scroll

puru-hk:
Testing the scrolling functionality it is working. But from the accessibility prospective not able to scroll from the keyboard. 
And I have raised some issue regarding the filter

libs/platform/src/lib/components/table/table.component.scss
dimamarksman: I probaly got you wrong.
We have mock scrollbar area in the header and need to track the scrolbar width and set it there in order to have the same header width as the body width. 
So just for simplicity I added `overflow-x: scroll` in order to keep this width as a constant.
In the future we could improve it and add a proper tracking of it.

PR #4511, opened 2021-01-28
fix: (Core|Platform) refactoring select component with  cdk key manager, selection model

### LINKED ISSUE #4425: fix(core): use selectionmodel, key mamanagers cdk in select component
enhancement

Enhance the current code to implement selection key manager cdk usage in select component.

10

sKudum:
closes #4425 

Refactored select component in core with cdks keymanager and selection model.

parallel this PR fixes build issues in platform due to this changes.

BREAKING CHANGE:

Not required 
**extendedBodyTemplate**- can be deprecated- as it is handled internally by select.component, controlTemplate!=null or undefined is considered as extendedBodyTemplate as true.

**Before:**
```
<fd-select [(value)]="selectedValue"  [extendedBodyTemplate]="true" [controlTemplate]="customSelectTemplate">
```

**Now:**
```
<fd-select [(value)]="selectedValue" [controlTemplate]="customSelectTemplate">
```

Documentation checklist:
- [n/a ] update `README.md`
- [x] Documentation Examples


libs/core/src/lib/select/select-key-manager.service.ts
```diff
@@ -0,0 +1,114 @@
...
+export class SelectKeyManagerService {
+
+    _component: any;
+    _keyManager: ActiveDescendantKeyManager<OptionComponent>;
+
+    /**
+    * Sets up a key manager to listen to keyboard events on the overlay panel.
+    * @hidden
+    */
+    _initKeyManager(_component: any): void {
```
JKMarkowski, line 20: Could you pass some interface, instead of any here?

libs/core/src/lib/select/option/option.component.ts
```diff
@@ -1,24 +1,53 @@
...
+        readonly source: OptionComponent,
+        /** Whether the change in the option's value was a result of a user action. */
+        readonly isUserInput = false
+    ) { }
+}
+
+/**
+ * Injection token used to provide the select component to options.
+ */
+export const SELECT_COMPONENT = new InjectionToken<SelectInterface>('SELECT_COMPONENT');
```
JKMarkowski, line 47: Is it still used anywhere?

libs/core/src/lib/select/select-key-manager.service.ts
```diff
@@ -0,0 +1,114 @@
...
+        this._keyManager = new ActiveDescendantKeyManager<OptionComponent>(this._component._options)
```
JKMarkowski, line 21: It would be great to have some interface, instead of `OptionComponent`, just to not restrict this service to select component

libs/core/src/lib/select/select-key-manager.service.ts
```diff
@@ -0,0 +1,114 @@
...
+import { KeyUtil } from '../utils/functions';
+import { UP_ARROW, DOWN_ARROW, ENTER, SPACE, HOME, END, hasModifierKey } from '@angular/cdk/keycodes';
+
+import { OptionComponent } from './option/option.component';
+
+
+@Injectable()
+export class SelectKeyManagerService {
+
+    _component: any;
```
JKMarkowski, line 13: Could you pass some interface, instead of any here?

libs/core/src/lib/select/select.component.ts
```diff
@@ -18,34 +17,55 @@ import {
...
+import { OptionComponent, FdOptionSelectionChange } from './option/option.component';
+import { SelectMobileComponent } from './select-mobile/select-mobile.component';
 
 let selectUniqueId = 0;
 
 export type SelectControlState = 'error' | 'success' | 'warning' | 'information';
 
+const SELECT_HEADER_IDENTIFIER = '.fd-list__group-header';
+
 export interface OptionStatusChange {
```
JKMarkowski, line 53: Please recheck if this is still used

libs/core/src/lib/select/select.component.ts
```diff
@@ -178,166 +194,320 @@ export class SelectComponent implements ControlValueAccessor, SelectInterface, O
...
+    popperOptions: PopperOptions = {
+        placement: 'bottom-start',
+        modifiers: {
+            preventOverflow: {
+                enabled: true,
+                escapeWithReference: true,
+                boundariesElement: 'scrollParent'
+            }
+        }
+    };
```
JKMarkowski, line 210: Feel free to remove it and its usages, `[options]` has been deprecated from popover and will be removed in future

libs/core/src/lib/select/select.component.html
```diff
@@ -1,99 +1,62 @@
...
-</ng-template>
-
-<ng-template #selectMobileTemplate>
-    <ng-container [ngTemplateOutlet]="selectInputControlTemplate"></ng-container>
+<ng-container *ngTemplateOutlet="mobile ? mobileTemplate : desktopTemplate"></ng-container>
+
+<ng-template #desktopTemplate>
+    <div class="fd-select" [class.fd-select--compact]="compact" (keydown)="_handleKeydown($event)" (click)="toggle()">
+        <fd-popover additionalBodyClass="fd-popover-custom-list" [isOpen]="_isOpen" [triggers]="[]"
+            [focusTrapped]="true" [disabled]="disabled" [appendTo]="appendTo" [options]="popperOptions"
```
JKMarkowski, line 6: let's remove `[options]="popperOptions"`

libs/core/src/lib/select/select.component.ts
```diff
@@ -178,166 +194,320 @@ export class SelectComponent implements ControlValueAccessor, SelectInterface, O
...
+    /**@hidden
+     * Referent to the option;s container
+     */
+    @ViewChild('optionPanel', { read: ElementRef })
+    _optionPanel: ElementRef;
+
+      /** Text value displayed
+     * @hidden
+     * in select control */
+    _selectViewValue: string;
```
JKMarkowski, line 253: Could you check if it's still used

libs/core/src/lib/select/select.component.ts
```diff
@@ -178,166 +194,320 @@ export class SelectComponent implements ControlValueAccessor, SelectInterface, O
...
+    _calculatedMaxHeight: number;
 
-    /** @hidden Cashed options as as Array */
-    private _options: OptionComponent[];
+    /** @hidden */
+    _triggerRect: ClientRect;
 
     /** @hidden */
-    private _subscriptions: Subscription = new Subscription();
+    _ngZone: NgZone;
```
JKMarkowski, line 276: Could you check if it's still used?

libs/core/src/lib/select/select.component.ts
```diff
@@ -178,166 +194,320 @@ export class SelectComponent implements ControlValueAccessor, SelectInterface, O
...
+    }
+
+    /** Whether control can be interacted with */
+    get isInteractive(): boolean {
+        return !(this.readonly || this.disabled);
+    }
+
+    get isCancelableMobileSelect(): boolean {
+        return this.mobile && !!this.mobileConfig.approveButtonText
+    }
```
JKMarkowski, line 360: Could you check if it's still used?

libs/core/src/lib/select/select.component.ts
```diff
@@ -178,166 +194,320 @@ export class SelectComponent implements ControlValueAccessor, SelectInterface, O
...
+    set compareWith(fn: (o1: any, o2: any) => boolean) {
+        if (typeof fn !== 'function') {
+            throw Error('compareWith` must be a function.');
+        }
+        this._compareWith = fn;
+        if (this._selectionService._selectionModel) {
+            // A different comparator means the selection could change.
+            this._initializeSelection();
+        }
+    }
```
JKMarkowski, line 399: Could you check if it's still used?

libs/core/src/lib/select/select.component.ts
```diff
@@ -178,166 +194,320 @@ export class SelectComponent implements ControlValueAccessor, SelectInterface, O
...
+
+    get calculatedMaxHeight(): number {
+        return this._maxHeight || this._calculatedMaxHeight;
+    }
+
+
+
     constructor(
+        public viewportRuler: ViewportRuler,
```
JKMarkowski, line 412: Do you need `viewportRuler` here?

libs/core/src/lib/select/select.component.ts
```diff
@@ -178,166 +194,320 @@ export class SelectComponent implements ControlValueAccessor, SelectInterface, O
...
         private _elementRef: ElementRef,
-        private _selectProxy: SelectProxy,
+        @Attribute('tabindex') _tabIndex: string,
+        public dir: Directionality,
+        public _KeySerivce: SelectKeyManagerService,
+        public _selectionService: SelectSelectionModelService,
```
JKMarkowski, line 417: Could you make them `private`? I can't see any usages outside component and tests

JKMarkowski (changes requested):
Hello @sKudum, amazing job :rocket:! This is exactly what we needed :) I added some minor comments regarding code

libs/core/src/lib/select/select.component.ts
sKudum: During selecting value using it.

libs/core/src/lib/select/select.component.ts
sKudum: It is used in selection service test specs.

libs/core/src/lib/select/option/option.component.html
```diff
@@ -1,9 +1 @@
-<ng-template #contentTemplateRef><ng-content></ng-content></ng-template>
-
-<span *ngIf="!_extendedTemplate" fd-list-title>
-    <ng-container *ngTemplateOutlet="contentTemplateRef"></ng-container>
-</span>
-
-<ng-container *ngIf="_extendedTemplate">
-    <ng-container *ngTemplateOutlet="contentTemplateRef"></ng-container>
-</ng-container>
+<ng-content></ng-content>
```
fkolar, line 1: Why are we removing this part?  Are we removing customization item level with custom template

fkolar, line 1: Why are we removing this part? Are we removing customization item level with custom template

libs/core/src/lib/select/select.component.ts
```diff
@@ -58,15 +72,22 @@ export interface OptionStatusChange {
         '[class.fd-select-custom-class--mobile]': 'mobile',
     },
     providers: [
-        SelectProxy,
         {
-            provide: NG_VALUE_ACCESSOR,
-            useExisting: forwardRef(() => SelectComponent),
-            multi: true
-        }
+            provide: SELECT_COMPONENT,
```
fkolar, line 76: What is this? you cant really use useExisting here. You are creating a provider which will not work with Forms. there is a reason why its called NG_VALUE_ACCESOR. it is used internally by form API

fkolar (changes requested):

libs/core/src/lib/select/option/option.component.ts
```diff
@@ -63,97 +88,147 @@ export class OptionComponent implements OnInit, OnDestroy {
...
-        private _elementRef: ElementRef,
-        private _selectProxy: SelectProxy,
-        private _changeDetRef: ChangeDetectorRef
-    ) {}
+    private _active = false;
 
     /** @hidden */
-    ngOnInit(): void {
-        this._observeValue();
+    constructor(private _elementRef: ElementRef, private _changeDetectorRef: ChangeDetectorRef) { }
```
dimamarksman, line 121:
```suggestion
    constructor(private _elementRef: ElementRef<HTMLElement>, private _changeDetectorRef: ChangeDetectorRef) { }
```

libs/core/src/lib/select/option/option.component.ts
```diff
@@ -63,97 +88,147 @@ export class OptionComponent implements OnInit, OnDestroy {
...
+        return this.viewValue;
     }
 
-    /** Returns HTMLElement representation of the component. */
-    getHtmlElement(): HTMLElement {
+    /** @hidden
+     * Returns HTMLElement representation of the component. */
+    _getHtmlElement(): HTMLElement {
         return this._elementRef.nativeElement as HTMLElement;
     }
```
dimamarksman, line 190: Looks like unused method

libs/core/src/lib/select/select.component.ts
```diff
@@ -178,166 +191,292 @@ export class SelectComponent implements ControlValueAccessor, SelectInterface, O
...
+    get calculatedMaxHeight(): number {
+        return this._maxHeight || this._calculatedMaxHeight;
     }
 
     constructor(
         private _elementRef: ElementRef,
-        private _selectProxy: SelectProxy,
+        @Attribute('tabindex') _tabIndex: string,
+        public dir: Directionality,
+        public _KeySerivce: SelectKeyManagerService,
```
dimamarksman, line 385: should be `_keyManagerSerivce`

libs/core/src/lib/select/select-selection-model.service.ts
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    /**
+    * Returns true if select has any value selected
+    */
+    clear(): void {
+        this._selectionModel.clear();
+    }
+
+}
```
dimamarksman, line 24: What is the benefit of moving selection model in this service?
This service does nothing and exposes `SelectionModel` anyway.

libs/core/src/lib/select/select.interface.ts
```diff
@@ -1,8 +1,12 @@
...
 import { OptionComponent } from './option/option.component';
-import { OptionStatusChange } from './select.component';
+import { FdSelectChange } from './select.component';
 import { DialogConfig } from '../dialog/utils/dialog-config.class';
 import { MobileMode } from '../utils/interfaces/mobile-control.interface';
+import { MobileModeConfig } from '../utils/public_api';
+import { OptionsInterface } from './options.interface';
+import { LiveAnnouncer } from '@angular/cdk/a11y';
 
 export const SELECT_COMPONENT = new InjectionToken<string[]>('SelectInterface');
```
dimamarksman, line 11: Why `string[]` and not `new InjectionToken<SelectInterface>('SelectInterface')`?

libs/core/src/lib/select/select.component.ts
dimamarksman: We probbale need to keep both providers: `NG_VALUE_ACCESOR` as well as `SELECT_COMPONENT`

libs/core/src/lib/select/select.component.ts
```diff
@@ -178,166 +191,292 @@ export class SelectComponent implements ControlValueAccessor, SelectInterface, O
...
+    /** @hidden */
+    _internalValue: any;
 
     /** @hidden */
-    onChange: Function = () => {};
+    _calculatedMaxHeight: number;
 
     /** @hidden */
-    onTouched: Function = () => {};
+    _triggerRect: ClientRect;
```
dimamarksman, line 252: Where is it used?

libs/core/src/lib/select/select.component.ts
```diff
@@ -351,230 +490,271 @@ export class SelectComponent implements ControlValueAccessor, SelectInterface, O
...
-            )
-        })
+    _highlightCorrectOption(): void {
+        if (this._KeySerivce._keyManager) {
+            if (this._selectionService.empty) {
+                this._KeySerivce._keyManager.setFirstItemActive();
+            } else {
+                this._KeySerivce._keyManager.setActiveItem(this.selected);
+            }
+        }
```
dimamarksman, line 563: let's rid of nested `If`

dimamarksman (changes requested):
God job,
Got several questions

libs/core/src/lib/select/select.component.ts
sKudum: Added to avoid naming conflicts in side service and component.. used these names.

libs/core/src/lib/select/select-selection-model.service.ts
sKudum: keyManager and selection models are provided as service with commonly used functions.
By targeting future  enhancements if any and for other components for reusability.

libs/core/src/lib/select/option/option.component.html
sKudum: Yes Frank, as Custom template provides more flexibility.

sKudum: Yes Frank, as Custom template provides more flexibility.

libs/core/src/lib/select/select.component.ts
sKudum: As we added selection model, and keyManagers as services adding NG_VALUE_ACCESOR and usage of ngControl  to display preselected values are resulting to cycled dependency , ng Control inject errors. To avoid that, I have provided SELECT_COMPONENT.

libs/core/src/lib/select/select.component.ts
dimamarksman: Sorry, not sure if I got you right.
Which one name conflicts do you mean?
Can we avoid using uppercase from the begining?

libs/core/src/lib/select/select.component.ts
JKMarkowski: I see, in spec tests you can use `any`  casting, like [here](https://github.com/SAP/fundamental-ngx/blob/fb48296246fa53d6578f87017351e6dfa91ea47c/libs/core/src/lib/alert/alert.component.spec.ts#L61). Could you apply it?

libs/platform/src/lib/components/select/select.component.ts
```diff
@@ -68,19 +81,25 @@ export class SelectPlatformComponent extends fdSelect implements OnInit, AfterCo
             this._value = newValue;
             this.onChange(newValue);
             this.onTouched();
-            this.valueChange.emit(newValue);
+           // this.valueChange.emit(newValue);
```
JKMarkowski, line 84: Commented code

libs/core/src/lib/select/select-selection-model.service.ts
dimamarksman: In my understanding the main purpose of this service should be incapsulation.
If we want to incapsulate `SelectionModel` we need to hide direct reference and define public api (actually I don't see much sense in it cause `SelectionModel` already incapsulated well by itself)

In case if you want to extract some selection component logic I don't see any in this service.
As for me it's useless.

By the way `selection.clear` method is not used at all. All manipulations with selectionModel goes directly through reference to the `_selectionModel` itself.

libs/core/src/lib/select/select.component.ts
dimamarksman: @sKudum could you please comment from your account name not editing my comments

dimamarksman: Otherwise it looks like I talking to myself :)

libs/core/src/lib/select/select-selection-model.service.ts
sKudum: With respect to selectionmodel I agree with you, 

 @JKMarkowski could you please share your thoughts on it, I can see Dimitry also sharing same concern as I did earlier.

libs/core/src/lib/select/option/option.component.ts
dimamarksman: So it's used in specs, got it.
In this case no need for `as` cause _elementRef already has type reference

libs/core/src/lib/select/select.component.ts
sKudum: :-), where ever message thread have gaps or hidden am trying to give reply there itself. So that it will be easy to track. Will update as you mentioned.

libs/core/src/lib/select/select.component.ts
```diff
@@ -178,166 +192,292 @@ export class SelectComponent implements ControlValueAccessor, SelectInterface, O
...
+
+    get calculatedMaxHeight(): number {
+        return this._maxHeight || this._calculatedMaxHeight;
     }
 
     constructor(
         private _elementRef: ElementRef,
-        private _selectProxy: SelectProxy,
+        @Attribute('tabindex') _tabIndex: string,
+        public dir: Directionality,
```
JKMarkowski, line 385: How does it work - it somehow detects the actual direction? I would encourage to use current `RtlService`. We can decide about migrating to `Directionality` later

libs/core/src/lib/select/select.component.ts
dimamarksman: So we need to fix somehow this `cycled dependency`.
We can't drop `NG_VALUE_ACCESOR` provider cause it wont work with angular forms

sKudum: Please suggest on this, 
I tried couple of ways like creating instance of ngControl, had both select_component and ng_value_accessor, tried only having ng_value_accessor etc and ended up with infinite error loops.

dimamarksman: Oh, I see.
So we don't need this provider since we have the linking in the constructor as :
```
constructor(@Optional() @Self() public ngControl: NgControl) {
        if (this.ngControl) {
            this.ngControl.valueAccessor = this;
        }
    }
```

fkolar: I missed this one, as we started to the same thing we have in platform code to connect it manually to forms api . thanks.

fkolar (approved):

libs/core/src/lib/select/select-selection-model.service.ts
sKudum: Hi @JKMarkowski , as discussed we won't be having separate service for selectioonmodel

libs/core/src/lib/select/select.component.ts
```diff
@@ -178,166 +191,293 @@ export class SelectComponent implements ControlValueAccessor, SelectInterface, O
...
-        private _selectProxy: SelectProxy,
+        @Attribute('tabindex') _tabIndex: string,
+        @Optional() private _rtlService: RtlService,
+        private _keyManagerService: SelectKeyManagerService,
         private _changeDetectorRef: ChangeDetectorRef,
         @Optional() private _dynamicComponentService: DynamicComponentService,
-        @Optional() public dialogConfig: DialogConfig
-    ) { }
+        @Optional() public dialogConfig: DialogConfig,
+        @Optional() @Self() public ngControl: NgControl,
```
JKMarkowski, line 392: Do the they have to be public?

dimamarksman (approved):

JKMarkowski (approved):
LGTM

PR #4638, opened 2021-02-10
fix: expose avatar inputs as part of card header

droshev:
The avatar in card header doesn't allow setting icons and different sizes

BREAKING CHANGE:
The avatar in card header needs to set size

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/card/card-header.component.ts
+++ b/libs/core/src/lib/card/card-header.component.ts
@@ -1,11 +1,13 @@
 import {
     Component,
     OnInit,
+    AfterContentInit,
     ChangeDetectionStrategy,
     ElementRef,
     ContentChild,
     Input,
-    HostBinding
+    HostBinding,
+    Renderer2
 } from '@angular/core';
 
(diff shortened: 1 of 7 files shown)


InnaAtanasova (approved):

libs/core/src/lib/card/card-header.component.ts
```diff
@@ -37,13 +39,22 @@ export class CardHeaderComponent implements OnInit, CssClassBuilder {
...
     }
 
+    /** @hidden */
+    ngAfterContentInit(): void {
+        /** Add fd-card__avatar class to fd-avatar */
+        const avatar = this.elementRef().nativeElement.querySelector('fd-avatar');
+        if (avatar) {
+            this.renderer.addClass(avatar, 'fd-card__avatar');
+        }
+    }
```
JKMarkowski, line 56: What about not having query selector, but `@ContentChild`? So reference to avatar class is available and `s` size can be added, when it's missing.

JKMarkowski (approved):

PR #4693, opened 2021-02-15
docs: (Core) refactor dialog example to remove box shadow issue

### LINKED ISSUE #4692: Wizard in dialog - refactor so that progress bar should not have box-shadow on top
Edit: The designers (Kalin) need to be contacted for the design of the Wizard in Dialog and we should decide if we will implement it using the Dynamic Page
Sprint 55 - set up a meeting with the designers and investigate the design

The wizard progress bar currently has a box-shadow on top, resulting in this when placed in a dialog:
Screen Shot 2021-01-25 at 10 14 01 AM

Should look like this:
Screen Shot 2021-01-25 at 10 15 22 AM

mikerodonnell89:
fixes #4692 

Moves the wizard title into a dialog header and adds some custom styles to remove the unwanted box-shadow between the title and the progress bar.


InnaAtanasova (approved):
Looks good

JKMarkowski (approved):

PR #4716, opened 2021-02-18
fix: (Core) Non interactive cards

JKMarkowski:
There is added non-interactive cards example with necessary styles added

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/apps/docs/src/app/core/component-docs/card/card-docs.component.html
+++ b/apps/docs/src/app/core/component-docs/card/card-docs.component.html
@@ -106,7 +106,7 @@
 <description>
     In order to display bar chart list in a card we can use list styles along with a custom bar component.
     For this example in order to display bar chart we created a simple <code>fd-card-bar</code> component.
-    For card with list t is not recommended to keep header interactive, or navigable. Such a header can be achieved
+    For card with list it is not recommended to keep header interactive, or navigable. Such a header can be achieved
     by adding <code>[interactive]="false"</code> to <code>fd-card-header</code> element.
 </description>
 
--- a/apps/docs/src/app/core/component-docs/card/card-docs.component.html
+++ b/apps/docs/src/app/core/component-docs/card/card-docs.component.html
@@ -106,6 +106,8 @@
 <description>
     In order to display bar chart list in a card we can use list styles along with a custom bar component.
     For this example in order to display bar chart we created a simple <code>fd-card-bar</code> component.
+    For card with list t is not recommended to keep header interactive, or navigable. Such a header can be achieved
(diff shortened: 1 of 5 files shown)


droshev (approved):

libs/core/src/lib/card/card-header.component.ts
```diff
@@ -21,11 +23,15 @@ import { CardSubtitleDirective } from './card-subtitle.directive';
...
+export class CardHeaderComponent implements OnInit, OnChanges, CssClassBuilder, AfterContentInit {
+    /** Whether card header is interactive */
+    @Input()
+    interactive = true;
+
     /** @hidden */
     @Input()
     @HostBinding('attr.tabindex')
-    tabindex = '0';
+    tabindex = this.interactive ? '0' : '-1';
```
dimamarksman, line 34: If `interactive="false"` should take precedence over `tabindex` input it might be better to have `tabindex` as `getter setter`.
```
@Input()
@HostBinding('attr.tabindex')	
set tabindex(tabindex: string) {
  this._tabindex =  tabindex;
}
get tabindex(): string {
  return !this.interactive ? '-1' : this._tabindex;
}
private _tabindex = '0';
```
Otherwise the user is able to override `tabindex` when `interactive="false"` just assigning tabindex  later.

If `tabindex` must be always based on `interactive` value then we shouldn't have tabindex as `input`;

@JKMarkowski  what do you think?

JKMarkowski: Hi @dimamarksman you are totally right there, I will add setters/getters, just to let user decide whether this should be tabbable

dimamarksman: In this case we don't need this assinment in ngOnChanges:
```
if ('interactive' in changes) {
   this.tabindex = this.interactive ? '0' : '-1' ;
}
```

apps/docs/src/app/core/component-docs/card/card-docs.component.html
```diff
@@ -106,9 +106,11 @@
 <description>
     In order to display bar chart list in a card we can use list styles along with a custom bar component.
     For this example in order to display bar chart we created a simple <code>fd-card-bar</code> component.
+    For card with list t is not recommended to keep header interactive, or navigable. Such a header can be achieved
```
dimamarksman, line 109:
```suggestion
    For card with list it is not recommended to keep header interactive, or navigable. Such a header can be achieved
```

InnaAtanasova (approved):

dimamarksman (approved):

PR #4749, opened 2021-02-22
fix: selecting multi-input option should clear search term

### LINKED ISSUE #4565: [Bug]: Multi Input Control - search term not being removed after selection

Bug

Expected behaviour (search string is being removed after selection and focus out (ESC, click outside, etc.)):
MultiInputControl

v0.26.0

Currently, the search string the user typed in remains in the input field after selection and focusing out:
Bug

Time estimation: 1 day

mikerodonnell89:
fixes #4565 

Resets the search term when the user makes a selection or the input loses focus


libs/core/src/lib/multi-input/multi-input.component.ts
```diff
@@ -379,6 +382,14 @@ export class MultiInputComponent implements
             this.popoverRef.refreshPosition();
         }
 
+        this._resetSearchTerm();
+
+        if (this.popoverRef) {
+            this.popoverRef.close();
+        }
+
+        this.searchInputElement.nativeElement.focus();
```
dimamarksman, line 391: Why do you close the popover once an option gets selected?
Shouldn't we allow multi-selection?

dimamarksman (changes requested):
Please see my comment.
Also it would be great to cover this by a unit test

libs/core/src/lib/multi-input/multi-input.component.ts
mikerodonnell89: I was going off of sapui5's behavior, but looking at it again today, it appears as though they keep the popover open if you click one of the checkboxes but close it if you click any other part of the list item.  I doubt that is actually the intended behavior so I will remove this line

dimamarksman (approved):

InnaAtanasova (approved):

JKMarkowski (approved):

PR #4758, opened 2021-02-23
feat(store): Entity Query Interface

### LINKED ISSUE #4584: Query Interface
Implement basic predicate specification to be able to translate expression into AST.

KevinOkamoto:
Closes #4584

This PR contains the query API for the state management solution as described here:

https://app.gitbook.com/@fundamental-ngx/s/state-management/design/entityquery-persistence


libs/store/src/lib/domain/query/query-builder.ts
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * This could be part of some common functionality shared with EntityStore where we
+ * create entity store out of type.
+ *
+ * This is where my original unitof works could jump in. a usecase  => unitof work untill its completed.
+ * all is cached as part of this.
+ *
+ *
+ */
+// export function newQueryBuilder < TModel > (resultType: Type < TModel > ): QueryBuilder < TModel > {
```
fkolar, line 26: Let' delete these old leftovers.

libs/store/src/lib/domain/query/query-builder.ts
```diff
@@ -0,0 +1,88 @@
...
+     * Add keyword to builder.
+     * @param keyword Keyword for search
+     */
+    keyword(keyword: string): QueryBuilder<TModel> {
+        this._keyword = keyword;
+        return this;
+    }
+
+    /*
+    newSubQuery <TSubModel> (withId: string, resultType: Type < TSubModel > ): QueryBuilder < TModel > {
```
fkolar, line 73: Let's delete this one too

libs/store/src/lib/domain/query/query.ts
```diff
@@ -0,0 +1,243 @@
...
+     */
+    expand<TP extends keyof TModel> (...expand: Array<TP> ): Query<TModel> {
+        this._expand = expand as string[];
+        return this;
+    }
+
+    /**
+     * Initiate query and return observable
+     */
+    fetch(): Observable<TModel | Array<TModel>> {
```
fkolar, line 148: is this our original .find() ?

libs/store/src/lib/domain/query/query.ts
```diff
@@ -0,0 +1,243 @@
...
+    _keyword: string;
+
+    /** @hidden - stores current predicate */
+    _predicate: Predicate<TModel>;
+
+    /** @hidden - stores current page size */
+    _pageSize: number;
+
+    /** @hidden - stores current index offset */
+    _offset: number;
```
fkolar, line 36: What if we name these internal properties to match what you ahve in tests and what is also in Ariba REST API? 

top, skip

libs/store/src/lib/domain/query/query.ts
```diff
@@ -0,0 +1,243 @@
...
+    _suppressPageReset: boolean;
+
+    /** @hidden - stores curernt order bys */
+    _orderByFields: Array<OrderBy<TModel, keyof TModel>>;
+
+    /** @hidden - stores current enable count flag */
+    _includeCount: boolean;
+
+    /** @hidden - stores current selection of properties */
+    _select: string[];
```
fkolar, line 48: We should work here with typings too. 
Array<<TModel, keyof TModel> or something like this..

fkolar (changes requested):
Looks great. Only few minor comments and questions. 

* Also lets move domain/query under persistence/query. 
* Domain will contains mostly some base classes fro BaseEntity, Value (ValueObject), decorators, and in the future some additional things. 

Persistence layer contains both Private API, and client facing API all related to persistence. @dimamarksman will also have its folder persistence/store,

libs/store/src/lib/domain/query/query.ts
dimamarksman, line 148: Doesn't `this.service.getWithQuery(query)` always return entity collection?

libs/store/src/lib/domain/query/query.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    /** @hidden - stores current expand properties */
+    _expand: string[];
+
+    constructor(
+        private resultType: Type<TModel>,
+        private service: EntityCollectionService<TModel>,
```
dimamarksman, line 55: In order not to have a tight coupling with internal details we could abstract this `service` to something that has one method to fetch (`getWithQuery`) and returns `Observable<TModel[]>`.
As a benefit, it becomes much easier to test this class and we don't need `TestBed` stuff in the spec file.

dimamarksman (changes requested):
Looks great!
Have several questions, please take a look.

libs/store/src/lib/domain/query/grammer/predicate.ts
```diff
@@ -0,0 +1,108 @@
+export interface Predicate < TModel > {
+    operands: Predicate < TModel > | Array < Predicate < TModel >> ;
+    readonly query: string;
+
+    and(...operands: Array < Predicate < TModel >> ): Predicate < TModel > ;
```
dimamarksman, line 5: We can use `this` as returned type:
```
and(...operands: Array < Predicate < TModel >> ): this ;
```

libs/store/src/lib/domain/query/query.ts
```diff
@@ -0,0 +1,243 @@
...
+    orderBy<TProperty extends keyof TModel> (...orderBys: Array<OrderBy<TModel, TProperty>> ): Query < TModel > {
+        this._orderByFields = orderBys;
+        return this;
+    }
+
+    /**
+     * Set first index of result set for paging.
+     * @param offset Index number of first result.
+     */
+    firstResult(offset: number): Query<TModel> {
```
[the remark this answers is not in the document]
fkolar: I see now also some addition point. we should call withFirstResult, withMaxresult to be more fluent. I did not noticed this one in my result

libs/store/src/lib/domain/query/grammer/predicate.ts
```diff
@@ -0,0 +1,108 @@
...
+        return '';
+    }
+}
+
+export class NotPredicate < TModel > extends BasePredicate < TModel > {}
+
+export class ComparisonPredicate < TModel, TProperty extends keyof TModel, TPropertyValue extends TModel[TProperty] >
+    extends BasePredicate < TModel > {
+
+        constructor(public readonly property: TProperty, value: TPropertyValue) {
```
dimamarksman, line 44:
```suggestion
        constructor(public readonly property: TProperty, public readonly value: TPropertyValue) {
```

libs/store/src/lib/domain/query/grammer/predicate.ts
```diff
@@ -0,0 +1,108 @@
...
+            super();
+        }
+    }
+
+export class EqPredicate < TModel, TProperty extends keyof TModel, TPropertyValue extends TModel[TProperty] >
+    extends ComparisonPredicate < TModel, TProperty, TPropertyValue > {
+
+        constructor(public readonly property: TProperty, public readonly value: TPropertyValue) {
```
dimamarksman, line 52: we don't need these modifiers since superclass already keep them for us

libs/store/src/lib/domain/query/query.ts
KevinOkamoto: Yes. I changed it to "fetch" because I thought it sounded better. When I see "find" I think of the Javascript array "find" method.

KevinOkamoto: If anyone has a differing opinion, I can change it back.

fkolar: No. It looks good.

libs/store/src/lib/domain/query/query.ts
KevinOkamoto: I can rename to "top" and "skip".

libs/store/src/lib/domain/query/query.ts
KevinOkamoto: yes you're correct. It should just be `Array<TModel>>`

libs/store/src/lib/domain/query/grammer/predicate.ts
```diff
@@ -0,0 +1,108 @@
...
+    operands: Predicate < TModel > | Array < Predicate < TModel >> ;
+    readonly query: string;
+
+    and(...operands: Array < Predicate < TModel >> ): Predicate < TModel > ;
+
+    or(...operands: Array < Predicate < TModel >> ): Predicate < TModel > ;
+
+    not(operand: Predicate < TModel > ): Predicate < TModel > ;
+
+    test(target: TModel): boolean;
```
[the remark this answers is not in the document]
KevinOkamoto: I didn't implement the Predicate logic for the Query interface. I'm just using the typing and structure of the Predicate to construct the query strings. 

Can I leave these methods here for future development?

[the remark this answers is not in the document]
dimamarksman: Sure, thanks for the explanation

dimamarksman (approved):

KevinOkamoto:
Now that I've created a `QueryService` I'm thinking of restructuring the code to move all the URL query string generation into the `QueryService`. I would then change the method `QueryService.getWithQuery(query: string)` to `QueryService.getWithQuery(query: QueryParams)` where `QueryParams` would be redefined:

```
interface QueryParams<T> {
  keyword: string;
  filter: Predicate;
  skip: number;
  top: number;
  select: <Array<keyof T>>
  ...
}
```

This way we could replace the `DefaultQueryService` which is REST specific, with another kind of service e.g. GraphQL or even a service which operates on a local array of data.

This will probably take me an additional day to implement with unit tests. Do think a) this is a worthwhile idea b) do it in this PR or defer it till later.

fkolar (approved):

Issue #4755, opened 2021-02-23
Implement Entity store, Entity store server


manjunathanagaraj:
Implement Entity Store

https://app.gitbook.com/@fundamental-ngx/s/state-management/design/entitystore

PR #4771, opened 2021-02-24
feat: (Store) Entity Store API

dimamarksman:
Closes: https://github.com/SAP/fundamental-ngx/issues/4755
This PR contains entity store API:
 - Entity Store
 - Entity Store builder
 - Entity Store Server
 - HTTP URL Generator
 - Basic entity decorators
Documentation https://app.gitbook.com/@fundamental-ngx/s/state-management/design/entitystore

Documentation checklist:
- [ ] update `README.md`
- [ ] Documentation Examples
- [ ] Stackblitz works for all examples


libs/store/src/lib/infrastructure/persistence/store/entity-store-server/entity-store-server.ts
```diff
@@ -0,0 +1,264 @@
...
+export declare type HttpMethods = 'DELETE' | 'GET' | 'POST' | 'PUT' | 'PATCH';
+
+export interface RequestData {
+    method: HttpMethods;
+    url: string;
+    data?: any;
+    options?: any;
+}
+
+/**
```
fkolar, line 22: We dont need additional level `store/entity-store-server/` I would probably prefer instead of having this folder I would just use /store

fkolar: if you feel you have many files, you can move generators and builder to /store/<builder | rest | memory>

libs/store/src/lib/infrastructure/persistence/store/entity-store-server/entity-store.ts
```diff
@@ -0,0 +1,94 @@
...
+export interface EntityStore<T> {
+    /**
+     * Query builder reference.
+     * Use it to create a query to underlying entity collection
+     */
+    readonly queryBuilder: any;
+    /**
+     * Get collection of entities
+     */
+    getAll(): Observable<T[]>;
```
fkolar, line 23: We should hide getAll, as to fetch more records which probably will require some pagination, you need Query, if this is used by other methods we should not expose this to user

libs/store/src/lib/infrastructure/persistence/store/entity-store-server/entity-store.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Get collection of entities that match given query
+     */
+    getWithQuery(query: QueryParams): Observable<T[]>;
```
fkolar, line 28: getWithQuery should be internal API that is used by Query, but user will not use this really

libs/store/src/lib/infrastructure/persistence/store/entity-store-server/entity-store.ts
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * Get entity by id
+     * @param id identity key
+     */
+    getBy(id: IdentityKey): Observable<T>;
```
fkolar, line 33: how about we rename getBy to get, to make easier to read. 
When I see together getAll(), and getBy() it does not feel right.

libs/store/src/lib/infrastructure/persistence/store/entity-store-server/entity-store-builder.ts
```diff
@@ -0,0 +1,116 @@
...
+    useCachePolicy(policy: CachePolicy): this;
+    /**
+     * Apply Fetch Policy
+     * @param policy FetchPolicy settings
+     */
+    useFetchPolicy(policy: FetchPolicy): this;
+    /**
+     * Created store will not be persisted
+     */
+    isTransient(): this;
```
fkolar, line 27: Let's remove this `isTransient` as we dont have use for it yet. The idea behind this, was that I configure EntityStore Server and tell this EntityStore is Transient. Meaning it does not store anything, its not cached. etc.. 

this was related to one of our discussion we had. EntityStore<XXX> will maintain some local map /cache and it will be aware of what entity you currently retrieve using `.getBy` . But I am not sure about it now..

fkolar (changes requested):

libs/store/src/lib/infrastructure/persistence/store/entity-store-server/http-url-generator.ts
```diff
@@ -0,0 +1,84 @@
...
+        pluralize = false
+    ): string {
+        let path: string = pluralize ? this.pluralizer.pluralize(entityName) : entityName;
+
+        if (entityPath) {
+            path = this.getEntityOperationPath(entityPath, operation) || path;
+        }
+
+        return `${normalize(root)}/${normalize(path)}/`.toLowerCase();
+    }
```
KevinOkamoto, line 61: Are we going to support subresource paths, where there is a parent id in the URL like `GET /requisitions/123/lineitems`? If so, won't we need to pass in the entity itself as input?

KevinOkamoto (commented):
Overall it looks good. I just had one comment about URL generation.

libs/store/src/lib/infrastructure/persistence/store/entity-store-server/http-url-generator.ts
dimamarksman: Yes, you are right, we need it.
Can we add this functionality a little bit later?

libs/store/src/lib/infrastructure/persistence/store/entity-store-server/entity-store-builder.ts
dimamarksman: changed

libs/store/src/lib/infrastructure/persistence/store/entity-store-server/entity-store.ts
dimamarksman: done

libs/store/src/lib/infrastructure/persistence/store/entity-store-server/entity-store.ts
dimamarksman: removed

libs/store/src/lib/infrastructure/persistence/store/entity-store-server/entity-store.ts
dimamarksman: removed

libs/store/src/lib/infrastructure/persistence/store/entity-store-server/entity-store-server.ts
dimamarksman: removed `entity-store-server` folder

libs/store/src/lib/infrastructure/persistence/store/entity-store-server/http-url-generator.ts
fkolar: I think this is one of the core functionality we need to support. but I aggree to get moving, we can add this as next task. not really later but more next PR

fkolar (approved):

PR #4777, opened 2021-02-25
feat: (Platform) Table Phase 4

### LINKED ISSUE #4429: Platform Table phase 4 implementation
Phase 4 of the Table component introduces the Tree Table feature.

For details, refer the below technical specs link.
Technical Specs: https://app.gitbook.com/@fundamental-ngx/s/docs/platform-table-phase-4

IlyaSurmay:

Closes #4429.

This PR contains implementation on Phase 4 functionality of Tree table and Drag & Drop support.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

##### Note for the reviewers
Should not be tested for the tree table (as will be implemented in the next phase/separate ticker):
* Sorting rows.
* Filtering rows.
* Grouping rows.
* Frozen columns alongside Drag & Drop.
* Navigation over the three with the keyboard: for now you only can `tab` between actionable elements. Native navigation with the arrows will be implemented separately to handle all the tree structures.

### DIFF
--- a/apps/docs/src/app/platform/component-docs/platform-table/platform-table-examples/platform-table-tree-example.component.ts
+++ b/apps/docs/src/app/platform/component-docs/platform-table/platform-table-examples/platform-table-tree-example.component.ts
@@ -1,0 +1,339 @@
+import { Component } from '@angular/core';
+import { Observable, of } from 'rxjs';
+
+import { FdDate } from '@fundamental-ngx/core';
+import {
+    TableDataSource,
+    TableDataProvider,
+    TableState,
+    TableRowSelectionChangeEvent
+} from '@fundamental-ngx/platform';
+
+@Component({
+    selector: 'fdp-platform-table-tree-example',
+    templateUrl: './platform-table-tree-example.component.html'
+})
+export class PlatformTableTreeExampleComponent {
+    source: TableDataSource<ExampleItem>;
+
+    constructor() {
+        this.source = new TableDataSource(new TableDataProviderExample());
+    }
+
+    alert(message: string): void {
+        alert(message);
+    }
+
+    onRowSelectionChange(event: TableRowSelectionChangeEvent<ExampleItem>): void {
+        console.log(event);
+    }
+}
+
+interface BaseItem {
+    name: string;
+}
+
+export interface ExampleItem extends BaseItem {
+    name: string;
+    description?: string;
+    price?: {
+        value: number;
+        currency: string;
+    };
+    status?: string;
+    statusColor?: string;
+    date?: FdDate;
+    verified?: boolean;
+    children?: ExampleItem[]
+}
+
+/**
+ * Table Data Provider Example
+ *
+ */
+export class TableDataProviderExample extends TableDataProvider<ExampleItem> {
+    items: ExampleItem[] = [...ITEMS];
+    totalItems = ITEMS.length;
+
+    fetch(tableState: TableState): Observable<ExampleItem[]> {
+        this.items = [...ITEMS];
+
+        // apply searching
+        if (tableState.searchInput) {
+            this.items = this.search(tableState);
+        }
+
+        this.totalItems = this.items.length;
+
+        return of(this.items);
+    }
+
+    private search({ searchInput, columns }: TableState): ExampleItem[] {
+        const items = this.items;
+        const searchText = searchInput?.text || '';
+        const keysToSearchBy = columns;
+
+        if (searchText.trim() === '' || keysToSearchBy.length === 0) {
+            return items;
+        }
+
+        return items.filter((item) => {
+            const valuesForSearch = keysToSearchBy.map((key) => getNestedValue(key, item));
+            return valuesForSearch
+                .filter((value) => !!value)
+                .map((value): string => value.toString())
+                .some((value) => value.toLocaleLowerCase().includes(searchText.toLocaleLowerCase()));
+        });
+    }
+}
+
+function getNestedValue<T extends {}>(key: string, object: T): any {
+    return key.split('.').reduce((a, b) => a[b], object);
+}
+
+// Example items
+const ITEMS: ExampleItem[] = [
+    {
+        name: 'Laptops',
+        children: [
+            {
+                name: 'Astro Laptop 1516',
+                description: 'pede malesuada',
+                price: {
+                    value: 489.01,
+                    currency: 'EUR'
+                },
+                status: 'Out of stock',
+                statusColor: 'negative',
+                date: new FdDate(2020, 2, 5),
+                verified: true
+            },
+            {
+                name: 'Benda Laptop 1408',
+                description: 'suspendisse potenti cras in',
+                price: {
+                    value: 243.49,
+                    currency: 'CNY'
+                },
+                status: 'Stocked on demand',
+                statusColor: 'informative',
+                date: new FdDate(2020, 9, 22),
+                verified: true
+            }
+        ]
+    },
+    {
+        name: 'Screens',
+        children: [
+            {
+                name: 'Bending Screen 21HD',
+                description: 'nunc nisl duis bibendum',
+                price: {
+                    value: 66.46,
+                    currency: 'EUR'
+                },
+                status: 'Available',
+                statusColor: 'positive',
+                date: new FdDate(2020, 8, 14),
+                verified: false
+            },
+            {
+                name: 'Broad Screen 22HD',
+                description: 'ultrices posuere',
+                price: {
+                    value: 458.18,
+                    currency: 'CNY'
+                },
+                status: 'Available',
+                statusColor: 'positive',
+                date: new FdDate(2020, 5, 4),
+                verified: true
+            },
+            {
+                name: 'Ergo Screen E-I',
+                description: 'massa quis augue luctus tincidunt',
+                price: {
+                    value: 387.23,
+                    currency: 'NZD'
+                },
+                status: 'Stocked on demand',
+                statusColor: 'informative',
+                date: new FdDate(2020, 3, 23),
+                verified: true
+            },
+            {
+                name: 'Ergo Screen E-II',
+                description: 'orci eget',
+                price: {
+                    value: 75.86,
+                    currency: 'EUR'
+                },
+                status: 'No info',
+                date: new FdDate(2020, 3, 20),
+                verified: false
+            }
+        ]
+    },
+    {
+        name: 'Other 1',
+        children: [
+            {
+                name: '10 Portable DVD player',
+                description: 'diam neque vestibulum eget vulputate',
+                price: {
+                    value: 66.04,
+                    currency: 'CNY'
+                },
+                status: 'Stocked on demand',
+                statusColor: 'informative',
+                date: new FdDate(2020, 1, 7),
+                verified: true
+            },
+            {
+                name: 'Astro Phone 6',
+                description: 'penatibus et magnis',
+                price: {
+                    value: 154.1,
+                    currency: 'IDR'
+                },
+                status: 'Stocked on demand',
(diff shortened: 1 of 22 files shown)


libs/platform/src/lib/utils/object.ts
```diff
@@ -1,3 +1,13 @@
 export function getNestedValue(key: string, object: any): any {
-    return key.split('.').reduce((a, b) => a[b], object);
+    return key
+        .split('.')
+        .reduce((a, b, i, arr) => {
+            // Break loop if we cannot dive inside
+            if (!a[b]) {
+                // Mutating original array will break reduce
+                arr.pop();
+            }
```
dimamarksman, line 9: For what is it?
If there is no `b` in `a` we can return `undefined` without array modification.

dimamarksman: `getNestedValue('x.y.z', {})` will throw `Uncaught TypeError: Cannot read property ...` anyway.

apps/docs/src/app/platform/component-docs/platform-table/platform-table-examples/platform-table-tree-example.component.ts
```diff
@@ -0,0 +1,339 @@
...
+        console.log(event);
+    }
+}
+
+interface BaseItem {
+    name: string;
+}
+
+export interface ExampleItem extends BaseItem {
+    name: string;
```
olshansky, line 37: You do not need `name` in the `ExampleItem` and `ExampleItem` at the same time, or you do not need `BaseItem` at all

libs/platform/src/lib/components/table/table.component.scss
```diff
@@ -65,31 +108,104 @@
...
+            @include fd-table-nested-cell-padding(20, true, true);
+        }
+
+        .fdp-table__cell--expand {
+            cursor: pointer;
+            display: inline-flex;
+            justify-content: center;
+            align-items: center;
+            font-size: 0.75rem;
+            color: var(--sapContent_IconColor, #0854A0);
```
olshansky, line 124: Please, add color with plain hex _(without variable)_ for IE
 
```suggestion
            color: #0854A0;
            color: var(--sapContent_IconColor, #0854A0);
```

libs/platform/src/lib/utils/object.ts
platon-rov: You right, we cannot break reduce so regular loop used here.

olshansky (approved):

apps/docs/src/app/platform/component-docs/platform-table/platform-table-examples/platform-table-tree-example.component.ts
```diff
@@ -0,0 +1,335 @@
...
+        this.source = new TableDataSource(new TableDataProviderExample());
+    }
+
+    alert(message: string): void {
+        alert(message);
+    }
+
+    onRowSelectionChange(event: TableRowSelectionChangeEvent<ExampleItem>): void {
+        console.log(event);
+    }
```
dimamarksman, line 29: Let's remove this handler and add `rowToggleOpenState` event handler instead

dimamarksman (commented):
LGTM.

Shouldn't we have a drag-and-drop event to react somehow on rearenging?

dimamarksman (approved):

Seamoo13:
- [ ] 1. case where dropdown arrow appears when it shouldn't
**STR**
- go to the tree list example 
- click to expand 'Other 1'
- drag 'beam breaker b-1' into '10 portable dvd player'
- check for expand/collapse arrow (it is present)
- drag the 'beam breaker b-1' from '10 portable dvd player' back to 'Other 1'
- look at '10 portable dvd player' item again

**actual results:** expand/collapse arrow present when there are no items under it

https://user-images.githubusercontent.com/47522152/113417898-17b13400-93cd-11eb-84a3-656fff85df24.mp4

**Expected results:** expand/collapse arrow will disappear when no child elements present

https://user-images.githubusercontent.com/47522152/113417985-3d3e3d80-93cd-11eb-8c6f-6f261a1b6280.mp4

Note: same issue for drag and drop example


- [ ] 2. missing example of tree with selection. Should it be implemented now, or will it be done later?


- [ ] 3. tab navigation skips bottom level child items

https://user-images.githubusercontent.com/47522152/113421716-3f57ca80-93d4-11eb-9b64-d365d8005540.mp4

- [ ] 4. icon missing title and aria-expanded

platon-rov:
> 3. tab navigation skips bottom level child items

@Seamoo13, not fully clear what exactly wrong. On the video I see that expandable items (which contains children rows) are focusable - and it's correct and all others (which not contains children) are not. Do you think it should behave in another way?

Seamoo13:
> > 1. tab navigation skips bottom level child items
> 
> @Seamoo13, not fully clear what exactly wrong. On the video I see that expandable items (which contains children rows) are focusable - and it's correct and all others (which not contains children) are not. Do you think it should behave in another way?

Yeah, I think all should be able to be focusable, so that a screenreader could catch them via keyboard navigation.

platon-rov:
@Seamoo13 

1 Fixed

2 Example adjusted.

3 Navigation by `Tab` works great and we really don't need to jump on table cells where any action can't be applied.

But from the point of accessibility - navigation by table's content should be done with arrows, not tab. And expanding/collapsing also should be done with arrows. You can see an example [here](https://cdn.rawgit.com/aleventhal/treegrid-example/0.1.23/treegrid.html). 

My proposition - to implement a **separate** feature (directive I guess) to be able to work with trees more natively. It's related not only to the tree table so it should not be done within the table component. I will raise a question about it on the call.

4 I've updated the tech spec. The icon just visually represents state of the row (expanded/collapsed), but `aria-expanded` should be set not for the icon, but for the row/cell.

I propose to leave `aria-expanded` on the first cell of the row so the screen-reader will read like `{cell content}, row x of n, collapsed (expanded)`.

Then, after implementing directive from point 3 we'll adapt to have the whole row in focus and the screen-reader will read like so `{row content}, row x of n, collapsed (expanded)`.

Please share your thoughts if you want to add something.

InnaAtanasova (changes requested):
1. How is the "Select All" option working? Why are not all rows selected?

Screen Shot 2021-04-05 at 4 22 36 PM

2. After "drop" event, the focus remains 
Screen Shot 2021-04-05 at 4 24 14 PM

3. The styling of the dragged item is off. The checkbox is detached from the element and the first column items gets a different focus:
Screen Shot 2021-04-05 at 4 27 30 PM
Screen Shot 2021-04-05 at 4 29 10 PM
4. The drag and drop for columns is missing

InnaAtanasova:
Please provide in the documentation examples buttons that will:
1. Expand all first level items
2. Expand only selected rows
3. Collapse selected rows
4. Collapse all

platon-rov:
> How is the "Select All" option working? Why are not all rows selected?

Fixed.

> After "drop" event, the focus remains

> The styling of the dragged item is off. The checkbox is detached from the element and the first column items gets a different focus:

I've done some investigations - `(focus)` event comes and it comes before dragging, in other words - no way to skip it. In any way - blinking of element's focus styles will be present (see that [reference](https://stackblitz.com/edit/angular-apxiuj?file=src%2Fapp%2Fcdk-drag-drop-sorting-example.ts) with pure Angular Material Drag & Drop CDK, which we're using as the base). But we also have a second focusing after the mouse button released (when item dropped, not present in the example but you can see it in our app), after a fast investigation I didn't find the reason - so all those improvements in Drag & Drop directive can be done in the separate ticket as it's not related to table itself.

> The column handling is missing - Show/hide, rearrange, resize - https://experience.sap.com/fiori-design-web/tree-table/#Tree-ColumnHandling

> Missing Context Menu - https://experience.sap.com/fiori-design-web/tree-table/#context-menu

It wasn't in [tech spec](https://app.gitbook.com/@fundamental-ngx/s/docs/platform-table-phase-4), should it be implemented in the scope of that ticket?



> 1. Expand all first level items
> 2. Expand only selected rows
> 3. Collapse selected rows
> 4. Collapse all

Every one of that requirements pushes us to expose the inner structure of rows, which isn't a great idea. Currently, we don't have an implementation of tracking rows uniqueness feature by some field (`id` or like that). So, I will prefer to implement tracking, and then if the user provides data with identifiers - allow him to collapse/expand/select/etc rows programmatically by needed id.

> Although being present in the normal Table documentation, it's good to have examples of Tree Table with filtering, sorting and Highlight items. This way we can easily test/catch bugs.

Highlighting - fixed.
Filtering, sorting - not implemented (as not implemented for the table with grouping enabled, the same algorithm will be used for both of them) - it wasn't in tech spec, so should it be implemented in the scope of the current ticket (phase)?

InnaAtanasova:
Screen Shot 2021-04-07 at 3 58 49 PM

Programatic control is something that many users require from us. UI5 documentation have this functionality and examples and I am pretty sure at some point it will be requested from us and will have to be included. I understand it's a very complex component with a lot of features and functionalities already in the PR but if we push it to the next phase, it might require refactoring the html or introducing classes which will cause Breaking changes and our goal is to minimize these Breaking changes, esp for recently added components.

InnaAtanasova:
Screen Shot 2021-04-07 at 4 05 11 PM

yes, it's the same algorithm but the other tables don't have "nesting" and by providing an example in the documentation we make sure that the sorting algorithm is working properly and we don't wait for a user to report a bug before we notice it. If you prefer you can write a test. I don't have preference, just want to make sure these functionalities are working for the Tree Table.

InnaAtanasova:
1. When you select "Laptops" row all child rows are also selected. Expected behaviour: only one row should be selected.
Screen Shot 2021-04-08 at 11 25 20 AM

2. Different focus outline style
Screen Shot 2021-04-08 at 11 12 49 AM
Screen Shot 2021-04-08 at 11 12 44 AM

platon-rov:
> When you select "Laptops" row all child rows are also selected. Expected behaviour: only one row should be selected. Different focus outline style

Fixed.

[Tech spec for Phase 4.1](https://app.gitbook.com/@fundamental-ngx/s/docs/platform-table-phase-4/platorm-table-phase-4-point-i). All in that spec relates to the tree table, that's why it's the 4.**1**. 

Also I plan to implement some [features](https://fundamental-ngx.gitbook.io/docs/platform-table-phase-4/platorm-table-phase-4-point-i#future-step-last-one-for-that-component) in phase 5 because it's not related to tree table (so it should be available for all tables).

So that's my plan and current PR ready to it's final review. Tree table hidden in the docs.

InnaAtanasova (approved):
I won't be blocking this PR anymore but just want to mention that when you click on the checkbox to select a row, the whole table gets a focus for a moment:
Screen Shot 2021-04-13 at 1 50 52 PM

The filtering is not working:
Screen Shot 2021-04-13 at 1 49 10 PM

Great job with this PR!

platon-rov:
> I won't be blocking this PR anymore but just want to mention that when you click on the checkbox to select a row, the whole table gets a focus for a moment:

That's the issue with checkbox itself. Investigated, [issue](https://github.com/SAP/fundamental-ngx/issues/5120) created.

> The filtering is not working:

That's right, it's planned in scope of 4.1 and some clarification needed before implementing. It's pointed in the [tech spec](https://app.gitbook.com/@fundamental-ngx/s/docs/platform-table-phase-4/platorm-table-phase-4-point-i) of 4.1

I done things only was described in Phase 4 tech spec and all other stuff related to tree table and drag & drop functionality described in 4.1 and will be implemented with the separate PR. And then we also have some features, not related to tree table and drag & drop functionality which I propose to implement in scope of Phase 5 (another PR). That's it.

Seamoo13 (approved):

PR #4782, opened 2021-02-27
feat: (Core|Platform) add content density switcher service to fundamental-ngx core module

### LINKED ISSUE #4062: Introduce a central Content Density switcher
feature request

It would be very useful to offer a service that allows applications of fundametnal-ngx to set centrally content density for all components they use in one application. This should work for both `core` and `platform`. 
A similar example of implementation would be RTL Service
Possible values: cozy, compact, and condensed. If condensed is missing then we should set to compact.

mikerodonnell89:
fixes #4062 

Adds `ContentDensityService` to core utils, which allows the developer to maintain application-wide content density that set on any relevant components.  `ContentDensityService` simply maintains a `BehaviorSubject<ContentDensity>` where `ContentDensity = 'cozy' | 'condensed' | 'compact'`. Developer can then subscribe to this BehaviorSubject and use it however they please - some examples lie in the library source and there is an additional example in the documentation.  

Most components that already have compact or cozy density are automatically subscribed to this service and will adjust their density if the developer has not provided a density via input. If the developer provides density via input, that will override the value stored in the service.

Some kind of composition should be added in future. Description is added in this issue: https://github.com/SAP/fundamental-ngx/issues/5041


libs/core/src/lib/breadcrumb/breadcrumb.component.ts
```diff
@@ -194,12 +197,24 @@ export class BreadcrumbComponent implements AfterContentInit, OnInit {
...
     }
 
     constructor(
         public elementRef: ElementRef,
-        private _cdr: ChangeDetectorRef,
-        @Optional() private rtlService: RtlService
+        @Optional() private rtlService: RtlService,
+        @Optional() private _contentDensityService: ContentDensityService,
+        private _cdRef: ChangeDetectorRef
     ) {}
```
JKMarkowski, line 218: Could you move constructor before all of other methods?

JKMarkowski (changes requested):
Hello @mikerodonnell89, Incredible job! works very well. Could you also add support to `ButtonBarComponent`? I see it's used in several places, but `[compact]` is hardcoded.

nikvalor:
Project: allure-results CORE 
 Url: https://allure-results--core-pr-4782-wu60x48o.web.app 
 This link will expire at Sat 03 Apr 2021 11:38:27 PM GMT

nikvalor:
Project: allure-results Platform B
 Url: https://allure-results--platform-b-pr-4782-am5iqwpl.web.app 
 This link will expire at Sun 04 Apr 2021 12:24:19 AM GMT

nikvalor:
Project: allure-results Platform A 
 Url: https://allure-results--platform-a-pr-4782-0rnu24n9.web.app 
 This link will expire at Sun 04 Apr 2021 12:28:11 AM GMT

libs/platform/src/lib/components/base.ts
```diff
@@ -58,7 +69,36 @@ export abstract class BaseComponent {
...
+    ngOnDestroy(): void {
+        this._subscriptions.unsubscribe();
+    }
+
+    /** @hidden */
+    private _setupDensitySubscriptions(): void {
+        if (this._contentDensity === null && this._contentDensityService) {
+            this._subscriptions.add(this._contentDensityService.contentDensity.subscribe(density => {
+                this.contentDensity = density;
+                this.detectChanges();
```
kavya-b, line 98: can this be changed to `markForCheck`? Dynamic page tabbed [example](https://deploy-preview-4782--fundamental-ngx.netlify.app/#/core/dynamic-page#tabbed) adds additional subscriptions and they don't take affect with `this.detectChanges()` here triggering change detection immediately thereby breaking the example.

fkolar:
Does this needs to be a service? Instead of plain injectable which you configure at the root level? We have so many services already. What is the advantage Regular Injectable Provider vs. rxjs? Do we need reactive behavior for this functionality? It will increase the load on the application as each of these subscription is created and might not be released properly.

mikerodonnell89:
> Does this needs to be a service? Instead of plain injectable which you configure at the root level? We have so many services already. What is the advantage Regular Injectable Provider vs. rxjs? Do we need reactive behavior for this functionality? It will increase the load on the application as each of these subscription is created and might not be released properly.
> 
> The same applies for RTL service we have. There are behaviors which you need to be reactive that changes and they are behavior that usually changes once a life time of the application.
> 
> First let's try to define some guideline and be consistent over applying async patterns. Each of us using this differently.

How could subscriptions not be released properly?  Each file here unsubscribes on destroy.  I agree that this behavior will usually not change frequently while an application is being used but I think we need to play it safe and assume that there can potentially be cases where the developer would be changing density often

kavya-b:
LGTM, requesting other @SAP/fundamental-ngx-ariba  members to verify this as well.

libs/core/src/lib/utils/services/content-density.service.ts
```diff
@@ -0,0 +1,16 @@
...
+import { BehaviorSubject } from 'rxjs';
+
+export type ContentDensity = 'cozy' | 'condensed' | 'compact';
+
+@Injectable()
+/**
+ * Service taking care of ContentDensity
+ */
+export class ContentDensityService {
+    contentDensity: BehaviorSubject<ContentDensity>;
```
valorkin, line 11: it seems it's a single reason service, and you could hide `contentDensity`
```suggestion
    private readonly contentDensity: BehaviorSubject<ContentDensity> =  new BehaviorSubject('cozy');;
```

libs/core/src/lib/utils/services/content-density.service.ts
```diff
@@ -0,0 +1,16 @@
...
+
+    constructor() {
```
valorkin, line 13: then you will not need `constructor`, and the only method to add is `subscribe`, and getter `value` for sync access, maybe `next` and `pipe`
you can check this article https://dev.to/alfredoperez/angular-service-to-handle-state-using-behaviorsubject-4818

libs/core/src/lib/utils/services/content-density.service.ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject } from 'rxjs';
+
+export type ContentDensity = 'cozy' | 'condensed' | 'compact';
+
+@Injectable()
+/**
```
valorkin, line 7: class comments are usually above decorators

libs/core/src/lib/input-group/input-group-directives.ts
```diff
@@ -5,30 +5,42 @@ import {
...
 })
-export class InputGroupInputDirective implements CssClassBuilder, OnInit, OnChanges {
+export class InputGroupInputDirective implements CssClassBuilder, OnInit, OnChanges, OnDestroy {
     /** user's custom classes */
     @Input()
     class: string;
 
     @Input()
-    compact = false;
+    compact: boolean = null;
```
valorkin, line 28: please use undefined and a proper typings, ng 12 will have strict enabled by default, so we should comply
```suggestion
    compact?: boolean;
```

libs/core/src/lib/input-group/input-group-directives.ts
```diff
@@ -5,30 +5,42 @@ import {
...
-    constructor(private _elementRef: ElementRef) {}
+    private _subscriptions = new Subscription();
+
+    /** @hidden */
+    constructor(private _elementRef: ElementRef, @Optional() private _contentDensityService: ContentDensityService) {}
 
     /** @hidden */
     ngOnInit(): void {
+        if (this.compact === null && this._contentDensityService) {
+            this._subscriptions.add(this._contentDensityService.contentDensity.subscribe(density => {
```
valorkin, line 39: taking in account change proposed to the service, `if` check can be removed and this line will be

```suggestion
            this._subscriptions.add(this?._contentDensityService?.subscribe(density => {
```

valorkin: what is more important here, is that you are loosing a default value for `compact`

if you move subscription to constructor, then value which is set from service will be overridden in `onInit`,
and then you can restore default value for compact

valorkin (changes requested):
on one hand I really like this serivce, on other hand I am worried about precedence of input vs service setting of content density.

because service will override input's in child components, so you will not be able to show compact elements on cozy parents

libs/core/src/lib/utils/services/content-density.service.ts
mikerodonnell89: Could you elaborate on the advantage of changing this and adding those methods to this service over what we're doing now?

libs/core/src/lib/input-group/input-group-directives.ts
mikerodonnell89: I think you may have this backwards, what's happening right now with this PR is that the value set from the service is overridden whenever an input value is provided for the density (typically the [compact] input but occasionally [cozy]).  Is this not the desired behavior?  Moving this logic to the constructor then tells the component to listen to changes from the content density service and ignore whatever is there for the input.

droshev:
`condensed` should default in `compact` if there is not a specific styles. at the moment it sets to `cozy`

AntonOlkhovskyi:
@JKMarkowski Tests for several components are fixed. 
Value Help Dialog component examples are broken, and need to be fixed. 
Search and Filter section in overlapping the table. And Ok / Cancel buttons are missing. 
Снимок экрана 2021-03-30 в 08 55 27

JKMarkowski (approved):

PR #4809, opened 2021-03-02
chore: update package.json lib versions

droshev:
update package.json lib versions


.ci-env/flags.sh-e
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+# This flag indicates if the tag should be treated as a latest one, which is going to trigger
+# - npm publish with `latest` tag, which will become new default version
+# - main version incrementation
+latest=false
```
JKMarkowski, line 5: Should there be new file named like that?

droshev: this is autogenerated

InnaAtanasova (approved):

.ci-env/flags.sh-e
droshev: this was autogenerated

JKMarkowski (approved):

nikvalor:
Project: null CORE 
 Url: null 
 This link will expire at

nikvalor:
Project: null Platform A 
 Url: null 
 This link will expire at

Issue #4869, opened 2021-03-08
Status Indicator: Focus is not going for all the icons


puru-hk:
Description: Focus is not going for all the icons

Expected: Focus should go for all the icons

droshev:
I don't understand the issue here? @puru-hk can you provide mode content?

puru-hk:
When I do tab from the keyboard tab order is not focusing for all the icons. This is one of the a11y requirement.

droshev:
i think the focus should go only if the item is actionable. I don't think the icons are in this case?

puru-hk:
Yes I agree but for screen reader if images or icon are used it should be focused then only screen reader will read the icon/images. This is accessibility requirement.

PR #4942, opened 2021-03-17
fix: (Core) updating the focusable object of status indicator and adding clickable

Lokanathan-k:

https://github.com/SAP/fundamental-ngx/issues/4869

updating the focusable object of status indicator
Focusable applied to all the tab able component and adding click feature to the component with cursor-type as pointer.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


nikvalor:
Project: null Platform B
 Url: null 
 This link will expire at

nikvalor:
Project: null Platform A 
 Url: null 
 This link will expire at

nikvalor:
Project: null Platform A 
 Url: null 
 This link will expire at

nikvalor:
Project: null Platform B
 Url: null 
 This link will expire at

nikvalor:
Project: null CORE 
 Url: null 
 This link will expire at

nikvalor:
Project: null CORE 
 Url: null 
 This link will expire at

apps/docs/src/app/core/component-docs/status-indicator/example/status-indicator-clickable-type.component.html
```diff
@@ -12,5 +12,6 @@
     title="35% with small size"
     [path]="euroImage"
     viewBox="0 0 26 25"
+    (click)="onCick()"
```
InnaAtanasova, line 15: there's a typo, also it's good to have a more meaningful name for this function

apps/docs/src/app/core/component-docs/status-indicator/example/status-indicator-clickable-type.component.ts
```diff
@@ -5,6 +5,7 @@ import { Component } from '@angular/core';
     templateUrl: './status-indicator-clickable-type.component.html'
 })
 export class StatusIndicatorClickableTypeComponent {
+    fillePercentage = 35;
```
InnaAtanasova, line 8: I think there's a typo?

apps/docs/src/app/core/component-docs/status-indicator/example/status-indicator-clickable-type.component.ts
```diff
@@ -19,4 +20,8 @@ export class StatusIndicatorClickableTypeComponent {
     cartImage = [
         'M22.9,17.2c-1.7,0-3.2,1.2-3.5,2.8h-4.8c-0.4-1.6-1.8-2.8-3.6-2.8c-0.1,0-0.2,0-0.3,0l-0.5-1.7h16L30.5,5H7L5.8,1l0,0l0,0l0-0.1L5.6,0.9l-5-0.4l0.2,1.7l3.4,0.3l4.7,15.4c-0.9,0.7-1.5,1.7-1.5,2.9c0,2,1.6,3.7,3.7,3.7c1.7,0,3.2-1.2,3.6-2.8h4.8c0.4,1.6,1.8,2.8,3.5,2.8c2,0,3.7-1.6,3.7-3.7C26.6,18.8,24.9,17.2,22.9,17.2z'
     ];
+
+    onCick(): void {
```
InnaAtanasova, line 24: typo

libs/core/src/lib/status-indicator/status-indicator.component.scss
```diff
@@ -2148,4 +2148,7 @@
 .fd-status-indicator--xl__text {
```
InnaAtanasova, line 2148: It was missed during the prev PR review, but this class makes no sense in BEM. You can't have .block--modifier__element like in your case. `.fd-status-indicator` is the block. Then it should be followed by an element, something like: `.fd-status-indicator__text` and the modifier comes last as it's modifying the default state of an element: `.fd-status-indicator__text--xl`.

InnaAtanasova (changes requested):

nikvalor:
Project: null Platform A 
 Url: null 
 This link will expire at

nikvalor:
Project: null Platform B
 Url: null 
 This link will expire at

nikvalor:
Project: null CORE 
 Url: null 
 This link will expire at

puru-hk (approved):

KevinOkamoto (approved):

libs/core/src/lib/status-indicator/status-indicator.component.scss
```diff
@@ -2083,27 +2078,30 @@
...
+.fd-status-indicator__text--critical {
   color: #e9730c;
   color: var(--sapCriticalElementColor,#e9730c)
 }
 .fd-status-indicator--negative {
   fill: #b00;
   fill: var(--sapNegativeElementColor,#b00)
 }
-.fd-status-indicator--negative__text {
+.fd-status-indicator__text--negative {
```
InnaAtanasova, line 2104: This is again the same discussion we have in fund-styles PR. 
You have the following markup:
```
<div class="fd-status-indicator fd-status-indicator--negative">
    <span class="fd-status-indicator__text--lg fd-status-indicator__text--negative">35% </span>
</div>
```
You have the `--negative` modifier applied on two places which is not ok. 
You need to set it only on the parent level and then use specificity to change the color. It's not the same as for the size, as size can be different from the size of the parent (which I doubt will happen) but the color of the label should be the same as the color of the status indicator.

So you need to have:
```
.fd-status-indicator--negative {
    .fd-status-indicator__text {
         color: #b00;
         color: var(--sapNegativeElementColor,#b00)
     }
}

.fd-status-indicator--positive {
    .fd-status-indicator__text {
         color: #107e3e;
         color: var(--sapPositiveElementColor,#107e3e)
     }
}
```

This way you will have a modifier only on the parent level:
```
<div class="fd-status-indicator fd-status-indicator--negative">
    <span class="fd-status-indicator__text--lg">35% </span>
</div>
```

libs/core/src/lib/status-indicator/status-indicator.component.scss
```diff
@@ -2145,7 +2143,7 @@
   width: 9.0625rem;
   height: 6.25rem
 }
-.fd-status-indicator--xl__text {
+.fd-status-indicator__text--xl {
```
InnaAtanasova, line 2146: these were renamed in fund-styles to label and this should be reflected here

InnaAtanasova (changes requested):

libs/core/src/lib/status-indicator/status-indicator.component.html
```diff
@@ -1,31 +1,25 @@
...
-        *ngIf="hasLabel && (labelPosition === 'right' || labelPosition === 'bottom')"
-        [ngClass]="'fd-status-indicator--' + size + '__text' + ' ' + 'fd-status-indicator--' + status + '__text'"
-        >{{ statusLabel }}
-    </span>
-</div>
+    <ng-container [ngTemplateOutlet]="svgTemplate"></ng-container>
+</svg>
+<span
+    *ngIf="hasLabel && (labelPosition === 'right' || labelPosition === 'bottom')"
+    [ngClass]="'fd-status-indicator__label--' + size"
```
InnaAtanasova, line 19: missing `class="fd-status-indicator__label"`

InnaAtanasova (changes requested):

InnaAtanasova (approved):

PR #4977, opened 2021-03-22
feat: (Core) introduce Facets component

### LINKED ISSUE #4286: Key Value and Form facets - component development
Following is the scope:
- key value facets
- Form facets
- Image facets
- Rating indicator facets

kavya-b:
closes #4286 
This PR introduces the following facet types: Form, Key-value, Image, and Rating Indicator. Facets are generally used in Object Page/Dynamic Page headers. 

Note: 
- An example of Facet usage with Platform Dynamic Page will be taken up in a separate PR due to dependency on some structure/markup changes that need to be incorporated first. Having everything in this PR will increase number of files to review > 50, therefore splitting into 2 PRs.
- Since the PR is dependent on styles that are not yet available in ngx,  this PR hardcodes the styles temporarily. These hardcoded styles will be removed when v0.17.0 of Fundamental Styles is consumed by ngx.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


libs/core/src/lib/facets/facet/facet.component.html
```diff
@@ -0,0 +1,31 @@
+<ng-container *ngIf="type === 'image'">
+    <ng-container *ngTemplateOutlet="image"></ng-container>
```
InnaAtanasova, line 2: why didn't you just put `<ng-container *ngTemplateOutlet="customContentTemplate"></ng-container>` here?

libs/core/src/lib/facets/content/facet-content.component.ts
```diff
@@ -0,0 +1,28 @@
+import { Component, ChangeDetectionStrategy, ViewEncapsulation, ElementRef, Renderer2, OnInit } from '@angular/core';
+import { FACET_CLASS_NAME } from '../constants';
+import { addClassNameToFacetElement } from '../utils';
+
+@Component({
+    selector: 'fd-facet-content',
+    template: ` <ng-content></ng-content> `,
```
InnaAtanasova, line 7: Just a question, not asking to change it :) Why did you choose to make this a component and not a directive? If it's a directive you will "expose" the native element to the user, for example `<div>` and will allow them to add props, classes, etc.

libs/core/src/lib/facets/facet-group.component.html
```diff
@@ -0,0 +1,3 @@
+<div class="fd-facet-group" role="group" [attr.aria-label]="ariaLabel">
```
InnaAtanasova, line 1: same as 'fd-facet-content'. If this is a directive the user will have the option to add labels, id, etc without fund-ngx exposing these and Input props

InnaAtanasova (commented):
Great work!

apps/docs/src/app/core/component-docs/facets/facet-examples/custom-facet-example.component.html
```diff
@@ -0,0 +1,25 @@
+<fd-facet-group ariaLabel="Custom Facets">
+    <fd-facet type="custom" [title]="'Product Description'" class="fd-docs-custom-facet-paragraph">
```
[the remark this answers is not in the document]
kavya-b: The reason why only `title` is this way at some places is because of incorrect a11y. If written as `title="some string"`, the tooltip 'some string' hovers on every element as `title` is also a HTML attribute and it just places the tooltip on every element in `fd-facet`. When used with the binding `[title]` it no longer considers it as a HTML tooltip attribute and doesn't show the same tooltip all over the facet elements and works as intended.

libs/core/src/lib/facets/facet-group.component.html
kavya-b: Same as above,  except I suppose I can bind the class and properties directly to the host than have an extra `<div>` here so that this can also support user adding their own classes/props to the same `fd-facet-group` tag level. Is that okay?

libs/core/src/lib/facets/content/facet-content.component.ts
kavya-b: If we have these as directives, we would have the extra effort of wrapping these into components again in platform(to support metaui) without any major enhancement. 😅  Also, in case of form and rating indicator facet where `fd-facet-content` is used, any extra classes can still be applied at the `fd-facet-content` tag level and no special inputs are present for this component.

libs/core/src/lib/facets/facet/facet.component.html
kavya-b: removed and simplified the template a bit

libs/core/src/lib/facets/facet-group.component.html
kavya-b: removed the `<div>` and this html file entirely and added the props to the host from the .ts file itself.

apps/docs/src/app/core/api-files.ts
```diff
@@ -72,6 +72,9 @@ export const API_FILES = {
         'DynamicPageFooterComponent',
         'DynamicPageContentComponent'
     ],
+    facets: [
+      'FacetComponent'
```
JKMarkowski, line 76: Could you add `'FacetGroupComponent'` there as well?

libs/core/src/lib/facets/facet/facet.component.html
```diff
@@ -0,0 +1,25 @@
...
+<ng-container *ngIf="type !== 'image'">
+    <h1 *ngIf="title" fd-title headerSize="5" class="fd-margin-bottom--sm" [id]="titleId">{{ title }}</h1>
+    <ng-container *ngIf="type === 'form'">
+        <ng-container *ngTemplateOutlet="contentTemplate"></ng-container>
+    </ng-container>
+    <ng-container *ngIf="type === 'key-value' || 'custom'">
+        <ng-container *ngTemplateOutlet="customContentTemplate"></ng-container>
+    </ng-container>
+    <ng-container *ngIf="type === 'rating-indicator'">
+        <h2 class="fd-form-label">{{ subtitle }}</h2>
```
JKMarkowski, line 14: Could you add `fd-form-label` direcive there, instead of class?

libs/core/src/lib/facets/facet.module.ts
```diff
@@ -0,0 +1,23 @@
...
+import { FacetComponent } from './facet/facet.component';
+import { TitleModule } from '../title/title.module';
+
+@NgModule({
+    declarations: [
+        FacetComponent,
+        FacetGroupComponent,
+        FacetContentComponent
+    ],
+    imports: [CommonModule, AvatarModule, ButtonModule, TitleModule],
```
JKMarkowski, line 16: Do you need `ButtonModule` and `AvatarModule` there?

JKMarkowski (changes requested):
Hello @kavya-b Great job! I added some comments to code. it also would be great to have example with facets usage on core's `dynamic page` component

libs/core/src/lib/facets/facet.module.ts
kavya-b: not needed, removed this :)

libs/core/src/lib/facets/facet/facet.component.html
kavya-b: replaced with `fd-form-label` directive.

apps/docs/src/app/core/api-files.ts
kavya-b: thanks for noticing, added this too.

apps/docs/src/app/core/component-docs/facets/facet-examples/custom-facet-example.component.html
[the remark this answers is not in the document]
InnaAtanasova: What about renaming `title` in this case? It will be less confusing if the user wants to add a tooltip.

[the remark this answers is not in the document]
kavya-b: sure, I will rename it as `facetTitle` instead.

mikerodonnell89 (approved):
Looks great! Just missing one thing Inna mentioned

apps/docs/src/app/core/component-docs/facets/facet-examples/custom-facet-example.component.html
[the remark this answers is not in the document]
kavya-b: pushed in changes with `facetTitle` now, please check.

kavya-b:
> Hello @kavya-b Great job! I added some comments to code. it also would be great to have example with facets usage on core's `dynamic page` component

@JKMarkowski As discussed, will work on this in a separate PR because some extra logic in dynamic page would have to be written for handling image facet on expand/collapse of header. Will create an issue for the same.

apps/docs/src/app/core/component-docs/facets/facet-examples/facet-group-example.component.html
```diff
@@ -0,0 +1,68 @@
+<fd-facet-group ariaLabel="Facet Group">
+    <fd-facet type="form" facetTitle="Technical Data" id="facet1">
+        <fd-facet-content>
+            <label class="fd-form-label" for="form-value-10">Base unit:</label>
+            <div class="fd-text" id="form-value-10">Each</div>
+        </fd-facet-content>
+        <fd-facet-content>
+            <label class="fd-form-label" for="form-value-11">Length:</label>
+            <div class="fd-text" id="form-value-11">23.24 Centimeter</div>
```
JKMarkowski, line 9: I know it's only example, but could you use `fd-text` and `fd-form-label` directives? Instead of classes

JKMarkowski (changes requested):
Hi @kavya-b There is one thing that can be changed. Rest LGTM

apps/docs/src/app/core/component-docs/facets/facet-examples/facet-group-example.component.html
kavya-b: On adding `fd-form-label` directive, it adds other styles by default that give extra padding between form and label:
image
Fiori3 recommends only 0.5rem spacing between form and label for a Form facet so adding the directive doesn't comply.

kavya-b: @JKMarkowski only replaced `fd-text` and not `fd-form-label` because of the above issue. Please check.

InnaAtanasova (approved):

InnaAtanasova:
@kavya-b the build is failing. My comments have been addressed. Great work!

kavya-b:
@JKMarkowski could you review/approve this? Your comments were addressed.

Betrozov (approved):

JKMarkowski (approved):
LGTM

puru-hk:
Avatar and rating indicator does not have the alternative text. This issues has been separately tracked to the respective components. #5033 #5032

puru-hk (approved):

Seamoo13:
@puru-hk is this already tested? if so, please change the label from 'ready for QA' to 'QA approved'. Thanks.

Issue #4991, opened 2021-03-24
Provide documentation example of Facets usage with Core Dynamic Page


kavya-b:
#### Is this a bug, enhancement, or feature request?
docs

#### Briefly describe your proposal.
Newly developed Facets component needs to be shown as part of Dynamic Page header, as it is the most common usecase for Facets. This needs to be part of both core and platform's dynamic page documentation.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
latest

#### Is there anything else we should know?
not creating a separate issue for Platform Dynamic Page since an older issue #3744 is already present describing the missing feature.

PR #4997, opened 2021-03-24
fix: (Platform) revisit Dynamic Page structure and add Facets example

### LINKED ISSUE #3744: Missing features: Dynamic Page Layout: The optional header content & Flexible column layout
Feature request, Enhancement

**1 Feature** 
-The optional header content located below the header title is missing. 
Reference link:  https://ux.wdf.sap.corp/fiori-design-web/dynamic-page-layout/#header-content

**2 Feature**
- Combining the dynamic page with the flexible column layout is missing
https://ux.wdf.sap.corp/fiori-design-web/dynamic-page-layout/#combining-the-dynamic-page-with-the-flexible-column-layout

### LINKED ISSUE #4972: Platform Dynamic Page markup needs refactoring
The Dynamic Page markup has been simplified and various issues had been addressed in styles. At the same time the core package released its implementation. It would be great if the platform implementation is revisited and reuse the implementation from core.

@manjunathanagaraj @kavya-b

kavya-b:
closes #4972 
closes #3744 
This PR addresses the following:
1. Dynamic Page went through some markup/structure changes in [Fundamental Styles](https://github.com/SAP/fundamental-styles/pull/2154) with few breaking changes that need to be incorporated before ngx consumes v0.17 of Fundamental Styles. This PR aligns these changes for Platform's Dynamic Page.
2. Facets are most commonly used with Dynamic Page/Object Page. Therefore it is necessary to provide an example usage of Facets in the Dynamic Page header.  This PR provides a new example in the Platform Dynamic Page documentation and implements logic for the Image facet to appear next to the title when the DP header is collapsed.

Note: Since this PR is coded on top of Facets changes which are also open in another PR, this PR will show all the changes from the other PR as well. 
To review changes related to this PR alone, please go through the following commit: [fix: (Platform) revisit Dynamic Page structure and add Facets example](https://github.com/SAP/fundamental-ngx/commit/34f0e605422baf5ccf074c9e2e9d69a72e61baf9 )

BREAKING CHANGES: 
1.  Removed `background` and `size` properties from the following components: 
`<fdp-dynamic-page-header>`
`<fdp-dynamic-page-title>`
`<fdp-dynamic-page-content>`
Specify `background` and `size` properties at the `<fdp-dynamic-page>` level only.

2. Added `autoResponsive` property to `<fdp-dynamic-page>` component. It is enabled by default.

- [x] Run npm run build-pack-library and test in external application - not detecting FacetModule correctly; will recheck after Facets PR is merged

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


Betrozov (approved):

KevinOkamoto (approved):

JKMarkowski (approved):
LGTM

droshev (changes requested):
We need to spend some time analyzing the breaking changes and aligning

kavya-b:
> We need to spend some time analyzing the breaking changes and aligning

okay, removing the `ready for qa` label.

droshev:
@kavya-b 
There are few things we need to consider before merging this PR:
 - I didn't see the breaking change wiki being update
 - Can't we deprecate the inputs first (in this release) and removing them in the future? The component is getting adoption and each breaking change will be very painful for our stakeholders
 -

kavya-b:
> @kavya-b
> There are few things we need to consider before merging this PR:
> 
> * I didn't see the breaking change wiki being update

I had only updated the [version-specific breaking change file](https://github.com/SAP/fundamental-ngx/wiki/0.29.0-Breaking-Changes) at that time since I didn't see a v0.29.0 section in the main one 😅 , will update the main file too.
> * Can't we deprecate the inputs first (in this release) and removing them in the future? The component is getting adoption and each breaking change will be very painful for our stakeholders

The inputs being removed are from the child components of dynamic page such as `title`, `content`, and `header`. The documentation shows these inputs used from the parent `fdp-dynamic-page` level only and not from the child levels. If we choose to deprecate this, we would need to retain the functionality of these inputs at child components level, whereas the latest Styles version being consumed does not even have the classes for these inputs anymore. We would then have to hardcode some <v0.17 dynamic page styles to support this deprecation. I feel the usecase where stakeholders are explicitly using these child-level inputs are not present currently, we can also ask @manjunathanagaraj to verify the same from 1P team that uses Platform dynamic page.
All in all, I think these particular inputs that are being removed are not going to affect stakeholders as they are already probably using it at the parent level, whose design has not changed. If you do know of some stakeholders using these inputs for the `title`, `content`, and `header` components, please let me know.
> * We need to bring the Motion Design - Snapping header for the facets. I have to admin without the animation the transition doesn't look great

Okay, but this is not just related to Facets, it is the same for other examples where scrolling happens. Do we need this handled in this PR itself or can it be taken up in a separate PR?
> * We need to enrich the documentation explaining all those cases in a bit more details. The Dynamic Page doc is quite dry

Could you elaborate a bit more on what needs to be added? I can add some more information for the Facet example, but could you give an idea of what more you are expecting?
> * At the moment the facets that is alternating while the user scroll up/down is added in 2 places. What happens if the dev doesn't add in `dynamic-page-header` or `dynamic-page-title`. I think this doesn't look mature yet

I see. Seems like we missed this during our components review meeting. I can think of the following ways of handling this:
1. We provide an input like `imageFacetSource` in `fdp-dynamic-page-title` and `fdp-dynamic-page-header` and have this `imgFacetSource` value map to `fd-avatar` inside `fd-facet` in our library implementation. We share this image source with both `title` and `header` components. This way, user won't have to write in the image facet in either places as `<fd-facet type="image"><fd-avatar></fd-avatar></fd-facet>` themselves. If this input is provided by user in only one component(say `title`), we internally add the image facet to the other component(like `header`).
But avatar takes `glyph`, `image`, and `label`, so how do we map one single input `imageFacetSource`  to these three types and understand which one the user is referring to? 
Also, should we allow user to add `<fd-facet type="image"><fd-avatar></fd-avatar></fd-facet>` explicitly in the header or title? The spec says there should only be one image facet in header and title, so this is something we might have to enforce. 

2. The second way could be to make the user provide `<fd-avatar>` themselves as part of `ng-template` and somehow inject this into the header and title through an Input, and map this to `fd-facet` within the library implementation. This way the problem of mapping  `imageFacetSource` to `glyph`, `image` or `label` as mentioned in the 1st option is no longer there, but at the cost of some implementation complexity. Also, when we inject the template, we would have to overrride the avatar size to small in title and large in header.

What do you think, @droshev ?
> * After all we need to bring this for Core Dynamic page too(after we agree how we will implement the facets)

Yes, we have a separate issue created for core example. #4991

kavya-b:
Hi @droshev,
Another PR #5216 was opened that internally modifies the dynamic page component structure to make it more modularized. Since it works on the same files that are present in this PR, and there is still work pending on this PR(some of which needs your response to my reply [here](https://github.com/SAP/fundamental-ngx/pull/4997#issuecomment-822341303)); I think it would be best if I move this PR back to [WIP](or if you want to, close it) and mark it as blocked until PR 5216 is merged. After PR 5216 is merged, I would have to additionally resolve conflicts, bring in this PR's changes on to the new internal structure, and continue addressing your comments. Please let me know if this is okay.

I had already started working on Motion Design and this is where I got to until yesterday:


It is still glitchy, somehow on scrolling down it isn't showing the fading up animation; still looking into that.

Meanwhile, if you could also respond to the above [comments](https://github.com/SAP/fundamental-ngx/pull/4997#issuecomment-822341303), it would help! Thanks!

kavya-b:
@droshev could you take a look at the comments above and share your thoughts? Thanks!

Issue #4908, opened 2021-03-11
Defect Hunting Issues for Sprint 58 - Stockholm


InnaAtanasova:
Please document the issues found during the defect hunting below
Here are some issues from the previous sprint:
- [x] #4871 fixed with https://github.com/SAP/fundamental-ngx/pull/5008
- [x] #4872 fixed with https://github.com/SAP/fundamental-ngx/pull/5009
- [x] #4873 fixed with https://github.com/SAP/fundamental-ngx/pull/4946
- [x] #4875 fixed with https://github.com/SAP/fundamental-ngx/pull/5012

mikerodonnell89:
- [ ] Feed input max height not working
![Screen Shot 2021-03-25 at 3 48 34 PM](https://user-images.githubusercontent.com/2471874/112548312-a4327580-8d81-11eb-96be-686a41f89d93.png)

- [ ] Grid list single select mode - strange tabbing behavior where it is unclear what is being focused every 3rd press of the tab key

- [ ] Considerable lag when pressing a key to change the selected option in a Select (for example Apple is selected and press "P" to select pineapple)

- [ ] Side nav - space/enter does not expand/collapse multi-level option (maybe this is intended behavior)

- [ ] Split button - clearing field returns "undefined".  Also should plus and minus buttons be accessible via keyboard?
![Screen Shot 2021-03-25 at 3 55 34 PM](https://user-images.githubusercontent.com/2471874/112548883-8ca7bc80-8d82-11eb-923c-ee89cb4d4b13.png)

- [ ] Datepicker, datetimepicker, and timepicker pages throw warning `[DOM] Found 7 elements with non-unique id #: `s.

PR #5013, opened 2021-03-25
fix: (Core) combobox minor tweaks for defect hunting

mikerodonnell89:
part of #4908 

because we removed the max width for multi-input i assume we need it for this component as well

### DIFF
--- a/libs/core/src/lib/combobox/combobox.component.html
+++ b/libs/core/src/lib/combobox/combobox.component.html
@@ -10,6 +10,7 @@
         [focusTrapped]="true"
         [triggers]="triggers"
         [disabled]="disabled || readOnly"
+        [maxWidth]="640"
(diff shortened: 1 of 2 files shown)


JKMarkowski:
Do we need it as well on `ComboboxComponent`? I thought it's only for MultiInput. @InnaAtanasova @droshev

InnaAtanasova:
> Do we need it as well on `ComboboxComponent`? I thought it's only for MultiInput. @InnaAtanasova @droshev

@mikerodonnell89 and @JKMarkowski I checked the specs and for ComboBox drop down the specs say to follow the design of Select component. In Select component there's a section called "Min and Max Width" and it currently says the following:
Screen Shot 2021-03-30 at 4 39 20 PM

InnaAtanasova (approved):

Issue #4581, opened 2021-02-05
bug: AppShell cannot use global urls to load assets


skocharian:
#### Is this a bug, enhancement, or feature request?

Bug

#### Briefly describe your proposal.

In content item app, we use images from assets folder in the component template.
When these components are rendered through Shellapp routing, images are not loaded.
Reason for this is, Shellapp tries find these images from its own asset folder.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)

@fundamental-ngx/app-shell: 0.2.35

#### If this is a bug, please provide steps for reproducing it.

1. Add assets with a global url
2. Build the application
3. Observe final asset url — it ir replaced with an app-shell base href, asset doesn't open.

manjunathanagaraj:
Solution for this is found, but implementation to be done in next sprint.

manjunathanagaraj:
The only feasible option seems to be to make it work with absolute URLs.

manjunathanagaraj:
It was exploratory. One solution is implemented, but needs review from Dima. This may require reimplementation or another implementation approach.

manjunathanagaraj:
To publish guidelines for adoption teams

omfgnuts:
^Proofreaded by me! @manjunathanagaraj

skocharian:
Final review is expected to happen today by @valorkin

PR #5023, opened 2021-03-27
feat: (AppShell) Relative to absolute urls

nowant:
Issue: https://github.com/sap/fundamental-ngx/issues/4581


libs/app-shell/src/lib/api/urls/url-dom-overrider.service.ts
```diff
@@ -0,0 +1,87 @@
...
+import { isAbsoluteUrl, resolveUrl, walkDomTree } from './url-utils';
+
+/**
+ *  Service is responsible to traverse the whole component's DOM tree
+ *  with prepending a base URL prefix to an every element with a relative URL source
+ */
+@Injectable({
+    providedIn: 'root'
+})
+export class UrlDomOverriderService {
```
valorkin, line 11: all methods could be extracted to url-utils because they don't use `this`

libs/app-shell/src/lib/api/urls/url-dom-overrider.service.ts
```diff
@@ -0,0 +1,87 @@
...
+     *              http://localhost/avatar-480w.jpg 480w,
+     *              http://localhost/avatar-800w.jpg 800w/>
+     */
+    private overrideElementSrcsetSources(element: HTMLElement, srcset: string, baseUrl: string) {
+        const overridedSrcset = [];
+
+        for (const imgDescriptor of srcset.split(/,/)) {
+            const [imgUrl, imgSize] = imgDescriptor.trim().split(/ +/g);
+
+            if (!isAbsoluteUrl(imgUrl)) {
```
valorkin, line 76: revert `if` and use early exit from loop to avoid `else` usage

libs/app-shell/src/lib/api/urls/url-overrider.pipe.ts
```diff
@@ -0,0 +1,35 @@
...
+       avatarImgSize = '30x30';
+
+       ngOnInit(): void {
+           this.avatarImgSrc = `assets/avatar-${this.avatarImgSize}.jpg`;
+       }
+    }
+ */
+@Pipe({
+    name: 'fdsUrlOverrider',
+    pure: false
```
valorkin, line 25: why impure?

libs/app-shell/src/lib/components/plugin-launcher/plugin-launcher.component.ts
```diff
@@ -238,4 +254,29 @@ export class PluginLauncherComponent implements OnChanges, AfterViewChecked {
...
+
+    /**
+     * Injects the domain name to dynamically created component
+     * To be used for UrlOverriderPipe
+     */
+    private injectBaseUrlToComponent() {
+        this._ngComponentInjector = Injector.create({
+            providers: [
+                {
+                    provide: 'REMOTE_BASE_URL',
```
valorkin, line 274: should be injection token?

valorkin (changes requested):

libs/app-shell/src/lib/api/urls/url-dom-overrider.service.ts
nowant: It could be but I would like to keep only common and reusable functions in the utils.  I just encaplsulated all specific logic (non reusable) within a service to use it as a facade with only one method.

valorkin: the thing is: same thing we will need in custom elements, not only ng-elements, where this one will work

nowant: ok, it makes sense

valorkin (approved):

PR #5027, opened 2021-03-28
chore: remove old styling

droshev:
remove old styling


libs/core/src/lib/avatar-group/avatar-group.component.scss
```diff
@@ -1,769 +1 @@
...
-}
-.fd-avatar-group .fd-avatar-group__more-button--accent-color-10 {
-    background-color: #647987;
-    background-color: var(--sapAccentColor10,#647987);
-    border-color: #647987;
-    border-color: var(--sapAccentColor10,#647987);
-    color: #fff;
-    color: var(--fdAvatar_Accent_Color,var(--sapContent_ContrastTextColor,#fff))
-}
+@import '~fundamental-styles/dist/avatar-group';
```
JKMarkowski, line 1: `0.16.1` version doesn't bring valid styles. Let's for `0.17.0` styles migration
image

libs/core/src/lib/table/table.component.scss
```diff
@@ -14,11 +14,6 @@
 	}
 }
 
-// Remove after merging and migrate https://github.com/SAP/fundamental-styles/pull/2192
```
JKMarkowski, line 17: Table changes will come with `0.17.0`, so let's uncommit this file as well

droshev: we brought the latest https://github.com/SAP/fundamental-ngx/blob/7940614e37ba0f49eb746ab2fcb887c98fe83db0/package.json#L116

droshev: i reverted it

InnaAtanasova (approved):

Issue #4583, opened 2021-02-05
Domain - Entity Format


manjunathanagaraj:
1. Define and implement some basic structure for Entity class (e.g. Requisition). How are we going to treat Entity ( id and can be updated) vs Value (Money, LineItems, is always immutable we always create new copy)
2. EntityComposite, ValueComposite

PR #5035, opened 2021-03-29
feat: (Store) Entity format

genereg1:
https://github.com/SAP/fundamental-ngx/issues/4583

- Basic implementation BaseEntity and BaseValue (Value Object)
- New store method `createEntityInstance` 
- Proxy instance wrapper to treat Entity properties (nested properties e.g arrays)

Documentation checklist:
- [x] update `README.md`
- [x ] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)
- [x] Documentation Examples


libs/store/src/lib/infrastructure/persistence/domain/base-classes/base-entity.ts
```diff
@@ -0,0 +1,47 @@
+import { Composite } from './composite';
+
+/**
+ * BaseEntityDTO that extends passed Entity DTO
+ * */
+export interface BaseEntityDTO {
+    id: number;
```
KevinOkamoto, line 7: Are we expecting the primary key of a DTO to always be named `id` and be of type `number`? In many cases the primary key will be named something else and be of type `string`.

I think we allow the user to set what the primary key is in the Entity decorator:
https://github.com/SAP/fundamental-ngx/blob/ab0dad82f415e3fc2150c5f45b507c6c9d01276d/libs/store/src/lib/domain/entity.ts#L11

libs/store/src/lib/infrastructure/persistence/store/entity-store.ts
```diff
@@ -90,7 +99,7 @@ export class DefaultQueryService<TModel> extends QueryService<TModel> {
 
     getWithQuery(query: QuerySnapshot<TModel>): Observable<TModel[]> {
         // Pass query payload instead of QueryParams
-        // TODO: Should we override EntityCollectionService interface for that? 
+        // TODO: Should we override EntityCollectionService interface for that?
         return this.service.getWithQuery(query as any);
```
KevinOkamoto, line 103: We need to map the return data here (and within the `getByKey` method) from the EntityCollectionService to call `instanceForType`, so that the end-user interacts with Entity objects and not DTOs.

KevinOkamoto (changes requested):

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
```diff
@@ -0,0 +1,51 @@
...
+    get(target: any, prop: PropertyKey): any {
+        // Check is an array and wraps a new one proxy
+        if (['[object Object]', '[object Array]']
+            .indexOf(Object.prototype.toString.call(this.targetState[prop])) > -1) {
+            return new Proxy(this.targetState[prop], this);
+        }
+        return target[prop] || this.targetState[prop];
+    }
+
+    set<T>(target: TModel, key: PropertyKey, value: Type<T> | Array<Type<T>>): boolean {
```
dimamarksman, line 28: As for me, there is no much sense in `T` in this case.
Let's handle `value` as `unknown` and use `instanceof` to guess what it actually is.

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
```diff
@@ -0,0 +1,51 @@
...
+    constructor(
+        private targetState?: TModel
+    ) {
+    }
+
+    get(target: any, prop: PropertyKey): any {
+        // Check is an array and wraps a new one proxy
+        if (['[object Object]', '[object Array]']
+            .indexOf(Object.prototype.toString.call(this.targetState[prop])) > -1) {
+            return new Proxy(this.targetState[prop], this);
```
dimamarksman, line 23: That means that every read operation will create a new instance of proxy, Am I right?
For example:
```
read entityProxy.lineItems ---> create new instance of Proxy
...
read entityProxy.lineItems ---> create new instance of Proxy 
```

I wonder if we could create only one Proxy instance for each property that needs to be wrapped. Can we do that?

libs/store/src/lib/infrastructure/persistence/store/entity-store-builder.ts
```diff
@@ -42,7 +42,7 @@ export abstract class EntityStoreBuilderFactory {
 /**
  * Entity Store Builder default implementation
  */
-export class DefaultEntityStoreBuilder<T extends BaseEntity> implements EntityStoreBuilder<T> {
```
dimamarksman, line 45: Why did you remove `extends BaseEntity` constraint?

libs/store/src/lib/infrastructure/persistence/store/entity-store.ts
```diff
@@ -33,6 +34,12 @@ export interface EntityStore<T> {
      * @param entity Entity to be deleted
      */
     delete(entity: T): Observable<T>;
+
+    /**
+     * Create new entity instance wrapped Proxy
+     * @param fromState initial state
+     */
+    createEntityInstance(fromState?: any): T;
```
dimamarksman, line 42: can we avoid `any`?

dimamarksman (changes requested):
Have got some question, please take a look

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
dimamarksman, line 23: Can we really share the same Proxy handler with the new child Proxy?
Asking because `this` handler shares the same `targetState` for all underlying proxies. 
At the same time we use `this.targetState` to make an assignment for each particular proxy which represents different model shapes. 
Does it really work? Just can't get it...

libs/store/src/lib/infrastructure/persistence/domain/base-classes/base-entity.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Base Entity follows the DDD definition of Entity
+ */
+export abstract class BaseEntity<EntityProps extends BaseEntityDTO> extends Composite<EntityProps> {
+     constructor(dto: EntityProps) {
+         super();
+         this._dto = dto;
+     }
+     _dto: EntityProps;
```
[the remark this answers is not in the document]
genereg1: `this._dto` used by `equals` method in `BaseEntity` class just to check id equality. Probably, a better way to assign just only id property `this._id = dto.id`

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
genereg1: Good point. I will check what we can do with it

libs/store/src/lib/infrastructure/persistence/domain/base-classes/composite.ts
```diff
@@ -0,0 +1,4 @@
+export abstract class Composite<T> {
+    value: T;
```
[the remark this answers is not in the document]
genereg1: Actually, it doesn't use. Initially, the idea was to create some interface/class to get access to proxy value or get errors from there. But currently, I don't see a case when we could use it.

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
```diff
@@ -0,0 +1,51 @@
...
+        }
+        return target[prop] || this.targetState[prop];
```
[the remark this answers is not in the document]
genereg1: Yea. I think it's a better way to do it

libs/store/src/lib/infrastructure/persistence/domain/base-classes/base-value.ts
```diff
@@ -0,0 +1,28 @@
...
+    protected constructor(dto: T) {
+        super();
+        this.dto = Object.freeze(dto);
+    }
+
+    /**
+     * Clone Value Object since they are immutable
+     */
+    clone() {
+        return { ...this.dto };
```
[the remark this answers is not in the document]
genereg1: From my perspective, the `clone` method will be used in the case when we want to edit our VO. Currently, the method just creates a copy of non-editable dto that would be editable, but the original dto will be still non-editable.

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
```diff
@@ -0,0 +1,51 @@
...
+    set<T>(target: TModel, key: PropertyKey, value: Type<T> | Array<Type<T>>): boolean {
+        if (value instanceof BaseValue) {
+            if (!value.dto) {
+                this.targetState[key] = value;
+                return true;
+            }
+            const isEqual = Array.isArray(target) && target.some(elem => value.equals(elem)
+                || value.equals(target[key]));
+
+            let clonedVO = target[key] && value.clone();
```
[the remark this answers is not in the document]
genereg1: Removed

libs/store/src/lib/infrastructure/persistence/domain/base-classes/base-value.ts
```diff
@@ -0,0 +1,28 @@
+import { Composite } from './composite';
+
+export abstract class BaseValue<T> extends Composite<T>{
+    readonly dto: T;
+
+    protected constructor(dto: T) {
+        super();
+        this.dto = Object.freeze(dto);
```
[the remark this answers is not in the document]
genereg1: Added deepFreeze function

libs/store/src/lib/infrastructure/persistence/domain/base-classes/base-entity.ts
```diff
@@ -0,0 +1,49 @@
+import { Composite } from './composite';
+
+/**
+ * BaseEntityDTO that extends passed Entity DTO
+ * */
+export interface BaseEntityDTO {
+    id: number;
+    primaryKey?: string;
```
KevinOkamoto, line 8: You don't need to add the `primaryKey` property here. What I was trying to say was that the DTO from a service might not have `id` as its identifier, it may be some other field, like:
```ts
{
   uuid: '1a3dd8b8-bb2a-4e62-92f7-6f5f5cb2ea7' ,
   name: 'John',
   age: 24
}
```
Here `uuid` is the identifier for the DTO, and it is a string.

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
```diff
@@ -34,15 +33,12 @@ class StateHandler<TModel> {
...
+                target[key] = new (Object.getPrototypeOf(value).constructor)(value.dto);
                 return true;
             }
         } else {
-            this.targetState[key] = value;
-            return true;
+            return Reflect.set(this.targetState as any, key, value);
         }
 
         return true;
```
dimamarksman, line 44: Should be `false` probably

apps/store-app/src/app/entities/requisition.entity.ts
```diff
@@ -0,0 +1,27 @@
+import 'reflect-metadata';
```
fkolar, line 1: I have already commented something similar in other PR but angular is getting rid of this polyfils.  Should not we do the same? 

https://angular.io/guide/deprecations#reflect-metadata

libs/store/src/lib/infrastructure/persistence/domain/base-classes/base-entity.ts
fkolar: There could be variety of ways to specify which key is primary key. In ariba especially we used many times. This is the reason we started originally with `Identity`  so we can describe what field and type. as kevin says we need to be able to specify which field is primary key

libs/store/src/lib/infrastructure/persistence/domain/base-classes/composite.ts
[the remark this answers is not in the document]
fkolar: Its either Composite<T> or our BaseEntity/BaseValue. It's the same, both have a internal DTO state and we read values from it or set value. so composite supposed to be abstract class that BaseEntity/BaseValue inherit from as both have internal DTO State. 

and Yes original idea was to handle also the NULL/undefined problem (the articles we posted into slack channel). + expose some kind of xx$ state to this entity level . 

So the `value: T` is our `readonly dto: T` it is duplicate information.

libs/store/src/lib/infrastructure/persistence/domain/base-classes/base-value.ts
```diff
@@ -0,0 +1,45 @@
+import { Composite } from './composite';
+
+export abstract class BaseValue<T> extends Composite<T>{
+    readonly dto: T;
```
fkolar, line 4: see comment bellow about `Composite`

fkolar (changes requested):
Just a few minor comments most of the things were covered already

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
fkolar: I am not sure if this can be the same instance, as proxy does not handle dotted notation, of the Target, key, value is always for the main object

libs/store/src/lib/infrastructure/persistence/domain/state-handler.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+
+interface ReqDTO extends BaseEntityDTO {
+    id: number;
+    title: string;
+    amount: number;
+    lineItems: Array<LineItem>;
+}
+
+class Requisition extends BaseEntity<ReqDTO> {
+    id: number;
```
fkolar, line 51: How we ended up with the discussion about the ID?  In Ariba I see we can have :

```
class Requisition {uniqueName: string}

class SourcingEvent {internalId: string;}

class SomeOtherEntity {id: string)

Now we need to have way to point to the correct key. 

What if our baseEntity will have method/getter/setter. 

identity(): string, so that in the Requisition about you will implement it and call the right DTO attribute? 

```
class Requisition extends BaseEntity {
   identity(): string {
    return  this.value.id;
   }
 
}
```

or maybe just getter/setter

**Review by fkolar — CHANGES_REQUESTED** [review]

File: libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts [hunk 18]
```diff
@@ -0,0 +1,46 @@
...
+            const isEqual = Array.isArray(target) && target.some(elem => value.equals(elem)
+                || value.equals(target[key]));
+
+            if (!isEqual) {
+                // target[key] = value;
+                // return true;
+                return Reflect.set(target as any, key, value);
+            }
+        } else {
+            return Reflect.set(this.state as any, key, value);
```
dimamarksman, line 40: Why do we use `this.state` as a target in this case?

dimamarksman:
```
it('Test nesting', () => {
        class Child {
            title: string;
        }
        
        interface DTO {
            id: string;
            title: string;
            child: Child;
        }
        class Primary extends BaseEntity<DTO> {
            title: string;
            child: Child;
        }

        const initState: DTO = {
            id: '123',
            title: 'Primary',
            child: {
                title: 'Child'
            }
        };

        const primary: Primary = instanceForType(Primary, initState);

        expect(primary.title).toBe('Primary');

        primary.child.title = 'New Child';

        expect(primary.title).toBe('Primary'); --> will fail cause 'primary.title' will be 'New Child'
    });
```

dimamarksman (changes requested):

libs/store/src/lib/domain/decorators.ts
```diff
@@ -1,40 +1,44 @@
...
 const ENTITY_META_MAP = new Map<Type<any>, EntityMetaOptions<any>>();
 const RESOURCE_MAP = new Map<Type<any>, EntityResourceMetaOptions>();
 
-export type EntityType<T extends BaseEntity> = Type<T>;
+export type EntityType<T> = Type<T>;
 
-export function Entity<T extends BaseEntity>(config: EntityMetaOptions<T>) {
+export interface EntityClass<T> extends BaseEntity<T> {
+    new (...arg: any[]): EntityClass<T>;
+}
```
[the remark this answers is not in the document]
dimamarksman: Instead you might need something like that:
```
export type EntityClass<E extends BaseEntity<K>, K = any> = new (...args: any[]) => E;
```

dimamarksman (changes requested):

libs/store/src/lib/infrastructure/persistence/store/entity-store-builder.spec.ts
```diff
@@ -5,37 +5,47 @@ import {
...
-    getEntityResourceMetadata<T extends BaseEntity>(entity: string | EntityType<T>): EntityResourceMetaOptions {
+    getEntityResourceMetadata<T>(entity: string | EntityType<T>): EntityResourceMetaOptions {
         throw new Error('Method not implemented.');
     }
-    getEntityMetadata<T extends BaseEntity>(entity: string | EntityType<T>): EntityMetaOptions<T> {
+    getEntityMetadata<T>(entity: string | EntityType<T>): EntityMetaOptions<T> {
+        throw new Error('Method not implemented.');
+    }
+
+    getEntityTypeByName(entityName: string): EntityBaseType {
```
dimamarksman, line 39: return type should be `EntityClass` probably

libs/store/src/lib/infrastructure/persistence/store/entity-store-builder.ts
```diff
@@ -46,6 +46,7 @@ export interface EntityStoreBuilder<T extends BaseEntity> {
  * Entity Store Builder Factory interface
  */
 export abstract class EntityStoreBuilderFactory {
+    // abstract create<T extends BaseEntity<T>>(entity: EntityType<T>): EntityStoreBuilder<T>;
```
dimamarksman, line 49: please remove this comment

libs/store/src/lib/infrastructure/persistence/store/entity-store.ts
```diff
@@ -37,13 +39,19 @@ export interface EntityStore<T> {
      * @param entity Entity to be deleted
      */
     delete(entity: T): Observable<T>;
+
+    /**
+     * Create new entity instance wrapped Proxy
+     * @param fromState initial state
+     */
+    createEntityInstance(fromState?: T): T;
```
dimamarksman, line 47: Actually `fromState` should be based on EntityDTO generic type, `T` in this context means entity class type, not a DTO type

libs/store/src/lib/infrastructure/persistence/store/entity-store.spec.ts
```diff
@@ -37,14 +45,15 @@ class UserCollectionServiceMock implements Partial<EntityCollectionService<User>
 class QueryBuilderMock extends QueryBuilder<User> {}
 
 describe('Default Entity Store', () => {
+    let entity: Type<User>;
```
dimamarksman, line 48: We have to cover `createEntityInstance` method by unit tests

dimamarksman (changes requested):

libs/store/src/lib/infrastructure/persistence/store/entity-store.ts
```diff
@@ -70,7 +79,7 @@ export class DefaultEntityStore<T extends BaseEntity> implements EntityStore<T>
     }
```
KevinOkamoto, line 79: When getting an entity from the EntityService here:
```ts
    get(id: IdentityKey): Observable<T> {
        return this._entityService.getByKey(id);
    }
```

Does the EntityService return an Observable with the Entity or EntityDTO?

Shouldn't we be calling `instanceForType` here? For example:

```ts
    get(id: IdentityKey): Observable<T> {
        return this._entityService.getByKey(id).pipe(
            map(dto: EntityDTO => instanceForType(this._entity, dto))
        );
    }
```

fkolar: I think we should be returning Entity, EntityDTO is an internal thing

libs/store/src/lib/domain/decorators.ts
```diff
@@ -1,40 +1,44 @@
-import { BaseEntity } from './entity';
 import { EntityMetaOptions } from './entity-meta-options';
 import { EntityResourceMetaOptions } from './rest-resource';
 import { Type } from './utility';
+import { BaseEntity } from '../infrastructure/persistence';
```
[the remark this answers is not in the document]
fkolar: I agree here.

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
```diff
@@ -0,0 +1,47 @@
...
+    ) {
+    }
+
+    get(target: any, prop: PropertyKey): any {
+        // Check is nested object and wraps a new one proxy
+        if (this.state[prop] && typeof this.state[prop] === 'object') {
+            return new Proxy(this.state[prop], this);
+        }
+
+        return Reflect.get(target as any, prop);
```
fkolar, line 26: Did not we say that we are not going to use Reflect? and Dima removed it from his PR as well?

libs/store/src/lib/infrastructure/persistence/query/query.ts
```diff
@@ -1,5 +1,5 @@
 import { Observable } from 'rxjs';
-import { BaseEntity, ChainingStrategyFieldsMap } from '../../../domain/public_api';
+import { ChainingStrategyFieldsMap } from '../../../domain/public_api';
```
fkolar, line 2: Let's use direct import not from barrel files or these public API files.

libs/store/src/lib/infrastructure/persistence/store/entity-collection-service-base.spec.ts
```diff
@@ -6,7 +6,7 @@ import { EntityMetaOptions, EntityMetaOptionsService } from '../utils/entity-opt
 import { EntityCollectionService } from './entity-collection-service';
 import { DefaultEntityCollectionService } from './entity-collection-service-base';
 import { EntityCollectionsService } from './entity-collections-service';
-import { BaseEntity } from './entity-server/interfaces';
+import { BaseEntity, IdentityKey } from '@fundamental-ngx/store';
```
fkolar, line 9: for local test you should not be importing it from @xxxx/store

libs/store/src/lib/infrastructure/persistence/store/entity-store.spec.ts
```diff
@@ -2,10 +2,18 @@ import { of } from 'rxjs';
 
 import { DefaultEntityStore } from './entity-store';
 import { QueryBuilder } from '../query/query-builder';
+import { IdentityKey, Type } from '../../../domain/public_api';
+import { BaseEntity } from '../domain/base-classes/base-entity';
```
fkolar, line 6: as above , use direct imports

fkolar (changes requested):

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
dimamarksman: Maybe I missed something but why we shouldn't use Reflect?

dimamarksman: The only thing the Reflect has to use accessor option otherwise there is no much sense in it

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
```diff
@@ -0,0 +1,47 @@
+import 'reflect-metadata';
```
dimamarksman, line 1: We don't need `reflect-metadata`

dimamarksman (changes requested):

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
dimamarksman: @fkolar it's a native Reflect, not that one that comes from `reflect-metadata`

dimamarksman (approved):

KevinOkamoto (approved):

fkolar (approved):

PR #5110, opened 2021-04-12
docs: (Core) create missing list variant examples of card

### LINKED ISSUE #3863: Card: Missing variant of s list card
Feature request

It's missing a variant of List card:
https://ux.wdf.sap.corp/fiori-design-web/overview-page-list-cards/

DeepakSap14:
fixes #3863 

This PR creates list card with select control.

Documentation checklist:
- na update `README.md`
- na [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)
- [x] Documentation Examples


apps/docs/src/app/core/component-docs/card/examples/bar-chart-list-card/card-bar.component.ts
```diff
@@ -14,4 +14,10 @@ export class CardBarComponent {
 
     /** bar hex color */
     @Input() color = '#1866b4';
```
InnaAtanasova, line 16: You can't just create a component `fd-card-bar`. This would mean for the user that it's coming from Fundamental library.

droshev (changes requested):
I don't understand the value of this PR. It doesn't bring any value to the library.
It even creates more confusion as it has styling done in the documentation, which means it is not delivered as part of the library and each application has to copy-paste it.

I think the card documentation needs to be revisited and cleaned from things like `https://github.com/SAP/fundamental-ngx/blob/main/apps/docs/src/app/core/component-docs/card/examples/bar-chart-list-card/bar-chart-list-card-example.component.scss`

DeepakSap14:
> @DeepakSap14 I'm afraid this PR should be closed. If you check the wiki page of Visual Core (the specs) you will see that there's a Card type called List Card and it has a subsection `List Card with extended Attributes`. This needs to be implemented in Fundamental-styles first and then brought to Fund-ngx. I am afraid an example only will not be enough. The design of this Component is still `IN PROGRESS` so you might want to discuss this with Manju and plan it in the next sprints

Yes @InnaAtanasova ,
i see https://wiki.wdf.sap.corp/wiki/display/FXU/Comparison+Chart+Specification and https://ux.wdf.sap.corp/fiori-design-web/micro-chart/
design is available now.

earlier while creating components it was may be not available and we created a sample example.
i will close this PR till above is getting implemented in styles.

PR #5186, opened 2021-04-22
fix: (Core) refactor Notifications to follow the latest design specs

InnaAtanasova:
refactored Notifications component to reflect the latest css from Fundamental-styles

BREAKING CHANGES:
- Changes in markup
- Removed `fd-notification-avatar` directive
- The directive `fd-notification-actions` is now a component
- Removed `fd-notification-text` directive
- Removed `fd-notification-description` directive
- Removed f`d-notification-metadata` directive
- Removed the Notifications From Object option together with the NotificationDefault.
- Notifications Group is no longer a banner displayed in the right corner but displayed inside a popover triggered by clicking the bell icon in the shell bar at the top right of the screen.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


libs/core/src/lib/notification/notification/notification.component.scss
```diff
@@ -1,621 +1,10 @@
...
-}
-.fd-notification--group .fd-notification__indicator--error,
-.fd-notification--group .fd-notification__indicator--information,
-.fd-notification--group .fd-notification__indicator--success,
-.fd-notification--group .fd-notification__indicator--warning {
-    padding-top: 0.75rem;
-    padding-bottom: 0.75rem;
+// To be removed with the new fund-styles release
+.fd-notification--group .fd-notification__body:last-child {
+    border-bottom: var(--sapList_BorderWidth) solid var(--sapList_BorderColor);
```
InnaAtanasova, line 9: This is fixed in PR in styles: https://github.com/SAP/fundamental-styles/pull/2272

libs/core/src/lib/notification/notification-group-header/notification-group-header.component.ts
```diff
@@ -0,0 +1,111 @@
...
+    @HostBinding('class.fd-notification__group-header')
+    fdNotificationGroupHeaderClass = true;
+
+    /** @hidden */
+    _rtl = false;
+
+    /** @hidden */
+    _subscription = new Subscription();
+
+    /** Whether the expand button is incompact mode */
```
N1XUS, line 50: Did you mean "in compact"?

libs/core/src/lib/notification/notification-group-list/notification-group-list.component.ts
```diff
@@ -0,0 +1,55 @@
+import { 
+    AfterContentInit,
+    ChangeDetectionStrategy, 
+    ChangeDetectorRef,
+    Component, 
+    ContentChild,
+    forwardRef,
+    Input,
```
N1XUS, line 8: Not used

libs/core/src/lib/notification/notification-group/notification-group.component.ts
```diff
@@ -1,14 +1,60 @@
...
+    buildComponentCssClass(): string[] {
+        return [
+            'fd-notification fd-notification--group fd-notification-custom-block',
+            this.mobile ? 'fd-notification--mobile' : '',
+            this.class
+        ];
+    }
+
+    /** @hidden */
+    elementRef(): ElementRef<any> {
```
N1XUS, line 52: Do we need an explicit type argument "any" here?

libs/core/src/lib/notification/notification-service/notification.service.ts
```diff
@@ -4,37 +4,29 @@ import { NotificationContainer } from '../notification-utils/notification-contai
...
-     * @param content Content of the alert component, or NotificationDefault object.
+     * Opens a notification component with a content of type TemplateRef or Component Type
      * @param notificationConfig Configuration of the notification component.
-     * @param notificationGroup Configuration of the notification component.
      */
     public open(
-        content: TemplateRef<any> | Type<any> | NotificationDefault,
-        notificationConfig: NotificationConfig = new NotificationConfig(),
-        notificationGroup?: ComponentRef<NotificationGroupComponent>
+        content: TemplateRef<any> | Type<any>,
```
N1XUS, line 23: "content" parameter was removed from documentation while still present in argument list

N1XUS: Also noticed that isAnyOpened method is unused anywhere, and maybe should be removed, or at least added to tests?

libs/core/src/lib/notification/notification/notification.component.ts
```diff
@@ -38,108 +44,128 @@ export type NotificationSize = 's' | 'm';
...
-        this.componentRef = this.containerRef.createComponent(componentFactory);
-        this.componentRef.instance.defaultConfigurationNotification = conf;
-        this.componentRef.instance.type = this.type;
+    /** @hidden Listen on NavigationStart event and dismiss the dialog */
+    private _listenAndCloseOnNavigation(): void {
+        if (this._router && this.notificationRef) {
+            this._subscriptions.add(
+                this._router.events.pipe(
+                    filter(event => event instanceof NavigationStart && this.closeOnNavigation)
+                ).subscribe(event => this.notificationRef.dismiss())
```
N1XUS, line 138: Event here is not used, do we need it in arguments?

N1XUS (changes requested):
Great job! I've commented couple of small issues I've noticed.

libs/core/src/lib/notification/notification-group/notification-group.component.ts
InnaAtanasova: yes. That's also how it's done in the other components:
Screen Shot 2021-05-03 at 11 06 51 AM

libs/core/src/lib/notification/notification-service/notification.service.ts
InnaAtanasova: `isAnyOpened()` was added in the previous implementation of the component in case the user wants to check if there are open notifications before reacting to some event (for example, not allowing the user to change the page, etc.)

InnaAtanasova: > "content" parameter was removed from documentation while still present in argument list

I'm not sure I understand your comment.

InnaAtanasova: The content here refers to a template or a component. In the documentation we have 2 examples -  "Template as content" and "Component as content".

InnaAtanasova:
@N1XUS thanks for your review, your comments have been addressed.

libs/core/src/lib/notification/notification-service/notification.service.ts
N1XUS: I meant this line: @param content Content of the alert component, or NotificationDefault object.
It was removed from the JsDoc, while the parameter wasn't

libs/core/src/lib/notification/notification-group/notification-group.component.ts
N1XUS: image
But the generic type of ElementRef is any by default, so this code is copying the default behavior

mikerodonnell89:
Notification w/ message strip - notification body has padding-top after message strip is dismissed so it looks a bit asymmetrical vertically
Screen Shot 2021-05-03 at 2 14 34 PM

I also noticed that if I open the "component as content" example then open the grouping example, then close the component as content popup, the grouping example closes as well.  Might be a problem if multiple notifications are being displayed at once

libs/core/src/lib/notification/notification-group-header/notification-group-header.component.ts
```diff
@@ -0,0 +1,111 @@
...
+        return this.expanded
+            ? 'slim-arrow-down'
+            : this._rtl
+                ? 'slim-arrow-left'
+                : 'slim-arrow-right'
+    }
+
+    /** @hidden */
+    private _listenRtl(): void {
+        if (this._rtlService) {
```
mikerodonnell89, line 102: This component should also depend on the content density service and set button sizes accordingly

mikerodonnell89 (changes requested):

InnaAtanasova:
> Notification w/ message strip - notification body has padding-top after message strip is dismissed so it looks a bit asymmetrical vertically
> Screen Shot 2021-05-03 at 2 14 34 PM
> 
> I also noticed that if I open the "component as content" example then open the grouping example, then close the component as content popup, the grouping example closes as well. Might be a problem if multiple notifications are being displayed at once

This is a css issue. The body of the notification that has a message strip has a margin-top.

libs/core/src/lib/notification/notification-group/notification-group.component.ts
InnaAtanasova: ok...will remove it but it will be inconsistent with the rest of the app. Prob need to create an issue to refactor it for all components.

libs/core/src/lib/notification/notification-service/notification.service.ts
InnaAtanasova: gotcha. Put back the content param

N1XUS (approved):

mikerodonnell89 (approved):
After investigating, input actually seems to be the best way to deal with message strip margin

Issue #5172, opened 2021-04-20
DynamicPage modularity problem


fkolar:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.
Currently `DynamicPageComponent` is implemented as monoliths component without possibility to scale this into individual and smaller (reusable) fragments. Everything needs to be defined inside the content, which makes this hard to add some dynamic behavior. 

When using this `DynamicPage` all these constructs should be possible, as with larger content you want to break it down into smaller pieces. 

1.

````html
<fdp-dynamic-page>
    <ng-container *ngTemplateOutlet="titleTemplate">
    </ng-container>

    <ng-container *ngTemplateOutlet="content">
    </ng-container>

    <ng-container *ngTemplateOutlet="footer">
    </ng-container>
</fdp-dynamic-page>

```


````html
<fdp-dynamic-page >

    <fdp-dynamic-page-title >     
        <ng-container *ngTemplateOutlet="titleContentTemplateWithActionsAndBredCrumb">
       </ng-container>
    </fdp-dynamic-page-title>
  </fdp-dynamic-page> 
    ....

```


3. 

````html
<fdp-dynamic-page >

  <div>. <!--  wrapper element should not break the structure-->
    <fdp-dynamic-page-title >     
         ....
    </fdp-dynamic-page-title>
  <div>
  </fdp-dynamic-page> 
    ....

```


#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)


```
Angular: 11.2.9
... animations, common, compiler, compiler-cli, core, forms
... language-service, localize, platform-browser
... platform-browser-dynamic, router
Ivy Workspace: <error>

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1102.8
@angular-devkit/build-angular   0.1102.8
@angular-devkit/core            11.2.8
@angular-devkit/schematics      11.2.8
@angular/cdk                    11.2.8
@angular/cli                    11.2.8
@angular/material               11.2.8
@schematics/angular             10.1.7
@schematics/update              0.1102.8
ng-packagr                      11.2.4
rxjs                            6.6.7
typescript                      4.1.5
    
```

dimamarksman:
@fkolar The third use case is not clear. 
For what do we need this `div`? Part of which DOM area it should be?

fkolar:
> @fkolar The third use case is not clear.
> For what do we need this `div`? Part of which DOM area it should be?

[...]

Issue #5169, opened 2021-04-20
DynamicPageModule


fkolar:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.
Dynamic page components are using DynamicPageService which is not part of Module but provided by component. Module should contain everything needed and related to create DynamicPage. When we have this on the component level it limits the usage . 

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)

<pre><code>
Angular CLI: 11.2.8
Node: 12.0.0
OS: darwin x64

Angular: 11.2.9
... animations, common, compiler, compiler-cli, core, forms
... language-service, localize, platform-browser
... platform-browser-dynamic, router
Ivy Workspace: <error>

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1102.8
@angular-devkit/build-angular   0.1102.8
@angular-devkit/core            11.2.8
@angular-devkit/schematics      11.2.8
@angular/cdk                    11.2.8
@angular/cli                    11.2.8
@angular/material               11.2.8
@schematics/angular             10.1.7
@schematics/update              0.1102.8
ng-packagr                      11.2.4
rxjs                            6.6.7
typescript                      4.1.5

</code>
</pre>

dimamarksman:
@fkolar `DynamicPageService` should be dedicated for each particular dynamic page that's why it is in component providers section

manjunathanagaraj:
closing this, as per comments and confirmation from @dimamarksman

PR #5216, opened 2021-04-28
fix: (Platform) dynamic page component modularity

dimamarksman:
Closes: https://github.com/SAP/fundamental-ngx/issues/5172

This PR changes how Platform `Dynamic Page` component projection works.
This changes allows us to add extra wrappers around projected content.
```html
<fdp-dynamic-page>
            <!-- Extra wrapper should not break projection-->
            <ng-container>
                <fdp-dynamic-page-content></fdp-dynamic-page-content>
           </ng-container>
</fdp-dynamic-page> 
```
The PR does not change Public API so signature remains the same.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples


fkolar (approved):
I am pretty happy with this, removing dependencies to ng-content and relaying, ViewChild/ContentChild. Even we had all this already we did not utilized the power of ng-container projection which is more flexible, than static `ng-content selector`.

Now I just need to to test it out..

fkolar (approved):

sKudum (approved):

libs/platform/src/lib/components/dynamic-page/dynamic-page.module.ts
```diff
@@ -35,7 +39,6 @@ import { DynamicPageComponent } from './dynamic-page.component';
         DynamicPageLayoutActionsComponent,
         DynamicPageHeaderComponent,
         DynamicPageContentComponent,
-        DynamicPageTabbedContentComponent,
         DynamicPageFooterComponent
     ]
 })
```
kavya-b, line 44: can we also handle issue [5169](https://github.com/SAP/fundamental-ngx/issues/5169) here? Adding `DynamicPageService` in the `providers` array should be enough right?

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
```diff
@@ -69,42 +71,44 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
     @Input()
     offset = 0;
```
kavya-b, line 72: if `offset` is not being used, we should probably also remove from the @Input and add to breaking change?

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
```diff
@@ -69,42 +71,44 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
...
+    @ViewChild('tabsContainer')
+    set _tabsContainer(tabsContainer: ElementRef<HTMLElement>) {
+        if (!tabsContainer?.nativeElement) {
+            return;
+        }
+        this._setTabStyles(tabsContainer.nativeElement);
+    }
 
     /**
      * tracks whether the header was toggled or not
```
kavya-b, line 114: can also remove `isHeaderCollapsed`, seems like it is not used anywhere

libs/platform/src/lib/components/dynamic-page/dynamic-page-header/header/dynamic-page-header-internal.component.ts
```diff
@@ -0,0 +1,433 @@
...
+     * subscription for when collapse header is called
+     */
+    private _collapseSubscription: Subscription;
+
+    /** @hidden */
+    constructor(
+        private _cd: ChangeDetectorRef,
+        private _elementRef: ElementRef<HTMLElement>,
+        private _renderer: Renderer2,
+        protected _dynamicPageConfig: DynamicPageConfig,
```
kavya-b, line 210: `_dynamicPageConfig` not being used anywhere in this file, can be removed.

kavya-b (commented):
Incredible work! 
Just couple of minor things:
- In the flexible column layout example, the breadcrumb is getting cut off instead of truncating like before:
 current: 
Screenshot 2021-04-29 at 3 29 05 PM 
 expected:

Screenshot 2021-04-29 at 3 29 24 PM

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
dimamarksman: I would prefer just to add `@deprecated` for the beginning

libs/platform/src/lib/components/dynamic-page/dynamic-page.module.ts
dimamarksman: I think it's not what we need in this case. 
`DynamicPageService` must be dedicated (not singleton) for each instance of the dynamic page, otherwise we`ll get a singleton.

dimamarksman:
> Incredible work!
> Just couple of minor things:
> 
> * In the flexible column layout example, the breadcrumb is getting cut off instead of truncating like before:
>   current:
> 
> Screenshot 2021-04-29 at 3 29 05 PM
> 
> expected:
> Screenshot 2021-04-29 at 3 29 24 PM

@kavya-b  I can reproduce the same problem in the current doc as well

image

It's connected to the breadcrumb resize logic directly and not to the dynamic page itself

libs/platform/src/lib/components/dynamic-page/dynamic-page.component.ts
dimamarksman: added `@deprocated`

libs/platform/src/lib/components/dynamic-page/dynamic-page-header/header/dynamic-page-header-internal.component.ts
dimamarksman: removed

kavya-b (approved):
LGTM!

fkolar:
I think we are ready to merge. What do you think ?

droshev:
@N1XUS Can you take a look at this PR? thanks

dimamarksman:
> @dimamarksman does core need similar refactoring?

@droshev the core component has the same projection `<ng-content select="projection-selector"></ng-content>` approach.
It would be good to use the same projection approach as proposed in this PR but not sure if it's required right now.
@fkolar initiated this issue cause he tries to instantiate fdp-dynamic-page dynamically using some mixe of `ng-container + ng-template`

PR #5259, opened 2021-04-30
fix: (Platform) make formcontrol non mandatory  in checkbox group for non form cases

### LINKED ISSUE #5173: bug: Checkbox-Group (platform) - formcontrol binding is expected for without form-field as well
bug

Not able is use non template driven checkbox group outside form fields. 
Usage of checkbox-group outside form displays errors in console as shown below.

11

1) Create a checkbox-group component outside forms. In non template driven approach.

sKudum:
Closes #5173 

Not able is use non template driven checkbox group outside form fields.

Documentation checklist:
- [n/a ] update `README.md`
- [n/a ] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)
- [x] Documentation Examples

### DIFF
--- a/libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.ts
+++ b/libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.ts
@@ -54,6 +54,18 @@ export class CheckboxGroupComponent extends CollectionBaseInput {
     @Input()
     isInline = false;
 
+    /**
+     * Establishes two way binding, when checkbox group used outside form.
+     */
+    @Input()
+    get checked(): string[] {
+        return this._checked;
+    }
+    set checked(checkedValues: string[]) {
+        this._checked = checkedValues;
+        this.value = checkedValues;
+    }
+
     /** Children checkboxes passed as content */
     @ContentChildren(CheckboxComponent)
     contentCheckboxes: QueryList<CheckboxComponent>;
@@ -64,6 +76,13 @@ export class CheckboxGroupComponent extends CollectionBaseInput {
 
     @Output()
     readonly valueChange: EventEmitter<PlatformCheckboxChange> = new EventEmitter<PlatformCheckboxChange>();
+
+    /** Emits checked change event */
+    @Output()
+    readonly checkedChange: EventEmitter<String[]> = new EventEmitter<String[]>();
+
+    /** @hidden used for two way binding, when used outside form */
+    private _checked: string[];
 
     constructor(
         cd: ChangeDetectorRef,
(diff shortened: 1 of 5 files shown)


DeepakSap14 (approved):
LGTM

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.ts
```diff
@@ -65,6 +77,13 @@ export class CheckboxGroupComponent extends CollectionBaseInput {
     @Output()
     readonly valueChange: EventEmitter<PlatformCheckboxChange> = new EventEmitter<PlatformCheckboxChange>();
 
+    /** Emits checked change event */
+    @Output()
+    readonly checkedChange: EventEmitter<String[]> = new EventEmitter<String[]>();
+
```
KevinOkamoto, line 83: Why do we need the `checkedChange` EventEmitter when we already have the `valueChange` EventEmitter? Don't they do the same thing?

KevinOkamoto: On second thought, I guess this is to enable two-way binding.

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.ts
```diff
@@ -54,6 +54,18 @@ export class CheckboxGroupComponent extends CollectionBaseInput {
     @Input()
     isInline = false;
 
+    /**
+     * Establishes two way binding, when checkbox group used outside form.
+     */
+    @Input()
```
fkolar, line 60: Why we can not use valueChanged ?

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.ts
KevinOkamoto: `String[]` should be lower case -> `string[]`

KevinOkamoto (changes requested):

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.ts
sKudum: > Why do we need the `checkedChange` EventEmitter when we already have the `valueChange` EventEmitter? Don't they do the same thing?

In valueChanged, we have passed an object which has a property which denotes which checkbox was changed and the final values of checked checkboxes. 

Here `value` is not `Input` for two way binding we need only checked checkboxes value. 
So for avoiding any breaking change, we need to introduce `checkedChange` property.

libs/platform/src/lib/components/form/checkbox-group/checkbox-group.component.ts
sKudum: In valueChanged, we have passed an object which has a property which denotes which checkbox was changed and the final values of checked checkboxes. 

Here `value` is not `Input` for two way binding we need only checked checkboxes value. 
So for avoiding any breaking change, we need to introduce `checkedChange` property.

KevinOkamoto (approved):

InnaAtanasova (approved):

Issue #5077, opened 2021-04-06
Accessibility:Text


Sanjay-1312:
Accessibility: Text
 Component (There are couple of accessibility issues found while testing, so i have attached error document for you reference.)
Screenshot:
[Text_ARIBA Controls.docx](https://github.com/SAP/fundamental-ngx/files/6262295/Text_ARIBA.Controls.docx)

Sanjay-1312:
Hello colleagues,
The issue is fixed .
Hence Closing the ticket.

URL:https://deploy-preview-5271--fundamental-ngx.netlify.app/#/core/text

Thanks and Reagrds
Sanjay

PR #5271, opened 2021-05-03
fix: (Core) Text component's more or less label for expand/collapse functionality is not readable for accessibility.

manu-kr:
Closes SAP/fundamental-ngx https://github.com/SAP/fundamental-ngx/issues/5077

Text component more or less label for expand/collapse functionality is not readable for accessibility.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/text/text.component.html
+++ b/libs/core/src/lib/text/text.component.html
@@ -10,6 +10,7 @@
     </span>
 
     <a fd-link
+       tabindex="0"
        class="fd-text__link--more"
        *ngIf="_expandable && _hasMore"
        (click)="toggleTextView()"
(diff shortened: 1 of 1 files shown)


libs/core/src/lib/text/text.component.html
```diff
@@ -10,6 +10,7 @@
     </span>
 
     <a fd-link
+       tabindex="0"
```
dimamarksman, line 13: `tabindex` makes a link focusable but does not allow us to trigger click using `enter` key down.
Instead, we could add `href="#"` attribute that would make a link focusable and react on `enter` by default but we should prevent the default behaviour on click

dimamarksman: Or keep `tabindex="0"` and listen to `keydown eneter` event as well

manu-kr: Modified. Can you please re review.

dimamarksman (approved):

KevinOkamoto (approved):

Issue #5217, opened 2021-04-28
StateHandler does not handle initialized data properly


KevinOkamoto:
#### Is this a bug, enhancement, or feature request?

bug

#### Briefly describe your proposal.

In the following scenario, I'm getting `undefined` when I access any entity attributes

```ts
class Pet extends BaseEntity {
  id: string;
  name: string;
  type: string;
  identity() { return this.id }
}

const pet1 = instanceForType(Pet, { name: 'Cuddles', type: 'dog' });

// pet1.name -> undefined
```

I should be able to access any of the properties as if it is the actual entity.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)

n/a

#### If this is a bug, please provide steps for reproducing it.

1. Create Entity class definition
2. Instantiate an entity using `instanceForType` with proper arguments
3. Try to access the properties of entity instance,

#### Please provide relevant source code if applicable.

https://github.com/SAP/fundamental-ngx/blob/store/libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts

PR #5341, opened 2021-05-10
fix: (Store) Fix entity format

dimamarksman:
Closes SAP/fundamental-ngx#5217

Fix for Entity Format. It should give the ability access entity properties directly e.g. `myEntity.entityProperty`

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/store/src/lib/infrastructure/persistence/domain/state-handler.spec.ts
+++ b/libs/store/src/lib/infrastructure/persistence/domain/state-handler.spec.ts
@@ -1,15 +1,11 @@
-import { DefaultEntityStore } from '../store';
-import { QueryBuilder } from '../query/query-builder';
 import { BaseEntity, BaseValue } from '../domain/base-classes';
 import { RESTResource } from '../../../domain/decorators';
-import { Type } from '../../../domain/utility';
-import { EntityCollectionService } from '../store';
-
-type IdentityKey = string | number;
+import { IdentityKey } from '../../../domain/public_api';
+import { instanceForType } from './state-handler';
 
 interface MoneyDTO {
     amount: number;
-    currency: string
+    currency: string;
 }
 
 interface Item {
@@ -25,15 +21,18 @@ interface ReqDTO {
     lineItems: Array<LineItem>;
 }
 
-class Money extends BaseValue<MoneyDTO> {
-    constructor(value?: MoneyDTO) {
-        super(value);
+class Money extends BaseValue<MoneyDTO> implements MoneyDTO {
+    amount: number;
+    currency: string;
+
+    constructor(dto: MoneyDTO) {
+        super(dto);
     }
 }
 
 @RESTResource({
     root: 'http://www.example.com/v1/',
-    path: '/requisitions/:reqId/lineItems',
+    path: '/requisitions/:reqId/lineItems'
 })
 export class LineItem extends BaseValue<Item> {
     amount: Money;
@@ -48,6 +47,7 @@ interface ReqDTO {
     uniqueName: IdentityKey;
     title: string;
     amount: number;
+    total: Money;
     lineItems: Array<LineItem>;
 }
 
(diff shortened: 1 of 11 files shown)


libs/store/src/lib/infrastructure/persistence/domain/state-handler.spec.ts
```diff
@@ -25,15 +21,18 @@ interface ReqDTO {
...
 }
 
-class Money extends BaseValue<MoneyDTO> {
-    constructor(value?: MoneyDTO) {
-        super(value);
+class Money extends BaseValue<MoneyDTO> implements MoneyDTO {
+    amount: number;
+    currency: string;
+
+    constructor(dto: MoneyDTO) {
```
fkolar, line 28: how are. you using DTO here? I dont think its used at all right ?

libs/store/src/lib/infrastructure/persistence/domain/state-handler.spec.ts
```diff
@@ -48,100 +47,137 @@ interface ReqDTO {
...
     amount: number;
+    total: Money;
     lineItems: Array<LineItem>;
 }
 
 class Requisition extends BaseEntity<ReqDTO> {
     uniqueName: IdentityKey;
     title: string;
     amount: number;
+    total: Money;
```
fkolar, line 58: I am not sure how this this Requisition work and how we are using Parent BaseEntity

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
```diff
@@ -1,46 +1,69 @@
...
+            // otherwise redirect to the default value.
+            // Need it to make working entity.getter flow
+            if (propertyKey in value) {
+                return Reflect.get(value, propertyKey, value);
+            }
         }
-
-        return Reflect.get(target as any, prop);
+        // default flow
+        return Reflect.get(target, propertyKey, receiver);
```
fkolar, line 31: Did not we discuss that we will not use Reflect.***

libs/store/src/lib/infrastructure/persistence/store/entity-store.spec.ts
```diff
@@ -2,18 +2,23 @@ import { of } from 'rxjs';
...
-class User extends BaseEntity<{}>{
-    constructor(public id: IdentityKey, public name: string, public age: number) {
-        super({ id, name, age });
+class UserDTO {
+    id: IdentityKey;
+    name: string;
+    age: number;
+}
+class User extends BaseEntity<UserDTO> {
+    constructor(dto: UserDTO) {
```
fkolar, line 16: These examples are not aligned, somewhere we are constructor and accessing .value and somewhere not.

fkolar (changes requested):

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
fkolar: Yes. But are you using we need it ? With you previous asnwer ?

libs/store/src/lib/infrastructure/persistence/domain/state-handler.spec.ts
dimamarksman: I removed this constructor here so it's not required

libs/store/src/lib/infrastructure/persistence/domain/state-handler.spec.ts
dimamarksman: I left the core things the same, so it uses the same approach as previously. 
A proxied entity relies on internal `composite.value` namespace.
`Proxy.get` trap tries to find the property in the `composite.value` namespace, if it’s not there then redirect to the default flow.
`Proxy.set` trap tries to set a new value in the `composite.value` namespace if possible, if not redirect to the default assignment process

libs/store/src/lib/infrastructure/persistence/domain/state-handler.ts
dimamarksman: sorry didn't get your last question

libs/store/src/lib/infrastructure/persistence/domain/state-handler.spec.ts
dimamarksman: Parent class constructor is used by default

genereg1 (approved):

fkolar (approved):

Issue #5420, opened 2021-05-19
Defect hunting for Sprint 62


InnaAtanasova:
Please record the bugs you found with a short explanation and a screenshot or gif.

mikerodonnell89:
- [ ] Avatar group RTL - missing margin between text and icon in popover 
<img width="519" alt="Screen Shot 2021-05-19 at 8 23 03 AM" src="https://user-images.githubusercontent.com/2471874/118829642-8530fb00-b87b-11eb-88c9-51cb5e462271.png">

aberikashvili:
[ ] Alert icons disappearing when RTL is on

RTL off
![image](https://user-images.githubusercontent.com/3863691/118865529-15aa2280-b8f2-11eb-8c26-244f41633869.png)

RTL on
![image](https://user-images.githubusercontent.com/3863691/118865562-1f338a80-b8f2-11eb-9288-97929aeca5c8.png)

N1XUS:
- [ ] Calendar: On mobile devices, mobile calendar navigation container is not centered when opened in landscape mode
![image](https://user-images.githubusercontent.com/6586561/118944134-f5637d80-b95c-11eb-808d-907169841383.png)
![image](https://user-images.githubusercontent.com/6586561/118944304-1deb7780-b95d-11eb-9e76-7cab77e1331b.png)

- [ ] Card: RTL mode, wrong positioning of counter
![image](https://user-images.githubusercontent.com/6586561/118944384-30fe4780-b95d-11eb-9b82-1b87ec26a801.png)
![image](https://user-images.githubusercontent.com/6586561/118944473-45dadb00-b95d-11eb-805a-921772eaae8c.png)

- [ ] Carousel: RTL mode breaks navigation with gestures
![image](https://user-images.githubusercontent.com/6586561/118944817-93574800-b95d-11eb-8391-618120d5620c.png)
https://user-images.githubusercontent.com/6586561/118945094-cd284e80-b95d-11eb-8c0b-437195beeacd.mov

- [ ] Dialog: RTL not affecting dialog itself 
![image](https://user-images.githubusercontent.com/6586561/118945848-84bd6080-b95e-11eb-98e7-305c43800cdb.png)

- [ ] Facets: RTL for counter looks odd 
![image](https://user-images.githubusercontent.com/6586561/118946085-bcc4a380-b95e-11eb-8a17-96068708473b.png)

- [ ] Feed list: no space before "Reply link" 
[...]

PR #5434, opened 2021-05-20
fix: (Core) add fix for Multi-Input defect hunting Sprint 62 bugs

mikerodonnell89:
fixes two multi-input bugs as part of #5420:

- selecting a new item by clicking it anywhere other than the checkbox should close the popover (popover should stay open if checkbox is clicked)
- Tokenizer (and multiinput) - cmd/ctrl+A to select all tokens then you must press delete twice to delete - should only happen once


Betrozov (approved):

Betrozov:
@mikerodonnell89 please remove package-lock.json from this PR

libs/core/src/lib/multi-input/multi-input.component.ts
```diff
@@ -419,29 +419,38 @@ export class MultiInputComponent implements
     }
 
     /** @hidden */
-    handleSelect(checked: any, value: any, event?: MouseEvent): void {
+    checkboxChanged(event: boolean, value: any): void {
+        this.handleSelect(event, value, null, true)
+    }
+
+    /** @hidden */
+    handleSelect(checked: any, value: any, event?: MouseEvent, fromCheckbox?: boolean): void {
```
platon-rov, line 427: `checked` should have boolean type.

platon-rov:
I'm not sure that it should be done in the scope of current PR but I see the component has a lot of `any[]` usage, which can be refactored to use `T[]` to have better type checking.

libs/core/src/lib/multi-input/multi-input.component.ts
```diff
@@ -417,29 +417,38 @@ export class MultiInputComponent implements
...
-        if (this._shouldPopoverBeUpdated(previousLength, this.selected.length)) {
-            this.popoverRef.refreshPosition();
-        }
+            // Handle popover placement update
+            if (this._shouldPopoverBeUpdated(previousLength, this.selected.length)) {
+                this.popoverRef.refreshPosition();
+            }
 
-        this._resetSearchTerm();
+            if (!fromCheckbox) {
```
Betrozov, line 444: please remove empty `if`

Betrozov (commented):
also, please squash all your commits in this PR
and then rename your commit using the new format

Betrozov (changes requested):
also, please squash all your commits in this PR
and then rename your commit using the new format

PR #5437, opened 2021-05-21
fix: (Core) remove wizard padding logic from dialog body component

### LINKED ISSUE #5402: Dialog body has unnecessary wizard component import
Currently the dialog body checks for a wizard component ContentChild and applies some styles if it exists.  At the same time the wizard ascends its nativeElement's parents to check to see if it's in a dialog in order to calculate the height of itself.  We can remove the WizardComponent import from the dialog body and handle the styling logic there in the wizard itself, solving the circular dependency issue

mikerodonnell89:
fixes #5402 

Moves wizard padding logic away from the dialog body to remove circular dependency between components

### DIFF
--- a/libs/core/src/lib/wizard/wizard.component.ts
+++ b/libs/core/src/lib/wizard/wizard.component.ts
@@ -9,6 +9,7 @@ import {
     HostListener,
     Input,
     OnDestroy,
+    Optional,
     QueryList,
     TemplateRef,
     ViewChild,
@@ -19,6 +20,7 @@ import { Subscription } from 'rxjs';
 import { WizardProgressBarDirective } from './wizard-progress-bar/wizard-progress-bar.directive';
 import { scrollTop } from '../utils/functions/scroll';
 import { ACTIVE_STEP_STATUS, CURRENT_STEP_STATUS, UPCOMING_STEP_STATUS, COMPLETED_STEP_STATUS } from './constants';
+import { DialogBodyComponent } from '../dialog/dialog-body/dialog-body.component';
 
 export const STEP_MIN_WIDTH = 168;
 export const STEP_STACKED_TOP_CLASS = 'fd-wizard__step--stacked-top';
@@ -93,7 +95,11 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
     /** @hidden */
     private _previousWidth: number;
 
-    constructor(private _elRef: ElementRef, private readonly _cdRef: ChangeDetectorRef) {}
+    constructor(
+        private _elRef: ElementRef,
+        private readonly _cdRef: ChangeDetectorRef,
+        @Optional() private _dialogBodyComponent: DialogBodyComponent
+    ) {}
 
     /** @hidden */
     @HostListener('window:resize')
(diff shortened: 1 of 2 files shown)


libs/core/src/lib/wizard/wizard.component.ts
```diff
@@ -201,6 +201,10 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
         const dialogBody = this._elRef.nativeElement.parentElement;
         let retVal = 0;
         if (dialogBody.tagName.toLowerCase() === 'fd-dialog-body') {
+            const dialogBodyStyle = this._elRef.nativeElement.parentElement.style;
```
N1XUS, line 204: This can be done with @Inject(DialogBodyComponent), or with some providers, which might be more feasible here.
After you inject parent component, you can call parent component's method to set additional class to the host, which will apply this styling changes, instead of doing this with JS.

mikerodonnell89: Great suggestion, I will use `@Optional` to inject it into the constructor

libs/core/src/lib/wizard/wizard.component.ts
```diff
@@ -198,14 +204,21 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
 
     /** @hidden */
     private _getDialogOffset(): number {
-        const dialogBody = this._elRef.nativeElement.parentElement;
         let retVal = 0;
-        if (dialogBody.tagName.toLowerCase() === 'fd-dialog-body') {
-            if (dialogBody.querySelector('.' + 'fd-title--h2')) {
-                retVal = dialogBody.querySelector('.' + 'fd-title--h2').offsetHeight;
+        if (this._dialogBodyComponent) {
```
N1XUS, line 208: I would do it like this:
`this._dialogBodyComponent.addClass('fd-dialog-body--no-vertical-padding')`
As for height, resizeObserver might help you out with window resize and dynamic content

libs/core/src/lib/wizard/wizard.component.ts
```diff
@@ -198,14 +204,21 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
+            const dialogBody = this._dialogBodyComponent.elementRef().nativeElement;
+            if (dialogBody.tagName.toLowerCase() === 'fd-dialog-body') {
+                const dialogBodyStyle = dialogBody.style;
+                dialogBodyStyle.paddingTop = '0';
+                dialogBodyStyle.paddingBottom = '0';
+                dialogBodyStyle.overflowY = 'hidden';
+                if (dialogBody.querySelector('.' + 'fd-title--h2')) {
```
N1XUS, line 215: Get element once and set it to some const. Also, no need to contact the string here

libs/core/src/lib/wizard/wizard.component.ts
mikerodonnell89: this is actually a property of dialogConfig so I will use that

InnaAtanasova (approved):

Issue #3123, opened 2020-08-28
Accessibility Bug: When the message strips are accessed, the information inside the message strips are not conveyed (screen reader)


puru-hk:
Fundamental core component: "Message Strip"

Description: When the message strips are accessed, the information inside the message strips are not conveyed (screen reader)

Expected: The content inside the message strips should be conveyed properly when the message strips are accessed(screen reader)

Attached the screenshot

![image](https://user-images.githubusercontent.com/32538291/91556741-db0ded80-e950-11ea-8a1e-17d2464ab595.png)

Note: - Issue persists for all the message strips present within the ‘Message strip’ control.

puru-hk:
@Sanjay-1312 Please test this issue in this url https://deploy-preview-5458--fundamental-ngx.netlify.app

Sanjay-1312:
Hello Colleagues,

The issue is not fixed, hence i am re opening the issue.
On accessing the 'Message Strip', the close information is only read, the message strip information is not read.
There is no keyboard support for the message strip where there is Close button is not available.

URL: https://deploy-preview-5458--fundamental-ngx.netlify.app
JAWS 2021

Screenshot :
![image](https://user-images.githubusercontent.com/78714518/120278503-fee0d400-c2d2-11eb-998a-5343a9b34d79.png)

puru-hk:
@kavya-b This issue is not fixed it-seems. Kindly have a look.

kavya-b:
As discussed with @puru-hk and @Sanjay-1312, the only way we see to fix this in JAWS is to provide `tabindex="0"` on the message strip. This means adding focus on message strip, which is not specified in the Fiori3 visual specs. Without this, even the UI5 example for message strip does not read the strip if a close button is not present.
Another issue is that UI5 reads the entire message strip content on tabbing to the close button. Is this correct behaviour? Shouldn't a button read only what it is trying to indicate rather than the entire strip's contents? Also the message strip should read what it is trying to indicate. I think this is a question for designers as well. cc: @manjunathanagaraj @kenykim 

@Sanjay-1312 will follow up with his team to understand whether we need focus on the entire message strip and how can they test a strip without a close button, if it is expected that close button will read out the entire strip's information.

kavya-b:
@InnaAtanasova @droshev currently, we do not have focus specification by Fiori for Message Strip. However, it seems like this is necessary in order for screen readers(especially JAWS) to read the message strip content. 

[...]

PR #5458, opened 2021-05-25
fix(core): set default ariaLabel to Message Strip for screen reader a11y

kavya-b:
Closes SAP/fundamental-ngx#3123

This PR adds the `ariaLabel` property to some of the examples for the screen readers to properly convey the message strip content. PR also handles logic for when `ariaLabel` property is not specified, the library will set the text of the message strip as the aria-label.

Documentation checklist:
- [n/a] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/message-strip/message-strip.component.html
+++ b/libs/core/src/lib/message-strip/message-strip.component.html
@@ -1,3 +1,6 @@
+<p class="fd-message-strip__text" #messageStripContent role="alert">
+    <ng-content></ng-content>
+</p>
 <button
     class="fd-message-strip__close"
     fd-button
@@ -10,6 +13,3 @@
     [attr.aria-label]="dismissLabel"
     [attr.title]="dismissLabel"
 ></button>
-<p class="fd-message-strip__text">
-    <ng-content></ng-content>
-</p>
--- a/libs/core/src/lib/message-strip/message-strip.component.ts
+++ b/libs/core/src/lib/message-strip/message-strip.component.ts
@@ -7,7 +7,10 @@ import {
     OnInit,
     ViewEncapsulation,
     EventEmitter,
-    Output
+    Output,
+    AfterViewInit,
+    ChangeDetectorRef,
+    ViewChild
 } from '@angular/core';
 import { applyCssClass } from '@fundamental-ngx/core/utils';
 import { CssClassBuilder } from '@fundamental-ngx/core/utils';
@@ -22,18 +25,15 @@ let messageStripUniqueId = 0;
     templateUrl: './message-strip.component.html',
     styleUrls: ['./message-strip.component.scss'],
     host: {
-        '[attr.aria-labelledby]': 'ariaLabelledBy',
-        '[attr.aria-label]': 'ariaLabel',
         '[style.width]': 'width',
         '[style.min-width]': 'minWidth',
         '[style.margin-bottom]': 'marginBottom',
-        role: 'alert',
         '[attr.id]': 'id'
     },
     encapsulation: ViewEncapsulation.None,
     changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class MessageStripComponent implements OnInit, OnChanges, CssClassBuilder {
+export class MessageStripComponent implements OnInit, AfterViewInit, OnChanges, CssClassBuilder {
     /** User's custom classes */
     @Input()
     class = '';
@@ -63,13 +63,13 @@ export class MessageStripComponent implements OnInit, OnChanges, CssClassBuilder
     @Input()
     ariaLabelledBy: string = null;
 
-    /** Aria label for the message-strip component element. */
+    /** Aria label for the message-strip component element. If not specified, it will be set to the content. */
     @Input()
     ariaLabel: string = null;
 
     /** Aria label for the dismiss button. */
     @Input()
-    dismissLabel = 'Dismiss';
+    dismissLabel = 'Close';
 
     /** Width of the message-strip. */
     @Input()
@@ -88,7 +88,11 @@ export class MessageStripComponent implements OnInit, OnChanges, CssClassBuilder
     onDismiss: EventEmitter<void> = new EventEmitter<void>();
 
     /** @hidden */
-    constructor(private _elementRef: ElementRef) {}
+    @ViewChild('messageStripContent', { read: ElementRef })
+    private _messageStripContent: ElementRef;
+
+    /** @hidden */
+    constructor(private _cd: ChangeDetectorRef, private _elementRef: ElementRef) {}
 
     /** @hidden */
     ngOnInit(): void {
@@ -98,6 +102,16 @@ export class MessageStripComponent implements OnInit, OnChanges, CssClassBuilder
     /** @hidden */
     ngOnChanges(): void {
         this.buildComponentCssClass();
+    }
+
+    /** @hidden */
+    ngAfterViewInit(): void {
+        if (this.ariaLabel) {
+            this._elementRef.nativeElement.setAttribute('aria-label', this.ariaLabel);
+        } else {
+            this._elementRef.nativeElement.setAttribute('aria-label', this._getMessage());
+        }
+        this._cd.detectChanges();
     }
 
     /** @hidden */
@@ -127,5 +141,32 @@ export class MessageStripComponent implements OnInit, OnChanges, CssClassBuilder
             this.noIcon ? 'fd-message-strip--no-icon' : '',
             this.class
         ];
+    }
+
+    /**
+     * @hidden
+     * get the message that screen reader should speak when icons or types are present
+     */
+    _getMessage(): string {
+        let message = '';
+        if (!this.type) {
+            message = 'Message Strip normal' + (this.dismissible ? ' closeable' : '') + '.';
+        } else if (this.type && this.noIcon) {
+            message = 'Message Strip ' + this.type + (this.dismissible ? ' closeable' : '') + '.';
+        } else {
+            message =
+                'Message Strip ' + this.type + (this.dismissible ? ' closeable' : '') + ' with ' + this.type + ' icon.';
+        }
+        // also read the text
+        message += this._messageStripContent.nativeElement.innerText;
+        if (this.dismissible) {
+            message +=
+                ' Message Strip ' +
+                this.dismissLabel +
+                ' information bar. ' +
+                (this.type ? this.type : 'Normal') +
+                ' Message Strip button.';
+        }
+        return message;
     }
 }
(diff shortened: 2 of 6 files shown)


platon-rov (approved):

DeepakSap14 (approved):

InnaAtanasova (approved):

InnaAtanasova (approved):

mikerodonnell89 (changes requested):
Would be nice if the ariaLabel was set automatically - this can be done by giving the <ng-content> for the text a template variable, accessing it via ContentChild in the component, and then if the developer has not explicitly set the ariaLabel via input then to set it in AfterContentInit

kavya-b:
> Would be nice if the ariaLabel was set automatically - this can be done by giving the for the text a template variable, accessing it via ContentChild in the component, and then if the developer has not explicitly set the ariaLabel via input then to set it in AfterContentInit

@mikerodonnell89 using ElementRef reference to set the aria-label value to the message strip text in AfterViewInit. ContentChild was not holding the reference correctly, so used ViewChild instead. Please take a look. 

@InnaAtanasova @DeepakSap14 @platon-rov please re-review since code was changed after your approvals.

platon-rov (approved):

InnaAtanasova (commented):
Now that ariaLabel is set by default (to the innerText) I think only 1 example can have a custom ariaLabel and the rest can be as they were before. The user might get the impression that they have to always provide `ariaLabel`. Also, I think it's a good idea to mention this functionality in the documentation (that if the aria label is omitted it will be set to the content).

libs/core/src/lib/message-strip/message-strip.component.ts
```diff
@@ -99,6 +107,14 @@ export class MessageStripComponent implements OnInit, OnChanges, CssClassBuilder
         this.buildComponentCssClass();
     }
 
+    /** @hidden */
+    ngAfterViewInit(): void {
+        if (!this.ariaLabel) {
+            this.ariaLabel = this.messageStripContent.nativeElement.innerText;
```
InnaAtanasova, line 113: Just a question. What will happen if the user passes a more complex content like text+button or text+link or even another component?

mikerodonnell89:
Testing the current build at https://fundamental-ngx.netlify.com/ with Chrome's screenreader, it is actually reading the message strips correctly when I click them, and with the deployment preview on this PR, the screenreader is reading the text from the message strip twice. @kavya-b what screen reader are you testing with?

mikerodonnell89 (changes requested):
Testing the current build at https://fundamental-ngx.netlify.com/ with Chrome's screenreader, it is actually reading the message strips correctly when I click them, and with the deployment preview on this PR, the screenreader is reading the text from the message strip twice. @kavya-b what screen reader are you testing with?

libs/core/src/lib/message-strip/message-strip.component.ts
kavya-b: hmm, not sure about this. How do you think we should handle this then?

kavya-b:
> Testing the current build at https://fundamental-ngx.netlify.com/ with Chrome's screenreader, it is actually reading the message strips correctly when I click them, and with the deployment preview on this PR, the screenreader is reading the text from the message strip twice. @kavya-b what screen reader are you testing with?

I am testing it with Voiceover on Mac.. what I understood from the issue was the screen reader was earlier only reading the button(‘Dismiss’) on entering the strip, but it should also read the text. In VoiceOver, we can explicitly navigate through Ctrl+option left/right arrows and it reads the text(without needing any aria label which is probably why it works on the latest netlify too), but apparently the issue requires that it be read out on entering the whole message strip. So now with the aria label, it reads the text on entering the strip through Ctrl+option left/right arrows(in Mac) and then it reads the button and text as well explicitly,  which is why it seems to be repeating twice.

libs/core/src/lib/message-strip/message-strip.component.ts
InnaAtanasova: Screen Shot 2021-05-26 at 1 31 01 PM

I tested this with adding a button to the text. It takes the content of the button and append it to the other text.

InnaAtanasova: @mikerodonnell89 FYI

kavya-b:
Had a call with the a11y team member alongwith Puru, and this is basically what they are expecting from a message strip to read out:
- an invisible text for the semantic icon at the start (Message Strip Error Closable)
- the text(There are no values distributed to characteristic Location)
- The type of message strip(Error Message Strip Close Information Bar)
- the close button(Error Message Strip Close Button)
Screenshot 2021-05-27 at 10 22 55 AM
(the image is an example from a UI5 project that a11y member showed us)

we may need to place some invisible span tags to read out these information @InnaAtanasova @mikerodonnell89 
cc: @puru-hk @manjunathanagaraj

Betrozov (approved):

kavya-b:
@InnaAtanasova @mikerodonnell89  made the following changes:
- [text that is already visible should not have aria-label on it](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute), therefore the previous approach of attaching aria-label directly to the text element by reading its contents was wrong. The suggested approach is to use aria-labelledby on the parent(`fd-message-strip`) to read out the content through the id. 
- so now what we do is we attach the `fd-message-strip__text` class to the host message strip’s `aria-labelledby`.
since we accept `aria-labelledby` from user as well, if provided, we can add both these ids to the host aria-labelledby. 
- we have 2 invisible spans now: one that reads out the strip information with type/icon info, and another that holds the custom aria label that user may provide through `ariaLabel` property. With the latter, we want the screen reader to read out the custom aria label and not what is provided from  aria-labelledby. An example of this is the 2nd message strip of the last example.
image

On voiceover, it reads only once for me now, please check.

libs/core/src/lib/message-strip/message-strip.component.html
```diff
@@ -1,3 +1,13 @@
+<span class="fd-accessible-hidden-message" [attr.aria-label]="_getMessage()"></span>
```
mikerodonnell89, line 1: I do not think this invisible element is necessary, this aria-label can be placed on the <fd-message-strip> host element

libs/core/src/lib/message-strip/message-strip.component.html
```diff
@@ -1,3 +1,13 @@
...
+<span
+    *ngIf="_customAriaLabel"
+    class="fd-accessible-hidden-message"
+    [id]="_customAriaLabelId"
+    [attr.aria-label]="_customAriaLabel"
+></span>
+<p class="fd-message-strip__text" [id]="_messageContentId" aria-hidden="true" role="alert">
```
mikerodonnell89, line 8: after removing the invisible element and moving its aria-label to the host, I think we can get rid of this aria-hidden="true"

libs/core/src/lib/message-strip/message-strip.component.ts
```diff
@@ -127,4 +159,20 @@ export class MessageStripComponent implements OnInit, OnChanges, CssClassBuilder
             this.class
         ];
     }
+
+    /**
+     * @hidden
+     * get the message that screen reader should speak when icons or types are present
+     */
+    _getMessage(): string {
```
mikerodonnell89, line 167: This is never read on the chrome screenreader

mikerodonnell89 (changes requested):

libs/core/src/lib/message-strip/message-strip.component.html
kavya-b: have followed your approach and tested with chrome screen reader. Thank you!
Will get in touch with a11y member to test on JAWS as well. Seems like voiceover is not a good way to test screen reader stuff!

libs/core/src/lib/message-strip/message-strip.component.ts
```diff
@@ -127,4 +159,20 @@ export class MessageStripComponent implements OnInit, OnChanges, CssClassBuilder
...
+        let message = '';
+        if (!this.type) {
+            message = 'Normal Message Strip bar';
+        } else if (this.type && this.noIcon) {
+            message = 'Message Strip ' + this.type + ' bar';
+        } else {
+            message = 'Message Strip ' + this.type + ' bar with ' + this.type + ' icon';
```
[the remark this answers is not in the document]
kavya-b: with [Mike's approach](https://github.com/SAP/fundamental-ngx/pull/5458#discussion_r641021336) the `ariaLabel` itself will now take the custom message.

libs/core/src/lib/message-strip/message-strip.component.ts
```diff
@@ -87,7 +88,16 @@ export class MessageStripComponent implements OnInit, OnChanges, CssClassBuilder
     onDismiss: EventEmitter<void> = new EventEmitter<void>();
 
     /** @hidden */
-    constructor(private _elementRef: ElementRef) {}
+    _customAriaLabel: string = null;
+
+    /** @hidden */
+    _messageContentId: string = 'fd-message-strip-content' + messageStripContentUniqueId++;
```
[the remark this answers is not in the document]
kavya-b: there was no need of these fields anymore, so have removed it.

InnaAtanasova (approved):

Betrozov (approved):

kavya-b:
This PR is blocked until we can confirm whether to add focus to Message Strip or not, in order for screen reader to read out correctly in JAWS(Windows). Please check the comments in the issue #3123 and provide your inputs. @InnaAtanasova @droshev

kavya-b:
awaiting a11y document for Message Strip from a11y team for reference.

kavya-b:
Closing this PR for now. Will reopen it if the issue gets reopened again.

Veerakarthick:
Member
[Veerakarthick](https://github.com/Veerakarthick) commented [26 minutes ago](https://github.com/SAP/fundamental-ngx/issues/3123#issuecomment-1074673441)[](https://user-images.githubusercontent.com/37998745/159398889-f5db029b-dce8-430f-8188-596f461fb615.png)
Hello Colleague,

Retest has been taken reported issue still persist the same so please check and do the needful.
Description: When the message strips are accessed, the information inside the message strips are not conveyed (screen reader)

Expected: The content inside the message strips should be conveyed properly when the message strips are accessed(screen reader)

Screenshot:
image

Retest envornment:
URL: https://deploy-preview-5458--fundamental-ngx.netlify.app/
JAWS 2021

Regards,
Karthikeyan.c

Karthickc1994:
Hello Colleague,

Retest has been taken with the below environment test results are below:

Issue: When the message strip is accessed the information within the message strip is not read out.
Reading made as:
Close Button
To activate press spacebar.
Close Button
To activate press spacebar.
Close Button
To activate press spacebar.
 Show Code Button
To activate press spacebar.

Screen shot:
image

Expected behavior:
when the message strip is accessed the information within the message strip should read out by the screen reader.

Retest Environment:
URL:[https://deploy-preview-5458--fundamental-ngx.netlify.app](https://deploy-preview-5458--fundamental-ngx.netlify.app/?utm_source=github&utm_campaign=bot_dp)
Browser: Version 101.0.4951.54 (Official Build) (64-bit)
Screen Reader: JAWS 2022.2204.20 ILM

Regards,
Karthikeyan.c

Issue #5513, opened 2021-05-27
bug: (platform) disabled menu button and split menu button opens menu on click


DeepakSap14:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.
Even when menu button/split menu button is disabled, clicking on it opens the associated menu.

#### If this is a bug, please provide steps for reproducing it.
![image](https://user-images.githubusercontent.com/57661487/119857200-000bad00-bf31-11eb-87e6-23c0dddcbcd0.png)

puru-hk:
Tested in this url https://deploy-preview-5514--fundamental-ngx.netlify.app/#/platform/split-menu-button . it is working as expected.

PR #5514, opened 2021-05-27
fix(platform): do not open menu on disabled menu button and split menu button

DeepakSap14:
Closes SAP/fundamental-ngx#5513

Even when menu button/split menu button is disabled, clicking on it opens the associated menu.

Documentation checklist:
- [na] update `README.md`
- [na] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)
- [x] Documentation Examples


sKudum (approved):

maxym-valor:
Approved!

Betrozov (approved):

platon-rov (approved):

N1XUS (approved):

libs/platform/src/lib/components/menu-button/menu-button.component.html
```diff
@@ -6,10 +6,9 @@
     [label]="label"
     [compact]="contentDensity === 'compact'"
     [glyph]="icon"
-    [disabled]="disabled"
```
dimamarksman, line 9: Why did you remove `[disabled]` button state?
You are trying to move this `is-disabled` class to the host element itself, but I believe it's not safe.
`.fd-button.is-disabled` has some css overrides but in this case you miss it.
Let's avoid it and try to find another way.

dimamarksman (changes requested):

libs/platform/src/lib/components/menu-button/menu-button.component.ts
```diff
@@ -30,10 +39,26 @@ export class MenuButtonComponent extends BaseComponent {
...
+    tabindex = 0;
+
     constructor(_cd: ChangeDetectorRef) {
         super(_cd);
     }
 
+    /** @hidden */
+    ngOnInit(): void {
+        this.menuButtonDisabled = this.disabled;
+        this.tabindex = this.disabled ? -1 : 0;
```
dimamarksman, line 59: probably we need a getter for it to keep it up to date cause `disabled` state might happen after `OnInit` phase

libs/platform/src/lib/components/menu-button/menu-button.component.html
dimamarksman: Also in this case you are introducing A11 issue as well.

DeepakSap14: keeping [disabled] now.
needed to disabled cursor/pointer-event at menu-button level, to avoid click when disabled. corrected it now.

libs/platform/src/lib/components/menu-button/menu-button.component.ts
DeepakSap14: using getter and setter now.

dimamarksman: It's not what I meant.
In your case when `disabled` gets changed after OnInit `tabindex` and `menuButtonDisabled` remains the same and doesn't reflect this change. Please correct me if I'm wrong.

What I mean:
```typescript
@HostBinding('class.menu-button-disabled')
get  menuButtonDisabled(): boolean {
    return this.disabled;
}

get tabindex(): number {
    return this.disabled ? -1 : 0;
}
```

DeepakSap14: okay.. understood your point. have corrected the code now.

dimamarksman (approved):
Thanks

maxym-valor:
Approved!

InnaAtanasova (dismissed):
`menu-button-disabled` this is not a valid class name. It should be coming from fund-styles and have the `fd-` prefix.

DeepakSap14:
> `menu-button-disabled` this is not a valid class name. It should be coming from fund-styles and have the `fd-` prefix.

@InnaAtanasova 
this class needed in platform only. should i create a component in styles for 1 class? what do you think?
corrected class name.

droshev:
Closing old PRs - feel free to reopen it if/when needed

Issue #4643, opened 2021-02-11
Core: Grid List - clicking on the edit button makes the whole card active


Betrozov:
#### Is this a bug, enhancement, or feature request?
bug
#### Briefly describe your proposal.
Clicking on the `edit` button makes the whole card active
#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
latest
#### Is there anything else we should know?
Screenshot:
<img width="1075" alt="Screen Shot 2021-02-10 at 3 02 26 PM" src="https://user-images.githubusercontent.com/39598672/107565632-dc0aa280-6bb1-11eb-997b-26a6d79713ac.png">

PR #5520, opened 2021-05-28
fix(core): click on edit button in Grid List active card

Betrozov:
Closes SAP/fundamental-ngx#4643

issue about Grid list clicking on different elements of active card has been fixed.

Documentation checklist:
- [X] update `README.md`
- [X] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)
- [X] Documentation Examples


libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.html
```diff
@@ -187,6 +205,8 @@
                 label="Locked"
                 [compact]="true"
                 (click)="_clickOnLocked($event)"
+                (mouseover)="_isActive = false"
```
N1XUS, line 208: I would move this assignment into ts file, so we could apply ChangeDetector if needed later

libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.ts
```diff
@@ -118,6 +118,12 @@ export class GridListItemComponent<T> implements OnChanges, AfterViewInit, OnDes
     @Input()
     isNavigated = false;
 
+    /** @hidden
+     * The active state of the list item.
+     * If set to true, the whole card has active state. Becomes false only when the Edit button is clicked.
+     */
+    _isActive = true;
```
N1XUS, line 125: only private properties should have _ prefix

platon-rov: Would not agree, methods without underscore at the start are **public**, so the user can use them to work with components programmatically and all others should be started with an underscore. But, of course, some of them are public in the meaning of the component (used in the template). 

What do you think?

Betrozov: agree with @platon-rov 
and it is already discussed and approved but not updated the wiki

Smiranin: @Betrozov @platon-rov If we use private variable inside template we will get error during AOT compilation. Angular disallow using private vars inside template. We have Code Contribute Guide which [tells](https://github.com/SAP/fundamental-ngx/wiki/Code-Contribution-Guideline#method-and-variable-naming-convention) that we should use underscore as prefix for private vars.We shouldn't use underscore as a prefix for public variables.

libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.html
```diff
@@ -199,6 +219,8 @@
                 label="Draft"
                 [compact]="true"
                 (click)="_clickOnDraft($event)"
+                (mouseover)="_isActive = false"
+                (mouseout)="_isActive = true"
```
Smiranin: I don't know which rules applied in current project, but on my previous project we had a rule: don't change state of component inside template.

libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.ts
Betrozov: @Smiranin
I'm talking about using `_` prefix for private variables and for variables that should be used in template but hidden from outside

libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.html
Betrozov: already updated

libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.ts
Smiranin: I had talk with Yury and as I figure out there is one addition rule related to @hidden variables: "use underscore prefix for @hidden variable". 
In this case need to update Code Contribute Guide.

**But,** In my opinion, it is better to add an new prefix for hidden variables if they need to be labeled in a special way. And do not interfere with them with private ones, because this is confusing.

Smiranin (approved):

dimamarksman:
Do we really need this fix?
We have `fundamental-styles` and there I can reproduce the same `active` state by clicking on the action button.
At first glance removing the item's `type` class on `mouseover` looks a little bit hacky.

dimamarksman:
Right now clicking on `checkbox` or `radio` triggers the `press` event but clicking on the actions button prevents it.
@Betrozov  Is it by design?

Betrozov:
> Do we really need this fix?
> We have `fundamental-styles` and there I can reproduce the same `active` state by clicking on any action button.
> At first glance removing the item's `type` class on `mouseover` looks a little bit hacky.
> 
> image

https://experience.sap.com/fiori-design-web/grid-list/#click-an-item

platon-rov (approved):

InnaAtanasova (approved):

N1XUS (approved):

maxym-valor:
Approved!

Issue #5275, opened 2021-05-03
bug: action sheet compact items change to cozy after being clicked


mikerodonnell89:
#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)

v0.29.1

#### If this is a bug, please provide steps for reproducing it.

Open the compact action sheet and click an item.  The popover will close. Open the action sheet again and you will see that the clicked item has changed to cozy size.

![Screen Shot 2021-05-03 at 2 10 28 PM](https://user-images.githubusercontent.com/2471874/116927879-84f8f480-ac19-11eb-989e-1d6659edef9c.png)

puru-hk:
Tested in this URL https://deploy-preview-5552--fundamental-ngx.netlify.app it is working as expected.

PR #5552, opened 2021-06-03
fix(core): fix compact items change to cozy after being clicked in Action Sheet

kavya-b:
Closes SAP/fundamental-ngx#5275

This PR fixes the logic of how `compact` is used in action sheet to fix the issue 5275. 
Now, clicking on an item in the compact example, and opening it again will retain the content density.

Documentation checklist:
- [n/a] update `README.md`
- [x] Documentation Examples


libs/core/src/lib/action-sheet/action-sheet.component.ts
```diff
@@ -53,7 +53,7 @@ export class ActionSheetComponent implements AfterContentInit, AfterViewInit, On
         this._initializeChildrenState();
         this._setItemsProperties();
     }
-    private _compact = false;
+    private _compact: boolean = undefined;
```
platon-rov, line 56: It's initialized to undefined by default.

platon-rov: Maybe we also should have a getter for the `compact` property for different purposes when the customer will work with the component programmatically?

Betrozov (approved):

libs/core/src/lib/action-sheet/action-sheet.component.ts
kavya-b: it is initialized to undefined because the content density service logic is working that way.[ In these lines](https://github.com/SAP/fundamental-ngx/pull/5552/files/38089fdb1394bd59d8423e3d640ce75be8859a8e#diff-0eeca74488ee851d7cf8c80beaf7f7866f1e618f965eb8846eabc89d9a44a69eR146), there is a check whether user has defined a value for `compact`, if not, it will register to the content density service. 
Having it set to `false` by default would mean it will not register to the content density service even if user has not specified a `compact` value.

platon-rov: I fully agree with you but `private _compact: boolean` also will initialize `_compact` with `undefined` value.

kavya-b: ah, yes, i had it like that before but then felt i should initialize it to something. Will revert the `undefined` :)

kavya-b: have made the changes, please check. @platon-rov

InnaAtanasova (approved):

platon-rov (approved):

puru-hk:
Tested this issue 5275. it is working fine.

puru-hk (approved):

Issue #4666, opened 2021-02-15
Grid List Accessibility Issues


puru-hk:
Grid List Accessibility Issues

Attached the file
[Grid List.docx](https://github.com/SAP/fundamental-ngx/files/5980066/Grid.List.docx)

kavya-b:
- [ ] OB: on accessing the ‘Sort’ button, the table header details is not conveyed., highlighted in red.
EB: Ideally, all visually visible content should be conveyed to the screen reader user.
- [ ] OB: on accessing the card the second tab there is an additional information
EB: Ideally, the additional information must be removed that may be confusing for the user
- [ ] OB: on accessing the Cards, the tile count is not read out.
EB: Ideally, The list count must be read for the screen reader user
- [ ] OB: on accessing the “sort’ button, the table header details ‘ Products(7) is not read
EB:Ideally, the table header details must be read for the screen reader user
- [ ] OB:On accessing the ‘Sort’, ’Filter’, ’Group’, ’Navigation’ buttons, there is no action performed by using ‘Enter/Space Bar’ key.
EB: Ideally, the Button must be activated using Enter/Space Bar key.
**From Kavya: assuming that click/enter action is missing. Currently logs are placed in console, will replace these with alerts**

Sanjay-1312:
Issue 1:
OB: Additional information is conveyed.
![image](https://user-images.githubusercontent.com/78714518/122026930-49407580-cde8-11eb-8071-3e0d993b8293.png)
EB: ideally, the table name Products(6) and the Sort button information should be read.

Issue 2:
Fixed 

issue 3:
Fixed

Issue 4:
![image](https://user-images.githubusercontent.com/78714518/122028004-37130700-cde9-11eb-8705-0b4593ee3d37.png)
OB: Additional information is conveyed.
EB: ideally, the table name Products(7) and the Sort button information should be read.

Issue 5:
Fixed

The cards are not navigable by arrow keys.

kavya-b:
> Issue 1:
> OB: Additional information is conveyed.
> ![image](https://user-images.githubusercontent.com/78714518/122026930-49407580-cde8-11eb-8071-3e0d993b8293.png)
> EB: ideally, the table name Products(6) and the Sort button information should be read.
> 
[...]

PR #5643, opened 2021-06-09
fix(core): fix a11y for Grid List component

kavya-b:
Closes SAP/fundamental-ngx#4666

This PR does the following:
- adds appropriate `aria-labelledby` fields to convey the header details
- for single select mode, radio button info was being read; therefore hiding the display of radio button from screen readers.
- added appropriate roles for the list items count to be read.
- clicks on buttons were not performing any action according to the issue: console logs were present, but added alerts instead so that it looks like an action has taken place.
- added `aria-selected` attribute conditionally to read for the different grid list modes.

Documentation checklist:
- [n/a] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/apps/docs/src/app/core/component-docs/grid-list/examples/default/grid-list-example.component.html
+++ b/apps/docs/src/app/core/component-docs/grid-list/examples/default/grid-list-example.component.html
@@ -1,25 +1,36 @@
 <fd-grid-list>
-    <fd-grid-list-title-bar title="Products">
+    <fd-grid-list-title-bar [title]="'Products'">
         <span fd-grid-list-title-bar-additional-title-item [innerText]="'(' + list.length + ')'"></span>
 
         <fd-grid-list-title-bar-spacer></fd-grid-list-title-bar-spacer>
 
-        <button fd-button fdType="transparent" glyph="sort" aria-label="Sort" title="Sort" [compact]="true"></button>
+        <button
+            fd-button
+            fdType="transparent"
+            glyph="sort"
+            aria-labelledby="grid-list-title-bar-1 grid-list-button-1"
+            aria-label="Sort"
+            title="Sort"
+            [compact]="true"
+            (click)="showAlert('Sort')"
+        ></button>
         <button
             fd-button
             fdType="transparent"
             glyph="filter"
-            aria-title="Filter"
+            aria-label="Filter"
             title="Filter"
             [compact]="true"
+            (click)="showAlert('Filter')"
         ></button>
         <button
             fd-button
             fdType="transparent"
             glyph="group-2"
-            aria-title="Group"
+            aria-label="Group"
             title="Group"
             [compact]="true"
+            (click)="showAlert('Group')"
         ></button>
     </fd-grid-list-title-bar>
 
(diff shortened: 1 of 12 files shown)


apps/docs/src/app/core/component-docs/grid-list/examples/default/grid-list-example.component.html
```diff
@@ -1,25 +1,36 @@
 <fd-grid-list>
-    <fd-grid-list-title-bar title="Products">
+    <fd-grid-list-title-bar [title]="'Products'">
```
InnaAtanasova, line 2: do you know why the binding is not working? It should work with `title="Products"`

libs/core/src/lib/grid-list/components/grid-list/grid-list.component.html
```diff
@@ -9,7 +9,7 @@
     [class.fd-grid-list--mode-single-select-right]="selectionMode === 'singleSelectRight'"
     [class.fd-grid-list--mode-multi-select]="selectionMode === 'multiSelect'"
 >
-    <div class="fd-container">
+    <div class="fd-container" role="listbox">
```
InnaAtanasova, line 12: There are keyboard requirements associated with this role which are not handled in our component. Please check the https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role#keyboard_interactions 

We are missing the "default" focus and the navigation using the arrows.

apps/docs/src/app/core/component-docs/grid-list/examples/default/grid-list-example.component.html
kavya-b: The simple binding works but shows a hover tooltip across the entire title bar, which I was trying to avoid.

libs/core/src/lib/grid-list/components/grid-list/grid-list.component.html
kavya-b: Probably a separate issue can be created for these, since the current structure makes it difficult to add `aria-activedescendant` and `aria-multiselectable`. The current structure does not place the list items under a separate element, and they are siblings with the `title-bar` component; placing `aria-activedescendant` and `aria-multiselectable` on the parent listbox has no effect on the list items. The structure needs to be modified to put the items under another container(which can then serve as a `ul`/listbox).
image

InnaAtanasova: I understand but in this case either a new issue needs to be created for addressing the a11y issues for this component or the current PR template should say `Part of #4666` not `Closes #4666`. Because with the current PR we are not adding all required aria roles and keyboard navigation. Please let me know so I know how to proceed with the PR.

kavya-b: @InnaAtanasova raised https://github.com/SAP/fundamental-ngx/issues/5655 for the other a11y issues.

InnaAtanasova (approved):

N1XUS (approved):

olshansky (approved):

aberikashvili (approved):

Betrozov (approved):

puru-hk:
Issues 1, 4 and 5 will be tracking in issue 5655. Hence this issue can be closed

puru-hk (approved):

Issue #3143, opened 2020-08-30
Accessibility Bug: When the ‘Open Dialog’ is activated, ‘popover Example’ is opened and no announcement is made when the input field within the popover is accessed


puru-hk:
Fundamental core component: "Popover"

Description: When the ‘Open Dialog’ is activated, ‘popover Example’ is opened and no announcement is made when the input field within the popover is accessed

Expected: When the input field within the popover is accessed, the UI type should be conveyed.

Attached screen shot:

![image](https://user-images.githubusercontent.com/32538291/91651452-7ff00e00-eaaa-11ea-939f-dc5d982d844a.png)

Sanjay-1312:
![image](https://user-images.githubusercontent.com/78714518/119811156-675f3800-bf04-11eb-8d9e-446967706834.png)
![image](https://user-images.githubusercontent.com/78714518/119811190-70500980-bf04-11eb-83e2-83634b9ebdcf.png)
![image](https://user-images.githubusercontent.com/78714518/119811206-75ad5400-bf04-11eb-8666-9822783dbb54.png)

Hello Colleagues,

The issue still persists.
On activating the 'Open Dialog' button, the dialog appears and dialog name is not read.
On activating the 'Click me' button, the pop up appears and the pop message should have focus and should be read.
The name of the UI is not read and  no tool tip for the 'Close' button.
So please fix the issue .

Thanks and Regards
Sanjay

Lokanathan-k:
hi Sanjay,

the netlify URL is not updated with the latest build can u test it in the below URL.

https://deploy-preview-5464--fundamental-ngx.netlify.app/#/core/popover

Sanjay-1312:
Hello Colleagues,

The issue persists the dialog name is read twice and for close button tool tip is not provided.
On activating the 'Click Me' button the pop up appears and no information is read.
So please fix the issue.
The pop should receive focus and should be read to the user.
URL: https://deploy-preview-5464--fundamental-ngx.netlify.app/#/core/popover
JAWS 2021
Screenshot:
![image](https://user-images.githubusercontent.com/78714518/119929142-988a4780-bf9a-11eb-87b8-62adb07cf241.png)
![image](https://user-images.githubusercontent.com/78714518/119929156-9fb15580-bf9a-11eb-94f7-f45bb158ba91.png)

sKudum:
[...]

PR #5693, opened 2021-06-17
fix(core): add ariaLabelledby to the popover header for popover component and adding Tabindex to the list items

Lokanathan-k:
Closes: https://github.com/SAP/fundamental-ngx/issues/3143

add ariaLabelledby to the dialog header for popover component to handle accessibility and work along and introduced the initfocus to focus the click me button 

- [NA] tests for the changes that have been done

Documentation checklist:
- [NA] update `README.md`
- [x] Documentation Examples


puru-hk (approved):

apps/docs/src/app/core/component-docs/popover/examples/popover-dialog/popover-dialog-example.component.html
```diff
@@ -1,7 +1,7 @@
 <button fd-button label="Open Dialog" (click)="openDialog(template)"></button>
 
 <ng-template #template let-dialog let-dialogConfig="dialogConfig">
-    <fd-dialog [dialogConfig]="dialogConfig" [dialogRef]="dialog">
+    <fd-dialog [dialogConfig]="dialogConfig" [dialogRef]="dialog" role="dialog">
```
olshansky, line 4: Do you really need `role="dialog"` here? `dialog.component.html` has the same role as well, looks unnecessary

Lokanathan-k: yes, we need this without 'role="dialog"' the contents are read multiple times, with this header content is read only once

olshansky: @Lokanathan-k nice to know the reason, thank you!
So we need to raise a separate issue, which would describe a double role for dialog component _(for `<fd-dialog>` and inside fd-dialog template)_ and it needs investigating (and resolve if possible)

olshansky (approved):

apps/docs/src/app/core/component-docs/popover/examples/popover-dialog/popover-dialog-example.component.html
Lokanathan-k: no need to raise issue, this issue resides in the Jaws Application for screen reading, on defining the role eliminates the duplicate reading.

Betrozov (approved):

sKudum (changes requested):
Hi @Lokanathan-k , 

     Considering all the elements will give complete accessibility to this example or else it looks partially accessible. What do you say?

puru-hk:
@Lokanathan-k There are other examples is not working for screen reader. Hence i have the change the label as required rework.

Lokanathan-k:
> @Lokanathan-k There are other examples is not working for screen reader. Hence i have the change the label as required rework.

@puru-hk fixed please check

sKudum:
Hi @Lokanathan-k ,

Using keyboard not able to access value help dialog button by which drop down is not showing up. Could you please have a look.

Lokanathan-k:
> 
> 
> Hi @Lokanathan-k ,
> 
> Using keyboard not able to access value help dialog button by which drop down is not showing up. Could you please have a look.

@sKudum multi input does not allow the user to open the dropdown using the key press, once the drop down is opened we it allows the keyboard to access the items in the drop, Its the limitation of the multi input

sKudum (approved):

puru-hk:
Looks fine

puru-hk (approved):

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
```diff
@@ -15,10 +15,10 @@
     <fd-popover [focusTrapped]="true" [focusAutoCapture]="true">
         <fd-popover-control>
             <button fd-button label="With Header"></button>
-        </fd-popover-control>
+        </fd-popover-control>x
```
mikerodonnell89, line 18: accidental 'x' character added here

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
```diff
@@ -15,10 +15,10 @@
...
         <fd-popover-body>
             <div fd-popover-body-header>
-                <div fd-bar [cozy]="true">
+                <div fd-bar [cozy]="true" tabindex="0">
```
mikerodonnell89, line 21: Why do we want the header to receive tab focus?

InnaAtanasova:
Screen Shot 2021-06-25 at 4 46 06 PM
Screen Shot 2021-06-25 at 4 45 56 PM
Screen Shot 2021-06-25 at 4 45 52 PM

The following focus outlines should not have this styling.

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
Lokanathan-k: @mikerodonnell89 

the focus on key arrow down press we dont get the focus to the header part the focus is directed only to the list elements and not the header

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
```diff
@@ -17,16 +17,12 @@
             <button fd-button label="With Header"></button>
         </fd-popover-control>
         <fd-popover-body>
-            <div fd-popover-body-header>
-                <div fd-bar [cozy]="true">
-                    <div fd-bar-left>
-                        <fd-bar-element>Cozy Header</fd-bar-element>
-                    </div>
-                </div>
-            </div>
```
InnaAtanasova, line 26: Why do you remove this?

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
```diff
@@ -17,16 +17,12 @@
...
             <ul fd-list>
+				<li fd-list-group-header>
+					<span fd-list-title>Cozy Header</span>
+				</li>
                 <li fd-list-item *ngFor="let option of list">
-                    <a [attr.href]="option.url" fd-list-title>{{ option.text }}</a>
+                    <span fd-list-title>{{ option.text }}</span>
```
InnaAtanasova, line 25: I don't think this is solving the problem. You can't just change the link to a span.

InnaAtanasova (changes requested):

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
Lokanathan-k: on using the fd-bar-element the focus in not directed to the header part so the change is required to migrate to list header.

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
Lokanathan-k: fixed

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
InnaAtanasova: I don't understand your answer. Please provide more info. Thank you

apps/docs/src/app/core/component-docs/popover/popover-docs.module.ts
```diff
@@ -32,6 +32,7 @@ import { BarModule } from '@fundamental-ngx/core/bar';
 import { SegmentedButtonModule } from '@fundamental-ngx/core/segmented-button';
 import { FormModule } from '@fundamental-ngx/core/form';
 import { CheckboxModule } from '@fundamental-ngx/core/checkbox';
+import { LinkModule } from '@fundamental-ngx/core';
```
InnaAtanasova, line 35: I don't think this is the correct way to import modules. Please use the new approach

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
InnaAtanasova: The header in this case is part of the Popover, it's not part of the List. You should be able to have a header without a List:
https://fundamental-styles.netlify.app/?path=/docs/components-popover--alignment

Screen Shot 2021-07-05 at 3 30 55 PM

InnaAtanasova (changes requested):

InnaAtanasova:
Also, the title of the PR mentions `dialog header` but it's not the case. Please correct accordingly. Thanks

InnaAtanasova:
A few problems:
1. The subheader focus:
Screen Shot 2021-07-06 at 12 55 41 PM

2. List elements gets double focus:
Screen Shot 2021-07-06 at 12 55 55 PM
Screen Shot 2021-07-06 at 12 55 51 PM

3. Are you sure the Header and Subhader are read by the screen reader?

Lokanathan-k:
> A few problems:
> 
> 1. The subheader focus: **yes, As per Accessibility team the  header item has to be read out.**
> 
> Screen Shot 2021-07-06 at 12 55 41 PM
> 
> 1. List elements gets double focus: **fixed**
> 
> Screen Shot 2021-07-06 at 12 55 55 PM
> 
> Screen Shot 2021-07-06 at 12 55 51 PM
> 
> 1. Are you sure the Header and Subhader are read by the screen reader?: **Yes**

Screenshot 2021-07-07 at 9 31 59 AM

InnaAtanasova:
> > A few problems:
> > 
> > 1. The subheader focus: **yes, As per Accessibility team the  header item has to be read out.**
> > 
> > Screen Shot 2021-07-06 at 12 55 41 PM
> > 
> > 1. List elements gets double focus: **fixed**
> > 
> > Screen Shot 2021-07-06 at 12 55 55 PM
> > Screen Shot 2021-07-06 at 12 55 51 PM
> > 
> > 1. Are you sure the Header and Subhader are read by the screen reader?: **Yes**
> 
> Screenshot 2021-07-07 at 9 31 59 AM

:D yes, I know it has to be focussed but don't you find something wrong? The styling maybe?

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
mikerodonnell89: tabIndex="0" makes the element tabbable, there is nothing in the header we want tabbable, the user can do this: 

Screen Shot 2021-07-14 at 3 37 58 PM

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
```diff
@@ -17,16 +17,16 @@
...
                     <div fd-bar-left>
                         <fd-bar-element>Cozy Header</fd-bar-element>
                     </div>
                 </div>
             </div>
-            <ul fd-list>
-                <li fd-list-item *ngFor="let option of list">
-                    <a [attr.href]="option.url" fd-list-title>{{ option.text }}</a>
+            <ul fd-list tabindex="-1">
+                <li fd-list-item *ngFor="let option of list" tabindex="0">
```
mikerodonnell89, line 28: Arrow keys are already working correctly here on main branch, no need for these changes

mikerodonnell89 (changes requested):
see comment

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
mikerodonnell89: tabindex="0" makes the element tabbable, there is nothing in the header we want tabbable, the user can do this:

Screen Shot 2021-07-14 at 4 26 43 PM

We should instead use aria roles to get the screenreader to read the content

mikerodonnell89:
@Lokanathan-k wondering if there must be a better approach than to make elements that should not be tabbable, tabbable... I understand making an element tabbable will get it read, however there is nothing in the header we want tabbable, the user can do this:

Screen Shot 2021-07-14 at 4 26 43 PM

We should instead use aria roles to get the screenreader to read the content

Lokanathan-k:
> @Lokanathan-k wondering if there must be a better approach than to use tabindex="0"... I understand making an element tabbable will get it read by the screenreader, however there is nothing in the header we want truly tabbable, because the user can do this:
> 
> Screen Shot 2021-07-14 at 4 26 43 PM
> 
> We should instead use aria roles to get the screenreader to read the content

>fixed the issue with aria-labelled by

InnaAtanasova (commented):
The examples with List still get double focus:
Screen Shot 2021-07-21 at 2 22 36 PM
Screen Shot 2021-07-21 at 2 22 25 PM

InnaAtanasova:
Screen Shot 2021-07-21 at 2 25 13 PM

Are you sure the Header and the subheader are read by the screen reader?

InnaAtanasova:
Screen Shot 2021-07-21 at 2 28 01 PM

Why are both Header and Subheader with the same class (fd-popover__body-header)? Subheader should have different class. `.fd-popover__body-header` class adds rounded corners on the top and results in this:


Screen Shot 2021-07-21 at 2 28 07 PM

Lokanathan-k:
> Screen Shot 2021-07-21 at 2 28 01 PM
> 
> Why are both Header and Subheader with the same class (fd-popover__body-header)? Subheader should have different class. `.fd-popover__body-header` class adds rounded corners on the top and results in this:
> 
> Screen Shot 2021-07-21 at 2 28 07 PM

fixed

Lokanathan-k:
> The examples with List still get double focus:
> Screen Shot 2021-07-21 at 2 22 36 PM
> Screen Shot 2021-07-21 at 2 22 25 PM

fixed

InnaAtanasova:
Screen Shot 2021-07-22 at 11 18 11 AM

I can still reproduce it in latest version of Chrome

Lokanathan-k:
> Screen Shot 2021-07-22 at 11 18 11 AM
> 
> I can still reproduce it in latest version of Chrome

fixed

aberikashvili (approved):
LGTM

InnaAtanasova (changes requested):
Screen Shot 2021-07-29 at 2 24 30 PM

Focus problem

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
```diff
@@ -6,7 +6,7 @@
         <fd-popover-body>
             <ul fd-list>
                 <li fd-list-item *ngFor="let option of list">
-                    <a [attr.href]="option.url" fd-list-title>{{ option.text }}</a>
+                    <a [attr.href]="option.url" fd-list-title tabindex="-1">{{ option.text }}</a>
```
mikerodonnell89, line 9: I think we need to find a way to do this automatically, so the dev doesn't have to enter this every time

mikerodonnell89 (changes requested):

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
Lokanathan-k: Screenshot 2021-08-03 at 9 58 58 PM

by default <a> tag gets focus, to over ride the focus we can add tabindex in the fd-link at the host, but the font color changes as shown in the screen shot.

current implementation 
Screenshot 2021-08-03 at 10 02 37 PM

mikerodonnell89: It could be done programmatically by drilling down through ContentChildren/ViewChildren and setting the tabindex manually.  A bit extreme but helps remove unnecessary code for the developer

puru-hk:
Looks fine

puru-hk (approved):

sKudum (approved):

Betrozov (approved):

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
InnaAtanasova: In NGX the developer should not add `tabindex="-1"`, this should be handled by fund-ngx behind the scene.

apps/docs/src/app/core/component-docs/popover/examples/popover-complex-example/popover-complex-example.component.html
```diff
@@ -17,16 +17,16 @@
...
                     <div fd-bar-left>
                         <fd-bar-element>Cozy Header</fd-bar-element>
                     </div>
                 </div>
             </div>
-            <ul fd-list>
+            <ul fd-list aria-labelledby="popover-bar-header-1">
                 <li fd-list-item *ngFor="let option of list">
-                    <a [attr.href]="option.url" fd-list-title>{{ option.text }}</a>
+                    <a [attr.href]="option.url" fd-link fd-list-title tabindex="-1">{{ option.text }}</a>
```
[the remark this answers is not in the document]
Lokanathan-k: @mikerodonnell89 @InnaAtanasova @aberikashvili fixed and screen reading is also fine
Screenshot 2021-08-06 at 6 48 22 PM

libs/core/src/lib/popover/popover.component.ts
```diff
@@ -159,6 +166,16 @@ export class PopoverComponent extends BasePopoverClass implements AfterViewInit,
         if (this.popoverControl && this.triggers.includes('click')) {
             this.popoverControl.makeTabbable();
         }
+        this.listItems.forEach((li) => {
+            const listItem = this._getHtmlElement(li);
+            const listTitle = listItem.querySelector('.fd-list__title');
+            listTitle.setAttribute('tabindex', '-1');
+        });
+    }
```
InnaAtanasova, line 174: does not work with ContentChildren/ViewChildren as Mike suggested?

Lokanathan-k: It does uses the contentChildrens as the object to access the listItems, when adding the contentChildren to the directive(ListTitleDirective) the object is not created since there is not constructor defined in the directive, this throws a runtime exception. in order to avoid these i have used ListItemComponent and added the tab index.

Screenshot 2021-08-09 at 9 14 06 AM

fkolar:
Let's keep it as it is now (tabindexes). Rather provide overall solution for both `core and platform`. When you look at the e.g. `core` in terms of `tabindex`, there is not much difference between solution from loki and what is already in master.

I would suggest, provide guideline for core  and for platform (here we have parent component for all of the components so it might be easier), when and how ot use tabindex and let's add it to the coding guideline, otherwise code in core and in platform is going to be inconsistent. 

- I would stick with binding ( in platform) as material has this
 - I would add it to the html referring this tabindex
 - For places where we know we need to turn it off the tabindex, we can write directly a value `tabindex=-1`
 
 I don't think we can find some solution that will do this magically behind the scenes. What do you think?

InnaAtanasova (approved):

InnaAtanasova:
> Let's keep it as it is now (tabindexes). Rather provide overall solution for both `core and platform`. When you look at the e.g. `core` in terms of `tabindex`, there is not much difference between solution from loki and what is already in master.
> 
> I would suggest, provide guideline for core and for platform (here we have parent component for all of the components so it might be easier), when and how ot use tabindex and let's add it to the coding guideline, otherwise code in core and in platform is going to be inconsistent.
> 
> * I would stick with binding ( in platform) as material has this
> * I would add it to the html referring this tabindex
> * For places where we know we need to turn it off the tabindex, we can write directly a value `tabindex=-1`
> 
> I don't think we can find some solution that will do this magically behind the scenes. What do you think?

It is done with 5 lines of code. 6 days ago, just above my comments about handling it magically behind the scenes, @mikerodonnell89 explained what needs to be done:
Screen Shot 2021-08-10 at 12 42 56 PM

Today he added the code:
Screen Shot 2021-08-10 at 12 40 26 PM

mikerodonnell89:
@Lokanathan-k @InnaAtanasova @fkolar  this is possible via ContentChild and no need to hardcode the list title classname, see this commit: https://github.com/SAP/fundamental-ngx/pull/5693/commits/684ac60bff4c9194278756fe118d2e30ddd391b6

However I think we should take it one step further and actually remove the tabIndex from the list title itself, if the list title is an anchor element, which I've done in this commit: https://github.com/SAP/fundamental-ngx/pull/5693/commits/bc1650885fcef546f54865a86791dd0a29e6c502

If anyone objects we can roll back the most recent commit I've linked

libs/core/src/lib/popover/popover.component.ts
Lokanathan-k: > @Lokanathan-k @InnaAtanasova @fkolar this is possible via ContentChild and no need to hardcode the list title classname, see this commit: [684ac60](https://github.com/SAP/fundamental-ngx/commit/684ac60bff4c9194278756fe118d2e30ddd391b6)
> 
> However I think we should take it one step further and actually remove the tabIndex from the list title itself, if the list title is an anchor element, which I've done in this commit: [bc16508](https://github.com/SAP/fundamental-ngx/commit/bc1650885fcef546f54865a86791dd0a29e6c502)
> 
> If anyone objects we can roll back the most recent commit I've linked

I had this solution before i had this introduced my changes as @droshev had shared comments in my Previous Pr's not to introduce changes in different component when the issue is pointed to the Specific Component. So i found the best alternative solution without introducing the change in the ListComponent.

Thank you, @mikerodonnell89 @InnaAtanasova Helping resolving the issue.

mikerodonnell89 (approved):

droshev:
@mikerodonnell89 thanks

Issue #3456, opened 2020-09-29
Accessibility: Combobox: Group headers within the combo box list are not reachable via keyboard. 


puru-hk:
Description: Group headers within the combo box list are not reachable via keyboard. 

Expected: Group headers within the combo box list should be reachable via keyboard

Screenshot:

![image](https://user-images.githubusercontent.com/32538291/94523088-5283b500-024e-11eb-9c6a-f3a1fe239d44.png)

![image](https://user-images.githubusercontent.com/32538291/94523101-557ea580-024e-11eb-8795-d80cf32f9053.png)

Sanjay-1312:
Hello Colleagues,

The issue is not fixed.
The group header name is not read.
So please fix the issue.
URL: https://fundamental-ngx.netlify.app/#/platform/combobox
JAWS 2021

Screenshot:
![image](https://user-images.githubusercontent.com/78714518/119931939-fcfbd580-bf9f-11eb-8bd1-31d43f269093.png)

Betrozov:
Hi @Sanjay-1312
please test using this link
https://deploy-preview-5206--fundamental-ngx.netlify.app/

Sanjay-1312:
The Issue still persists i can able to reproduce the issue.
The list header such as 'Vegetable' and 'Fruits' is not read
when accessed it is read as 'Open List box' when 'Vegetable is accessed and nothing is read on accessing the 'Vegetable' Header.

![image](https://user-images.githubusercontent.com/78714518/120318689-c3a7ca80-c2fd-11eb-9faa-97f4d24b6e7b.png)

URL:https://deploy-preview-5206--fundamental-ngx.netlify.app/
JAWS 2021

puru-hk:
@Betrozov Please see the above comments from the accessibility team.

olshansky:
Assigned to @OleksiyDrozdov

Sanjay-1312:
Issue:
OB: There is no visible focus provided for the 'Fruits' and 'vegetables'.
EB: Ideally, four sided focus should be provided so that user will be able to identify on which element the focus is available.
![image](https://user-images.githubusercontent.com/78714518/127625695-80e44166-9f10-4f02-9283-d91ba71f1b50.png)
![image](https://user-images.githubusercontent.com/78714518/127625778-f74de421-cdad-4c0c-ba5a-5070616503b8.png)

PR #5772, opened 2021-06-24
fix(platform): add attribute option into combobox

OleksiyDrozdov:
Closes [#3456](https://app.zenhub.com/workspaces/fundamental-library-60761c0894e25c0011e828b4/issues/sap/fundamental-ngx/3456) 

Documentation checklist:
- [n/a] update `README.md`
- [n/a] Documentation Examples


Betrozov (approved):

aberikashvili (approved):
LGTM

InnaAtanasova (changes requested):
Screen Shot 2021-06-25 at 4 23 32 PM
Screen Shot 2021-06-25 at 4 24 01 PM

Group header should be focusable so that the screen reader can read them but they should not have the interaction states (hover, active) like the other elements.

OleksiyDrozdov:
> Screen Shot 2021-06-25 at 4 23 32 PM
> 
> Screen Shot 2021-06-25 at 4 24 01 PM
> 
> Group header should be focusable so that the screen reader can read them but they should not have the interaction states (hover, active) like the other elements.





could you please clarify expected result:
Option1: as in fundamental-styles (picture1 and 2)
Option2: without underlining and with boarder dotted when user clicks/presses tab (picture 3)
Now I implemented with Option2. If it is ok please let me know and I will create PR for fundamental-style and move there styles from fundamental-ngx.

libs/platform/src/lib/components/form/combobox/combobox/combobox.component.scss
```diff
@@ -25,3 +25,12 @@
         }
     }
 }
+
+.fd-list__group-header.ng-star-inserted{
```
InnaAtanasova, line 29: You can't have selector `.ng-star-inserted`. It's Angular specific and if we add it in Fundamental-styles it won't work in the other libs (fundamental-react and fundamental-vue)

InnaAtanasova (changes requested):

InnaAtanasova:
> > Group header should be focusable so that the screen reader can read them but they should not have the interaction states (hover, active) like the other elements.
> > picture1
> > 
> > picture2
> > 
> > picture3
> > 
> > @InnaAtanasova
> > could you please clarify expected result:
> > Option1: as in fundamental-styles (picture1 and 2)
> > Option2: without underlining and with boarder dotted when user clicks/presses tab (picture 3)
> > Now I implemented with Option2. If it is ok please let me know and I will create PR for fundamental-style and move there styles from fundamental-ngx.

Hey @OleksiyDrozdov it's still not the correct implementation. Please check Fundamental-styles and Combobox with grouping: https://fundamental-styles.netlify.app/?path=/docs/patterns-combobox-input--cozy-and-compact
Screen Shot 2021-07-05 at 3 18 36 PM

The group header (**Fruits** and **Vegetables** in your case) should not get the focus outline, it should not have any sort of interaction from the user - hover, selected, active state, etc. This is just an element for information purpose. If you need the screen reader to be able to read it by focusing the element, you then need to visually remove the focus line from this element. You can also check here how it should behave: https://sapui5.hana.ondemand.com/#/entity/sap.m.ComboBox/sample/sap.m.sample.ComboBoxGrouping

libs/platform/src/lib/components/form/combobox/combobox/combobox.component.html
```diff
@@ -96,7 +96,7 @@
         <ng-container *ngIf="isGroup">
             <ng-container *ngFor="let group of _suggestions">
                 <ng-container *ngIf="!groupItemTemplate">
-                    <li fd-list-group-header role="group" tabindex="0">
+                    <li fd-list-group-header role="option" tabindex="0">
```
mikerodonnell89, line 99: @InnaAtanasova are we sure that this element needs to be tab-able to be compliant with accessibility requirements? I think just using roles to have the screenreader read it out may be sufficient

mikerodonnell89 (changes requested):
see comment

Betrozov (approved):

libs/platform/src/lib/components/form/combobox/combobox/combobox.component.html
OleksiyDrozdov: if we use role="group" screenreader will not read it

mikerodonnell89: I'm saying, can we remove `tabindex="0"` ?

OleksiyDrozdov: tabindex was removed

InnaAtanasova (approved):

libs/core/src/lib/list/directives/list-group-header.directive.spec.ts
```diff
@@ -39,6 +39,6 @@ describe('ListHeaderComponent', () => {
     });
 
     it('should assign class', () => {
-        expect(component.ref.nativeElement.className).toBe('fd-list__item fd-list__group-header');
+        expect(component.ref.nativeElement.className).toBe('fd-list__group-header');
```
aberikashvili, line 42: How trustworthy is this assertion?
There can be only 1 class name there?

OleksiyDrozdov: @aberikashvili there can be only 1 class name because I deleted class 'fd-list__item' as it was adding not nessasary styles that were described at the above comments.

mikerodonnell89 (approved):

puru-hk:
Issue:
OB: There is no visible focus provided for the 'Fruits' and 'vegetables'.
EB: Ideally, four sided focus should be provided so that user will be able to identify on which element the focus is available.
image
image

OleksiyDrozdov:
> Issue:
> OB: There is no visible focus provided for the 'Fruits' and 'vegetables'.
> EB: Ideally, four sided focus should be provided so that user will be able to identify on which element the focus is available.
> image
> image

@puru-hk implementation was done according to Inna's comments please see below: https://github.com/SAP/fundamental-ngx/pull/5772#issuecomment-874293132

puru-hk:
Looks fine..

puru-hk (approved):

Issue #5457, opened 2021-05-24
bug: (click) event goes through on disabled fd-button-bar component


mikerodonnell89:
Realized while testing the wizard in dialog example that the `<fd-button-bar>` component will still allow for clicks when `disabled="true"`.  To demonstrate - go to the 2nd step on the wizard in dialog example - "next step" button appears disabled visually while the user has not entered the required fields, however clicking it still progresses to the third step.

Smiranin:
@mikerodonnell89 I tested it and not reproduced.(see video) Can you retest it or provide more specific info?!
https://user-images.githubusercontent.com/14183958/120173275-8c0d2580-c20c-11eb-9de9-2644bf4aca59.mp4

KeszleriJudit:
@Smiranin I have the same issue with the fd-button-bar. In the wizard it's just a simple button tag, but in the Dialog Wizard you can reproduce the same issue, because fd-button-bar is used there.

PR #5780, opened 2021-06-24
fix(core): set real disabled state for button-bar element

Smiranin:
Closes SAP/fundamental-ngx#5457

Users expect that if disabled state set on fd-button-bar, any clicks wouldn't work on element, but it didn't work because disabled attribute works only for specific html elements. So I set real disabled state for fd-button-bar element if disabled is true.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/bar/button-bar/button-bar.component.ts
+++ b/libs/core/src/lib/bar/button-bar/button-bar.component.ts
@@ -51,6 +51,11 @@ export class ButtonBarComponent extends BaseButton implements OnInit, OnDestroy 
     @HostBinding('class.fd-bar__element')
     _barElement = true;
 
+    @HostBinding('style.pointer-events')
+    get pointerEvents(): string {
+        return this._disabled ? 'none' : 'auto';
+    }
+
     /** @hidden */
     @ViewChild(ButtonComponent)
     _buttonComponent: ButtonComponent;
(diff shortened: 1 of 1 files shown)


platon-rov (approved):

package-lock.json
```diff
@@ -20084,7 +20084,7 @@
           "dependencies": {
             "get-stream": {
               "version": "3.0.0",
-              "resolved": "https://registry.npmjs.org/get-stream/-/get-stream-3.0.0.tgz",
+              "resolved": "http://registry.npmjs.org/get-stream/-/get-stream-3.0.0.tgz",
```
olshansky, line 20087: Could you please revert `package-lock.json` and re-test after `npm ci`?

libs/core/src/lib/bar/button-bar/button-bar.component.ts
```diff
@@ -51,14 +51,21 @@ export class ButtonBarComponent extends BaseButton implements OnInit, OnDestroy
     @HostBinding('class.fd-bar__element')
     _barElement = true;
 
+    @HostBinding('style.pointer-events') get disabledState(): string {
+        return this._disabled ? 'none' : 'auto';
+    }
```
dimamarksman, line 56:
```suggestion
    @HostBinding('style.pointer-events') 
    get pointerEvents(): string {
        return this._disabled ? 'none' : 'auto';
    }
```

aberikashvili (approved):
LGTM

package-lock.json
Smiranin: @olshansky @Betrozov Why should I revert this file if it doesn't add to `.gitignore`? I have just run `npm i`.

libs/core/src/lib/bar/button-bar/button-bar.component.ts
Smiranin: @dimamarksman fixed

package-lock.json
dimamarksman: package-lock.json is not ignored intentionally, we need it to keep all of us with the same dependencies.
we always run `npm ci` in such cases.

dimamarksman (approved):

Issue #3155, opened 2020-08-30
Accessibility Bug: This nav does not have a unique accessible name as required when two or more nav elements are present.


puru-hk:
Fundamental core component: "Pagination"

Description: This nav does not have a unique accessible name as required when two or more nav elements are present.

![image](https://user-images.githubusercontent.com/32538291/91651833-17eff680-eaaf-11ea-9801-bb3765d4e66a.png)

Expected: When there is ambiguity about an ARIA region, landmark or HTML section, developers must ensure that these elements contain meaningful accessible labels to assist users of screen readers to identify and distinguish the section, region or landmark

Screenshot 1:

![image](https://user-images.githubusercontent.com/32538291/91651840-348c2e80-eaaf-11ea-95d4-a43d0a77c882.png)

Screenshot 2:

![image](https://user-images.githubusercontent.com/32538291/91651846-3bb33c80-eaaf-11ea-9155-13e967b98825.png)

kavya-b:
This seems to be fixed with https://github.com/SAP/fundamental-ngx/pull/5460.
@Sanjay-1312 @puru-hk please verify with latest netlify https://fundamental-ngx.netlify.app/#/core/pagination

Sanjay-1312:
![image](https://user-images.githubusercontent.com/78714518/123457934-12493b80-d602-11eb-885f-067877d5308a.png)
i can still able to reproduce the issue.

URL: https://fundamental-ngx.netlify.app/#/core/pagination

PR #5845, opened 2021-06-29
fix(core): add ariaLabel to Pagination component

kavya-b:
Closes SAP/fundamental-ngx#3155

Each `nav` element is supposed to have a label [if there are multiple `nav` landmarks in a page.](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role#labeling_landmarks). This PR brings in an `ariaLabel` property to address the issue.
Other minor issues like maintaining order of heading element, and clearer label for the 'items per page' buttons are also handled from the documentation side.

Documentation checklist:
- [n/a] update `README.md`
- [x] Documentation Examples


InnaAtanasova (approved):

platon-rov (approved):

libs/core/src/lib/pagination/pagination.component.ts
```diff
@@ -137,6 +137,9 @@ export class PaginationComponent implements OnChanges, OnInit, OnDestroy {
     @Input()
     pageLabel = 'Page';
 
+    /** Aria label for the navigation element */
+    @Input()
+    ariaLabel = 'Pagination with id ' + this.id;
```
dimamarksman, line 142: Why do we use `Pagination with id {id}` as the default one?

kavya-b: We need some label by default when multiple `<nav>`s are used and user does not specify `ariaLabel`. From the issue description and [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role#description),  it is not very clear whether the label has to be unique or not. To be on the safe side, I let it be unique by including the `id`.
image

dimamarksman: just `Pagination`?

kavya-b: just `Pagination` also works when I tried, but if the expectation is that each label must be unique, wouldn't having just `Pagination` reflect the same on all the different `<nav>`s?

dimamarksman: So if pagination must be unique then lib consumer must provide a unique descriptive `aria-label`, this is how I see it.
component.id is not descriptive by itself, it just generates something like `fd-pagination-0` `fd-pagination-1` and so on.
By default we would have `Pagination with id fd-pagination-0`, as for me it looks weird.

kavya-b: I see, I guess it makes sense that we leave it to consumer to provide unique labels, if necessary. Have kept it as just `Pagination` now, please take a look.

dimamarksman (approved):

puru-hk:
Looks fine.

puru-hk (approved):

Issue #5059, opened 2021-03-31
Date-/Datetime-/Time-picker: format-value validation cannot be checked and not normal behaviours


zaheral:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
The format-value validation cannot be checked when a user is entering the values manually instead of using the calendar. Some additional invalid behaviours are also found. See more details in this stackblitz example (point 3): https://stackblitz.com/edit/fundamental-ngx-date-time-fields?embed=1&file=src/app/datetime-program-example.component.ts

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
"@fundamental-ngx/core": "^0.28.1"

#### If this is a bug, please provide steps for reproducing it.
Checkout the above mentioned example in stackblitz

#### Please provide relevant source code if applicable.
Checkout the above mentioned example in stackblitz

zaheral:
Hi @droshev & @JKMarkowski 
do we have news here?

PR #5851, opened 2021-06-30
fix(core): handle invalid date inside date's components

Smiranin:
Closes SAP/fundamental-ngx#5059

This request fixed this issues(read point 3) [stackblitz](https://stackblitz.com/edit/fundamental-ngx-date-time-fields?embed=1&file=src/app/datetime-program-example.component.ts).
What was done:

- Implemented standard behavior(as it works for Date calss of native JS or moment.js) for invalid date. Now we don't get error exception if we pass invalid date during creating instance of FdDate class. If FdDate instance contains invalid date, we get NaN value for all number formats and string  message "Invalid Date" for string format.
- Adapted date's components for new behavior 
- Fixed several bugs of date's components.

**BREAKING CHANGES:**

FdDate class invalid date handling:
- Before: 
You set invalid date params -> You are getting throwing error.
- After:
You set invalid date params -> You are getting an invalid date instance.

DateTimePickerComponent invalid date handling:
- Before: 
You type in input invalid date format -> You are getting date instance with current date.
- After:
You type in input invalid date format -> You are getting an invalid date instance.

DatePickerComponent invalid date handling:
- Before: 
You type in input invalid date format -> You are getting null
- After:
You type in input invalid date format -> You are getting an invalid date instance.

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/datetime-picker/datetime-picker.component.ts
+++ b/libs/core/src/lib/datetime-picker/datetime-picker.component.ts
@@ -490,18 +490,13 @@ export class DatetimePickerComponent<D> implements OnInit, OnDestroy, OnChanges,
      * Function that provides support for ControlValueAccessor that allows to use [(ngModel)] or forms
      */
     writeValue(selected: D): void {
-        if (!this._dateTimeAdapter.isValid(selected)) {
-            this._inputFieldDate = '';
-            this._changeDetRef.detectChanges();
-            return;
-        }
         this.date = selected;
-        this._setTempDateTime();
+        this._isInvalidDateInput = !this._isModelValid(this.date);
         if (this.isCurrentModelValid()) {
+            this._setTempDateTime();
             this._refreshCurrentlyDisplayedCalendarDate(this.date);
-            this._setInput(this.date);
         }
-        this._changeDetRef.detectChanges();
+        this._setInput(this.date);
     }
 
     /**
@@ -577,27 +572,19 @@ export class DatetimePickerComponent<D> implements OnInit, OnDestroy, OnChanges,
      * validation the results are different. It also changes to state of _isInvalidDateInput.
      */
     handleInputChange(inputStr: string): void {
-        const date = this._dateTimeAdapter.parse(inputStr, this._dateTimeFormats.parse.dateTimeInput);
-        this._isInvalidDateInput = !this._isModelValid(date);
+        if (!inputStr) {
+            this._isInvalidDateInput = !this.allowNull;
+            this.onChange(null);
+            return;
+        }
+        this.date = this._dateTimeAdapter.parse(inputStr, this._dateTimeFormats.parse.dateTimeInput);
+        this._isInvalidDateInput = !this._isModelValid(this.date);
 
         if (!this._isInvalidDateInput) {
-            this.date = date;
-            this._setTempDateTime();
-            this.onChange(date);
-            this._refreshCurrentlyDisplayedCalendarDate(date);
-        } else {
-            this.onChange(this.date);
-        }
-
-        if (!inputStr && this.allowNull) {
-            this._isInvalidDateInput = false;
-            this.date = this._dateTimeAdapter.now();
             this._setTempDateTime();
             this._refreshCurrentlyDisplayedCalendarDate(this.date);
-            this.onChange(null);
-        } else if (!inputStr && !this.allowNull) {
-            this._isInvalidDateInput = true;
         }
+        this.onChange(this.date);
     }
 
     /** @hidden */
@@ -620,7 +607,7 @@ export class DatetimePickerComponent<D> implements OnInit, OnDestroy, OnChanges,
     }
 
     private _setInput(dateTime: D): void {
-        this._inputFieldDate = this._formatDateTime(dateTime);
+        this._inputFieldDate = this._isModelValid ? this._formatDateTime(dateTime) : '';
         this._changeDetRef.detectChanges();
     }
 
--- a/libs/core/src/lib/datetime/fd-date.ts
+++ b/libs/core/src/lib/datetime/fd-date.ts
@@ -1,4 +1,5 @@
-import { toIso8601 } from './fd-date.utils';
+import { isValidByParams, toIso8601 } from './fd-date.utils';
+import { INVALID_DATE_ERROR } from '@fundamental-ngx/core/utils';
 
 export class FdDate {
     /**
@@ -30,6 +31,12 @@ export class FdDate {
      * Date seconds. 0 - 59.
      */
     second: number;
+
+    /**
+     * Validation state
+     */
+    /** @hidden */
+    private _isValid: boolean;
 
     /**
      * Create FdDate instance of the current moment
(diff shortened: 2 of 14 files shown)


libs/core/src/lib/datetime-picker/datetime-picker.component.ts
```diff
@@ -490,18 +490,13 @@ export class DatetimePickerComponent<D> implements OnInit, OnDestroy, OnChanges,
...
-        this._setTempDateTime();
+        this.date = selected
+        this._isInvalidDateInput = !this._isModelValid(this.date);
         if (this.isCurrentModelValid()) {
+            this._setTempDateTime();
             this._refreshCurrentlyDisplayedCalendarDate(this.date);
-            this._setInput(this.date);
         }
-        this._changeDetRef.detectChanges();
+        this._setInput(this.date);
```
dimamarksman, line 499: what will be printed in the input field if the date is invlid?

libs/core/src/lib/datetime/fd-date.ts
```diff
@@ -31,6 +32,11 @@ export class FdDate {
      */
     second: number;
 
+    /**
+     * Validation state
+     */
+    private _isValid: boolean;
```
dimamarksman, line 38: @hidden?

libs/core/src/lib/datetime/fd-date.ts
```diff
@@ -160,24 +163,34 @@ export class FdDate {
...
+            return INVALID_DATE;
+        }
         if (!this.year || !this.month || !this.day) {
             return '';
         }
 
         return toIso8601(this);
     }
 
+    valueOf(): number {
```
dimamarksman, line 183: please add method description

libs/core/src/lib/datetime/fd-date.ts
```diff
@@ -59,6 +65,18 @@ export class FdDate {
         );
     }
 
+    static isValidByParams(params: {year: number, month?: number, day?: number, hour?: number, minute?: number, second?: number}): boolean {
```
dimamarksman, line 68: Not sure if it worths exposing to the public API.
Let's hide it from the public API

libs/core/src/lib/datetime/fd-date.ts
```diff
@@ -59,6 +65,18 @@ export class FdDate {
...
+        const {year, month, day, hour, minute, second} = params;
+        const yearLength = year.toString().length;
+        // 1, 2 and 4 length are correct length of year.
+        // e.g: one digital convert correctly 1 -> 2001; two digital converted correctly 23 -> 2023.
```
dimamarksman, line 72: native js Date converts 1 to 1901, not 2001

libs/core/src/lib/datetime/fd-date.ts
```diff
@@ -118,16 +131,6 @@ export class FdDate {
      * @param second
      */
     setTime(hour: number, minute: number, second: number): FdDate {
```
dimamarksman, line 133: let's add `isValid` calculation for this setter as well

libs/core/src/lib/utils/consts/error-messages.ts
```diff
@@ -1,3 +1,5 @@
 export const MOBILE_CONFIG_ERROR = `Missing configuration object for mobile version of the component. Consult the documentation on how to provide MobileModeConfiguration object for this component.`;
 
-export const THEME_SWITCHER_ROUTER_MISSING_ERROR = `You need to import RouterModule to enable theme settings via the url.`
+export const THEME_SWITCHER_ROUTER_MISSING_ERROR = `You need to import RouterModule to enable theme settings via the url.`;
+
+export const INVALID_DATE = 'Invalid Date';
```
dimamarksman, line 5:
```suggestion
export const INVALID_DATE_ERROR = 'Invalid Date';
```

InnaAtanasova:
The checkboxes in the PR template should be all checked or marked NA as the `Mergeable` check will always fail.
Dima's comments need to be addressed, the rest looks good.

libs/core/src/lib/datetime/fd-date.ts
InnaAtanasova: yes, should be hiddem

libs/core/src/lib/datetime/fd-date.ts
Smiranin: @dimamarksman @InnaAtanasova Should we add "hidden" to private properties? Some time ago I had a talk with @Betrozov and he told me the we add hidden for public properties which we want to mark that it uses only inside controller and template.  Perhaps I misunderstood it. Could you please describe cases when we should use @hidden?

dimamarksman: We use @hidden annotation if we want to avoid its appearing in the public documentation.
properties that start with `_` are considered as private but in order to make it accessible from a template we are forced to keep it public

libs/core/src/lib/datetime/fd-date.ts
dimamarksman: I suggest keeping it as a helper function.

libs/core/src/lib/datetime/fd-date.utils.ts
```diff
@@ -21,3 +21,34 @@ export function toIso8601(fdDate: {
...
+ * @param day 1 - 31
+ * @param hour 0 - 23
+ * @param minute 0 - 59
+ * @param second 0 - 59
+ */
+// tslint:disable-next-line:max-line-length
+export function isValidByParams(params: {year: number, month?: number, day?: number, hour?: number, minute?: number, second?: number}): boolean {
+    const { year, month, day, hour, minute, second } = params;
+    if (
+        year < 1 || year > 9999
```
dimamarksman, line 37: Why do you find years below 1 and upper 9999 as invalid years?
With native `Date` I'm able to create such dates and `date` instance is valid.

libs/core/src/lib/datetime/fd-datetime-adapter.spec.ts
```diff
@@ -413,12 +414,12 @@ describe('FdDatetimeAdapter', () => {
         expect(adapter.isValid((NaN as unknown) as FdDate)).toBe(false);
     });
 
-    it('should not throw when attempting to format a date with a year less than 1', () => {
-        expect(() => adapter.format(new FdDate(-1, 1, 1), {})).not.toThrow();
+    it('should be invalid when attempting to format a date with a year less than 1', () => {
```
dimamarksman, line 417: Wy did you change it?
-1 year means 1 year b.c.

libs/core/src/lib/datetime/fd-datetime-adapter.spec.ts
```diff
@@ -413,12 +414,12 @@ describe('FdDatetimeAdapter', () => {
...
+        expect(adapter.format(new FdDate(-1, 1, 1), {})).toBe(INVALID_DATE_ERROR);
     });
 
-    it('should not throw when attempting to format a date with a year greater than 9999', () => {
-        expect(() => adapter.format(new FdDate(10000, 1, 1), {})).not.toThrow();
+    it('should be invalid date when attempting to format a date with a year greater than 9999', () => {
```
dimamarksman, line 421: why 9999 year should be invalid?

libs/core/src/lib/datetime/fd-date.utils.ts
dimamarksman: With these changes we are getting this:

image

libs/core/src/lib/datetime-picker/datetime-picker.component.ts
```diff
@@ -490,18 +490,13 @@ export class DatetimePickerComponent<D> implements OnInit, OnDestroy, OnChanges,
...
      */
     writeValue(selected: D): void {
-        if (!this._dateTimeAdapter.isValid(selected)) {
-            this._inputFieldDate = '';
-            this._changeDetRef.detectChanges();
-            return;
-        }
-        this.date = selected;
-        this._setTempDateTime();
+        this.date = selected
```
aberikashvili, line 493: Missing `;`

dimamarksman (approved):
Looks good.
The only thing that we are introducing `breaking changes` here since now instead of throwing error we are getting an invalid date instance.

Please add corresponding Breaking Changes description in the PR

InnaAtanasova (approved):

PR #5865, opened 2021-07-03
feat(core): bring new qr code icon

droshev:
bring new qr code icon


apps/docs/src/app/core/component-docs/icon/examples/icon-example.component.ts
```diff
@@ -7,626 +7,666 @@ import { Component } from '@angular/core';
 })
 export class IconExampleComponent {
     icons: string[] = [
```
droshev, line 9: the order changed as I kept the original order from the source

InnaAtanasova (approved):

Issue #5776, opened 2021-06-24
Card Component bugs


droshev:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.

- [ ] when list component is used inside the Card body the List Items do not get focus with Tab key. The focus goes from the Card header to the List body
![image](https://user-images.githubusercontent.com/4380815/123263317-d493c800-d4c6-11eb-94cf-2c20d2e591cb.png)

- [ ] The card counter reads "of 10 3" in RTL mode.
All browsers.
![image](https://user-images.githubusercontent.com/4380815/123263376-e4131100-d4c6-11eb-9252-16e8fff898ee.png)

Lokanathan-k:
the list body is focus is part of list component, so new issue is created to follow up 
https://github.com/SAP/fundamental-ngx/issues/5945

Lokanathan-k:
issue 1: fixed as part of the issue,
issue 2: the issue has been already addressed by @dimamarksman  and comments has been approve and closed by @droshev (https://github.com/SAP/fundamental-ngx/issues/3418#issuecomment-700573734)

PR #5885, opened 2021-07-05
docs(core): adding tabable to the cards with list

Lokanathan-k:
Closes: https://github.com/SAP/fundamental-ngx/issues/5776

adding tab index to handle the tab key accessibility from card body to card item

- [NA] tests for the changes that have been done

Documentation checklist:
- [NA] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/apps/docs/src/app/core/component-docs/card/examples/card-compact-example.component.html
+++ b/apps/docs/src/app/core/component-docs/card/examples/card-compact-example.component.html
@@ -5,8 +5,8 @@
         <span fd-card-counter>3 of 10</span>
     </fd-card-header>
     <fd-card-content>
-        <ul fd-list [noBorder]="true" [compact]="true">
-            <li fd-list-item>
+        <ul fd-list [noBorder]="true" [compact]="true" tabindex="-1">
+            <li fd-list-item tabindex="0">
                 <span fd-list-title>
                     item 1
                 </span>
--- a/apps/docs/src/app/core/component-docs/card/examples/card-example.component.html
+++ b/apps/docs/src/app/core/component-docs/card/examples/card-example.component.html
@@ -3,8 +3,8 @@
         <h2 fd-card-title>Card Title</h2>
     </fd-card-header>
     <fd-card-content>
-        <ul fd-list [noBorder]="true">
-            <li fd-list-item>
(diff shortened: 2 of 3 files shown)


platon-rov:
For some reason when you're getting on the card list first item arrow down navigation works only after 2nd click - could you please fix it if it's related to the card or create an appropriate ticket if it's the list-item issue?

mikerodonnell89 (changes requested):
I think it would be better to deal with this in the library's code rather than in the example documentation so the developer doesn't have to type this every time.  Is there a reason it has to happen here in the examples?

Lokanathan-k:
> I think it would be better to deal with this in the library's code rather than in the example documentation so the developer doesn't have to type this every time. Is there a reason it has to happen here in the examples?

the reason is, the card component body container can have any type of component inside, it could be list, or avatar or any other component, so it will be easier if this can be handled by user.

apps/docs/src/app/core/component-docs/card/examples/card-compact-example.component.html
```diff
@@ -5,8 +5,8 @@ <h3 fd-card-subtitle>Card Subtitle</h3>
         <span fd-card-counter>3 of 10</span>
     </fd-card-header>
     <fd-card-content>
-        <ul fd-list [noBorder]="true" [compact]="true">
-            <li fd-list-item>
+        <ul fd-list [noBorder]="true" [compact]="true" tabindex="-1">
+            <li fd-list-item tabindex="0">
```
InnaAtanasova, line 9: I don't think you are supposed to do this. Check List implementation.

InnaAtanasova:
> > I think it would be better to deal with this in the library's code rather than in the example documentation so the developer doesn't have to type this every time. Is there a reason it has to happen here in the examples?
> 
> the reason is, the card component body container can have any type of component inside, it could be list, or avatar or any other component, so it will be easier if this can be handled by user.

@mikerodonnell89 has a point here. This is a problem coming from List component. You can observe the behaviour there, too.  The fix should be done in the component (List) not in the documentation of Card component.

InnaAtanasova:
Please open this PR again when you address the comments. Thanks

Issue #5650, opened 2021-06-10
bug: (platform) content density for select component in platform is wrong


N1XUS:
#### Is this a bug, enhancement, or feature request?
Bug
#### Briefly describe your proposal.
When we change the content density parameter on select page, select component have some wrong appearance:
![image](https://user-images.githubusercontent.com/6586561/121529005-a7deab80-ca04-11eb-9851-a2e30069ceac.png)

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
v0.30.1
#### If this is a bug, please provide steps for reproducing it.
1. Go to https://fundamental-ngx.netlify.app/#/platform/select
2. Change content density to Condensed
3. Compare select from platform and core: https://fundamental-ngx.netlify.app/#/core/select

PR #5936, opened 2021-07-12
fix(platform): display default content density  as core for components

### LINKED ISSUE #5877: Platform components don't default to same content density value as core components
bug

There are components in platform that the default content density value is not the same as the value in core.
Examples: button, select,..

All the platform components should default to the same content density as it core

latest

sKudum:
Closes #5877

When platform button is used in any user angular application the button should have a content density as cozy by default.

Before:

After:

- [n/a] Run npm run build-pack-library and test in external application

Documentation checklist:
- [n/a] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/platform/src/lib/components/button/button.component.html
+++ b/libs/platform/src/lib/components/button/button.component.html
@@ -1,6 +1,6 @@
 <button #fdButton fd-button [attr.id]="id" [attr.title]="title" [attr.name]="name" [attr.aria-label]="ariaLabel"
     [glyphPosition]="glyphPosition" [label]="label" [type]="type" [attr.value]="value"
-    [attr.aria-selected]="ariaSelected" [attr.aria-disabled]="ariaDisabled" [compact]="contentDensity !== 'cozy'"
+    [attr.aria-selected]="ariaSelected" [attr.aria-disabled]="ariaDisabled" [compact]="contentDensity ? contentDensity !== 'cozy' : ''"
     [glyph]="glyph" [disabled]="disabled" (click)="onBtnClick($event)" [fdType]="buttonType" [class.is-disabled]="ariaDisabled"
     [ngStyle]="{ width: width }">
 </button>


droshev (changes requested):
I found 2 components and i didn't check the others. Maybe you can check the rest of them in platform and fix them all.


Hi @droshev ,

    I had tested for 40 components in platform in which button and button included components are showing contentDensity as compact by default. fixing in button will address other dependent component issues as well.

Attaching screen shot of components(presented in core and platform) tested/tried to reproduce in angular applications.

sKudum:
> I found 2 components and i didn't check the others. Maybe you can check the rest of them in platform and fix them all.


Hi @droshev ,

    I had tested for 40 components in platform in which button and button included components are showing contentDensity as compact by default. fixing in button will address other dependent component issues as well.

Attaching screen shot of components tested/tried to reproduce in angular applications.

libs/platform/src/lib/components/button/button.component.html
```diff
@@ -1,6 +1,6 @@
 <button #fdButton fd-button [attr.id]="id" [attr.title]="title" [attr.name]="name" [attr.aria-label]="ariaLabel"
     [glyphPosition]="glyphPosition" [label]="label" [type]="type" [attr.value]="value"
-    [attr.aria-selected]="ariaSelected" [attr.aria-disabled]="ariaDisabled" [compact]="contentDensity !== 'cozy'"
+    [attr.aria-selected]="ariaSelected" [attr.aria-disabled]="ariaDisabled" [compact]="contentDensity ? contentDensity !== 'cozy' : ''"
```
aberikashvili, line 3: I think it's better to extract the logic to separate method for better readability, like:
```
[compact]="isCompact()"
```

sKudum: Created function as suggested.

olshansky (approved):

Betrozov (approved):

libs/platform/src/lib/components/button/button.component.ts
```diff
@@ -92,4 +92,11 @@ export class ButtonComponent extends BaseComponent implements AfterViewInit {
     public onBtnClick($event: any): void {
         this.buttonClicked.emit($event);
     }
+
+    /**
+     *  Handles default contentDensity as core
+     */
```
dimamarksman, line 98: @hidden?

sKudum: Added @hidden.

dimamarksman (approved):
It took me a while to understand why we have this discrepancy.

Since this is a common logic for all platform components it would be good to keep `isCompactModeByContentDensity` helper function separately

sKudum:
> It took me a while to understand why we have this discrepancy.
> 
> Since this is a common logic for all platform components it would be good to keep `isCompactModeByContentDensity` helper function separately

Hi Dimitry,

  Except button for all other components in platform it is already handled (attached screen shots above for reference.)
.It could be because of it was first component created (around 2 yrs back).

N1XUS (approved):

dimamarksman:
> > It took me a while to understand why we have this discrepancy.
> > Since this is a common logic for all platform components it would be good to keep `isCompactModeByContentDensity` helper function separately
> 
> Hi Dimitry,
> 
> Except button for all other components in platform it is already handled (attached screen shots above for reference.)
> .It could be because of it was first component created (around 2 yrs back).

I noticed that the logic of how we handle `compact` in core and platform for the same component types is different.

For example checkbox component. 
In the `core` we have it compact when contentDensity services gives `this.compact = density !== 'cozy'';
In the `platform` we have it compact ONLY if `[compact]="density === 'compact'"`;

So I mean these kinds of discrepancies.

sKudum:
> > > It took me a while to understand why we have this discrepancy.
> > > Since this is a common logic for all platform components it would be good to keep `isCompactModeByContentDensity` helper function separately
> > 
> > 
> > Hi Dimitry,
> > Except button for all other components in platform it is already handled (attached screen shots above for reference.)
> > .It could be because of it was first component created (around 2 yrs back).
> 
> I noticed that the logic of how we handle `compact` in core and platform for the same component types is different.
> 
> For example checkbox component.
> In the `core` we have it compact when contentDensity services gives `this.compact = density !== 'cozy'`;
> In the `platform` we have it compact ONLY if `[compact]="density === 'compact'"`;
> 
> So I mean these kinds of discrepancies.

Got Your point, 

1). In this PR we are targeting to address  issue #5877 , that platform components should be cozy by default similar as core when used in any angular (user) application.

 2). The logic pointed above is w.r.t contentDensity service. Here we need to handle whether user uses condensed or compact it will be compact all the time (other than table as per fiori). Already separate issues are raised for them and they are in track too.

dimamarksman:
Thanks @sKudum 

1. I see that this particular pr is about the default state for the button component so that's why I approved it.
I just wanted to highlight that we need to keep such kind of code separately and reuse it among others.
2. As for me the provided example  [#5650](https://github.com/SAP/fundamental-ngx/issues/5650) is not about the issue I [mention](https://github.com/SAP/fundamental-ngx/pull/5936#issuecomment-882581052) previously. The issue I described is more generic for platform components and is relative not to `select` only. So I created [this one](https://github.com/SAP/fundamental-ngx/issues/6036) do not miss it

sKudum:
> Thanks @sKudum
> 
> 1. I see that this particular pr is about the default state for component so that's why I approved it.
>    I just wanted to highlight that we need to keep such kind of code separately and reuse it among others.
> 2. As for me the provided example  [#5650](https://github.com/SAP/fundamental-ngx/issues/5650) is not about the issue I [mentioned](https://github.com/SAP/fundamental-ngx/pull/5936#issuecomment-882581052) previously. The issue I described is more generic for platform components and is relative not to `select` only. So I created [this one](https://github.com/SAP/fundamental-ngx/issues/6036) do not miss it

_Thank You @dimamarksman , will keep track of it._

Seamoo13 (approved):

libs/platform/src/lib/components/button/button.component.html
mikerodonnell89: > I think it's better to extract the logic to separate method for better readability, like:
> 
> ```
> [compact]="isCompact()"
> ```

It is best to bind to avoid functions for data binding, and instead try to use properties or subscribe to observables with `observableName$ | async` because using functions results in the function being called every time a digest cycle is ran, which is slightly less performant than binding to a property.  It is some extra code in our library but it's better for our users in the long term.  Search through the core library for `$ | async"` to get an idea of how we prefer to handle this

mikerodonnell89 (changes requested):
see comment

libs/platform/src/lib/components/button/button.component.html
sKudum: Removed function, reused contentDensity property, which gets value from core (common) contentDensityService (subscribes the changes already.)

droshev (approved):

mikerodonnell89 (approved):

PR #5937, opened 2021-07-12
fix(core,docs): add local rtl service for notification component

### LINKED ISSUE #5515: Notification component: RTL is not working properly
Description: RTL is not working 

Expected: RTL should work properly 

Attached the screenshot.

Screenshot 2021-05-28 at 14 15 34

N1XUS:
Closes #5515 

Added locally-scoped RTL service to the notification service and generated component.

Documentation checklist:
- n/a update `README.md`
- n/a [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)
- [x] Documentation Examples
- n/a Stackblitz works for all examples

### DIFF
--- a/libs/core/src/lib/notification/notification-service/notification.service.ts
+++ b/libs/core/src/lib/notification/notification-service/notification.service.ts
@@ -1,9 +1,9 @@
-import { ComponentRef, Injectable, TemplateRef, Type } from '@angular/core';
+import { ComponentRef, Injectable, Optional, TemplateRef, Type } from '@angular/core';
 import { NotificationComponent } from '../notification/notification.component';
 import { NotificationContainer } from '../notification-utils/notification-container';
 import { NotificationConfig } from '../notification-utils/notification-config';
 import { NotificationRef } from '../notification-utils/notification-ref';
-import { DynamicComponentService } from '@fundamental-ngx/core/utils';
+import { DynamicComponentService, RtlService } from '@fundamental-ngx/core/utils';
 
 @Injectable()
 export class NotificationService {
@@ -13,7 +13,10 @@ export class NotificationService {
 
     public containerRef: ComponentRef<NotificationContainer>;
 
-    constructor(private dynamicComponentService: DynamicComponentService) {}
+    constructor(
+        private dynamicComponentService: DynamicComponentService,
+        @Optional() private _rtlService: RtlService
+        ) {}
 
     /**
      * Opens a notification component with a content of type TemplateRef or Component Type
@@ -47,7 +50,7 @@ export class NotificationService {
             content,
             NotificationComponent,
             notificationConfig,
-            { services: [notificationService, notificationConfig] }
+            { services: [notificationService, notificationConfig, this._rtlService] }
         );
 
         // Add To array
--- a/libs/core/src/lib/notification/notification-service/notification.service.ts
+++ b/libs/core/src/lib/notification/notification-service/notification.service.ts
@@ -13,8 +13,11 @@ export class NotificationService {
 
     public containerRef: ComponentRef<NotificationContainer>;
 
+    /**
+     * @hidden
+     */
     constructor(
-        private dynamicComponentService: DynamicComponentService,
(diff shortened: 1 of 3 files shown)


libs/core/src/lib/notification/notification/notification.component.ts
```diff
@@ -90,14 +96,23 @@ export class NotificationComponent extends AbstractFdNgxClass implements OnInit,
...
         this._setNotificationConfig(notificationConfig);
     }
 
     /** @hidden */
     ngOnInit(): void {
+
+        this._subscriptions.add(
+            this._rtlService?.rtl.subscribe(isRtl => {
+                this._dir = isRtl ? 'rtl' : 'ltr';
+                this.cdRef.markForCheck();
```
Smiranin, line 112: Do we have some code style guide how to use subscription? 
I mean, the subscription and the success handler must be on the same line, or vice versa, the subscription must always start on a new line, and each handler in the subscription?

Smiranin (approved):

libs/core/src/lib/notification/notification/notification.component.ts
N1XUS: I don't think that we have such code style. It was copy-pasted from different component, and as I see, majority of components use the same code style.

libs/core/src/lib/notification/notification-service/notification.service.ts
```diff
@@ -13,7 +13,10 @@ export class NotificationService {
 
     public containerRef: ComponentRef<NotificationContainer>;
 
-    constructor(private dynamicComponentService: DynamicComponentService) {}
+    constructor(
```
InnaAtanasova, line 16: Should be hidden

libs/core/src/lib/notification/notification-service/notification.service.ts
```diff
@@ -13,7 +13,10 @@ export class NotificationService {
...
+        private dynamicComponentService: DynamicComponentService,
```
InnaAtanasova, line 17: should be prefixed with `_` as it's private

libs/core/src/lib/notification/notification/notification.component.ts
```diff
@@ -90,14 +96,23 @@ export class NotificationComponent extends AbstractFdNgxClass implements OnInit,
         private cdRef: ChangeDetectorRef,
         private _router: Router,
         @Optional() private notificationConfig: NotificationConfig,
-        @Optional() private notificationRef: NotificationRef
+        @Optional() private notificationRef: NotificationRef,
```
InnaAtanasova, line 99: should be prefixed with `_`

libs/core/src/lib/notification/notification/notification.component.ts
InnaAtanasova: cdRef is also private and should be prefixed with `_`

InnaAtanasova (changes requested):

libs/core/src/lib/notification/notification-service/notification.service.ts
```diff
@@ -43,11 +49,11 @@ export class NotificationService {
         let notificationComponentRef: ComponentRef<NotificationComponent>;
 
         // Create Notification Component
-        notificationComponentRef = this.dynamicComponentService.createDynamicComponent(
+        notificationComponentRef = this._dynamicComponentService.createDynamicComponent(
             content,
             NotificationComponent,
             notificationConfig,
-            { services: [notificationService, notificationConfig] }
+            { services: [notificationService, notificationConfig, this._rtlService] }
```
dimamarksman, line 56: I suggest using the `injector` option instead.
`services` option is so unclear (not obvious what will be used as an injection token in this case). I would remove `services` option at all as the `injector` option is much more flexible and transparent

N1XUS: Done. Please check

dimamarksman (approved):

PR #6018, opened 2021-07-16
docs(core): improvements to branching wizard focus/selection logic

### LINKED ISSUE #5808: Bug: (core) Wizard - When you try to select another radio button and click 'Cancel' in confirmation pop-up, focus disappears
Bug
Wizard component, branching example:
When you selected one payment type and would like change it to another - you will get confirmation pop-up. First time if you click 'Cancel' - focus from radio button disappears(no selected radio-buttons), but it should not. Focus should return to previous selected button. Next time (second, third etc.) it will work as expected, focus will be returned to the previous button
1. Go to https://fundamental-ngx.netlify.app/#/core/wizard#Wizard-Branching
2. Open Wizard
3.  Click 'Go to the payment section page'
4. Select first radio-button
5. Select second radio-button
6. Click 'Cancel' in pop-up
Expected result: focus returned to the first radio-button
Actual result: focus disappeared, no chosen button and when we want to select radio-button we will get pop-up again

Запись_экрана_2021-06-25_в_11 19 04

mikerodonnell89:
Closes #5808 

Various improvements to the documentation for better handling of selection/deselection/focusing of the payment option radio buttons.

### DIFF
--- a/apps/docs/src/app/core/component-docs/wizard/examples/wizard-branching-example.component.html
+++ b/apps/docs/src/app/core/component-docs/wizard/examples/wizard-branching-example.component.html
@@ -48,7 +48,7 @@
                             <fd-wizard-step-indicator>2</fd-wizard-step-indicator>
                             <fd-wizard-content size="xl">
                                 Please select a payment type:
-                                <fd-radio-button
+                                <fd-radio-button #bankButton
                                     value="bank"
                                     id="radio-1"
                                     name="radio-name-1"
@@ -57,7 +57,7 @@
                                 >
                                     Bank Transfer
                                 </fd-radio-button>
-                                <fd-radio-button
+                                <fd-radio-button #creditButton
                                     value="credit"
                                     id="radio-2"
                                     name="radio-name-2"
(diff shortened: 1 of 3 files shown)


Betrozov (approved):

apps/docs/src/app/core/component-docs/wizard/examples/wizard-branching-example.component.html
```diff
@@ -112,7 +112,7 @@
 </div>
 
 <ng-template let-dialog let-dialogConfig="dialogConfig" #confirmationDialog>
-    <fd-dialog [dialogConfig]="dialogConfig" [dialogRef]="dialog">
+    <fd-dialog [dialogConfig]="dialogConfig" [dialogRef]="dialog" style="z-index: 1002;">
```
InnaAtanasova, line 115: Why did you add this? Will the user be required to add z-index in their project? If yes, we need to move this in fund-styles

N1XUS (approved):

apps/docs/src/app/core/component-docs/wizard/examples/wizard-branching-example.component.html
mikerodonnell89: @InnaAtanasova Good question, I do not believe this will be necessary for the user to add.  It is because the z-index of the "wizard overlay" class is 1000, and the z-index of the fd-bar in the previous example is 1001 - so this is to ensure the dialog is over top of both the wizard overlay and the buttons from the fd-bar in the "mobile wizard" examples

InnaAtanasova (approved):

PR #6161, opened 2021-07-30
fix(core): do not preventDefault when activeElement is select control

### LINKED ISSUE #6160: Bug(core): Select cannot be opened with keyboard
Due to https://github.com/SAP/fundamental-ngx/pull/5854/ the select popover options cannot be opened with keyboard.

mikerodonnell89:
Fixes #6160 

Fixes a bug where keyboard would no longer open the select popover body


droshev (approved):

libs/core/src/lib/popover/popover.component.ts
```diff
@@ -178,7 +179,8 @@ export class PopoverComponent extends BasePopoverClass implements AfterViewInit,
     onKeyDown(event: KeyboardEvent): void {
         if (
             this.popoverControl.elRef.nativeElement.children[0] === document.activeElement &&
-            document.activeElement.tagName !== 'INPUT' && document.activeElement.tagName !== 'TEXTAREA'
+            document.activeElement.tagName !== 'INPUT' && document.activeElement.tagName !== 'TEXTAREA' &&
+            !document.activeElement.classList.contains(SELECT_CLASS_NAMES.selectControl)
```
Betrozov, line 183:
```suggestion
            const { activeElement } = document;
            activeElement.tagName !== 'INPUT' && activeElement.tagName !== 'TEXTAREA' &&
            !activeElement.classList.contains(SELECT_CLASS_NAMES.selectControl)
```

Betrozov (changes requested):

kavya-b:
something strange happens after interacting with the mobile example:
- go to https://deploy-preview-6161--fundamental-ngx.netlify.app/#/core/select#mobile and click on the select input field to open the mobile view
- click on any item
- close the mobile view
- try to click on the select field in any other example, say [max height one](https://deploy-preview-6161--fundamental-ngx.netlify.app/#/core/select#max-height), it always opens the mobile view

libs/core/src/lib/popover/popover.component.ts
```diff
@@ -176,9 +180,12 @@ export class PopoverComponent extends BasePopoverClass implements AfterViewInit,
     /** @hidden */
     @HostListener('keydown', ['$event'])
     onKeyDown(event: KeyboardEvent): void {
+        const activeElement = document.activeElement;
         if (
-            this.popoverControl.elRef.nativeElement.children[0] === document.activeElement &&
-            document.activeElement.tagName !== 'INPUT' && document.activeElement.tagName !== 'TEXTAREA'
+            this.popoverControl.elRef.nativeElement.children[0] === activeElement &&
+            activeElement.tagName !== 'INPUT' &&
```
aberikashvili, line 186: May be it's better to extract these `INPUT`, `TEXTAREA` to constants?

Betrozov (approved):

libs/core/src/lib/popover/popover.component.ts
mikerodonnell89: I think we can count on these tagNames never changing so there is no added benefit of moving these to constants

PR #6164, opened 2021-07-31
fix(platform): table loading state improvements

### LINKED ISSUE #5623: bug(platform): disable table toolbar when table in the loading state

Bug.

When table in the loading state there is no much sense from the working toolbar as any actions should be processed after data get loaded.

0.30.0-rc.60
Chrome 91 (MacOS Big Sur)

Go to the https://fundamental-ngx.netlify.app/#/platform/table#loading-table
Tab to the search field above the table

platon-rov:

Closes #5623.

* Table rows not shown in the loading state (Fiori3 design).
* Table toolbar not interactable in the loading state.

Documentation checklist:
- [n/a] update `README.md`
- [n/a] Documentation Examples

### DIFF
--- a/apps/docs/src/app/platform/component-docs/platform-table/platform-table-examples/platform-table-loading-example.component.html
+++ b/apps/docs/src/app/platform/component-docs/platform-table/platform-table-examples/platform-table-loading-example.component.html
@@ -5,6 +5,7 @@
 <fdp-table
     [dataSource]="source"
     [loading]="loading"
+    bodyHeight="10rem"
     contentDensity="compact"
     emptyTableMessage="No data found"
 >
--- a/libs/platform/src/lib/components/table/components/table-toolbar/table-toolbar.component.html
+++ b/libs/platform/src/lib/components/table/components/table-toolbar/table-toolbar.component.html
@@ -13,7 +13,14 @@
         [hasTitle]="!!title"
         [size]="_getCozyCompactSize(size)"
     >
-        <label fd-toolbar-label [id]="tableToolbarTitleId">{{ title }} <span *ngIf="!hideItemCount">({{ counter }})</span></label>
+        <label
+            fd-toolbar-label
+            class="fdp-table__title"
+            [id]="tableToolbarTitleId"
+        >
+            {{ title }}
+            <span *ngIf="!hideItemCount">({{ counter }})</span>
+        </label>
 
         <fd-toolbar-spacer></fd-toolbar-spacer>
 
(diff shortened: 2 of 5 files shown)


KevinOkamoto (approved):

Betrozov (approved):

olshansky (approved):

N1XUS (approved):

Smiranin (approved):

dimamarksman:
Why do we need to hide rows completely during the loading state?
Looking at sapui5 examples tables rows are not hidden completely...
 
This change breaks lazyloading behaviour: 
every time the user scrolls to the end it triggers a loading, the loading removes all rows and gets it back breaking current scroll position. So the user gets forced to start scrolling from the beginning again.

dimamarksman:
also since we don't have virtual scrolling yet, if the user has a lot of rows it means every `loading` state will rerender all these rows over and over degrading performance

dimamarksman:
Also, it isn't clear enough why the table toolbar shouldn't be accessible during loading.

Let's imagine the case when user has some filter settings in the toolbar and submit certain criteria to serach items. If this request takes too much time and user doesn't want to wait for it or for some reason decides to change search criteria he just needs to change settings and submit it again. In this case the previous request will be cancelled and a new one will be sent.

platon-rov:
> Why do we need to hide rows completely during the loading state?
> Looking at sapui5 examples tables rows are not hidden completely...
> 
> This change breaks lazyloading behaviour:
> every time the user scrolls to the end it triggers a loading, the loading removes all rows and gets it back breaking current scroll position. So the user gets forced to start scrolling from the beginning again.
> Also since we don't have virtual scrolling yet, if the user has a lot of rows it means every `loading` state will rerender all these rows over and over degrading performance

It's the way how it's implemented in the Fiori 3: https://experience.sap.com/fiori-design-web/grid-table/#loading-data
But I agree with you on these points, seems like it was a mistake on my side to implement it like this.

> Also, it isn't clear enough why the table toolbar shouldn't be accessible during loading.
> 
> Let's imagine the case when user has some filter settings in the toolbar and submit certain criteria to serach items. If this request takes too much time and user doesn't want to wait for it or for some reason decides to change search criteria he just needs to change settings and submit it again. In this case the previous request will be cancelled and a new one will be sent.

Also a great catch! Seems like we didn't realize all these items during declaring the ticket.

platon-rov:
I think PR should be closed due to the reasons described in my previous and Dmitry's comments.

PR #6196, opened 2021-08-02
fix(core): fix Pagination component a11y to announce currently chosen page

### LINKED ISSUE #6096: Bug: Accessibility Testing_Core_Pagination
Accessibility: Pagination-Core Component (There are couple of accessibility issues found while testing, so I have attached error document for you reference.)
[Pagination_Core_Ariba Controls.docx](https://github.com/SAP/fundamental-ngx/files/6877053/Pagination_Core_Ariba.Controls.docx)

Pagination- Core

1.1	Message Details

Raise Ticket	
Category	
Preprocessed By	
Description	

1.2 Application & Test Environment Details

Application Name:	ARIBA Controls- Fundamental Library for Angular
URL:	https://fundamental-ngx.netlify.app/#/core/pagination

Technology:	Angular 11
Version: 	V 0.31.0-rc.125
Web browser:	Google Chrome 92.0.4515.107 (Official Build) (64-bit)
Operating System:	Microsoft Windows 10 x64
Screen Reader:	JAWS 2021.2105.53 ILM

 

Issue 1

Priority:	Medium
Checkpoint:	ACC-264

(264.1)

•	Search ‘Pagination’ from the left panel

Observed Behavior:

When the ‘page number’ is selected, the selected information ‘page 4 is selected’ not conveyed.
Expected Behavior:
The state ‘selected’ information should be conveyed when a page number is selected.
Screenshot:
image

 

Note: The above reported issue persists with JAWS 2019 as well.

Issue 2

Priority:	Medium
Checkpoint:	ACC-264

(264.1)

•	Search ‘Pagination’ from the left panel

Observed Behavior:

When the user is on the last page and upon accessing ‘next’ button in virtual pc mode, still the tutor message ‘To activate press spacebar’ is conveyed.
Expected Behavior:
When the user is on the last page and upon accessing ‘next’ button in virtual pc mode, the tutor message ‘To activate press spacebar’ is not required to be announced.

Screenshot:
image

 
Issue 3

Priority:	Medium
Checkpoint:	ACC-264

(264.1)

•	Search ‘Pagination’ from the left panel

Observed Behavior:

The information regarding the pages ‘Showing 1-15 of 50’, ‘showing 49-50 of 50’… are read out twice in virtual pc mode.
Expected Behavior:
The information regarding the pages ‘Showing 1-15 of 50’, ‘showing 49-50 of 50’… should read out only once in virtual pc mode.

Screenshot:

 
image

kavya-b:
Closes #6096 

This PR modifies the `aria-current` attribute value and uses LiveAnnouncer from CDK to announce the currently selected page.

Unfortunately, `aria-current` does not announce `"current page"` each time the page is changed; it seems to only announce the actively(or first time) selected `aria-current` element and doesn’t announce on page switch when handled dynamically. For this, there were two ways to handle this:
1. To add `current page` to the element’s aria-label attribute itself. This works okay for the pages themselves, but when we try to associate the previous and next buttons to also announce the page navigated to, it announces something like `Previous, Page 4, current page; button`; including the page’s aria-label before completely describing the previous button.
2. To use CDK’s LiveAnnouncer and announce the current page from code, in the method `goToPage` which is also called by previous and next buttons, thereby eliminating the problem explained in point 1. 

I have chosen to use LiveAnnouncer for its clarity of announcing the message in the correct order. It may seem that on selecting a page, it reads `Page 2, link; Page 2, current page`, but this is okay in my opinion because if you just tab through the pages instead, it reads `Page 2, link` first and only on pressing enter/space, it reads `Page 2, current page`(which isn’t happening currently in the latest netlify )

For the other two issues described in the issue, it has been clarified with the issue reporter that this is expected behaviour.

Documentation checklist:
- [n/a] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/libs/core/src/lib/pagination/pagination.component.html
+++ b/libs/core/src/lib/pagination/pagination.component.html
@@ -28,7 +28,7 @@
                         (keypress)="onKeypressHandler(page, $event)"
                         (click)="goToPage(page, $event)"
                         *ngIf="page !== -1; else more"
-                        [attr.aria-current]="currentPage == page"
+                        [attr.aria-current]="currentPage === page ? 'page': undefined"
                         [attr.aria-label]="pageLabel +' '+ page"
                     >
                         {{ page }}
--- a/libs/core/src/lib/pagination/pagination.component.ts
+++ b/libs/core/src/lib/pagination/pagination.component.ts
@@ -97,9 +97,14 @@ export class PaginationComponent implements OnChanges, OnInit, OnDestroy {
     @Input()
     itemsPerPageTemplate: TemplateRef<any>;
 
-    /** Label for options for items per page. */
+    /**
+     * Label for options for items per page.
+     * This property is mainly provided to support reading in the right language for screen reader.
+     * Here, Angular i18n's `$localize` tag is used with a default text of `Results per page`.
+     * Application developer is expected to pass in their own value if they use any other localization tool.
+     */
     @Input()
-    itemsPerPageLabel = 'Results per page';
+    itemsPerPageLabel = $localize`:@@corePaginationItemsPerPageLabel:Results per page`;
 
     /** Represents the options for items per page. */
     @Input()
@@ -129,21 +134,50 @@ export class PaginationComponent implements OnChanges, OnInit, OnDestroy {
     @Input()
     displayTextTemplate: TemplateRef<any>;
 
-    /** Label for the 'previous' page button. */
+    /**
+     * Label for the 'previous' page button.
+     * This property is mainly provided to support reading in the right language for screen reader.
+     * Here, Angular i18n's `$localize` tag is used with a default text of `Previous`.
+     * Application developer is expected to pass in their own value if they use any other localization tool.
+     */
     @Input()
-    previousLabel = 'Previous';
+    previousLabel = $localize`:@@corePaginationPreviousLabel:Previous`;
 
-    /** Label for the 'next' page button. */
+    /**
+     * Label for the 'next' page button.
+     * This property is mainly provided to support reading in the right language for screen reader.
+     * Here, Angular i18n's `$localize` tag is used with a default text of `Next`.
+     * Application developer is expected to pass in their own value if they use any other localization tool.
+     */
     @Input()
-    nextLabel = 'Next';
+    nextLabel = $localize`:@@corePaginationNextLabel:Next`;
 
-    /** Label for the 'Page' page button. */
+    /**
+     * Label for the 'Page' page button.
+     * This property is mainly provided to support reading in the right language for screen reader.
+     * Here, Angular i18n's `$localize` tag is used with a default text of `Page`.
+     * Application developer is expected to pass in their own value if they use any other localization tool.
+     */
     @Input()
-    pageLabel = 'Page';
+    pageLabel = $localize`:@@corePaginationPageLabel:Page `;
 
-    /** Aria label for the navigation element */
+    /**
+     * Aria label for the navigation element
+     * This property is mainly provided to support reading in the right language for screen reader.
+     * Here, Angular i18n's `$localize` tag is used with a default text of `Pagination`.
+     * Application developer is expected to pass in their own value if they use any other localization tool.
+     */
     @Input()
-    ariaLabel = 'Pagination';
+    ariaLabel = $localize`:@@corePaginationAriaLabel:Pagination`;
+
+    /**
+     * The current page label that should be read when a page is selected.
+     * This property is mainly provided to support reading in the right language for screen reader.
+     * Here, Angular i18n's `$localize` tag is used with a default text of `Current page`.
+     * Application developer is expected to pass in their own value if they use any other localization tool.
+     */
+    @Input()
+    currentPageLabel = $localize`:@@corePaginationCurrentPageLabel: Current page`;
 
     /** Event fired when the page is changed. */
     @Output()
@@ -262,8 +296,7 @@ export class PaginationComponent implements OnChanges, OnInit, OnDestroy {
         }
         this._refreshPages();
 
-        const currentPageMessage = $localize`:@@corePaginationCurrentPage:Current page`;
-        this._liveAnnouncer.announce(this.pageLabel + ' ' + page + currentPageMessage);
+        this._liveAnnouncer.announce(this.pageLabel + page + this.currentPageLabel);
 
         this.pageChangeStart.emit(page);
     }
--- a/libs/core/src/lib/pagination/pagination.component.ts
+++ b/libs/core/src/lib/pagination/pagination.component.ts
@@ -16,6 +16,7 @@ import {
     TemplateRef,
     ViewEncapsulation
 } from '@angular/core';
+import '@angular/localize/init';
 import { KeyUtil, RtlService } from '@fundamental-ngx/core/utils';
 import { Subscription } from 'rxjs';
 import { Pagination } from './pagination.model';
@@ -261,7 +262,8 @@ export class PaginationComponent implements OnChanges, OnInit, OnDestroy {
         }
         this._refreshPages();
 
-        this._liveAnnouncer.announce('Page ' + page + ', current page');
+        const currentPageMessage = $localize`:@@corePaginationCurrentPage:Current page`;
+        this._liveAnnouncer.announce(this.pageLabel + ' ' + page + currentPageMessage);
 
         this.pageChangeStart.emit(page);
     }
--- a/libs/core/src/lib/pagination/pagination.component.ts
+++ b/libs/core/src/lib/pagination/pagination.component.ts
@@ -177,7 +177,7 @@ export class PaginationComponent implements OnChanges, OnInit, OnDestroy {
      * Application developer is expected to pass in their own value if they use any other localization tool.
      */
     @Input()
-    currentPageLabel = $localize`:@@corePaginationCurrentPageLabel: Current page`;
+    currentPageAriaLabel = $localize`:@@corePaginationCurrentPageAriaLabel: Page ${this.currentPage}:page: is current page`;
 
     /** Event fired when the page is changed. */
     @Output()
@@ -296,9 +296,9 @@ export class PaginationComponent implements OnChanges, OnInit, OnDestroy {
         }
         this._refreshPages();
 
-        this._liveAnnouncer.announce(this.pageLabel + page + this.currentPageLabel);
-
         this.pageChangeStart.emit(page);
+
+        this._liveAnnouncer.announce(this.currentPageAriaLabel);
     }
 
     /**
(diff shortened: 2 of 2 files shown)


libs/core/src/lib/pagination/pagination.component.ts
```diff
@@ -257,6 +257,8 @@ export class PaginationComponent implements OnChanges, OnInit, OnDestroy {
         }
         this._refreshPages();
 
+        this._liveAnnouncer.announce('Page ' + page + ', current page');
```
mikerodonnell89, line 260: I do not think we should use the liveAnnouncer here, I think instead of doing this we need to add an `aria-selected="true"` property to the currently selected page's HTML so the screen reader will say 'page 3, link, selected'.  That will also help with language issues, the above solution will only work in for English websites

mikerodonnell89 (dismissed):

libs/core/src/lib/pagination/pagination.component.ts
kavya-b: I did look into `aria-selected` initially, but following the [w3 page](https://www.w3.org/TR/wai-aria-1.1/#aria-current), we currently use `aria-current` that is supposed to be added to navigation/pagination elements and not `aria-selected`(which announces 'Selected'). We are also using the `<nav>` element which [does not support `aria-selected`](https://www.w3.org/TR/wai-aria-1.2/#navigation), so i don't know if it is okay to use `aria-current` and `aria-selected` both.

kavya-b: @mikerodonnell89 please share your thoughts on the above comment, thanks!

kavya-b: @mikerodonnell89  using `$localize` in typescript now. Verified that on running `ng extract-i18n`, it extracts the localize id in `messages.xlf`. 
Screenshot 2021-08-09 at 3 17 42 PM

libs/core/src/lib/pagination/pagination.component.html
```diff
@@ -28,7 +28,7 @@
                         (keypress)="onKeypressHandler(page, $event)"
                         (click)="goToPage(page, $event)"
                         *ngIf="page !== -1; else more"
-                        [attr.aria-current]="currentPage == page"
+                        [attr.aria-current]="currentPage === page ? 'page': undefined"
```
platon-rov, line 31: Shouldn't we use `location` here? 

https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaCurrent#value

kavya-b: hmm.. although they say `Represents the current location, for example the current page in a breadcrumbs hierarchy` for `location`, but if you see [this w3 link](https://www.w3.org/TR/wai-aria-1.1/#aria-current), they say `A page token used to indicate a link within a set of pagination links, where the link is visually styled to represent the currently-displayed page.` which is what i followed for `aria-current` value since it even mentioned "**pagination**". For `location`, w3 says that it is `used to indicate the image that is visually highlighted as the current component of a flow chart.`

platon-rov: On the W3 site a little below they propose second meaning `Represents the current location within an environment or context`.

There is also great examples here https://www.aditus.io/aria/aria-current They say that 'page' is used to mark an active page from the main menu and 'location` to mark an active nested page.

Screenshot 2021-08-10 at 10 28 55

kavya-b: The way we use it in pagination is not by having nested links/pages. We are not expecting a pagination scenario with 1>2>3>4 nested pages; for us here each page is independent of the other in pagination component. 
- I tried using `aria-current="location"` instead of `page` but JAWS will read it as `Page x, current location link`. Note that similar to `page` it does not announce on each page switch `Page x, current location link` and only announces the first time(as described in the PR description). 
- There is another [example here](https://a11y-style-guide.com/style-guide/section-navigation.html#kssref-navigation-pagination) showing a pagination component using `aria-current="page"`
- Also, the [example](https://www.service-public.fr/particuliers/vosdroits/N23493) you pointed out is actually using `aria-current="page"` ;)

libs/core/src/lib/pagination/pagination.component.ts
mikerodonnell89: @kavya-b I am not sure about using `$localize` in our library - sets an interesting precedent, and does this also assume that the application developer using our library will be using Angular's built in localization rather than an alternative tool such as ngx-translate? I need to do some more research on that as I'm not sure.  Perhaps rather than assuming the developer needs to use English and providing an English default but give them a backup option via `$translate`, we need to make no assumptions but still provide a way for the application developer to handle their own accessibility and language requirements.

mikerodonnell89 (commented):
See my comment here: https://github.com/SAP/fundamental-ngx/pull/6196#discussion_r686189050

libs/core/src/lib/pagination/pagination.component.ts
kavya-b: @mikerodonnell89 using `$localize` in fundamental ngx would be similar to using `i18n` markers, which we are already using in ngx, so the questions you pose are equally applicable to the usage of `i18n` markers which also assume English as the default currently. 
I am also not sure I have the answers to your questions; I thought fundamental ngx would be supporting Angular's i18n by default and migrating to any other translation/localization tool would be upto the end-user. Therefore, this probably needs to be part of a bigger discussion involving more people.
@droshev @InnaAtanasova @dimamarksman @fkolar @KevinOkamoto @Betrozov @manjunathanagaraj @platon-rov  what are your thoughts on this?

manjunathanagaraj: For ngx-platform, we had a detailed discussion before rolling out the translation guidelines and publishing it in github wiki. We decided to use i18n directive for the hardcoded strings we introduce in any of the components and expect the app developers to use i18n extraction command to externalise the strings to be translated into the xlf file and then they proceed with providing values for translations.
There could be many other translation tools, but we decided to go with angular's i18n.
I hope the same can be extended to ngx-core as well so that both core and platform libraries have similar architecture for translation and it becomes easy for app developers to mix and match components from both core and platform.

If app developers decide to chose any other translation tool, they need to handle it on their own.

@droshev and team, your thoughts?

kavya-b:
> See my comment here: [#6196 (comment)](https://github.com/SAP/fundamental-ngx/pull/6196#discussion_r686189050)

Hi @mikerodonnell89  please take a look at[ Manju's](https://github.com/SAP/fundamental-ngx/issues/6196#discussion_r686495406) and [my](https://github.com/SAP/fundamental-ngx/issues/6196#discussion_r686458925) comments on the same thread.

libs/core/src/lib/pagination/pagination.component.ts
InnaAtanasova: Sending this for Core team discussion Monday. Apologies for the delay, we need to discuss it first. It's good that you had a discussion for platform, but the current PR is in Core so we need to see all the options and then come back with an answer.

droshev (dismissed):
@kavya-b @manjunathanagaraj  we would need some time to discuss and revisit that.

libs/core/src/lib/pagination/pagination.component.ts
platon-rov: Hi there, we had a discussion on that, then I've tried to investigate how it can be done in a different (without `$localize`) way and I find your approach the most suitable here.

But we're not going to force customers to use the Angular i18n package, so this string should be getting from the @Input property and described in the docs so the customers will decide which localization package they want to use. 

cc @droshev @mikerodonnell89

droshev: Yes, so we are fine using Live Announcer. 
We think we should not be opinionated with i18n. Instead, we can offer inputs with default values. We want to avoided hardcoded texts. @kavya-b can you introduce the right inputs for the texts?

libs/core/src/lib/pagination/pagination.component.ts
```diff
@@ -261,6 +262,9 @@ export class PaginationComponent implements OnChanges, OnInit, OnDestroy {
         }
         this._refreshPages();
 
+        const currentPageMessage = $localize`:@@corePaginationCurrentPage:Current page`;
```
mikerodonnell89, line 265: Please allow for the developer to provide the message as an input

libs/core/src/lib/pagination/pagination.component.ts
```diff
@@ -1,27 +1,27 @@
...
-    TemplateRef, OnDestroy, ChangeDetectorRef
+    TemplateRef,
+    ViewEncapsulation
 } from '@angular/core';
-import {
-    ENTER,
-    SPACE
-} from '@angular/cdk/keycodes';
-import { coerceNumberProperty, coerceArray } from '@angular/cdk/coercion';
+import '@angular/localize/init';
```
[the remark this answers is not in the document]
kavya-b: > But it seems we should be safe to import localize function directly by
import { $localize } from '@angular/localize/src/localize'; 

Unfortunately, with this, the `ng extract-i18n` command fails to detect the module:
Screenshot 2021-08-30 at 9 09 24 AM

I also looked around a bit, and there is [not really any support ](https://github.com/angular/angular-cli/issues/17140#issuecomment-726080457)for adding `$localize` to libraries currently. There are some [hacks/workarounds](https://github.com/angular/angular/issues/29536#issuecomment-775162817) but I'm not sure if we should do that now or wait for angular to come up with a cleaner solution and we can just try to use hidden `<span>`s in template for the time being.

Also, Pete from Angular team [here](https://github.com/angular/angular/issues/36757#issuecomment-660519427) had suggested to just use the import in the lib .ts file for it to be detected, like how I had done. Also, the linked issue has added an `under consideration` label now so maybe there might be some work on this front soon.

cc: @droshev @manjunathanagaraj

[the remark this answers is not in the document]
dimamarksman: @kavya-b we asked Pete (petebacondarwin, angular team developer) and he confirmed that adding `import '@angular/localize/init'` into angular lib should be safe enough.
So let's keep it

dimamarksman (approved):

libs/core/src/lib/pagination/pagination.component.ts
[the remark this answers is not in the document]
kavya-b: oh, that's awesome! Thanks for confirming, Dima!

libs/core/src/lib/pagination/pagination.component.ts
kavya-b: @mikerodonnell89 @droshev resolving this conversation as the changes have been made in accordance to the comments received.

puru-hk:
All issues are fixed

puru-hk (approved):

Issue #3494, opened 2020-09-30
Accessibility:Panel: Name of the button “Panel Title” is not read when accessed in backward navigation using Shift+Tab key.


puru-hk:
Description: Name of the button “Panel Title” is not read when accessed in backward navigation using Shift+Tab key.

Expected: Name of the UI should be read in backward navigation also. 

Screenshot:

![image](https://user-images.githubusercontent.com/32538291/94650979-4d3d6d80-0315-11eb-9497-687b15d593e1.png)

PR #6259, opened 2021-08-06
fix(core,platform): fix a11y for Panel component

### LINKED ISSUE #2944: fix: (core) Add title to core Panel toggle button

enhancement

Taking into account that panel toggle button doesn't have text inside it (only icons) it would be nice to have `title` on this toggle button with the same value as `expandAriaLabel`.
In this case when the button is hovered it will show a title pointing an user what click action will lead to.

### LINKED ISSUE #6152: Accessibility Testing_Panel _Platform
Accessibility: Panel-Platform Component (There are couple of accessibility issues found while testing, so i have attached error document for you reference.) 
Screenshot:
[Panel.docx](https://github.com/SAP/fundamental-ngx/files/6907064/Panel.docx)

### LINKED ISSUE #6176: Accessibility Testing - Panel - Core
Accessibility: Panel -Core Component (There are couple of accessibility issues found while testing, so i have attached error document for you reference.)
Screenshot:
[Panel_Core_Ariba Controls.docx](https://github.com/SAP/fundamental-ngx/files/6917790/Panel_Core_Ariba.Controls.docx)

Panel - Core

1.1	Message Details

Raise Ticket	

Category	
Preprocessed By	
Description	

1.2 Application & Test Environment Details

Application Name:	ARIBA Controls- Fundamental Library for Angular
URL:	https://fundamental-ngx.netlify.app/#/core/panel

Technology:	Angular 11
Web browser:	Google Chrome Version 92.0.4515.107 (Official Build) (64-bit)
Operating System:	Microsoft Windows 10 x64
Screen Reader:	JAWS 2021.2105.53 ILM

 

Issue 1

Priority:	High
Checkpoint:	ACC-270 , ACC-264

(270.5)(264.1)

•	Search ‘Panel’ from the left panel
Observed behavior:
Case 1:
The textual header ‘Panel Header’ is included in navigation order.
Case 2:
On accessing the textual header 'Panel header' , an additional information 'contains text' is announced.
Speech output:
Panel Header
Contains text

Expected Behavior:    
The textual header ‘Panel Header’ should not be included in navigation order as it’s a non-interactive element and only relevant and meaningful information should be announced.
NOTE: the issue persists in all panels  
Screenshot:  
image

Issue 2

Priority:	High
Checkpoint:	ACC-270

(270.5)

•	Search ‘Panel’ from the left panel

Observed Behavior:
The textual content information ‘Lorem ipsum’ within the panel header is also included in the navigation order

Expected Behavior:
The textual contents should not be included in navigation order as it’s a non-interactive element.
NOTE: the issue persists in all panels  
Screenshot:
image

 

Issue 3

Priority:	Medium
Checkpoint:	ACC-264

(264.1)

•	Search ‘Panel’ from the left panel

Observed Behavior:

On accessing the expand button with the expandable panel, the associated header ‘Panel Header’ is not announced. 

Expected Behavior:
On accessing the expand button with the expandable panel, the associated header ‘Panel Header’ should be announced. 
NOTE: the issue persists in all panels  
Screenshot:

 
image

kavya-b:
Closes #6176 
Closes #6152 
Closes #2944 

**Core:**
_Issue 1, Issue 2:_ removed tabindex from title and content elements, and also the invisible focus style
_Issue 3:_  associating title and button with `expandAriaLabelledBy`
**Platform:**
_Issue 1:_  fdp-button was not propagating the aria-controls, aria-expanded and aria-pressed states to the underlying native button, so added these attributes as properties to platform button component. Also, associated id correctly to platform feed input content for aria-controls to work properly. 
edit:  there is a problem with JAWS in reading the expanded/collapsed state when `aria-pressed` is used. Have elaborated [here](https://github.com/SAP/fundamental-ngx/pull/6259#issuecomment-895014100).
_Issue 2:_ removed tabindex in core fixes this

additionally, also added title attribute to expand/collapse button and some formatting changes.

Documentation checklist:
- [n/a] update `README.md`
- [x] Documentation Examples


kavya-b:
@puru-hk @Sanjay-1312 please note that in core and platform Panel examples with the `Toggle Expand` button, JAWS does not read the `expanded` or `collapsed` state. It only announces `pressed` or `not pressed` states. This is a limitation from JAWS and same can be reproduced in other websites as well. For example, in http://www.mit.edu/~rjc/aria/expandingSections.html this website, if I add `aria-expanded` through `Inspect Element` and have JAWS read the 'toggle button' button at the end of the page, JAWS does not read out `expanded` or `collapsed`. Here is the speech history:
image

Therefore the issue 1 of Platform Panel(#6152) cannot be fixed due to this known issue. However, the toggle expand button shows the pressed state correctly now.

Betrozov (approved):

libs/core/src/lib/panel/panel-content/panel-content.directive.ts
```diff
@@ -53,7 +53,4 @@ export class PanelContentDirective {
     @HostBinding('attr.id')
     id: string = 'fd-panel-content-' + panelContentUniqueId++;
 
-    /** Tabindex of the content section. */
-    @HostBinding('attr.tabindex')
-    tabindex = 0;
```
InnaAtanasova, line 58: Why did you remove this? The idea of having it the first place was that these elements (section or header) can get some interaction and then they will need the focus.

kavya-b: @InnaAtanasova 
I understand but this was not the right thing to do, I think. 
- non-interactive elements should not get focus [or have tabindex](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex#accessibility_concerns): the panel content and the panel title are clearly visible and do not need to be read out by the screen reader by explicitly focusing on it or by adding tabindex. For JAWS, a11y team tests using something called virtual PC cursor mode in JAWS specifically for reading out static content like this.
- Adding tabindex with an invisible focus line also has the side effect of adding these non-interactive elements to the a11y/navigation tree (which is what the first two issues in #6176 talk about). Even [now](https://fundamental-ngx.netlify.app/#/core/panel), after tabbing to the expand/collapse button, it takes two tabs to reach `Show Code`; this shouldn't happen and can confuse the user.
- the [issue](https://github.com/SAP/fundamental-ngx/issues/3494) for which this change was added(pr https://github.com/SAP/fundamental-ngx/pull/4555) talks about title not being read when focus was on the expand/collapse arrow. We could have fixed this by using aria-labelledby instead of adding tabindex.

kavya-b:
@InnaAtanasova could you please review this; I have responded to your comment [here](https://github.com/SAP/fundamental-ngx/pull/6259#discussion_r685397728)

libs/core/src/lib/panel/panel-content/panel-content.directive.ts
InnaAtanasova: @kavya-b thanks for explaining it. I was just confused why we are now removing smth that was added to fix an a11y issue :)

InnaAtanasova (approved):

puru-hk:
All issues are fixed.

puru-hk (approved):

PR #6311, opened 2021-08-12
fix(platform): trigger detection when field state changes

### LINKED ISSUE #6275: Form generator - error message not shown if use with Observable-based values
bug
Error message must be shown 
latest
1. Go to https://deploy-preview-6199--fundamental-ngx.netlify.app/#/platform/form-generator
2. Scroll down to `Form Generator usage with Observable-based values` example
3. Enter not valid text in the `Your name` field
Error message not shown
An error message is displayed after defocusing and then focusing again
n/a
n/a

N1XUS:
Closes #6275

Added detection when field status changes.

Documentation checklist:
- n/a update `README.md`
- n/a [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)
- n/a Documentation Examples
- n/a Stackblitz works for all examples

### DIFF
--- a/libs/platform/src/lib/components/form/base.input.ts
+++ b/libs/platform/src/lib/components/form/base.input.ts
@@ -209,6 +209,15 @@ export abstract class BaseInput extends BaseComponent
 
     /** @hidden */
     ngAfterViewInit(): void {
+
+        if (this.ngControl) {
+            this._subscriptions.add(
+                this.ngControl.statusChanges.subscribe(() => {
+                    this.markForCheck();
+                })
+            );
+        }
+
         const labelAndHelpId = `fdp-form-label-${this.id} fdp-form-hint-${this.id}`;
         // if not specified, associate label and inline help ids with the input,
         // else add these ids to the specified ones


olshansky (approved):

libs/platform/src/lib/components/form/base.input.ts
```diff
@@ -209,6 +209,15 @@ export abstract class BaseInput extends BaseComponent
 
     /** @hidden */
     ngAfterViewInit(): void {
+
+        if (this.ngControl) {
```
fkolar, line 213: Why do we need this here? Is this because we are using our control outside of FormField? The idea was always to use FormField that controls everything, this way there will be too much activity going on. 

If yes, then we should turn this on only and if input is used as standalone.  Or what are the usecaes for this ?

droshev (approved):

libs/platform/src/lib/components/form/base.input.ts
```diff
@@ -209,6 +209,15 @@ export abstract class BaseInput extends BaseComponent
...
+            this._subscriptions.add(
+                this.ngControl.statusChanges.subscribe(() => {
+                    this.markForCheck();
```
fkolar, line 216: I think we should make it conditional. Otherwise will be firing like crazy.

fkolar:
@droshev You should not merge this yet. As there are open question. When you approved this have you considered why we even use `this.ngControl.statusChanges` for?

libs/platform/src/lib/components/form/base.input.ts
N1XUS: It fires only when the previous validation status does not equal the current, so there shouldn't be the case when it's triggered on every user keydown.

PR #6322, opened 2021-08-12
fix(core): fix top round corners for combobox and multi input

### LINKED ISSUE #5985: Combobox defect hunting issues
Top 2 corners have a border radius when they should not (applies to multi-input as well):
Screen Shot 2021-07-14 at 2 30 38 PM

Clicking the combobox search field 'x' to clear the form erroneously opens the popover

Console error thrown and the user's first input key did not register when using custom filter or display object property

The cursor is focused back to the combobox after the user attempted to leave the input

aberikashvili:
Closes #5985

- [N/A] tests for the changes that have been done

Documentation checklist:
- [N/A] update `README.md`
- [N/A] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)
- [N/A] Documentation Examples


droshev:
@aberikashvili do you think the stakeholders can understand what the fix is from this title: `Combobox defect hunting issues`

aberikashvili:
@droshev edited

libs/core/src/lib/popover/popover-body/popover-body.component.scss
```diff
@@ -128,6 +128,11 @@ $fd-popover-background-color: #fff;
 			bottom: $fd-popover-arrow-offset;
 		}
 	}
+
+    .fd-popover__popper--no-border-radius & {
+        border-top-left-radius: 0;
+        border-top-right-radius: 0;
+    }
```
InnaAtanasova, line 135: Didn't we get rid of popper.js?

libs/core/src/lib/combobox/combobox.component.html
```diff
@@ -15,7 +15,9 @@
         <fd-popover-control>
             <ng-container *ngTemplateOutlet="controlTemplate"></ng-container>
         </fd-popover-control>
-        <fd-popover-body [hidden]="!displayedValues || !displayedValues.length">
+        <fd-popover-body
+            [hidden]="!displayedValues || !displayedValues.length"
+            [class.fd-popover__popper--no-border-radius]="true">
```
Betrozov, line 20: Could we use it like that?
```suggestion
            class="fd-popover__popper--no-border-radius">
```

libs/core/src/lib/multi-input/multi-input.component.html
```diff
@@ -20,7 +20,10 @@
                     </ng-container>
                 </form>
             </fd-popover-control>
-            <fd-popover-body [attr.aria-hidden]="!open" [class.fd-popover__body--hidden]="!displayedValues || !displayedValues.length">
+            <fd-popover-body
+                [attr.aria-hidden]="!open"
+                [class.fd-popover__body--hidden]="!displayedValues || !displayedValues.length"
+                [class.fd-popover__popper--no-border-radius]="true">
```
Betrozov, line 26: same

Betrozov (changes requested):

mikerodonnell89 (changes requested):
If the combobox is near the bottom of the page when the user opens it, the popover will be placed above the input, in which case this fix doesn't work well

Screen Shot 2021-08-17 at 1 40 01 PM

libs/core/src/lib/popover/popover-body/popover-body.component.scss
aberikashvili: I don't know

libs/core/src/lib/combobox/combobox.component.html
aberikashvili: Done

libs/core/src/lib/multi-input/multi-input.component.html
aberikashvili: Done

Betrozov (approved):

aberikashvili:
> If the combobox is near the bottom of the page when the user opens it, the popover will be placed above the input, in which case this fix doesn't work well
> 
> Screen Shot 2021-08-17 at 1 40 01 PM

fixed:
image
image

mikerodonnell89 (commented):
The popover body seems to stick out a bit from the edges of the input - this was happening previously as well but was less obvious due to the corners having the border radius.  Can this be addressed in this PR?  I've circled here in red.
Screen Shot 2021-08-24 at 9 53 51 AM


And here is the combobox in fundamental-styles, notice we aren't having that issue:
Screen Shot 2021-08-24 at 9 50 22 AM

libs/core/src/lib/popover/popover-service/popover.service.ts
```diff
@@ -370,7 +378,7 @@ export class PopoverService extends BasePopoverClass {
     /** @hidden */
     private _applyWidthOverlay(): void {
         const maxWidthLimit = this.maxWidth ? this.maxWidth : MAX_BODY_SIZE;
-        const width = Math.min(this._getTriggerWidth(), maxWidthLimit);
+        const width = Math.min(this._getTriggerWidth(), maxWidthLimit) - WIDTH_OVERLAY_OFFSET;
```
dimamarksman, line 381: what is WIDTH_OVERLAY_OFFSET?

dimamarksman: I think we shouldn't use this hardcoded offset, instead we should rely on styles.
As far as I can see the original problem comes from the usage of `shadow` property and width on the same element.
We set min-width as popover-control width and set shadow style (mimic border) on the same node, which leads to such an effect.

In fundamnetal-styles we don't have this problem because examples there rely on `fd-popover_body` class and this class has a [modifier to align this thing](https://github.com/SAP/fundamental-styles/blob/9fed66352a4034aed39466714eb98af5c253c4da/src/popover.scss#L206-L210).
We need the same for `fd-popover__popper` but additionally we need to set `min-width/max-width` to the `fd-popover__popper` parent element instead of `fd-popover__popper` itself

libs/core/src/lib/popover/popover-service/popover.service.ts
```diff
@@ -301,16 +301,24 @@ export class PopoverService extends BasePopoverClass {
...
+            filter(() => !this.noArrow && !!this._getPopoverBody()),
+            distinctUntilChanged(
+                (previous, current) =>
+                    previous.connectionPair === current.connectionPair
+            )
+        ).subscribe(event => this._getPopoverBody()._setArrowStyles(event.connectionPair, this._getDirection()));
+
+        // non popover cases, like: combobox, multi-input etc.
+        observer.pipe(
+            filter(() => this.noArrow && !!this._getPopoverBody()),
```
dimamarksman, line 319: not always `noArrow` means combobox or multi-input.
We might have just dropdown as it's shown here:
image


But with this change we have this:
image

dimamarksman: We probably need some dedicated input in popover itself to make it clear.
@mikerodonnell89 @platon-rov what do you think?

libs/core/src/lib/popover/popover-body/popover-body.component.html
```diff
@@ -1,6 +1,7 @@
 <div class="fd-popover__popper fd-popover__popper--cdk-custom"
      [class.fd-popover__popper--no-arrow]="_noArrow"
      [ngClass]="_additionalBodyClass"
+     [ngClass]="_angleClass"
```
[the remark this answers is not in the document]
aberikashvili: @dimamarksman added method which will return `_additionalBodyClass` and `_angleClass` as one string

libs/core/src/lib/popover/popover-service/popover.service.ts
aberikashvili: @dimamarksman you mean it's not a bug? or how we should deal with it?

platon-rov:
Should be blocked until https://github.com/SAP/fundamental-ngx/issues/5985#issuecomment-908277191 discussed.

platon-rov:
@aberikashvili okay, according to the Fiori 3 Visual Specs we should have round borders for all corners. But the ticket your PR closes also contains some other items. Feel free to remove the round corners fix and left only fixes for the other items.

InnaAtanasova:
Please reopen the PR when it's ready to be reviewed. The CLA needs to be signed.

PR #6357, opened 2021-08-17
refactoring(core,platform): table navigatable rows

platon-rov:

Previously implemented navigatable table rows feature needs to be adjusted a bit because of:

* In the core component, there was no ability to make row activable, hoverable if the row is not navigatable.
* In the platform component, there was no ability to make row navigatable without TS code, just from the template.
* [Requested] Ability to trigger custom function on row navigation is needed, not only go by the provided route.

Documentation checklist:
- [X] update `README.md`
- [X] Documentation Examples


apps/docs/src/app/platform/component-docs/platform-table/platform-table-examples/platform-table-navigatable-row-indicator-example.component.ts
```diff
@@ -1,20 +1,17 @@
...
 
-import { FdDate } from '@fundamental-ngx/core';
-import {
-    TableDataSource,
-    TableDataProvider,
-    TableState,
-    TableComponent
-} from '@fundamental-ngx/platform';
+import { Observable, of } from 'rxjs';
+import { TableRowActivateEvent } from '../../../../../../../../libs/platform/src/lib/components/table/models';
```
dimamarksman, line 7: please use absolute path

libs/platform/src/lib/components/table/table.component.html
```diff
@@ -423,6 +425,17 @@
...
                         ></td>
+
+                        <td
+                            *ngIf="_isShownNavigationColumn"
+                            fd-table-cell
+                        >
+                            <i
+                                fd-table-icon
+                                [navigation]="true"
+                                class="fdp-table__navigation-indicator sap-icon--slim-arrow-left"
```
dimamarksman, line 436: should we handle rtl?

libs/platform/src/lib/components/table/table.component.ts
```diff
@@ -426,6 +428,11 @@ export class TableComponent<T = any> extends Table implements AfterViewInit, OnD
     /** @hidden */
     _scrollBarWidth = 0;
 
+    /** @hidden */
+    get _isShownNavigationColumn(): boolean {
+        return this._tableRows.some(row => row.navigatable);
```
dimamarksman, line 433: Iteration in the getter potentially might be a problem.
Let's keep it as standalone value (e.g. observable and async in the template)?

apps/docs/src/app/core/component-docs/table/examples/table-navigatable-row-example.component.html
```diff
@@ -36,68 +62,98 @@ <h4 class="table-example-popin-title">Row with navigation button</h4>
...
+                {{ row.date }}
+            </td>
+
             <td fd-table-cell [fitContent]="true">
-                <button aria-label="navigation" class="fd-button fd-button--transparent" [disabled]="!row.navigatable" [routerLink]="'/core/home'">
-                    <i [ngClass]="getClass()"></i>
+                <button
+                    class="fd-button fd-button--transparent"
+                    [disabled]="!row.navigatable"
+                    [routerLink]="'/core/home'"
```
Smiranin, line 88: extra bindings. Just without brackets

libs/platform/src/lib/components/table/table.component.ts
```diff
@@ -1771,4 +1744,17 @@ export class TableComponent<T = any> extends Table implements AfterViewInit, OnD
...
+    private _isRowNavigatable(row: T, rowNavigationDisabled: string | boolean): boolean {
+        if (!row) {
+            return false;
+        }
+
+        if (isString(rowNavigationDisabled)) {
+            return !!row[rowNavigationDisabled];
+        }
+
+        return !!rowNavigationDisabled;
```
dimamarksman, line 1758: From this code if `rowNavigationDisabled` means `isRowNavigatable`.
Looks illogical to me.

libs/platform/src/lib/components/table/table.component.ts
```diff
@@ -244,9 +244,9 @@ export class TableComponent<T = any> extends Table implements AfterViewInit, OnD
     @Input()
     relationKey: string;
 
-    /** Whether table row has navigation button. */
+    /** Whether row is navigatable. */
     @Input()
-    navigationButton = false;
+    rowNavigatable: string | boolean;
```
dimamarksman, line 249: what does `string` mean here?
could you please add description for it

libs/platform/src/lib/components/table/table.component.ts
```diff
@@ -501,6 +507,10 @@ export class TableComponent<T = any> extends Table implements AfterViewInit, OnD
         if ('selectionMode' in changes || 'freezeColumnsTo' in changes) {
             this.recalculateTableColumnWidth();
         }
+
+        if ('navigation' in changes) {
```
dimamarksman, line 511: 'navigation'? I don't see such @input

libs/platform/src/lib/components/table/table.component.ts
Smiranin, line 249: It's not entirely clear what can come as a string? Which options of the string?

platon-rov: Updated typedoc comment:

```
/** Whether row is navigatable. 
      * Pass boolean value to set state for the all rows.
      * Pass string value with the key of the row item's field to compute state for every single row.
      */
      ```

platon-rov: https://github.com/SAP/fundamental-ngx/pull/6357#discussion_r692745286

libs/platform/src/lib/components/table/table.component.ts
platon-rov: Great catch, agree!

libs/platform/src/lib/components/table/table.component.html
platon-rov: It's inside the scroller container so we don't really care about what's inside, it's just to calculate the container width and the scroller width properly.

libs/platform/src/lib/components/table/table.component.ts
platon-rov: Now if found that this is different things.

We may have not set "rowNavigatable" and then call "setRowNavigation(rowIndex, rowNavigationState)" with boolean value - in that case `return !!this.rowNavigatable` won't calculate right value for the `row.navigatable`.

platon-rov: Naming, improved, thank you!

dimamarksman: let's change rowNavigationDisabled to rowNavigatable

dimamarksman (approved):

Smiranin (approved):

Issue #6206, opened 2021-08-03
Accessibility_Testing_Button_Core


Wemitha1234:
Accessibility: Button-Core Component (There are couple of accessibility issues found while testing, so i have attached error document for you reference.) Screenshot:
[Button_Core_ARIBA Controls.docx](https://github.com/SAP/fundamental-ngx/files/6920577/Button_Core_ARIBA.Controls.docx)

Issue 1

Priority:	Medium
Checkpoint:	ACC-271

Observed Behavior:
No focus indication provided when Disabled State Button is accessed. 

The speech output is:
Disabled State Button  Unavailable  
To activate press spacebar.

Expected Behavior:
Ideally, clear dotted focus indication should be provided for each & every UI when accessed.  

Screenshot:
![image](https://user-images.githubusercontent.com/88035260/127947100-7b22e118-d378-4835-8c55-8a174fc1b425.png)

Issue 2

Priority:	Medium
Checkpoint:	ACC-264

Observed Behavior:
Color information about buttons is not read out when buttons are accessed. 

The speech output is:
Filter Button  
To activate press spacebar.
Filter Button  
To activate press spacebar.
Filter Button  
To activate press spacebar.
Filter Button  
To activate press spacebar.

Expected Behavior:
Ideally, color information such as success, warning, error, etc. should be conveyed when buttons are accessed. 

Screenshot:
![image](https://user-images.githubusercontent.com/88035260/127947129-1e83b40f-0f74-42ca-abe5-3a083a07170c.png)

Issue 3

Priority:	High
Checkpoint:	ACC-264

Observed Behavior:
Button looks emphasized. This ‘emphasized’ information about button is not read out when accessed. 

The speech output is:
Add to Cart Button  
To activate press spacebar.

Expected Behavior:
Ideally, the emphasized information about the UI should be conveyed.
 
Screenshot:
![image](https://user-images.githubusercontent.com/88035260/127947175-87b530cb-3272-43bc-ac10-0ef9fa71324f.png)

Issue 4

Priority:	High
Checkpoint:	ACC-264

Observed Behavior:

UI type menu button is not read out when menu buttons are accessed. 

The speech output is:
Add to Cart Button  
To activate press spacebar.
Add to Cart Button  
To activate press spacebar.
Add to Cart Button  
To activate press spacebar.
Filter Button  
To activate press spacebar.
Filter Button  
To activate press spacebar.
Filter Button  
To activate press spacebar.
Filter Button  
[...]

PR #6385, opened 2021-08-25
fix(core): adding accessability label and reading button type

Lokanathan-k:
Closes: https://github.com/SAP/fundamental-ngx/issues/6206

adding accessability label and reading button type

Documentation checklist:
- [NA] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/apps/docs/src/app/core/component-docs/button/examples/button-icons-example.component.html
+++ b/apps/docs/src/app/core/component-docs/button/examples/button-icons-example.component.html
@@ -1,15 +1,15 @@
-<button fd-button label="Add to Cart" [glyph]="'cart'"></button>
-<button fd-button label="Add to Cart" [fdType]="'emphasized'" [glyph]="'cart'"></button>
-<button fd-button label="Add to Cart" [fdType]="'ghost'" [glyph]="'cart'"></button>
-<button fd-button label="Filter" [fdType]="'positive'" [glyph]="'filter'"></button>
-<button fd-button label="Filter" [fdType]="'negative'" [glyph]="'filter'"></button>
-<button fd-button label="Filter" [fdType]="'attention'" [glyph]="'filter'"></button>
-<button fd-button label="Filter" [fdType]="'transparent'" [glyph]="'filter'"></button>
+<button fd-button label="Add to Cart"  glyph="cart" ariaLabel="Add to Cart"></button>
+<button fd-button label="Add to Cart" fdType="emphasized" glyph="cart"></button>
+<button fd-button label="Add to Cart" fdType="ghost"  glyph="cart" ariaLabel="Add to Cart Ghost"></button>
+<button fd-button label="Filter" fdType="positive" glyph="filter" ariaLabel="Positive Filter"></button>
+<button fd-button label="Filter" fdType="negative" glyph="filter" ariaLabel="Negative Filter"></button>
+<button fd-button label="Filter" fdType="attention" glyph="filter" ariaLabel="Attention Filter"></button>
+<button fd-button label="Filter" fdType="transparent" glyph="filter" ariaLabel="Transparent Filter"></button>
 
-<button fd-button [glyph]="'cart'" aria-label="Add to Cart" title="Add to Cart"></button>
-<button fd-button [fdType]="'emphasized'" [glyph]="'cart'" aria-label="Add to Cart" title="Add to Cart"></button>
-<button fd-button [fdType]="'ghost'" [glyph]="'cart'" aria-label="Add to Cart" title="Add to Cart"></button>
-<button fd-button [fdType]="'positive'" [glyph]="'accept'" aria-label="accept" title="accept"></button>
-<button fd-button [fdType]="'negative'" [glyph]="'filter'" aria-label="filter" title="filter"></button>
-<button fd-button [fdType]="'attention'" [glyph]="'filter'" aria-label="filter" title="filter"></button>
-<button fd-button [fdType]="'transparent'" [glyph]="'menu2'" aria-label="menu2" title="menu2"></button>
+<button fd-button glyph="cart" ariaLabel="Add to Cart" ariaLabel="Add to Cart" title="Add to Cart"></button>
+<button fd-button fdType="emphasized" glyph="cart" title="Cart Emphasized"></button>
+<button fd-button fdType="ghost" glyph="cart" ariaLabel="Cart Ghost" title="Cart Ghost"></button>
+<button fd-button fdType="positive" glyph="accept" ariaLabel="Accept Positive" title="Accept Positive"></button>
+<button fd-button fdType="negative" glyph="filter" ariaLabel="Filter Negative" title="Filter Negative"></button>
+<button fd-button fdType="attention" glyph="filter" ariaLabel="Filter Attention" title="Filter Attention"></button>
+<button fd-button fdType="transparent" glyph="menu2" ariaLabel="Menu Transparent" title="Menu Transparent"></button>
--- a/apps/docs/src/app/core/component-docs/button/examples/button-menu-example.component.html
+++ b/apps/docs/src/app/core/component-docs/button/examples/button-menu-example.component.html
@@ -1,14 +1,14 @@
-<button fd-button [fdMenu]="true" [glyph]="'cart'" label="Add to Cart"></button>
-<button fd-button [fdMenu]="true" [fdType]="'emphasized'" [glyph]="'cart'" label="Add to Cart"></button>
-<button fd-button [fdMenu]="true" [fdType]="'ghost'" [glyph]="'cart'" label="Add to Cart"></button>
-<button fd-button [fdMenu]="true" [fdType]="'positive'" [glyph]="'filter'" label="Filter"></button>
-<button fd-button [fdMenu]="true" [fdType]="'negative'" [glyph]="'filter'" label="Filter"></button>
-<button fd-button [fdMenu]="true" [fdType]="'attention'" [glyph]="'filter'" label="Filter"></button>
-<button fd-button [fdMenu]="true" [fdType]="'transparent'" [glyph]="'filter'" label="Filter"></button>
+<button fd-button [fdMenu]="true" glyph="cart" label="Add to Cart" ariaLabel="Add to cart"></button>
+<button fd-button [fdMenu]="true" fdType="emphasized" glyph="cart" label="Add to Cart" ariaLabel="Add to cart emphasized Menu"></button>
+<button fd-button [fdMenu]="true" fdType="ghost" glyph="cart" label="Add to Cart" ariaLabel="Add to cart Ghost Menu"></button>
+<button fd-button [fdMenu]="true" fdType="positive" glyph="filter" label="Filter" ariaLabel="Success Filter Menu"></button>
+<button fd-button [fdMenu]="true" fdType="negative" glyph="filter" label="Filter" ariaLabel="Negative Filter Menu"></button>
+<button fd-button [fdMenu]="true" fdType="attention" glyph="filter" label="Filter" ariaLabel="Attention Filter Menu"></button>
+<button fd-button [fdMenu]="true" fdType="transparent" glyph="filter" label="Filter" ariaLabel="Transparent Filter Menu"></button>
 
-<button fd-button [fdMenu]="true" [fdType]="'emphasized'" [glyph]="'cart'" aria-label="cart" title="cart"></button>
-<button fd-button [fdMenu]="true" [fdType]="'ghost'" [glyph]="'cart'" aria-label="cart" title="cart"></button>
-<button fd-button [fdMenu]="true" [fdType]="'positive'" [glyph]="'accept'" aria-label="accept" title="accept"></button>
-<button fd-button [fdMenu]="true" [fdType]="'negative'" [glyph]="'filter'" aria-label="filter" title="filter"></button>
-<button fd-button [fdMenu]="true" [fdType]="'attention'" [glyph]="'filter'" aria-label="filter" title="filter"></button>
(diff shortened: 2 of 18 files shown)


Betrozov (approved):

mikerodonnell89 (changes requested):
In my opinion the aria-label should be an input on the button (`@Input() ariaLabel: string = null`).  And if the developer has not provided a value for this input, then the button could automatically set the `aria-label` using the value given by the developer for `label` (and the `fdType`, if provided. They should be comma-separated).  If the dev has provided the input `ariaLabel` then use that

Lokanathan-k:
> provided a value for this input, th

@InnaAtanasova @mikerodonnell89 
There are few button with no label meaning the button with only icon in that case what should be done? if that is not handled aria label is something like this aria-label="undefined, emphasised"

mikerodonnell89:
> There are few button with no label meaning the button with only icon in that case what should be done? if that is not handled aria label is something like this aria-label="undefined, emphasised"

Good question, I didn't think of this, maybe use the property passed in to the glyph input, instead of the label, thoughts @InnaAtanasova ? This should simply be the default behavior and the docs should make it clear to the developer that providing the appropriate label is their responsibility

InnaAtanasova:
> > There are few button with no label meaning the button with only icon in that case what should be done? if that is not handled aria label is something like this aria-label="undefined, emphasised"
> 
> Good question, I didn't think of this, maybe use the property passed in to the glyph input, instead of the label, thoughts @InnaAtanasova ? This should simply be the default behavior and the docs should make it clear to the developer that providing the appropriate label is their responsibility

It was always the case. This is the current state of the buttons:
Screen Shot 2021-09-01 at 12 10 21 PM

The ones with no label (text) have `aria-label`.

InnaAtanasova:
For this case, where we have label (text):
```
<button fd-button [fdMenu]="true" [fdType]="'positive'" [glyph]="'filter'" label="Filter"></button>
```

We can check if there's aria-label provided by the user and if not set it to the label text behind the scene.
If there's aria-label, do nothing.

In this case, where you have only icon, the user must provide aria-label, just like they did before:
```
<button fd-button [fdMenu]="true" [fdType]="'positive'" [glyph]="'accept'" aria-label="accept" title="accept"></button>
```

mikerodonnell89 (approved):

Seamoo13:
pushed a fix for 1 test, the other failing test is due to an issue in your PR

1 issue causing core flexible column layout test to fail -  aria-label for button in dynamic example not present until after clicking the button

**STR:**

1. go to https://deploy-preview-6385--fundamental-ngx.netlify.app/#/core/flexible-column-layout
2. go to the 2nd example (Flexible Column Layout With Dynamic Page)
3. click to open the example
4. click 'open column 2' button
5. inspect the button with the icon of a pin
6. notice that there is no aria-label (on main branch there is)
7. click the pin button
8. notice that the aria-label has appeared
9. click button again
10. aria-label present

**Actual results:** pin button missing aria-label until after being clicked

**Expected results:** button should have aria-label on load

https://user-images.githubusercontent.com/47522152/132020633-d2374dfc-0fc5-4909-9130-c44e6375a6c2.mp4

libs/core/src/lib/bar/button-bar/button-bar.component.ts
```diff
@@ -23,8 +23,8 @@ import { ContentDensityService } from '@fundamental-ngx/core/utils';
               [compact]="compact"
               [fdType]="fdType"
               [label]="label"
-              [attr.title]="title"
-              [attr.aria-label]="ariaLabel"
+              [attr.title]="ariaLabel"
```
InnaAtanasova, line 26: Shouldn't this first check if the title prop is set and then assign it to ariaLabel? 
We should not take away the option for the user to set the value of title to what they want

libs/core/src/lib/bar/button-bar/button-bar.component.ts
```diff
@@ -23,8 +23,8 @@ import { ContentDensityService } from '@fundamental-ngx/core/utils';
...
+              [ariaLabel]="ariaLabel"
```
InnaAtanasova, line 27: I'm confused. Why did you change [attr.aria-label] to [ariaLabel]?

InnaAtanasova (dismissed):

libs/core/src/lib/button/button.component.ts
```diff
@@ -37,7 +37,8 @@ import { applyCssClass } from '@fundamental-ngx/core/utils';
     changeDetection: ChangeDetectionStrategy.OnPush,
     host: {
         '[attr.type]': 'type',
-        '[attr.disabled]': '_disabled || null'
+        '[attr.disabled]': '_disabled || null',
+        '[attr.aria-label]': 'ariaLabel ? ariaLabel : label !== undefined ? label +","+ fdType : glyph === undefined || glyph === null  ? title : glyph.split("-").join(" ") +","+ fdType '
```
droshev, line 41: why do we need this?

libs/core/src/lib/button/base-button.ts
```diff
@@ -61,6 +61,10 @@ export class BaseButton {
     @Input()
     fdMenu = false;
 
+    /** adding native aria-label to the component */
+    @Input()
+    ariaLabel: string = null;
```
droshev, line 66: can't we use the native attribute and avoid the creation iof an input

apps/docs/src/app/core/component-docs/button/examples/button-types-example.component.html
```diff
@@ -1,7 +1,7 @@
-<button fd-button label="Standard Button"></button>
-<button fd-button label="Emphasized Button" [fdType]="'emphasized'"></button>
-<button fd-button label="Ghost Button" [fdType]="'ghost'"></button>
-<button fd-button label="Positive Button" [fdType]="'positive'"></button>
-<button fd-button label="Negative Button" [fdType]="'negative'"></button>
-<button fd-button label="Attention Button" [fdType]="'attention'"></button>
-<button fd-button label="Transparent Button" [fdType]="'transparent'"></button>
+<button fd-button label="Standard Button" ariaLabel="Standard"></button>
+<button fd-button label="Emphasized Button" fdType="emphasized" ariaLabel="Emphasized"></button>
```
droshev, line 2: those updates of the doc where `[fdType]` is replaced with `fdType` should be done in separate Pr

apps/docs/src/app/core/component-docs/button/examples/button-menu-example.component.html
```diff
@@ -1,14 +1,14 @@
-<button fd-button [fdMenu]="true" [glyph]="'cart'" label="Add to Cart"></button>
-<button fd-button [fdMenu]="true" [fdType]="'emphasized'" [glyph]="'cart'" label="Add to Cart"></button>
-<button fd-button [fdMenu]="true" [fdType]="'ghost'" [glyph]="'cart'" label="Add to Cart"></button>
-<button fd-button [fdMenu]="true" [fdType]="'positive'" [glyph]="'filter'" label="Filter"></button>
-<button fd-button [fdMenu]="true" [fdType]="'negative'" [glyph]="'filter'" label="Filter"></button>
-<button fd-button [fdMenu]="true" [fdType]="'attention'" [glyph]="'filter'" label="Filter"></button>
-<button fd-button [fdMenu]="true" [fdType]="'transparent'" [glyph]="'filter'" label="Filter"></button>
+<button fd-button [fdMenu]="true" glyph="cart" label="Add to Cart" ariaLabel="Add to cart"></button>
```
droshev, line 1: those updates of the doc where `[glyph]` is replaced with `glyph` should be done in separate PR

apps/docs/src/app/core/component-docs/button/examples/button-icons-example.component.html
```diff
@@ -1,15 +1,15 @@
-<button fd-button label="Add to Cart" [glyph]="'cart'"></button>
-<button fd-button label="Add to Cart" [fdType]="'emphasized'" [glyph]="'cart'"></button>
-<button fd-button label="Add to Cart" [fdType]="'ghost'" [glyph]="'cart'"></button>
-<button fd-button label="Filter" [fdType]="'positive'" [glyph]="'filter'"></button>
-<button fd-button label="Filter" [fdType]="'negative'" [glyph]="'filter'"></button>
-<button fd-button label="Filter" [fdType]="'attention'" [glyph]="'filter'"></button>
-<button fd-button label="Filter" [fdType]="'transparent'" [glyph]="'filter'"></button>
+<button fd-button label="Add to Cart"  glyph="cart" ariaLabel="Add to Cart"></button>
```
droshev, line 1: those updates of the doc where `[glyph]` is replaced with `glyph` should be done in separate PR

droshev (dismissed):
i think the PR is mixing few things together

libs/core/src/lib/button/button.component.ts
Lokanathan-k: @mikerodonnell89 
https://github.com/SAP/fundamental-ngx/pull/6385#pullrequestreview-742950708
requested for this change.

apps/docs/src/app/core/component-docs/button/examples/button-types-example.component.html
Lokanathan-k: @droshev Previously a comment was shared saying if we are touching any file and that file has unwanted braces can be removed and this change was introduced as part of the previously shared comment.

puru-hk:
All issues are fixed.

libs/core/src/lib/bar/button-bar/button-bar.component.ts
Lokanathan-k: this change is overridden when ariaLabel is passed or label is passed as a parameter.

libs/core/src/lib/bar/button-bar/button-bar.component.ts
Lokanathan-k: fixed

apps/docs/src/app/core/component-docs/button/examples/button-types-example.component.html
droshev: I don't see why you would touch this file for the initial idea of your PR

Lokanathan-k:
waiting for the feedback from the Ngx core and accessibility team.

droshev:
@Lokanathan-k can we focus only on reading the button types(Emphasized and the semantic ones) the other types are just design types and they may confused the user (e.g. ghost button)?
So what we need is to have an internal logic to add the "Emphasized, warning, positive, and negative".

Lokanathan-k:
> @Lokanathan-k can we focus only on reading the button types(Emphasized and the semantic ones) the other types are just design types and they may confused the user (e.g. ghost button)?
> So what we need is to have an internal logic to add the "Emphasized, warning, positive, and negative".

@droshev done, pls have a look

libs/core/src/lib/button/button.component.ts
```diff
@@ -37,14 +37,21 @@ import { applyCssClass } from '@fundamental-ngx/core/utils';
...
+        '[attr.aria-label]': 'ariaLabel ? ariaLabel : specialButtonType.includes(fdType) ? label !==null ?  label +" ,"+ fdType : glyph !== undefined || glyph !== null ? fdType +","+ glyph.split("-").join(" ")  : title : title'
     }
 })
 export class ButtonComponent extends BaseButton implements OnChanges, CssClassBuilder, OnInit,  OnDestroy {
     /** The property allows user to pass additional css classes. */
     @Input()
     class = '';
 
+    /** @hidden */
+    specialButtonType: Array<string> = ['emphasized', 'positive', 'negative', 'attention'];
```
platon-rov, line 50: Could you please care about supporting i18n.

libs/core/src/lib/button/button.component.ts
```diff
@@ -37,14 +37,21 @@ import { applyCssClass } from '@fundamental-ngx/core/utils';
     changeDetection: ChangeDetectionStrategy.OnPush,
     host: {
         '[attr.type]': 'type',
-        '[attr.disabled]': '_disabled || null'
+        '[attr.disabled]': '_disabled || null',
+        '[attr.aria-label]': 'ariaLabel ? ariaLabel : specialButtonType.includes(fdType) ? label !==null ?  label +" ,"+ fdType : glyph !== undefined || glyph !== null ? fdType +","+ glyph.split("-").join(" ")  : title : title'
```
platon-rov, line 41: Looks to hard to read.

Please use `!= null` for checking for not null and not undefined.

Instead of `a ? a : b` you can use `a || b`.

Please rewrite this expression.

platon-rov (dismissed):

libs/core/src/lib/button/button.component.ts
Lokanathan-k: this array is used to handle only button types that has to be read out by default., other button types are not required since it creates unwanted confusion this array.this is used to handle internally.

libs/core/src/lib/button/button.component.ts
Lokanathan-k: @platon-rov this(a ? a : b you can use a || b) doesnt work as expected, when ariaLabel is passed the ariaLabel has to be retained. a || b doesn't do that.

libs/core/src/lib/button/button.component.ts
platon-rov: Got it, so maybe we need to rename the field to avoid such misunderstanding in feature. I guess we also has alll the button types as enum, do we? In that case we can use as items for that array enum values from there.

puru-hk:
Issue 1: Not fixed
Issue 2: Fixed
Issue 3: Fixed
Issue 4: Fixed
Issue 5: Fixed

puru-hk (approved):

Issue #6219, opened 2021-08-03
Accessibility_Testing_Inline Help_Core


Wemitha1234:
Accessibility: Inline Help-Core Component (There are couple of accessibility issues found while testing, so i have attached error document for you reference.) Screenshot:

[Inline Help_Core_ARIBA Controls.docx](https://github.com/SAP/fundamental-ngx/files/6922327/Inline.Help_Core_ARIBA.Controls.docx)

Issue 1

Priority:	Medium
Checkpoint:	https://wiki.wdf.sap.corp/wiki/display/Acc/ACC-284+Details

Observed Behavior:
Inline help highlighted in red appears only when mouse hovered on it. 

Expected Behavior:
Ideally, each & every UI should be accessible via keyboard. 

Screenshot:
![image](https://user-images.githubusercontent.com/88035260/127980173-c225738f-7367-43cc-904d-ff79aca92ff6.png)

Issue 2

Priority:	Medium
Checkpoint:	ACC-264

Observed Behavior:
The red color information (for example: error) about inline help is not conveyed to the user.  

Speech output is:
Inline Help Tooltip

Expected Behavior:
Ideally, the color information should be conveyed to the user.  

Screenshot:

![image](https://user-images.githubusercontent.com/88035260/127980230-3683e107-75af-4f06-b32b-492fcaf0792b.png)

Issue 3

Priority:	Medium
Checkpoint:	ACC-264

Observed Behavior:
UI type is not read when inline help buttons are accessed. 

Speech output is:
Inline Help Tooltip 
Inline Help Tooltip 
Inline Help Tooltip 
Inline Help Tooltip 
Inline Help Tooltip

Expected Behavior:
Ideally, each & every UI should be read out based upon its type when accessed.  

Screenshot:

![image](https://user-images.githubusercontent.com/88035260/127980279-b31c908c-a65c-4036-8175-a327bd0e65ff.png)

Lokanathan-k:
issue 1: fixed
issue 2: not reproduceable
issue 3: fix

puru-hk:
![image](https://user-images.githubusercontent.com/32538291/134454454-9e646919-448e-470b-99b7-ecd183aefe3d.png)

PR #6434, opened 2021-09-02
fix(core): adding accessability label to the inline help component

Lokanathan-k:
Closes: https://github.com/SAP/fundamental-ngx/issues/6219

adding accessability label and  removing the hardcoded background

Documentation checklist:
- [x] update `README.md`
- [x] Documentation Examples

### DIFF
--- a/apps/docs/src/app/core/component-docs/inline-help/examples/inline-help-example.component.html
+++ b/apps/docs/src/app/core/component-docs/inline-help/examples/inline-help-example.component.html
@@ -1,26 +1,26 @@
 <div class="fd-inline-help-example">
     <span>Default</span>
-    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" tabindex="0" ariaLabel="Inline Help Tooltip"></fd-icon>
+    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" tabindex="0" ariaLabel="Default Inline Help Tooltip"></fd-icon>
 </div>
 
 <div class="fd-inline-help-example">
     <span>Left Position</span>
-    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="left" tabindex="0" ariaLabel="Inline Help Tooltip"></fd-icon>
+    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="left" tabindex="0" ariaLabel="Left Position Inline Help Tooltip"></fd-icon>
 </div>
 
 <div class="fd-inline-help-example">
     <span>Right Position</span>
-    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="right" tabindex="0" ariaLabel="Inline Help Tooltip" ></fd-icon>
+    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="right" tabindex="0" ariaLabel="Left Position Inline Help Tooltip" ></fd-icon>
 </div>
 
 <div class="fd-inline-help-example">
     <span>Top Position</span>
-    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="top" tabindex="0" ariaLabel="Inline Help Tooltip"></fd-icon>
+    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="top" tabindex="0" ariaLabel="Top Position Inline Help Tooltip"></fd-icon>
 </div>
 
 <div class="fd-inline-help-example">
     <span>Bottom Position</span>
-    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="bottom" tabindex="0" ariaLabel="Inline Help Tooltip"></fd-icon>
+    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="bottom" tabindex="0" ariaLabel="Bottom Position Inline Help Tooltip"></fd-icon>
 </div>
 
 <div class="fd-inline-help-example">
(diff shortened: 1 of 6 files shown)


apps/docs/src/app/core/component-docs/inline-help/examples/inline-help-example.component.html
```diff
@@ -1,26 +1,26 @@
...
 <div class="fd-inline-help-example">
     <span>Left Position</span>
-    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="left" tabindex="0" ariaLabel="Inline Help Tooltip"></fd-icon>
+    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="left" tabindex="0" ariaLabel="Left Position Inline Help Tooltip"></fd-icon>
 </div>
 
 <div class="fd-inline-help-example">
     <span>Right Position</span>
-    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="right" tabindex="0" ariaLabel="Inline Help Tooltip" ></fd-icon>
+    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="right" tabindex="0" ariaLabel="Left Position Inline Help Tooltip" ></fd-icon>
```
platon-rov, line 13: `Right Position`?

platon-rov (changes requested):

apps/docs/src/app/core/component-docs/inline-help/examples/inline-help-template-example/inline-help-template-example.component.html
```diff
@@ -1,4 +1,4 @@
-<span [fd-inline-help-template]="templateRef" placement="bottom" aria-describedby="tooltip-message1">Hover on me to see inline help</span>
+<a fd-object-status clickable="true" label="Hover on me to see inline help" [fd-inline-help-template]="templateRef" placement="bottom" aria-describedby="tooltip-message1"></a>
```
platon-rov, line 1: It can be just a span with `tabindex=0`, wouldn't using object-status be the overkill here?

platon-rov (changes requested):

apps/docs/src/app/core/component-docs/inline-help/examples/inline-help-template-example/inline-help-template-example.component.html
Lokanathan-k: if we use tab index the focus is not in sync with the Fiori 3 spec, so the Object status is used

platon-rov: Okay, let it be like that :)

apps/docs/src/app/core/component-docs/inline-help/examples/inline-help-example.component.html
```diff
@@ -1,26 +1,26 @@
...
 <div class="fd-inline-help-example">
     <span>Top Position</span>
-    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="top" tabindex="0" ariaLabel="Inline Help Tooltip"></fd-icon>
+    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="top" tabindex="0" ariaLabel="Top Position Inline Help Tooltip"></fd-icon>
 </div>
 
 <div class="fd-inline-help-example">
     <span>Bottom Position</span>
-    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="bottom" tabindex="0" ariaLabel="Inline Help Tooltip"></fd-icon>
+    <fd-icon glyph="question-mark" fd-inline-help="Inline Help Tooltip" placement="bottom" tabindex="0" ariaLabel="Bottom Position Inline Help Tooltip"></fd-icon>
```
InnaAtanasova, line 23: Just wondering how helpful is the information "Bottom Position/right position, etc" to a blind person. Like in the other cases, can we set the aria-label value automatically, behind the scene, to the value provided for `fd-inline-help`. If the user wants something different they can set manually the `araiLabel` attribute. @mikerodonnell89 what do you think?

Betrozov (approved):

apps/docs/src/app/core/component-docs/inline-help/examples/inline-help-example.component.html
Lokanathan-k: @InnaAtanasova ariaLabel is obtained from the icon component, We need to create a new @input variable as InlineAriaLabel cause the ariaLabel is already used in the Fd-icon component.

mikerodonnell89 (approved):

mikerodonnell89 (changes requested):
(Chrome screenreader) The tooltip isn't being read until the cursor is clicked - I think it should be read on display

apps/docs/src/app/core/component-docs/inline-help/examples/inline-help-example.component.html
InnaAtanasova: How is ariaLabel "obtained" when you provide a string input?

mikerodonnell89 (changes requested):
The tooltip is still not being read on initial display with chrome screenreader

Lokanathan-k:
> The tooltip is still not being read on initial display with chrome screenreader

@mikerodonnell89 i see all the example tooltip being read, can u be specific which example is being read.

apps/docs/src/app/core/component-docs/inline-help/examples/inline-help-example.component.html
platon-rov: @Lokanathan-k please process this one.

mikerodonnell89:
> > The tooltip is still not being read on initial display with chrome screenreader
> 
> @mikerodonnell89 i see all the example tooltip being read, can u be specific which example is being read.

On all the examples, using Chrome's default screenreader, I have to click the element in order for the tooltip to be read.  In my opinion it should just be read when the tooltip is first displayed.  But after looking at bootstrap and angular material both of those libraries perform the same way with Chrome screenreader.  Perhaps a ChromeVox bug.  Approving

platon-rov (approved):

platon-rov:
Please fix travis.

puru-hk:
All issues are fixed

puru-hk (approved):

PR #6552, opened 2021-09-15
docs(core): remove wrong button state example

droshev:
enhance the button state doc with references related to #6206 Issue 1

### DIFF
--- a/apps/docs/src/app/core/component-docs/button/button-docs.component.html
+++ b/apps/docs/src/app/core/component-docs/button/button-docs.component.html
@@ -44,8 +44,7 @@
     Button State
 </fd-docs-section-title>
 <description>
-    'disabled' and 'selected' states can be achieved through native attribute 'disabled' and 'aria-selected'
-    attributes.
+    'disabled' and 'selected' states can be achieved through native 'disabled' / 'aria-disabled' and 'aria-selected' attributes. Regarding the disabled buttons there are 2 flavours reflecting two opinions how to make a button disabled. Feel free to check Mozilla Doc for <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled">disabled attribiut</a> and for <a href="https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-disabled">aria-disabled</a>.
 </description>
 <component-example>
     <fd-button-state-example></fd-button-state-example>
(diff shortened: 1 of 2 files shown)


InnaAtanasova (approved):

InnaAtanasova:
Screen Shot 2021-09-15 at 10 02 13 AM

Source: https://codepen.io/stevef/post/short-note-on-the-disabled-attribute


`disabled`: The disabled attribute often appears a little bit dimmed or transparent. By using the tabulator key and navigating within a form the screen readers will ignore the existence of any disabled button. The screen reader doesn't ever reach the button, then he will never indicate that there's a button.

With `aria-disabled` the screen reader will get the fact that there is a button - by navigating through the page with the tabulator - and it will tell the user that it's dimmed (VoiceOver) or disabled (ChromeVox).

Source: https://a11y-101.com/development/aria-disabled

InnaAtanasova:
Reading more about `disabled` state I think we should not change anything in our documentation. 
As per MDN web docs "[The Boolean disabled attribute, when present, makes the element not mutable, focusable, or even submitted with the form. The user can neither edit nor focus on the control, nor its form control descendants. If the disabled attribute is specified on a form control, the element and its form control descendants do not participate in constraint validation. Often browsers grey out such controls and it won't receive any browsing events, like mouse clicks or focus-related ones.](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled)", which is exactly what our example is doing. 

As for aria-disabled, this is what w3 says: "[Disabled elements might not receive focus from the tab order. For some disabled elements, applications might choose not to support navigation to descendants. In addition to setting the aria-disabled attribute, authors SHOULD change the appearance (grayed out, etc.) to indicate that the item has been disabled.](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-disabled)"

`aria-disabled` is used for elements that do not natively support disabled. Button component supports natively disabled and should use is. img tag needs to use aria-disabled and be styled properly.

InnaAtanasova:
Here's an interesting discussion about disabled state:
https://ux.stackexchange.com/questions/103239/should-disabled-elements-be-focusable-for-accessibility-purposes

Betrozov (approved):

Smiranin (approved):

InnaAtanasova (approved):

droshev:
@InnaAtanasova i reverted the example and added the 2 links from Mozilla as a reference

Issue #5774, opened 2021-06-24
Accessibility Testing_Core_Grid List


fanjum-acc:
Hello Colleagues,

Please find attached issue document for control: Core_Grid List

[Grid List_Core_ARIBA_Controls_Part-1.docx](https://github.com/SAP/fundamental-ngx/files/6709129/Grid.List_Core_ARIBA_Controls_Part-1.docx)

[Grid List_Core_ARIBA_Controls_Part-2.docx](https://github.com/SAP/fundamental-ngx/files/6709134/Grid.List_Core_ARIBA_Controls_Part-2.docx)

[Grid List_Core_ARIBA_Controls_Part-3.docx](https://github.com/SAP/fundamental-ngx/files/6709143/Grid.List_Core_ARIBA_Controls_Part-3.docx)

Veerakarthick:
Hello Colleague,

Retest has been taken results are below.


Issue1:
Observed Behavior:

Text alternative is missing from images present within ‘None Mode’, ‘Single Select Mode’, ‘Single Select Left Mode’, ‘Single Select Right Mode’, ‘Multi Select Mode’, ‘Delete Mode’, ‘States’, ‘Status’, ‘More Button’, ‘Footer’, ‘Drag & Drop’ & ‘Layout’ sections and human icon present under ‘State: Locked’ group area within ‘State’ section

Expected Behavior:

Ideally, all the non-textual UI element and images should have clear and meaningful text alternative in the form of tooltip which will provide the necessary information to the user.


Screenshot:
Single Select Mode:
![image](https://user-images.githubusercontent.com/37998745/152106886-e6e67e96-4ad6-46f1-8fc3-a212c02eda9e.png)

Single Select Left Mode:
![image](https://user-images.githubusercontent.com/37998745/152106978-29062edd-cc16-4a24-b1ea-bc9b44c7a507.png)

Single Select Right Mode:
![image](https://user-images.githubusercontent.com/37998745/152106990-2fb4b092-278c-413e-a779-f4e4a85f9506.png)

Multi Select Mode:
![image](https://user-images.githubusercontent.com/37998745/152107020-eb16eafe-9f5f-4c3d-84f5-68d4919591ce.png)

Delete Mode:
![image](https://user-images.githubusercontent.com/37998745/152107095-f0b20a44-a92a-4dab-ab8b-a0c731ca6181.png)

Group:
[...]

PR #6568, opened 2021-09-16
fix(core): grid list, fix a11y issues

Betrozov:
## Related Issue.
Closes https://github.com/SAP/fundamental-ngx/issues/5774

## Description
Fixed grid list a11y issues

Overview of the changes that were done.
Issues:
1. done
2. not done, this is managed by browser itself and can't me configured in any way
3. done
4. not done, current state matches exactly the fiori 3 design https://sapui5.hana.ondemand.com/#/entity/sap.f.GridList/sample/sap.f.sample.GridListModes https://experience.sap.com/fiori-design-web/grid-list/#select. Correct naming of controls was fixed within the following ISSUE 5
5. done
6. cannot reproduce neither on macOS, nor on windows
7. cannot reproduce neither on macOS, nor on windows
8. not done, this is the limitation of jaws,you should use tab to interact with elements in the list. It works correctly using voiceover
9. not done, this is a correct behavior. Screen reader reads the list as a whole, if you need to read through the options, you should use tab for jaws or step into details in voiceover
10. done
11. done
12. done

##### During Implementation
1. Visual Testing:
- [n/a] visual misalignments/updates
- [n/a] check Light/Dark/HCB/HCW themes
- [n/a] RTL/LTR - proper rendering and labeling
- [n/a] responsiveness(resize)
- [n/a] Content Density (Cozy/Compact/(Condensed))
- [n/a] States - hover/disabled/focused/active/on click/selected/selected hover/press state
- [n/a] Interaction/Animation - open/close, expand/collapse, add/remove, check/uncheck
- [n/a] Mouse vs. Keyboard support
- [n/a] Text Truncation
- [n/a] check for console logs (warnings, errors)
- [n/a] API boundary values
- [n/a] different combinations of components - free style
- [n/a] change the API values during testing
- [n/a] missing API documentation or it is not understandable
- [n/a] poor examples
- [n/a] Stackblitz works for all examples
4. Accessibility testing

##### PR Quality
- [X] the commit message(s) follows the guideline:
- [n/a] Run npm run build-pack-library and test in external application
- [n/a] update `README.md`

### DIFF
--- a/libs/core/src/lib/grid-list/components/grid-list-group-header/grid-list-group-header.component.html
+++ b/libs/core/src/lib/grid-list/components/grid-list-group-header/grid-list-group-header.component.html
@@ -1,6 +1,6 @@
 <div
     class="fd-grid-list__group-header fd-grid-list__group-header--no-margin"
-    tabindex="0"
+    tabindex="-1"
     [attr.aria-label]="ariaLabel"
 >
     <ng-content></ng-content>
--- a/libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.html
+++ b/libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.html
@@ -17,6 +17,7 @@
     (mouseover)="_setActive($event, false)"
     (mouseout)="_setActive($event, true)"
     [attr.aria-selected]="selectionMode === 'none' ? null : !!_selectedItem || _selectedItem === 0"
+    [attr.aria-describedby]="ariaDescribedby"
 >
     <span
         *ngIf="status"
@@ -143,6 +144,8 @@
         [id]="id + '-radio'"
         [name]="'fd-grid-list-item-toolbar-' + selectionMode"
         [attr.aria-label]="ariaLabel"
+        [attr.aria-posinset]="posInSet"
+        [attr.aria-setsize]="_totalItemsCount"
         [value]="value"
         [ngModel]="_selectedItem"
         (ngModelChange)="$event && _selectionItem($event)"
@@ -162,6 +165,8 @@
         class="fd-checkbox fd-checkbox--compact fd-grid-list__checkbox-input"
         [id]="id + '-checkbox'"
         [attr.aria-label]="ariaLabel"
+        [attr.aria-posinset]="posInSet"
+        [attr.aria-setsize]="_totalItemsCount"
         [ngModel]="_selectedItem"
         (ngModelChange)="_selectionItem($event)"
     />
--- a/libs/core/src/lib/grid-list/components/grid-list/grid-list.component.html
+++ b/libs/core/src/lib/grid-list/components/grid-list/grid-list.component.html
@@ -8,9 +8,9 @@
     [class.fd-grid-list--mode-single-select-right]="selectionMode === 'singleSelectRight'"
     [class.fd-grid-list--mode-multi-select]="selectionMode === 'multiSelect'"
 >
-    <div class="fd-container" role="listbox" [attr.aria-label]="ariaLabel">
+    <form class="fd-container" role="listbox" [attr.aria-label]="ariaLabel">
         <div class="fd-row">
             <ng-content></ng-content>
         </div>
-    </div>
+    </form>
 </div>
--- a/libs/core/src/lib/grid-list/components/grid-list/grid-list.component.html
+++ b/libs/core/src/lib/grid-list/components/grid-list/grid-list.component.html
@@ -2,7 +2,6 @@
     [id]="id"
     [style.width]="width"
     class="fd-grid-list"
-    [attr.aria-label]="ariaLabel"
     [class.fd-grid-list--mode-delete]="selectionMode === 'delete'"
     [class.fd-grid-list--mode-single-select]="selectionMode === 'singleSelect'"
     [class.fd-grid-list--mode-single-select-left]="selectionMode === 'singleSelectLeft'"
--- a/libs/core/src/lib/grid-list/components/grid-list/grid-list.component.html
+++ b/libs/core/src/lib/grid-list/components/grid-list/grid-list.component.html
@@ -9,7 +9,7 @@
     [class.fd-grid-list--mode-single-select-right]="selectionMode === 'singleSelectRight'"
     [class.fd-grid-list--mode-multi-select]="selectionMode === 'multiSelect'"
 >
-    <div class="fd-container" role="listbox">
+    <div class="fd-container" role="listbox" [attr.aria-label]="ariaLabel">
         <div class="fd-row">
             <ng-content></ng-content>
         </div>
--- a/apps/docs/src/app/core/component-docs/grid-list/examples/default/grid-list-example.component.html
+++ b/apps/docs/src/app/core/component-docs/grid-list/examples/default/grid-list-example.component.html
@@ -1,5 +1,5 @@
-<fd-grid-list>
-    <fd-grid-list-title-bar [title]="'Products'">
+<fd-grid-list ariaLabel="Products">
+    <fd-grid-list-title-bar title="Products">
         <span fd-grid-list-title-bar-additional-title-item [innerText]="'(' + list.length + ')'"></span>
 
         <fd-grid-list-title-bar-spacer></fd-grid-list-title-bar-spacer>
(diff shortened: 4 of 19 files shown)


InnaAtanasova (changes requested):
Screen Shot 2021-09-16 at 4 47 35 PM

Please explain this. How many aria-labels are needed per element?

libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.html
```diff
@@ -17,6 +17,7 @@
     (mouseover)="_setActive($event, false)"
     (mouseout)="_setActive($event, true)"
     [attr.aria-selected]="selectionMode === 'none' ? null : !!_selectedItem || _selectedItem === 0"
+    [attr.aria-describedby]="ariaDescribedby"
```
InnaAtanasova, line 20: What's the purpose of this? The aria-describedby attribute is used to indicate the IDs of the elements that describe the object. Will there be an element that will have an id that describes this element?

libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.html
```diff
@@ -143,6 +144,8 @@
         [id]="id + '-radio'"
         [name]="'fd-grid-list-item-toolbar-' + selectionMode"
         [attr.aria-label]="ariaLabel"
+        [attr.aria-posinset]="posInSet"
+        [attr.aria-setsize]="_totalItemsCount"
```
InnaAtanasova, line 148: Please check [this documentation](https://www.digitala11y.com/aria-posinset-properties/) and explain why these attributes are added.
I think this is the case for Grid List: "If all items in a set are present in the document structure, it is not necessary to set this attribute, as the user agent can automatically calculate the set size and position for each item."

libs/core/src/lib/grid-list/components/grid-list/grid-list.component.html
```diff
@@ -9,7 +9,7 @@
     [class.fd-grid-list--mode-single-select-right]="selectionMode === 'singleSelectRight'"
     [class.fd-grid-list--mode-multi-select]="selectionMode === 'multiSelect'"
 >
-    <div class="fd-container" role="listbox">
+    <div class="fd-container" role="listbox" [attr.aria-label]="ariaLabel">
```
InnaAtanasova, line 12: I'm not sure the container should have the listbox role. This role is applied in a div that holds elements with role option.In your case the container has also the title bar. 
The element with this role (listbox) needs `aria-activedescendant` and/or `aria-multiselectable` and the element with role option that is selected should have aria-selected.
Please check here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role
Note: I am not an a11y expert so don't take my word for 100% true. That's just what I understood from reading the docs. The people who logged the issue should be able to explain in details what you are supposed to do.

libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.html
dmitry-stepanenko: It looks like screen reader fails to automatically calculate the set size here. This is a fix to ISSUE 11 from the attached doc.

libs/core/src/lib/grid-list/components/grid-list/grid-list.component.html
dmitry-stepanenko: I agree. However, I didn't do related changes are they're in the scope of another issue (please refer https://github.com/SAP/fundamental-ngx/pull/5643#discussion_r648608796)

libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.html
dmitry-stepanenko: Adding this attribute to address ISSUE 12 from the attached docs
> Ideally the sub header should be associated to the list and should be read on navigating to the grid list.

As the most straightforward way of solving this, I'm adding an option to explicitly tie the header with corresponding options. Let me explain why.

Current structure of elements within the grid is as follows

```
<fd-grid-list-group-header>Header 1</fd-grid-list-group-header>
<fd-grid-list-item>....</fd-grid-list-item>
<fd-grid-list-item>....</fd-grid-list-item>
<fd-grid-list-item>....</fd-grid-list-item>
<fd-grid-list-group-header>Header 2</fd-grid-list-group-header>
<fd-grid-list-item>....</fd-grid-list-item>
<fd-grid-list-item>....</fd-grid-list-item>
<fd-grid-list-item>....</fd-grid-list-item>
```

Because of this I don't see any reliable ways to have an implicit association of items with their headers. In my opinion, this can be properly solved in 2 ways:
- by adding some fuzzy js logic to determine which header a particular item belongs to
- by changing the `fd-grid-list-group-header` to `fd-grid-list-group`, that would contain all corresponding items instead of keeping it as a sibling of those items (this would be the best IMO)

Both these options require significant changes, which would be too much for the scope of particular PR. If you think that's reasonable, I can create a separate issue to address this point in the way you see fit. 

Please let me know

dmitry-stepanenko:
@InnaAtanasova 
> Screen Shot 2021-09-16 at 4 47 35 PM
> 
> Please explain this. How many aria-labels are needed per element?

Removed aria-label from `div.fd-grid-list`

libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.html
InnaAtanasova: But the is it really an a11y issue or a screen reader issue? I think we need to comply with the a11y official documentation and guidelines, not to a certain version of a screen reader. 
We are having a meeting with the a11y team next week. Blocking this PR for now. Sorry about that.

libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.html
InnaAtanasova: I see. You can leave the current implementation

InnaAtanasova: Thanks for explaining

InnaAtanasova:
Blocking the PR till we have the meeting with the a11y team.

dmitry-stepanenko:
@InnaAtanasova can I remove blocked label and move the PR to QA?

InnaAtanasova:
> @InnaAtanasova can I remove blocked label and move the PR to QA?

the a11y team needs to verify the changes. I can remove the blocked label but it can't be merged before they say it's ok and answer my questions in the PR.

libs/core/src/lib/grid-list/components/grid-list-item/grid-list-item.component.html
Sanjay-1312: Hello Colleagues, 

The radio button count is reading precise now.

libs/core/src/lib/grid-list/components/grid-list-group-header/grid-list-group-header.component.html
```diff
@@ -1,6 +1,6 @@
 <div
     class="fd-grid-list__group-header fd-grid-list__group-header--no-margin"
-    tabindex="0"
+    tabindex="-1"
```
platon-rov, line 3: Removing `tabindex` attribute at all will make this element not focusable as it's the div.

platon-rov (approved):
Accordingly to the new rules we don't have QA review step, so i guess it should be reviewed by the pr opener.

Veerakarthick:
Hello Colleague,

The reported issue is duplicate of 5774 so i am tracking this issue with the PR5774.

Thanks and regards,
Karthikeyan.c

Karthickc1994:
Hello Colleague,

Retest has been taken with the below environment test results are below:

Issue1: Fixed.
Issue2: Fixed.
Issue3: Fixed.
Issue4: Partially fixed.

Observed behavior:
Associated label has been provided but the label information is not read out by the screen reader.

Screen shot:
Single select left mode:
image

Single select right mode:
image

Multi select mode:
image

Expected behavior:
When accessing the radio button, check box associated label should read out.

Issue5: Not Reproducible

Issue6: Fixed.

Issue7: Fixed.

Issue8: Still reproducible:

Observed behavior:
The list item within the groups are not accessible via Arrow key. only accessible via tab key.

Screenshot:
image

Expected behavior:
The list item within the group should be accessible via arrow keys.

issue9: Fixed.

Issue10: Fixed.

Issue11: Fixed.

Issue12: Fixed.

Retest Environment:

URL:https://deploy-preview-6568--fundamental-ngx.netlify.app/#/core/grid-list
Browser: Chrome Version 101.0.4951.54 (Official Build) (64-bit)
Screen Reader: JAWS2022.2204.20ILM

Regards,
Karthikeyan.,c

Issue #6546, opened 2021-09-15
bug(platform): Table - column sorting bug


droshev:
#### Is this a bug, enhancement, or feature request?
bug originally reported in #6498

#### Briefly describe your proposal.
- [ ] Missing an indication when the column is filtered through the column option. when we sort from the column we have the icon but there is no cue when it is filtered from there

**Steps for reproducing it:**

1. Go to https://fundamental-ngx.netlify.app/#/platform/table#sortable
2. Filter by name, description or price
**Actual result:**
![image](https://user-images.githubusercontent.com/74654139/134303055-cf4576f7-8a67-4cfd-a618-d62f9d3aacad.png)

**Expected result in core is similar example:**
<img width="362" alt="Screen Shot 2021-09-14 at 10 35 35 PM" src="https://user-images.githubusercontent.com/4380815/133361059-78a005c0-ad52-48ba-9ddb-bba89ed28940.png">

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)latest

PR #6703, opened 2021-09-23
fix(platform): table, fix column sorting issue (#6702)

Betrozov:
## Related Issue.
Closes https://github.com/SAP/fundamental-ngx/issues/6546

## Description
FIxed an issue with missing an indication when the column is filtered through the column option. When we sort from the column we have the icon. Now there's the same behavior when it is filtered from there

### Before:
before

### After:

after
##### During Implementation
1. Visual Testing:
- [x] visual misalignments/updates
- [x] check Light/Dark/HCB/HCW themes
- [x] RTL/LTR - proper rendering and labeling
- [x] responsiveness(resize)
- [x] Content Density (Cozy/Compact/(Condensed))
- [x] States - hover/disabled/focused/active/on click/selected/selected hover/press state
- [x] Interaction/Animation - open/close, expand/collapse, add/remove, check/uncheck
- [n/a] Mouse vs. Keyboard support
- [n/a] Text Truncation
- [x] check for console logs (warnings, errors)
- [n/a] API boundary values
- [n/a] different combinations of components - free style
- [n/a] change the API values during testing
- [n/a] missing API documentation or it is not understandable
- [n/a] poor examples
- [n/a] Stackblitz works for all examples
4. Accessibility testing

##### PR Quality
- [x] the commit message(s) follows the guideline:
- [n/a] update `README.md`

### DIFF
--- a/libs/platform/src/lib/table/table.component.scss
+++ b/libs/platform/src/lib/table/table.component.scss
@@ -377,4 +377,21 @@ $fd-table-header-color-var: var(--sapList_HeaderBackground, #f2f2f2);
     .fd-dnd-item {
         position: inherit;
     }
+
+    .fd-table__inner__icons {
+        display: flex;
+
+        fd-icon:not(:last-child) {
+            margin-right: 0.25rem;
+        }
+
+        [dir="rtl"] & {
+            fd-icon {
+                margin-right: 0;
+            }
+            fd-icon:not(:last-child) {
+                margin-left: 0.25rem;
+            }
+        }
+    }
 }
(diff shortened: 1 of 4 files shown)


libs/platform/src/lib/table/table.component.scss
```diff
@@ -415,6 +415,23 @@ $fd-table-border-var: var(--sapList_BorderWidth, 0.0625rem) solid var(--sapList_
     .fd-dnd-item {
         position: inherit;
     }
+
+    .fd-table__inner__icons {
```
platon-rov, line 419: According to the [BEM](https://en.bem.info/methodology/quick-start/#guidelines-for-using-elements) we cannot have elements of element. 

> An element is always part of a block, not another element. This means that element names can't define a hierarchy such as block__elem1__elem2.

But basically these styles should be applied in Fundamental Styles, otherwise we need to refactor it later.

platon-rov (changes requested):

platon-rov (approved):

InnaAtanasova (approved):

PR #6720, opened 2021-09-24
fix(platform): replace duplicate i18n ids

JenniferStamm:
## Related Issue.
Closes SAP/fundamental-ngx/#6467

## Description
Replace duplicate i18n ids with a new unique id, because the localized strings are different from the first usage of the previous ids.


droshev (approved):

droshev:
@JenniferStamm thank you for the contribution! We appreciate it!

Betrozov (approved):

PR #7130, opened 2021-11-08
fix(core): disable field clear at suggestion select

### LINKED ISSUE #6718: multi combo improvement during filtering
enhancement

1. go to multi combo box and type a string to filter the items where you have more than one matching items
2. select the first item

The field clears out the typed string

Would be good to keep the string after the selection so the user can select another item

xandrio:
## Related Issue(s)
closes #6718

## Description

disable field clear at suggestion select

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [x] poor examples

4. Accessibility testing

### DIFF
--- a/libs/core/src/lib/multi-input/multi-input.component.html
+++ b/libs/core/src/lib/multi-input/multi-input.component.html
@@ -111,7 +111,7 @@
         >
             <fd-checkbox
                 [compact]="compact"
-                (ngModelChange)="_handleSelect($event, value)"
+                (ngModelChange)="_handleSelect($event, value, false)"
                 [ngModel]="selected.indexOf(value) !== -1"
             >
             </fd-checkbox>


mikerodonnell89 (changes requested):
If I understand #6718 correctly, the desired behavior would be, taking the first multi-input example: type 'B' and the resulting list would display Banana and Blueberry.  Selecting either option would not clear the input so the user can select the other without having to re-type 'B'. That is not the case I am seeing in the preview deployment

xandrio:
@mikerodonnell89 after discussion was decided do the same as ui5 combobox - there exist difference between click on checkbox and on label, so currently it follow to ui5 logic

mikerodonnell89 (approved):

droshev (approved):

PR #7171, opened 2021-11-11
chore(core,platform): styles version bump

platon-rov:
Fundamental Styles version bump to get fixes & features.

## For the QA

Could you please test that core checkboxes are not broken.


genereg1 (approved):

droshev (changes requested):
@platon-rov when you bump like that do you check the breaking changes (e.g. checkbox)?
if there are we need to bring them

platon-rov:
> @platon-rov when you bump like that do you check the breaking changes (e.g. checkbox)? if there are we need to bring them

agree, my miss, sorry. But shouldn't we change [minor] version in case of breaking change as we do in ngx?

InnaAtanasova (approved):

Issue #6182, opened 2021-08-01
Accessibilty Testing - Core - Time Picker


george861:
Accessibility: Time Picker -Core Component (There are couple of accessibility issues found while testing, so i have attached error document for you reference.)

[Time Picker _Core_Ariba Controls.docx](https://github.com/SAP/fundamental-ngx/files/6915104/Time.Picker._Core_Ariba.Controls.docx)

Time Picker - Core

1.1	Message Details

Raise Ticket	
Category	
Preprocessed By	
Description	

1.2 Application & Test Environment Details

Application Name:	ARIBA Controls- Fundamental Library for Angular
URL:	https://fundamental-ngx.netlify.app/#/core/object-marker

Technology:	Angular 11
Web browser:	Google Chrome 92.0.4515.107 (Official Build) (64-bit)
Operating System:	Microsoft Windows 10 x64
Screen Reader:	JAWS 2021.2105.53 ILM

 

Issue 1

Priority:	Medium
Checkpoint:	ACC-264

(264.1)

•	Search ‘Time Picker’ from the left panel

Observed Behavior:
Case 1
No information is conveyed about the accessed list box ‘Hrs’, ‘Mins’ and ‘Period’  and upon changing the value within the hours , minutes and period, no information is conveyed about changing values.
Case 2
The values within the ‘Hrs’,’Mins’ and ‘Period’ can be changed via up and down arrow keys, no information is conveyed about this behavior to the user.
Case 3
No information is conveyed as the user can navigate from the list ‘hours’ to ‘Mins’ or from ‘Mins’ to ‘period’ via (Left and Right) arrow keys 
Speech:
Decrease hours Button  
To activate press spacebar.

Expected Behavior:
The information should be conveyed about the accessed list box ‘Hrs’, ‘Mins’ and ‘Period’ and upon changing the time in ‘Hours’, in ‘Minutes’ and in Seconds, the information should be conveyed about changing values.
The information should be conveyed as the user navigates from the list ‘hours’ to ‘Mins’ or from ‘Mins’ to ‘Sec’ via (Left and Right) arrow keys. 

NOTE: the issues persists in all similar UI elements ‘Time Picker’

Screenshot:
 
![image](https://user-images.githubusercontent.com/88019003/127782570-61ee2daa-08c5-4b46-b5b8-0f4d6025da02.png)

Issue 2

Priority:	Medium
Checkpoint:	ACC-255

(255.1)

[...]

PR #7280, opened 2021-11-30
fix(core): time component accessibility

dimamarksman:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/6187

## Description

This PR fixes some time component accessibility issues

##### During Implementation

1. Visual Testing:

-   [x] visual misalignments/updates
-   [x] check Light/Dark/HCB/HCW themes
-   [x] RTL/LTR - proper rendering and labeling
-   [x] responsiveness(resize)
-   [x] Content Density (Cozy/Compact/(Condensed))
-   [x] States - hover/disabled/focused/active/on click/selected/selected hover/press state
-   [x] Interaction/Animation - open/close, expand/collapse, add/remove, check/uncheck
-   [x] Mouse vs. Keyboard support
-   [x] Text Truncation

2. API and functional correctness

-   [x] check for console logs (warnings, errors)
-   [x] API boundary values
-   [x] different combinations of components - free style
-   [x] change the API values during testing

3. Documentation and Example validations

-   [x] missing API documentation or it is not understandable
-   [x] poor examples
-   [x] Stackblitz works for all examples

4. Accessibility testing
5. Browser Testing - Edge, Safari, Chrome, Firefox

##### PR Quality

-   [x] the commit message(s) follows the guideline:
        https://github.com/SAP/fundamental-ngx/blob/main/CONTRIBUTING.md
-   [na] tests for the changes that have been done
-   [x] all items on the PR Review Checklist are addressed :
        https://github.com/SAP/fundamental-ngx/wiki/PR-Review-Checklist
-   [x] Run npm run build-pack-library and test in external application
-   [na] update `README.md`
-   [na] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)

PR #7304, opened 2021-12-02
fix(core): core time-picker accessibility

dimamarksman:
## Related Issue(s)

closes https://github.com/sap/fundamental-ngx/issues/6182

## Description

This PR fixes some accessibility issues with in core time-picker

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [na] missing API documentation or it is not understandable
-   [na] poor examples

4. Accessibility testing

##### PR Quality

-   [na] update `README.md`
-   [na] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)

### DIFF
--- a/libs/core/src/lib/shared/services/value-state-message.service.ts
+++ b/libs/core/src/lib/shared/services/value-state-message.service.ts
@@ -1,0 +1,28 @@
+import { Injectable } from '@angular/core';
+
+/**
+ *
+ * This service holds default
+ * value state aria messages
+ * that are needed for accessibility purposes.
+ *
+ * It's intendant to be used with form controls
+ * which supports different value states
+ *
+ */
+@Injectable({
+    providedIn: 'root'
+})
+export class ValueStateAriaMessageService {
+    /** Value state "success" message */
+    readonly success = 'Value State Success';
+
+    /** Value state "information" message */
+    readonly information = 'Value State Information';
+
+    /** Value state "warning" message */
+    readonly warning = 'Value State Warning';
+
+    /** Value state "error" message */
+    readonly error = 'Value State Error';
+}
--- a/libs/core/src/lib/time-picker/time-picker.component.ts
+++ b/libs/core/src/lib/time-picker/time-picker.component.ts
@@ -21,7 +21,7 @@ import { AbstractControl, ControlValueAccessor, NG_VALIDATORS, NG_VALUE_ACCESSOR
 import { Subject, Subscription } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 
-import { Placement, FormStates } from '@fundamental-ngx/core/shared';
+import { Placement, FormStates, ValueStateAriaMessageService } from '@fundamental-ngx/core/shared';
 import { DatetimeAdapter, DATE_TIME_FORMATS, DateTimeFormats } from '@fundamental-ngx/core/datetime';
 import { TimeComponent } from '@fundamental-ngx/core/time';
 import { PopoverFormMessageService } from '@fundamental-ngx/core/form';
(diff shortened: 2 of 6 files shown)


dimamarksman:
blocked until https://github.com/SAP/fundamental-ngx/pull/7280 is merged

dimamarksman:
Issue 3 with the backslash is not fixed. Please see [my comment here](https://github.com/sap/fundamental-ngx/issues/6182#issuecomment-984515503) 
Issue 4 can not be reproduced.

libs/core/src/lib/shared/services/value-state-message.service.ts
```diff
@@ -0,0 +1,28 @@
...
+    readonly success = 'Value State Success';
+
+    /** Value state "information" message */
+    readonly information = 'Value State Information';
+
+    /** Value state "warning" message */
+    readonly warning = 'Value State Warning';
+
+    /** Value state "error" message */
+    readonly error = 'Value State Error';
```
InnaAtanasova, line 27: How are we handling i18n?

InnaAtanasova (commented):
Just one question I added for i18n, the rest looks good

libs/core/src/lib/shared/services/value-state-message.service.ts
dimamarksman: This can be overridden on app level with translated strings

mikerodonnell89: This goes against our current paradigm of allowing the devs to provide translations via inputs, this could be confusing

mikerodonnell89 (changes requested):
see comment above re: i18n

libs/core/src/lib/shared/services/value-state-message.service.ts
dimamarksman: @mikerodonnell89 there are still inputs for these fields which take precedence over these service values.
This service gives us default values for this kind of inputs.
So if app developer wants to redefine default values for components that rely on this it is needed only to override this service and not inputs for every occurrence in the template

dimamarksman: we have the same approach for core calendar labels

nx.json
```diff
@@ -486,7 +486,7 @@
         },
         "core-time-picker": {
             "tags": ["fd"],
-            "implicitDependencies": ["core-utils", "core-time", "core-popover"]
+            "implicitDependencies": ["core-shared", "core-utils", "core-time", "core-popover"]
```
dmitry-stepanenko, line 489: Why is this needed? In PR for Angular 13 migration all implicit dependencies for core packages are being removed now

dimamarksman: the changes you are referencing are not merged yet

dimamarksman: reverted it

dmitry-stepanenko (approved):

libs/core/src/lib/time-picker/time-picker.component.ts
```diff
@@ -220,6 +226,30 @@ export class TimePickerComponent<D>
     @Input()
     parseFormat: unknown;
 
+    /**
+     * Value state "success" aria message.
+     */
+    @Input()
+    valueStateSuccessMessage: string = this._valueStateAriaMessagesService.success;
```
platon-rov, line 233: As `_valueStateAriaMessagesService` is Optional so please use `?.` for optional chaining here and below.

dimamarksman: ValueStateAriaMessageService is not optional
The service itself uses `provideIn: root`

Betrozov (approved):

InnaAtanasova (approved):

dimamarksman:
@george861 @Sanjay-1312 could you please test it

Sanjay-1312:
Issue 1: Fixed.

Issue 2: Fixed.

Issue 3: Fixed.

Issue 4: Fixed.

Retest Details:
https://deploy-preview-7304--fundamental-ngx.netlify.app/#/core/time-picker
JAWS 2021

PR #7407, opened 2021-12-20
fix(core): remove obsolete utils

dmitry-stepanenko:
## Related Issue(s)

n/a

## Description
In core/utils there're a few functions that are no longer needed:
- cloneDeep has been replaced with lodash's implementation. Current implementation could not copy classes properly. 
- `compareObjects` has been replaced with `fast-deep-equal`, because it's the fastest object comparer
- `closestElement` has been removed in favour of `Element.closest()` method, that is supported by all browsers except IE (which we're not targeting anymore)
- `getNestedValue` has been replaced with lodash's `get`
- `mergeDeep` has been replaced with lodash's `merge`

Each change is shipped as a separate commit

##### During Implementation

1. Visual Testing:

-   [n/a] Text Truncation

-   [x] API boundary values

-   [n/a] poor examples

4. Accessibility testing

##### PR Quality

-   [n/a] update `README.md`


dimamarksman:
I would suggest keeping these utils anyway.
If we want to update the implementation of these helpers we can update just it, so imports and interfaces remain the same.
In this case we don't need so many changes in code.

@dmitry-stepanenko what do you think?

dmitry-stepanenko:
@dimamarksman in my opinion it will be quite messy to keep functions that simply wraps other functions without any need. What are the benefits of this? Moreover, `cloneDeep` and `fast-deep-equal` even have the same interfaces

dimamarksman:
@dmitry-stepanenko it gives us encapsulation and strict interfaces for consumers so end consumer (components) doesn't know about third paty libs.
In the future it gives us ability to update helpers implementation without changing 10+ files.
Of course if we can rely on native API we don't need any helper at all in this case

dmitry-stepanenko:
@dimamarksman as discussed, retained the current implementation

apps/docs/src/app/platform/component-docs/platform-icon-tab-bar/examples/platform-icon-tab-bar-configurable-paddings-example/platform-icon-tab-bar-configurable-paddings-example.component.ts
```diff
@@ -1,5 +1,5 @@
 import { Component } from '@angular/core';
-import { cloneDeep } from '@fundamental-ngx/core';
+import { cloneDeep } from 'lodash-es';
```
N1XUS, line 2: please use deep import from lodash-es/cloneDeep

N1XUS: I would also update blacklist of imports to restrict import from the root of lodash

libs/platform/src/lib/table/table.component.ts
```diff
@@ -35,8 +35,9 @@ import {
     RtlService
 } from '@fundamental-ngx/core/utils';
 import { TableRowDirective } from '@fundamental-ngx/core/table';
-import { getNestedValue, isDataSource, isFunction, isString } from '@fundamental-ngx/platform/shared';
+import { isDataSource, isFunction, isString } from '@fundamental-ngx/platform/shared';
 import { PopoverComponent } from '@fundamental-ngx/core/popover';
+import { get } from 'lodash';
```
N1XUS, line 40: maybe lodash-es?

package.json
```diff
@@ -128,6 +128,7 @@
         "intl": "1.2.5",
         "jasminewd2": "2.2.0",
         "karma-viewport": "1.0.8",
+        "lodash-es": "4.17.21",
```
N1XUS, line 131: Please also update subpackages package.json files to reflect this dependency

N1XUS (changes requested):

libs/platform/src/lib/table/pipes/value-by-path.pipe.ts
```diff
@@ -1,10 +1,9 @@
 import { Pipe, PipeTransform } from '@angular/core';
-
-import { getNestedValue } from '@fundamental-ngx/platform/shared';
+import { get } from 'lodash';
```
dimamarksman, line 2: lodash-es?

apps/docs/src/app/platform/component-docs/platform-icon-tab-bar/examples/platform-icon-tab-bar-configurable-paddings-example/platform-icon-tab-bar-configurable-paddings-example.component.ts
dmitry-stepanenko: Tree-shaking works fine, when dependencies are imported from `lodash-es` directly. I've verified.

libs/platform/src/lib/table/pipes/value-by-path.pipe.ts
dmitry-stepanenko: fixed

apps/docs/src/app/platform/component-docs/platform-icon-tab-bar/examples/platform-icon-tab-bar-configurable-paddings-example/platform-icon-tab-bar-configurable-paddings-example.component.ts
dmitry-stepanenko: > I would also update blacklist of imports to restrict import from the root of lodash
This is not easily doable at this point, I've created a separate ticket
https://github.com/SAP/fundamental-ngx/issues/7414

package.json
dmitry-stepanenko: done

dimamarksman (approved):

N1XUS (approved):

Betrozov (approved):

mikerodonnell89 (approved):

Issue #6156, opened 2021-07-30
[Accessibility][Sourcing] Accessibility issue: fd-list inside fd-popover-body doesnt work on hitting enter key


I543348:
#### Is this a bug, enhancement, or feature request?
bug
#### Briefly describe your proposal.
I am using  "@fundamental-ngx/core": "^v0.31.0-rc.125",
for the component fd-popover, referring to the link https://fundamental-ngx.netlify.app/#/core/popover
` <fd-popover-body *ngIf="list1 && list1.length">
            <ul fd-list>
                <li fd-list-item *ngFor="let option of list1">
                    <a [attr.href]="option.url" fd-list-title>{{ option.text }}</a>
                </li>
            </ul>
        </fd-popover-body>`
On hitting enter key when focus is on any option , it doesnt navigate to the option.url but works only on mouse click.
Its more on accessibility issue. Please fix it.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
"@fundamental-ngx/core": "^v0.31.0-rc.125", angular 10
#### If this is a bug, please provide steps for reproducing it.
1) Get the focus on the fd-popover trigger button , hit enter 
2) opens popover body
3) default focus is on the 1st option li 
4) hit enter on it 
5) nothing happens , expected behaviour: should have navigated to some link 
#### Please provide relevant source code if applicable.
`  <fd-popover [focusTrapped]="true" [focusAutoCapture]="true" title="photo 1 in circle">
        <fd-popover-control>
            <fd-avatar size="s" [circle]="true" ariaLabel="photo 1 in circle" image="http://picsum.photos/id/1018/400"></fd-avatar>
        </fd-popover-control>
        <fd-popover-body *ngIf="list1 && list1.length">
            <ul fd-list>
                <li fd-list-item *ngFor="let option of list1">
                    <a [attr.href]="option.url" fd-list-title>{{ option.text }}</a>
                </li>
            </ul>
        </fd-popover-body>
    </fd-popover>`
#### Is there anything else we should know?
As an alternate i am currently adding tabindex=0 and (keydown.enter)="option.url" directly on li tag

skaquib123:
@I543348 , If you are using **fd-list**  in <ul> then you have to use **fd-list-item** in <li> and **fd-checkbox** not fdp-checkbox. There is a compatibility issue between fd and fdp components.

PR #7416, opened 2021-12-21
fix(core): fd-list-link handle by enter key

genereg1:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/6156

## Description

- For now, `fd-list-link` can handle links only by click. Added ability to handle it by Enter key as well 
- Added example `With Link` in Popover doc 

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values
-   [n/a] change the API values during testing

-   [x] poor examples

4. Accessibility testing

##### PR Quality

-   [n/a] update `README.md`

### DIFF
--- a/apps/docs/src/app/core/component-docs/popover/examples/popover-simple/popover-example.component.html
+++ b/apps/docs/src/app/core/component-docs/popover/examples/popover-simple/popover-example.component.html
@@ -122,4 +122,17 @@
             </ul>
         </fd-popover-body>
     </fd-popover>
+
+    <fd-popover placement="bottom" [noArrow]="false" [focusTrapped]="true" [focusAutoCapture]="true">
+        <fd-popover-control>
+            <button fd-button label="With Link"></button>
+        </fd-popover-control>
+        <fd-popover-body *ngIf="list1 && list1.length">
+            <ul fd-list>
+                <li fd-list-item *ngFor="let option of list1">
+                    <a [attr.href]="option.url" fd-list-link>{{ option.text }}</a>
+                </li>
+            </ul>
+        </fd-popover-body>
+    </fd-popover>
 </div>
--- a/libs/core/src/lib/list/directives/list-link.directive.ts
+++ b/libs/core/src/lib/list/directives/list-link.directive.ts
@@ -1,4 +1,4 @@
-import { Attribute, Directive, ElementRef, HostBinding, Input, OnInit } from '@angular/core';
+import { Directive, ElementRef, HostBinding, Input, OnInit } from '@angular/core';
 
 @Directive({
     selector: '[fd-list-link], [fdListLink]'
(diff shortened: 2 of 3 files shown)


libs/core/src/lib/list/list-item/list-item.component.ts
```diff
@@ -161,6 +169,9 @@ export class ListItemComponent extends ListFocusItem implements AfterContentInit
                 this.radio.labelClicked(event);
                 this._muteEvent(event);
             }
+            if (this.link && this._url) {
+                window.open(this._url, '_self');
```
Betrozov, line 173: maybe we need to add `@Input` for target to `list-link.directive` so that the user can define how this link will open?

apps/docs/src/app/core/component-docs/popover/examples/popover-simple/popover-example.component.html
```diff
@@ -122,4 +122,17 @@
             </ul>
         </fd-popover-body>
     </fd-popover>
+
+    <fd-popover placement="bottom" [noArrow]="false" [focusTrapped]="true" [focusAutoCapture]="true">
+        <fd-popover-control>
+            <button fd-button label="With Link"></button>
+        </fd-popover-control>
+        <fd-popover-body *ngIf="list1 && list1.length">
```
Betrozov, line 130:
```suggestion
        <fd-popover-body>
```

Betrozov (changes requested):

libs/core/src/lib/list/list-item/list-item.component.ts
mikerodonnell89: Yes, the application developer needs to be able to choose between this, Angular router etc

libs/core/src/lib/list/directives/list-link.directive.ts
```diff
@@ -22,4 +22,14 @@ export class ListLinkDirective {
     /** @hidden */
     @HostBinding('class.fd-list__link')
     fdListLinkClass = true;
+
+    /** Keeps href string */
+    href: string;
+
+    constructor(private _elementRef: ElementRef) {}
```
N1XUS, line 29:
```suggestion
constructor(private _elementRef: ElementRef, @Attribute('href') public href: string) {}
```

libs/core/src/lib/list/directives/list-link.directive.ts
```diff
@@ -22,4 +22,14 @@ export class ListLinkDirective {
...
+
+    /** @hidden */
+    ngOnInit(): void {
+        this.href = this._elementRef.nativeElement.getAttribute('href');
```
platon-rov, line 33: With @N1XUS's suggestion we can get rid of this.

platon-rov (commented):
I'm not fully getting the idea of having list's items being focusable as ul/ol items neither focusable. Respectively, when we have a link inside of list item - only the link will be focusable and will have `href` attribute. By default it will navigate with mouse click & enter, that' the default browser behavior.

With all words said this pr doesn't make sense also. Share your thought and let me raise the question to ensure that we want have list items focusable.

PR #7428, opened 2021-12-24
feat(core): pagination long & short modes, improvements, styles version bump

### LINKED ISSUE #7376: feat(pagination): improvements

Enhancement.

1. Emit event when itemsPerPage change.
2. Improve filtering of the itemsPerPageOptions.
  * Do not filter option that equals totalItems.
  * If totalItems changes update the available options.
3. improve the visual part following the figma [link](https://www.figma.com/file/YWGZcfjnGy2vLslsghEHXY/Pagination-Pattern?node-id=1%3A6)

0.33.0-rc.214

platon-rov:
## Related Issue(s)

Closes #7376

## Description

* (feat) long & short modes
  If `<= 9` pages all pages shown, button for current page; otherwise `...` shown, input for current page
* (fix) **breaking change**`aria-label`s inputs change
  Previously some strings were created by `ariaLabelInput + value` but for some languages word order may be different
* (chore) styles version bump
* (fix) styles breaking changes adoption

### Before:

image

### After:

image

## Breaking changes

[Wiki page](https://github.com/SAP/fundamental-ngx/wiki/0.34.0-Breaking-Changes#core-pagination-component-aria-label-inputs-update-7428)

* `currentPageAriaLabel` previously wanted specific string with `${currentPage}` inside of it to be passed, now replaced with the function with  `currentPage` as the parameter.
* `pageLabel` previously string value was expected, now replaced with the function with `page` as the parameter.
* `labelBeforeInputMobile` (string), `labelAfterInputMobile` (function, `pagesCount` is the parameter), `inputAriaLabel` (function, `currentPage`, `pagesCount` are the parameters) were added for better a11y support.

### DIFF
--- a/apps/docs/src/app/core/component-docs/pagination/pagination-header/pagination-header.component.html
+++ b/apps/docs/src/app/core/component-docs/pagination/pagination-header/pagination-header.component.html
@@ -5,7 +5,10 @@
 
     Pagination links are navigable via the tab key and selected with the space bar or enter key.<br /><br />
 
-    Pres Enter or Space after filling the input with page number.<br /><br />
+    If <= 9 pages present - all pages shown, button in active state shown for the current page, otherwise "..." (more)
+    elements shown, input with the current page value shown, can be used to navigate to certain page.<br /><br />
+
+    Pres Enter or Space after filling the input with page number to navigate to the certain page.<br /><br />
 
     The pagination component's <code>currentPage</code> value must be updated from your component for the component to
     work properly.<br /><br />
--- a/libs/core/src/lib/pagination/pagination.component.ts
+++ b/libs/core/src/lib/pagination/pagination.component.ts
@@ -111,7 +111,7 @@ export class PaginationComponent implements OnChanges, OnInit, OnDestroy {
      * This property is mainly provided to support reading in the right language for screen reader.
      */
     @Input()
-    itemsPerPageLabel = 'Results per page';
+    itemsPerPageLabel = 'Results per Page';
 
     /** Represents the options for items per page. */
     @Input()
--- a/libs/core/src/lib/pagination/pagination.component.ts
+++ b/libs/core/src/lib/pagination/pagination.component.ts
@@ -50,7 +50,8 @@ let paginationUniqueId = 0;
     providers: [PaginationService],
     host: {
         class: 'fd-pagination',
-        '[class.fd-pagination--mobile]': 'mobile'
+        '[class.fd-pagination--mobile]': 'mobile',
+        '[class.fd-pagination--short]': '_lastPage <= 9'
     },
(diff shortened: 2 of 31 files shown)


apps/docs/src/app/core/component-docs/pagination/pagination-header/pagination-header.component.html
```diff
@@ -5,7 +5,10 @@
 
     Pagination links are navigable via the tab key and selected with the space bar or enter key.<br /><br />
 
-    Pres Enter or Space after filling the input with page number.<br /><br />
+    If <= 9 pages present - all pages shown, button in active state shown for the current page, otherwise "..." (more)
+    elements shown, input with the current page value shown, can be used to navigate to certain page.<br /><br />
+
+    Pres Enter or Space after filling the input with page number to navigate to the certain page.<br /><br />
```
dmitry-stepanenko, line 11:
```suggestion
    Press Enter or Space after filling the input with page number to navigate to the certain page.<br /><br />
```

dmitry-stepanenko (approved):

libs/core/src/lib/pagination/pagination.component.ts
```diff
@@ -270,6 +254,48 @@ export class PaginationComponent implements OnChanges, OnInit, OnDestroy {
         return this._rtlService?.rtl.value;
     }
 
+    /**
+     * Label for the 'Page ' page button. Page number passed as function parameter.
+     * This property is mainly provided to support reading in the right language for screen reader.
+     */
+    @Input()
+    pageLabel: (page: number) => string = (page: number) => `Page ${page}`;
```
dimamarksman, line 262: I'm trying to imaging the main usage of this...
The functional approach pushes a user to keep this function somewhere in his component and use some translation services for it.
As for me it would be much easier to keep this sort of inputs somewhere in the template, so-called declarative approch. In this case user can derive this as a template reference where template itself can provide a page number or whatever is needed for this.
For example:
```
<fd-pagination [pageLabel]="pageLabelTemplateRef" />
<ng-template let-pageNumber="pageNumber" i18n="@@myPagePagination" #pageLabelTemplateRef>
  Page {{pageNumber}}
</ng-template>
```
Does it make sense?

platon-rov: It makes sense, i totally agree with you, but exactly this input used for `[attr.aria-label]` so it cannot be done.

From all of the inputs of this component that are used for labels\aria, only one can be provided in the way proposed by you, so i'm not sure to have both approaches in the component - functions & templates. What do you think?

Betrozov (approved):

platon-rov:
blocked until https://github.com/SAP/fundamental-styles/pull/3067 not merged.

mikerodonnell89 (approved):

Issue #5727, opened 2021-06-22
Calendar: allow to disable back and/or forward buttons 


droshev:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
At the moment the Calendar allows to pass `disableFunction ` so certain days can be disabled. It makes sense to introduce similar functionality for the back and forth buttons of the navigation. (e.g. disable back button if older date is not allowed to be selected.

![Screen Shot 2021-06-21 at 9 24 25 PM](https://user-images.githubusercontent.com/4380815/122848278-37bf0800-d2d7-11eb-924e-0dff56b44487.png)

PR #7574, opened 2022-01-19
fix(core): added disable functions for previous/next buttons in the calendar

Lukakatche:
## Related Issue(s)

closes https://github.com/sap/fundamental-ngx/issues/5727

## Description

added 2 disable functions for previous and next buttons


InnaAtanasova (approved):

libs/core/src/lib/calendar/calendar-header/calendar-header.component.ts
```diff
@@ -212,6 +218,23 @@ export class CalendarHeaderComponent<D> implements OnDestroy, OnInit, OnChanges
     /** Get information about amount of years displayed at once on year view  */
     private _amountOfYearsPerPeriod = 1;
 
+    /**
+     * Function used to disable previous button in the calendar header.
+     */
+
+    @Input()
+    set previousButtonDisableFunction(disableFunc: (date?: D) => boolean) {
+        this.previousButtonDisabled = disableFunc();
```
platon-rov, line 227: Current date not passed to the function, but i guess this is the idea of this feature. Also the function should be invoked every time new date is selected, not only once when binding.

platon-rov (changes requested):

libs/core/src/lib/calendar/calendar-header/calendar-header.component.ts
dmitry-stepanenko, line 227: Agree with @platon-rov:
1. Checks should be applied whenever bindings are changed, date is changed or date view is changed
2. As a context it would be better to pass selected date and selected view type 
3. Arguments should be passed to the function
4. You should also check if provided value is actually a function
5. Extract type of the function to a separate type, that can be used outside of the component

libs/core/src/lib/calendar/calendar.component.ts
```diff
@@ -232,6 +232,18 @@ export class CalendarComponent<D> implements OnInit, ControlValueAccessor, Valid
     @Input()
     disableFunction: (date: D) => boolean = () => false;
 
+    /**
+     * Function used to disable previous button in the calendar header.
+     */
+    @Input()
+    previousButtonDisableFunction: (date?: D) => boolean = () => false;
```
dmitry-stepanenko, line 239: I don't think default value is needed, you should just handle the case provided value is not a function in `calnedar-header`

apps/docs/src/app/core/component-docs/calendar/calendar-docs.component.html
```diff
@@ -164,3 +164,15 @@
     <fd-calendar-form-example></fd-calendar-form-example>
 </component-example>
 <code-example [exampleFiles]="calendarFormSource"></code-example>
+
+<fd-docs-section-title id="reactiveForm" componentName="calendar">
```
dmitry-stepanenko, line 168: wrong id

apps/docs/src/app/core/component-docs/calendar/calendar-docs.component.html
```diff
@@ -164,3 +164,15 @@
...
+    Calendar with disabled previous/next buttons</fd-docs-section-title
+>
+<description
+    >You can disable previous/next buttons on the calendar by passing in previousButtonDisableFunction or
+    nextButtonDisableFunction.</description
+>
+<component-example>
+    <fd-calendar-disabled-navigation-buttons-example></fd-calendar-disabled-navigation-buttons-example>
+</component-example>
+<code-example [exampleFiles]="calendarFormSource"></code-example>
```
dmitry-stepanenko, line 178: wrong source files

apps/docs/src/app/core/component-docs/calendar/examples/calendar-disabled-navigation-buttons-example/calendar-disabled-navigation-buttons-example.html
```diff
@@ -0,0 +1,10 @@
+<form class="flex-form">
+    <div>
+        <fd-calendar
+            calType="single"
+            [startingDayOfWeek]="1"
+            [previousButtonDisableFunction]="previousButtonDisableFunction"
+            [nextButtonDisableFunction]="nextButtonDisableFunction"
+        ></fd-calendar>
+    </div>
+</form>
```
dmitry-stepanenko, line 10: why do we need such complex layout? can it be just the `<fd-calendar>` in root?

e2e/wdio/core/pages/calendar.po.ts
```diff
@@ -18,6 +18,7 @@ export class CalendarPo extends CoreBaseComponentPo {
     mondayCalendar = 'fd-calendar-monday-start-example ';
     internationalCalendar = 'fd-calendar-i18n-example ';
     reactiveCalendarExamples = 'fd-calendar-form-example ';
+    disabledNavigationButtonsCalendar = 'fd-calendar-disabled-navigation-buttons-example';
```
dmitry-stepanenko, line 21: it's not used in e2e tests, so why is this added?

libs/core/src/lib/calendar/calendar.component.html
```diff
@@ -8,6 +8,8 @@
     (closeClicked)="closeClicked.emit()"
     (nextClicked)="handleNextArrowClick()"
     (previousClicked)="handlePreviousArrowClick()"
+    [previousButtonDisabled]="previousButtonDisableFunction(selectedDate)"
+    [nextButtonDisabled]="nextButtonDisableFunction(selectedDate)"
```
dmitry-stepanenko, line 12: This will be invoked on every CD cycle. Any reasons for this? 
According to my comment above:

1. Checks should be applied whenever bindings are changed, date is changed or date view is changed
2. As a context it would be better to pass selected date and selected view type
3. Arguments should be passed to the function
4. You should also check if provided value is actually a function
5. Extract type of the function to a separate type, that can be used outside of the component

apps/docs/src/app/core/component-docs/calendar/examples/calendar-disabled-navigation-buttons-example/calendar-disabled-navigation-buttons-example.component.ts
```diff
@@ -0,0 +1,22 @@
...
+
+@Component({
+    selector: 'fd-calendar-disabled-navigation-buttons-example',
+    templateUrl: 'calendar-disabled-navigation-buttons-example.html'
+})
+export class CalendarDisabledNavigationButtonsExampleComponent {
+    /**
+     * Function used to disable previous button in the calendar header.
+     */
+    previousButtonDisableFunction(): boolean {
```
dmitry-stepanenko, line 11: You should add some logic, so it's clear how to use these functions. Would also be nice to see different use cases for prev and next buttons. And please, use typed arrow functions:
```typescript
import { DisableHeaderButtonFunc } from '@fundamental-ngx/core/calendar';
...
export class ExampleComponent {

previousButtonDisableFunction: DisableHeaderButtonFunc = () => { .... };
nextButtonDisableFunction: DisableHeaderButtonFunc = () => { .... };

```

dmitry-stepanenko (commented):
@Lukakatche my comments above are not really addressed. Also noticed few issues with added docs

mikerodonnell89 (changes requested):
Let's add the inputs to datepicker and datetimepicker as well, most devs will not be using the calendar component on its own

libs/core/src/lib/calendar/calendar-header/calendar-header.component.ts
```diff
@@ -24,6 +24,12 @@ import { CalendarCurrent } from '../models/calendar-current';
 import { CalendarYearGrid } from '../models/calendar-year-grid';
 import { CalendarService } from '../calendar.service';
 
+export type NavigationButtonDisableFunction<D> = (
```
dmitry-stepanenko, line 27: why this type is declared inside header component, where it's not used? Let's move it to separate file in models folder

libs/core/src/lib/calendar/calendar-header/calendar-header.component.html
```diff
@@ -66,7 +66,8 @@
                 fdType="transparent"
                 glyph="slim-arrow-right"
                 [attr.id]="_nextButtonId"
-                [attr.aria-disabled]="false"
+                [attr.aria-disabled]="nextButtonDisabled ? 'true' : 'false'"
```
dmitry-stepanenko, line 69: same thing with aria-disabled not needed

libs/core/src/lib/calendar/calendar.component.ts
```diff
@@ -212,12 +212,36 @@ export class CalendarComponent<D> implements OnInit, ControlValueAccessor, Valid
...
+    /**
+     * @hidden
+     */
+    get currentlyDisplayed(): CalendarCurrent {
+        return this._currentlyDisplayed;
+    }
+
+    set currentlyDisplayed(currentlyDisplayed: CalendarCurrent) {
+        this._currentlyDisplayed = currentlyDisplayed;
+        if (this.previousButtonDisableFunction && this.nextButtonDisableFunction) {
```
dmitry-stepanenko, line 224: remove the check condition from this setter and apply it inside `_setNavigationButtonsStates` function. Additionally, you should check for presence of `previousButtonDisableFunction` and `nextButtonDisableFunction` functions apart

libs/core/src/lib/calendar/calendar.component.ts
```diff
@@ -232,6 +256,58 @@ export class CalendarComponent<D> implements OnInit, ControlValueAccessor, Valid
...
+    set nextButtonDisableFunction(disableFunc: NavigationButtonDisableFunction<D>) {
+        if (typeof disableFunc === 'function') {
+            this._nextButtonDisableFunction = disableFunc;
+            if (this.currentlyDisplayed) {
+                this.nextButtonDisabled = this.nextButtonDisableFunction(
+                    this.selectedDate,
+                    this.currentlyDisplayed,
+                    this.activeView
+                );
+            }
```
dmitry-stepanenko, line 305: just wondering, wouldn't it be more concise to have just check for changed inputs inside `onChanges`?

It can look like this:
```typescript
ngOnChanges(changes: SimpleChanes): void {
  if ('nextButtonDisableFunction' in changes || 'previousButtonDisableFunction' in changes || 'currentlyDisplayed' in changes) {
      this. _setNavigationButtonsStates();
  }
}
```

libs/core/src/lib/calendar/calendar.component.ts
```diff
@@ -643,4 +727,19 @@ export class CalendarComponent<D> implements OnInit, ControlValueAccessor, Valid
...
+        this.previousButtonDisabled = this.previousButtonDisableFunction(
+            this.selectedDate,
+            this.currentlyDisplayed,
+            this.activeView
+        );
+        this.nextButtonDisabled = this.nextButtonDisableFunction(
+            this.selectedDate,
+            this.currentlyDisplayed,
+            this.activeView
+        );
```
dmitry-stepanenko, line 742:
```suggestion
        this.previousButtonDisabled = typeof previousButtonDisableFunction === 'function' && this.previousButtonDisableFunction(
            this.selectedDate,
            this.currentlyDisplayed,
            this.activeView
        );
        this.nextButtonDisabled = typeof this.nextButtonDisableFunction === 'function' && this.nextButtonDisableFunction(
            this.selectedDate,
            this.currentlyDisplayed,
            this.activeView
        );
```

libs/core/src/lib/calendar/calendar.component.ts
```diff
@@ -206,18 +208,51 @@ export class CalendarComponent<D> implements OnInit, ControlValueAccessor, Valid
...
+     * @hidden
+     */
+    get currentlyDisplayed(): CalendarCurrent {
+        return this._currentlyDisplayed;
+    }
+
+    set currentlyDisplayed(currentlyDisplayed: CalendarCurrent) {
+        this._currentlyDisplayed = currentlyDisplayed;
+    }
+
```
dmitry-stepanenko, line 245: in the recent changes you removed all logic from the setter. at this point we can work with `_currentlyDisplayed` directly as it is in main

libs/core/src/lib/calendar/calendar.component.ts
```diff
@@ -286,6 +321,16 @@ export class CalendarComponent<D> implements OnInit, ControlValueAccessor, Valid
         }
     }
 
+    ngOnChanges(changes: SimpleChanges): void {
```
dmitry-stepanenko, line 324: add `@hidden`

apps/docs/src/app/core/component-docs/calendar/examples/calendar-disabled-navigation-buttons-example/calendar-disabled-navigation-buttons-example.component.ts
```diff
@@ -0,0 +1,52 @@
...
+        activeView: FdCalendarView
+    ): boolean {
+        if (date.month === 1) {
+            return true;
+        }
+        if (activeView === 'year') {
+            return true;
+        }
+        return currentlyDisplayedDate?.year > 2023;
+    }
```
dmitry-stepanenko, line 51: sorry, I didn't notice this before: don't use hardcoded dates, because this example will stop working in 2024
```suggestion
     * will disable choosing dates before current year also if we are in year view or if we selected January
     */
    previousButtonDisableFunction(
        date: FdDate,
        currentlyDisplayedDate: CalendarCurrent,
        activeView: FdCalendarView
    ): boolean {
        if (date.month === 1) {
            return true;
        }
        if (activeView === 'year') {
            return true;
        }
        return currentlyDisplayedDate?.year < new Date().getFullYear();
    }

    /**
     * Function used to disable next button in the calendar header.
     * @param date selected date
     * @param currentlyDisplayedDate currently displayed date
     * @param activeView current view of calendar
     *  will disable choosing dates after next year, also if we are in year view or if we selected January
     */
    nextButtonDisableFunction(
        date: FdDate,
        currentlyDisplayedDate: CalendarCurrent,
        activeView: FdCalendarView
    ): boolean {
        if (date.month === 1) {
            return true;
        }
        if (activeView === 'year') {
            return true;
        }
        return currentlyDisplayedDate?.year > (new Date().getFullYear() + 1);
    }
```

dmitry-stepanenko (commented):
looks much better. still noticing few minor things. and this comment is not addressed https://github.com/SAP/fundamental-ngx/pull/7574#discussion_r800566995

dmitry-stepanenko (approved):

mikerodonnell89 (approved):

libs/core/src/lib/calendar/calendar.component.ts
```diff
@@ -218,6 +238,10 @@ export class CalendarComponent<D> implements OnInit, ControlValueAccessor, Valid
     /** @hidden */
     private _adapterStartingDayOfWeek: DaysOfWeek;
 
+    public previousButtonDisabled: boolean;
```
platon-rov, line 241: `@hidden`?

platon-rov (commented):
Minor items.

platon-rov (approved):

Issue #7587, opened 2022-01-21
Defect Hunting for Sprint 80 - Lima


droshev:
Please list all the findings as a thread to this issue.

We will be testing getting inspired [Exploratory Testing Tours](https://github.com/SAP/fundamental-ngx/wiki/Exploratory-Testing-Tours)
- Landmark Tour
- Intellectual Tour
- Back-alley Tour
- Money Tour
- All-Nighter Tour
- Saboteur Tour
- Guidebook Tour
- Obsessive Compulsive Tour
- Supermodel Tour
- Rained-Out Tour
- Antisocial Tour

Browsers:
- MS Edge
- Firefox
- Safari
- Chrome

mikerodonnell89:
- [ ] Multi-input: [this PR](https://github.com/SAP/fundamental-ngx/pull/7588) allows the user to tab to the expand key to open the dropdown of selectable options. However the user then must shift+tab back to the input part of the multi-input, then press tab, to focus the first item in the list. Instead we should have an additional tab keypress, or down arrow press, focus the first item in the dropdown

InnaAtanasova:
- [ ] (Platform) Standard Combobox. Label and Input are not vertically aligned:
<img width="1187" alt="Screen Shot 2022-01-24 at 2 02 47 PM" src="https://user-images.githubusercontent.com/39598672/150847294-3ee9b830-d3f5-452a-a401-1a248f1b21f7.png">
- [ ] (Platform) Combobox with Two Columns example: the selected item is not fully displayed: 
<img width="890" alt="Screen Shot 2022-01-24 at 2 06 04 PM" src="https://user-images.githubusercontent.com/39598672/150847875-382d01aa-63bb-4458-9d89-c5fcab8f65f6.png">
- [ ] (Platform) Date Picker: the validation is not working properly. Invalid year, such as 202 or 20255 is not detected as invalid. 
<img width="395" alt="Screen Shot 2022-01-24 at 2 09 28 PM" src="https://user-images.githubusercontent.com/39598672/150848469-713e6fdb-4e9e-402b-9ce7-47cb9d106ab6.png">
<img width="414" alt="Screen Shot 2022-01-24 at 2 09 16 PM" src="https://user-images.githubusercontent.com/39598672/150848472-feb9d73e-0dc1-43ab-a6b3-b76f49d8aaf2.png">
[...]

PR #7624, opened 2022-01-26
feat(core): tile clickable flag & event emit on click

platon-rov:
## Related Issue(s)

Part of https://github.com/SAP/fundamental-ngx/issues/7587
>  (Core) Tile, has unneeded tab stop (or event on click needed)

## Description

As tile is focusable in [SAPUI5](https://sapui5.hana.ondemand.com/#/entity/sap.m.GenericTile/sample/sap.m.sample.GenericTileAsLaunchTile) i decided to left such behavior but also add:
* event emitting on click (as in the SAPUI5)
* ability to make tile not focusable (of course clicking is not possible also).

### DIFF
--- a/apps/docs/src/app/core/component-docs/tile/examples/tile-examples.component.ts
+++ b/apps/docs/src/app/core/component-docs/tile/examples/tile-examples.component.ts
@@ -83,3 +83,13 @@ export class BadgeTileExampleComponent {}
     templateUrl: './feed-tile-example.component.html'
 })
 export class FeedTileExampleComponent {}
+
+@Component({
+    selector: 'fd-clickable-tile-example',
+    templateUrl: './clickable-tile-example.component.html'
+})
+export class ClickableTileExampleComponent {
+    onTileClick(): void {
+        console.log('tile clicked!');
+    }
+}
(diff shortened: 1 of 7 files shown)


apps/docs/src/app/core/component-docs/tile/examples/tile-examples.component.ts
```diff
@@ -83,3 +83,13 @@ export class BadgeTileExampleComponent {}
...
 })
 export class FeedTileExampleComponent {}
+
+@Component({
+    selector: 'fd-clickable-tile-example',
+    templateUrl: './clickable-tile-example.component.html'
+})
+export class ClickableTileExampleComponent {
+    onTileClick(): void {
+        console.log('tile clicked!');
```
N1XUS, line 93: Maybe use alert to be more noticeable?

N1XUS (approved):

apps/docs/src/app/core/component-docs/tile/examples/tile-examples.component.ts
InnaAtanasova: +1

InnaAtanasova (approved):

Issue #7550, opened 2022-01-14
Critical Defect Hunting issues for Fundmantal-ngx 0.33.0. release


InnaAtanasova:
# Sprint 78
## Platform Schematics

* Adding platform package fails if `add translations` prompt accepted.
https://github.com/SAP/fundamental-ngx/issues/7420
https://github.com/SAP/fundamental-ngx/pull/7421

## Platform Thumbnail
* Dialog looks not correct on small screens, `380px` example
<img width="389" alt="Screenshot 2021-12-22 at 17 26 18" src="https://user-images.githubusercontent.com/20265336/147115919-035586f6-6eca-413f-b416-f234cd9ceacf.png">

# Sprint 77
- [x] No keyboard navigation for pagination. Once you click on a page you can't go to another page using only the keyboard.
<img width="1046" alt="Screen Shot 2021-12-13 at 4 25 53 PM" src="https://user-images.githubusercontent.com/39598672/145891442-428f449f-aa88-4a15-8ae0-fa96c268e127.png">

- [x] clicking on More does not show dropdown with the overflow items
<img width="1058" alt="Screen Shot 2021-12-13 at 5 51 11 PM" src="https://user-images.githubusercontent.com/39598672/145902041-1a4847dd-fdd1-417b-af9c-6bb17a45daf8.png">

platon-rov:
@mikerodonnell89 did you have chance to check why we have these two approaches for tokenizer component, with and without horizontal scrolling?

platon-rov:
Only platform issues left, unassigning myself.

PR #7630, opened 2022-01-28
feat(core): pagination keyboard navigation

platon-rov:
## Related Issue(s)

Part of https://github.com/SAP/fundamental-ngx/issues/7550

## Description

* Shared directives created to provide keyboard navigation (one items focusable and others accessible using arrows).
* Directives applied to the pagination component.

### DIFF
--- a/libs/core/src/lib/pagination/pagination.component.html
+++ b/libs/core/src/lib/pagination/pagination.component.html
@@ -10,78 +10,166 @@
     </div>
 
     <nav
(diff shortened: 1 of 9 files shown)


InnaAtanasova:
probably need to discuss it in the daily but in my opinion the focus should not skip the buttons on tab. Currently when you tab the focus goes from input to input but skip the buttons. Let's discuss it Monday

PR #7651, opened 2022-01-31
test(core,platform): e2e fixes

platon-rov:
## Related Issue(s)

None

## Description

* fixes for broken e2e on main


dmitry-stepanenko:
just curious, why should we get rid of retries?

platon-rov:
> just curious, why should we get rid of retries?

I was testing, but yea, there is no any benefits (time) from disabling retrying. Reverted. Just fixed & skipped some tests.

dmitry-stepanenko (approved):

InnaAtanasova (approved):

PR #7701, opened 2022-02-08
fix(core): fixed card layout performance

### LINKED ISSUE #5511: Fixed Card Layout performance/behaviour issue
**Describe the bug**
When the the browser changes its size, the cards are destroyed and built again, causing the data inside them to disappear (shows busy indicator, makes a call to the data) and appear again. 

**Expected behavior**
Cards content must be stable when the browser is resized and just adapt to the responsive design

platon-rov:
## Related Issue(s)

Closes #5511

## Description

Fixed card layout rework due to the performance issues. Previously there were nested `*ngFor` that were updated on the window's resize doing recalculation/reinitialization of the columns. 

New approach based on the https://github.com/angular/components/issues/13372#issuecomment-838220343, with the layout approach from https://tobiasahlin.com/blog/masonry-with-css. That makes possible to have only one array with all the cards, that positioned using `flex-direction: column` while keeping Z flow of the cards.

Also, due to the new approach `fdDndKeyboardGroup` & `fdDndKeyboardItem` directives that provide dnd with keyboard were extended and rtl support added.

## Screenshots

No visual changes.

## Breaking changes

* Interface CardDropped of the event emitted after the cards reorganised using drag & drop feature changed, container & prevContainer fields removed.
* Input property dndKeyboardDisabled removed from the DndKeyboardItemDirective, please use input property disableKeyboardDragDrop of the DndKeyboardGroupDirective instead.

Wiki https://github.com/SAP/fundamental-ngx/wiki/0.34.0-Breaking-Changes


libs/core/src/lib/fixed-card-layout/fixed-card-layout.component.ts
```diff
@@ -273,66 +392,51 @@ export class FixedCardLayoutComponent implements OnInit, AfterContentInit, After
...
-        this._cardsArray?.sort(comparator);
-        this._cardsArray?.forEach((card, i) => {
-            const index = i % numberOfColumns;
-            columns[index].push(card);
-        });
+/** @hidden Returns number of columns that can fit in current available width for fd-card-layout */
+function getNumberOfColumns(containerWidth: number, cardWidth: number): number {
+    const numberOfCardsWithNoGap = Math.floor(containerWidth / cardWidth); // 960 / 320 = 3
+    const requiredWidthWithGap = numberOfCardsWithNoGap * cardWidth + (numberOfCardsWithNoGap - 1) * CARD_GAP_WIDTH;
+    const columnCount = requiredWidthWithGap > containerWidth ? numberOfCardsWithNoGap - 1 : numberOfCardsWithNoGap;
```
dmitry-stepanenko, line 432: this is quite unlikely to happen, but if containerWidth or cardWidth is 0, or cardWidth > containerWidth, we will have columnCount < 0 or columnCount === Infinity. 
I think with these changes it will be slightly safer
```typescript
function getNumberOfColumns(containerWidth: number, cardWidth: number): number {
    containerWidth ||= 1;
    cardWidth ||= 1;
    const numberOfCardsWithNoGap = Math.max(Math.floor(containerWidth / cardWidth), 1); // 960 / 320 = 3
    const requiredWidthWithGap = numberOfCardsWithNoGap * cardWidth + (numberOfCardsWithNoGap - 1) * CARD_GAP_WIDTH;
    const columnCount = requiredWidthWithGap > containerWidth ? numberOfCardsWithNoGap - 1 : numberOfCardsWithNoGap;
    return Math.max(columnCount, 1)
}
```

Also can you please add unit tests for it?

libs/core/src/lib/utils/drag-and-drop/dnd-keyboard-group/dnd-keyboard-group.directive.ts
```diff
@@ -1,27 +1,97 @@
...
+            event.preventDefault();
+
+            const nextGroupIndex = groupIndex + (isRtl ? -1 : 1);
+            const nextGroup = this.groups[nextGroupIndex];
+
+            this.customTransferFn(group, nextGroup, itemIndex, indexInNextGroup);
+            this._focusDndItem(nextGroupIndex, indexInNextGroup);
+        }
+
+        if (KeyUtil.isKeyCode(event, LEFT_ARROW) && (isRtl ? nextGroupExists : prevGroupExists)) {
```
dmitry-stepanenko, line 66: please use else if, since these conditions are exclusive

apps/docs/src/app/core/component-docs/fixed-card-layout/fixed-card-layout-docs-header/fixed-card-layout-docs-header.component.html
```diff
@@ -1,13 +1,14 @@
 <header>Fixed Card Layout</header>
 <description>
     The fixed card layout supports automatic, easy and fast card design. The cards are ordered in responsive and
-    collapsible columns.The cards are arranged as a “Z” flow: from left to right, starting with the first card on the
-    top left of the page.
+    collapsible columns. The cards are arranged into a “Z” flow: from left to right, starting with the first card on the
+    top left of the page. The maximum number of columns is 10.
```
dmitry-stepanenko, line 5: maybe I'm missing something, but I don't see where it's handled in code. This is just relying on css, right?

dmitry-stepanenko (commented):
nice one! 

noticed a few minor things

libs/core/src/lib/fixed-card-layout/fixed-card-layout.component.scss
```diff
@@ -1 +1,60 @@
 @import '~fundamental-styles/dist/fixed-card-layout';
+
+/** TODO: Move to styles */
```
mikerodonnell89, line 3: @platon-rov can you open a PR for this todo?

apps/docs/src/app/core/component-docs/fixed-card-layout/fixed-card-layout-docs-header/fixed-card-layout-docs-header.component.html
platon-rov: Right. I could find information only about 1-3 columns in wiki, so 10 it's more than enough i think.

libs/core/src/lib/fixed-card-layout/fixed-card-layout.component.scss
platon-rov: Opened https://github.com/SAP/fundamental-styles/pull/3208

dmitry-stepanenko (approved):

libs/core/src/lib/fixed-card-layout/fixed-card-layout.component.scss
platon-rov: Merged, this pr can be reviewed.

InnaAtanasova (approved):

g-cheishvili:
I do not know what, but something is wrong. When I launch locally I get this:
image
Plus I could not make any of the examples, both on hosted and local versions drag at all. Am I doing something wrong? @platon-rov connect with me

platon-rov:
> I do not know what, but something is wrong. When I launch locally I get this. Plus I could not make any of the examples, both on hosted and local versions drag at all. Am I doing something wrong? @platon-rov connect with me

Try now, but do `npm i` before.

PR #7725, opened 2022-02-11
chore(deps): replace master with main in hotfix-publish script

droshev:
replace master with main in hostfix-publish script


g-cheishvili (approved):
Good thing you noticed it

PR #7728, opened 2022-02-12
docs(core,platform): update version menu

droshev:
update version menu

### DIFF
--- a/apps/docs/src/app/documentation/core-helpers/toolbar/toolbar.component.ts
+++ b/apps/docs/src/app/documentation/core-helpers/toolbar/toolbar.component.ts
@@ -82,8 +82,11 @@ export class ToolbarDocsComponent implements OnInit, OnDestroy {
 
     ngOnInit(): void {
         this.versions = [
+            { id: '0.33.2', url: 'https://620423a8f2458b000724fd5f--fundamental-ngx.netlify.app/' },
+            { id: '0.32.0', url: 'https://6130e294b2dc5c00086828de--fundamental-ngx.netlify.app/' },
+            { id: '0.31.0', url: 'https://6116c4d420b1d40007ce5fd9--fundamental-ngx.netlify.app/' },
             { id: '0.30.0', url: 'https://6099b79d6eba2400084bb441--fundamental-ngx.netlify.app/' },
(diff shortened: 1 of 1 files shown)


apps/docs/src/app/documentation/core-helpers/toolbar/toolbar.component.ts
```diff
@@ -82,8 +82,11 @@ export class ToolbarDocsComponent implements OnInit, OnDestroy {
 
     ngOnInit(): void {
         this.versions = [
+            { id: '0.32.2', url: 'https://620423a8f2458b000724fd5f--fundamental-ngx.netlify.app/' },
```
g-cheishvili, line 85: This one is for 0.33.2, but here states that it is for 0.32.2?

apps/docs/src/app/documentation/core-helpers/toolbar/toolbar.component.ts
```diff
@@ -82,8 +82,11 @@ export class ToolbarDocsComponent implements OnInit, OnDestroy {
...
+            { id: '0.32.0', url: 'https://6130e294b2dc5c00086828de--fundamental-ngx.netlify.app/' },
+            { id: '0.31.0', url: 'https://6116c4d420b1d40007ce5fd9--fundamental-ngx.netlify.app/' },
             { id: '0.30.0', url: 'https://6099b79d6eba2400084bb441--fundamental-ngx.netlify.app/' },
-            { id: '0.29.1', url: 'https://60d4a57fa0ac322070e9bc6e--fundamental-ngx.netlify.app/' },
+            { id: '0.29.0', url: 'https://6040dbed0b78130008f102b5--fundamental-ngx.netlify.app/' },
```
g-cheishvili, line 89: I am getting 0.28.2-rc.5 version when navigating here

g-cheishvili (changes requested):
Urls for 0.29.0 and 0.32.2 are incorrect, or ids typed in versions are incorrect. either way, those two should be looked at

apps/docs/src/app/documentation/core-helpers/toolbar/toolbar.component.ts
droshev: that is the closest

g-cheishvili (approved):
Now it's correct 👍

Issue #7609, opened 2022-01-25
Platform Date Picker


droshev:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.

Internationalization of Date Picker - the validation is not working properly. Change the language to German, remove the last digit for the year, the field becomes red (invalid). Add the digit again. The field remains invalid. 
<img width="898" alt="Screen Shot 2022-01-24 at 2 13 52 PM" src="https://user-images.githubusercontent.com/39598672/150849135-bcea3202-c612-40b2-b181-7971f04bea95.png">

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
latest

PR #7745, opened 2022-02-15
fix(core,platform): date & time pickers - fix i18n

dmitry-stepanenko:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/7609

## Description
Currently i18n of Date Picker - the validation is not working properly. Change the language to German, remove the last digit for the year, the field becomes red (invalid). Add the digit again. The field remains invalid.

This is caused by the `FdDateAdapter` not able to parse dates in any locale except the US one. The reason is actually the native date parsing api, that adapter relies on. In order to be able to parse dates in other locales, we have to use `MomentDatetimeAdapter` in our examples.

##### During Implementation

1. Visual Testing:

-   [n/a] Mouse vs. Keyboard support
-   [n/a] Text Truncation

-   [x] API boundary values
-   [n/a] change the API values during testing

-   [x] poor examples

4. Accessibility testing

##### PR Quality

-   [n/a] update `README.md`


mikerodonnell89 (commented):
Our examples using the moment datetime adapter handle validation differently than the fd datetime adapter. With the moment adapter the field doesn't check for validation until it's been cleared entirely then the user types in some invalid value. Should we open a separate issue to get these more in sync? @droshev @dmitry-stepanenko

dmitry-stepanenko:
> Our examples using the moment datetime adapter handle validation differently than the fd datetime adapter. With the moment adapter the field doesn't check for validation until it's been cleared entirely then the user types in some invalid value. Should we open a separate issue to get these more in sync? @droshev @dmitry-stepanenko

Hi, @mikerodonnell89. Validation is there. The thing is `moment` is really flexible in terms of what it can parse, so a lot of expressions that looks invalid, will still be resolved. I've added raw date output for i18n examples, so it's more visual
Feb-16-2022 13-08-49

platon-rov (commented):
I'm okay with the changes but if our own adapter doesn't fully meet the requirements & we propose to switch out of using it we need to consider if it's should be present at all. 

I will raise that question on the todays call.

platon-rov:
Okay, we discussed this a bit.

Moment is currently deprecated, so there is not much sense to stick to it tighter. Let's think what we can do, go with native or any other custom solution.

cc @droshev

platon-rov:
We can start using dayjs to avoid implementing (i.e. spending time) our own on top of native. Also, native has some limitations, especially, about locales, so i think it's not the option here. From what i see dayjs is the smallest alternative but its api very close to the momentjs and some not included initially features can be achieved using plugins.

Another option - date-fns, that, accordingly to the comparison table, has all the features but is a way bigger by its size, has less locales than dayjs and its api haven't been designed to mimic momentjs.

here is the whole comparison table: https://github.com/you-dont-need/You-Dont-Need-Momentjs/blob/master/README.md#feature-parity

cc @droshev @g-cheishvili @N1XUS @dmitry-stepanenko

platon-rov:
MomentJS issue ticket https://github.com/SAP/fundamental-ngx/issues/7904

platon-rov:
We discussed it yesterday, most probably we will be switching to the another library (dayjs or date-fns).

dmitry-stepanenko:
@mikerodonnell89 @platon-rov now when we have updated dayjs adapter, I think we can proceed with this one

platon-rov (approved):

Issue #7670, opened 2022-02-03
Fiori Next Next


droshev:
**Note:** Please put your name BEFORE you start working on a component and mark as the done (checkbox) when the PR for the component is merged.

Tasks related to the component design updates:

- [ ] bring the latest fn package (update the API, update the Breaking Changes (class names, html structure, etc.)
Update existing components
- [x] Button - @g-cheishvili 
    - [ ] Update according to new styleguide
- [ ] Text Field
- [x] Search - @N1XUS 
- [x] Slider - @N1XUS 
- [ ] Tabs - @g-cheishvili 
- [x] Tag - @g-cheishvili 
- [x] Toggles - Steven
    - [ ] switch - 
    - [ ] checkbox - 
    - [ ] radio - 

Enhance existing components
- [x] Button types and semantics -  @g-cheishvili 
- [ ] Menu Button - 
- [ ] Split Button - 
- [ ] Range slider - talk to Inna before you start working on it
- [ ] Tabs - semantic, types, instance tabs, tab container, multi-instance  - @g-cheishvili 
- [ ] Inputs 
    - [ ]  Select - @StevenZhao318
    - [x] Text Input - @mikerodonnell89 
    - [ ] Combo Box - 
    - [ ] Step Input - 
    - [ ] Multi-Input - @mikerodonnell89 
    - [ ] Search - @StevenZhao318
    - [ ] Token - 
    - [ ] Textarea - @StevenZhao318

Implement new components
- [ ] Dialog - 
- [ ] message box -
- [x] Segmented button -  @g-cheishvili 
- [x] List - [PR](https://github.com/SAP/fundamental-ngx/pull/7767)  @g-cheishvili 
- [ ] Illustrated Messages
- [ ] Shell - 
- [ ] Page - footer, title, group title, toolbar
- [---] Cards - the html and classes will be updated in styles
- [ ] Forms - 
- [x] Avatar - @N1XUS
- [x] Progress Bar - @N1XUS
- [x] Generic Tag - @N1XUS

newly added
- [x] Message Strip (Messaging page) - @N1XUS 
- [x] Object Status (Messaging page) - @N1XUS
- [x] Info Label (Messaging page) - @N1XUS
- [ ] Rating Indicator - 
- [x] Notification (Messaging page) - @N1XUS
- [x] Message Toast (Messaging page) - @N1XUS 
- [ ] Popover (Popups) - @platon-rov 
- [ ] Table

N1XUS:
[...]

PR #7767, opened 2022-02-16
feat(fn): added fn-list component and generic approaches

g-cheishvili:
## Related Issue(s)

Part of #7670 

## Description

Implemented Fiori next list component. Intentionally did not wrote unit tests and also there is mixin extension in ListItemComponent class, that will need to be replaced with recommended approach, once we recommend how to deal with such scenario.

### After:
image


mikerodonnell89 (changes requested):
The readonly list items aren't working, should render as a grey list item with no hover interactions https://fundamental-styles.netlify.app/?path=/docs/btp-fn-components-list--byline&globals=fioriVersion:fioriNext#list

mikerodonnell89:
I really like the fnDisabled directive. In a separate PR, we should add a "Utilities" section to the side nav, documentation on how to use the disabled/readonly directives, which demonstrates buttons to turn on and off the disabling like this

```
<ul fn-list>
    <li fn-list-item [fnDisabled]="isInfoBarDisabled" infoBar>
        <span *fnListItemTitle> Disabled info bar </span>
    </li>
</ul>

<button fd-button (click)="isInfoBarDisabled = !isInfoBarDisabled">Toggle info bar disabled</button>
```

and some examples on other components such as button etc.

libs/fn/src/lib/cdk/mixins/CombineBehaviors.ts
```diff
@@ -0,0 +1,10 @@
+// import { Constructor } from '../interfaces/Constructor';
```
InnaAtanasova, line 1: if this file is not used maybe remove it?

InnaAtanasova:
Screen Shot 2022-02-18 at 12 39 33 PM

disabled item is focussable

g-cheishvili:
> 

That happens on netlify deploy, it not up to date. look [here](https://fundamental-ngx-gh--pr7767-feat-fn-list-kbhl2j6r.web.app/fundamental-ngx#/fn/list)

mikerodonnell89:
> > The readonly list items aren't working, should render as a grey list item with no hover interactions https://fundamental-styles.netlify.app/?path=/docs/btp-fn-components-list--byline&globals=fioriVersion:fioriNext#list
> 
> seems like only infoBar has readonly support. is that how it should be or styles is missing something?

You are correct, my bad, but let's make this line an infobar as well so that the fnReadonly directive actually does something

g-cheishvili:
Go away stale bot

mikerodonnell89 (approved):

PR #7790, opened 2022-02-22
feat(fn): introduce Message Toast and Notification components in Fiori Next

N1XUS:
## Related Issue(s)

relates #7670 

## Description
Added Fiori next Message Toast and Notification components.
## Screenshots

image

##### During Implementation

1. Visual Testing:

-   [x] visual misalignments/updates
-   [x] check Light/Dark/HCB/HCW themes
-   [x] RTL/LTR - proper rendering and labeling
-   [x] responsiveness(resize)
-   [x] Content Density (Cozy/Compact/(Condensed))
-   [x] States - hover/disabled/focused/active/on click/selected/selected hover/press state
-   [x] Interaction/Animation - open/close, expand/collapse, add/remove, check/uncheck
-   [x] Mouse vs. Keyboard support
-   [x] Text Truncation

2. API and functional correctness

-   [x] check for console logs (warnings, errors)
-   [x] API boundary values
-   [x] different combinations of components - free style
-   [x] change the API values during testing

3. Documentation and Example validations

-   [x] missing API documentation or it is not understandable
-   [x] poor examples
-   [x] Stackblitz works for all examples

4. Accessibility testing
5. Browser Testing - Edge, Safari, Chrome, Firefox

##### PR Quality

-   [x] the commit message(s) follows the guideline:
        https://github.com/SAP/fundamental-ngx/blob/main/CONTRIBUTING.md
-   [x] tests for the changes that have been done
-   [x] all items on the PR Review Checklist are addressed :
        https://github.com/SAP/fundamental-ngx/wiki/PR-Review-Checklist
-   [x] Run npm run build-pack-library and test in external application
-   [x] update `README.md`
-   [x] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)

PR #7797, opened 2022-02-24
feat(fn): implement fn-input

mikerodonnell89:
Part of #7670 

Implements the fn-input component for Fiori Next. Also adds a class InputBase which other input style components can import to avoid code duplication with some common inputs that most inputs will use.

Blocked, waiting for CDK utils pr


libs/fn/src/lib/input/input-base.ts
```diff
@@ -0,0 +1,31 @@
...
+    @Input()
+    placeholder: string;
+
+    /** Whether or not this input is the 'display' type. */
+    @Input()
+    display = false;
+
+    /** Whether or not this input is disabled. */
+    @Input()
+    fnDisabled = false;
```
platon-rov, line 22: This & `fnReadonly` input properties redundant as for me. 

There is no coercion but it's passed as is to the appropriate directives, in case these fields will be used before coercion we may face unexpected behavior.

Also, appropriate directives can be used theirselves and we can get them (to pass further or anything) in the construction, of course with the `@Optional()`.

cc @g-cheishvili @N1XUS.

mikerodonnell89: Good call, see my recent commits for the update

libs/fn/src/lib/input/input-base.ts
```diff
@@ -0,0 +1,71 @@
...
+        this._listenToDisablingEvents();
+    }
+
+    /** @hidden */
+    ngOnDestroy(): void {
+        this._subscriptions.unsubscribe();
+    }
+
+    /** @hidden */
+    private _listenToDisablingEvents(): void {
```
platon-rov, line 54: Now it's much better, but some same parts of code i found when reviewing @g-cheishvili's `feat/fn-tabs` branch. Maybe there is also some duplications with https://github.com/SAP/fundamental-ngx/pull/7790 I'm aware of getting into the same mess we currently have in core & platform.

platon-rov (commented):
There are no comments about your code, but let's manage first all the architecture stuff.

libs/fn/src/lib/input/input.component.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+        expect(component.inputText).toBe('test');
+        expect(component.onTouched).toHaveBeenCalled();
+        expect(component.onChange).toHaveBeenCalled();
+    });
+
+    it('should setDisabledState', () => {
+        component.setDisabledState(true);
+        expect(component.disabledByForm).toBeTruthy();
+    });
+});
```
dmitry-stepanenko, line 50: can you please add `ngx-cva-test-suite` to this component? just to ensure it works correctly with angular forms. you may refer `libs/platform/src/lib/form/input/input.component.spec.ts`

InnaAtanasova (approved):

PR #7812, opened 2022-02-28
fix(platform): dynamic page onCollapseChange & improvements

platon-rov:
## Related Issue(s)

Closes.
! **MERGE AS**:

BREAKING CHANGE:
rename dynamic page header output

## Description

Platform Dynamic Page

* Header output `onCollapseChange` renamed to `onCollapsedChange` to be consistent with `collapsed` Input property and to be able to use two-way binding. The previous one was marked as deprecated.
* Header internal component removed because it's not used anywhere.
* Example minor fixes.


libs/platform/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.ts
```diff
@@ -104,7 +104,7 @@ export class DynamicPageHeaderComponent {
 
     /** Collapse/Expand change event raised */
     @Output()
-    collapseChange: EventEmitter<DynamicPageCollapseChangeEvent> = new EventEmitter<DynamicPageCollapseChangeEvent>();
+    collapsedChange: EventEmitter<DynamicPageCollapseChangeEvent> = new EventEmitter<DynamicPageCollapseChangeEvent>();
```
dmitry-stepanenko, line 107: I don't think we should really break something
```suggestion
    collapsedChange: EventEmitter<DynamicPageCollapseChangeEvent> = new EventEmitter<DynamicPageCollapseChangeEvent>();
    
    /** @deprecated - use `collapsedChange` instead */
    @Output()
    collapseChange = this.collapsedChange;
```

dmitry-stepanenko (approved):

libs/platform/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.ts
InnaAtanasova: I agree with Dima

InnaAtanasova (changes requested):
Please check Dima's comment and also document in the Wiki page any Breaking changes (if any). Users check mostly the Wiki page when they do a Fund-ngx migration.

InnaAtanasova (approved):

PR #7819, opened 2022-03-01
feat(core,platform): bring horizon themes

droshev:
bring horizon themes

### DIFF
--- a/libs/core/src/lib/utils/services/themes.service.ts
+++ b/libs/core/src/lib/utils/services/themes.service.ts
@@ -50,7 +50,19 @@ export class ThemesService {
         },
         {
             id: 'sap_horizon',
-            name: 'Horizon'
+            name: 'Morning Horizon'
+        },
+        {
+            id: 'sap_horizon_dark',
+            name: 'Evening Horizon'
+        },
+        {
+            id: 'sap_horizon_hcb',
+            name: 'High Contrast Black Horizon'
+        },
+        {
+            id: 'sap_horizon_hcw',
+            name: 'High Contrast White Horizon'
         }
     ];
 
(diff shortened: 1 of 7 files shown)


g-cheishvili (approved):

libs/core/src/lib/utils/services/themes.service.ts
```diff
@@ -50,7 +50,19 @@ export class ThemesService {
...
+            id: 'sap_horizon_dark',
+            name: 'Evening Horizon'
+        },
+        {
+            id: 'sap_horizon_hcb',
+            name: 'High Contrast Black Horizon'
+        },
+        {
+            id: 'sap_horizon_hcw',
+            name: 'High Contrast While Horizon'
```
InnaAtanasova, line 65: `While` typo

InnaAtanasova (changes requested):

libs/core/src/lib/utils/services/themes.service.ts
droshev: good catch~ :)

libs/core/src/lib/utils/services/themes.service.ts
```diff
@@ -50,7 +50,19 @@ export class ThemesService {
...
+        {
+            id: 'sap_horizon_dark',
+            name: 'Evening Horizon'
+        },
+        {
+            id: 'sap_horizon_hcb',
+            name: 'High Contrast Black Horizon'
+        },
+        {
+            id: 'sap_horizon_hcw',
```
mikerodonnell89, line 64: I'm not seeing anything change when I select any of the new themes other than "morning horizon"?

mikerodonnell89 (approved):

Issue #7789, opened 2022-02-21
Write documentation for fnDisabled/fnReadonly directives


mikerodonnell89:
In this PR https://github.com/SAP/fundamental-ngx/pull/7767 Giorgi implemented two new directives, fnDisabled and fnReadonly that allow us, our the developers using our library, to quickly and easily disable components or make them readonly. We should add some documentation for these directives under the "utilities" section.

g-cheishvili:
Not only that, but also there is fnFocusable directive, which gives you ability to give focus feature to the elements, that would otherwise be non-focusable(div, section, span or anything). That other directive also respects above mentioned directives, so much so, it both are applied focusability is not applied

PR #7822, opened 2022-03-01
feat(fn): added helpful directives and utilities

g-cheishvili:
## Related Issue(s)

part of #7789

## Description

Enhanced previously existing directive [fnFocusable], added parent directive. Added selectable list concept. Moved helper functions into the DI. Added support for enhancing base [fnSelectableItem] data, using provider tokens.


libs/fn/src/lib/cdk/selectable-list/SelectComponentRootToken.ts
```diff
@@ -0,0 +1,9 @@
+import { Observable } from 'rxjs';
```
platon-rov, line 1: dash-case-naming instead of UpperCamelCase, please.

platon-rov (approved):

libs/fn/src/lib/cdk/selectable-list/SelectComponentRootToken.ts
g-cheishvili: In case of the interfaces too?

libs/fn/src/lib/cdk/selectable-list/selectable-item.directive.ts
```diff
@@ -0,0 +1,117 @@
...
+import { SelectableItemToken } from './SelectableItemToken';
+import { FN_SELECTABLE_ITEM_PROVIDER } from './selectable-list.tokens';
+import { FN_DISABLED } from '../tokens/disabled';
+import { DisabledBehavior } from '../interfaces/DisabledBehavior';
+import { ReadonlyBehavior } from '../interfaces/ReadonlyBehavior';
+import { FN_READONLY } from '../tokens/readonly';
+import { SelectionService } from './selection.service';
+
+@Directive({
+    selector: '[fnSelectableItem]',
```
platon-rov, line 12: `fnSelectableItem` but `fnFocusable`, let's make them consistent.

g-cheishvili: Which one would be better? Leave "item" postfix or nah?

libs/fn/src/lib/cdk/selectable-list/SelectComponentRootToken.ts
g-cheishvili: I was thinking about having also only [fnSelectable] in case of grouping you know. Every item could have been as well as root, as well as item. So that they would resolve children and add selection logic to them. On self selection it would select probably itself and all of it's children or something like that. Like a Tree

libs/fn/src/lib/cdk/selectable-list/selectable-item.directive.ts
g-cheishvili: I was thinking about having also only [fnSelectable] in case of grouping you know. Every item could have been as well as root, as well as item. So that they would resolve children and add selection logic to them. On self selection it would select probably itself and all of it's children or something like that. Like a Tree

InnaAtanasova: I'd prefer with `item` postfix. This will allow us to add `fnSelectable`  for a group

InnaAtanasova (dismissed):

libs/fn/src/lib/cdk/selectable-list/selectable-item.directive.ts
g-cheishvili: I was having some fun with tree data structure visualisation couple months ago. Check out my profile, it should be there. Same principles can be applied to these too

platon-rov: My thought is that if we have `fnSelectableList` then we should have `fnSelectableItem` and the same for any similar cases. 

If we go with `list` and ` ` (nothing, i mean `fdSelectable`, without the suffix) there will be less coherence between them.

In the same time i don't see any issues if we will be using `fnSelectableItem` without the list, `item` name looks okay even without the container.

g-cheishvili: Makes sense. In any ways we always can update. I will take these inco account in next push sequence

platon-rov: `fnSelectable` sounds more common, when we will be discussing them we will say `selectable` & `selectable list`, there is a big chance to make a mistake, but when it's the `item` and `list` there are no changes to missunderstant it.

mikerodonnell89 (dismissed):
It would be nice to see each new directive used in an example, not even necessarily a new example but on existing component docs

apps/docs/src/app/fn/component-docs/segmented-button/examples/custom-button-component/custom-button.directive.ts
```diff
@@ -1,6 +1,6 @@
 import { Directive, ElementRef, Self } from '@angular/core';
```
platon-rov, line 1: what's the purpose of this example actually (segmented button with custom button directive)? it looks very similar to others (multiple values example), where custom button is not needed. let's enhance this example so we will see the difference or remove it.

libs/fn/src/lib/cdk/HasElementRef.ts
```diff
@@ -1,5 +1,5 @@
 import { ElementRef } from '@angular/core';
```
platon-rov, line 1: dash-case, please.

libs/fn/src/lib/list/selectable-item/selectable-item.directive.ts
```diff
@@ -3,11 +3,16 @@ import {
     ChangeDetectorRef,
```
platon-rov, line 3: for further: i would ask you to split changes to do not have several components updated in the same pr if it's not about refactoring. this pr about the cdk so we can review & merge it then start using cdk in these components.

it's my opinion, we can discuss it if needed.

libs/fn/src/lib/segmented-button/selectable-button.directive.ts
```diff
@@ -0,0 +1,93 @@
+import {
```
platon-rov, line 1: it should be named using `segmented-button` in name to be consisted with the component name, otherwise there may be missunderstanding

platon-rov: `segmented-button-item` or something

platon-rov (commented):
i saw you changed a lot since my last review, in that case, please remove all approvals so doesn't get pr merged accidentally without the review.

g-cheishvili:
> It would be nice to see each new directive used in an example, not even necessarily a new example but on existing component docs

Added for focusable and selectable lists. Will add for directives too

libs/fn/src/lib/cdk/focusable-list/focusable-list.directive.ts
```diff
@@ -0,0 +1,43 @@
...
+    selector: '[fnFocusableList]',
+    exportAs: 'fnFocusableList',
+    providers: [FocusableListService, DestroyedBehavior]
+})
+export class FocusableListDirective implements AfterViewInit {
+    @Input()
+    navigationDirection: 'horizontal' | 'vertical' = 'vertical';
+
+    @Input()
+    contentDirection: 'ltr' | 'rtl' | null = 'ltr';
```
mikerodonnell89, line 17: Will we be using something similar to fundamental-ngx's RtlService for the FN library? BidiModule from Angular CDK? Would be nice for directionality to come automatically and not as an input

mikerodonnell89, line 17: Other components are using fundamental-ngx's RtlService, it would be nice for directionality to come automatically and not as an input

g-cheishvili: Yes, just we have not yet decided how we will gonna handle directionality issue globally on fn

libs/fn/src/lib/cdk/selectable-list/selection.service.ts
```diff
@@ -123,31 +134,17 @@ export class SelectionService<ValueType = any> implements OnDestroy {
...
-                )
-            );
-            events
-                .pipe(
-                    tap(() => this._itemClicked(item)),
-                    takeUntil(unsubscribe$)
-                )
-                .subscribe();
-        }
+    private _isMultipleMode(): boolean {
```
platon-rov, line 146: can be a getter.

platon-rov (commented):
good job!

libs/fn/src/lib/cdk/focusable-list/focusable-list.directive.ts
platon-rov: service was is pretty usable as in my opinion.

apps/docs/src/app/fn/component-docs/utilities/selectable-list/selectable-list-header/selectable-list-header.component.html
```diff
@@ -0,0 +1,9 @@
+<header>Selectable List</header>
+<description>
+    You can quickly build out selectable list collection for your components. For that you will need to use combination
+    of <code>[fnSelectableList]</code> and <code>[fnSelectableItem]</code>
```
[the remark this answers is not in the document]
g-cheishvili: Selectable Item could be used if someone implements it's root component, yes, but that all depends on implementation. Focusable can be independent and planning to move that single directive into separate module for that reason

libs/fn/src/lib/cdk/focusable-list/focusable-list.directive.ts
g-cheishvili: Yes, but that exists on core side yet, needs to be moved and somehow docs adapted to it too

g-cheishvili:
Needs some work on focusable item and selectable item side to correctly recognize state, respecting fnDisabled and fnReadonly

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/examples/di-example/di-example.component.html
```diff
@@ -0,0 +1,9 @@
+<ul fnDisabledRecipient [fnDisabled]="true">
+    <li fnDisabledRecipient>child 1</li>
+    <!--    <li fnDisabledRecipient>-->
```
platon-rov, line 3: should commented code be removed/uncommented?

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/examples/di-example/di-example.component.ts
```diff
@@ -0,0 +1,25 @@
...
+    selector: '[fnDisabledRecipient]',
+    providers: [FnDisabledProvider]
+})
+export class FnDisabledRecipientDirective implements DisabledViewModifier {
+    constructor(private _disabled$: FnDisabledProvider, private _elementRef: ElementRef) {
+        this._disabled$.addViewModifier(this);
+    }
+
+    setDisabledState(isDisabled: boolean): void {
+        // console.log({ fromRecipientDirective: isDisabled, element: this._elementRef.nativeElement });
```
platon-rov, line 23: should commented code be removed/uncommented?

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/fn-disabled-header/fn-disabled-header.component.ts
```diff
@@ -0,0 +1,7 @@
+import { Component } from '@angular/core';
+
+@Component({
+    templateUrl: './fn-disabled-header.component.html',
+    styleUrls: ['./fn-disabled-header.component.scss']
```
platon-rov, line 5: can be removed.

libs/fn/src/lib/cdk/selectable-list/selection.service.ts
platon-rov: still actual.

mikerodonnell89 (changes requested):
- [ ] On the list focus examples: http://localhost:4200/fundamental-ngx#/fn/focusable-list

When I click the "switch background" button, then hit "tab" to focus the first item in the list - i (sometimes, not always) have to press the down arrow twice to focus the nextitem. Same for when tabbing to the horizontal list.  Again this doesn't always happen but sometimes

- [ ] The disabled link can be focused via tab and clicked via enter. I think it is correct for the link to be focusable (https://ux.stackexchange.com/a/103246) but it should not be "clickable" via the keyboard

Screen Shot 2022-03-08 at 10 10 30 AM

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/examples/di-example/di-example.component.html
```diff
@@ -0,0 +1,18 @@
+<ul [fnDisabled]="rootElementDisabled">
+    <li fnDisabledRecipient [fnDisabled]="firstElementDisabled">child 1</li>
+    <li fnDisabledRecipient>child 2</li>
+    <li fnDisabledRecipient>child 3</li>
+</ul>
+
+<div style="margin-top: 20px">
+    <button
```
mikerodonnell89, line 8:
```suggestion
    <button style="margin-bottom: 0.5rem"
```

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/fn-disabled-docs.component.html
```diff
@@ -0,0 +1,52 @@
+<fd-docs-section-title id="simple" componentName="FnDisabled">Basic example</fd-docs-section-title>
+<description>
+    <p>
+        <code>[fnDisabled]</code> is implementation of default view modifier. When passed true or used as an attribute
+        to <b>any</b> element, which is rendered in DOM will have similar traits. <code>[fnDisabled]</code> will add
+        <code>.is-disabled</code>, <code>[disabled]</code> and <code>[aria-disabled]</code> to the target element.
```
mikerodonnell89, line 6:
```suggestion
        Any element can be set as disabled using the <code>[fnDisabled]</code> directive. When passed as true or as an attribute, the <code>[fnDisabled]</code> directive will add
        <code>.is-disabled</code>, <code>[disabled]</code> and <code>[aria-disabled]</code> to the target element.
```

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/fn-disabled-docs.component.html
```diff
@@ -0,0 +1,52 @@
...
+    </p>
+    <p>
+        <code>[fnDisabled]</code> is independent directive and parent or child directives do not have any impact on it.
+        Everything is quite opposite.
```
mikerodonnell89, line 10:
```suggestion
        <code>[fnDisabled]</code> is an independent directive and the parent or child directives do not have any impact on it.
```

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/fn-disabled-docs.component.html
```diff
@@ -0,0 +1,52 @@
...
+</component-example>
+<code-example [exampleFiles]="defaultExample"></code-example>
+
+<fd-docs-section-title id="di" componentName="FnDisabled">FnDisabledProvider</fd-docs-section-title>
+<description>
+    <p>
+        There are cases when you need to tap into the process of declaring certain element disabled. For example, you
+        want to add class <code>is_very_disabled</code>&nbsp;<u>alongside</u> default behavior when disabled and remove
+        when not. You can achieve that two ways, one recommended and one side effect of having notifier. We will gonna
+        go through recommended way in below.
```
mikerodonnell89, line 24:
```suggestion
        There are cases when you need to tap into the process of declaring certain elements disabled. For example, if you
        want to add the class <code>is_very_disabled</code>&nbsp;<u>alongside</u> default behavior when disabled and remove it
        when not. You can achieve that in two ways, one recommended and one side effect of having a notifier. We will 
        go through the recommended way below.
```

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/fn-disabled-docs.component.html
```diff
@@ -0,0 +1,52 @@
...
+        </li>
+        <li>
+            <code>FnDisabledProvider</code> will detect if any ancestor has <code>[fnDisabled]="true"</code> and will
+            also set itself as <b>disabled</b>
+        </li>
+        <li>
+            When you add your recipient component as <code>FnDisabledViewModifier</code> it will be called immediately
+            with current state and also after every consequent state change. You can add as many
+            <code>FnDisabledViewModifier</code> as you want, they will be called one after another, in an order of
+            adding.
```
mikerodonnell89, line 40:
```suggestion
            <code>FnDisabledViewModifier</code> as you want, they will be called one after another, in the order they have been added.
```

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/fn-disabled-docs.component.html
```diff
@@ -0,0 +1,52 @@
...
+    </p>
+    <p>Things to note:</p>
+    <ul>
+        <li>
+            <code>FnDisabledProvider</code> will also detect native HTML <code>[disabled]</code>
+            attribute and will act upon it's change.
+        </li>
+        <li>
+            <code>FnDisabledProvider</code> will detect if any ancestor has <code>[fnDisabled]="true"</code> and will
+            also set itself as <b>disabled</b>
```
mikerodonnell89, line 34: This line contradicts a statement made earlier above, that `fnDisabled` is independent of parent and child directives

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/fn-disabled-header/fn-disabled-header.component.html
```diff
@@ -0,0 +1,17 @@
+<header>Disabled directive</header>
+<description>
+    <p>
+        When you apply <code>[fnDisabled]</code> directive to the element it will get effectively disabled. It will have
```
mikerodonnell89, line 4:
```suggestion
        When you apply the <code>[fnDisabled]</code> directive to the element it will be effectively disabled. It will have
```

apps/docs/src/app/fn/component-docs/utilities/focusable-list/focusable-list-header/focusable-list-header.component.html
```diff
@@ -0,0 +1,11 @@
+<header>Focusable List</header>
+<description>
+    You can quickly build out focusable list collection for your components. For that you will need to use combination
+    of <code>[fnFocusableList]</code> and <code>[fnFocusableItem]</code>. It has vertical and horizontal directions, as
+    well as respects <code>[fnDisabled]</code> and <code>[fnReadonly]</code> directives, so that those elements will be
+    excluded from the selection sequence.
```
mikerodonnell89, line 6:
```suggestion
    You can quickly build out focusable list collection for your components. For that you will need to use a combination
    of <code>[fnFocusableList]</code> and <code>[fnFocusableItem]</code> directives. It has vertical and horizontal directions, and
    it also respects <code>[fnDisabled]</code> and <code>[fnReadonly]</code> directives, so that those elements will be
    excluded from the selection sequence.
```

apps/docs/src/app/fn/component-docs/utilities/selectable-list/selectable-list-docs.component.html
```diff
@@ -0,0 +1,15 @@
+<fd-docs-section-title id="simple" componentName="selectableList">Basic example</fd-docs-section-title>
+<component-example [hasBackground]="false">
+    <fundamental-ngx-selectable-list-default-example></fundamental-ngx-selectable-list-default-example>
+</component-example>
+<code-example [exampleFiles]="defaultExample"></code-example>
+
+<fd-docs-section-title id="advanced" componentName="selectableList">Advanced example</fd-docs-section-title>
+<description>
+    If you have case, where you want to tap in and modify the behavior of selectable item, here is the example of how it
+    can be done
```
mikerodonnell89, line 10:
```suggestion
    This example demonstrates how to tap in to and modify the behavior of a selectable item
```

mikerodonnell89 (changes requested):
Proposed a few suggestions for grammar etc. in the docs. Also please note the comment about about whether or not fnDisabled affects parent/child elements

apps/docs/src/app/fn/component-docs/list/examples/default/list-default-example.component.html
```diff
@@ -1,6 +1,6 @@
 <ul fn-list>
     <li fn-list-group-header>group header</li>
-    <li fn-list-item>
+    <li fn-list-item class="additional_class">
```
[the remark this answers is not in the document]
g-cheishvili: just showing it has that. Previously some components were ignoring classes on components

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/fn-disabled-docs.component.html
g-cheishvili: but fnDisabled has effect on other components in DI tree. Should we skip that scenario here?

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/fn-disabled-docs.component.html
g-cheishvili: No. FnDisabled by itself will not set disabled state when parent changes, but when component has FnDisabledProvider in it, then it will look through dep tree and FnDisabled will have effect on other elements. If component has BOTH FnDisabledProvider and [fnDisabled], then FnDisabledProvider will take effect

libs/fn/src/lib/cdk/focusable-item/focusable.observer.ts
```diff
@@ -0,0 +1,38 @@
...
+    constructor(private _attributeObserver: AttributeObserver) {}
+
+    static isFocusable(element: Element): boolean {
+        if (ReadonlyObserver.isReadonly(element)) {
+            return false;
+        }
+        if (DisabledObserver.isDisabled(element)) {
+            return false;
+        }
+        const tabIndex = parseInt(element.getAttribute('tabindex') + '', 10);
```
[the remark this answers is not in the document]
g-cheishvili: Adde some helper for that

g-cheishvili:
> * [ ]  On the list focus examples: http://localhost:4200/fundamental-ngx#/fn/focusable-list
> 
> When I click the "switch background" button, then hit "tab" to focus the first item in the list - i (sometimes, not always) have to press the down arrow twice to focus the nextitem. Same for when tabbing to the horizontal list. Again this doesn't always happen but sometimes. This problem may not be related to the PR?
> 
> * [ ]  The disabled link can be focused via tab and clicked via enter. I think it is correct for the link to be focusable (https://ux.stackexchange.com/a/103246) but it should not be "clickable" via the keyboard
> 
> Screen Shot 2022-03-08 at 10 10 30 AM
I think these two issues should be addressed separately, or else this pr will never end 😄

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/fn-disabled-docs.component.html
mikerodonnell89: Ok now I understand, thanks

mikerodonnell89:
> > I think these two issues should be addressed separately, or else this pr will never end 😄

Ok, this is fine with me, let's create issues for this after merging

apps/docs/src/app/fn/component-docs/utilities/fn-disabled/fn-disabled-docs.component.html
g-cheishvili: It is nicely visible in FnDisabledProvider implementation. It depends on parent and self injection tokens, but [fnDisabled] does not have any deps, it just adds disabled state default. That's one of the use cases of the FnDisabledProvider, when you want to both watch for outside changes and respect parent changes and act upon their arrival. I plan to add configuration capability to that Provider, so that you'll be able to customize behavior of that service, even add ability to exclude directive from update sequence. But that's for the future

g-cheishvili: In general provider can be used without directive and directive can be used with or without provider. Those are just different sides of the feature. One more concerned on self and nothing else and second one is concerned about everything, even native updates

mikerodonnell89 (approved):

PR #7830, opened 2022-03-02
feat(fn): implement fn-select

StevenZhao318:
## Related Issue(s)

Part of https://github.com/SAP/fundamental-ngx/issues/7670

## Description

Component to be updated and incorporate fn-input after #7797, 
extend functionality of select component and add additional styling to it.

### Before:

image

### After:

image

image

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation


libs/fn/src/lib/select/option/option.component.scss
```diff
@@ -0,0 +1,24 @@
+.fn-select__item {
```
mikerodonnell89, line 1: select styles should go in select scss

mikerodonnell89 (changes requested):
All of the CSS changes in this PR should actually be happening on the styles lib and should be marked with TODO's here in the meantime

mikerodonnell89 (approved):
Approving this now to get fn-select up to date with the current designs, but we will need some updates in fundamental-styles, and fn-select in ngx will need further updates when the base input PR is merged, as well as the new CDK utils

Issue #7815, opened 2022-02-28
Defect Hunting for Sprint 82


InnaAtanasova:
#### Components to be tested:
- [ ] core Calendar - Mike
- [ ] core Menu - Mike
- [x] core Grid List - Inna
- [x] platform Smart Filter Bar - Inna
- [x] core Upload Collection - Inna
- [x] core Avatar - Inna
- [ ] core Table - Mike
- [x] core List - Inna
- [x] core Message Box - Inna
- [x] core Time Picker - Inna

InnaAtanasova:
- [ ] Avatar - zoom icon is not clickable. The zoom icon should trigger an event specified by the user and the Avatar itself should get a ficus. The zoom icon should have a pointer cursor. UI5 example:
<img width="314" alt="Screen Shot 2022-02-28 at 12 23 29 PM" src="https://user-images.githubusercontent.com/39598672/156029993-92f86d49-6fe3-4f35-b4d6-bb39017541dd.png">

mikerodonnell89:
- [ ] Datepicker, reactive form - focusing the input and even changing the value in the input does not consider the input "touched", only considered touched when the popover is openee

InnaAtanasova:
- [ ] The links in the documentation for Link are not working. List (Standard & Byline) documentation pages are not loaded.
<img width="235" alt="Screen Shot 2022-02-28 at 1 58 44 PM" src="https://user-images.githubusercontent.com/39598672/156042391-6cd453d0-47f1-44bf-aaa4-c1b4b209fe63.png">

InnaAtanasova:
- [ ] Action List - the `Show More` item gets double focus. To reproduce: navigate the list with the Tab key. When you reach `List Item 5`, tab and the `Show More` item will get focus. Tab again and the same element is focussed for second time:
<img width="1227" alt="Screen Shot 2022-02-28 at 2 37 30 PM" src="https://user-images.githubusercontent.com/39598672/156047620-d2fc9037-f4cb-468b-954f-43c60014ae81.png">

InnaAtanasova:
- [ ] Time Picker - not sure if it's a bug or desired behaviour. To reproduce, use the keyboard to open the dropdown menu. Use the tab key to switch from hours to minutes to seconds. You need to press the Tab key **twice** to switch from one to another. Also the buttons (arrows) never get focus.

[...]

PR #7833, opened 2022-03-03
feat(core): avatar clickable & event, zoomGlyph event on click

platon-rov:
## Related Issue(s)

Part of #7815.

## Description

* Avatar `clickable` input property & `avatarClicked` output property.
* Avatar `zoomGlyphClicked` output propert, emits only if `zoomGlyph` is set.


dmitry-stepanenko (approved):

InnaAtanasova (approved):
The `zoom icon` should have pointer cursor but it should also come from fund-styles so I will approve this PR.

PR #7856, opened 2022-03-08
fix(core): make state input optional

droshev:
make state input optional -> it allows passing `undefined` for `state`


droshev:
> Just wondering, why only these components? There are many more with a state property without the optional marker

in other places like [this](https://github.com/SAP/fundamental-ngx/blob/b4d937da6e07f850e543ff5e6687d1ebf6f70a77/libs/core/src/lib/datetime-picker/datetime-picker.component.ts#L193) is handled better

platon-rov (approved):

mikerodonnell89 (approved):

PR #7857, opened 2022-03-09
fix(core): upload collection overflow

platon-rov:
## Related Issue(s)

Part of #7815

Pr in styles https://github.com/SAP/fundamental-styles/pull/3227

## Description

Upload collection overflow fixes.

### Before:

image

### After:

image

## BREAKING CHANGE:

* Upload collection item object status now needs `fd-upload-collection-status-item` directive to be applied.


apps/docs/src/app/core/component-docs/upload-collection/examples/upload-collection-complex-example.component.html
```diff
@@ -69,6 +69,8 @@
                     label="Negative Object Status"
                     title="Object status Negative"
                     aria-label="Object status Negative"
+                    class="fd-upload-collection__status-group-item"
+                    textClass="fd-upload-collection__status-group-item-text"
```
dmitry-stepanenko, line 73: these classes bring behavior users should have by default. should we actually force users to manually set them? maybe it's better to apply them internally to have truncation?

platon-rov: We can try to remove `fd-upload-collection__status-group-item-text` but `fd-upload-collection__status-group-item` still will be present as it sets margins.

platon-rov: Update on this one. 

I don't think we can get rid of it, unfortunately. Original object status shouldn't has `text-overflow: ellipsis` and it should has it here. 

First approach - use selector `.fd-upload-collection__status-group .fd-object-status__text` and apply styles **implicitly**. It goes against BEM methodology.

Second approach - apply classes **explicitly**. Requires more actions to do from users but at least is doesn't conflict any other items.

dmitry-stepanenko: @platon-rov maybe we should have someone else's opinion here, but I would go with implicit approach, because:
a) it won't fix the issue for existing use cases
b) since this is considered as a bug, using upload collection without this class will lead to incorrect behavior. Which means it should never be the case

g-cheishvili: I think we should have additional Directive here, called `fd-upload-collection-status-group-item` or something in this manner. It should inject `fd-object-status` and then set these classes. Implicit is the way to go here in opinion, but also for some reason if user wants to override classes, he should have that possibility for sure. Forcing user to explicitly type these classes will be very uncomfortable in my opinion

dmitry-stepanenko (approved):

InnaAtanasova (approved):

PR #7866, opened 2022-03-12
feat(fn,docs): added custom angular event and added docs for fnClicked

g-cheishvili:
## Related Issue(s)
none

## Description

With this update from now on there is no need for existence of (fnClicked) directive on component to be able to gain those unified events. @HostListened, Renderer2.listen, fromEvent will work on fnClicked, just like it works on angular's native (click). Also there is left way for a user to use provider and have previous available methods. No deprecations

### DIFF
--- a/libs/fn/src/lib/cdk/clicked/clicked-event.plugin.ts
+++ b/libs/fn/src/lib/cdk/clicked/clicked-event.plugin.ts
@@ -1,0 +1,74 @@
+import { Inject, Injectable, PLATFORM_ID } from '@angular/core';
+import { ENTER, SPACE } from '@angular/cdk/keycodes';
+import { DOCUMENT } from '@angular/common';
+
+type EventHandlerFunction = ($event: Event) => void;
+type HandlerRemoveFunction = () => void;
+
+@Injectable()
+export class ClickedEventPlugin {
+    constructor(@Inject(DOCUMENT) private document: Document, @Inject(PLATFORM_ID) private platformId: any) {}
+
+    private parseHigherOrderElement(selector: string): EventTarget {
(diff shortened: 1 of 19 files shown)


droshev (approved):

libs/fn/src/lib/cdk/clicked/clicked-event.plugin.ts
```diff
@@ -0,0 +1,74 @@
...
+            target.removeEventListener('keydown', proxyFunction, false);
+        };
+
+        const proxyFunction = (event: Event | MouseEvent | KeyboardEvent): void => {
+            if (event instanceof KeyboardEvent) {
+                if (event.keyCode !== ENTER && event.keyCode !== SPACE) {
+                    return;
+                }
+            }
+            console.log({ event });
```
platon-rov, line 66: redundant.

libs/fn/src/lib/cdk/clicked/clicked-event.plugin.ts
```diff
@@ -0,0 +1,74 @@
...
+import { DOCUMENT } from '@angular/common';
+
+type EventHandlerFunction = ($event: Event) => void;
+type HandlerRemoveFunction = () => void;
+
+@Injectable()
+export class ClickedEventPlugin {
+    constructor(@Inject(DOCUMENT) private document: Document, @Inject(PLATFORM_ID) private platformId: any) {}
+
+    private parseHigherOrderElement(selector: string): EventTarget {
```
platon-rov, line 12: minor: order of methods. public then private.

platon-rov (approved):
good job!

PR #7890, opened 2022-03-18
docs(core): update the documentation for form message

ynnenu:


apps/docs/src/app/core/component-docs/form-message/form-message-header/form-message-header.component.html
```diff
@@ -1,4 +1,8 @@
 <header>Form Message</header>
+<description>
+    The form message is a control that is used as an information bar. It contains information about an object or a
```
InnaAtanasova, line 3: Form message is not a control, it can't be used individually, it's always attached to a control such as input field, textarea, etc.
You can check here https://experience.sap.com/internal/fiori-design-web/input-field/#value-state-and-value-state-message

Issue #7824, opened 2022-03-02
bring more convenience for  i18n in fundamental-ngx components


dmitry-stepanenko:
Currently fundamental-ngx doesn't have unified convenient approach of working with translations:
- core components use inputs 
- platform components use `@angular/localize` package

This brings several issues.

Firstly, since there's no unified approach for working with translations, there's an ambiguity in what approach should be used in a particular case

Secondly, usage of `@angular/localize` brings certain difficulties for those, who does not use it in their apps. Popular packages, like `ngx-translate` or `transloco` packages use json files for translations, which is absolutely incompatible with angular's internal approach. Because of this, in fact, users are forced to use `@angular/localize` in their apps to be able to translate UI components. 

Instead of this we should be technology agnostic in what translation utilities are used.
This can be done by using internal translation vocabulary for our components. It can work similarly to transloco/ngx-translate: we will have a json/js objects, that contain all strings needed for UI components. This data will be supplied to components by using DI mechanism:
- we should have predefined vocabularies for popular languages (en/de/fr/es/ru/ua/it etc)
- if needed, users will be able to override vocabulary for a certain language on any injection level by overriding the injection token

This is a **breaking change** for platform components, that rely on `@angular/localize` tool. However, negative impact should be minimal, if we supply translations for necessary languages.

This tool can be used by core, platform and experimental packages to standardise the requirements.

droshev:
Let's think about that and come up with the next steps

PR #7900, opened 2022-03-21
feat: add i18n package

dmitry-stepanenko:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/7824

## Description
Currently fundamental-ngx doesn't have unified convenient approach of working with translations:
- core components use inputs 
- platform components use `@angular/localize` package

This brings several issues.

Firstly, since there's no unified approach for working with translations, there's an ambiguity in what approach should be used in a particular case

Secondly, usage of `@angular/localize` brings certain difficulties for those, who does not use it in their apps. Popular packages, like `ngx-translate` or `transloco` packages use json files for translations, which is absolutely incompatible with angular's internal approach. Because of this, in fact, users are forced to use `@angular/localize` in their apps to be able to translate UI components. 

Instead of this we should be technology agnostic in what translation utilities are used.
This can be done by using internal translation vocabulary for our components. It can work similarly to transloco/ngx-translate: we will have a json/js objects, that contain all strings needed for UI components. This data will be supplied to components by using DI mechanism:
- we should have predefined vocabularies for popular languages (en/de/fr/es/ru/ua/it etc)
- if needed, users will be able to override vocabulary for a certain language on any injection level by overriding the injection token

This is a **breaking change** for platform components, that rely on `@angular/localize` tool. However, negative impact should be minimal, if we supply translations for necessary languages.

This tool can be used by core, platform and experimental packages to standardise the requirements.


platon-rov (approved):

mikerodonnell89 (commented):
Maybe I am missing something but I think there needs to be a way to display the values that can be translated in the API docs section of each component. Currently, does the developer have to look into the source for each component to find this information?

dmitry-stepanenko:
> Maybe I am missing something but I think there needs to be a way to display the values that can be translated in the API docs section of each component. Currently, does the developer have to look into the source for each component to find this information?

@mikerodonnell89 good point. Added separate docs section for each component, that has translations
image

g-cheishvili:
Better!

mikerodonnell89:
@dmitry-stepanenko should this still be a "draft" pull request?

dmitry-stepanenko:
> @dmitry-stepanenko should this still be a "draft" pull request?

@mikerodonnell89 it's a draft because I don't want it to be accidentally merged. We have to publish i18n package to npm and I'm waiting for the response from @droshev

libs/i18n/src/lib/pipes/fd-translate.pipe.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    transform(key: string, args?: FdLanguageKeyArgs, defaultValue = ''): string {
+        this._key$.next(key);
+        this._args$.next(args);
+
+        return this._value ?? defaultValue;
+    }
+
+    /** @hidden */
+    _interpolate(expression: string, args: FdLanguageKeyArgs = {}): string {
```
platon-rov, line 56: Shouldn't this method be a private?

platon-rov (approved):

libs/i18n/src/lib/pipes/fd-translate.pipe.ts
dmitry-stepanenko: @platon-rov well, it's left public for testing purposes. Generally it's better to avoid testing internals of a class, but this function is pretty important by itself

g-cheishvili (approved):
I can live with this 😄

InnaAtanasova (approved):

g-cheishvili (changes requested):

mikerodonnell89 (approved):

libs/core/src/lib/pagination/pagination.component.ts
```diff
@@ -273,58 +284,65 @@ export class PaginationComponent implements OnChanges, OnInit, OnDestroy {
...
-    inputAriaLabel: (currentPage: number, pagesCount: number) => string = (currentPage: number, pagesCount: number) =>
-        `Page input, Current page, Page ${currentPage} of ${pagesCount}`;
+    inputAriaLabel: (currentPage: number, pagesCount: number) => string;
 
     /** @hidden */
     constructor(
         private readonly paginationService: PaginationService,
         private readonly _cdr: ChangeDetectorRef,
         private readonly _liveAnnouncer: LiveAnnouncer,
+        @Inject(FD_LANGUAGE) private readonly _language: Observable<FdLanguage>,
```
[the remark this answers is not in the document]
mikerodonnell89: This will be addressed in a separate issue

PR #7878, opened 2022-03-15
feat(platform, core): advanced inline help management

g-cheishvili:
## Related Issue(s)

closes #7793

## Description

Added FormFieldLayoutService for managing the layouts

### Before:
image

### After:
image

##### During Implementation

1. Visual Testing:

-   [x] visual misalignments/updates
-   [x] check Light/Dark/HCB/HCW themes
-   [x] RTL/LTR - proper rendering and labeling
-   [x] responsiveness(resize)
-   [x] Content Density (Cozy/Compact/(Condensed))
-   [x] States - hover/disabled/focused/active/on click/selected/selected hover/press state
-   [x] Interaction/Animation - open/close, expand/collapse, add/remove, check/uncheck
-   [x] Mouse vs. Keyboard support
-   [x] Text Truncation

PR #7905, opened 2022-03-23
fix(platform,core): form mis-alignments

### LINKED ISSUE #7825: Form Generator visual mis-alignments
bug

figma link - Forms Review
ping @droshev for more details

g-cheishvili:
## Related Issue(s)

Closes #7825 

### Before:
Spacing-between-inputs-Before
select-and-margin-position-before
label-positions-before
Label-cropping-before
label-case-before
header-box-shadow-before

### After:
Spacing-between-inputs-After
select-and-margin-position-after
label-positions-after
Label-cropping-after
label-case-after
header-box-shadow-after

### DIFF
--- a/libs/core/src/lib/form/form-label/form-label.component.scss
+++ b/libs/core/src/lib/form/form-label/form-label.component.scss
@@ -31,7 +31,7 @@ $form-label-bottom-spacing: 0.125rem;
     display: flex;
     padding-bottom: $form-label-bottom-spacing;
 
-    &__align-end {
+    &--align-end {
         justify-content: flex-end;
     }
 
--- a/libs/core/src/lib/form/form-label/form-label.component.scss
+++ b/libs/core/src/lib/form/form-label/form-label.component.scss
@@ -31,6 +31,10 @@ $form-label-bottom-spacing: 0.125rem;
     display: flex;
     padding-bottom: $form-label-bottom-spacing;
 
+    &__align-end {
+        justify-content: flex-end;
+    }
+
     span.fd-form-label {
         margin-bottom: -$form-label-bottom-spacing;
     }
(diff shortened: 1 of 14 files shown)


g-cheishvili:
https://github.com/SAP/fundamental-ngx/pull/7878 should be merged first

libs/core/src/lib/form/form-label/form-label.component.scss
```diff
@@ -1,6 +1,7 @@
 @import '~fundamental-styles/dist/form-label';
 
 $form-label-inline-help-placement-space: 0rem;
+$form-label-bottom-spacing: 0.125rem;
```
N1XUS, line 4: Maybe this should be also reflected in fundamental-styles?

N1XUS (approved):

libs/core/src/lib/form/form-label/form-label.component.scss
g-cheishvili: Wrapper does not exist on fundamental side at all. It would be a breaking change with no value to it I think

platon-rov (approved):

libs/core/src/lib/form/form-label/form-label.component.scss
```diff
@@ -28,6 +29,15 @@ $form-label-inline-help-placement-space: 0rem;
 .fd-form-label__wrapper {
     position: relative;
     display: flex;
+    padding-bottom: $form-label-bottom-spacing;
+
+    &__align-end {
```
InnaAtanasova, line 34: I know this is Fund-ngx but we should still maintain the BEM methodology for our classes. 
`fd-form-label__wrapper__align-end` is a wrong BEM. You can't have BLOCK__ELEMENT__ELEMENT. In your case "align-end" is a modifier, so you need to have:
`fd-form-label__wrapper--align-end`

PR #7934, opened 2022-04-01
feat(core): added ignore click on selection directive

### LINKED ISSUE #7910: Dynamic Page issue with copy clickable header area
bug

Allow the select text from the clickable header area in Dynamic Page

latest

1. go to https://fundamental-ngx.netlify.app/#/core/dynamic-page#tabbed
2. click on the button for `Default tabs example`
3. try to select text from the header - you will see that instead of having the text selected the subheader collapses/expands

dxp/core/issues/483

g-cheishvili:
## Related Issue(s)

closes #7910 

## Description

Added directive, which will listen on element click and if document has active selection from that element it will not propagate click events. This way we do not have to cancel every event, but only the ones which result selection

### DIFF
--- a/libs/core/src/lib/utils/directives/ignore-click-on-selection/ignore-click-on-selection.directive.ts
+++ b/libs/core/src/lib/utils/directives/ignore-click-on-selection/ignore-click-on-selection.directive.ts
@@ -2,6 +2,11 @@ import { Directive, ElementRef, HostListener, Inject } from '@angular/core';
 import { DOCUMENT } from '@angular/common';
 import { IgnoreClickOnSelectionDirectiveToken } from './tokens';
 
+/**
+ * Directive will stop propagation and prevent default of click, if selected area is coming from
+ * directive's host element. This gives ability to not cancel all clicks on element, just the ones
+ * which include selection
+ */
 @Directive({
     selector: '[fdIgnoreClickOnSelection]',
     providers: [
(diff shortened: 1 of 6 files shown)


droshev (approved):

libs/core/src/lib/utils/directives/ignore-click-on-selection/ignore-click-on-selection.directive.ts
```diff
@@ -0,0 +1,34 @@
...
+        {
+            provide: IgnoreClickOnSelectionDirectiveToken,
+            useExisting: IgnoreClickOnSelectionDirective
+        }
+    ]
+})
+export class IgnoreClickOnSelectionDirective {
+    constructor(@Inject(DOCUMENT) private document: Document, private _elementRef: ElementRef<HTMLElement>) {}
+
+    @HostListener('click', ['$event'])
```
InnaAtanasova, line 17: missing typedocs

libs/core/src/lib/utils/directives/ignore-click-on-selection/ignore-click-on-selection.directive.ts
```diff
@@ -0,0 +1,34 @@
...
+                this._elementRef.nativeElement.isSameNode(selection.anchorNode) ||
+                this._elementRef.nativeElement.isSameNode(selection.anchorNode.parentElement)
+            ) {
+                $event.stopPropagation();
+                $event.preventDefault();
+            }
+        }
+    }
+
+    nativeElement(): Element {
```
InnaAtanasova, line 31: missing typedocs

libs/core/src/lib/utils/directives/ignore-click-on-selection/ignore-click-on-selection.directive.ts
g-cheishvili: dammit

InnaAtanasova:
Please maintain the right-hand section of the PR - labels, project, milestone, assignee...

PR #7956, opened 2022-04-05
fix(core): time picker locale change incorrect behavior

### LINKED ISSUE #6428: bug(core): Open issues on date-/datetime-/time-picker
Mainly bugs

There are some new issues on Date-, Time- and DateTime-Picker. To demonstrate the issues I created an stackblitz example with more details and demo-fields: https://stackblitz.com/edit/fundamental-issue-6428

Fundamental-ngx: 0.31.0. You can see the other deps on stackblitz

Checkout the above mentioned example in stackblitz

Checkout the above mentioned example in stackblitz

g-cheishvili:
## Related Issue(s)

closes #6428 

## Description

This fix will make time picker respect null values, when locale is changed

### DIFF
--- a/libs/core/src/lib/time-picker/time-picker.component.ts
+++ b/libs/core/src/lib/time-picker/time-picker.component.ts
@@ -426,11 +426,8 @@ export class TimePickerComponent<D>
         let formattedTime = '';
 
         try {
-            if (this.allowNull) {
-                const parsed = this._dateTimeAdapter.parse(time, this.getDisplayFormat());
-                if (!this._dateTimeAdapter.isValid(parsed) && time === null) {
-                    return formattedTime;
-                }
+            if (this.allowNull && time === null) {
+                return '';
             }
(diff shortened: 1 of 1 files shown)


libs/core/src/lib/time-picker/time-picker.component.ts
```diff
@@ -422,6 +426,12 @@ export class TimePickerComponent<D>
         let formattedTime = '';
 
         try {
+            if (this.allowNull) {
+                const parsed = this._dateTimeAdapter.parse(time, this.getDisplayFormat());
+                if (!this._dateTimeAdapter.isValid(parsed) && time === null) {
+                    return formattedTime;
+                }
```
platon-rov, line 433: Can be simplified

```
if (this.allowNull && time === null) {
    return '';
}
```

platon-rov (changes requested):

platon-rov (approved):

PR #7964, opened 2022-04-06
feat(platform): enhancement of form generator and form container

### LINKED ISSUE #7955: feat(platform): enhancement of form generator and form container
enhancement

After meeting with UIX team they had few notes for further enhancements of form components, namely:

- Icon should be `hint` instead of `message-information` as a default in inline help
- Icon should have "interactive" color
- Default triggers for inline help should be "click" instead of "hover"
- Inline help popover's default placement should not hide object, to which it is attached to
- Form label should not enforce TitleCase programmatically, but examples, which are shown in the Docs should have TitleCase
- Labels with inline help needs fix to better align vertically with each other

newest

Additionally UIX team will come up with a solution to cases when inline help is attached to the inputs, which do not extend to the allocated space maximally. In those cases currently inline help icon is far away from input and visually it appears detached from the form item

g-cheishvili:
## Related Issue(s)

closes #7955 

## Description

- [x] Icon should be hint instead of message-information as a default in inline help
- [x] Icon should have "interactive" color
- [x] Default triggers for inline help should be "click" instead of "hover"
- [x] Inline help popover's default placement should not hide object, to which it is attached to
- [x] Form label should not enforce TitleCase programmatically, but examples, which are shown in the Docs should have TitleCase
- [x] Labels with inline help needs fix to better align vertically with each other

### Before:
popover-position-before
icon-before
form-label-before

### After:
popover-position-after
icon-after
form-label-after

### DIFF
--- a/libs/core/src/lib/link/link.component.ts
+++ b/libs/core/src/lib/link/link.component.ts
@@ -39,6 +39,9 @@ export class LinkComponent implements OnChanges, OnInit, CssClassBuilder {
     @Input()
     subtle: boolean;
 
+    @Input()
+    undecorated: boolean;
+
     /** @hidden */
     constructor(private _elementRef: ElementRef) {}
 
@@ -58,6 +61,7 @@ export class LinkComponent implements OnChanges, OnInit, CssClassBuilder {
             this.disabled ? 'is-disabled' : '',
             this.inverted ? `fd-link--inverted` : '',
             this.subtle ? 'fd-link--subtle' : '',
+            this.undecorated ? 'fd-link--undecorated' : '',
             this.class
         ];
     }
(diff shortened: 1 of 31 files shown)


libs/core/src/lib/link/link.component.ts
```diff
@@ -39,6 +39,9 @@ export class LinkComponent implements OnChanges, OnInit, CssClassBuilder {
     @Input()
     subtle: boolean;
 
+    @Input()
```
platon-rov, line 42: Typedocs.

platon-rov (approved):

N1XUS (approved):

droshev (approved):

mengbingguo:
> for point 3 I suggest we open a separate feature request as it will require further discussions and changes

Thanks @g-cheishvili @droshev I opened a new GH issue for this. Pls let me know if more input is needed from my side.

PR #7967, opened 2022-04-07
fix(platform): add limitless option to multi combobox

### LINKED ISSUE #7862: <fdp-multi-comboxbox>: [selectedItems] property populated but view is not rendering selected values
## Is this a bug, enhancement, or feature request?
Bug

## Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Our app is currently using v0.31.x, but issue also found on latest release v0.33.x

## If this is a bug, please provide steps for reproducing it.
- As a user, they can only see a small set of values (from the whole data source value set) upon opening the dropdown
 
- User could not scroll the dropdown to see the rest of the data source values, but they could type to search and find their desired values. It appears that the component is not rendering all the data source values upon opening the dropdown. Is this the expected behavior of the component (due to perf concern if the data source set is too large)?
 
- When a user has typed and selected a matching value, the (selectionChange) event is triggered, and as per our business logic requirement, we will need to perform validation API call and then refresh the component view (basically doing some re-initialization)
 
- Now, after refreshing the view, even though we have already chosen to bind the [selectedItems] property to a certain array value, we could not see these multi-select values being rendered inside the input box.
However, if user now types certain keyword and it matches with their previously selected choices, we can see that from opening the dropdown, these previous values are pre-selected and highlighted (since we have bound the ‘selectedItems’ properties already). Like below:
Screen Shot 2022-03-08 at 5 03 56 PM

- So our main question is: since we have the data already bound to the component’s ‘selectedItems’ property, why isn’t the view rendering these values? Basically, the model is there, but the view is not updated or rendered

## Is there anything else we should know?
Similar issues were also found for the \<fd-multi-input\> component

aberikashvili:
## Related Issue(s)

closes #7862 

##### During Implementation

1. Visual Testing:

-   [ ] visual misalignments/updates
-   [ ] check Light/Dark/HCB/HCW themes
-   [ ] RTL/LTR - proper rendering and labeling
-   [ ] responsiveness(resize)
-   [ ] Content Density (Cozy/Compact/(Condensed))
-   [ ] States - hover/disabled/focused/active/on click/selected/selected hover/press state
-   [ ] Interaction/Animation - open/close, expand/collapse, add/remove, check/uncheck
-   [ ] Mouse vs. Keyboard support
-   [ ] Text Truncation

-   [ ] check for console logs (warnings, errors)
-   [ ] API boundary values
-   [ ] different combinations of components - free style
-   [ ] change the API values during testing

-   [ ] poor examples
-   [ ] Stackblitz works for all examples

4. Accessibility testing

##### PR Quality

-   [ ] update `README.md`


libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
```diff
@@ -170,6 +172,9 @@ export abstract class BaseMultiCombobox extends CollectionBaseInput implements A
     @Input()
     invalidEntryMessage = 'Invalid entry';
 
+    @Input()
+    limitless = true;
```
genereg1, line 176: typedoc

libs/platform/src/lib/shared/domain/data-source.ts
```diff
@@ -162,6 +162,15 @@ export class ComboBoxDataSource<T> implements DataSource<T> {
 
     protected _dataLoading = false;
 
+    protected _limitless = false;
```
g-cheishvili, line 165: Why have setter and getter when nothing happens in them?

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
```diff
@@ -714,7 +725,11 @@ export abstract class BaseMultiCombobox extends CollectionBaseInput implements A
         // initial data fetch
         const map = new Map();
         map.set('query', '*');
-        map.set('limit', 12);
+
+        if (!this.limitless) {
+            map.set('limit', MAP_LIMIT);
```
g-cheishvili, line 730: Is there any particular reason behind that "12"? or it can be anything user wants and can be put out to configuration?

aberikashvili: It was implemented before I touch this component, I just extracted it to constant

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
```diff
@@ -169,6 +172,14 @@ export abstract class BaseMultiCombobox extends CollectionBaseInput implements A
     /** Sets invalid entry message. */
     @Input()
     invalidEntryMessage = 'Invalid entry';
+    private _limitless = false;
+    @Input()
+    set limitless(val: boolean) {
```
platon-rov, line 177: typedoc.

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
```diff
@@ -368,6 +382,10 @@ export abstract class BaseMultiCombobox extends CollectionBaseInput implements A
         super(_cd, ngControl, ngForm, formField, formControl);
     }
 
+    ngOnInit(): void {
+        this._initializeDataSource(this._data);
```
platon-rov, line 386: If datasource isn't set till this moment it wont be opened, and previously we had this in setted, why is this changed?

aberikashvili: So, initialization of the datasource, depends on the value of limitless property, which is passed using @Input decorator as well. Leaving DS initialization in the setter, will cause regression, for example: if users will call the component using the following order of the inputs ` <... [datasource]=".." [limitless]="..">`, we won't have limitless value ready for the initialization, that's why I moved it to the `ngOnInit`

aberikashvili: @dmitry-stepanenko I re-tested it and it's working without `ngOnChanges`

aberikashvili: @dmitry-stepanenko added button for toggling `limitless` input

platon-rov: Yea, changing limitless will be working as we're processing then that change in `searchTermChanged` method. 

What i'm talking about - we're not reinitializing datasource on `dataSource` input change anymore. User has to set it once at the beginning and cannot change it. That's the regression.

platon-rov (changes requested):

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
aberikashvili: @platon-rov ah, got it, will handle this case

aberikashvili: @platon-rov done, please check again

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
```diff
@@ -367,6 +386,12 @@ export abstract class BaseMultiCombobox extends CollectionBaseInput implements A
         super(_cd, ngControl, ngForm, formField, formControl);
     }
 
+    ngOnChanges(changes: SimpleChanges): void {
+        if (changes['dataSource']) {
```
g-cheishvili, line 390: well, you already have input setter on dataSource, you can use that and shrink lines by removing ngOnChanges

g-cheishvili: Or you can remove dataSource setter, if always getting the filled input is not something very important and shrink code that way

g-cheishvili: Either way, in this case, one of them should go in my opinion

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
```diff
@@ -169,6 +173,15 @@ export abstract class BaseMultiCombobox extends CollectionBaseInput implements A
     @Input()
     invalidEntryMessage = 'Invalid entry';
 
+    /** Sets _limitless value */
+    @Input()
+    set limitless(val: boolean) {
```
g-cheishvili, line 178: Setter just sets value to the private property, could be removed and `limitless` can be just an input

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
```diff
@@ -461,7 +486,11 @@ export abstract class BaseMultiCombobox extends CollectionBaseInput implements A
 
         const map = new Map();
         map.set('query', text);
-        map.set('limit', 12);
+
+        if (!this.limitless) {
+            map.set('limit', MAP_LIMIT);
```
g-cheishvili, line 491: It would be better if MAP_LIMIT became configurable by user via DI Token

g-cheishvili (changes requested):
Overall few small notes

platon-rov (commented):
Please check Giorgi's comments.

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
aberikashvili: done

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
aberikashvili: done

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
```diff
@@ -91,14 +97,7 @@ export abstract class BaseMultiCombobox extends CollectionBaseInput implements A
...
     @Input()
-    set dataSource(value: FdpMultiComboboxDataSource<any>) {
-        if (value) {
-            this._initializeDataSource(value);
-        }
-    }
-    get dataSource(): FdpMultiComboboxDataSource<any> {
-        return this._dataSource;
-    }
+    dataSource: any;
```
g-cheishvili, line 100: Why `any`?

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
```diff
@@ -63,6 +67,8 @@ import {
 import { TextAlignment } from '../../combobox';
 import { MultiComboboxConfig } from '../multi-combobox.config';
 
+export const MAP_LIMIT = new InjectionToken<number>('Map limit≥');
```
g-cheishvili, line 70:
```suggestion
export const MAP_LIMIT = new InjectionToken<number>('Map limit≥', { factory: () => 12 });
```

g-cheishvili (changes requested):

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
g-cheishvili: Can you please also add one small example or description of that feature to example?

g-cheishvili (approved):

apps/docs/src/app/platform/component-docs/platform-multi-combobox/platform-multi-combobox-docs.module.ts
```diff
@@ -45,7 +46,10 @@ const routes: Routes = [
         BusyIndicatorModule
     ],
     exports: [RouterModule],
-    providers: [{ provide: MOBILE_MODE_CONFIG, useValue: MULTI_COMBOBOX_MOBILE_CONFIG, multi: true }],
+    providers: [
+        { provide: MOBILE_MODE_CONFIG, useValue: MULTI_COMBOBOX_MOBILE_CONFIG, multi: true },
+        { provide: MAP_LIMIT, useValue: 12 }
```
dmitry-stepanenko, line 51: since you're using the factory as @g-cheishvili suggested, you don't need to provide it anymore

aberikashvili: @dmitry-stepanenko done

libs/platform/src/lib/shared/domain/data-source.ts
```diff
@@ -181,7 +185,7 @@ export class ComboBoxDataSource<T> implements DataSource<T> {
             throw new Error('DataSource.match() predicate can only accepts string and Map');
         }
 
-        if (!searchParam.has('limit')) {
+        if (!searchParam.has('limit') && !this.limitless) {
```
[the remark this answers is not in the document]
platon-rov: I mean, we can keep `@Input() limitless` in component and if it's not provided then use the default limit. But here, in data-source i think just one of them is enough.

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
```diff
@@ -362,11 +365,18 @@ export abstract class BaseMultiCombobox extends CollectionBaseInput implements A
...
-        @Optional() @SkipSelf() @Host() formControl: FormFieldControl<any>
+        @Optional() @SkipSelf() @Host() formControl: FormFieldControl<any>,
+        @Inject(MAP_LIMIT) private _mapLimit: number
     ) {
         super(_cd, ngControl, ngForm, formField, formControl);
     }
 
+    ngOnChanges(changes: SimpleChanges): void {
+        if (changes['dataSource']) {
+            this._initializeDataSource(this.dataSource);
```
platon-rov, line 376: This is still will be different. You have `this._dataSource` changed in `_initializeDataSource` and previously, with getter approach, we were returning `this._datasource` in getter of `this.datasource`. And now `this.datasource` will be returned. Please, just revert this to its oiriginal state.

platon-rov (changes requested):

libs/platform/src/lib/shared/domain/data-source.ts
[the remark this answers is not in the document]
aberikashvili: @platon-rov do you offer to throw away `ComboBoxDataSource.MaxLimit` ? because, if I remove `&& !this.limitless` and consider `!searchParam.has('limit')` means limitless, than this max limit is an obstacle here

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
aberikashvili: @platon-rov which original state? you mean to remove `ngOnChanges` at all and return to setters and getters for datasource?

platon-rov: yea, exactly

libs/platform/src/lib/shared/domain/data-source.ts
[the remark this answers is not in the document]
platon-rov: We have limitless input property in component, we can set it to false by default so we will limit results to certain number using `searchParam.set('limit', ...)`. 

But user will be able to set limitless to true then we won't be setting this map key-value pair. That's  what i meant.

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
aberikashvili: @platon-rov If we return to setter and getter approach, then we'll face the following issue again: 
image

mikerodonnell89 (changes requested):
This breaks the first few examples, clicking the input group addon button to open the dropdown on the first example throws this error:
Screen Shot 2022-05-24 at 2 53 04 PM

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
platon-rov: Okay, i see. Reasonable.

Anyway, we need to partially change this behavior to its previous state. Let's not initialize ds in setter, but we should left those accessors and also set `_dataSource` in `_initializeDataSource` method. 

```js
set dataSource(...
  // do nothing
}
get dataSource(...
  // we wanna persist this one
  return this._dataSource;
}

ngOnChanges(...
  if(changes.dataSource) {
    // in this method we will set this._dataSource as previously
    this._initializeDataSource(changes.dataSource.currentValue);
}
```

aberikashvili: @platon-rov done, please check

aberikashvili:
> This breaks the first few examples, clicking the input group addon button to open the dropdown on the first example throws this error: Screen Shot 2022-05-24 at 2 53 04 PM

@mikerodonnell89 fixed, please check again

dmitry-stepanenko (approved):

PR #7970, opened 2022-04-08
fix(core, platform): dynamic page spacer fix, update for tabs

platon-rov:
## Related Issue(s)

Follow up https://github.com/SAP/fundamental-ngx/pull/7962

## Description

I missed that dynamic page may contain tabs and then there will be several dynamic page content components, so the approach of showing spacer should be different a bit. 

### Before:

Spacer is not shown in stacked tabs example

image

### After:

Spacer is shown for only the last content component when there is stacked tabs

image

### DIFF
--- a/libs/core/src/lib/dynamic-page/dynamic-page.component.ts
+++ b/libs/core/src/lib/dynamic-page/dynamic-page.component.ts
@@ -27,7 +27,7 @@ import { addClassNameToElement, dynamicPageWidthToSize } from './utils';
 import { TabListComponent } from '@fundamental-ngx/core/tabs';
 import { FlexibleColumnLayoutComponent } from '@fundamental-ngx/core/flexible-column-layout';
 
-import { fromEvent, Observable, Subject } from 'rxjs';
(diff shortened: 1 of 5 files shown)


N1XUS (approved):

libs/core/src/lib/dynamic-page/dynamic-page.component.ts
```diff
@@ -317,4 +310,27 @@ export class DynamicPageComponent implements AfterViewInit, OnDestroy, WithDynam
...
         }
     }
+
+    /** @hidden */
+    private _setContentFooterSpacer(): void {
+        this._contentComponent.changes
+            .pipe(
+                startWith(this._contentComponent.toArray()),
+                // to avoid NG0100: Expression has changed after it was checked
+                debounceTime(0),
```
dmitry-stepanenko, line 320: just FYI, it makes more sense to use `observeOn(asyncScheduler)` as we don't need to filter anything (which is the main purpose of debounceTime)

dmitry-stepanenko (approved):

PR #7979, opened 2022-04-12
fix(core): pagination announcement, examples

### LINKED ISSUE #7578: When navigating in pagination a weird text is shown in body
bug

If navigating in pagination a weird text is appended in the `<body>`-tag which says `Page undefined is current page`.

#### Which versions of Angular and Fundamental Library for Angular are affected?
@fundamental-ngx/core : 0.32.5

1. Visit https://sap.github.io/fundamental-ngx/#/core/pagination
2. Open "Source Code" for the Basic Pagination example
3. Click on "StackBlitz"
4. Run the example on "StackBlitz"
5. Click on "Page 2"

Screen Shot 2022-01-20 at 09 47 43

https://user-images.githubusercontent.com/12843264/150304775-09c5a748-8ed1-4c32-adc8-1ca4df360667.mov

platon-rov:
## Related Issue(s)

Closes #7578

## Description

Pagination live announcer fix, example fix. Previously announcer announced wrong page, see console logs on screenshots.

### Before:

image

### After:

image

### DIFF
--- a/libs/core/src/lib/pagination/pagination.component.scss
+++ b/libs/core/src/lib/pagination/pagination.component.scss
@@ -10,3 +10,7 @@ $block: fd-pagination;
         }
     }
 }
+
+.cdk-live-announcer-element {
+    display: none;
+}
--- a/libs/core/src/lib/pagination/pagination.component.scss
+++ b/libs/core/src/lib/pagination/pagination.component.scss
@@ -10,7 +10,3 @@ $block: fd-pagination;
         }
     }
 }
-
-.cdk-live-announcer-element {
-    display: none;
-}
--- a/apps/docs/src/app/core/component-docs/pagination/examples/pagination-example.component.ts
+++ b/apps/docs/src/app/core/component-docs/pagination/examples/pagination-example.component.ts
@@ -1,5 +1,5 @@
 import { Component, ViewChild } from '@angular/core';
-import { HttpClient } from '@angular/common/http';
+import { of } from 'rxjs';
 import { tap, delay } from 'rxjs/operators';
 import { PaginationComponent } from '@fundamental-ngx/core/pagination';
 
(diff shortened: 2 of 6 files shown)


libs/core/src/lib/pagination/pagination.component.scss
```diff
@@ -10,3 +10,7 @@ $block: fd-pagination;
         }
     }
 }
+
+.cdk-live-announcer-element {
```
dmitry-stepanenko, line 14: why do we need this? announcer's elements are not visible

And even if we do, this style will start working once pagination component is loaded, affecting the entire app. We should not have global styles like this

platon-rov: Yea, you right, thanks, fixed it. 

Anyway, it visible in stackblitz examples, so i kept this rule only moved in the different file.

dmitry-stepanenko: @platon-rov I've noticed that now messages won't be announced by screen reader. Maybe we should have something else instead of `display: none`? We have `screenreader-only` style in a few places, maybe we should reuse it and apply here?
```
.screenreader-only {
    position: absolute !important; /* Outside the DOM flow */
    height: 1px;
    width: 1px; /* Nearly collapsed */
    overflow: hidden;
    opacity: 0;
    clip: rect(1px 1px 1px 1px); /* IE 7+ only support clip without commas */
    clip: rect(1px, 1px, 1px, 1px); /* All other browsers */
}
```

platon-rov: Good catch, thank you, please check :)

platon-rov: I've added these styles to the stackblitz styles file as only there we see this live announcer text, in our apps we don't have it.

platon-rov: Also checked in external app, text reads & element isn't visible.

dmitry-stepanenko (approved):

mikerodonnell89 (approved):

PR #7987, opened 2022-04-15
chore(core,platform): styles bump & breaking changes adoption

platon-rov:
## Related Issue(s)

Closes none.

## Description

Fund styles version bump, breaking changes adoption. Unblocks issues that depends on the fund styles fixes.

_Note: Shellbar's buttons color on hover will be fixed automatically with the https://github.com/SAP/fundamental-styles/pull/3328 where the specificity will be changed for the button component._

### Before:

Calendar:
image

Shellbar:
image

### After:

Calendar:
image

Shellbar:
image

_Other components haven't been affected visually_

## BREAKING CHANGE
* Core, Toolbar: Toolbar title now requires `fd-toolbar__title` class
* Core, Tables: Wrap text in table cell in `fd-table-text` directive to have the text-shadow


mikerodonnell89 (changes requested):
I think we want to use an fd-action-sheet rather than an fd-list for the overflow options, as we don't want the separator / dividing line between the items

libs/core/src/lib/calendar/calendar.component.scss
```diff
@@ -1,4 +1,5 @@
 @import '~fundamental-styles/dist/calendar';
+@import '~fundamental-styles/dist/button';
```
N1XUS, line 2: Not sure it's a good idea to load button styles into calendar styles. It make sense only if calendar component doesn't use fd-button component itself. Otherwhise it will create some overrides to the styles

platon-rov: Yea, we're actually applying button classes to the calendar elements. 

It's a good idea to apply only certain button mixins to that elements instead of importing whole component styles, but to be honest it will take one more refactoring iteration for the button component and i'm not sure that we have time for this now. What do you think about refactoring it later after we finished with the initial horizon support phase?

mikerodonnell89 (approved):

N1XUS (approved):

PR #7999, opened 2022-04-18
fix(docs): fixed missing variables file in stackblitz examples

g-cheishvili:
## Related Issue(s)
It was reported from community slack channel

## Description

Added missing delta variables files

### DIFF
--- a/apps/docs/src/app/documentation/core-helpers/stackblitz/stackblitz.service.ts
+++ b/apps/docs/src/app/documentation/core-helpers/stackblitz/stackblitz.service.ts
@@ -102,6 +102,7 @@ export class StackblitzService {
 <html>
     <head>
         <link rel="stylesheet" href="node_modules/@sap-theming/theming-base-content/content/Base/baseLib/sap_fiori_3/css_variables.css" />
+        <link rel="stylesheet" href="node_modules/fundamental-styles/dist/theming/sap_fiori_3.css" />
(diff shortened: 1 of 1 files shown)


22shubham22:
> ## Related Issue(s)
> It was reported from community slack channel
> 
> ## Description
> Added missing delta variables files

Was it just missing in the documentation ? because as soon as i upgraded the fundamentals ngx versions , i lost cards from my project also. 
Here's the screenshot provided -:

MicrosoftTeams-image (20)

g-cheishvili:
> > ## Related Issue(s)
> > It was reported from community slack channel
> > ## Description
> > Added missing delta variables files
> 
> Was it just missing in the documentation ? because as soon as i upgraded the fundamentals ngx versions , i lost cards from my project also. Here's the screenshot provided -:
> 
> MicrosoftTeams-image (20)

Previously border was visible in your case because cards had borders regardless of active theme, but after update border depends on theme you have chosen. So both previously and now, you have to have `node_modules/fundamental-styles/dist/theming/{themeName}.css` in your app loaded

g-cheishvili:
Right now cards comply with both fiori 3 and horizon themes, they have different ways for showing borders. In case of regular themes, it is achieved with box-shadow, in case of high contrast variants borders are used. Becuase of this difference I had to move them to the delta theming variables. So you have to have both `@sap-theming` library and `fundamental-styles` theme variables in your end build

mikerodonnell89 (approved):

droshev (approved):

PR #8005, opened 2022-04-21
fix(core): card horizon breaking changes

g-cheishvili:
## Description

This PR adds footer to the card component and also added feature to card horizon component with list

### DIFF
--- a/libs/core/src/lib/card/card-content.component.ts
+++ b/libs/core/src/lib/card/card-content.component.ts
@@ -19,7 +19,6 @@ export class CardContentComponent implements OnInit, CssClassBuilder {
     /** @hidden */
     @ContentChild(FD_LIST)
     set containsList(list: any) {
-        console.log(list);
         this._containsList = !!list;
     }
 
--- a/libs/core/src/lib/card/card-content.component.ts
+++ b/libs/core/src/lib/card/card-content.component.ts
@@ -1,17 +1,33 @@
-import { Component, OnInit, ChangeDetectionStrategy, ElementRef } from '@angular/core';
+import { Component, OnInit, ChangeDetectionStrategy, ElementRef, ContentChild } from '@angular/core';
 import { applyCssClass } from '@fundamental-ngx/core/utils';
 import { CssClassBuilder } from '@fundamental-ngx/core/utils';
 
 import { CLASS_NAME } from './constants';
+import { FD_CARD_CONTAINER } from './card.tokens';
+import { FD_LIST } from '@fundamental-ngx/core/list';
 
 @Component({
     selector: 'fd-card-content',
     template: '<ng-content></ng-content>',
-    changeDetection: ChangeDetectionStrategy.OnPush
+    changeDetection: ChangeDetectionStrategy.OnPush,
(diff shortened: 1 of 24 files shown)


platon-rov:
Looks like diff calculated between your branch & main, but not the `pr/chore/styles-bump-breaking-changes-adoption`, could you rebase once again?

libs/core/src/lib/card/card-footer.component.html
```diff
@@ -0,0 +1,10 @@
+<ng-content></ng-content>
+<div class="fd-card__footer--start">
```
InnaAtanasova, line 2: Where are these classes coming from? They don't follow BEM. In order to have the modifier `fd-card__footer--start` you need a base class `fd-card__footer` on the same element. For example:
```
<div class="fd-card__footer fd-card__footer--start">...</div>
```

libs/core/src/lib/card/card-content.component.ts
```diff
@@ -1,18 +1,33 @@
...
+    @ContentChild(FD_LIST)
+    set containsList(list: any) {
+        this._containsList = !!list;
+    }
+
+    get containsList(): any {
+        return this._containsList;
+    }
+
+    _containsList: boolean;
```
InnaAtanasova, line 29: missing typedocs

InnaAtanasova: shouldn't this be private?

InnaAtanasova (changes requested):
I'm sorry, you can't have an element with the following classes:
```
class="fd-card__footer fd-card__header fd-card__header--non-interactive..."
```
`fd-card__footer` and `fd-card__header` are two "BLOCKS" from the BEM,  can't be applied together.
Screen Shot 2022-04-25 at 5 54 20 PM

Also, the paddings of the footer are not per specs. Please double check.

libs/core/src/lib/card/card-footer.component.html
g-cheishvili: Those classes are not even necessary, because __footer does `space-between`. I messed up modifier with element

InnaAtanasova (approved):

mikerodonnell89 (approved):

libs/core/src/lib/card/card-footer.component.html
```diff
@@ -0,0 +1,6 @@
+<ng-content></ng-content>
+<div class="fd-card__footer-actions" *ngIf="actionItems?.length > 0">
+    <div class="fd-card__footer-actions-item" *ngFor="let actionItem of actionItems">
```
mikerodonnell89, line 3: Can we do `<ng-content select="fdCardFooterActionItem">`? And apply the `fd-card__footer-actions-item` class in the host of the `CardFooterActionItemDirective` directive. Then we would not need the AfterViewInit in `CardFooter`

mikerodonnell89 (commented):
one question in the code^

libs/core/src/lib/card/card-footer.component.html
g-cheishvili: But then how to decide if actions div should exist or not?

PR #8006, opened 2022-04-21
fix(platform): icon tab bar horizon theme - removed unnecessary more(extra) button directive

g-cheishvili:
## Description

Removed more button directive, because recalculations are not required anymore, button is not absolute


InnaAtanasova:
Looks good, but some visual problem (no spacing for the More button)

Before:
Screen Shot 2022-04-25 at 6 04 15 PM

After:
Screen Shot 2022-04-25 at 6 04 21 PM

g-cheishvili:
> Looks good, but some visual problem (no spacing for the More button)
> 
> Before: Screen Shot 2022-04-25 at 6 04 15 PM
> 
> After: Screen Shot 2022-04-25 at 6 04 21 PM

That's how it should be. Previously position for that button was determined by the `left` CSS rule, it was absolute. By specs it seems it always should be in the end of the bar, so now it is changed to be static positioned + `margin-left(or right in rtl): auto`

InnaAtanasova (approved):

platon-rov (approved):

PR #8010, opened 2022-04-22
(fix platform)  wizardGenerator dots in labels are not supported

MichaelBaySAP:
Currently, dots in labels are not supported as these are interpreted as delimiters of subpaths by angular/forms. Therefore, the dot notation should no longer be used to identify paths in wizard items.

angular/angular#21950

There exists an issue '[WizardGenerator] Dots in labels are not supported in wizard', which describes this with more detailed information in the repository whose name, for confidentiality reasons, I will not post here.

### DIFF
--- a/libs/platform/src/lib/form/form-generator/form-generator.service.ts
+++ b/libs/platform/src/lib/form/form-generator/form-generator.service.ts
@@ -251,18 +251,19 @@ export class FormGeneratorService implements OnDestroy {
      * @returns Found form control.
      */
     getFormControl(form: DynamicFormGroup, controlName: string): DynamicFormGroupControl {
-        let control = form?.get(controlName);
+        let control = form?.get([controlName]);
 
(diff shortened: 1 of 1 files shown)


g-cheishvili:
form.get(path) !== form.get([path])
[Take a look at this](https://github.com/angular/angular/blob/9e6538c4ba6fc4aeba22d70216c689dcf36a73ec/packages/forms/src/model/abstract_model.ts#L1075)

If someone used to use this function with `'some.path.to.field'` he/she will face breaking change. I'm not sure if that breaking change is required and can not be avoided

MichaelBaySAP:
It is not the same but this is also intentional. The split on '.' is what causes the error. As if the controlName aka the label contains a '.' in the name, the function will split on the name interpret it as two different paths when it is not. Therefore, the function accepting an array should be used.

PR #8022, opened 2022-04-25
fix(platform): value help dialog search

### LINKED ISSUE #7908: Search bar behavior is different in for firefox and chrome.

Bug

Search bar behaviour is different in different browsers.

1. Chrome (x appears when we enter some text and on pressing enter searches for the text)
2. Firefox and safari ( x does not appear when we type something and on pressing enter nothing happens )

Chrome 

Screenshot (386)

Firefox & Safari 

Screenshot (387)

Angular 12.2.1
Ngx version : 0.32.7

bug can be noticed while using search bar across different browser .

platon-rov:
## Related Issue(s)

Closes #7908.

## Description

Previously search button was used only to focus search field input, now by clicking on it search is performed by the value in the search field input without any other filter values. Button "Go" still performs search by all of the filters, including search, nothing changed here.

### DIFF
--- a/libs/platform/src/lib/value-help-dialog/value-help-dialog/value-help-dialog.component.ts
+++ b/libs/platform/src/lib/value-help-dialog/value-help-dialog/value-help-dialog.component.ts
@@ -591,7 +591,6 @@ export class PlatformValueHelpDialogComponent<T = any> implements OnChanges, OnD
             const dsSub = this.openDataStream()
                 .pipe(takeUntil(this._destroyed))
                 .subscribe((data) => {
-                    console.log(data);
                     this._displayedData = data.slice();
                     this._changeDetectorRef.markForCheck();
                 });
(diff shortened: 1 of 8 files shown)


mikerodonnell89 (commented):
The linked issue in the description seems only semi-related. Also should the input clear after a search is performed? I don't think that is typical UX behavior

libs/platform/src/lib/value-help-dialog/value-help-dialog/value-help-dialog.component.ts
```diff
@@ -584,23 +584,27 @@ export class PlatformValueHelpDialogComponent<T = any> implements OnChanges, OnD
...
     private _initializeDS(ds: FdpValueHelpDialogDataSource<any> = this.dataSource): void {
         this._resetSourceStream();
+
         if (this.showSelectionTab) {
             this._dsSubscription = new Subscription();
 
             const dsSub = this.openDataStream()
                 .pipe(takeUntil(this._destroyed))
                 .subscribe((data) => {
+                    console.log(data);
```
InnaAtanasova, line 594: Maybe not needed? :)

g-cheishvili (changes requested):
Clicking enter, when you are focused on search input, refreshes grid, but for some reason it does not search for anything and after refresh input is cleaned

dmitry-stepanenko (approved):

g-cheishvili (approved):

g-cheishvili:
@platon-rov please check failing unit tests

mikerodonnell89 (approved):

Issue #7984, opened 2022-04-13
Horizon Adoption/Release Defect Hunting + Bug fixes


InnaAtanasova:
The defect hunting should be done after the fundamental-styles latest version is adopted with its breaking changes and the components have the Horizon theme.
- [ ] Defect Hunting
- [ ] Identify the release blockers
- [ ] Bug fixes

platon-rov:
* [ ] - Core, Dynamic Page inside Column Layout, doubled scrollbars.

<img width="56" alt="image" src="https://user-images.githubusercontent.com/20265336/165604958-3f5e48a2-d645-4f09-a31e-8cb9e93986db.png">

N1XUS:
### Core Card:

- [ ] Not rounded corners with image example: 
- [ ] White strip is visible (the size of the image is smaller than the size of the card)

<img width="417" alt="image" src="https://user-images.githubusercontent.com/6586561/165722515-d77e987d-0f12-4d35-b3f0-95e49c75daa6.png">

### Core Checkbox:

- [ ] Focus outline should have border radius: 
<img width="144" alt="image" src="https://user-images.githubusercontent.com/6586561/165751460-206b27c5-082f-430f-b33d-9ba7dee45c85.png">
<img width="602" alt="image" src="https://user-images.githubusercontent.com/6586561/165751473-2fe18c68-3abd-4c36-b340-f50c529c1a4e.png">

### Core Radio Button:

- [ ] Focus outline should have border radius: 
<img width="154" alt="image" src="https://user-images.githubusercontent.com/6586561/165751804-cc0bad45-c388-40f3-a10e-03c606491f10.png">
<img width="340" alt="image" src="https://user-images.githubusercontent.com/6586561/165751830-9e369cad-ee2f-4cb5-9f39-f86f16bd4947.png">

g-cheishvili:
## Platform Icon tab bar
Steps to reproduce:
* [Navigate here](https://fundamental-ngx.netlify.app/#/platform/icon-tab-bar)
* Resize screen so that "More" button appears
* Select any of the items from "More" menu
* Observe that, that item is now moved to visible items
* Resize to bigger screen, to make "More" hidden
* Then resize back to smaller screen
* Observe that selected item is not now in Visible items, but hidden in "More" menu

Expected Behavior:
Selected tab should always be in visible tab headers list

platon-rov:
* [ ] Calendar, Mobile mode, weird border radius

Landscape:
[...]

PR #8034, opened 2022-04-27
fix(core,platform): missing class on split button

g-cheishvili:
## Related Issue(s)

part of #7984

## Description

Type class should have been applied to the parent class too

### Before:
image

### After:
image


g-cheishvili:
e2e tests, which are not related to this component will be fixed on target branch separately before merging with main

PR #8035, opened 2022-04-28
fix(core): add new Message Box icons

N1XUS:
Updated icons according to this breaking change: https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes#message-box-3257


InnaAtanasova (approved):

libs/core/src/lib/message-box/message-box-semantic-icon/message-box-semantic-icon.component.ts
```diff
@@ -40,15 +35,15 @@ export class MessageBoxSemanticIconComponent {
     get _semanticIcon(): string {
         switch (this.messageBoxConfig.type) {
```
g-cheishvili, line 36: As a side note, we can transfer this switch to object lookup and then we could make it configurable by user

g-cheishvili (approved):

mikerodonnell89 (approved):

Issue #7931, opened 2022-03-30
Horizon themes parameters adaptation from fundamental-styles (part 1)


droshev:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.

List of the components that have some design for Horizon:
- [x] Avatar 
- [x] Busy Indicator
- [x] Checkbox 
- [x] Message Box 
- [x]  Message Toast 
- [x] Message Strip
- [x] Radio Button 
- [x] Switch 
- [x] Icons - Giorgi (No changes)
- [x] Date Picker - @platon-rov had fixed this already
- [x] Dialog - Giorgi (No changes)
- [x] Tab Bar - Giorgi - https://github.com/SAP/fundamental-ngx/pull/8006
- [x] Cards - Giorgi - https://github.com/SAP/fundamental-ngx/pull/8005
- [x] Tiles - Giorgi - There is no dedicated section in wiki for tiles. As per css variables, current styles implementation complies with Semantic Color Parameters. On NGX side, it does not need any involvement, since it had not changed
- [x] Popover - Mike https://github.com/SAP/fundamental-ngx/pull/8055
- [x] Toolbar and Bar - Mike https://github.com/SAP/fundamental-ngx/pull/8051
- [x] List View - Mike (no changes necessary)
- [x] Text Components - Mike (no changes)
- [x] Home - Mike (no changes)
- [x] Page and Layout - Denis https://github.com/SAP/fundamental-ngx/pull/8052
- [x] Shellbar - Denis 

Components dependant on future work in styles
- [x] Badge on Button
- [x] Button 
- [x] Button Shortcuts - no such feature
- [x] Menu Button
- [x] Segmented Button
- [x] Toggle Button 
- [x] Input Components - Combobox 
- [x] Input Components - Input Field 
    - [x] Step Input 
    - [x] Input Group 
- [x] Input Components - Search Field 
- [x] Input Components - Select 
- [x] Input Components - Text Area 
- [x] Input Components - Token 
- [x] Input Components - Tokenizer

PR #8051, opened 2022-05-02
fix(core): toolbar bar horizon updates

mikerodonnell89:
part of #7931 

Adds the option to simply pass a title via string input, or the developer can use the existing method of passing via ng-content


g-cheishvili:
So, `When the text inside a Toolbar is set to h4, the [hasTitle]="true" must be used as seen in the exmaple below.`, meaning hasTitle should always be applied or only in case of h4? Can't `ToolbarComponent` check it by itself if it contains fd-title?

N1XUS (approved):

mikerodonnell89:
> So, `When the text inside a Toolbar is set to h4, the [hasTitle]="true" must be used as seen in the exmaple below.`, meaning hasTitle should always be applied or only in case of h4? Can't `ToolbarComponent` check it by itself if it contains fd-title?

@g-cheishvili thanks for pointing this out, I didn't look through the docs

mikerodonnell89:
> So, `When the text inside a Toolbar is set to h4, the [hasTitle]="true" must be used as seen in the exmaple below.`, meaning hasTitle should always be applied or only in case of h4? Can't `ToolbarComponent` check it by itself if it contains fd-title?

I think the docs are implying that the dev should use h4 in the event that they'd like to use a title. In my opinion this could simply be a string input - I have committed a change for this, wondering if anyone has thoughts @InnaAtanasova @g-cheishvili @N1XUS

g-cheishvili:
@mikerodonnell89 that will limit capabilities. what if user wants to have icon in header or have inline help, or anythin. if its input funtionality is limited

mikerodonnell89:
> @mikerodonnell89 that will limit capabilities. what if user wants to have icon in header or have inline help, or anythin. if its input, funtionality is limited

@g-cheishvili The previous docs and design specs for both horizon and fiori 3 imply to me that the toolbar is to have a text-only title. But I suppose going this route might be too opinionated, which goes against the idea of fundamental-ngx/core. Thoughts @InnaAtanasova ?

mikerodonnell89:
@InnaAtanasova has suggested allowing both a simple input or allowing more complex templates via ng-content so I will do that and document both

libs/core/src/lib/toolbar/toolbar.component.ts
```diff
@@ -118,6 +125,10 @@ export class ToolbarComponent implements OnInit, AfterViewInit, OnDestroy, After
     @ContentChildren(forwardRef(() => ToolbarItemDirective))
     toolbarItems: QueryList<ToolbarItemDirective>;
 
+    /** @hidden */
+    @ContentChild(TitleComponent)
+    titleComponent: TitleComponent;
```
platon-rov, line 130: Querying like this prevent from tree-shaking `TitleComponent` when compiling this component.

Here some more about it https://angular.io/guide/lightweight-injection-tokens

libs/core/src/lib/toolbar/toolbar.component.ts
```diff
@@ -206,6 +217,14 @@ export class ToolbarComponent implements OnInit, AfterViewInit, OnDestroy, After
         this.buildComponentCssClass();
     }
 
+    /** @hidden */
+    ngAfterContentInit(): void {
+        if (this.titleComponent) {
+            this.titleComponent.elementRef?.nativeElement.classList.add('fd-toolbar__title');
```
platon-rov, line 223: I think it should be done in `TitleComponent` itself. 

There we may get toolbar as an optional dependency and check if it's present - apply class. Also, this dependency should rely not on the implementation but on the injection token, see my comment above.

mikerodonnell89: It's a toolbar class so in my opinion makes more sense to happen here

libs/core/src/lib/toolbar/toolbar.component.ts
mikerodonnell89: done

platon-rov (approved):

InnaAtanasova (approved):

InnaAtanasova:
I noticed the new components that go in the overflow menu don't have a margin between them:
Screen Shot 2022-05-20 at 2 00 17 PM

But I guess it's out of the scope for this PR and will go to the defect hunting issue (or separately).

InnaAtanasova:
> I noticed the new components that go in the overflow menu don't have a margin between them: Screen Shot 2022-05-20 at 2 00 17 PM
> 
> But I guess it's out of the scope for this PR and will go to the defect hunting issue (or separately).

Linked the issue here: https://github.com/SAP/fundamental-ngx/issues/8119

PR #8052, opened 2022-05-03
fix(core,platform): adjust toolbar component for dynamic page

N1XUS:
Updated toolbar component to reflect correct class for adjusting the background color if it's used inside dynamic page header


libs/core/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.ts
```diff
@@ -33,9 +36,15 @@ export const ActionSquashBreakpointPx = 1280;
     encapsulation: ViewEncapsulation.None,
     host: {
         '[attr.tabindex]': '0'
-    }
+    },
+    providers: [
+        {
+            provide: DYNAMIC_PAGE_HEADER_TOKEN,
+            useExisting: forwardRef(() => DynamicPageHeaderComponent)
```
dmitry-stepanenko, line 43: just FYI, forwardRef is not required when referring the component directly in the director

g-cheishvili (approved):

dmitry-stepanenko (approved):

PR #8055, opened 2022-05-03
fix(core): update popover docs to not use fd-list

mikerodonnell89:
part of #7984

Popover bodies should not have been containing fd-lists, this PR only changes the docs. No breaking changes to popover from fundamental-styles

### DIFF
--- a/apps/docs/src/app/core/component-docs/popover/popover-docs.module.ts
+++ b/apps/docs/src/app/core/component-docs/popover/popover-docs.module.ts
@@ -33,8 +33,6 @@ import { PopoverFocusExampleComponent } from './examples/popover-focus-example/p
 import { PopoverMobileExampleComponent } from './examples/popover-mobile/popover-mobile-example.component';
 import { PopoverTriggerExampleComponent } from './examples/popover-trigger-example/popover-trigger-example.component';
 import { PopoverDynamicContainerHeightExampleComponent } from './examples/popover-dynamic-container-height/popover-dynamic-container-height-example.component';
-import { BusyIndicatorModule } from '@fundamental-ngx/core/busy-indicator';
-import { ListModule } from '@fundamental-ngx/core/list';
 
 const routes: Routes = [
     {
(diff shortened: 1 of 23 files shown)


apps/docs/src/app/core/component-docs/popover/popover-docs.module.ts
```diff
@@ -34,6 +33,8 @@ import { PopoverFocusExampleComponent } from './examples/popover-focus-example/p
 import { PopoverMobileExampleComponent } from './examples/popover-mobile/popover-mobile-example.component';
 import { PopoverTriggerExampleComponent } from './examples/popover-trigger-example/popover-trigger-example.component';
 import { PopoverDynamicContainerHeightExampleComponent } from './examples/popover-dynamic-container-height/popover-dynamic-container-height-example.component';
+import { BusyIndicatorModule } from '@fundamental-ngx/core/busy-indicator';
+import { ListModule } from '@fundamental-ngx/core/list';
```
platon-rov, line 37: It shouldn't be needed anymore.

apps/docs/src/app/core/component-docs/popover/popover-docs.module.ts
```diff
@@ -34,6 +33,8 @@ import { PopoverFocusExampleComponent } from './examples/popover-focus-example/p
 import { PopoverMobileExampleComponent } from './examples/popover-mobile/popover-mobile-example.component';
 import { PopoverTriggerExampleComponent } from './examples/popover-trigger-example/popover-trigger-example.component';
 import { PopoverDynamicContainerHeightExampleComponent } from './examples/popover-dynamic-container-height/popover-dynamic-container-height-example.component';
+import { BusyIndicatorModule } from '@fundamental-ngx/core/busy-indicator';
```
platon-rov, line 36: Also don't see where it's used

mikerodonnell89:
@platon-rov fixed, thanks!

PR #8058, opened 2022-05-04
fix(docs): card examples

g-cheishvili:
## Related Issue(s)

part of #7984 

## Description

Fixed card image example. Updating fundamental styles version so that after that cards do not have overflown content

### Before:
analytical_before
calendar_before
image_before

### After:
analytical_afterjpg
calendar_after
image_after

## Blocked
https://github.com/SAP/fundamental-styles/pull/3379

### DIFF
--- a/apps/docs/src/app/core/component-docs/card/card-docs.component.ts
+++ b/apps/docs/src/app/core/component-docs/card/card-docs.component.ts
@@ -3,6 +3,8 @@ import { Component } from '@angular/core';
 import { ExampleFile } from '../../../documentation/core-helpers/code-example/example-file';
 
 import cardExampleHtml from '!./examples/card-example.component.html?raw';
+import cardExampleTs from '!./examples/card-example.component.ts?raw';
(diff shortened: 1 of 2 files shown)


N1XUS:
@g-cheishvili I would disagree on the default overflow behaviour. The design clearly states that the borders of the card should be rounded, and thus, the content inside it too. If user would want to have `overflow-x: scroll`, he'd still need to change the styles no matter if we provide default overflow behaviour or not.
In short words: I think we should have default `overflow: hidden` behaviour added to the card content component.

g-cheishvili:
> @g-cheishvili I would disagree on the default overflow behaviour. The design clearly states that the borders of the card should be rounded, and thus, the content inside it too. If user would want to have `overflow-x: scroll`, he'd still need to change the styles no matter if we provide default overflow behaviour or not. In short words: I think we should have default `overflow: hidden` behaviour added to the card content component.

Valid point. Take a look at this then https://github.com/SAP/fundamental-styles/pull/3379

InnaAtanasova (approved):

PR #8066, opened 2022-05-06
fix(docs): use injection tokens instead of plain strings

dmitry-stepanenko:
Should not be using plain strings for injection tokens


g-cheishvili (approved):
I like it. Good Job!

PR #8103, opened 2022-05-17
test: use soft page refresh in e2e tests

dmitry-stepanenko:
## Related Issue(s)

closes none

## Description
Switched to using in-app redirects instead of full page reload. This reduced runtime of e2e test suites by ~40%.

Also added 90s timeouts for test cases and 100m timeout for entire test suits

### Before:
image

### After:
image


e2e/wdio/core/pages/multi-input.po.ts
```diff
@@ -16,12 +16,12 @@ export class MultiInputPo extends CoreBaseComponentPo {
     approveButton = '.fd-button--emphasized';
     multiSelectButton = '.custom-multi-input-select-all-bar-element button';
     mobileInputOptions = 'fd-multi-input-mobile-example .fd-token__text span';
-    displayObjectOptions = 'div#background-ex3 .fd-token__text span';
-    searchTermOptions = 'div#background-ex4 .fd-token__text span';
-    customFilterOptions = 'div#background-ex5 .fd-token__text span';
+    displayObjectOptions = 'fd-multi-input-displaywith-example .fd-token__text span';
```
InnaAtanasova, line 19: missed a `-` in `displaywith`

InnaAtanasova (approved):

e2e/wdio/core/pages/multi-input.po.ts
dmitry-stepanenko: @InnaAtanasova it's the actual selector of the component. Do you want me to rename it?
image

InnaAtanasova: no, it's ok. Leave it as it is. Thanks for replying.

e2e/wdio/driver/wdio.ts
```diff
@@ -61,7 +61,21 @@ export function goBack(): void {
...
+export function refreshPage(isFullRefresh = false): void {
+    try {
+        browser.dismissAlert();
+    } catch {}
+    if (!isFullRefresh) {
+        const url = browser.getUrl();
+        click(`#toolbar-home-btn`);
+        if (browser.getUrl().includes(`/home`)) {
+            goBack();
+            return;
```
platon-rov, line 73: If we're in this branch we're omitting `if (browserIsSafari())`, is it intended?

e2e/wdio/driver/wdio.ts
```diff
@@ -61,7 +61,21 @@ export function goBack(): void {
...
+        } else {
+            // failed to navigate, reset the url and perform full page refresh
+            browser.url(url);
```
platon-rov, line 76: If we're in this branch we will be refreshing page 2 times. Here & `browser.refresh();` later.

platon-rov (commented):
good job!

e2e/wdio/driver/wdio.ts
dmitry-stepanenko: Nice catch. I was just handling a weird edge cases if any and didn't notice `browser.url` performs normal page redirects. Fixed

e2e/wdio/driver/wdio.ts
dmitry-stepanenko: I'm wondering why they actually added that extra pause for safari, probably it was a bug somewhere, it works fine for me even without it. But it definitely not needed, when we're basically clicking urls on the page, so yes, it's intended.

platon-rov (approved):

PR #8106, opened 2022-05-17
fix(platform) wizardgenerator dots in labels are not supported

MichaelBaySAP:


libs/platform/src/lib/form/form-generator/form-generator.service.ts
```diff
@@ -255,13 +255,13 @@ export class FormGeneratorService implements OnDestroy {
      * @returns Found form control.
      */
     getFormControl(form: DynamicFormGroup, controlName: string): DynamicFormGroupControl {
```
N1XUS, line 257:
```suggestion
    getFormControl(form: DynamicFormGroup, ...controlNames: string[]): DynamicFormGroupControl {
        let control = form?.get(controlNames);

        // If no control found, try to find it in ungrouped group
        if (!control) {
            control = form?.get([UNGROUPED_FORM_GROUP_NAME, ...controlNames]);
        }

        return control as DynamicFormGroupControl;
    }
```

MichaelBaySAP: Is this actually working, I understood that usually the input to getFormControl is a single element?

MichaelBaySAP: I think, to correctly implement this change, many more places in the code need to be changed. I understand that for cleanliness, this is required. However, I do not know how to locally test the code. I only tested the change in the browser. I believe that the change I proposed would work. If it does not, it would be great if you could open another PR/push the changes required.

PR #8120, opened 2022-05-20
fix(docs): reset component search results when X clicked

mikerodonnell89:
closes none

### DIFF
--- a/apps/docs/src/app/documentation/core-helpers/sections-toolbar/sections-toolbar.component.html
+++ b/apps/docs/src/app/documentation/core-helpers/sections-toolbar/sections-toolbar.component.html
@@ -9,7 +9,7 @@
             [disabled]="false"
             [ngModel]="search"
             (ngModelChange)="onSearchChange($event)"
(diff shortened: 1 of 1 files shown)


apps/docs/src/app/documentation/core-helpers/sections-toolbar/sections-toolbar.component.html
```diff
@@ -9,7 +9,7 @@
             [disabled]="false"
             [ngModel]="search"
             (ngModelChange)="onSearchChange($event)"
-            (addOnButtonClicked)="search = ''"
+            (addOnButtonClicked)="search = ''; onSearchChange('')"
```
dmitry-stepanenko, line 12: I guess it should just work without direct assignment to `search` 
```suggestion
            (addOnButtonClicked)="onSearchChange('')"
```

g-cheishvili (changes requested):
Just like Dmitry noted

g-cheishvili (approved):

PR #8133, opened 2022-05-23
fix(platform): allow dots in control name

N1XUS:
closes

## Description
Currently, dots in labels are not supported as these are interpreted as delimiters of subpaths by angular/forms. Therefore, the dot notation should no longer be used to identify paths in wizard items.

https://github.com/angular/angular/issues/21950

There exists an issue '[WizardGenerator] Dots in labels are not supported in wizard', which describes this with more detailed information in the repository whose name, for confidentiality reasons, I will not post here.
##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [x] poor examples

4. Accessibility testing

##### PR Quality

-   [ ] update `README.md`


g-cheishvili (approved):

libs/platform/src/lib/form/form-generator/form-generator.service.ts
```diff
@@ -51,8 +53,14 @@ export class FormGeneratorService implements OnDestroy {
...
         @Inject(FORM_GENERATOR_ITEM_CONFIG)
-        private _defaultConfig: Partial<DynamicFormFieldItem>
-    ) {}
+        private _defaultItemConfig: Partial<DynamicFormFieldItem>,
+        @Optional()
+        @SkipSelf()
+        @Inject(FORM_GENERATOR_CONFIG)
+        private _providedConfig: Partial<FormGeneratorConfig>
+    ) {
+        this._config = Object.assign({ ...this._config }, { ...this._providedConfig });
```
dmitry-stepanenko, line 62: Some inner props may be omitted, when config is created using `Object.assign`. Maybe it would be better to use lodash's merge, which is recursive? Also please note that config can be null, right?

N1XUS: @dmitry-stepanenko config cannot be null since we use default provider in form generator module, and override it with `withConfig` method, which will not accept null, but can accept empty object which shouldn't break anything

dmitry-stepanenko (approved):

PR #8136, opened 2022-05-24
fix(core): layout grid issues

platon-rov:
## Related Issue(s)

Part of https://github.com/SAP/fundamental-ngx/issues/7984#issuecomment-1132079100

## Description

Layout Grid component fixes, nested rows issue fix.

### Before:

image

### After:

image

### DIFF
--- a/apps/docs/src/app/core/component-docs/layout-grid/examples/layout-grid-basic-example.component.html
+++ b/apps/docs/src/app/core/component-docs/layout-grid/examples/layout-grid-basic-example.component.html
@@ -1,25 +1,27 @@
 <fd-layout-grid>
-    <div [fdLayoutGridCol]="2">
-        <example-layout-grid-block>Size: 2</example-layout-grid-block>
-    </div>
+    <div fdLayoutGridRow>
+        <div [fdLayoutGridCol]="2">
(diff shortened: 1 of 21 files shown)


apps/docs/src/app/core/component-docs/layout-grid/examples/layout-grid-basic-example.component.html
```diff
@@ -1,25 +1,27 @@
 <fd-layout-grid>
-    <div [fdLayoutGridCol]="2">
-        <example-layout-grid-block>Size: 2</example-layout-grid-block>
-    </div>
+    <div fdLayoutGridRow>
```
N1XUS, line 2: Can we add check in fd-layout-grid component that if there's no fdLayoutGridRow directive, we should wrap everything? Just to be backward-compatible with older versions?

N1XUS (approved):

apps/docs/src/app/core/component-docs/layout-grid/examples/layout-grid-basic-example.component.html
platon-rov: Good catch, breaking change removed.

InnaAtanasova (approved):

Issue #7765, opened 2022-02-16
Smart Filter Bar - content density support


droshev:
- [ ] smart filter bar fields are not responding to Content Density
<img width="1316" alt="Screen Shot 2022-02-14 at 22 22 04" src="https://user-images.githubusercontent.com/4380815/153986592-7b2a0862-f81c-4a42-9f96-eeee93622afe.png">
<img width="1322" alt="Screen Shot 2022-02-14 at 22 21 54" src="https://user-images.githubusercontent.com/4380815/153986594-7f04d1e2-b6a1-45f9-b73d-7a99fd90f837.png">

droshev:
@platon-rov #8230 has been merged

PR #8146, opened 2022-05-27
feat(core,platform): local content density, smart filter bar density support

platon-rov:
## Related Issue(s)

Part of https://github.com/SAP/fundamental-ngx/issues/7765

## Description

This pr introduces local (provided on the sub-package level) content density service that handles global content density changes and shares content density value with child components. Explicit subscription to the global content density service is not needed anymore. Local content density service to be provided on the root component of the sub-package and content density value should be consumed using `listenDensity` pipe in the template. 

Local content density service set up for the Smart Filter Bar Component.

### Before:

image

### After:

image


InnaAtanasova (changes requested):
In Cozy mode the the inputs (Description and status) they stay in compact mode

Screen Shot 2022-05-27 at 1 24 28 PM
Screen Shot 2022-05-27 at 1 24 21 PM

Issue #5574, opened 2021-06-04
Accessibility:  Approval Flow-Platform Component


Sanjay-1312:
Accessibility: Approval Flow Component (There are couple of accessibility issues found while testing, so i have attached error document for you reference.)
Screenshot:
[Approval flow_Platform_ARIBA Controls.docx](https://github.com/SAP/fundamental-ngx/files/6654147/Approval.flow_Platform_ARIBA.Controls.docx)

Sanjay-1312:
<meta name="ProgId" content="Word.Document">
<meta name="Generator" content="Microsoft Word 15">
<meta name="Originator" content="Microsoft Word 15">
<link rel="File-List" href="file:///C:/Users/C5310468/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

<link rel="themeData" href="file:///C:/Users/C5310468/AppData/Local/Temp/msohtmlclip1/01/clip_themedata.thmx">
<link rel="colorSchemeMapping" href="file:///C:/Users/C5310468/AppData/Local/Temp/msohtmlclip1/01/clip_colorschememapping.xml">

<style>

</style>






<h2 style="mso-collapsed-heading:yes"><span lang="EN-US">Issue 1<o:p></o:p></span></h2>

<h3><span lang="EN-US"><o:p>&nbsp;</o:p></span></h3>


Priority: | High
-- | --
Checkpoint: | ACC-264



<p class="MsoNormal" style="margin-left:18.0pt"><span lang="EN-US" style="color:white;mso-color-alt:windowtext">Navigate to the names available
within watchers</span><span lang="EN-US"><o:p></o:p></span></p>

<h3><span lang="EN-US">Observed Behavior:<o:p></o:p></span></h3>

<p class="MsoNormal"><span lang="EN-US">On navigating to the names within
‘watchers’ header, the information ‘watcher’ is not read. <o:p></o:p></span></p>

<p class="MsoNormal"><span lang="EN-US">Also, the information regarding the element
type is not read. i.e. it does not read the information that the element is
clickable.<o:p></o:p></span></p>

<p class="MsoNormal"><span lang="EN-US">The speech output is:<o:p></o:p></span></p>

<p class="MsoNormal"><span lang="EN-US">Sofia Hanson Graphic‑ <br>
Melissa Martin Graphic‑ <br>
Julie Peters Graphic‑ <br>
Fred Gibson Graphic‑ <br>
[...]

PR #8160, opened 2022-05-31
fix(core): assign labels as aria-labelledby for corresponding controls inside fd-form-item

dmitry-stepanenko:
## Related Issue(s)

closes none

## Description
Adding logic, that ties labels with controls inside `fd-form-item` block
 ```html
<div fd-form-item>
  <label fd-form-label for="input-id">Label Text</label>
  <fd-select></fd-select> <= should have "aria-labelledby" property pointing to the corresponding label
</div>
  ```

##### During Implementation

1. Visual Testing:

-   [n/a] States - hover/disabled/focused/active/on click/selected/selected hover/press state
-   [n/a] Interaction/Animation - open/close, expand/collapse, add/remove, check/uncheck
-   [n/a] Mouse vs. Keyboard support
-   [n/a] Text Truncation

-   [x] API boundary values

-   [n/a] poor examples

4. Accessibility testing

##### PR Quality

-   [n/a] update `README.md`

### DIFF
--- a/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
+++ b/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
@@ -23,6 +23,7 @@ import { LIST_ITEM_COMPONENT, ListItemInterface, ContentDensityService } from '@
 import equal from 'fast-deep-equal';
 import { Subscription } from 'rxjs';
 import { FormStates, Nullable } from '@fundamental-ngx/core/shared';
+import { registerFormItemControl, FormItemControl } from '@fundamental-ngx/core/form';
 import { FD_CHECKBOX_VALUES_DEFAULT } from './fd-checkbox-values.interface';
 
 let checkboxUniqueId = 0;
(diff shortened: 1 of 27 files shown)


InnaAtanasova:
Screen Shot 2022-06-01 at 1 42 43 PM

Hi @dmitry-stepanenko , could you briefly explain what you fixed with this PR? What is the issue you are trying to fix? Why do every list item need `aria-describedby`? I see an empty hidden span added. What's it's purpose when there's no label or content? Lastly, since this is Core, could you use IDs that start with `fd-`?

dmitry-stepanenko:
> Screen Shot 2022-06-01 at 1 42 43 PM
> 
> Hi @dmitry-stepanenko , could you briefly explain what you fixed with this PR? What is the issue you are trying to fix? Why do every list item need `aria-describedby`? I see an empty hidden span added. What's it's purpose when there's no label or content? Lastly, since this is Core, could you use IDs that start with `fd-`?

Hi @InnaAtanasova. Your points are not related to this PR, because I'm not touching list options, only tying actual inputs with their labels for screen readers. This was raised as part of #5574 (issue №8 there), but I identified this as a common problem across the UI library.
image


But you're right, it's not really reasonable to show empty element there, added `*ngIf`. And changed id's prefix to be `fd` instead

dmitry-stepanenko:
> Should we clean up in the docs all the components that have the ariaLabelledBy set manually where it doesn't match with anything? For example here: https://github.com/SAP/fundamental-ngx/blob/main/apps/docs/src/app/core/component-docs/combobox/examples/combobox-example.component.html#L7

@mikerodonnell89 good point, done


Had to create a new PR as this one cannot be reopened https://github.com/SAP/fundamental-ngx/pull/8223

Issue #7930, opened 2022-03-30
avatar with initials in Table 


droshev:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.

![image](https://user-images.githubusercontent.com/4380815/160844800-0a04fb81-6e2f-408f-9ea4-6b72db714961.png)

If using the fundamentals avatar inside a fd table the style changes:
image
![image](https://user-images.githubusercontent.com/4380815/160844869-1180f92e-0258-4b1c-88e8-e855828c160f.png)

![image](https://user-images.githubusercontent.com/4380815/160844900-953e001c-1a8c-4081-9ded-943aa9a47f52.png)

there is text-shadow, comming from the table-cell style
if there are more than 2 letters they are shown on two rows
image

image

/dxp/ngx-dxp/issues/84

tjbutz:
> Hi, could you please ping me when this is fixed, i will need to clean up a workaround in dxp frame side, thanks :)

@droshev any update on this one?

platon-rov:
@lpgarzonr 

fix is available in version `0.35.0-rc.11` and consists of two parts

* If you're using core table - from now you have to apply `fd-table-text` to the cell content to have this glow (shadow), otherwise there won't be any shadow as it was previously.

* if you're using platform table - from now you have to set `applyText` input property to false to the `fdp-column` you don't want to have shadow applied to (like those ones with avatars).

Generally this shadow is visible in High Contrast themes when row is selected, but it may cause glitches like those ones with the avatars, so now there is a way to not have a shadow in some cells\columns.

<img width="667" alt="image" src="https://user-images.githubusercontent.com/20265336/173336120-3269b64e-455f-45a9-bef5-9a06ef7adc43.png">

lpgarzonr:
@platon-rov great! I am gonna upgrade and fix our workaround :)

PR #8192, opened 2022-06-07
feat(core): ability to optionally apply table-text directive in table

platon-rov:
## Related Issue(s)

Closes https://github.com/SAP/fundamental-ngx/issues/7930

## Description

Ability to optionally apply table-text directive in platform table.

### Before:
image

### After:
image


libs/core/src/lib/table/directives/table-text.directive.ts
```diff
@@ -6,6 +6,7 @@ import { Directive, HostBinding, Input } from '@angular/core';
 export class TableTextDirective {
     /** @hidden */
     @HostBinding('class.fd-table__text')
+    @Input()
```
N1XUS, line 9: Please remove `@hidden` and add a description.

InnaAtanasova (approved):

PR #8196, opened 2022-06-08
feat(core): breadcrumbs updates

g-cheishvili:
## Description

Adopts breaking change from fundamental-ngx. added deprecation of fd-breadcrumb-link. Changed the way colapses are made. Also added feature, now icons are attached correctly to the menu as prefix and postfix. Now changes on content of link are passed to collapsed menu item


apps/docs/src/app/core/component-docs/breadcrumb/examples/breadcrumb-routerLink-example.component.html
```diff
@@ -1,12 +1,12 @@
 <fd-breadcrumb>
     <fd-breadcrumb-item>
-        <a fd-breadcrumb-link routerLink="../../avatar">Breadcrumb Level 1</a>
+        <a fd-link routerLink="../../avatar"> Breadcrumb Level <span style="font-weight: bold">1</span> </a>
```
mikerodonnell89, line 3: Why bold here?

libs/core/src/lib/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -1,131 +1,129 @@
...
-            <fd-breadcrumb-item>
-                <span>Breadcrumb Level 3</span>
-            </fd-breadcrumb-item>
-        </fd-breadcrumb>
-    `
-})
-class BreadcrumbWrapperComponent {
-    @ViewChild(BreadcrumbComponent) breadcrumb: BreadcrumbComponent;
-}
+// import { ComponentFixture } from '@angular/core/testing';
```
mikerodonnell89, line 1: Can we uncomment this test?

apps/docs/src/app/core/component-docs/breadcrumb/examples/breadcrumb-routerLink-example.component.html
g-cheishvili: Just to showcase that when it transforms to the menu, html is saved, previously it was not, it would just take inner text and that's it

libs/core/src/lib/breadcrumb/breadcrumb.component.spec.ts
g-cheishvili: Not yet

mikerodonnell89 (changes requested):
When I first click the ellipses to open the overflow popover, the popover is opened in the top left corner of the screen:
Screen Shot 2022-06-08 at 9 41 27 AM

libs/core/src/lib/breadcrumb/breadcrumb-hidden-item/breadcrumb-hidden-item.component.ts
```diff
@@ -0,0 +1,33 @@
...
+import { BreadcrumbItemComponent } from '../breadcrumb-item.component';
+import { DomPortal } from '@angular/cdk/portal';
+
+@Component({
+    selector: 'fd-breadcrumb-hidden-item',
+    templateUrl: './breadcrumb-hidden-item.component.html'
+})
+export class BreadcrumbHiddenItemComponent implements AfterViewInit {
+    @Input()
+    breadcrumbItem: BreadcrumbItemComponent;
```
platon-rov, line 11: typedocs.

libs/platform/src/lib/upload-collection/upload-collection/upload-collection.component.html
```diff
@@ -14,13 +14,13 @@
 
     <!-- Toolbar -->
     <fd-toolbar class="fdp-upload-collection__toolbar">
-        <div #responsiveBreadcrumbContainer>
+        <div #responsiveBreadcrumbContainer style="flex-grow: 1">
```
platon-rov, line 17: is setting styles in html the best approach we could use here?

libs/core/src/lib/utils/services/resize-observer.service.ts
```diff
@@ -0,0 +1,84 @@
+import { ElementRef, Injectable, OnDestroy } from '@angular/core';
```
platon-rov, line 1: is that service replacement for the `resize-observable.ts`? then we should mark that approach as deprecated. 

sorry, but what reason to have this service? i mean what not okay with previous approach?

libs/core/src/lib/link/link.component.ts
```diff
@@ -67,12 +85,20 @@ export class LinkComponent implements OnChanges, OnInit, CssClassBuilder, AfterV
     /** @hidden */
     private _destroyed$ = new Subject<void>();
 
+    prefixIconName: string;
```
platon-rov, line 88: typedocs.

g-cheishvili:
> When I first click the ellipses to open the overflow popover, the popover is opened in the top left corner of the screen: Screen Shot 2022-06-08 at 9 41 27 AM

That issue exists in app. I fixed it now

mikerodonnell89 (approved):

libs/core/src/lib/utils/services/resize-observer.service.ts
```diff
@@ -0,0 +1,84 @@
...
+    create(callback: ResizeObserverCallback): ResizeObserver | null {
+        return typeof ResizeObserver === 'undefined' ? null : new ResizeObserver(callback);
+    }
+}
+
+@Injectable({ providedIn: 'root' })
+export class ResizeObserverService implements OnDestroy {
+    private _observedElements = new Map<Element, ObservedElement>();
+
+    constructor(private _resizeObserverFactory: ResizeObserverFactory) {}
```
N1XUS, line 22: typedocs

N1XUS (approved):

InnaAtanasova (approved):

PR #8209, opened 2022-06-10
fix(core): truncate breadcrumb text when item text is longer than 30 chars

### LINKED ISSUE #7840: Truncate the title in fd-breadcrum if it's too long
enhancement
Ideally if the title is more than specific length, we should truncate and show title-truncated...
Screen Shot 2022-01-25 at 11 47 36 AM
Screen Shot 2022-01-25 at 11 50 18 AM

<fd-breadcrum>

shrvr:
## Related Issue(s)

closes #7840 

## Description

Truncates the breadcrumb text if item text is longer than 30 characters.

##### During Implementation

1. Visual Testing:

-   [ ] check Light/Dark/HCB/HCW themes
-   [ ] RTL/LTR - proper rendering and labeling
-   [ ] Content Density (Cozy/Compact/(Condensed))
-   [ ] States - hover/disabled/focused/active/on click/selected/selected hover/press state
-   [ ] Interaction/Animation - open/close, expand/collapse, add/remove, check/uncheck
-   [ ] Mouse vs. Keyboard support
-   [x] Text Truncation

-   [ ] check for console logs (warnings, errors)
-   [ ] API boundary values
-   [ ] different combinations of components - free style
-   [ ] change the API values during testing

-   [ ] poor examples
-   [ ] Stackblitz works for all examples

4. Accessibility testing

##### PR Quality

-   [ ] update `README.md`

### DIFF
--- a/libs/core/src/lib/breadcrumb/breadcrumb-item.directive.ts
+++ b/libs/core/src/lib/breadcrumb/breadcrumb-item.directive.ts
@@ -38,11 +38,25 @@ export class BreadcrumbItemDirective implements AfterContentInit {
 
     /** @hidden */
     ngAfterContentInit(): void {
+        const childElements = this._elementRef.nativeElement.getElementsByTagName('*');
+        for (const child of childElements) {
+            child.innerText = this._truncateBreadcrumbText(child.innerText);
+        }
         if (this.breadcrumbLink && this.breadcrumbLink.elementRef.nativeElement.getAttribute('href')) {
             this.href = this.breadcrumbLink.elementRef.nativeElement.getAttribute('href');
         }
         if (this.breadcrumbLink && this.breadcrumbLink.routerLink) {
             this.routerLink = this.breadcrumbLink.routerLink;
         }
+    }
+    /**
+     * @hidden
+     *  it truncates breadcrumb text when item text is longer than 30 chars
+     */
+    private _truncateBreadcrumbText(text: string): string {
+        if (text.length > 30) {
+            text = text.substring(0, 29) + '...';
+        }
+        return text;
     }
 }
--- a/libs/core/src/lib/breadcrumb/breadcrumb.component.spec.ts
+++ b/libs/core/src/lib/breadcrumb/breadcrumb.component.spec.ts
@@ -18,7 +18,9 @@ import { BreadcrumbItemDirective, BreadcrumbLinkDirective } from './public_api';
                 <a fd-breadcrumb-link [routerLink]="'#'">Breadcrumb Level 1</a>
             </fd-breadcrumb-item>
             <fd-breadcrumb-item>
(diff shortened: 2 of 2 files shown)


g-cheishvili:
This implementation is missing few points:
* If you truncate, there should be way to see whole text, besides aria title or something like it
* What about cases when text is changed? AFTER content init

Please take a look at the main branch now and try reimplementing

mikerodonnell89:
> * What about cases when text is changed? AFTER content init

@g-cheishvili True that the PR doesn't handle this, but I don't see a real-world case of this happening so in my opinion it isn't worth it to run the check more than once. If a developer needs this they can trick the breadcrumb into reloading with an *ngIf but I see that as a very rare edge-case

mikerodonnell89:
> * If you truncate, there should be way to see whole text, besides aria title or something like it

Agreed here, need some aria support and maybe a mouseover tooltip

g-cheishvili:
> > * What about cases when text is changed? AFTER content init
> 
> @g-cheishvili True that the PR doesn't handle this, but I don't see a real-world case of this happening so in my opinion it isn't worth it to run the check more than once. If a developer needs this they can trick the breadcrumb into reloading with an *ngIf but I see that as a very rare edge-case

Content might not be only text-based, it can be something else too, for example icon, object status or any HTML-based element. We can not handle all cases with just innerText

mikerodonnell89:
> > > * What about cases when text is changed? AFTER content init
> > 
> > 
> > @g-cheishvili True that the PR doesn't handle this, but I don't see a real-world case of this happening so in my opinion it isn't worth it to run the check more than once. If a developer needs this they can trick the breadcrumb into reloading with an *ngIf but I see that as a very rare edge-case
> 
> Content might not be only text-based, it can be something else too, for example icon, object status or any HTML-based element. We can not handle all cases with just innerText

True, and in that case it would probably be better to check for a large width of the element in pixels?  And I'm also thinking maybe rather than picking an arbitrary number of pixels (or characters or anything) to truncate on, we provide an input for the developer to configure the truncation point for themselves (perhaps with a default value).

g-cheishvili:
> > > > * What about cases when text is changed? AFTER content init
> > > 
> > > 
> > > @g-cheishvili True that the PR doesn't handle this, but I don't see a real-world case of this happening so in my opinion it isn't worth it to run the check more than once. If a developer needs this they can trick the breadcrumb into reloading with an *ngIf but I see that as a very rare edge-case
> > 
> > Content might not be only text-based, it can be something else too, for example icon, object status or any HTML-based element. We can not handle all cases with just innerText
> 
> True, and in that case it would probably be better to check for a large width of the element in pixels?  And I'm also thinking maybe rather than picking an arbitrary number of pixels (or characters or anything) to truncate on, we provide an input for the developer to configure the truncation point for themselves (perhaps with a default value).

Would be great, plus it should be additional separate directive

mikerodonnell89:
Closing in favor of this issue: https://github.com/SAP/fundamental-ngx/issues/8239

Issue #5041, opened 2021-03-30
Content Density Service Improvement - Composition


JKMarkowski:
Content Density usage inside core components should be improved.
One of composition features should be used for every single component, to get rid of redundant variables.
This will provide reusability for repetitive behaviours, because such will be in only one place.

IMO this could be done by creating base class with properties, that are included into every component.

- [ ] include  https://github.com/SAP/fundamental-ngx/issues/5689
- [ ] include https://github.com/SAP/fundamental-ngx/issues/7108
- [ ] review the PR for https://github.com/SAP/fundamental-ngx/issues/7765 and refactor if needed

droshev:
@g-cheishvili Could you summarize your proposal for the content density refactoring here?

g-cheishvili:
@droshev 
Goal: 
* Have centralised place for managing content density configuration - Done
* Have local configuration for certain elements in application
* Get rid of unnecessary variables across multiple components
* Have ability to store selected configuration out of the box
* Read content density configuration parameter from URL and preferably configurable name for it - Partially done

Implementation proposal:
Implement directives `[fdCompact]="true | false"` and `[fdCondensed]="true | false"` for managing locally enforced configuration. Value, provided to these variables will override globally selected one and on top of it, any component under these directives will detect local enforcement and will use it.
Implement interface, for storing selected configuration, something along the lines
```ts
interface ContentDensityStorage {
    get: () => Observable<'condensed' | 'cozy' | 'compact'>;
    set: (val: 'condensed' | 'cozy' | 'compact') => Observable<void>
}
```
This way, storage is up to the user's preference, we can implement three variants out of the box, `LocalContentDensityStorage`, `MemoryContentDensityStorage`, `UrlContentDensityStorage` and then add one generic implementation, something like `HttpContentDensityStorage`, which will be extendable class, with endpoints provided by user. Also all of the above out of the box implementations will have their own DI deps params, which will be configured by the user. Plus all of these functions can, and should be SSR compatible.
We should get rid of the inputs that are added currently to the components across the library, things such as `[cozy]`, or `[compact]` and so on. They should all be excluded from components themselves and replaced with one provider, which will be responsible for detecting global and local configs, applying them to host components.
Also for ease of development down the road, we should create decorator, something like
```
@SupportsCompact({
    compactClassName: string
    ...Some additional configs, that will be visible during development
})
```
This will just add required properties to the class and along with the up mentioned provider will make changes to the host and child components, depending on requirements

PR #8230, opened 2022-06-14
feat(core): new content density implementation

g-cheishvili:
## Related Issue(s)

Part of https://github.com/SAP/fundamental-ngx/issues/5041#issuecomment-1150397403

## Description

Implementing content density features


apps/docs/src/app/core/component-docs/content-density-2/content-density-user-2/content-density-user.component.scss
```diff
@@ -0,0 +1,24 @@
+//.example-component {
```
platon-rov, line 1: could be removed/uncommented.

apps/docs/src/app/core/component-docs/content-density-2/examples/directive-usage/directive-usage-example.component.html
```diff
@@ -0,0 +1,20 @@
+<fundamental-ngx-content-density-user fdCondensed>
+    This should be compact
+    <fundamental-ngx-content-density-user-two fdCozy> This should be cozy </fundamental-ngx-content-density-user-two>
+</fundamental-ngx-content-density-user>
+
+<!--<fundamental-ngx-content-density-user fdCompact>-->
```
platon-rov, line 6: could be uncommented/removed.

apps/docs/src/app/core/component-docs/content-density-2/content-density-user-2/content-density-user.component.ts
```diff
@@ -0,0 +1,30 @@
...
+                [ContentDensityMode.COMPACT]: 'example-component--compact',
+                [ContentDensityMode.CONDENSED]: 'example-component--condensed'
+            },
+            defaultContentDensity: ContentDensityMode.COMPACT
+        })
+    ],
+    encapsulation: ViewEncapsulation.None
+})
+export class ContentDensityUserTwoComponent {
+    constructor(@Inject(ContentDensityConsumer) readonly _contentDensityConsumer: ContentDensityConsumer) {}
```
platon-rov, line 29: i would recommend to notice that without explicitly specifying in the constructor consumer instance won't be created. kinda obvious thing but some developers may don' know this.

apps/docs/src/app/core/documentation/core-documentation-data.ts
```diff
@@ -112,7 +112,8 @@ export const utilities: SectionInterfaceContent[] = [
     { url: 'core/initial-focus', name: 'Initial Focus' },
     { url: 'core/scroll-spy', name: 'Scroll Spy' },
     { url: 'core/theme-switcher', name: 'Theme Switcher API' },
-    { url: 'core/theming', name: 'Theming' }
+    { url: 'core/theming', name: 'Theming' },
+    { url: 'core/content-density-2', name: 'Content Density 2' }
```
platon-rov, line 116: not sure about `-2` naming at all because we have to rename it in future. maybe we can rename the previous docs folder to `content-density-old` or something like that? and the new one will be just a `content-density`. then, somewhen in future we will just remove the old one.

libs/core/src/lib/content-density/directives/content-density.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+    providers: [
+        {
+            provide: CONTENT_DENSITY_DIRECTIVE,
+            useExisting: forwardRef(() => ContentDensityDirective)
+        }
+    ]
+})
+export class ContentDensityDirective extends BehaviorSubject<LocalContentDensityMode> implements OnDestroy {
+    @Input()
+    set fdContentDensity(val: LocalContentDensityMode | '') {
```
platon-rov, line 19: typedocs.

platon-rov (commented):
brilliant approach. left some minor comments. also i see that examples a bit broken

https://user-images.githubusercontent.com/20265336/173627331-7bf09394-1f73-4e4e-b240-ed6e8b867fe0.mov

g-cheishvili:
Yeah, that example should not be there at all, because I reverted back changes to fd-button :D

apps/docs/src/app/core/component-docs/content-density-2/content-density-user-2/content-density-user.component.ts
g-cheishvili: I would say Examples are not ready yet with their docs... I should probably ask @ynnenu for help

mikerodonnell89 (commented):
@platon-rov @g-cheishvili regarding naming, personally I would like to see this new service titled "ContentDensityGlobalService" as I think that would help devs understand the difference between this and the old

mikerodonnell89:
Maybe some examples with components that do have all 3 sizes would be helpful

g-cheishvili:
Guys, take a quick look again. Examples are more or less fixed

g-cheishvili:
> 

I think it would be better if it was called `GlobalContentDensityService` then

libs/core/src/lib/content-density/providers/local-content-density-storage.ts
```diff
@@ -0,0 +1,44 @@
+import { Inject, Injectable } from '@angular/core';
+import { LocalStorageService } from '@fundamental-ngx/core/utils';
+import { Observable, of, Subject } from 'rxjs';
+import { ContentDensityStorage } from '../classes/abstract-content-density-storage';
+import { DEFAULT_CONTENT_DENSITY } from '../tokens/default-content-density.token';
+import { CONTENT_DENSITY_STORAGE_KEY } from '../tokens/content-density-storage-key.token';
+import { ContentDensityMode } from '../content-density.types';
+
+@Injectable()
+export class LocalContentDensityStorage implements ContentDensityStorage {
```
dmitry-stepanenko, line 10: actually with local storage based implementation we can act directly on the change event in the storage

Check this implementation of reactive local storage, that also tracks cross-window events
[local-storage.service.txt](https://github.com/SAP/fundamental-ngx/files/8902726/local-storage.service.txt)

libs/core/src/lib/content-density/content-density.module.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+    return [];
+}
+
+@NgModule({
+    imports: [CommonModule],
+    exports: [ContentDensityDirective],
+    declarations: [ContentDensityDirective]
+})
+export class ContentDensityModule {
```
dmitry-stepanenko, line 42: since module can be configured, can you please add an example for this?

dmitry-stepanenko (commented):
Looks neat! Just a few minor comments/suggestions

apps/docs/src/app/core/component-docs/content-density-2/content-density-docs.component.html
```diff
@@ -0,0 +1,30 @@
+<fd-docs-section-title id="content-density-service" componentName="content-density-service">
+    ContentDensityControllerService
+</fd-docs-section-title>
+<description>
+    Content Density can be maintained app-wide using the <code>ContentDensityControllerService</code>, and components
+    can subscribe to a <code>Observable</code> managed by the service and use that to handle changes to the content
+    density. Before using it, you should add <code>ContentDensityModule</code> to imports array of
```
platon-rov, line 7: please notice that `forRoot` method should be used to provide a service.

apps/docs/src/app/core/component-docs/content-density-2/content-density-user/content-density-user.component.ts
```diff
@@ -0,0 +1,40 @@
+import { Component, Inject, ViewEncapsulation } from '@angular/core';
```
platon-rov, line 1: maybe we should remove to `content-density-example-consumer`?

apps/docs/src/app/core/component-docs/content-density-2/examples/directive-usage/directive-usage-example.component.ts
```diff
@@ -0,0 +1,7 @@
+import { Component } from '@angular/core';
+
+@Component({
+    selector: 'fd-directive-usage-example',
+    templateUrl: './directive-usage-example.component.html'
+})
+export class DirectiveUsageExampleComponent {}
```
platon-rov, line 7: minor: not properly works yet.

apps/docs/src/app/core/component-docs/content-density-2/content-density-user/content-density-user.component.ts
```diff
@@ -0,0 +1,40 @@
...
+import { DestroyedService } from '@fundamental-ngx/core/utils';
+import {
+    ContentDensityConsumer,
+    contentDensityConsumer,
+    ContentDensityMode
+} from '@fundamental-ngx/core/content-density';
+
+@Component({
+    selector: 'fundamental-ngx-content-density-user',
```
InnaAtanasova, line 10: Please use `fd-docs-.......`  so it's clear for the user that this is a documentation related component.

libs/core/src/lib/content-density/content-density.module.ts
InnaAtanasova: +1

InnaAtanasova:
The button should not have `condensed` state. This state belongs only to Table component (as far as I know, could be 1-2 more components, but for sure none of the components we have in Fundamentals). 
Screen Shot 2022-06-17 at 10 59 37 AM

InnaAtanasova:
The example is not working properly. This is compact state

Screen Shot 2022-06-17 at 12 01 44 PM

libs/core/src/lib/content-density/providers/content-density-consumer.ts
```diff
@@ -0,0 +1,68 @@
...
+
+            changesSource$.pipe(takeUntil(destroy$)).subscribe();
+
+            return new ContentDensityConsumer(changesSource$);
+        },
+        deps: [
+            ElementRef,
+            DestroyedService,
+            [new Optional(), new SkipSelf(), CONTENT_DENSITY_DIRECTIVE],
+            [new Optional(), new Self(), CONTENT_DENSITY_DIRECTIVE],
```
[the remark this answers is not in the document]
g-cheishvili: Yes, you are right. Was thinking about that exact thing today :D

platon-rov (commented):
please, edit `api-files.ts` file so user can see typedocs on the second tab

image

mikerodonnell89 (commented):
Toolbar and calendar examples still feature the old `ContentDensityService`, shall we update those as well?

g-cheishvili:
> Toolbar and calendar examples still feature the old `ContentDensityService`, shall we update those as well?

For replacing across core and docs I have separate branch. This is pure implementation and documentation of this feature

libs/core/src/lib/content-density/directives/content-density.directive.ts
```diff
@@ -0,0 +1,78 @@
...
+    selector: '[fdContentDensity], [fdCompact], [fdCondensed], [fdCozy]',
+    exportAs: 'fdContentDensity',
+    providers: [
+        {
+            provide: CONTENT_DENSITY_DIRECTIVE,
+            useExisting: forwardRef(() => ContentDensityDirective)
+        }
+    ]
+})
+export class ContentDensityDirective extends BehaviorSubject<LocalContentDensityMode> implements OnDestroy {
```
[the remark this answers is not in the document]
g-cheishvili: I will disable several inputs on same element

apps/docs/src/app/core/component-docs/content-density-2/content-density-user/content-density-user.component.ts
g-cheishvili: Nah, it works too 😁

libs/core/src/lib/content-density/providers/local-content-density-storage.ts
g-cheishvili: We can do that in utils separately. would be nice to have

dmitry-stepanenko: As you with 🤷 You're creating that local storage service anyway, so I thought it would be better to have more feature rich implementation

apps/docs/src/app/core/component-docs/old-content-density/content-density-docs.component.html
```diff
@@ -0,0 +1,20 @@
+<description>
+    Content Density can be maintained app-wide using the <code>ContentDensityService</code>, and components can
```
platon-rov, line 2: `GlobalContentDensityService`

libs/core/src/lib/content-density/classes/content-density-consumer.class.ts
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * Remove consumer callback from list of callbacks
+     */
+    removeConsumer(
+        ...consumerConfigs: Array<ContentDensityConsumerTarget | ContentDensityCallbackFn>
+    ): ContentDensityObserver {
+        consumerConfigs.forEach((consumerConfig) => {
+            this.callbacks.delete(consumerConfig);
+        });
+        this._callCallbacks$.next();
```
platon-rov, line 121: every time we add or remove a consumer we run all the consumers because of this emit. but in the typedocs said `to be called on density change`. is it intended?

libs/core/src/lib/content-density/classes/content-density-consumer.class.ts
```diff
@@ -0,0 +1,140 @@
...
+        if (!this._cozySubscription) {
+            this._cozySubscription = this._listenToChanges(this.isCozy$, (v) => (this._isCozy = v));
+        }
+        return this._isCozy;
+    }
+
+    /**
+     * Pass consumers callback function, or configuration object to be called on density change
+     */
+    consume(
```
platon-rov, line 101: i cannot fully understand why do we have this. 

example. we may have component `A` that is a **consumer**. classes applied and so on. 

_first scenario_: 
if there is a nested comp `B` - we can make it consumer too. 

_second scenario_: 
if there are some elements in template of `A` (not a components for some reason) and we want to apply classes to them depending on content density we could write `[class.cozy]="_contentDensityConsumer.isCozy$ | async"` and that's it. 
if we want to run custom code we may do `_contentDensityConsumer.isCozy$.subscribe(() => { ... })` and that's it.
or we could create a template variable, query for it with `ContentChild()` and only then pass it to the `_contentDensityConsumer.consume(...)` and we will have the same result?

libs/core/src/lib/content-density/classes/content-density-consumer.class.ts
```diff
@@ -0,0 +1,140 @@
...
+    ): ContentDensityObserver {
+        consumerConfigs.forEach((consumerConfig) => {
+            this.callbacks.delete(consumerConfig);
+        });
+        this._callCallbacks$.next();
+        return this;
+    }
+
+    /** @hidden */
+    _callCallbacks(density: ContentDensityMode): void {
```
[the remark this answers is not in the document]
g-cheishvili: Yes :/

libs/core/src/lib/content-density/classes/content-density-consumer.class.ts
g-cheishvili: Yeah, I encountered such scenarios, where content density classes are added to elements, which did not have directive or component, that's why If someone wants to have everything focused in TS, I created that support. If they want to have it on HTML, that is totally possible too

g-cheishvili: Plus, it's open source, I do not know what weird things are other people doing, they may be creating elements through document,createElement and then injecting in DOM, so if they want to add content density to their components, they can include that

libs/core/src/lib/content-density/classes/content-density-consumer.class.ts
g-cheishvili: Hmm, now when you mentioned it, I remember that I should not call everyone, one is enough :D

libs/core/src/lib/content-density/providers/content-density-observer.ts
```diff
@@ -0,0 +1,83 @@
...
+import { CONTENT_DENSITY_DIRECTIVE } from '../tokens/content-density-directive';
+import { ContentDensityObserver } from '../classes/content-density-consumer.class';
+import { defaultContentDensityObserverConfigs } from '../variables/default-content-density-consumer-config';
+import { getChangesSource$ } from '../helpers/get-changes-source.provider';
+
+/**
+ * Creates provider for ContentDensityObserver
+ * @param providedConfiguration
+ */
+export function contentDensityObserver(providedConfiguration?: ContentDensityObserverSettings): Provider {
```
platon-rov, line 15: we could have only one factory for all the consumers, something like this

```ts
export function consumerFactory(Consumer): Provider {
  return {
    provide: Consumer,
    useFactory: (injector: Injector) => new Consumer(injector),
    deps: [Injector],
  };
}
```

then, in our certain consumer we will use `injector.get(...)` to get any needed  stuff, so we would have something like that 

```ts
export class Consumer {
  constructor(private readonly injector: Injector) {
    // get custom token
    console.log(injector.get(Token));
  }

  consume(fn?: (injector: Injector) => any): void {
    if (fn) {
      // custom consumer logic
      fn(this.injector);
    } else {
      // default consumer logic
      this.injector
        .get(ElementRef)
        .nativeElement.classList.add('default-class');
    }
  }
}
```

very basic examples but i think logic is clear, let me know if not.

this way all the logic related to the consumer will be in the consumer itself and we won't need to create many factories. for sure some common logic for all the consumers may be moved to the factory (like destroyed service).

g-cheishvili: Could be, but the way it is done right now seemed more explicit to me

platon-rov (approved):
nice!

libs/core/src/lib/content-density/providers/consumer-provider-factory.ts
```diff
@@ -0,0 +1,11 @@
+import { Provider } from '@angular/core';
+import { consumerProviderFactory, DestroyedService } from '@fundamental-ngx/core/utils';
+import { ContentDensityObserverSettings } from '../content-density.types';
+import { ContentDensityObserver } from '../services/content-density-observer.service';
+
+/**
+ * Creates provider for ContentDensityObserver and adds DestroyedService provider
+ */
+export function contentDensityObserverProviders(params?: ContentDensityObserverSettings): Provider[] {
```
platon-rov, line 9: filename probably should be changed to be in sync with exported function.

Issue #7838, opened 2022-03-04
Skeleton Component


droshev:
#### Is this a bug, enhancement, or feature request?
feature request

#### Briefly describe your proposal.
We would expect that Fundamentals supports something similar as MUI currently supports:

https://mui.com/components/skeleton/

Skeleton loading should be available for various elements, such as table, forms, headers, list, side navigation.

(dxp/ngx-dxp/issues/75)

The first goal:
- [x] research how this is done in other libraries
- [x] gather all needed information/design/assets
- [x] research and document the impact for NGX (maybe doing PoC)
- [x] breaking changes
- [x] present the summary after scrum
- [x] trying to break into tasks the implementation

Next Sprint Task - 1w
- [ ] skeleton impl in simple components (e.g. button, avatar)
- [ ] skeleton impl in complex components (e.g. table, form container)
- [ ] introduce a skeleton switch in doc examples
- [ ] present it to the designers for feedback

platon-rov:
There exists two ways of providing this feature, let me describe pros & cons of both.

1. CSS Class. Something like here https://taiga-ui.dev/skeleton. It's a way simpler on the first glance, only applying of css class is needed. Because of the reusing existing HTML elements and putting this skeletons via pseudo elements they can inherit size & border radius what is a very handy. The downsides are that in case of a complex components where not all the elements initially visible there is nothing to apply CSS classes to so we have to create an empty elements on the our side. Also, maybe the biggest cons, is that we should manage all the classes applying for the complex components (where not all the elements exposed to the customer) in our library, for the every single component (add not related to the component itself inputs & manage all that stuff in the template). 
_I'm against this approach but have to mention it here._

2.

platon-rov:
Okay, after a call with @g-cheishvili he gave me idea that we can use both approaches together.

For some components we cannot use CSS Class approach because there is no html markup to apply those classes (exactly that i mentioned before). So we have to rely on the Separate Component approach like so:

```
<fd-table fdLoading>
    <tr *ngFor="let row of tableRows | async">
        <!-- Table cells are not visible before tableRows, cannot apply skeleton classes to the table cells -->
    </tr>

    <!-- Show loadingTemplate till tableRows aren't loaded -->
    <ng-template #loadingTemplate>
        <tr>
            <td><fd-skeleton></fd-skeleton></td>
            <td><fd-skeleton></fd-skeleton></td>
            <td><fd-skeleton></fd-skeleton></td>
        </tr>
[...]

PR #8258, opened 2022-06-21
[draft] feat(core,platform): skeleton & skeleton state for components

platon-rov:
## Related Issue(s)

Part of https://github.com/SAP/fundamental-ngx/issues/7838

## Description

Skeleton component, directives.

* Complex skeleton templates were created for all the components (core & platform) except the ones mentioned here (with reasons) https://github.com/SAP/fundamental-ngx/issues/8601. 

* Wrong colors issue https://github.com/SAP/theming-base-content/issues/12

* Inline CSS were used in skeleton templates until common styles are not available (not released).

**_Note:_**

* Gap issues are related to the components themselves, not the skeleton.

  image

### After:

image


libs/core/src/lib/avatar/avatar.component.ts
```diff
@@ -40,7 +42,8 @@ const ALTER_ICON_OPTIONS = {
     changeDetection: ChangeDetectionStrategy.OnPush,
     host: {
         '[attr.tabindex]': '_tabindex'
-    }
+    },
+    providers: [DestroyedService, skeletonConsumer()]
```
platon-rov, line 46: that's example how we add skeleton state for the simple components (one placeholder).

libs/core/src/lib/table/table.component.html
```diff
@@ -0,0 +1,63 @@
+<ng-template [fdSkeletonTemplate]="skeletonTemplate">
+    <ng-content></ng-content>
+</ng-template>
+
+<ng-template #skeletonTemplate>
```
platon-rov, line 5: that's example how we add skeleton state for the complex components (multiple placeholders).

apps/docs/src/app/core/component-docs/skeleton/examples/directive/skeleton-template-directive-example.component.html
```diff
@@ -0,0 +1,61 @@
...
+    </fd-card>
+
+    <fd-card badge="New" cardType="list" *fdSkeleton="'global'; template: loadingCard">
+        <fd-card-header>
+            <h2 fd-card-title>Card reacts to global skeleton state and it has a template on the card level.</h2>
+        </fd-card-header>
+
+        <fd-card-content>
+            <ul fd-list [noBorder]="true">
+                <li *ngFor="let i of [1, 2, 3, 4]; let index = index" fd-list-item>
```
dmitry-stepanenko, line 43: maybe it would be reasonable to include an "*fdRepeat" directive, that can be used like
```
<comp *fdRepeat="5"></comp>
```

I believe it will be pretty common scenario for skeletons to have such repeatable dummy structures

apps/docs/src/app/core/component-docs/skeleton/skeleton-header/skeleton-header.component.html
```diff
@@ -0,0 +1,22 @@
...
+    <code>forRoot</code> method of <code>SkeletonModule</code> in the root module of your application. It's also
+    possible to set <code>defaultSkeletonState</code>(<code>true | false (default)</code>) via this way.
+
+    <br /><br />
+
+    <b>Notice</b>: We will iteratively add skeleton state support out of the box to all the components but before it
+    happened please provide your own templates.
+</description>
+
+<import module="SplitterModule" subPackage="splitter"></import>
```
dmitry-stepanenko, line 18:
```suggestion
<import module="SkeletonModule" subPackage="skeleton"></import>
```

libs/core/src/lib/skeleton/directives/skeleton-template.directive.ts
```diff
@@ -0,0 +1,65 @@
...
+    @Input()
+    fdSkeletonTemplate: TemplateRef<any>;
+
+    /** @hidden */
+    private readonly _onDestroy$ = new Subject<void>();
+
+    /** @hidden */
+    constructor(
+        @Inject(SKELETON_DIRECTIVE) @Optional() private readonly _parentSkeletonDirective: SkeletonDirective,
+        @Optional() private readonly _skeletonService: SkeletonService,
```
dmitry-stepanenko, line 33: please make them nullable

libs/core/src/lib/skeleton/providers/skeleton-consumer.ts
```diff
@@ -0,0 +1,50 @@
...
+export function skeletonConsumer(providedConfiguration?: SkeletonConsumerConfig): Provider {
+    const configuration = { ...defaultSkeletonConsumerConfig, ...providedConfiguration };
+
+    return {
+        provide: SkeletonConsumer,
+        useFactory: (
+            elementRef: ElementRef<Element>,
+            destroy$: Observable<void>,
+            skeletonDirective?: SkeletonDirective,
+            contentDensityService?: SkeletonService
```
dmitry-stepanenko, line 26: should be skeletonService?

libs/core/src/lib/skeleton/components/skeleton.component.ts
```diff
@@ -0,0 +1,71 @@
...
+    /** @hidden */
+    @HostBinding('class.fd-skeleton--circle')
+    get _circleClass(): boolean {
+        return this.type === 'circle';
+    }
+}
+
+function getRandomNumber(min: number, max: number): number {
+    return Math.floor(Math.random() * (max - min + 1)) + min;
+}
```
dmitry-stepanenko, line 67: maybe it's better to switch to lodash since we're using it a lot in other places https://lodash.com/docs/4.17.15#random

libs/core/src/lib/skeleton/components/skeleton.component.ts
```diff
@@ -0,0 +1,71 @@
+import { ChangeDetectionStrategy, Component, HostBinding, Input, ViewEncapsulation } from '@angular/core';
+
+export type SkeletonType = 'circle' | 'rectangle' | 'text';
+export type SkeletonWidth = 'rand' | string;
+export type SkeletonHeight = 'auto' | string;
+
+@Component({
+    selector: 'fd-skeleton',
+    template: ``,
```
dmitry-stepanenko, line 9: maybe you could also add content projection here while making the content fully transparent? that will allow to have shapes based on fixed text size (e.g. exactly the width of `lorem ipsum` following the text size in the particular container)


Something like
```
<fd-skeleton width="inherit">lorem ipsum</fd-skeleton>


fd-skeleton.inherited-width {
    width: fit-content;
    color: transparent;
}
```

libs/core/src/lib/avatar/avatar.component.ts
```diff
@@ -200,7 +203,11 @@ export class AvatarComponent implements OnChanges, OnInit, CssClassBuilder {
     }
 
     /** @hidden */
-    constructor(private _elementRef: ElementRef, private _cdr: ChangeDetectorRef) {}
+    constructor(
+        private _elementRef: ElementRef,
+        private _cdr: ChangeDetectorRef,
+        private readonly _skeletonConsumer: SkeletonConsumer
```
dmitry-stepanenko, line 209: just thinking: it's quite not obvious from the first glance, why this service is imported without being references even once. Maybe in order to be more explicit it would be better to pass the elementRef in the constructor of the component instead of using DI? That will visually tie the component with the logic of the consumer

dmitry-stepanenko (commented):
Lovely! Left few thoughts here and there

libs/core/src/lib/skeleton/components/skeleton.component.ts
```diff
@@ -0,0 +1,73 @@
...
+export type SkeletonHeight = 'auto' | string;
+
+@Component({
+    selector: 'fd-skeleton',
+    template: `<ng-content></ng-content>`,
+    styleUrls: ['skeleton.component.scss'],
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    // eslint-disable-next-line @angular-eslint/no-host-metadata-property
+    host: {
```
N1XUS, line 15: Please move it to
```
@HostBinding('class')
private readonly _initialClass = 'fd-skeleton';

File: libs/core/src/lib/skeleton/components/skeleton.component.ts [hunk 11]
```diff
@@ -0,0 +1,73 @@
...
+        class: 'fd-skeleton'
+    }
+})
+export class SkeletonComponent {
+    /** Wight of the skeleton */
```
N1XUS, line 20: Typo in width

libs/core/src/lib/skeleton/components/skeleton.component.ts
```diff
@@ -0,0 +1,73 @@
...
+    set width(value: SkeletonWidth) {
+        if (value === 'rand') {
+            this._width = getRandomWidth();
+            return;
+        }
+
+        this._width = value;
+    }
+
+    /** Height of the skeleton. Relevant if type 'circle' or 'rectangle'. */
```
N1XUS, line 31: If the type is circle, is it really needed to use height and width? It's kinda round, so it either one of those values that needs to be used (not sure if we should support ellipses)

libs/core/src/lib/skeleton/components/skeleton.component.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    /** @hidden */
+    @HostBinding('class.fd-skeleton--circle')
+    get _circleClass(): boolean {
+        return this.type === 'circle';
+    }
+}
+
+function getRandomWidth(): string {
+    return `${random(30, 90)}%`;
```
N1XUS, line 72: Should we allow users to rewrite this values?

libs/core/src/lib/skeleton/directives/skeleton.directive.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    /** @hidden */
+    ngOnDestroy(): void {
+        this._onDestroy$.next();
+
+        this.complete();
+    }
+
+    /** @hidden */
+    _updateView(loadingState: boolean): void {
```
N1XUS, line 68: Is it used anywhere else? maybe make it private?

libs/core/src/lib/skeleton/directives/skeleton-template.directive.ts
```diff
@@ -0,0 +1,67 @@
...
+            .subscribe((loadingState) => this._updateView(loadingState));
+    }
+
+    /** @hidden */
+    ngOnDestroy(): void {
+        this._onDestroy$.next();
+    }
+
+    /** @hidden */
+    _updateView(loadingState: boolean): void {
```
N1XUS, line 54: Is it used anywhere else? maybe make it private?

libs/core/src/lib/skeleton/components/skeleton.component.ts
```diff
@@ -0,0 +1,73 @@
...
+    animation = true;
+
+    /** @hidden */
+    @HostBinding('style.width')
+    _width: string;
+
+    /** @hidden */
+    @HostBinding('style.height')
+    get _height(): string {
+        return this.type === 'text' ? 'auto' : this.height;
```
N1XUS, line 55: Following the above comment, maybe add a fallback to use width if height is not provided at all so it would draw perfect circle or square by just providing width OR height

apps/docs/src/app/core/component-docs/skeleton/examples/directive/skeleton-template-directive-example.component.html
g-cheishvili: Great idea!

libs/core/src/lib/avatar/avatar.component.ts
g-cheishvili: I do not know, whole idea behind the Consumers was that it can be low config implicit. We could add config function to the consumers and that could be optional thing

libs/core/src/lib/skeleton/components/skeleton.component.ts
platon-rov: i don't see much sense in that, let's add it in future if there will be such request, what do you think?

N1XUS: Fine with me :)

mikerodonnell89 (commented):
Build is failing I think on these lines https://github.com/SAP/fundamental-ngx/runs/7090921468?check_suite_focus=true#step:5:980 https://github.com/SAP/fundamental-ngx/runs/7090921468?check_suite_focus=true#step:5:2209

platon-rov:
`[BLOCKED]` https://github.com/SAP/fundamental-styles/pull/3766

droshev:
Comments:
- most components (e.g. title, switch, rating indicator, radio btn, checkbox show placeholders for example 20% of the width)
- Table  and List (show 3 rows for example and make the cells same width)
- Toolbar, Tabs, Action Bar - create a specific ( 2 placeholders left/right)
- vertical nav, side nav - specific (2-3 placeholders)
- Upload collection - specific look Header: Object Page
- Timeline like Control: Tiles
- fields (e.g input, timepicker, datepicker, textarea, text, switch, step input, status indicator, select, multi input, input group, file upload, FeedList, combo,  ... add some padding so the grey doesn't take 100%)
- Tile - Control Tile
- Shellbar - Control: Shell
- Quick View - Control: Tile
- Panel - Control: Tile upper part only
- Object Marker - it should not jump
- Notification: Control: Tile
- Grid List Item: Control: Tile
- Feed Input: Control: Toolbar
- Facet: Control: Object Page
- Cart: container

InnaAtanasova:
When reviewing the PR keep in mind the following: https://github.com/SAP/fundamental-ngx/issues/8601

mikerodonnell89:
Looking at the dialog examples, looks like some skeletons appear where there were not items previously (top right of header and bottom left of footer). Does not make sense to me to have skeletons that will disappear and show nothing when the skeleton state is set to false
Screen Shot 2022-08-30 at 2 11 40 PM
Screen Shot 2022-08-30 at 2 11 48 PM

platon-rov:
> Looking at the dialog examples, looks like some skeletons appear where there were not items previously (top right of header and bottom left of footer). Does not make sense to me to have skeletons that will disappear and show nothing when the skeleton state is set to false

Yea, that was the previous approach - to show skeletons where content is. But then requirements changed a bit so now we want to have predefined placeholders accordingly to the component, not to its content. 

For example, if it's a toolbar - show 2 placeholders on sides no matter what. 

It makes sense if we don't have content placed yet, because otherwise we won't see placeholders at all.

InnaAtanasova:
The following comments are `PERSONAL OPINION`. This said, it doesn't mean you have to address them, it's just something I don't like/thing needs improvement.

1. Circle Avatar should have circle Skeleton
2. Avatar Group can be mistaken with an input field. In my opinion stacked circles or just circles next to each other would make more sense:
Screen Shot 2022-09-23 at 11 15 08 AM
Screen Shot 2022-09-23 at 11 15 03 AM

3. Slider Skeleton also doesn't make sense. In my opinion a think line with a circle in the middle is more appropriate. 
Screen Shot 2022-09-23 at 11 22 26 AM
Screen Shot 2022-09-23 at 11 22 35 AM

I'd like to see something like this:
Screen Shot 2022-09-23 at 11 37 55 AM

4. Breadcrumb. I'd like to see:
Screen Shot 2022-09-23 at 11 39 38 AM

5. Labels shouldn't be this long (esp. compared to input)
I'd like to see:
Screen Shot 2022-09-23 at 11 41 14 AM

InnaAtanasova:
6. Carousel would look better like this:
Screen Shot 2022-09-23 at 11 46 53 AM

7. Calendar
Screen Shot 2022-09-23 at 11 50 42 AM

InnaAtanasova:
8. Date Picker, Select, Multi Combo, etc in my opinion should looks like this: 
Screen Shot 2022-09-23 at 11 54 54 AM

Having a veeeery long label compared to a short input field for me is weird

InnaAtanasova:
9. Radio and Checkbox
Screen Shot 2022-09-23 at 11 57 50 AM

InnaAtanasova:
10. File Uploader
Screen Shot 2022-09-23 at 11 59 43 AM

InnaAtanasova:
11. 
Screen Shot 2022-09-23 at 12 00 33 PM


should be something like:
Screen Shot 2022-09-23 at 12 01 34 PM

InnaAtanasova:
12. Micro Process Flow
Screen Shot 2022-09-23 at 12 03 17 PM

something like:
Screen Shot 2022-09-23 at 12 04 04 PM

InnaAtanasova:
13. Panel
Screen Shot 2022-09-23 at 12 06 45 PM
Screen Shot 2022-09-23 at 12 05 23 PM

InnaAtanasova:
14. Dialogs
Screen Shot 2022-09-23 at 12 24 18 PM

droshev:
Components:
- ✅ Action bar 
- Action Sheet - within the space of the btn
- Alert - maybe we don't need skeleton for that component
- Avatar - circle skeleton for circle avatar
- Avatar Group - apply the toolbar pattern (2-3 circle placeholders)
- ✅ Bar 
- Breadcrumb ?
- Busy Indicator - no skeleton
- Buton ?
- ✅ Card
- ✅ Segmented Button
- Carousel - maybe we can have 2 parts (bottom as a toolbar and the upper part can be one block)
- checkbox - no placeholder
- ✅ split button
- Calendar - no placeholder
- ✅ Combobox
- ✅ Datepicker
- ✅ DTPicker
- Label - the skeleton should be not bigger than the label

PR #8259, opened 2022-06-21
chore(core,docs): deprecated old content density inputs

g-cheishvili:
## Related Issue(s)

closes none

## Description

Deprecated all usages of content-density related inputs in core and updated docs related to it


libs/core/src/lib/action-sheet/action-sheet.component.ts
```diff
@@ -125,14 +110,11 @@ export class ActionSheetComponent implements AfterContentInit, AfterViewInit, On
...
 
-    /** @hidden */
-    private _compact: boolean;
-
     constructor(
         private readonly _keyboardSupportService: KeyboardSupportService<ActionSheetItemComponent>,
         private readonly _changeDetectionRef: ChangeDetectorRef,
         private readonly _viewContainerRef: ViewContainerRef,
-        @Optional() private _contentDensityService: ContentDensityService,
+        @Inject(ContentDensityConsumer) private _contentDensityConsumer: ContentDensityConsumer,
```
platon-rov, line 117: everywhere consumer is provided we also should mention the destroyed service so it would create its instance.

platon-rov (approved):

PR #8264, opened 2022-06-22
fix(core): update token tabindex behavior

### LINKED ISSUE #8171: fd-token accessibility: Not able to reach tokens using keyboard (tab key)
Bug

fd-tokens are not reachable via keyboard mostly using tab key, there is no element having tabindex=0 for fd-token

Angular 13
    "@fundamental-ngx/core": "0.33.4",
    "@fundamental-ngx/platform": "0.33.4",

https://sap.github.io/fundamental-ngx/#/core/token - try to reach all tokens using tab key - none of them are accessible

mikerodonnell89:
fixes #8171 

adds tab index to tokens without changing tab/keyboard arrow functionality of tokens in tokenizers


libs/core/src/lib/token/token.component.html
```diff
@@ -2,8 +2,10 @@
     #tokenWrapperElement
     class="fd-token"
     role="option"
+    tabindex="0"
```
I543348, line 5: hope the css for focus is taken care ?
Thank you.

Rest of the code looks fine to me and is as expected.

mikerodonnell89: Yes, the tokens receive the dotted focus within the token when they receive tab focus. Note that this PR only changes the functionality for tokens that are not part of a tokenizer

droshev (approved):

PR #8266, opened 2022-06-23
feat(fn): bring more FN components

N1XUS:
Add new fiori next components


InnaAtanasova:
Please use the proper naming for the PR title

mikerodonnell89 (commented):
This is looking nice! Many of the new components and cdk utils are missing the API docs (maybe coming later?)

jonas-meister-sap:
I walked through the demo / documentation and found some issues:

Bottom line of input / select is now very dark "black". Should be / was  gray:
image

Select Semantic State - touch feedback is not centered
image

Select Options appearing behind other elements:
image


Looks like the "normal tags" are no longer present in the documentation
image

Toasts appearing behind footer
image

droshev (approved):

PR #8289, opened 2022-06-27
feat(core): fd-truncate improvements

### LINKED ISSUE #8239: Implement fd-truncate directive
A helpful directive for the ngx libraries would be a truncation directive that the developer can apply to any component to limit the length of text (or some other content). The directive should accept as input a length specified in characters, or a length specified in pixels, in the event that the content they wish to be truncated contains some non-text items such as icons or small images.

shrvr:
## Related Issue(s)

Closes #8239 

## Description

Created fd-truncate directive that allow developer to truncate element by character length and pixel length of the element.

#### 1

Screen Shot 2022-06-25 at 4 53 09 PM

 #### Result

Screen Shot 2022-06-25 at 4 52 47 PM

#### 2

Screen Shot 2022-06-25 at 4 51 00 PM

 #### Result

Screen Shot 2022-06-25 at 4 50 28 PM


libs/core/src/lib/utils/directives/truncation/truncation.directive.ts
```diff
@@ -0,0 +1,55 @@
...
+            this._truncateByPixel(rootElement);
+        } else {
+            this._truncateByChar(rootElement);
+        }
+    }
+
+    /** @hidden
+     * Truncates element by width
+     */
+    private _truncateByPixel(_element): void {
```
mikerodonnell89, line 31:
```suggestion
    private _truncateByPixel(_element: HTMLElement): void {
```

Nice code! Let's do types for function parameters

libs/core/src/lib/utils/directives/truncation/truncation.directive.ts
```diff
@@ -0,0 +1,55 @@
...
+        _element.style.textOverflow = 'ellipsis';
+        _element.style.whiteSpace = 'nowrap';
+        _element.style.width = `${this.pixLength}px`;
+    }
+
+    /** @hidden
+     * Truncates element's text by characters
+     */
+    private _truncateByChar(_element): void {
+        console.log(this._elementRef);
```
mikerodonnell89, line 42: I know this is a WIP so maybe you were already planning on removing it, but we try to avoid console logs in the codebase

mikerodonnell89 (changes requested):

libs/core/src/lib/utils/directives/truncation/truncation.directive.ts
```diff
@@ -0,0 +1,55 @@
...
+    @Input() charLength: number;
+    @Input() pixLength: number;
+
+    /** @hidden */
+    private defaultTruncationByChar = 30;
+
+    constructor(private _elementRef: ElementRef) {}
+
+    /** @hidden */
+    ngAfterViewInit(): void {
```
mikerodonnell89, line 17: I think we'll want to run this with `ngAfterViewChecked` rather than init, in case the text changes after init

mikerodonnell89 (commented):
So I'm seeing this is a new directive entirely and now we have two fd-truncate directives - i think let's bring this logic into the existing `truncate.directive.ts` rather than making a new file

shrvr:
## Related Issue(s)

closes #8292
closes #8239 

## Description

Changed existing fd-truncate to add truncation of an element by pixel width length.

`<div  fd-truncate [fdTruncateState]="isTruncate" [fdTruncateWidth]="200">`

libs/core/src/lib/utils/directives/truncate/truncate.directive.ts
```diff
@@ -53,6 +61,12 @@ export class TruncateDirective implements OnChanges, AfterViewInit, OnDestroy {
     /** @hidden */
     private _truncateTarget: HTMLElement;
 
+    /** @hidden
+     * If fdTruncateTargetText is not provided this directive truncates Text Content of an element
+     * This Text Content of an element
+     */
+    private fdTruncateTargetTextContent: string;
```
InnaAtanasova, line 68: if this is private it should be prefixed with an underscore.

libs/core/src/lib/utils/directives/truncate/truncate.directive.ts
```diff
@@ -62,17 +76,43 @@ export class TruncateDirective implements OnChanges, AfterViewInit, OnDestroy {
...
+
+    /** @hidden
+     * saves default style of element before truncating
+     */
+    private defaultStyle: string;
+
+    /** @hidden
+     * truncation style for truncating element
+     */
+    private truncationStyle: string;
```
InnaAtanasova, line 115: prefix for private property is missing

libs/core/src/lib/utils/directives/truncate/truncate.directive.ts
```diff
@@ -84,12 +112,32 @@ export class TruncateDirective implements OnChanges, AfterViewInit, OnDestroy {
...
+            this._setTextContent();
+            this._setDefaultStyle();
+
+            this.refreshTarget(this);
+            this.reset();
+
             this._initTruncate();
 
-            this.windowResize$ = fromEvent(window, 'resize')
+            this._windowResize$ = fromEvent(window, 'resize')
```
platon-rov, line 140: instead of this, we could use `resizeObservable` from the `@fundamental-ngx/core/utils`.

platon-rov (commented):
Good!

However we don't react to element's content change because we replaced that content with cut string at the beginning. 

```
// user provided
<div fd-truncate fdTruncateChars="10">Some content...</div>

// then we cut the text and replaced nativeElement.textContent
<div fd-truncate fdTruncateChars="10">Some conte...</div>

// then user changed the text but we don't see it
<div fd-truncate fdTruncateChars="10">Short...</div>
```

So probably we have to change the `textContent` only if text provided via `fdTruncateTargetText` input property. Otherwise we should not change the content to don't lose the content provided by the user. What do you think?

After all looks like we could have two cases of directive usage
* `fdTruncateTargetText` + `fdTruncateChars` or `fdtruncatewudth` = alter the content.
* `fdTruncateWidth` = set styles, **don't** alter the content.

mikerodonnell89:
> Good!
> 
> However we don't react to element's content change because we replaced that content with cut string at the beginning.
> 
> ```
> // user provided
> <div fd-truncate fdTruncateChars="10">Some content...</div>
> 
> // then we cut the text and replaced nativeElement.textContent
> <div fd-truncate fdTruncateChars="10">Some conte...</div>
> 
> // then user changed the text but we don't see it
> <div fd-truncate fdTruncateChars="10">Short...</div>
> ```
> 
> So probably we have to change the `textContent` only if text provided via `fdTruncateTargetText` input property. Otherwise we should not change the content to don't lose the content provided by the user. What do you think?
> 
> After all looks like we could have two cases of directive usage
> 
> * `fdTruncateTargetText` + `fdTruncateChars` or `fdtruncatewudth` = alter the content.
> * `fdTruncateWidth` = set styles, **don't** alter the content.

There is a caveat in the documentation about how the developer should not attempt to change the text if they are using `fdTruncateChars`. I am not aware of a way to tell the truncate directive to watch for any changes to text passed within its curly brackets etc like so: 

```
        <p fd-truncate [fdTruncateState]="true" [fdTruncateChars]="45">
            {{ changeMe }}
        </p>
        <button (click)="changeText()">This button should change the above text to "Hello World"</button>
```

 If we need this then perhaps `fdTruncateChars` would work better as a pipe. I'm starting to think in 99% of cases it will be best for application developers to perform truncation on their end without the need of a pipe or a directive provided by our library.

platon-rov:
@mikerodonnell89 are we going to remove the directive at all or to keep pipe for truncating text? or any other options.

cc @droshev

mikerodonnell89:
> @mikerodonnell89 are we going to remove the directive at all or to keep pipe for truncating text? or any other options.
> 
> cc @droshev

Yep, pipe for text and directive for px

platon-rov:
is it ready for review? otherwise let's mark it as draft until it's ready.

shrvr:
> is it ready for review? otherwise let's mark it as draft until it's ready.

It's ready. Just need to fix the test fail.

apps/docs/src/app/core/component-docs/truncate/examples/truncate-example.component.ts
```diff
@@ -0,0 +1,18 @@
+import { Component } from '@angular/core';
```
platon-rov, line 1: shouldn't all the files related to the truncate directive be removed? as now we have pipe that we don't expose to the customers

shrvr: Directive is still there truncating elements by width. So it is example for that. If that is the case for pipe, I can remove pipe-examples.

platon-rov (approved):

InnaAtanasova (approved):

mikerodonnell89 (approved):
Good to merge 👍

PR #8326, opened 2022-07-06
fix(core): switching tabs issue

platon-rov:
## Related Issue(s)

Relates to Kafka.

## Description

Tabs switching issue when component inside flexible column layout component.

_Apply this patch file to change example to see the issue is actually fixed [tabs.txt](https://github.com/SAP/fundamental-ngx/files/9052770/tabs.txt)_

### Before:
image

### After:

image

### DIFF
--- a/libs/core/src/lib/tabs/tab-panel/tab-panel.component.ts
+++ b/libs/core/src/lib/tabs/tab-panel/tab-panel.component.ts
@@ -125,8 +125,14 @@ export class TabPanelComponent implements OnChanges {
     _expand(expanded: boolean): void {
         if (this._expanded !== expanded) {
             this._expanded = expanded;
-            expanded ? this.opened.emit() : this.closed.emit();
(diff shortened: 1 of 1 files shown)


libs/core/src/lib/tabs/tab-panel/tab-panel.component.ts
```diff
@@ -125,8 +125,14 @@ export class TabPanelComponent implements OnChanges {
...
             this._expanded = expanded;
-            expanded ? this.opened.emit() : this.closed.emit();
-            this._changeDetRef.markForCheck();
+
+            if (expanded) {
+                this.opened.emit();
+            } else {
+                this.closed.emit();
+            }
+
```
g-cheishvili, line 134:
```suggestion
expanded ? this.opened.emit() : this.closed.emit();
```
Why touch this? 😄

g-cheishvili (approved):

PR #8347, opened 2022-07-12
feat(core,docs): Shellbar enhancements

### LINKED ISSUE #7966: Shellbar enhancements
enhancement

- According to the design Overflow menu is carried inside a Popover. Popover can be responsive (on mobile become Dialog) or not. so we need to offer that through the Shellbar/overflow API

shrvr:
## Related Issue(s)

closes #7966

## Description

1. Added toggling `combobox`.
2. Enabled combobox (full-width) for `'s'` size.
3. Enabled responsiveness for shellbar and added example.
4. Introduced `select` component to shellbar and added example.


CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
 
+### [0.35.2-rc.31](https://github.com/SAP/fundamental-ngx/compare/v0.35.2-rc.30...v0.35.2-rc.31) (2022-07-19)
+
+
+### Bug Fixes
+
+* **core:** fixed card layout enhancements ([#8391](https://github.com/SAP/fundamental-ngx/issues/8391)) ([5896e29](https://github.com/SAP/fundamental-ngx/commit/5896e29409fb2ed6bee00e1eb665bf868a4f2966))
```
mikerodonnell89, line 7: Was this intended to be committed?

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -335,6 +336,14 @@ export class ComboboxComponent
      */
     inShellbar = false;
 
+    fullWidth = false;
+
+    @Input()
+    showInput = true;
```
mikerodonnell89, line 342: These do not need to be inputs and should be annotated with `/** @hidden */`

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -335,6 +336,14 @@ export class ComboboxComponent
...
+
+    @Input()
+    disableHideShowOfInput = false;
```
mikerodonnell89, line 345: Same as above

libs/core/src/lib/fixed-card-layout/fixed-card-layout.component.ts
```diff
@@ -24,7 +24,7 @@ import {
 import { FocusKeyManager } from '@angular/cdk/a11y';
 import { CdkDrag, CdkDragDrop, CdkDragEnter } from '@angular/cdk/drag-drop';
 import { Subject, Subscription } from 'rxjs';
-import { debounceTime, delay, distinct, takeUntil } from 'rxjs/operators';
+import { debounceTime, distinct, filter, takeUntil } from 'rxjs/operators';
```
mikerodonnell89, line 27: Unintended changes on this file

mikerodonnell89 (changes requested):
See my comments above ^

libs/core/src/lib/fixed-card-layout/fixed-card-layout.component.ts
shrvr: Due to the merge did earlier, other commits were appended. I made undo of that merge.

CHANGELOG.md
shrvr: Due to the merge did earlier, other commits were appended. I made undo of that merge.

libs/core/src/lib/combobox/combobox.component.ts
shrvr: Right. Actually, according to design we should  allow user to enable/disable the hiding behaviour and to show directly opened/closed input. For that I should add inputs to shellbar component.

InnaAtanasova (changes requested):
Looking good. Could you add an example with Select inside the ShellBar? The specs specify that either Search Filed or Select could be used and it's good to have both so we can detect bugs in future. Thanks

InnaAtanasova:
Also, FYI please check https://github.com/SAP/fundamental-ngx/issues/8607
The search field should open in a new full screen dialog in which items can be selected from a list of suggestions. But let's leave it out of the scope of this PR.

mikerodonnell89:
No margin between these two elements: 
Screen Shot 2022-08-31 at 2 25 35 PM

If the small size combobox is open and the screen size changes (this could happen when the user resizes the page or something), combobox still takes the full width
Screen Shot 2022-08-31 at 2 26 04 PM

libs/core/src/lib/select/select.component.scss
```diff
@@ -20,3 +20,9 @@
 .fd-popover-custom-select-body {
     max-width: 100%;
 }
+
+.fd-select_text-content-inShellbar {
```
mikerodonnell89, line 24: Let's not create a new class for this but instead add these styles to `fd-shellbar__input-group-input--select` via fundamental-style

mikerodonnell89 (changes requested):
The normal select arrow control button has some margin-left between the text and the bottom that seems to be missing here:
Screen Shot 2022-08-31 at 2 52 07 PM

libs/core/src/lib/select/select.component.scss
shrvr: https://github.com/SAP/fundamental-styles/pull/3843

libs/core/src/lib/input-group/input-group.component.html
```diff
@@ -26,7 +26,7 @@
                 fd-button
                 class="fd-input-group__button"
                 type="button"
-                [fdType]="inShellbar ? 'standard' : 'transparent'"
+                [fdType]="'transparent'"
```
platon-rov, line 29: `fdType="transparent"`

libs/core/src/lib/shellbar/shellbar-actions-mobile/shellbar-actions-mobile.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="fd-shellbar-collapse">
-    <fd-action-sheet placement="bottom-end">
+    <fd-action-sheet *ngIf="!isComboboxShowed" placement="bottom-end">
```
platon-rov, line 2: `isComboboxShown`

platon-rov (commented):
when hiding search field with "Cancel" button focus disappears

libs/core/src/lib/input-group/input-group.component.html
shrvr: It search for `transparent`  property in component. `[fdType]="'transparent'"` will pass as a value to `fdType`.

platon-rov: it won't. if you want to pass `transparent` value from the component - use it without `''`.

libs/core/src/lib/input-group/input-group.component.ts
```diff
@@ -181,6 +181,10 @@ export class InputGroupComponent implements ControlValueAccessor, AfterViewInit,
     @ContentChild(InputGroupAddOnDirective)
     addOnElement: InputGroupAddOnDirective;
 
+    /** @hidden */
+    @ViewChild(InputGroupAddOnDirective)
+    localAddOnElement: InputGroupAddOnDirective;
```
platon-rov, line 186: should start with `_`.

libs/core/src/lib/input-group/input-group.component.ts
```diff
@@ -278,10 +287,16 @@ export class InputGroupComponent implements ControlValueAccessor, AfterViewInit,
...
+    _setShowInput(value: boolean): void {
+        this._showInput = value;
+        this._changeDetectorRef.detectChanges();
+    }
+
     /** @hidden */
     setInShellbar(value: boolean): void {
         this.inShellbar = value;
-
+        this.localAddOnElement.elementRef().nativeElement.classList.add('fd-input-group__addon--inShelbar');
```
platon-rov, line 299: please don't mix `camelCase` and `dash-case`.

libs/core/src/lib/select/select.component.ts
```diff
@@ -260,6 +260,8 @@ export class SelectComponent
     /** @hidden */
     _selectionModel: SelectionModel<OptionComponent>;
 
+    inShellbar = false;
```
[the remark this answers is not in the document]
platon-rov: also should start with `_`.

platon-rov (changes requested):
a lot of components changed from the point if they placed inside a shellbar or not. it shouldn't be the case i think, we cannot have proper components composition if they will be assuming that comp A placed inside a comp B or vice versa, they should behave independently. probably to be discussed on the call.

platon-rov (commented):
In my opinion we should not modify the combobox, select and other components from the point if they are placed inside of certain component or not. They are independent components and should exist this way.

In this exact case they are kinda "controlled components" but any manipulations should exist in the controller.

For example we can query with `@ViewChild(...)` in Shellbar and then somehow manipulate components like select, combobox and others. The main thing here is to do it via the public API, but not calling the private methods to not break the encapsulation. We even can  assign classes to "child" components from Shellbar, but still - using public API. If we need to assign class to the select's button - okay, let's introduce a new input in select but it can be used without shellbar, users may also want to add some classes to that button, it's okay.

libs/docs/core/shellbar/examples/shellbar-select-example.component.html
```diff
@@ -0,0 +1,31 @@
+<div>
```
mikerodonnell89, line 1: Why the div?

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -275,10 +277,14 @@ export class ComboboxComponent
     @Output()
     readonly itemClicked: EventEmitter<ComboboxItem> = new EventEmitter<ComboboxItem>();
 
-    /** Event emitted, when the combobox's popover body is opened or closed */
+    /** Event emitted, when the combobox's popovprimaryButtonClickeder body is opened or closed */
```
mikerodonnell89, line 280: Typo here

mikerodonnell89 (changes requested):
Looking good! Two minor comments, and this isn't building but probably just a CI glitch

libs/docs/core/shellbar/examples/shellbar-select-example.component.html
shrvr: oh yeah, it might left when removed size segmented buttons.

libs/core/src/lib/shellbar/shellbar.component.ts
```diff
@@ -26,45 +31,154 @@ export type ShellbarSizes = 's' | 'm' | 'l' | 'xl';
...
+        this._applyShellbarModeToSelect();
+    }
+
+    /** to apply view change on select menu open */
+    handleSelectIsOpenChange(): void {
+        this._subscriptions.add(
+            this.selectComponent.isOpenChange.subscribe((isOpen) => {
+                if (isOpen) {
+                    setTimeout(() => {
+                        this.selectComponent._controlElementRef.nativeElement.classList.remove('is-expanded');
```
platon-rov, line 101: could you instead of manipulating native element classes add input properties to the select\combobox? i.e. `inputClass`, `buttonClass`, etc

The same for methods, instead of focusing by `nativeElement.focus()` just add `focus` method to combobox\select, so in case the inner structure of those components got changed you don't have to change everything outside of them.

mikerodonnell89: Inputs would be exposed to the app developers, not sure that's what we want, unless you mean not inputs but just hidden properties on select/combobox

libs/core/src/lib/combobox/combobox.component.html
```diff
@@ -89,17 +89,17 @@
         />
         <span
             class="fd-input-group__addon fd-input-group__addon--button"
+            [ngClass]="addOnClass"
```
platon-rov, line 92: `[class]` is preferable over `[ngClass]`

libs/core/src/lib/button/button.component.ts
```diff
@@ -146,4 +146,8 @@ export class ButtonComponent extends BaseButton implements OnChanges, CssClassBu
     detectChanges(): void {
         this._changeDetectorRef.detectChanges();
     }
+
+    focus(): void {
```
platon-rov, line 150: typedocs

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -348,6 +352,10 @@ export class ComboboxComponent
     /** @hidden */
     clearInputBtnFocused = false;
 
+    /** Can be used to add class to respective element */
+    addOnClass = '';
+    buttonClass = '';
```
platon-rov, line 357: is it a public api? then we have to write typedocs for each property, method, otherwise `@hidden` and `_` prefis are needed. because we have tool that generates documentation from the comments in the code.

libs/core/src/lib/combobox/combobox.component.html
shrvr: Hm, but not able to get the result with [class].

libs/core/src/lib/combobox/combobox.component.ts
shrvr: yeah, it's public api. Did the changes.

libs/core/src/lib/combobox/combobox.component.html
```diff
@@ -89,17 +89,17 @@
...
             [class.fd-input-group__addon--compact]="_contentDensityObserver.isCompact$ | async"
-            [class.fd-shellbar__input-group-addon]="inShellbar"
-            *ngIf="isSearch && showClearButton && inputText && inputText.length > 0"
+            [style.display]="isSearch && showClearButton && inputText && inputText.length > 0 ? '' : 'none'"
```
platon-rov, line 94: ternary here resolves to `''`, what this stands for?

platon-rov (commented):
Please add typedocs to new class fields and name internal (which are not exposed to users) fields with `_` prefix.

libs/core/src/lib/combobox/combobox.component.html
InnaAtanasova: @shrvr ?

shrvr: Hm, it's to set display to `none` only when condition is false. `''` won't add display property at all in case of true condition. Otherwise, I can add `'block'` instead `''`. @platon-rov @InnaAtanasova

libs/core/src/lib/combobox/combobox.component.ts
mikerodonnell89: @shrvr is there a reason we want this to be public? If all of this should happen under the hood, it would be better to hide this from the API. And using `@Hostbinding` would be more succinct than using public properties and the `_applyShellbarMode` functions

mikerodonnell89 (commented):
Looking good - three things:

- [ ] see my comment about the custom classes being public API
- [ ] i think we will need some changes on the styles repo to address these visual issues with the focus outline:
Screen Shot 2022-11-04 at 10 29 31 AM
Screen Shot 2022-11-04 at 10 29 12 AM

- [ ] When resizing the responsive shellbar, the combobox popover is opened automatically when moving to small mode

https://user-images.githubusercontent.com/2471874/200030076-13cc61bb-d876-43de-945c-0ac8a6591a82.mov

shrvr:
https://github.com/SAP/fundamental-ngx/pull/8347#discussion_r1014250538

@mikerodonnell89 
Hm, no. I agree that it should be under the hood and hidden from API. Will change that.

But I don't think `@HostBinding` can be used here, because we are applying css classes to different elements of the components e.g. `button, input, etc.`.

mikerodonnell89:
> [#8347 (comment)](https://github.com/SAP/fundamental-ngx/pull/8347#discussion_r1014250538)
> 
> @mikerodonnell89 Hm, no. I agree that it should be under the hood and hidden from API. Will change that.
> 
> But I don't think `@HostBinding` can be used here, because we are applying css classes to different elements of the components e.g. `button, input, etc.`.

Oops yes you are right, let's use hidden variables.

libs/core/src/lib/shellbar/shellbar-actions/shellbar-actions.component.scss
```diff
@@ -0,0 +1,13 @@
+.fd-shellbar-actions-combobox--mobile {
```
mikerodonnell89, line 1: Would be nice to have in the css repo

libs/core/src/lib/shellbar/shellbar-actions/shellbar-actions.component.ts
```diff
@@ -67,6 +73,14 @@ export class ShellbarActionsComponent {
     @Input()
     collapsedItemMenuLabel: string;
 
+    /** toggles combobox. Shows combobox if true.*/
```
mikerodonnell89, line 76:
```suggestion
    /** @hidden toggles combobox. Shows combobox if true.*/
```

libs/core/src/lib/shellbar/shellbar-actions/shellbar-actions.component.scss
shrvr: https://github.com/SAP/fundamental-styles/pull/4012
@mikerodonnell89

libs/core/src/lib/shellbar/shellbar.component.ts
```diff
@@ -26,45 +33,166 @@ export type ShellbarSizes = 's' | 'm' | 'l' | 'xl';
...
     @Input()
     size: ShellbarSizes = 'm';
 
+    /** It makes shellbar responsive */
+    @Input()
+    responsive = false;
+
+    /** Custom width can be given to 's' , 'm' , 'l' and 'xl' sizes.  */
+    @Input()
+    sizesWidth = {
```
platon-rov, line 47:
```suggestion
    breakpoints = {
```

platon-rov (commented):
Can we avoid resizing the search field?

Screen Recording 2022-11-10 at 12.09.48

platon-rov (commented):
On the smallest screens search takes the whole widths and logo goes hidden accordingly to the designs

CleanShot 2022-11-18 at 16 15 07@2x

CleanShot 2022-11-18 at 16 17 08

shrvr:
> Can we avoid resizing the search field?
> 
>  Screen.Recording.2022-11-10.at.12.09.48.mov

One solution to solve this to give full width to combobox. Check https://github.com/SAP/fundamental-ngx/pull/8347/commits/1a6e6599ea1f28c8dc314b7b395dc3638eba1e2f

Before:
msedge_URiGEkIQPB

After:
msedge_bvjEnHVQCj

libs/core/src/lib/shellbar/shellbar.component.ts
```diff
@@ -26,45 +34,186 @@ export type ShellbarSizes = 's' | 'm' | 'l' | 'xl';
...
+    constructor(private _elementRef: ElementRef, private _cdr: ChangeDetectorRef) {}
+
+    /** @hidden */
+    ngOnInit(): void {
+        this._attachResizeListener();
+    }
+
+    /** @hidden */
+    ngOnChanges(): void {
+        this._handleSizeChange();
```
platon-rov, line 94: We should check that `size` property present in `changes` otherwise we should not perform calculations in matter of optimization.

platon-rov (approved):

libs/core/src/lib/shellbar/shellbar-actions/shellbar-actions.component.scss
```diff
@@ -0,0 +1,19 @@
+.fd-shellbar__action-fullwidth {
```
mikerodonnell89, line 1: This should be BEM friendly and also in fundamental-styles but I think let's get this merged and handle it in a separate issue

libs/core/src/lib/shellbar/shellbar-actions/shellbar-actions.component.scss
```diff
@@ -0,0 +1,19 @@
...
+}
+
+.fd-shellbar__action-combobox {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 100%;
+}
+
+.fd-combobox-custom-class {
```
mikerodonnell89, line 17: I think this is unused

mikerodonnell89 (changes requested):
The select is not automatically being added to the overflow menu on resize for the responsive shellbar. And, when it does get collapsed, selecting it from the overflow menu should open the select in mobile mode like this: https://sap.github.io/fundamental-ngx/#/core/select#mobile

droshev (changes requested):
let's handover the PR to one of the team members to ensure proper transition

N1XUS:
Closing in favour of #9287

Issue #8342, opened 2022-07-11
Defect Hunting for Sprint 92


droshev:

mikerodonnell89:
- [ ] This should bring up a full-screen dialog menu. Buttons should also be cozy
<img width="468" alt="Screen Shot 2022-07-11 at 10 26 48 AM" src="https://user-images.githubusercontent.com/2471874/178312353-7d63fb92-b6af-4048-93e8-41e8d8e8bda8.png">

InnaAtanasova:
- [ ] Alert - the alignment for the success alert is not correct
<img width="1370" alt="Screen Shot 2022-07-11 at 2 35 02 PM" src="https://user-images.githubusercontent.com/39598672/178334428-e3e78b95-f33d-45b8-bbac-17d450521da8.png">

InnaAtanasova:
- [ ] Content density
When content density is changed to `condensed` the changes are applied to all elements (in this case it's Bar component). Condensed is only applied to Table.
<img width="1627" alt="Screen Shot 2022-07-11 at 2 47 26 PM" src="https://user-images.githubusercontent.com/39598672/178336358-37281778-8bb0-4553-b98b-a1a305339ce1.png">

mikerodonnell89:
> * [ ]  Content density
>   When content density is changed to `condensed` the changes are applied to all elements (in this case it's Bar component). Condensed is only applied to Table.

@InnaAtanasova The think the idea here is that condensed is the smallest of the 3 visually so it is assumed that if you want globally condensed visuals, you'll want compact not cozy for the components that do have a compact size but not condensed.

InnaAtanasova:
> > * [ ]   Content density
> >   When content density is changed to `condensed` the changes are applied to all elements (in this case it's Bar component). Condensed is only applied to Table.
> 
> @InnaAtanasova The think the idea here is that condensed is the smallest of the 3 visually so it is assumed that if you want globally condensed visuals, you'll want compact not cozy for the components that do have a compact size but not condensed.

I don't think so. Condensed is specific for Table and is not intended to be used separately. Here's a screenshot from the UX guidelines:
<img width="1208" alt="Screen Shot 2022-07-11 at 2 57 12 PM" src="https://user-images.githubusercontent.com/39598672/178338305-6e061a1c-4766-4143-bf15-7ed1ee0a667b.png">

And a link to the spec:
https://experience.sap.com/fiori-design-web/tree-table/

[...]

PR #8351, opened 2022-07-12
docs(core): fix text issue with alert from component

mikerodonnell89:
part of https://github.com/SAP/fundamental-ngx/issues/8342

before:
Screen Shot 2022-07-11 at 2 35 02 PM

after:
Screen Shot 2022-07-12 at 11 29 42 AM


g-cheishvili (approved):
It's cool and all, but is not alert deprecated?

N1XUS (commented):
@mikerodonnell89 I think putting `div` inside `p` is not a good option here. I would also update alert component itself so instead of `<p class="fd-alert__text">` we would have `<div class="fd-alert__text">`.

PR #8353, opened 2022-07-12
fix(core): remove breadcrumb hidden item component

mikerodonnell89:
part of #8342 

Two issues - one was incorrect focus outline and two was an accessibility problem.  The cause was an Angular wrapper element `fd-breadcrumb-hidden-item` getting between the `<ul>` and the `<li>`'s when generating the breadcrumb overflow menu. According to the html5 spec, `li`'s must always be direct children of `ul`'s. This PR removes the extra component and moves the "hidden breadcrumb item" logic back in to the breadcrumb and breadcrumb item

before:
Screen Shot 2022-07-12 at 2 33 34 PM

after:
Screen Shot 2022-07-12 at 2 28 34 PM


g-cheishvili (approved):

N1XUS (dismissed):
image
@mikerodonnell89 please also fix correct attribute name for `role` (it's in menu-item.component.ts)

InnaAtanasova (approved):

PR #8354, opened 2022-07-12
fix(core): small bug with combobox border radius

mikerodonnell89:
part of #8342 

before:
image

after:
Screen Shot 2022-07-12 at 3 24 21 PM


g-cheishvili (approved):

N1XUS (commented):
@mikerodonnell89 if I'm not mistaken, we agreed with the designers that we shouldn't round the borders for such elements.

PR #8355, opened 2022-07-12
fix(core): select options should be li elements (border radius, accessibility issue)

mikerodonnell89:
part of #8342 

Similar to https://github.com/SAP/fundamental-ngx/pull/8353

Changes `fd-option` to an attribute selector. All immediate children of `ul` elements should be `li`s. This is a breaking change to fd-option, however I have left the old selector in place so no one's selects will fail entirely when this is merged.

before
Screen Shot 2022-07-12 at 3 39 06 PM

after
Screen Shot 2022-07-12 at 3 48 13 PM


N1XUS (changes requested):
@mikerodonnell89 please check failing e2e tests

InnaAtanasova (approved):
Code looks good, please check the failing tests

PR #8394, opened 2022-07-19
feat(core, platform): datepicker today button

### LINKED ISSUE #8362: date picker enhancement - bring Today button
enhancement

We got feedback and requests to bring the Today button like it is in styles (core and platform)
https://fundamental-styles.netlify.app/?path=/docs/patterns-date-picker--default-and-compact-sizes#today-selection-button

shrvr:
## Related Issue(s)

closes #8362 

## Description

Added Today Selection Button in the footer selects today's date in the system or user-preferred timezone and closes the DatePicker popover.

`Important:`
For date range it sets start and end both to today's date.

### DIFF
--- a/libs/core/src/lib/date-picker/date-picker.component.html
+++ b/libs/core/src/lib/date-picker/date-picker.component.html
@@ -75,7 +75,7 @@
         <div *ngIf="showTodayButton" fd-bar barDesign="footer">
             <div fd-bar-right>
                 <fd-bar-element>
-                    <button fd-button label="Today" (click)="onTodayButtonClick()"></button>
+                    <button fd-button [label]="todayButtonLabel" (click)="onTodayButtonClick()"></button>
                 </fd-bar-element>
             </div>
         </div>
--- a/libs/core/src/lib/date-picker/date-picker.component.html
+++ b/libs/core/src/lib/date-picker/date-picker.component.html
@@ -72,7 +72,7 @@
             [nextButtonDisableFunction]="nextButtonDisableFunction"
         ></fd-calendar>
 
-        <div fd-bar barDesign="footer">
+        <div *ngIf="showTodayButton" fd-bar barDesign="footer">
             <div fd-bar-right>
                 <fd-bar-element>
                     <button fd-button label="Today" (click)="onTodayButtonClick()"></button>
--- a/libs/core/src/lib/date-picker/date-picker.component.html
+++ b/libs/core/src/lib/date-picker/date-picker.component.html
@@ -71,6 +71,14 @@
             [previousButtonDisableFunction]="previousButtonDisableFunction"
             [nextButtonDisableFunction]="nextButtonDisableFunction"
         ></fd-calendar>
+
+        <div fd-bar barDesign="footer">
+            <div fd-bar-right>
+                <fd-bar-element>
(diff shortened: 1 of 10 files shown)


libs/core/src/lib/date-picker/date-picker.component.html
```diff
@@ -71,6 +71,14 @@
             [previousButtonDisableFunction]="previousButtonDisableFunction"
             [nextButtonDisableFunction]="nextButtonDisableFunction"
         ></fd-calendar>
+
+        <div fd-bar barDesign="footer">
```
mikerodonnell89, line 75: I think it should be conditional, so we will need an `*ngIf` here, and a boolean input in `date-picker.component.ts`, something like `showTodayButton`? Probably default to false as well.  So in addition, another example showing the new input in action

libs/core/src/lib/date-picker/date-picker.component.ts
```diff
@@ -529,6 +529,18 @@ export class DatePickerComponent<D>
         }
     }
 
+    onTodayButtonClick(): void {
+        const todayDate = this._dateTimeAdapter.today();
+        const todayDateRange: DateRange<D> = { start: todayDate, end: todayDate };
```
mikerodonnell89, line 534: It is not necessary to create both these variables here, since we'll only end up using one of them

mikerodonnell89 (changes requested):
Looks pretty good so far! Left a couple comments

libs/core/src/lib/date-picker/date-picker.component.html
```diff
@@ -71,6 +71,14 @@
             [previousButtonDisableFunction]="previousButtonDisableFunction"
             [nextButtonDisableFunction]="nextButtonDisableFunction"
         ></fd-calendar>
+
+        <div *ngIf="showTodayButton" fd-bar barDesign="footer">
+            <div fd-bar-right>
+                <fd-bar-element>
+                    <button fd-button label="Today" (click)="onTodayButtonClick()"></button>
```
mikerodonnell89, line 78: We should (for now) provide an input for the label value that defaults to "Today" so the developer can pass their own strings in different languages etc

mikerodonnell89 (changes requested):
We should have a unit test for the new function

libs/core/src/lib/date-picker/date-picker.component.ts
```diff
@@ -202,6 +202,13 @@ export class DatePickerComponent<D>
     @Input()
     closeOnDateChoose = true;
 
+    /** Enables Today date selection button if true */
+    @Input()
+    showTodayButton = false;
+
+    @Input()
```
platon-rov, line 209: typedoc

libs/core/src/lib/date-picker/date-picker.component.ts
```diff
@@ -529,6 +536,16 @@ export class DatePickerComponent<D>
         }
     }
 
+    onTodayButtonClick(): void {
```
platon-rov, line 539: typedoc

g-cheishvili (approved):
looks good to me

mikerodonnell89 (approved):

platon-rov (approved):

PR #8458, opened 2022-07-26
fix(core): action-sheet focus bug

### LINKED ISSUE #8344: Defect Hunting: Action Sheet focus bugs
bug/enhancement

- [ ] In mobile view Action sheet is open in a full screen menu. One can navigate from element to element using the tab key. However, once reaching the last element (Cancel button), the focus is lost and it goes to the first open browser tab. 

Expected behaviour: the focus should be kept in the Action Sheet

- [ ] Action sheet (core): strange keyboard behavior. Open an action sheet popover via mouse click on the control. Note that the first item in the list is focused. Press tab twice to move focus to the 3rd item. Press the down arrow key and the 2nd item will be focused

shrvr:
## Related Issue(s)

closes #8344 

## Description

1. Added tab key handling.
2. Solved focus-lost on close for Mobile view.

### DIFF
--- a/libs/core/src/lib/action-sheet/action-sheet-body/action-sheet-body.component.ts
+++ b/libs/core/src/lib/action-sheet/action-sheet-body/action-sheet-body.component.ts
@@ -13,7 +13,7 @@ import { Nullable } from '@fundamental-ngx/core/shared';
 
 import { ActionSheetItemComponent } from '../action-sheet-item/action-sheet-item.component';
 import { ContentDensityObserver, contentDensityObserverProviders } from '@fundamental-ngx/core/content-density';
-import { TAB } from '@angular/cdk/keycodes';
+import { hasModifierKey, TAB } from '@angular/cdk/keycodes';
 
 let actionSheetBodyUniqueIdCounter = 0;
 
@@ -72,15 +72,16 @@ export class ActionSheetBodyComponent {
     @HostListener('keydown', ['$event'])
     keyDownHandler(event: KeyboardEvent): void {
         if (this._keyboardSupportService.keyManager) {
-            this._keyboardSupportService.onKeyDown(event);
-
-            if (event.shiftKey && KeyUtil.isKeyCode(event, TAB)) {
+            if (KeyUtil.isKeyCode(event, TAB)) {
                 event.preventDefault();
-                this._keyboardSupportService.keyManager?.setPreviousItemActive();
-            } else if (KeyUtil.isKeyCode(event, TAB)) {
-                event.preventDefault();
-                this._keyboardSupportService.keyManager?.setNextItemActive();
+                if (hasModifierKey(event, 'shiftKey')) {
+                    this._keyboardSupportService.keyManager.setPreviousItemActive();
+                } else {
+                    this._keyboardSupportService.keyManager.setNextItemActive();
+                }
+                return;
             }
+            this._keyboardSupportService.onKeyDown(event);
         }
     }
 }
--- a/libs/core/src/lib/action-sheet/action-sheet-body/action-sheet-body.component.ts
+++ b/libs/core/src/lib/action-sheet/action-sheet-body/action-sheet-body.component.ts
@@ -8,12 +8,11 @@ import {
     ViewEncapsulation
 } from '@angular/core';
 
-import { KeyboardSupportService, KeyUtil } from '@fundamental-ngx/core/utils';
+import { KeyboardSupportService } from '@fundamental-ngx/core/utils';
 import { Nullable } from '@fundamental-ngx/core/shared';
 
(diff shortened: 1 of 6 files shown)


libs/core/src/lib/action-sheet/action-sheet-body/action-sheet-body.component.ts
```diff
@@ -73,5 +89,17 @@ export class ActionSheetBodyComponent {
         if (this._keyboardSupportService.keyManager) {
             this._keyboardSupportService.onKeyDown(event);
         }
+
+        if (event.shiftKey && KeyUtil.isKeyCode(event, TAB)) {
```
N1XUS, line 93: KeyboardSupportService supports wrapping the keyboard navigation (cycling). You can try to adjust `this._keyboardSupportService.setKeyboardService(this.actionSheetItems, false);` and pass `true` to the second argument in `ActionSheetComponent`

N1XUS (changes requested):

libs/core/src/lib/action-sheet/action-sheet-body/action-sheet-body.component.ts
```diff
@@ -71,6 +72,15 @@ export class ActionSheetBodyComponent {
...
         if (this._keyboardSupportService.keyManager) {
+            if (KeyUtil.isKeyCode(event, TAB)) {
+                event.preventDefault();
+                if (hasModifierKey(event, 'shiftKey')) {
+                    this._keyboardSupportService.keyManager.setPreviousItemActive();
+                } else {
+                    this._keyboardSupportService.keyManager.setNextItemActive();
+                }
+                return;
+            }
```
g-cheishvili, line 83: Move this to the keyboardService.onKeyDown like this:

```ts
        this._keyManager.onKeydown(event);
        if (KeyUtil.isKeyCode(event, TAB)) {
            event.preventDefault();
            hasModifierKey(event, 'shiftKey')
                ? this._keyManager.setPreviousItemActive()
                : this._keyManager.setNextItemActive();
        }
    }
```

This way not only here, but everywhere this issue will be resolved, Problem generally is that on TAB ListKeyManager does not update activeItem, it just relies on native TAB behavior

g-cheishvili (changes requested):

mikerodonnell89 (changes requested):
When I first open the mobile action sheet and press tab, the 2nd action is focused and not the first

libs/core/src/lib/action-sheet/action-sheet.component.ts
```diff
@@ -171,6 +171,8 @@ export class ActionSheetComponent implements AfterContentInit, AfterViewInit, On
 
         if (isOpen) {
             this._setItemActive(0);
+        } else {
+            this.actionSheetControl._focus();
```
g-cheishvili, line 175: I would go for ternary, instead of if/else

libs/core/src/lib/action-sheet/action-sheet-control/action-sheet-control.component.ts
```diff
@@ -36,4 +37,12 @@ export class ActionSheetControlComponent {
     onClick(): void {
         this.clicked.emit();
     }
+
+    /** @hidden */
+    _focus(): void {
+        this._elementRef.nativeElement.firstChild.focus();
```
platon-rov, line 43: actually there are no limitations for the action sheet control content hence `firstChild` may be the not focusable element.

shrvr: @platon-rov  added logic to make element focusable if not.

libs/core/src/lib/action-sheet/action-sheet-control/action-sheet-control.component.ts
```diff
@@ -36,4 +37,16 @@ export class ActionSheetControlComponent {
     onClick(): void {
         this.clicked.emit();
     }
+
+    /** @hidden */
+    _focus(): void {
+        const elemChild = this._elementRef.nativeElement.firstChild;
+        if (elemChild && elemChild.getAttribute('tabindex') !== '-1') {
```
platon-rov, line 44: it will return `null` for buttons and other elements that are focusable by default and don't need `tabindex` to be explicitly added.

platon-rov (changes requested):

platon-rov (commented):
the first item in the list doesn't get focused after action sheet dialog opens, is it the intended behavior?

shrvr:
> the first item in the list doesn't get focused after action sheet dialog opens, is it the intended behavior?

Yes, to give clear visual appearance in mobile mode. Users can focus first element using tab key if they use keyboard.
Let me know your thoughts.

platon-rov (approved):

mikerodonnell89 (approved):

PR #8471, opened 2022-07-28
feat(core): combobox dropdown values list can containt duplicate display names

### LINKED ISSUE #8466: Combobox: Duplicate labels forces unwanted selection of first duplicate
bug

Issue: when duplicate labels in Combobox's dropdown exist, only the first of the duplicates can be selected
Expected: selected item from dropdown is selected

Screenshot 2022-07-28 at 16 39 51

[Current implementation](https://github.com/SAP/fundamental-ngx/blob/dbc852dda6f65bbe3268e7617034793c219033da/libs/core/src/lib/combobox/combobox.component.ts#L445) always selects the first matching label:

```typescript
    /** @hidden */
    onMenuClickHandler(value: any): void {
        if (value || value === 0) {
            const index: number = this.dropdownValues.findIndex((_value) => _value === value);  // <== selects first match
            this._handleClickActions(value);
            this.filterHighlight = false;
            this.itemClicked.emit({ item: value, index });
        }
    }
```

v0.35.3

I made this issue reproducible in the following stackblitz:
https://stackblitz.com/edit/angular-28nbwn?file=src%2Fapp%2Fcombobox-forms-example.component.html,src%2Fapp%2Fcombobox-forms-example.component.ts,src%2Fapp%2Fcombobox-forms-example.component.scss

it is not possible to select b5a!

Regardless of which "Tom Sawyer" you choose from the dropdown list => always "e2f" is selected.

g-cheishvili:
## Related Issue(s)

closes #8466

## Description
Now, instead of only saving the display name and then trying to guess to which one it belonged originally, we save full value and then use it for the output.

### DIFF
--- a/libs/core/src/lib/combobox/combobox.component.ts
+++ b/libs/core/src/lib/combobox/combobox.component.ts
@@ -611,6 +611,11 @@ export class ComboboxComponent
         this.clearInputBtnFocused = false;
     }
 
+    /** Current select value */
+    getValue(): any {
+        return this._value;
+    }
+
     /** Method that picks other value moved from current one by offset, called only when combobox is closed */
     private _chooseOtherItem(offset: number): void {
         const activeValue: any = this._getOptionObjectByDisplayedValue(this.inputTextValue);
@@ -694,16 +699,13 @@ export class ComboboxComponent
         this.onChange(this.getValue());
     }
 
+    /** @hidden */
     private setValue(value: any): void {
         if (this.communicateByObject) {
             this._value = value;
         } else {
             this._value = this.displayFn(value);
         }
-    }
-
-    getValue(): any {
-        return this._value;
     }
 
     /** @hidden */
--- a/libs/core/src/lib/combobox/combobox.component.ts
+++ b/libs/core/src/lib/combobox/combobox.component.ts
@@ -348,6 +348,9 @@ export class ComboboxComponent
     private _subscriptions = new Subscription();
 
     /** @hidden */
+    private _value: any;
+
+    /** @hidden */
     onChange: (value: any) => void = () => {};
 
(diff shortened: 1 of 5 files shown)


mikerodonnell89 (approved):
LGTM, possibly related but I am confused about the purpose of this example (already existing)
Screen Shot 2022-07-28 at 4 50 55 PM

N1XUS (approved):

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -709,4 +723,9 @@ export class ComboboxComponent
             injector
         );
     }
+
+    isSelected(term: any): boolean {
```
N1XUS, line 727: Typedoc

droshev (approved):

g-cheishvili:
> LGTM, possibly related but I am confused about the purpose of this example (already existing)
> Screen Shot 2022-07-28 at 4 50 55 PM
> 

I added that double example for validation. Now selection is only one of them, even though they have the same name

platon-rov (approved):

mikerodonnell89:
> > LGTM, possibly related but I am confused about the purpose of this example (already existing)
> > Screen Shot 2022-07-28 at 4 50 55 PM
> 
> I added that double example for validation. Now selection is only one of them, even though they have the same name

Right I get why you added the double list item, just wondering why this "communication via displayed value" is an example in the first place. Not sure when that'd be a desirable use case

desaroxx:
@mikerodonnell89 :

> Not sure when that'd be a desirable use case

I am the guy requesting this to be possible. We were facing issues with this component when populating people's names in the dropdown. Names look same, but different person representing each row.

Issue #8457, opened 2022-07-26
Theming Bug


droshev:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.

- [ ] SAP Icons are not loaded.

1. create a new angular app
2. add `@fundamental-ngx/core`
3. import `ButtonModule`
4. add a button with an icon
```
<button fd-button fdType="transparent" glyph="menu2" ariaLabel="Menu Transparent" title="Menu Transparent"></button>
```
The icons are not rendered

- [ ] Another issue which could be related to that is: changing the theme from Fiori 3 to Horizon doesn't change the icon file.

- [ ] Can we add an example how to set a (random) theme by default in an application (not through `ng add` schematics but through the API)?

N1XUS:
1. Icons are not loaded since they are loaded with ThemingService (from ThemingModule), which we aren't importing to the user's application, this needs to be done by the user manually. Added info message during schematics execution.
2. Depends on which theming provider they are using (deprecated one, or a new one). Old one will not import sap icons since this feature is a new one, and the service is deprecated now.
3. We have everything documented here for a manual process https://sap.github.io/fundamental-ngx/#/core/theming

Issue #8350, opened 2022-07-12
bug(core, platform): lib's dependencies & theming service issues


platon-rov:
## Description

There exist several issues with lib's dependencies and because of them result that we see using ThemingService may be unexpected.

### _We copy font stylesheets during `ng add @fundamental-ngx/core` installation process._
  That's not needed, we could keep font stylesheets in our library then copy it to the assets of the host application. 
  To do that copying we have to add `./node_modules/@fundamental-ngx/core/.../fonts` to the `styles` array in `angular.json` of the host app. That can be done using schematics.
### _We add `fundamental-styles` to the host app's dependencies during `ng add @fundamental-ngx/core` installation process._
  That's not needed because we have `fundamental-styles` as the `dependency` of our `@fundamental-ngx/core` library. This action should be removed.
### _We don't have `@fundamental-ngx/core` peer dependency in the package.json in `@fundamental-ngx/platform`._
  Despite we add proper version of `@fundamental-ngx/core` to the host app's `package.json` during the `ng add @fundamental-ngx/core` later, on update, user may forget to bump core's version. We need to avoid this.
### _We rely only on the `./node_modules/(fundamental-styles | @sap-theming/base-content)` paths when adding third party stylesheets to the host app's `styles` array in `angular.json`._
  Currently we may copy stylesheets not from the correct third-party library instance and see incorrect results when load a theme using `ThemingService`.

  Accordingly to "[How npm3 Works](https://npm.github.io/how-npm-works-docs/npm3/how-npm3-works.html)" and since we have these third-party libraries as `dependency` for our `@fundamental-ngx/core` then these third party libraries can be in 

  * `host-app/.node_modules/(fundamental-styles | @sap-theming/base-content)` or 
  * `host-app/.node_modules/@fundamental-ngx/core/.node_modules/(fundamental-styles | @sap-theming/base-content)`
  
   And it may change at any `npm install`.
   
  Only one approach that could be there is to put these dependencies as `peerDependencies` to our `@fundamental-ngx/core`'s `package.json`. There may be issues when customer installs another packages that relies on another versions of `(fundamental-styles | @sap-theming/base-content)` but i cannot image a way to solve it on our side.

PR #8479, opened 2022-07-29
fix(core,platform): schematics, deps, migrations, assets & styles

platon-rov:
Original PR https://github.com/SAP/fundamental-ngx/pull/8465 **`<= Please check comments there also`**

## Related Issue(s)

Closes https://github.com/SAP/fundamental-ngx/issues/8350
Closes https://github.com/SAP/fundamental-ngx/issues/8457
Refers to https://github.com/SAP/fundamental-ngx/pull/8258

## Description

* Library deps were made peerDeps to fix various issues with multiple deps instances.
* Core schematics updated to setup Theming out of the box.
* Core schematics styles and assets adding fixed to have all the files only on the library (our) side.
* Migrations created to remove old styles and assets.

**TDB**, Not done
* Removing the old deprecated Themes.

##### PR Quality

-   [X] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/0.36.0-Breaking-changes#schematics-8465)


libs/core/schematics/migrations/migration-0.36/index.ts
```diff
@@ -0,0 +1,129 @@
...
+
+                    tree.overwrite(styleSheet, contents);
+                    context.logger.info(`✅️ Removed font styles from stylesheets.`);
+                }
+            });
+        }
+
+        context.logger.info(
+            `⚠️ Notice: If you have imported font styles from @sap-theming, please remove it since now it controlled by fundamental-ngx`
+        );
```
g-cheishvili, line 117:
```suggestion
        context.logger.info(
            `⚠️ Notice: If you have imported font styles from @sap-theming, please remove it, since now it is controlled by fundamental-ngx`
        );
```

libs/core/schematics/migrations/migration-0.36/index.ts
```diff
@@ -0,0 +1,129 @@
...
+
+        return tree;
+    };
+}
+
+function noticeAddSchematics(): Rule {
+    return (_: Tree, context: SchematicContext) => {
+        context.logger.info(`ℹ️ Now you have run ng-add schematics once again to setup things in the right way.`);
+
```
g-cheishvili, line 126:
```suggestion
    return (_: Tree, context: SchematicContext) => {
        context.logger.info(`ℹ️ Now you have to run ng-add schematics once again to set up things in the right way.`);

```

libs/core/schematics/utils/file-utils.ts
```diff
@@ -1,10 +1,11 @@
 import { join, Path } from '@angular-devkit/core';
 import { Tree } from '@angular-devkit/schematics';
 
-const STYLESHEET_REGEX = /.*\.(css|scss|less)/;
+const STYLESHEET_REGEX = /.*\.(css|scss)/;
```
g-cheishvili, line 4: What do we have against "sass" and "less" extensions?

package.json
```diff
@@ -73,8 +73,8 @@
     "compile-typedoc-platform": "rm -rf apps/docs/src/assets/typedoc/platform && typedoc --out apps/docs/src/assets/typedoc/platform libs/platform/src/lib --tsconfig libs/platform/src/lib/tsconfig.lib.json --hideGenerator --theme apps/docs/src/fd-typedoc",
     "compile-typedoc-json": "typedoc --json apps/docs/src/assets/typedoc/typedoc.json --hideGenerator --theme apps/docs/src/fd-typedoc libs/core/src/lib",
     "build:schematics": "./node_modules/.bin/tsc -p libs/core/tsconfig.schematics.json && yarn run copy:schemas && yarn run copy:collection",
-    "copy:schemas": "cp libs/core/schematics/ng-add/schema.json dist/libs/core/schematics/ng-add/schema.json && yarn run copy:core-migration-shema",
-    "copy:core-migration-shema": "cp libs/core/schematics/migrations/migrate-theme-fonts/schema.json dist/libs/core/schematics/migrations/migrate-theme-fonts/schema.json",
+    "copy:schemas": "cp libs/core/schematics/ng-add/schema.json dist/libs/core/schematics/ng-add/schema.json && yarn run copy:core-migration-schema",
+    "copy:core-migration-schema": "cp libs/core/schematics/migrations.json dist/libs/core/schematics/migrations.json",
```
g-cheishvili, line 77: Thanks. without bash I will not be able to use these scripts on my machine even more(Windows...)

g-cheishvili (approved):
Only grammatical issues, nothing more

g-cheishvili (changes requested):
See above

package.json
platon-rov: actually I didn't change anything, just file names

platon-rov: in case we have issues with commands we could fix it in separate pr.

libs/core/schematics/utils/file-utils.ts
platon-rov: nothing :D 

fixed

libs/platform/src/lib/package.json
```diff
@@ -19,12 +19,13 @@
     "@angular/animations": "ANGULAR_VER_PLACEHOLDER",
     "@angular/router": "ANGULAR_VER_PLACEHOLDER",
     "@angular/cdk": "ANGULAR_VER_PLACEHOLDER",
-    "rxjs": "RXJS_VER_PLACEHOLDER"
+    "rxjs": "RXJS_VER_PLACEHOLDER",
+    "fundamental-styles": "FDSTYLES_VER_PLACEHOLDER",
```
InnaAtanasova, line 23: do we need the FD prefix?

g-cheishvili: Do not think we need, otherwise prepare executor should be updated as well to properly handle placeholder @platon-rov

libs/core/src/lib/package.json
```diff
@@ -19,16 +22,21 @@
...
-    "fundamental-styles": "FDSTYLES_VER_PLACEHOLDER",
     "@fundamental-ngx/i18n": "VERSION_PLACEHOLDER",
     "fast-deep-equal": "FAST_DEEP_EQUAL_VER_PLACEHOLDER",
     "lodash-es": "LODASH_ES_VER_PLACEHOLDER",
     "tslib": "^2.0.0",
     "compare-versions": "COMPARE_VERSIONS_VER_PLACEHOLDER"
+  },
+  "exports": {
+    ".": {
+      "sass": "./styles/fundamental-ngx-core.scss"
```
InnaAtanasova, line 39: why is this needed?

platon-rov: Currently, we have *raw* global styles `libs/core/schematics/ng-add/styles.ts` (fonts, icons, etc.) which we place on the customer side when installing lib with `ng add`. In case tomorrow we need to change these styles, we have to write migration for it. 

If we only be putting link to styles file from our lib, we can change its contents to whatever we want without need to write migrations etc.

Real world example: 
Tomorrow we will add Skeleton feature and these Skeletons will be rendered over the components when applying CSS classes. To have these CSS classes' values loaded, we either have to render a skeleton component somewhere which import these styles (bad approach) or add these styles to global ones (like fonts) so they will be accessible anywhere once global styles are loaded.

libs/platform/src/lib/package.json
platon-rov: What do you mean? I didn't change anything except moving from deps to peerDeps.

libs/core/schematics/migrations.json
```diff
@@ -0,0 +1,9 @@
+{
+    "schematics": {
+        "migration-0-37": {
+            "version": "0.37",
```
g-cheishvili, line 4: Does not this mean that if user updates from anything below 0.37.0 then this schematic will be called, but we're currently on 0.36.3. Am I getting this right?

platon-rov: This schematics update is a breaking change by itself, once we merge this pr `ng add` will start working differently and that's why I set version here to `0.37`, merge of this pr will promote version to `0.37`

InnaAtanasova (approved):

g-cheishvili (approved):

PR #8489, opened 2022-07-31
chore(platform,docs,core): content density deprecations

g-cheishvili:
## Related Issue(s)

Closes https://github.com/SAP/fundamental-ngx/issues/7765

## Description
- Replaced every usage of deprecated properties from the core and deprecated platform inputs "contentDensity".
- Added warnings both on the docs page and during usage.
- Got rid of prop drilling in various components, such as form-generator, table and etc
- Added support for DI resolved default content density in Core

No breaking changes


InnaAtanasova (approved):

mikerodonnell89 (commented):
`libs/platform/src/lib/table/table.component.ts` contains a mention of the old `contentDensity` input in a comment and `platform-forms/switch/switch-docs.component.html` mentions old `compact` input, think we need to remove these

platon-rov (commented):
Seems like platform smart filter bar still doesn't react to content density changes.

image

g-cheishvili:
> Seems like platform smart filter bar still doesn't react to content density changes.
> 
> image

If you are looking at the first example, it is forced into fdCompact, look in example. 
Regarding the `fd-tokenizer` it is always that height, bug from styles. It is actually `compact` in that image
Screenshot 2022-08-03 152936

platon-rov (approved):

Issue #7965, opened 2022-04-06
Notification as part of (responsive) popover


droshev:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.

According to the design `Notification is carried inside a Popover. Popover can be responsive (on mobile become Dialog) or not.` so we need to offer that through the Notification API

PR #8525, opened 2022-08-09
feat(core): notification mobile (in popover)

platon-rov:
## Related Issue(s)

Closes https://github.com/SAP/fundamental-ngx/issues/7965

## Description

Notification mobile mode (in Popover).

### After:

image

image

### DIFF
--- a/libs/core/src/lib/dialog/dialog.component.scss
+++ b/libs/core/src/lib/dialog/dialog.component.scss
@@ -22,4 +22,9 @@
             border-top-right-radius: 0;
         }
     }
+
+    .fd-dialog__body--no-horizontal-padding {
+        padding-left: 0;
+        padding-right: 0;
+    }
 }
--- a/libs/core/src/lib/notification/notification-group/notification-group.component.ts
+++ b/libs/core/src/lib/notification/notification-group/notification-group.component.ts
@@ -6,10 +6,12 @@ import {
     Input,
     ViewEncapsulation,
     OnChanges,
-    OnInit
+    OnInit,
+    Optional
 } from '@angular/core';
 import { applyCssClass } from '@fundamental-ngx/core/utils';
 import { CssClassBuilder } from '@fundamental-ngx/core/utils';
+import { PopoverComponent } from '@fundamental-ngx/core/popover';
 
 @Component({
     selector: 'fd-notification-group',
@@ -31,7 +33,12 @@ export class NotificationGroupComponent implements OnChanges, OnInit, CssClassBu
     width: string;
 
     /** @hidden */
-    constructor(private _elementRef: ElementRef) {}
+    constructor(private _elementRef: ElementRef, @Optional() private _popover: PopoverComponent) {
+        if (this._popover) {
+            this._popover.focusTrapped = true;
+            this._popover.focusAutoCapture = true;
+        }
+    }
 
     /** @hidden */
     ngOnChanges(): void {
--- a/libs/docs/core/notification/examples/notification-mobile/notification-mobile-example.component.html
+++ b/libs/docs/core/notification/examples/notification-mobile/notification-mobile-example.component.html
@@ -1,0 +1,263 @@
+<h4>Notification In Popover Dialog</h4>
+
+<button fd-button label="Notification" fdCompact [fdPopoverTrigger]="popoverMobile"></button>
+
+<fd-popover #popoverMobile [mobile]="true" [mobileConfig]="mobileConfig">
+    <ng-template #popoverBodyContent>
+        <fd-notification [mobile]="true">
+            <fd-notification-body>
+                <fd-notification-content>
+                    <fd-notification-header>
+                        <div fd-notification-indicator type="information"></div>
+                        <h2 fd-notification-title [unread]="true">Your leave request has been accepted</h2>
+                    </fd-notification-header>
+                    <p fd-notification-paragraph>
+                        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut
+                        labore et dolore magna aliqua.
+                    </p>
+                    <fd-notification-footer>
+                        <span fd-notification-footer-content>Author name</span>
+                        <span fd-notification-separator></span>
+                        <span fd-notification-footer-content>Timestamp</span>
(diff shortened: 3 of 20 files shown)


libs/core/src/lib/notification/notification-group/notification-group.component.ts
```diff
@@ -30,6 +30,10 @@ export class NotificationGroupComponent implements OnChanges, OnInit, CssClassBu
     @Input()
     width: string;
 
+    /** @hidden */
+    @HostBinding('style.height')
+    _height = '100%';
```
mikerodonnell89, line 35: Why as a hostbinding and not just styled in the css (in fundamental-styles)?

mikerodonnell89 (commented):
Looks good, did have one question see above

libs/core/src/lib/notification/notification-group/notification-group.component.ts
platon-rov: moved https://github.com/SAP/fundamental-styles/pull/3783

InnaAtanasova (changes requested):
I find this odd. If I see this in an app I'd immediately assume it's a bug. First of all, the specs say the following: "Notification is carried inside a Popover. Popover can be responsive (on mobile become Dialog) **or not.**"

The `or not` part means we need to have this behaviour as a property so the user can specify if they want the notification to appear in a dialog or keep the normal popover behaviour. 

Second, the box-shadow of the notification that creates the illusion of a border should probably be removed when it's in a responsive popover and I don't see the need of a footer in this case. 

Screen Shot 2022-08-10 at 3 08 52 PM
Screen Shot 2022-08-10 at 3 08 23 PM

platon-rov:
> I find this odd. If I see this in an app I'd immediately assume it's a bug. First of all, the specs say the following: "Notification is carried inside a Popover. Popover can be responsive (on mobile become Dialog) **or not.**"
> The `or not` part means we need to have this behaviour as a property so the user can specify if they want the notification to appear in a dialog or keep the normal popover behaviour.

Having popover mobile so much different from the normal popover is the architectural mistake, i would prefer to fix it in the popover component, otherwise any component where we want to use both - normal & mobile popover will look very odd.

More info: https://github.com/SAP/fundamental-ngx/issues/8574

> Second, the box-shadow of the notification that creates the illusion of a border should probably be removed when it's in a responsive popover and I don't see the need of a footer in this case.

Fixed.

libs/docs/core/notification/examples/notification-mobile/notification-mobile-example.component.html
```diff
@@ -0,0 +1,263 @@
...
+    </ng-template>
+</fd-popover>
+
+<h4>Notification Group In Popover Dialog</h4>
+
+<button fd-button label="Notification Group" fdCompact [fdPopoverTrigger]="popoverMobileGroup"></button>
+
+<fd-popover #popoverMobileGroup [mobile]="true" [mobileConfig]="mobileConfig">
+    <ng-template #popoverBodyContent>
+        <fd-notification-group width="100%">
```
mikerodonnell89, line 116: I don't think `width="100%"` is needed

mikerodonnell89 (changes requested):
Example code is not displaying

Screen Shot 2022-10-04 at 1 59 24 PM

libs/core/src/lib/dialog/dialog.component.scss
```diff
@@ -22,4 +22,9 @@
             border-top-right-radius: 0;
         }
     }
+
+    .fd-dialog__body--no-horizontal-padding {
```
mikerodonnell89, line 26: I think this should be in fundamental-styles, oddly it is mentioned in the documentation there but this class is missing

libs/docs/core/notification/examples/notification-mobile/notification-mobile-example.component.html
```diff
@@ -0,0 +1,263 @@
+<h4>Notification In Popover Dialog</h4>
+
+<button fd-button label="Notification" fdCompact [fdPopoverTrigger]="popoverMobile"></button>
```
mikerodonnell89, line 3: Some buttons are hardcoded compact and some respond to the content density observer. Is this intentional? Looks like in the specs they should all be compact

mikerodonnell89 (approved):
Approving but I do have two comments/questions, see above

InnaAtanasova:
what is the black border around the dialog? And why aren't the bottom corners rounded?
Screen Shot 2022-10-24 at 1 26 45 PM
Screen Shot 2022-10-24 at 1 26 50 PM

InnaAtanasova:
I find this very weird: 
Screen Shot 2022-10-24 at 1 27 52 PM
Shouldn't the elements have the same width?

InnaAtanasova:
Clicking outside of the dialog and ESC should close the notifications "dialog"

InnaAtanasova (changes requested):
See my comments

libs/docs/core/notification/examples/notification-mobile/notification-mobile-example.component.html
platon-rov: Good catch, fixed for all examples

libs/core/src/lib/dialog/dialog.component.scss
platon-rov: https://github.com/SAP/fundamental-styles/pull/3957

platon-rov:
> what is the black border around the dialog? 

Cannot reproduce, chrome/firefox.

> And why aren't the bottom corners rounded?

In mobile popover (i.e. full width dialog) it makes sense to have all the borders angled so i made exact this https://github.com/SAP/fundamental-styles/pull/3957

> Shouldn't the elements have the same width?

I think mobile popover may be 100vh and 100vw but this one of many questions to this component, let's maybe fix it separately? Added to this ticket https://github.com/SAP/fundamental-ngx/issues/8574

> Clicking outside of the dialog and ESC should close the notifications "dialog"

Mobile popover doesn't have such behavior opposite to the Dialog. I also described it there https://github.com/SAP/fundamental-ngx/issues/8574

InnaAtanasova (approved):

Issue #8557, opened 2022-08-15
Defect Hunting Sprint 94


InnaAtanasova:
Please list all the issues you find during the defect hunting specifying the browser you used (in case it's browser specific)

InnaAtanasova:
- [ ] MS Edge Grid List visual issues:
    - [ ] In active state the text should be changed to contract color 
    
<img width="950" alt="Screen Shot 2022-08-15 at 12 56 55 PM" src="https://user-images.githubusercontent.com/39598672/184680386-ef2526d9-046d-411f-8f29-18ba93b34477.png">

    - [ ] In active state the focus outline should be white
    
<img width="1179" alt="Screen Shot 2022-08-15 at 12 55 53 PM" src="https://user-images.githubusercontent.com/39598672/184680512-24b07b43-f9b2-49cc-8eb3-0a5f554cd814.png">

InnaAtanasova:
- [ ] Grid List in RTL - the status indicator gets rounded corners on all sides
<img width="438" alt="Screen Shot 2022-08-15 at 1 03 22 PM" src="https://user-images.githubusercontent.com/39598672/184681583-584529f2-3b21-4928-b614-1948089cf9c0.png">

InnaAtanasova:
- [ ] Grid List images issue
<img width="459" alt="Screen Shot 2022-08-15 at 1 02 52 PM" src="https://user-images.githubusercontent.com/39598672/184681624-9e41aedb-0f18-47dd-b4bb-5b66778cee43.png">

InnaAtanasova:
- [ ] Quick View focus issues
1. Visually it looks broken (maybe needs some offset)
2. The focus should be kept inside the element (aka focus trap is missing)
<img width="464" alt="Screen Shot 2022-08-15 at 1 08 11 PM" src="https://user-images.githubusercontent.com/39598672/184682065-68ade3ca-5453-4f69-a4b0-2d3327179364.png">

InnaAtanasova:
- [ ] Scrollbar - shouldn't scrollbar be controlled using the keyboard?

InnaAtanasova:
- [ ] Quick View in Horizon:
<img width="1111" alt="Screen Shot 2022-08-15 at 1 33 57 PM" src="https://user-images.githubusercontent.com/39598672/184686028-8da19ae5-0e69-4d79-b772-0cf3a98db68e.png">
[...]

PR #8562, opened 2022-08-16
fix(core): prevent page scroll on up/down arrow key in table

mikerodonnell89:
part of #8557 

Disable page scroll when user is navigating through a focusable table with up/down arrow keys


libs/core/src/lib/table/table.component.ts
```diff
@@ -118,6 +118,7 @@ export class TableComponent implements AfterViewInit, OnDestroy, FdTable {
     _onCellKeydown(event: KeyboardEvent, cell: TableCellDirective): void {
         const cellElement = cell.elementRef.nativeElement;
         if (KeyUtil.isKeyCode(event, [DOWN_ARROW, UP_ARROW])) {
+            event.preventDefault();
             const data = cell.getCellPosition();
             if (!data) {
                 return;
```
N1XUS, line 124: I would suggest that if there's no data from getCellPosition, I would allow native behaviour (page scroll), so maybe move `event.preventDefault();` a bit down, after line 125

N1XUS (approved):

InnaAtanasova (approved):

PR #8566, opened 2022-08-18
feat(core): tabs defaultTab and selectDefaultOnTabsChange

platon-rov:
## Related Issue(s)

Refers to DXP.

## Description

* `defaultTab` input property to be able to specify the default tab.
* `selectDefaultOnTabsChange` input property to be able to control tab selection behavior on tabs change.

### After:

image

### DIFF
--- a/libs/core/src/lib/tabs/tab-panel/tab-panel.component.ts
+++ b/libs/core/src/lib/tabs/tab-panel/tab-panel.component.ts
@@ -5,6 +5,7 @@ import {
     ContentChild,
     ElementRef,
     EventEmitter,
+    HostBinding,
     Input,
     OnChanges,
     Optional,
@@ -33,9 +34,7 @@ export class TabPanelStateChange {
     host: {
         role: 'tabpanel',
         class: 'fd-tabs__panel',
-        '[attr.id]': 'id',
-        '[class.is-expanded]': 'expanded',
-        '[attr.aria-expanded]': 'expanded ? true : null'
+        '[attr.id]': 'id'
     },
     encapsulation: ViewEncapsulation.None,
     changeDetection: ChangeDetectionStrategy.OnPush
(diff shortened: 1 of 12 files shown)


libs/core/src/lib/tabs/tab-panel/tab-panel.component.ts
```diff
@@ -134,6 +141,17 @@ export class TabPanelComponent implements OnChanges {
...
+            setTimeout(() => {
+                this._updateHost();
+            });
         }
     }
+
+    /** @hidden */
+    private _updateHost(): void {
+        this._expandedClass = this.expanded;
+        this._expandedAria = this.expanded ? true : null;
```
InnaAtanasova, line 155: why `null` but not `false`?

platon-rov: when binding to attribute null evaluates to "no attribute at all", when false only about setting value.

libs/core/src/lib/tabs/tab-panel/tab-panel.component.ts
```diff
@@ -134,6 +141,17 @@ export class TabPanelComponent implements OnChanges {
             }
 
             this._changeDetRef.detectChanges();
+
+            // Done this way to avoid "NG0100: Expression has changed after it was checked"
```
g-cheishvili, line 145: ngZone.onStable?

mikerodonnell89 (commented):
The `defaultTab` input looks good and makes sense, but I am confused by what exactly the next example (screenshot below) is supposed to do, can you expand on the description a bit?
Screen Shot 2022-08-23 at 10 10 29 AM

InnaAtanasova (approved):

mikerodonnell89 (approved):

PR #8569, opened 2022-08-18
fix(docs): busy indicator - remove common margin for buttons in examples

shrvr:
## Related Issue(s)
part of #8557 

## Description

Removed common margin applied to all buttons of examples.

### Before:
Screen Shot 2022-08-18 at 2 41 24 PM

### After:
Screen Shot 2022-08-18 at 2 42 28 PM

### DIFF
--- a/apps/docs/src/app/core/component-docs/busy-indicator/examples/busy-indicator-extended-example.component.html
+++ b/apps/docs/src/app/core/component-docs/busy-indicator/examples/busy-indicator-extended-example.component.html
@@ -1,4 +1,4 @@
-<button fd-button label="Open from Template" (click)="openFromTemplate(template)"></button>
+<button style="margin-right: 8px" fd-button label="Open from Template" (click)="openFromTemplate(template)"></button>
 <button fd-button label="Hide All" fdType="emphasized" (click)="messageToastService.hideAll()"></button>
 
 <ng-template #template>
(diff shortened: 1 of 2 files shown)


apps/docs/src/app/core/component-docs/busy-indicator/examples/busy-indicator-extended-example.component.html
```diff
@@ -1,4 +1,4 @@
-<button fd-button label="Open from Template" (click)="openFromTemplate(template)"></button>
+<button style="margin-right: 8px" fd-button label="Open from Template" (click)="openFromTemplate(template)"></button>
```
InnaAtanasova, line 1: There shouldn't be a margin at all:
```
<button fd-button label="Open from Template" (click)="openFromTemplate(template)"></button>
```

InnaAtanasova (changes requested):

apps/docs/src/app/core/component-docs/busy-indicator/examples/busy-indicator-extended-example.component.html
shrvr: I think we need for that example to have space between the buttons.
Screen Shot 2022-08-19 at 2 15 09 PM

g-cheishvili (approved):

mikerodonnell89 (approved):

InnaAtanasova (approved):

PR #8571, opened 2022-08-18
fix(core): avatar group defect hunting

### LINKED ISSUE #8395: Defect Hunting: Avatar Group bugs
bug

- [ ] Avatar group RTL back button should face the other way

Screen Shot 2021-08-26 at 4 32 09 PM

- [ ] Avatar group (core): group type. Open the popover then use the space bar to select an avatar to drill in to the details. Page will scroll down - need to prevent default.

mikerodonnell89:
fixes #8395


InnaAtanasova (approved):

g-cheishvili (approved):
This will work and it's ok to have it this way.
For the future(I do not know if it already exists or not). I see a pipe like this:
```html
someThingThatDependsOnRTL="LTRValue | rtl : RTLValue"
```
This pipe is needed because in many cases, we might not care about rtl on the TS side and we are just creating dependency on rtl service without any reason

PR #8572, opened 2022-08-18
fix(core): allow click of entire separator in flexible column layout

### LINKED ISSUE #8389: Flexible Column Layout - increase the clickable area between the sections
bug

The whole animation layout should be clickable and once hovering over activate the > button, now only the > area behaves that way

mikerodonnell89:
fixes #8389


InnaAtanasova (approved):

g-cheishvili (approved):
Ok, this is acceptable, but in the next iteration I think we also should propagate the hover of the column separator element to the button, otherwise, it is not clear before clicking on the separator if it will have any effect or not

PR #8593, opened 2022-08-22
fix(core): breadcrumb keyboard arrow support

### LINKED ISSUE #8565: arrow keyboard support for breadcrumbs

enhancement

Breadcrumbs should be navigable via keyboard arrows

mikerodonnell89:
fixes #8565 

Adds arrow support for breadcrumbs. Using this instead of cdk's FocusKeyManager because FocusKeyManager does not cooperate with tab keypresses, nor the functionality with the collapsed breadcrumb menu.

### DIFF
--- a/libs/core/src/lib/breadcrumb/breadcrumb.component.html
+++ b/libs/core/src/lib/breadcrumb/breadcrumb.component.html
@@ -61,8 +61,8 @@
                 fd-link
                 tabindex="0"
                 class="fd-breadcrumb__collapsed"
-                (keydown.enter)="_keyDownHandle($event)"
-                (keydown.space)="_keyDownHandle($event)"
+                (keydown.enter)="_overflowKeyDownHandle($event)"
+                (keydown.space)="_overflowKeyDownHandle($event)"
             >
                 …
                 <fd-icon glyph="slim-arrow-down"></fd-icon>
--- a/libs/core/src/lib/breadcrumb/breadcrumb.component.html
+++ b/libs/core/src/lib/breadcrumb/breadcrumb.component.html
@@ -1,7 +1,7 @@
 <fd-overflow-layout
     [reverseHiddenItems]="!reverse"
     showMorePosition="left"
-    [enableKeyboardNavigation]="false"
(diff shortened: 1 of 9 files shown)


g-cheishvili:
You could update the active element index on tab navigation for instance and not create arrow navigation

libs/core/src/lib/breadcrumb/breadcrumb.component.html
```diff
@@ -61,8 +61,8 @@
                 fd-link
                 tabindex="0"
                 class="fd-breadcrumb__collapsed"
-                (keydown.enter)="_keyDownHandle($event)"
-                (keydown.space)="_keyDownHandle($event)"
+                (keydown.enter)="_overflowKeyDownHandle($event)"
```
N1XUS, line 64: Overflow layout supports keyboard navigation via `enableKeyboardNavigation` input property and adding `fdOverflowLayoutFocusableItem` directive to the item that can be focused.

N1XUS (changes requested):

libs/core/src/lib/breadcrumb/breadcrumb.component.html
mikerodonnell89: This line you've highlighted here is for the overflow menu button. The actual breadcrumb links themselves are coming from the app developer through `ng-content`.  Correct me if I am wrong but I think they would need to provide this directive as well through the `fd-link`s, and import the overflow module, so I think what I have in this PR is simpler for the developer

N1XUS: You can provide focusableItem as fd-link. Moreover, you can extend focusableItem directive to search for the focusableItem element inside directives content, which would be much better on my opinion

InnaAtanasova (approved):

InnaAtanasova:
When reviewing this PR keep in mind the following: https://github.com/SAP/fundamental-styles/issues/3821

N1XUS:
@mikerodonnell89 Quick question: If we use key arrows navigation, should we keep tab navigation over items, or just to focus on the first item, and then use arrows only?

PR #8599, opened 2022-08-24
fix(core): breadcrumb overflow click

### LINKED ISSUE #8521: Breadcrumb overflow items do not navigate on click
bug

0.35.0+

1. Open https://sap.github.io/fundamental-ngx/#/core/breadcrumb
2. Decrease screen width until an overflow menu gets shown
3. Open overflow menu 
4. Breadcrumbs that are displayed in the overflow do not navigate on click

### Screenshot
breadcrumb-click

mikerodonnell89:
fixes #8521

### DIFF
--- a/libs/core/src/lib/breadcrumb/breadcrumb-item.component.ts
+++ b/libs/core/src/lib/breadcrumb/breadcrumb-item.component.ts
@@ -51,7 +51,7 @@ export class BreadcrumbItemComponent implements AfterViewInit {
     constructor(public readonly elementRef: ElementRef<HTMLElement>) {}
 
     /** @hidden */
-    get needsClickProxy(): boolean {
+    get _needsClickProxy(): boolean {
         return (
             !!this.breadcrumbLink?.elementRef().nativeElement.getAttribute('href') || !!this.breadcrumbLink.routerLink
         );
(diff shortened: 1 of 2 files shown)


platon-rov (approved):

libs/core/src/lib/breadcrumb/breadcrumb-item.component.ts
```diff
@@ -51,7 +51,7 @@ export class BreadcrumbItemComponent implements AfterViewInit {
     constructor(public readonly elementRef: ElementRef<HTMLElement>) {}
 
     /** @hidden */
-    get _needsClickProxy(): boolean {
+    get needsClickProxy(): boolean {
```
InnaAtanasova, line 54: why is a getter annotated by `@hidden`. If you remove the `_` prefix you should also remove the `@hidden` typedocs and provide some description for the API docs

mikerodonnell89: I should not have removed the `_`, forgot that we use that for hidden vars/functions (in addition to private vars/functions).

I assume it's a getter due to the expression:
```
    /** @hidden */
    get _needsClickProxy(): boolean {
        return (
            !!this.breadcrumbLink?.elementRef().nativeElement.getAttribute('href') || !!this.breadcrumbLink.routerLink
        );
    }
```

PR #8609, opened 2022-08-26
fix(core): focus select control on escape press

### LINKED ISSUE #8305: fd-select focus is lost on press of escape when fd-option is opened
#### Angular version - 13.3.4
Fundamental-ngx - 0.34.3-rc.47

https://sap.github.io/fundamental-ngx/#/core/select

-

mikerodonnell89:
fixes #8305


fkolar (approved):
Besides that merge conflicts it looks good.

platon-rov (approved):

PR #8611, opened 2022-08-26
fix(core): tabbing for lists, lists in cards

### LINKED ISSUE #8303: [Accessibility] List Card items are accessible via TAB key instead of arrow up/down keys only
--
A bug

Briefly describe your proposal
--
List items should be accessible via arrow up/down keys only

Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
--
Issue observed with Angular 13 + F-NGX 0.35.1

If this is a bug, please provide steps for reproducing it.
--
Focus onto the very first list item under this component, and then press TAB key will move on to the next list item

Screen Shot 2022-06-30 at 12 31 57 PM

Is there anything else we should know?
-- 
Related SAP Ariba JIRA defect: https://product-jira.ariba.com/browse/SS-36895

mikerodonnell89:
fixes #8303

### DIFF
--- a/libs/core/src/lib/card/card.component.ts
+++ b/libs/core/src/lib/card/card.component.ts
@@ -1,12 +1,15 @@
 import {
+    AfterContentInit,
     ChangeDetectionStrategy,
     Component,
+    ContentChildren,
     ElementRef,
     HostBinding,
     Input,
     OnChanges,
     OnDestroy,
     OnInit,
+    QueryList,
     ViewEncapsulation
 } from '@angular/core';
 
@@ -20,6 +23,7 @@ import {
     contentDensityObserverProviders,
     ContentDensityMode
 } from '@fundamental-ngx/core/content-density';
+import { ListItemComponent } from '@fundamental-ngx/core/list';
 
 let cardId = 0;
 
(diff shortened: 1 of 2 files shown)


libs/core/src/lib/card/card.component.ts
```diff
@@ -83,6 +91,11 @@ export class CardComponent implements OnChanges, OnInit, CssClassBuilder, OnDest
         this.buildComponentCssClass();
     }
 
+    /** @hidden */
+    ngAfterContentInit(): void {
+        this.listItems?.first?.setIsFirst(true);
```
platon-rov, line 96: the list can be added later (after content init), so it won't be caught. probably we have to manage the list item's tabindex in the list item code.

mikerodonnell89: `the list can be added later (after content init), so it won't be caught`

Confused what you mean here, the card won't necessarily have a list so we need the ? check

`probably we have to manage the list item's tabindex in the list item code.`

That is what's happening in the `setIsFirst` function

g-cheishvili: He meant that ngAfterContentInit is triggered only once. What if the component did not have a list in it during content init, but after some conditions were met in the client's app and the list appeared? In that case, this `setIsFirst` will not get triggered

g-cheishvili (changes requested):
What Platon said

mikerodonnell89:
Misunderstood the original linked issue so I'm closing this pr

Issue #8600, opened 2022-08-24
Issues with multi-combobox custom datasource


ankur-sheelwant:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
Observed these issue while using a MultiComboBoxDataSource as a custom defined data source for Multi-combobox component.
1. Multi-combobox component becomes very slow and sometimes unresponsive. The API response for data is already received. It prints a series of console log statement approximately once for each item in the list as shown in the recording.  
2. Previously selected items don't rendered correctly for multi-combobox. The checkbox for these items are selected in the list but they don't show up in input box as selection.

https://user-images.githubusercontent.com/112031085/186535686-47a5b9f1-4b6a-49f4-ad1c-b21d5535bd0a.mov

  
3. When I do a search query and select multiple items, only the last item is being selected. i wrote customized code for fetch method shown below.

https://user-images.githubusercontent.com/112031085/186538959-25fc5345-2742-4965-b4d8-78127a7e428e.mov

4. Mulit-combobox component throws exception '[dataSource] source did not match an array, Observable, or DataSource'  as shown below:

https://user-images.githubusercontent.com/112031085/186537060-c0d290bf-f99b-4794-922b-d1710f2cdbe3.mov

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Issues were found on latest release: v 0.35.4

#### Please provide relevant source code if applicable.
Screenshot of custom datasource code:
![Screen Shot 2022-08-24 at 6 12 31 PM](https://user-images.githubusercontent.com/112031085/186539628-a08fcf47-be76-4b38-b139-7a060388f8e9.png)

![Screen Shot 2022-08-24 at 6 12 58 PM](https://user-images.githubusercontent.com/112031085/186539636-67762905-3de9-4475-9824-a0489ff2b246.png)

![Screen Shot 2022-08-24 at 6 13 13 PM](https://user-images.githubusercontent.com/112031085/186539649-cd86f970-567a-41c2-8f84-5924b38cc390.png)

#### Is there anything else we should know?
These issues are not observed when we use simple array as a datasource input.

fkolar:
In term of performance, for this issue, you are loading all the records found in the backend right ? 

How many issues we have here? Let's try to isolate them:
* Performance - we can optimize the fetch by setting proper limit on how many records you want to retrieve at once. 
* Selection is also a problem? 

[...]

PR #8624, opened 2022-08-30
fix(platform): fixed multi-combobox same name issue

### LINKED ISSUE #8587: fdp-multicheckbox issue with same name
Bug

I am using fdp-multicheckbox component.

Example:
If Apple is from Fruits category and again Apple is from Fruits1 category, it looks like:
image

[Multicheckbox stackblitz link - Just type Apple and select](https://stackblitz.com/edit/angular-6mvwz6?file=src/app/multi-combobox-columns-example.component.ts
)

RealWorld example  - If two users have same name but belong to a different group. Then we are observing this issue.

13.3.4

[Multicheckbox stackblitz link - Just type Apple and select](https://stackblitz.com/edit/angular-6mvwz6?file=src/app/multi-combobox-columns-example.component.ts
)

Please fix this ASAP
 [Reference -](https://stackblitz.com/edit/primeng-multiselect-demo-vggwuh?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.html)

skaquib123:
## Related Issue(s)

closes #8587 and #8600

## Description
In multi-combobox when a list is having items with same name, there was an error adding those items. 
for example, if there are two items with name  "Apple" but types different, user should be able to check both the apple. 
It is fixed now. 

User have to provide a unique id in lookup key field from the datasource as shown in the picture which will be used to compare each values among items.
image

Datasource:
image

### Before:

image

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [x] poor examples

4. Accessibility testing

##### PR Quality

-   [x] update `README.md`


mikerodonnell89 (approved):

platon-rov (approved):

libs/platform/src/lib/form/multi-combobox/commons/base-multi-combobox.ts
```diff
@@ -685,7 +685,7 @@ export abstract class BaseMultiCombobox extends CollectionBaseInput implements O
 
                 this._suggestions = this._convertToOptionItems(data).map((optionItem: SelectableOptionItem) => {
                     const selectedElement = this._selectedSuggestions.find(
-                        (selectedItem: SelectableOptionItem) => selectedItem.label === optionItem.label
+                        (selectedItem: SelectableOptionItem) => selectedItem.value === optionItem.value
```
fkolar, line 688: In terms of `.value`, we should probably make something different, if object, then those values might not match. You have some selectedItems, and then you receive a new on from that stream and references will be different. we cannot rely it can be the same. 

So we need some uniquely identifiable value. or we do more deep object comparision

platon-rov (approved):

libs/docs/platform/multi-combobox/examples/multi-combobox-columns/multi-combobox-columns-example.component.ts
```diff
@@ -11,15 +11,15 @@ import { MultiComboboxSelectionChangeEvent } from '@fundamental-ngx/platform/for
 })
 export class MultiComboboxColumnsExampleComponent {
     dataSource = [
```
fkolar, line 13: Yes we need to provide best practises how to work with objects, they need to provide an ID, otherwise when lookup key is not set we should throw warning message

fkolar (changes requested):
Looks good, I just have one fundamental ask, is to add JS doc explaining how this works, with strings and objects, why its important for complex object to use lookup Value and what will happen when they don.t. 

We just had a bit now, both documentation app examples so its good to describe this. 

```
/**
 * This component can works with both string primitives as well as with complex objects. In order to... etc,,
 */
@Component({
    selector: 'fdp-multi-combobox',
    templateUrl: './multi-combobox.component.html',
    styleUrls: ['./multi-combobox.component.scss'],
    providers: [
        {
            provide: FormFieldControl,
            useExisting: MultiComboboxComponent,
            multi: true
        },
        contentDensityObserverProviders()
    ]
})
export class MultiComboboxComponent extends BaseMultiCombobox implements OnInit, AfterViewInit {
    /** @hidde

```

skaquib123:
@fkolar I have made changes and added js docs.

mikerodonnell89 (changes requested):
If the developer wants to use objects, but does not provide a lookupKey, I think we should automatically generate unique ID's for each item. But I would also like to keep this current setup if the developer does want to provide a lookupKey

fkolar (approved):
Looks good to me.

fkolar:
> If the developer wants to use objects, but does not provide a lookupKey, I think we should automatically generate unique ID's for each item. But I would also like to keep this current setup if the developer does want to provide a lookupKey

HI @mikerodonnell89 , This is impossible. there are some enterprise setup, when working with complex objects and you need ID, as 99% of the time you will be fetching data out of your REST API -> from DB. 

I am not sure how would you generate ID and then manage equality between objects. We try to come up with the most performant solution that could work for enterprise applications. We are talking about real objects that can contains 50 fields+ some additional nested structure, so trying to compare object my reference or deep comparison would not good. This is why there was this idea of LookupKey since we started to design this component. The only think we were not able to manage well this final implementation, how we worked with the objects.

fkolar:
Merged, as we had many discussion about this behavior and did not get any response for 6 days, so we went ahead and merged this. As this was blocking issue for our adoption teams where customers raised a few issues. .

PR #8626, opened 2022-08-31
fix(core): combobox scroll strategy

### LINKED ISSUE #8228: Positioning of list item popover in Combobox control
Bug
The positioning of the popover in Combobox gets distorted when the user scrolls through the page while the Combobox list is opened.
(link: https://sap.github.io/fundamental-ngx/#/core/combobox). I guess we should hide the Combobox list when the input is not visible or we should position the Combobox list accordingly.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use the current version.)
fundamental-ngx/core: ^0.33.1
angular: 13.2.2

1) Go to the example section of Combobox here (https://sap.github.io/fundamental-ngx/#/core/combobox).
2) Click on any Combobox list
3) While the Combobox list is opened scroll the page down/up
4) Please notice that the Combobox list is still shown even when the Combobox input is hidden

Attaching the image below for reference

image

platon-rov:
## Related Issue(s)

Closes #8228.

## Description

Combobox scroll strategy fix.

Ideally, the scroll strategy should be set to `BLOCK` the same as the default HTML select but it seems very buggy and doesn't respect when the scroll container if it's not a body. https://github.com/angular/components/issues/21195

So, the scroll strategy is set to `REPOSITION` but with the "close when scrolled out of viewport" flag enabled.

The issue about z-indexes probably should be fixed separately because it relates to our docs app and not any component. We have various z-indexes set for different components, and probably we have to review all those values.

### After:

https://user-images.githubusercontent.com/20265336/187691632-73ea3dfc-74d0-4909-b687-6f96b95f15a6.mov

### DIFF
--- a/libs/core/src/lib/combobox/combobox.component.html
+++ b/libs/core/src/lib/combobox/combobox.component.html
@@ -7,6 +7,7 @@
         [isOpen]="open && displayedValues && displayedValues.length > 0"
         (isOpenChange)="isOpenChangeHandle($event)"
         [fillControlMode]="fillControlMode"
+        [scrollStrategy]="_repositionScrollStrategy"
         [focusTrapped]="true"
         [triggers]="triggers"
         [disabled]="disabled || readOnly"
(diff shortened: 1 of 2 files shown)


g-cheishvili (changes requested):
If I open first example combobox and then scroll down, to make It jump outside viewport, dropdown closes, but whole document gets scrolled up to page top

g-cheishvili (approved):

PR #8628, opened 2022-08-31
fix(platform): object lists should never be compact

mikerodonnell89:
fixes none

before:
Screen Shot 2022-08-31 at 11 59 27 AM

after:
Screen Shot 2022-08-31 at 12 01 06 PM

### DIFF
--- a/libs/platform/src/lib/list/list.component.ts
+++ b/libs/platform/src/lib/list/list.component.ts
@@ -1,4 +1,5 @@
 import {
+    AfterViewChecked,
     AfterViewInit,
     ChangeDetectionStrategy,
     ChangeDetectorRef,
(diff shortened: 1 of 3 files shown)


droshev (approved):

libs/platform/src/lib/list/list.component.ts
```diff
@@ -430,6 +434,14 @@ export class ListComponent<T> extends CollectionBaseInput implements OnInit, Aft
         indicator?.setAttribute('aria-label', '');
     }
 
+    /** @hidden */
+    ngAfterViewChecked(): void {
+        if (this.hasObject) {
+            // object lists should never be compact
+            this._ulElement?.classList.remove('fd-list--compact');
```
platon-rov, line 441: i think it should be done in styles like so

```css
.fd-list--compact:not(.fd-object-list) { ... }
```

mikerodonnell89: That is a better idea! Let me give it a try, thanks

mikerodonnell89:
Closing in favor of https://github.com/SAP/fundamental-styles/pull/3842

PR #8630, opened 2022-09-01
feat(core,platform): require-jsdoc rule

platon-rov:
## Related Issue(s)

Closes none.

## Description

* [`require-jsdoc`](https://github.com/gajus/eslint-plugin-jsdoc#user-content-eslint-plugin-jsdoc-rules-require-jsdoc) eslint rule enabled for everything, disabled for docs, docs-e2e, e2e.
* [`grouped-accessor-pairs`](https://eslint.org/docs/latest/rules/grouped-accessor-pairs) rule enabled for everything, disabled for docs, docs-e2e, e2e.

_In conjunction with `require-jsdoc` settings, it gives us the ability to write typedocs only above setter if both are present. 
 It also ensures that the order of setter-getter pair is always the same, what's neat by itself._

* missing typedocs added in core, platform.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -144,10 +144,24 @@
                 "eol-last": "error",
                 "eqeqeq": ["error", "smart"],
                 "guard-for-in": "error",
+                "grouped-accessor-pairs": ["error", "setBeforeGet"],
                 "import/order": "off",
                 "jsdoc/check-alignment": "off",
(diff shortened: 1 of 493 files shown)


g-cheishvili (approved):
This will reduce so much headaches during reviews

mikerodonnell89 (approved):

InnaAtanasova (approved):

Issue #8469, opened 2022-07-28
Table Keyboard support review


droshev:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.

Let's review the current Table keyboard support. (e.g. the Tab should focus only one 1 row/cell and then with Up/Down navigate within the table) and Tabbing again should move to the next component

N1XUS:
@droshev do we have anything left here to do? Or we can close the issue?

N1XUS:
Closing this issue due to inactivity.

PR #8635, opened 2022-09-02
feat(core): table keyboard navigation

N1XUS:
## Related Issue(s)

relates #8469 

## Description
Changed core table keyboard navigation. Extended documentation with explanation of the keys that are used for the navigation inside table's elements.

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [x] poor examples

4. Accessibility testing

##### PR Quality

-   [x] update `README.md`

### DIFF
--- a/apps/docs/src/app/core/component-docs/table/table-docs.component.html
+++ b/apps/docs/src/app/core/component-docs/table/table-docs.component.html
@@ -81,6 +81,10 @@
         When pressing Tab key inside cell and there's no focusable elements, focus will be switched to the next
         focusable element outside the table.
     </p>
+    <p>
+        <strong>Note:</strong> If both row and cells in the row are focusable, only row will be navigatable via
+        keyboard.
+    </p>
 </description>
 <component-example>
     <fd-table-focusable-example></fd-table-focusable-example>
--- a/apps/docs/src/app/core/component-docs/table/table-docs.component.html
+++ b/apps/docs/src/app/core/component-docs/table/table-docs.component.html
@@ -71,7 +71,16 @@
 
 <fd-docs-section-title id="focusable" componentName="table"> Focusable </fd-docs-section-title>
 <description>
-    Table rows and cell can be focusable, by adding <code>[focusable]="true"</code> and <code>tabindex="0"</code> to it.
+    <p>
+        Table rows and cell can be focusable, by adding <code>[focusable]="true"</code> and <code>tabindex="0"</code> to
+        it.
+    </p>
+    <p>Navigation inside table works with arrow keys.</p>
+    <p>Navigating between focusable element inside table cell works with the Tab key.</p>
+    <p>
+        When pressing Tab key inside cell and there's no focusable elements, focus will be switched to the next
+        focusable element outside the table.
+    </p>
 </description>
 <component-example>
     <fd-table-focusable-example></fd-table-focusable-example>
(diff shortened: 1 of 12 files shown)


libs/core/src/lib/table/directives/table-body.directive.ts
```diff
@@ -8,13 +8,23 @@ export class TableBodyDirective {
...
 
-    /** Whether or not to show the table body's vertical borders */
+    /**  Whether to show the table body's vertical borders */
     @HostBinding('class.fd-table__body--no-vertical-borders')
     @Input()
     noBorderY = false;
+
+    @HostListener('focus')
+    private _onFocus(): void {
+        console.log('Table body focus');
```
platon-rov, line 23: not needed.

apps/docs/src/app/core/component-docs/table/table-docs.component.html
```diff
@@ -71,7 +71,16 @@
...
-    Table rows and cell can be focusable, by adding <code>[focusable]="true"</code> and <code>tabindex="0"</code> to it.
+    <p>
+        Table rows and cell can be focusable, by adding <code>[focusable]="true"</code> and <code>tabindex="0"</code> to
+        it.
+    </p>
+    <p>Navigation inside table works with arrow keys.</p>
+    <p>Navigating between focusable element inside table cell works with the Tab key.</p>
+    <p>
+        When pressing Tab key inside cell and there's no focusable elements, focus will be switched to the next
+        focusable element outside the table.
```
platon-rov, line 82: Please mention that if both cells & rows are focusable then only the rows will get focus.

platon-rov (approved):

mikerodonnell89 (commented):
This probably isn't the focus outline style we want, but I think we are going against Fiori specs with this clickable header so I'm not really sure what to do here

Screen Shot 2022-09-02 at 9 54 34 AM

mikerodonnell89:
I think in the interactive rows and cells example, the clickable cells should be navigable via keyboard

mikerodonnell89 (commented):
Should these get tab focus?

Screen Shot 2022-09-02 at 10 01 03 AM

mikerodonnell89 (changes requested):
Nice work! Just a couple comments above that are more like questions than change requests

N1XUS:
> This probably isn't the focus outline style we want, but I think we are going against Fiori specs with this clickable header so I'm not really sure what to do here
> 
> Screen Shot 2022-09-02 at 9 54 34 AM

This is actually non-navigatable table. Tab works there because there's tabbable elements such as links and some element with tabindex. I'll update the example to make it navigatable.

As to the wrong focus outline: we can introduce new class regarding sortable table header element, but I think it's outside the scope of this task.

N1XUS:
> Should these get tab focus? Only focusable via keyboard at the moment
> 
> Screen Shot 2022-09-02 at 10 01 03 AM

By default, any focusable element inside a table with focusable cells will not receive a focus on tab key. This is done in order for users not to tap on Tab key for 200 times if there's a large table. Instead, the process of table navigation is done in such logic:
When user focuses on the cell, we search for the first focusable element inside, and try to set focus to it. If not, set focus on the whole row. After that, if the cell contains more than one focusable element, users can use Tab key to navigate through those elements inside that cell. Navigation between cells happens with arrow keys. Pressing tab while focusing cell, or inner focusable element there, where there's no more focusable items will shift the focus from the entire table to the next focusable element. Pretty much how W3C recommends keyboard navigation for the tables.

g-cheishvili (approved):
Nice, I like it!

mikerodonnell89 (approved):
Thanks for your answers @N1XUS

PR #8637, opened 2022-09-02
fix(core): incorrect emitting of tab change events

### LINKED ISSUE #8050: bug: FDP Dynamic Page / FD Tab List: `tabChange` event triggered on scroll
bug
When using the FDP Dynamic Page or the FD Tab List, the `tabChange` event is triggered on the first scroll on the tab even when `[stackContent]="false"` is set.
"@angular/core": "13.1.3",
"@fundamental-ngx/core": "0.33.7",
"@fundamental-ngx/platform": "0.33.7",
1) Go to  https://fundamental-ngx.netlify.app/#/platform/dynamic-page#tabbed
2) open the devtools
3) open the `Default tabs example` example
4) see the message `tab changed to tab_1` in the console (expected)
5) adjust the width of the window to make sure you can scroll
6) Scroll
7) see the message `tab changed to tab_1` in the console again, not expected, the tab did not change

mikerodonnell89:
fixes #8050 

Fixes an issue where tab change events were not correctly firing. Event would be emitted when the tab did not actually change


libs/core/src/lib/tabs/tab-list.component.ts
```diff
@@ -205,10 +205,12 @@ export class TabListComponent implements AfterContentInit, AfterViewInit, OnDest
 
     /** @hidden */
     _highlightActiveTab({ id }: HTMLElement): void {
-        const tab = this._tabArray.find((_tab) => _tab.id === id);
+        const tab = this._tabArray.find((_tab) => _tab.panel._panelId === id);
         if (tab) {
+            if (!tab.active) {
+                this.selectedTabChange.emit(tab.panel);
+            }
             this._activateStackedTab(tab.panel, false);
```
g-cheishvili, line 213:
```suggestion
            this._activateStackedTab(tab.panel, false);
            if (!tab.active) {
                this.selectedTabChange.emit(tab.panel);
            }
```

g-cheishvili: Without this, it will just emit inactive tab and for some brief moment, tab-list user will consider emitted value as active, when in reality it is not active yet

g-cheishvili (changes requested):

platon-rov (approved):

PR #8638, opened 2022-09-02
feat(core,platform): add Chinese, Polish and Hindi language

droshev:
add Chinese, Polish and Hindi language


InnaAtanasova (approved):

libs/i18n/src/lib/languages/polish.ts
```diff
@@ -5,537 +5,543 @@ import { FdLanguage } from '../models/lang';
...
-        dateRangeInputLabel: 'Date range input',
-        displayCalendarToggleLabel: 'Open picker',
-        valueStateSuccessMessage: 'Value state Success',
-        valueStateInformationMessage: 'Value state Information',
-        valueStateWarningMessage: 'Value state Warning',
-        valueStateErrorMessage: 'Value state Error'
+        dateInputLabel: 'Data',
+        dateRangeInputLabel: 'Przedział czasowy',
+        displayCalendarToggleLabel: 'Otwórz selektor',
+        valueStateSuccessMessage: 'Value state Success', // I have no idea what it is about (?)
```
mikerodonnell89, line 15: It is for a hidden div indicating to the screenreader that the datepicker input has a success state

Screen Shot 2022-09-02 at 1 42 19 PM

mikerodonnell89 (changes requested):
see comment

libs/i18n/src/lib/languages/polish.ts
gkrajniak: sure but even in english I never have I ever hit into the labels like:

```
        valueStateSuccessMessage: 'Value state Success',   
        valueStateInformationMessage: 'Value state Information', 
        valueStateWarningMessage: 'Value state Warning',  
        valueStateErrorMessage: 'Value state Error' 
```

When the `Value state Success` makes some sense, the `Value state InformationWarrning` even in English for me is odd, so if I had to translate I would be:

```
        valueStateSuccessMessage: 'Stan wartości Poprawna',   
        valueStateInformationMessage: 'Stan wartości Informacja', 
        valueStateWarningMessage: 'Stan wartości Ostrzeżenie',  
        valueStateErrorMessage: 'Stan wartości Niepoprawna' 
```

PR #8651, opened 2022-09-06
fix(platform): multi-combobox displayKey  name property is taking com…

skaquib123:
…plete text

## Related Issue(s)

closes  #8631

## Description

In multi-combobox DisplayKey name property was not taking complete text. 
for data={" name:'apple<red>', type:'fruits1' "} it was showing only "apple" in list and now it is showing "apple<red>".

### Before:
Screenshot 2022-09-06 at 4 33 21 PM

### After:
MicrosoftTeams-image

##### During Implementation

1. Visual Testing:

-   [x] visual misalignments/updates
-   [x] check Light/Dark/HCB/HCW themes
-   [x] RTL/LTR - proper rendering and labeling
-   [x] responsiveness(resize)
-   [x] Content Density (Cozy/Compact/(Condensed))
-   [x] States - hover/disabled/focused/active/on click/selected/selected hover/press state
-   [x] Interaction/Animation - open/close, expand/collapse, add/remove, check/uncheck
-   [x] Mouse vs. Keyboard support
-   [x] Text Truncation

2. API and functional correctness

-   [x] check for console logs (warnings, errors)
-   [x] API boundary values
-   [x] different combinations of components - free style
-   [x] change the API values during testing

3. Documentation and Example validations

-   [x] missing API documentation or it is not understandable
-   [x] poor examples
-   [x] Stackblitz works for all examples

4. Accessibility testing
5. Browser Testing - Edge, Safari, Chrome, Firefox

##### PR Quality

-   [x] the commit message(s) follows the guideline:
        https://github.com/SAP/fundamental-ngx/blob/main/CONTRIBUTING.md
-   [x] tests for the changes that have been done
-   [x] all items on the PR Review Checklist are addressed :
        https://github.com/SAP/fundamental-ngx/wiki/PR-Review-Checklist
-   [x] Run npm run build-pack-library and test in external application
-   [x] update `README.md`
-   [x] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)

PR #8658, opened 2022-09-08
fix(core,platform): remove platform highlighting and account for lt and gt in highlight pipe

### LINKED ISSUE #8631: fdp-multi-combobox - displaykey name property is not taking the complete text
Bug

dataSource with name and type, where If the name is set to user<email>, the dropdown is just displaying user.

Angular 13.x
ngx 0.35.4

https://stackblitz.com/edit/angular-6mvwz6?file=src%2Fapp%2Fmulti-combobox-columns-example.component.ts
Open the dropdown and issue can be observed.

mikerodonnell89:
fixes #8631 

This is a replacement for #8651 

Removes platform's redundant highlighting pipes, and the core highlighting pipe now accounts for `<` and `>` which were not being rendered (see linked issue/pr above for more details)


libs/core/src/lib/utils/pipes/search-highlight.pipe.ts
```diff
@@ -5,6 +5,8 @@ import { Pipe, PipeTransform } from '@angular/core';
 })
 export class SearchHighlightPipe implements PipeTransform {
     transform(value: string, args: string, active: boolean = true, includeSpans: boolean = false): string {
+        value = value.replace(/</g, '&lt;');
```
g-cheishvili, line 8: Since we use lodash anyways, why not [utilize it escape](https://lodash.com/docs/#escape)?

skaquib123:
@mikerodonnell89 I don't think we can use innerHTML as it will convert any unicode to HTML tag.
 dataSource = [
        { name: 'Apple<hello>', type: 'Fruits' },
        { name: 'Banana&gt', type: 'Fruits' },
        { name: 'Pineapple', type: 'Fruits' }
        ];
for datasource even if we use replace functionality it will print as shown in picture below.
image

mikerodonnell89:
> @mikerodonnell89 I don't think we can use innerHTML as it will convert any unicode to HTML tag. dataSource = [ { name: `Apple<hello>`, type: 'Fruits' }, { name: 'Banana&gt', type: 'Fruits' }, { name: 'Pineapple', type: 'Fruits' } ]; for datasource even if we use replace functionality it will print as shown in picture below. image

In this case we need to refactor the pipe such that it does not require the components to use innerHTML. I think we could programmatically assign font-weight style

libs/core/src/lib/utils/pipes/search-highlight.pipe.ts
```diff
@@ -1,13 +1,15 @@
 import { Pipe, PipeTransform } from '@angular/core';
+import _ from 'lodash-es';
```
g-cheishvili, line 2:
```suggestion
import escape from 'lodash-es/escape'
```

libs/core/src/lib/utils/pipes/search-highlight.pipe.ts
```diff
@@ -1,13 +1,15 @@
...
 
 @Pipe({
     name: 'highlight'
 })
 export class SearchHighlightPipe implements PipeTransform {
     transform(value: string, args: string, active: boolean = true, includeSpans: boolean = false): string {
+        value = _.escape(value);
```
g-cheishvili, line 9:
```suggestion
        value = escape(value);
```

libs/core/src/lib/utils/pipes/search-highlight.pipe.ts
```diff
@@ -1,13 +1,15 @@
...
         let result: string = value;
         if (args && value && active) {
-            const testStr: string = args.trim().toLowerCase();
+            const testStr: string = _.escape(args.trim().toLowerCase());
```
g-cheishvili, line 12:
```suggestion
            const testStr: string = escape(args.trim().toLowerCase());
```

g-cheishvili (changes requested):

platon-rov (approved):

InnaAtanasova (approved):

g-cheishvili (approved):

Issue #8398, opened 2022-07-19
Defect Hunting: multi-input issue


droshev:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.
- [ ] Multi-input needs to disable browser autocomplete by default (see https://github.com/SAP/fundamental-ngx/issues/5889)
<img width="356" alt="Screen Shot 2021-08-26 at 4 50 57 PM" src="https://user-images.githubusercontent.com/2471874/131046040-b1f4b622-a1c3-4dfa-b586-2a86ee3a18c5.png">\

- [ ] Core multiinput - open the popover using the button on the right - you can no longer use the keyboard down arrow or tab keys to access items in the popover - we have an issue for tackling the multiinput keyboard.

PR #8686, opened 2022-09-14
fix(core): multi input focusing

platon-rov:
## Related Issue(s)

Closes https://github.com/SAP/fundamental-ngx/issues/8398

## Description

Core multi-input focusing issue.

### Before:

image

### After:

image

### DIFF
--- a/libs/core/src/lib/input-group/input-group.component.html
+++ b/libs/core/src/lib/input-group/input-group.component.html
@@ -39,6 +39,7 @@
                 [attr.aria-controls]="_inputId"
                 [disabled]="disabled"
                 (mousedown)="_preventFocus($event)"
+                (keydown.enter)="$event.preventDefault()"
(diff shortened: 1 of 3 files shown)


N1XUS (approved):

libs/core/src/lib/input-group/input-group.component.html
```diff
@@ -39,6 +39,7 @@
                 [attr.aria-controls]="_inputId"
                 [disabled]="disabled"
                 (mousedown)="_preventFocus($event)"
+                (keydown.enter)="$event.preventDefault()"
```
mikerodonnell89, line 42: This prevents the popover list from opening on enter press, I don't think we want that do we?

mikerodonnell89 (changes requested):

g-cheishvili (approved):

mikerodonnell89 (approved):

PR #8690, opened 2022-09-14
fix(core): remove list focus trap

### LINKED ISSUE #8687: List items traps focus
bug

At the moment the list traps the focus and the user cannot go out with the keyboard.

latest

shrvr:
## Related Issue(s)

closes #8687 

## Description

Removed list focus trap on tab key.

### DIFF
--- a/libs/core/src/lib/utils/services/keyboard-support/keyboard-support.service.ts
+++ b/libs/core/src/lib/utils/services/keyboard-support/keyboard-support.service.ts
@@ -27,7 +27,21 @@ export class KeyboardSupportService<T> {
     }
 
     /** @hidden */
-    setKeyboardService(queryList: QueryList<KeyboardSupportItemInterface & T>, wrap?: boolean): void {
+    private _itemList: QueryList<KeyboardSupportItemInterface & T>;
+
+    /** @hidden
+     * allow focus trap for tab key. default is true.
+     */
+    private _focusTrapOnTab = true;
+
+    /** @hidden */
+    setKeyboardService(
+        queryList: QueryList<KeyboardSupportItemInterface & T>,
+        wrap?: boolean,
+        focusTrapOnTab = true
+    ): void {
+        this._itemList = queryList;
+        this._focusTrapOnTab = focusTrapOnTab;
         this._keyManager = new FocusKeyManager(queryList).withWrap(wrap).withHomeAndEnd();
         queryList.changes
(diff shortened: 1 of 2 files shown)


mikerodonnell89 (approved):
This works well and addresses the issue with FocusKeyManager losing sight of its active item when tab/shift+tab is used

platon-rov (changes requested):
Before the action list fix pr (https://github.com/SAP/fundamental-ngx/pull/8458) we had only the first item in list focusable with tab and nov every item has tab stop.

shrvr:
> Before the action list fix pr (#8458) we had only the first item in list focusable with tab and nov every item has tab stop.

So to Navigate within the list,  should only arrow keys be used? Because for action sheet the navigation is by both Tab and arrow keys, and both share Keyboard Service.
Or I can add one more argument to `setKeyboardService ` to enable or disable tab key listening.

mikerodonnell89 (changes requested):
I think refactoring the onKeyDown function of `keyboard-support.service.ts` to this should do it: 

```
    /** @hidden */
        this._keyManager.onKeydown(event);
        if (KeyUtil.isKeyCode(event, TAB) && this._focusTrapOnTab) {
            event.preventDefault();
            hasModifierKey(event, 'shiftKey') ? this.keyManager.setPreviousItemActive() : this.keyManager.setNextItemActive();
        }
    }
```

libs/core/src/lib/utils/services/keyboard-support/keyboard-support.service.ts
```diff
@@ -38,13 +52,19 @@ export class KeyboardSupportService<T> {
     onKeyDown(event: KeyboardEvent): void {
```
mikerodonnell89, line 52: I think try refactoring this function to 

```suggestion
    /** @hidden */
    onKeyDown(event: KeyboardEvent): void {
        this._keyManager.onKeydown(event);
        if (KeyUtil.isKeyCode(event, TAB) && this._focusTrapOnTab) {
            event.preventDefault();
            hasModifierKey(event, 'shiftKey') ? this.keyManager.setPreviousItemActive() : this.keyManager.setNextItemActive();
        }
    }
```

mikerodonnell89 (changes requested):

mikerodonnell89 (approved):

platon-rov (approved):

PR #8692, opened 2022-09-14
fix(core): action sheet mobile tab bug, also list tab fix

### LINKED ISSUE #8687: List items traps focus

mikerodonnell89:
fixes #8687


mikerodonnell89:
Closing, doesn't address:

- [ ] Action sheet (core): strange keyboard behavior. Open an action sheet popover via mouse click on the control. Note that the first item in the list is focused. Press tab twice to move focus to the 3rd item. Press the down arrow key and the 2nd item will be focused

PR #8695, opened 2022-09-15
feat(core): add time selection using keyboard

### LINKED ISSUE #5645: Date Time Picker: Providing the time via Keyboard is not Possible
#### Comparing the date time picker with its UI5 version I recognized that providing the time values by typing on the keyboard doesn't work. This is quite an UX issue if you imagine that a user has to scroll all the way thru 60 minutes or seconds to pick the appropriate value. Either typing is not working properly or was forgotten.

See attached .gif of UI5 control as reference.
datetimepicker-typing

#### v0.30.0

shrvr:
## Related Issue(s)

closes #5645 

## Description

Added time input via keyboard.

### DIFF
--- a/libs/core/src/lib/time/time-column/time-column.component.ts
+++ b/libs/core/src/lib/time/time-column/time-column.component.ts
@@ -290,7 +290,7 @@ export class TimeColumnComponent<K, T extends SelectableViewItem<K> = Selectable
         } else if (KeyUtil.isKeyType(event, 'numeric')) {
             this._numericKeyInputHandler(event);
         } else if (KeyUtil.isKeyType(event, 'alphabetical')) {
-            this._queryKeyDownEvent.next(event.key);
+            this._alphaKeyInputHandler(event);
         }
     }
 
@@ -411,6 +411,17 @@ export class TimeColumnComponent<K, T extends SelectableViewItem<K> = Selectable
     }
 
     /** @hidden
+     * handles alphabetical key inputs to set period.
+     */
+    private _alphaKeyInputHandler(event: KeyboardEvent): void {
+        if (event.key === 'a' || event.key === 'A') {
+            this._pickTimeOnValue('AM');
+        } else if (event.key === 'p' || event.key === 'P') {
+            this._pickTimeOnValue('PM');
+        }
+    }
+
+    /** @hidden
      * handles numeric key inputs to set time.
      */
     private _numericKeyInputHandler(event: KeyboardEvent): void {
@@ -432,7 +443,7 @@ export class TimeColumnComponent<K, T extends SelectableViewItem<K> = Selectable
     /** @hidden
      * set the time for given item value. used by _numericKeyInputHandler.
      */
-    private _pickTimeOnValue(value: number): void {
+    private _pickTimeOnValue(value: string | number): void {
         this._pickTime(
             this.items.find((item) => item.value.value === value),
             false,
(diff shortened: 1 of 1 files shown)


libs/core/src/lib/time/time-column/time-column.component.ts
```diff
@@ -291,6 +303,38 @@ export class TimeColumnComponent<K, T extends SelectableViewItem<K> = Selectable
...
+            true,
+            false
+        );
+        this._keyLog = []; // clear key log
+    }
+
+    /** @hidden
+     * handles numeric key inputs to set time.
+     */
+    _numericKeyInputHandler(event: KeyboardEvent): void {
```
platon-rov, line 322: both methods could be private.

libs/core/src/lib/time/time-column/time-column.component.ts
```diff
@@ -291,6 +303,38 @@ export class TimeColumnComponent<K, T extends SelectableViewItem<K> = Selectable
         }
     }
 
+    /** @hidden
+     * set the time for given item value. used by _numericKeyInputHandler.
+     */
+    _pickTimeOnValue(value: number): void {
```
platon-rov, line 309: minor: it would feel better if it will be below the method that calls this method, because we read code from top to bottom.

libs/core/src/lib/time/time-column/time-column.component.ts
```diff
@@ -291,6 +303,38 @@ export class TimeColumnComponent<K, T extends SelectableViewItem<K> = Selectable
...
+        this._numericInputTimeout && clearTimeout(this._numericInputTimeout);
+        this._keyLog.push(event.key);
+        const inputValue = parseInt(this._keyLog.join(''), 10); // converts keyLog elements to a number
+
+        if (inputValue * 10 > lastItemValue || this._keyLog.length === lastItemValue.toString().length) {
+            this._pickTimeOnValue(inputValue);
+        } else {
+            this._numericInputTimeout = setTimeout(() => {
+                this._pickTimeOnValue(inputValue);
+            }, 1000);
```
platon-rov, line 334: minor: i think 300-500ms is enough.

platon-rov (approved):
Good, mentioned some minor items.

libs/core/src/lib/time/time-column/time-column.component.ts
shrvr: 500ms seems enough.

libs/core/src/lib/time/time-column/time-column.component.ts
```diff
@@ -277,7 +287,9 @@ export class TimeColumnComponent<K, T extends SelectableViewItem<K> = Selectable
         } else if (KeyUtil.isKeyCode(event, UP_ARROW)) {
             this.scrollUp();
             event.preventDefault();
-        } else if (KeyUtil.isKeyType(event, 'numeric') || KeyUtil.isKeyType(event, 'alphabetical')) {
+        } else if (KeyUtil.isKeyType(event, 'numeric')) {
+            this._numericKeyInputHandler(event);
+        } else if (KeyUtil.isKeyType(event, 'alphabetical')) {
             this._queryKeyDownEvent.next(event.key);
```
mikerodonnell89, line 293: I don't think this actually does anything when an alphabetical character is pressed, but could be cool to allow the user to switch from AM to PM with a P keypress and vice versa?

libs/core/src/lib/time/time-column/time-column.component.ts
```diff
@@ -398,6 +410,49 @@ export class TimeColumnComponent<K, T extends SelectableViewItem<K> = Selectable
         this.activeStateChange.emit();
     }
 
+    /** @hidden
+     * handles alphabetical key inputs to set period.
+     */
+    private _alphaKeyInputHandler(event: KeyboardEvent): void {
+        if (event.key === 'a' || event.key === 'A') {
```
N1XUS, line 417: You can lowercase the value so you are not depending on the letter case

N1XUS (approved):

mikerodonnell89 (approved):

PR #8730, opened 2022-09-22
fix(core): improve onTouched handling in core components and add docs

### LINKED ISSUE #8677: Step Input switches to touched state to early
Bug

The Step Input switches to touched state as soon as the user focuses the element. This is not how `touched` is defined in Angular.
> To prevent the validator from displaying errors before the user has a chance to edit the form, you should check for either the dirty or touched states in a control.
>
>When the user changes the value in the watched field, the control is marked as "dirty"
>**When the user blurs the form control element, the control is marked as "touched"**

The Datepicker and other input fields seem to handle this correctly and will only enter the `touched` state once the user removes the focus from the element.

- Angular 13.2.7
- fundamental-ngx 0.35.4

- Open https://sap.github.io/fundamental-ngx/#/core/step-input
- Scroll to the bottom of the page
- Touched is being shown correctly as `false`
- As soon as the user clicks on the field `touched` is being switched to `true`. It should only switch once the user looses focus though.

n/a

nope.

mikerodonnell89:
fixes #8677

A user of the library recently pointed out that the step input sets onTouched to true too early, and after combing through the library, I found this was the case with many components. According to the Angular documentation, onTouched for a control should be set to true after the control is blurred, not when it is first focused. For us with complex controls that means using focusout as blur only works on the parent element, while focusout works for child elements and bubbles.


platon-rov (approved):

InnaAtanasova (approved):

libs/core/src/lib/date-picker/date-picker.component.html
```diff
@@ -39,7 +39,7 @@
                 [ngModel]="_inputFieldDate"
                 (keyup.enter)="handleInputChange($any($event.target).value, false)"
                 (ngModelChange)="handleInputChange($event, true)"
-                (blur)="handleInputChange($any($event.target).value, false)"
+                (blur)="onTouched(); handleInputChange($any($event.target).value, false)"
```
g-cheishvili, line 42: create function `onBlur` and do `onTouched(); handleInputChange($any($event.target).value, false)`. Doing this on HTML side is not beautiful :D

g-cheishvili (changes requested):

Issue #6238, opened 2021-08-05
Approval flow: not started is not localized


I835291:
#### Is this a bug, enhancement, or feature request?
Yes
#### Briefly describe your proposal.
"not started" is not localized. This is reported by localization sourcing team.
https://product-jira.ariba.com/browse/SS-28661
#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
10
#### If this is a bug, please provide steps for reproducing it.
Add an approver, not started is shown as a status

Issue #5837, opened 2021-06-28
Bug: approval flow(platform) - localization issue on approval flow-2


I835291:
#### Steps:
Please see screen shots
![Create_review_type_task_added_approver](https://user-images.githubusercontent.com/86618211/123670410-0d4ade80-d7f2-11eb-904a-aca2257314ee.png)
![create_review_task_pending_hardcoded](https://user-images.githubusercontent.com/86618211/123670425-0fad3880-d7f2-11eb-8b89-6e12c467c0fc.png)
![create_review_task_add_watcher_1_moe_hardcoded](https://user-images.githubusercontent.com/86618211/123670437-1176fc00-d7f2-11eb-9e43-68f1e6e9ba01.png)
![create_review_task_add_approvers_3_more_hardcoded](https://user-images.githubusercontent.com/86618211/123670446-1340bf80-d7f2-11eb-83a4-e25faed97d6a.png)

PR #8733, opened 2022-09-23
fix(core,platform): localize hardcoded strings for tokenizer and approval flow

### LINKED ISSUE #8731: <fd-tokenizer> 'moreTerm' string is not localized
A bug
Need to localize this string

We're using Angular 13 and F-NGX 0.35.4

Using the `<fd-tokenizer>` component, add in more tokens so that they exceed the input field width, then focus out from the field until you see the `xxx more` string displayed inside the input field of the tokenizer

Screen Shot 2022-09-22 at 10 56 11 AM

https://github.com/SAP/fundamental-ngx/blob/main/libs/core/src/lib/token/tokenizer.component.ts#L111

This will cause translation issues with the `<fdp-multi-combobox>` component since itself is using the `<fd-tokenizer>`

mikerodonnell89:
fixes #8731, #5837, #6238

Prepares the tokenizer and approval flow components for translation

### DIFF
--- a/libs/core/src/lib/calendar/calendar.component.ts
+++ b/libs/core/src/lib/calendar/calendar.component.ts
@@ -2,6 +2,7 @@ import {
     ChangeDetectionStrategy,
     ChangeDetectorRef,
     Component,
+    ElementRef,
     EventEmitter,
     forwardRef,
     HostBinding,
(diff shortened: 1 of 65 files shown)


mikerodonnell89:
Closing, accidentally did not create a new branch off main

PR #8747, opened 2022-09-27
fix(ci): added is-npm dependency

g-cheishvili:
Testing if adding dev-dep helps


g-cheishvili:
This is not working, we will have to wait for firebase-tools update
https://github.com/firebase/firebase-tools/issues/5010

PR #8752, opened 2022-09-28
fix(core): Add keyboard support for mobile mode in menu

shrvr:
## Related Issue(s)

part of #8557 

## Description

Solves [this](https://github.com/SAP/fundamental-ngx/issues/8557#issuecomment-1220624224): Added  keyboard support for mobile mode in menu.

### DIFF
--- a/libs/core/src/lib/menu/menu-mobile/menu-mobile.component.ts
+++ b/libs/core/src/lib/menu/menu-mobile/menu-mobile.component.ts
@@ -4,6 +4,7 @@ import {
     Component,
     ElementRef,
     Inject,
+    NgZone,
     OnDestroy,
     OnInit,
     Optional,
@@ -11,11 +12,11 @@ import {
     ViewChild,
     ViewEncapsulation
 } from '@angular/core';
-import { DialogService } from '@fundamental-ngx/core/dialog';
+import { DialogBodyComponent, DialogService } from '@fundamental-ngx/core/dialog';
 import { Observable, of } from 'rxjs';
 import { MenuService } from '../services/menu.service';
 import { MenuItemComponent } from '../menu-item/menu-item.component';
-import { map, startWith, takeUntil } from 'rxjs/operators';
+import { map, startWith, take, takeUntil } from 'rxjs/operators';
 import { RtlService } from '@fundamental-ngx/core/utils';
 import { MENU_COMPONENT, MenuInterface } from '../menu.interface';
 import {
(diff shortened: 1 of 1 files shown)


platon-rov (commented):
two more items here

* when press "back" button being in 2nd level, you get to the 3rd level, press "back" again - you go somewhere, like you pressed on a link

https://user-images.githubusercontent.com/20265336/192984558-0dcbe080-0c70-4dc2-b4a6-74524c89af06.mov

* when you use right arrow key to get deeper to the 3rd level, then when you press left to get back to 2nd level - focus is lost

https://user-images.githubusercontent.com/20265336/192984824-76f595e1-5f13-417b-8b91-1d444bd43052.mov

libs/core/src/lib/menu/menu-mobile/menu-mobile.component.ts
```diff
@@ -82,10 +82,11 @@ export class MenuMobileComponent extends MobileModeBase<MenuInterface> implement
 
     /** Navigate back to parent level of submenu */
     backToParentLevel(): void {
-        this._menuService.setActive(
-            false,
-            this._menuService.activeNodePath[this._menuService.activeNodePath.length - 1].item
-        );
+        const activeNode = this._menuService.activeNodePath[this._menuService.activeNodePath.length - 1].item;
+        this._menuService.setActive(false, activeNode);
+        setTimeout(() => {
```
mikerodonnell89, line 87: I do not think this setTimeout is needed (can just use `activeNode?.focus();`)

mikerodonnell89 (changes requested):

mikerodonnell89:
Using setTimeout here is giving off some code smells. It would be better to wait for the presence of the desired element, or to use events, something along those lines

libs/core/src/lib/menu/services/menu.service.ts
```diff
@@ -256,7 +257,7 @@ export class MenuService {
         ) {
             if (this.focusedNode.parent?.item) {
                 this.setActive(false, this.focusedNode.parent.item);
-                this.setFocused(this.focusedNode.parent.item);
+                console.log('left arrow: focused node: ', this.focusedNode);
```
platon-rov, line 260: please remove log.

shrvr:
Current Issue (Help Needed for):

https://user-images.githubusercontent.com/65063487/197236672-432eb490-7220-4bc1-b594-4ce3ca677489.mp4

When we come back to Main Menu It focuses first item instead the item which we lastly visited.
In this case, it focused 'Fruits'  instead 'Milk Products' when we come back to main menu at last.

mikerodonnell89:
Can we have a variable to store the previously focused item's index when moving between submenus?

shrvr:
> Can we have a variable to store the previously focused item's index when moving between submenus?

Menu-service is keeping track of the `activepath` when we travel through submenus. There is already `focusedNode` that we can focus when we go backward. 
Issue is when we come back to main menu, it focuses first menu item. The `focus `function of menu-item is called one extra time when we come back to main menu.

mikerodonnell89 (approved):

PR #8753, opened 2022-09-28
fix(core): quick view should not use fd-input and should be focus trapped in popovers

mikerodonnell89:
part of defect hunting 94, this comment: https://github.com/SAP/fundamental-ngx/issues/8557#issuecomment-1215401360

Also relies on https://github.com/SAP/fundamental-styles/pull/3896


platon-rov (approved):

g-cheishvili (approved):
Looks good to me. Works like a charm

Issue #7445, opened 2021-12-27
bug(core): popover can go over the veiwport vertically


platon-rov:
#### Is this a bug, enhancement, or feature request?

Bug.

#### Briefly describe your proposal.

Popover is not limited vertically by the height of viewport. To see the footer controls we need to scroll down. In the SAPUI5 it's quite different, if popover if bigger vertically - popover's body has a scrollbar. See video for details.

Also check issue comments https://github.com/SAP/fundamental-ngx/issues/7445#issuecomment-1032452458

https://user-images.githubusercontent.com/20265336/147470977-482982c7-b8ec-45bb-8702-4aec4e7efaca.mov

It's even worse when the screen size is small, as we don't have scrollbar for popover body - we cannot see all the content and get to the popover's footer.

https://user-images.githubusercontent.com/20265336/147471126-74a5338e-a89d-4d32-9333-cb91725c760c.mov

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)

0.33.0-rc.239

#### If this is a bug, please provide steps for reproducing it.

One of many https://fundamental-ngx.netlify.app/#/core/datetime-picker
Adjust screen size to < 400px in height
Open picker

#### Please provide relevant source code if applicable.

Possible fix.

```
fitOverlayVertically() {
        const overlayBoundingClientRect = this._overlayRef.overlayElement.getBoundingClientRect();
        const viewportHeightPx = window.innerHeight;

        let newHeight = overlayBoundingClientRect.height;

        if (overlayBoundingClientRect.top < 0) {
            newHeight += overlayBoundingClientRect.top;
        }

        if (overlayBoundingClientRect.bottom > viewportHeightPx) {
            newHeight += overlayBoundingClientRect.bottom - viewportHeightPx;
        }

        if (overlayBoundingClientRect.height !== newHeight) {
            this._overlayRef.updateSize({ maxHeight: newHeight });
            this._detectChanges();
        }
    }
```

#### Is there anything else we should know?

Relevant issue https://github.com/SAP/fundamental-ngx/issues/7384

platon-rov:
Note:

When setting `overflow: auto` and scrollbars being added there is two different behaviors:

1. Chrome: popover width + scrollbar width
2. Firefox, Safari: popover width - scrollbar width

Considering this we should set `overflow: scroll` to get the Chrome's behavior, but to don't show scrollbar all the time we should track when popover goes too tall and only then apply this css rule.

[...]

PR #8755, opened 2022-09-29
fix(core): popover overflow

platon-rov:
## Related Issue(s)

Closes https://github.com/SAP/fundamental-ngx/issues/7445

## Description

Scrollbars for the popover body when it's too big for the viewport.

### Before:

image

### After:

image

_Note: Firefox doesn't support custom scrollbars, default ones are shown._

### DIFF
--- a/libs/core/src/lib/popover/popover-body/popover-body.component.scss
+++ b/libs/core/src/lib/popover/popover-body/popover-body.component.scss
@@ -9,6 +9,11 @@ $fd-popover-border-radius: var(--fdPopover_Border_Radius);
     border-#{$position}-left-radius: $radius;
 }
 
+fd-popover-body {
+    max-width: 100%;
+    max-height: 100%;
+}
+
 .#{$block} {
     &-custom {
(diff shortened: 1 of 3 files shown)


libs/core/src/lib/popover/popover-body/popover-body.component.scss
```diff
@@ -9,6 +9,11 @@ $fd-popover-border-radius: var(--fdPopover_Border_Radius);
     border-#{$position}-left-radius: $radius;
 }
 
+fd-popover-body {
+    max-width: 100%;
```
mikerodonnell89, line 13: Should these be vh/vw rather than % if we're concerned with the viewport?

platon-rov: this element is placed inside `cdk-overflay` material's element and the cdk is controlling resizing, max-height\width, we only want our element to be not bigger that its container.

g-cheishvili (approved):

Issue #7926, opened 2022-03-28
feat(docs): universal version switcher


g-cheishvili:
#### Is this a bug, enhancement, or feature request?
Enhancement

#### Briefly describe your proposal.
In general, when versions are deployed, we need to update version switcher manually with latest list of the versions. Plus when you switch from that switcher, it opens a new tab with desired version and url will not get transitioned. For instance if you're looking at the Button in v0.20 and want to see how it was in 0.19, and switch, you'll have to navigate to Button docs manually. Plus every version only has previous versions in dropdown. Meaning you can not view next versions.
Netlify has API, which gives you ability to fetch prod endpoints, but only with private API key. Their deployment version IDs are different from commits. So no way to find out deployment version from public information.

Proposal:
Build small application, which will contain only top bar, with version switcher and iframe, with current selected version's deploy url. Save user navigations between component docs and when user switches version try and save state when showing other version of it. This way every old version will have link to new one and new ones to every old versions. Switch between them will be seamless.
One gotcha: API key probably should be hidden from front-end. I think we should use Next.js and their SSR friendliness. As far as I'm aware netlify works with Next.js very well.
So in future fundamental-ngx.netlify.app would not target to fundamental-ngx repo, but to that small navigator app and that navigator app would be manager for the fundamental-ngx deployments

#### Please provide relevant source code if applicable.
related issue #7719

platon-rov:
Plus one level of complexity, I'm not sure about that. I think it can be done easier. 

After deploying we can have a workflow step that will make an api call to the netlify and get the url of the deployment. That url will be added to the url collection that stored here, in repository. Then, in our doc angular application we will just grab it, parse and show appropriate values in the dropdown.

In that way we won't be tied up to the wrapper app that may crash and make documentation app unreachable. We won't have separate title bar for the version dropdown, where dropdown itself should be the ngx component.

g-cheishvili:
@platon-rov that will work, but there will be issue with next versions. All previous versions will not have reference to the next versions

platon-rov:
@g-cheishvili why not? we will create a service that will fetch data from the file in the repository that will be updated with new versions thru the script. 

yea, it won't work for the deployments that were deployed before the service creation but there is a similar thing for the wrapper-based approach, customers who are using direct urls to get to the documentation of the old versions won't see the wrapper.

[...]

PR #8774, opened 2022-10-03
feat(docs): universal switcher

platon-rov:
## Related Issue(s)

Closes https://github.com/SAP/fundamental-ngx/issues/7926

## Description

Version switcher for the docs app, where new versions will be added automatically (for new releases).

## Screenshots

No visual changes.


.github/workflows/update-versions.yml
```diff
@@ -0,0 +1,58 @@
...
+            - name: Check if pre release
+              id: check-pre-release
+              run: |
+                  PRE_RELEASE=$(echo "${{ github.event.workflow_run.head_commit.message }}" | grep -e "-rc." -c) || :
+                  if [ $PRE_RELEASE -eq 0 ]; then
+                    echo "Not a pre-release, continuing..."
+                  else
+                    echo "Pre-release detected, skipping..."
+                  fi
+                  echo "::set-output name=pre-release::$PRE_RELEASE"
```
g-cheishvili, line 34: This can be small JS action, which will just check semver.prerelease

g-cheishvili (approved):
Left minor stuff. Good work

.github/workflows/update-versions.yml
g-cheishvili: or you can use actions/github-script@v6

libs/docs/shared/src/lib/core-helpers/toolbar/toolbar.component.ts
```diff
@@ -211,6 +191,21 @@ export class ToolbarDocsComponent implements OnInit, OnDestroy {
         this._contentDensityService.updateContentDensity(density);
     }
 
+    private _setVersions(): void {
+        this._http
+            .get<Version[]>('https://raw.githubusercontent.com/SAP/fundamental-ngx/main/versions.json')
```
N1XUS, line 196: Do we need to actually call a http request? File is in repo, and can be imported directly

platon-rov: Here is the trick, actually :) 

When new version will be released it will be added to the file, so the old deployments can show links to the newer versions.

platon-rov: Otherwise `v0.37` won't get `v0.51` deployment's url because at the moment of building the old one its `versions.json` has any versions not newer than `v0.37`.

.github/workflows/update-versions.yml
platon-rov: i think once we need it somewhere else we can extract this logic to an action.

libs/docs/shared/src/lib/core-helpers/toolbar/toolbar.component.ts
N1XUS: Got it, thanks

mikerodonnell89 (approved):

PR #8780, opened 2022-10-03
fix(core): add key control to scrollbar

shrvr:
## Related Issue(s)

part of #8557 

## Description

Solves [this](https://github.com/SAP/fundamental-ngx/issues/8557#issuecomment-1215412466): Added arrow key control to scrollbar.

### DIFF
--- a/libs/core/src/lib/scrollbar/scrollbar.component.ts
+++ b/libs/core/src/lib/scrollbar/scrollbar.component.ts
@@ -54,8 +54,9 @@ export class ScrollbarComponent {
         return this._alwaysVisible;
     }
 
+    /** @hidden */
     @HostBinding('attr.tabindex')
-    tabindex = 0;
+    _tabindex = 0;
 
     /** @hidden */
     @HostBinding('style.overflow-x')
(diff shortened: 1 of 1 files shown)


mikerodonnell89 (commented):
Works well but I wonder if this is the focus outline we want when tabbing to the container? Doesn't happen when clicking in. Not seeing anything in the designs @InnaAtanasova any thoughts
Screen Shot 2022-10-04 at 9 56 20 AM

shrvr:
> Works well but I wonder if this is the focus outline we want when tabbing to the container? Doesn't happen when clicking in. Not seeing anything in the designs @InnaAtanasova any thoughts Screen Shot 2022-10-04 at 9 56 20 AM

I think that's  the default focus `outline` by browser. We can set it to `none` for Scrollbar component in fd-styles. Let me know your thoughts. @mikerodonnell89  @InnaAtanasova

libs/core/src/lib/scrollbar/scrollbar.component.ts
```diff
@@ -54,6 +54,9 @@ export class ScrollbarComponent {
         return this._alwaysVisible;
     }
 
+    @HostBinding('attr.tabindex')
+    tabindex = 0;
```
platon-rov, line 58: typedocs or `@hidden` and `_` prefix.

platon-rov (approved):
place make scrollbar area focus outline is accordingly to the specs (apply focus mixin to the comp in styles).

PR #8805, opened 2022-10-07
fix(platform): fix broken two-columns options in select

### LINKED ISSUE #8801: bug(platform): select two columns looks broken

bug.

image

0.36.1-rc.26

shrvr:
## Related Issue(s)

closes #8801 

## Description

Platform: Fixed broken two-columns options in select.

### Before:
image

### After:
image

### DIFF
--- a/libs/platform/src/lib/form/select/select/select.component.html
+++ b/libs/platform/src/lib/form/select/select/select.component.html
@@ -64,7 +64,7 @@
 
 <ng-template let-optionItem="optionItem" let-index="index" #optionItemSource>
     <ng-container *ngIf="!_optionItemTemplate">
-        <span fd-list-title [innerHTML]="optionItem.label"></span>
+        <span fd-list-title id="primary-title" [innerHTML]="optionItem.label"></span>
     </ng-container>
 
     <ng-container *ngIf="_optionItemTemplate">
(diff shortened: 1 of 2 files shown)


libs/platform/src/lib/form/select/select/select.component.html
```diff
@@ -64,7 +64,7 @@
 
 <ng-template let-optionItem="optionItem" let-index="index" #optionItemSource>
     <ng-container *ngIf="!_optionItemTemplate">
-        <span fd-list-title [innerHTML]="optionItem.label"></span>
+        <span fd-list-title id="primary-title" [innerHTML]="optionItem.label"></span>
```
platon-rov, line 67: Having multiple same IDs is a bad practice.

libs/platform/src/lib/form/select/select/select.component.ts
```diff
@@ -101,6 +101,13 @@ export class SelectComponent extends BaseSelect implements AfterViewInit, AfterV
         // setting option items
         this.select._options.forEach((option) => {
             const optionItem = option._getHtmlElement();
+
+            // set title-text items to center - useful for two columns mainly
+            const listTitleText = optionItem.querySelector('.fd-list__title-text');
+            if (listTitleText) {
+                listTitleText.setAttribute('style', 'display:flex; align-items:center');
+            }
```
InnaAtanasova, line 109: I don't like this solution. We don't need to use `querySelector` and we can add this css to the component itself OR better fix it in the fund-styles

InnaAtanasova (changes requested):

platon-rov (approved):

InnaAtanasova (approved):

PR #8825, opened 2022-10-11
fix(docs): update versions workflow

platon-rov:
## Related Issue(s)

Follows up https://github.com/SAP/fundamental-ngx/pull/8774

## Description

Fixes for update versions workflow.

Issues fixed
* workflow triggered with `workflow_run` event didn't get the commit that made in process of previous workflow.
* getting the version done wrong.

## Screenshots

No visual changes.

### DIFF
--- a/.github/actions/wait-for-netlify/index.js
+++ b/.github/actions/wait-for-netlify/index.js
@@ -1,0 +1,137 @@
+const core = require('@actions/core');
+const axios = require('axios');
+
+const READY_STATES = ['ready', 'current'];
+
+function getNetlifyUrl(url) {
+    return axios.get(url, {
+        headers: {
+            Authorization: `Bearer ${process.env.NETLIFY_TOKEN}`
+        }
+    });
+}
+
+const waitForDeployCreation = (url, commitSha, MAX_TIMEOUT) => {
+    const increment = 15;
+
+    return new Promise((resolve, reject) => {
+        let elapsedTimeSeconds = 0;
+
+        const handle = setInterval(async () => {
+            elapsedTimeSeconds += increment;
+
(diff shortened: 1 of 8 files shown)


.github/actions/wait-for-netlify/index.js
```diff
@@ -0,0 +1,137 @@
...
+                clearInterval(handle);
+                return resolve();
+            }
+
+            console.log(`Not yet ready, waiting ${increment} more seconds...`);
+        }, increment * 1000);
+    });
+};
+
+const waitForUrl = async (url, MAX_TIMEOUT) => {
```
g-cheishvili, line 77: We have library wait-on in our dependencies, you can use that instead

g-cheishvili:
How can we test this?

g-cheishvili:
Idea is cool with netlify deployment waits. Maybe we could use netlify instead of firebase for e2e tests? In the case of firebase, we are waiting for local build to finish. We could just wait for netlify results and that's all

.github/actions/wait-for-netlify/index.js
platon-rov: I copied `probablyup/wait-for-netlify-action` action and added the commit-sha input parameter as we need it, we probably can rewrite the whole thing in a better manner, but do we want to invest time in it? :)

platon-rov:
> How can we test this?

Testing github actions is a tough deal but we can use `act` for it.

> Idea is cool with netlify deployment waits. Maybe we could use netlify instead of firebase for e2e tests? In the case of firebase, we are waiting for local build to finish. We could just wait for netlify results and that's all

https://github.com/SAP/fundamental-ngx/issues/8836

g-cheishvili (approved):

PR #8828, opened 2022-10-11
fix(fn): show selected option label in select

### LINKED ISSUE #8138: Fiori Next: Select - Renders Value instead of Option Label
Using internal values within a select component. The Options render the displayed label correctly.
Once an option is selected the select shows the internal value instead of the option label

image

```
<fn-select placeholder="Select a Parent Company" class="w-100" [(value)]="parentCompany"(ngModelChange)="onParentChanged($event)">
  <li fn-option *ngFor="let partner of partners" [value]="partner.peerId">{{ partner.displayName }}</li>
</fn-select>
```

Also tried binding with [(ngModel)].  

Version: 0.34.3

shrvr:
## Related Issue(s)

closes #8138

## Description

Added [label] property to fn-option.

### Before:
Was showing value of selected option in input-field.
image
image

### After:
image
image

### DIFF
--- a/libs/docs/fn/select/examples/select-example/select-semantic-example.component.html
+++ b/libs/docs/fn/select/examples/select-example/select-semantic-example.component.html
@@ -1,7 +1,7 @@
 <h4>Positive state</h4>
 <div style="width: 200px">
     <fn-select placeholder="Select an option" label="Standard Selector" state="positive" #selectSemanticPositive>
-        <fn-option *ngFor="let option of options" [value]="option" [label]="option"></fn-option>
+        <li fn-option *ngFor="let option of options" [value]="option" [label]="option"></li>
     </fn-select>
 </div>
 <p>
@@ -12,7 +12,7 @@
 <h4>Critical state</h4>
 <div style="width: 200px">
     <fn-select placeholder="Select an option" label="Standard Selector" state="critical" #selectSemanticCritical>
-        <fn-option *ngFor="let option of options" [value]="option" [label]="option"></fn-option>
+        <li fn-option *ngFor="let option of options" [value]="option" [label]="option"></li>
     </fn-select>
 </div>
 <p>
@@ -23,7 +23,7 @@
 <h4>Negative state</h4>
 <div style="width: 200px">
     <fn-select placeholder="Select an option" label="Standard Selector" state="negative" #selectSemanticNegative>
-        <fn-option fn-option *ngFor="let option of options" [value]="option" [label]="option"></fn-option>
+        <li fn-option fn-option *ngFor="let option of options" [value]="option" [label]="option"></li>
     </fn-select>
 </div>
 <p>
@@ -34,7 +34,7 @@
 <h4>Info state</h4>
 <div style="width: 200px">
     <fn-select placeholder="Select an option" label="Standard Selector" state="info" #selectSemanticInfo>
-        <fn-option *ngFor="let option of options" [value]="option" [label]="option"></fn-option>
+        <li fn-option *ngFor="let option of options" [value]="option" [label]="option"></li>
     </fn-select>
 </div>
 <p>
--- a/libs/docs/fn/select/examples/select-example/select-semantic-example.component.html
+++ b/libs/docs/fn/select/examples/select-example/select-semantic-example.component.html
@@ -1,7 +1,7 @@
 <h4>Positive state</h4>
 <div style="width: 200px">
     <fn-select placeholder="Select an option" label="Standard Selector" state="positive" #selectSemanticPositive>
-        <li fn-option *ngFor="let option of options" [value]="option">{{ option }}</li>
+        <fn-option *ngFor="let option of options" [value]="option" [label]="option"></fn-option>
     </fn-select>
 </div>
 <p>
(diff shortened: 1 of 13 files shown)


libs/fn/src/lib/select/option/option.component.ts
```diff
@@ -56,6 +56,13 @@ export class OptionComponent implements OnDestroy, FocusableOption {
     /** Whether this option is hidden */
     hidden = false;
 
+    /**
+     * The displayed value of the option. It shows the selected option in the select's trigger.
+     */
+    get viewValue(): string {
+        return (this.elementRef.nativeElement.textContent || '').trim();
```
N1XUS, line 63: Not really a fan of this solution. Text content may change whenever developer wants, yet with this approach those changes won't be reflected. Take a look at core's select implementation. There's option to pass custom templateRef for rendering the selected value. Another option would be to use structural directive for the title, and return it's templateRef for rendering

```html
<fn-select placeholder="Select an option" label="Standard Selector">
    <li fn-option *ngFor="let option of options" [value]="option.value">
        <span *fnSelectTitle>{{ option.title }}</span>
    </li>
</fn-select>
```

libs/fn/src/lib/select/select.component.ts
```diff
@@ -171,9 +191,11 @@ export class SelectComponent implements AfterContentInit, OnDestroy, ControlValu
         this.options.forEach((option) => {
             option === clickedOption ? (option.selected = true) : (option.selected = false);
         });
-        this.value = clickedOption.value;
+        clickedOption && this._selectionModel.select(clickedOption);
+        this.value = this.selected.value;
         this.hideMenu();
         this._cdRef.detectChanges();
+        console.log(this._selectionModel.selected);
```
N1XUS, line 198: Please remove console.log

N1XUS (changes requested):

libs/fn/src/lib/select/select.component.html
```diff
@@ -18,8 +18,8 @@
         [cdkConnectedOverlayOrigin]="trigger"
         [cdkConnectedOverlayOpen]="opened && !_optionsListEmpty"
     >
-        <ul [style.width.px]="_selectWidth" aria-label="Select menu" fnSelectMenu #menu>
+        <div [style.width.px]="_selectWidth" aria-label="Select menu" fnSelectMenu #menu>
```
mikerodonnell89, line 21: Assuming the original syntax for the options (using an `li`) can be used, I think we would want to keep this as a `ul`

mikerodonnell89 (commented):
I think it would be best to keep the original syntax so there would be no breaking changes, is changing it necessary?

libs/docs/fn/select/examples/select-example/select-semantic-example.component.html
```diff
@@ -1,7 +1,7 @@
 <h4>Positive state</h4>
 <div style="width: 200px">
     <fn-select placeholder="Select an option" label="Standard Selector" state="positive" #selectSemanticPositive>
-        <li fn-option *ngFor="let option of options" [value]="option">{{ option }}</li>
+        <fn-option *ngFor="let option of options" [value]="option" [label]="option"></fn-option>
```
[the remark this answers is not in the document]
shrvr: Yes, still it can be used.

libs/fn/src/lib/select/select.component.html
mikerodonnell89: I think it is more important that the rendered template has a `ul` with direct `li` descendants, otherwise we have invalid HTML. Accessibility tools need this `ul` `li` relationship to work properly.

libs/fn/src/lib/select/option/option.component.ts
```diff
@@ -45,6 +45,9 @@ export class OptionComponent implements OnDestroy, FocusableOption {
     @Input()
     selected = false;
 
+    @Input()
```
platon-rov, line 48: typedoc.

libs/fn/src/lib/select/select.component.html
N1XUS: @shrvr What you can do to keep the markup intact is to use ContentChildren for the fn-option, and iterate over it inside ul tag and creating own li elements

N1XUS: @mikerodonnell89 we may have issues with both options:
if we keep li[fn-option], then we'll have invalid markup in the template
If we have fn-option, we'll introduce a breaking change, yet, markup will be valid.

libs/fn/src/lib/select/option/option.component.ts
```diff
@@ -56,6 +60,14 @@ export class OptionComponent implements OnDestroy, FocusableOption {
     /** Whether this option is hidden */
     hidden = false;
 
+    /**
+     * @hidden
+     * The displayed value of the option. Used when label not passed by user. e.g. <li fn-option [value]='Apple'> Apple </li>
+     */
+    get _viewValue(): string {
```
platon-rov, line 67: We can enhance this to `return this.label ?? this.elementRef...` to remove check `option.label ? option.label : option._viewValue` in `SelectComponent`.

InnaAtanasova (approved):

N1XUS (approved):

PR #8830, opened 2022-10-12
fix(platform): table drag and drop

### LINKED ISSUE #8602: [sourcing]: fdp-table - Drag and drop is not working properly
Bug
- Unable to drag before or after a row. It always drag into.
- Unable to drag a last row to the first row as the table does not scroll.
- When drag and drop, different newIndex or previousIndex are found when the parent expanded or not expanded. This is not correct. It should give the same index regardless of the parent’s open state or not.
- It freezes when drag a row inside a parent out and put on the top of the table.
- Create a table with parent/child rows
- Drag and drop
- If drag to a child, it will insert into
- If expand the parent, newIndex is different if parent is not expanded
Had a meeting with Frank, Kevin and Manju

skaquib123:
## Related Issue(s)

closes #8602 

## Description

-Able to draw before and after a row. It is not going inside.
-Able to draw last row to first row and vice versa.

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [x] poor examples

4. Accessibility testing

##### PR Quality

-   [x] update `README.md`

### DIFF
--- a/libs/platform/src/lib/table/table.component.spec.ts
+++ b/libs/platform/src/lib/table/table.component.spec.ts
@@ -951,7 +951,7 @@ describe('TableComponent internal', () => {
             expect(hostComponent.source.fetch).toHaveBeenCalledTimes(1); // 1 means initial fetch
             const container = tableBodyContainer.nativeElement as HTMLElement;
             await tableBodyScrollTop(container.scrollHeight);
-            expect(hostComponent.source.fetch).toHaveBeenCalledTimes(2);
+            expect(hostComponent.source.fetch).toHaveBeenCalledTimes(1);
         });
 
         // TODO: flaky test  https://github.com/SAP/fundamental-ngx/issues/7534
@@ -1252,8 +1252,6 @@ class TreeTableDataProviderMock extends TableDataProvider<SourceTreeItem> {
                 const draggedRow = tableComponent._tableRowsVisible[1];
                 const row3 = tableComponent._tableRowsVisible[2];
 
-                console.log('before: ', tableComponent._tableRowsVisible);
-
                 tableComponent._dragDropItemDrop({
                     items: [],
                     replacedItemIndex: 2,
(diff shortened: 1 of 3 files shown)


platon-rov (commented):
Although this is nice to have feature, it may be better to have both modes - place before/after and place inside, so we have to somehow distinct between those two.

fkolar:
> Although this is nice to have feature, it may be better to have both modes - place before/after and place inside, so we have to somehow distinct between those two.

I agree. Let's move to this direction and maybe to unblock anybody who is using it and needs this so badly, plus currently see this as blocker or defect (that you can't simply drag now at all  - as now its limited).

 Let's take this as a part I. and the D&D "into & out" address  in the next PR. I am sure there will be other cases as well

libs/platform/src/lib/table/table.component.spec.ts
```diff
@@ -1201,8 +1201,8 @@ class TreeTableDataProviderMock extends TableDataProvider<SourceTreeItem> {
                 calculateTableElementsMetaData();
```
fkolar, line 1202: My only comment to this PR, would be add more unit tests and clearly validate before and after, first to last, last to first, last in between, first in between works correctly. I don't see any added tests here.

fkolar (changes requested):

mikerodonnell89 (commented):
Not unique to this PR, but when I drag Laptops here, why does Screens get expanded?

https://user-images.githubusercontent.com/2471874/199282575-181e23aa-0997-4608-ba4b-7f2efe709d33.mov

skaquib123:
@mikerodonnell89 That feature was already existing that of dropped row will get expanded so i did not change it. Let me know if that also need to be changed.

mikerodonnell89:
I am now only seeing changes to a test file

skaquib123:
Hi @mikerodonnell89 ,  Can you please check in the commits and let me know if you can see?
link:
[https://github.com/SAP/fundamental-ngx/pull/8830/commits/6a4df64dd428900140fb17f806e0287b0aa41416](url)

PR #8845, opened 2022-10-14
feat(core, docs): add Input() to SelectComponent's compareWith accessor

dushanpantic:
## Related Issue(s)

Resolves issue https://github.com/SAP/fundamental-ngx/issues/8846

## Description

Adds `Input()` to the SelectComponent's `compareWith` field, so the compare function can be provided via template.


dushanpantic:
image
@mikerodonnell89 I have added an example that demonstrates the usage of the `compareWith` input (image attached).
Please let me know if there is anything else that needs to be done.

libs/docs/core/select/select-docs.component.html
```diff
@@ -116,3 +116,25 @@
...
 </component-example>
 <code-example [exampleFiles]="selectForm"></code-example>
+
+<separator></separator>
+
+<fd-docs-section-title id="custom-comparator" componentName="select">
+    Select With A Custom Comparator
+</fd-docs-section-title>
+<description>
+    When the list of options provided to the Select Component consists of objects (instead of primitive elements), a
```
mikerodonnell89, line 126:
```suggestion
    When the list of options provided to the Select Component consists of objects (instead of primitive types), a
```

mikerodonnell89 (approved):

N1XUS (approved):

platon-rov (approved):

dushanpantic:
Seems like the https://github.com/SAP/fundamental-ngx/pull/8845/commits/2b322f187d56e49fba504888695536ef5b354022 commit broke the formatting rule for desired characters per line.
The change was applied straight from the "suggested change" window, which made it skip all the pre-commit hooks 😐.

I pushed a fix.

InnaAtanasova (approved):

PR #8863, opened 2022-10-17
fix(core): combobox: set value to null on clear button

### LINKED ISSUE #8858: REGRESSION Combobox: Pressing "Clear" Button Does Not Reset NgModel
bug

When using `[communicateByObject]="true"` and `[isSearch]="true"` clicking the "X" (clear button) I expect ngModel to be set to null.
Instead: The previously selected object is still in the ngModel. Only the searchTerm is cleared. 

https://user-images.githubusercontent.com/1498400/196110644-9665983e-5311-41c3-bd08-e4c3cf5a4c50.mp4

Screenshot 2022-10-17 at 09 02 10

I consider this a bug, because ngModel is reset to empty string, when `[communicateByObject]="false"`. So I expect that communicatedByObject mode will also have a resetting behaviour in regards to ngModel.

**Expected Behaviour**
Hitting the "X" button will reset searchTerm AND ngModel (as well as FromControl.value)

@fundamental-ngx/core: 0.36.0
@angular/core: 14.0.2

(@fundamental-ngx/core: 0.34.2 works as expected https://624793d2f1d02d000925c339--fundamental-ngx.netlify.app/#/core/combobox , so regression was introduced after 0.34.2)

https://stackblitz.com/edit/angular-dcxrb9?file=src%2Fapp%2Fcombobox-forms-example.component.html

https://stackblitz.com/edit/angular-dcxrb9?file=src%2Fapp%2Fcombobox-forms-example.component.html

shrvr:
## Related Issue(s)

closes #8858 

## Description

Set value to null on clear button in Combobox.

### DIFF
--- a/libs/core/src/lib/combobox/combobox.component.ts
+++ b/libs/core/src/lib/combobox/combobox.component.ts
@@ -699,10 +699,14 @@ export class ComboboxComponent
 
     /** @hidden */
     private _propagateChange(): void {
-        if (!this.communicateByObject) {
-            this.onChange(this.inputText);
-        } else {
+        if (this.communicateByObject) {
+            const value = this._getOptionObjectByDisplayedValue(this.inputText);
+            if (this.displayFn(value) !== this.displayFn(this.getValue())) {
+                this.setValue(value);
+            }
             this.onChange(this.getValue());
+        } else {
+            this.onChange(this.inputText);
         }
     }
 
--- a/libs/core/src/lib/combobox/combobox.component.spec.ts
+++ b/libs/core/src/lib/combobox/combobox.component.spec.ts
@@ -212,8 +212,8 @@ describe('ComboboxComponent', () => {
         expect(addOns.length).toBe(1);
         component.isSearch = true;
         component.communicateByObject = true;
-        component.displayFn = (item: any): string => item.displayedValue;
-        component.inputText = 'displayedValue2';
+        component.displayFn = (item: any): string => item?.displayedValue ?? '';
+        component.inputText = 'test';
         (<any>component)._cdRef.detectChanges();
         addOns = fixture.nativeElement.querySelectorAll('button');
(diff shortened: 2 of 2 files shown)


libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -534,6 +534,10 @@ export class ComboboxComponent
 
     /** @hidden */
     handleSearchTermChange(): void {
+        if (this.isEmptyValue) {
+            this._handleClearSearchTerm();
+        }
+
```
desaroxx, line 540: before the regression the old behaviour was to set the value to `null` as soon as the label aka searchTerm is not an exact match to any items in the dropdown list.


https://user-images.githubusercontent.com/1498400/196356169-5881990b-e7a0-4f77-b025-5b0a16d5f4e1.mp4

libs/core/src/lib/combobox/combobox.component.ts
```diff
@@ -705,10 +701,12 @@ export class ComboboxComponent
 
     /** @hidden */
     private _propagateChange(): void {
-        if (!this.communicateByObject) {
-            this.onChange(this.inputText);
-        } else {
+        if (this.communicateByObject) {
+            const value = this._getOptionObjectByDisplayedValue(this.inputText);
```
mikerodonnell89, line 705: If we want to communicate by object, we don't want to just get the object by the displayed value. Take a look at the reactive example, which uses these two values:
```
        { displayedValue: 'Apple', value: 'AppleValue' },
        { displayedValue: 'Apple', value: 'AppleValue2' },
```
Now we can no longer select the option with AppleValue2. You'll see this is working on main

mikerodonnell89 (changes requested):

mikerodonnell89 (approved):

InnaAtanasova (approved):

Issue #8728, opened 2022-09-22
[Wizard when Branching] Review pages listed also not visible inputs


droshev:
#### Is this a bug, enhancement, or feature request?
bug

#### Briefly describe your proposal.
We have created a wizards content, with when conditions, and
the field Namespace is shown when project is selected and not shown when library is selected.

In preview Page we see parameters for the two possible inputs

`dxp/template-ui/issues/245`

we will try to tackle it in sprint 98

platon-rov:
cannot reproduce, dxp uses outdated ngx version.

platon-rov:
reproduced, fixed and raised pr

PR #8891, opened 2022-10-24
fix(platform): wizard generator hidden form groups visible

platon-rov:
## Related Issue(s)

Refers to DXP.
Closes https://github.com/SAP/fundamental-ngx/issues/8728

## Description

Fix for the issue when hidden form groups being shown on the summary step and in the resulting object.

Prerequisite
Screenshot 2022-10-24 at 15 04 42

### Before:
Screenshot 2022-10-24 at 15 04 51

### After:
Screenshot 2022-10-24 at 15 09 20

### DIFF
--- a/libs/platform/src/lib/wizard-generator/components/wizard-generator-step/wizard-generator-step.component.ts
+++ b/libs/platform/src/lib/wizard-generator/components/wizard-generator-step/wizard-generator-step.component.ts
@@ -197,18 +197,15 @@ export class WizardGeneratorStepComponent implements OnInit, OnDestroy, OnChange
     }
 
     /**
-     * @returns Step submitted forms values.
+     * Returns visible forms in the step.
      */
-    getFormValues(): WizardStepSubmittedForms {
-        return this._submittedForms;
-    }
-
-    /**
-     *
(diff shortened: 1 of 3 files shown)


libs/platform/src/lib/wizard-generator/components/wizard-generator-step/wizard-generator-step.component.ts
```diff
@@ -197,18 +197,15 @@ export class WizardGeneratorStepComponent implements OnInit, OnDestroy, OnChange
     }
 
     /**
-     * @returns Step submitted forms values.
+     * Returns visible forms in the step.
      */
-    getFormValues(): WizardStepSubmittedForms {
```
droshev, line 202: is this a breaking change?

platon-rov: wouldn't say so, although it's exposed we don't use `fdp-wizard-generator-step` anywhere

InnaAtanasova (approved):

droshev (approved):

Issue #8893, opened 2022-10-24
Defect Hunting Sprint 99


InnaAtanasova:
Please list the issues you've found during the defect hunting.

shrvr:
- [ ] Arrow overflows in popover

![image](https://user-images.githubusercontent.com/65063487/197872676-fff452f5-c10a-47b6-a232-8b4537da971a.png)

PR #8900, opened 2022-10-26
fix(core): solve popover arrow overflow

shrvr:
## Related Issue(s)

part of https://github.com/SAP/fundamental-ngx/issues/8893#issuecomment-1291088741

## Description

Solved popover arrow overflow.
Fixed double scrollbar for popover with list.

### Before:
image
image

### After:
image
image

### DIFF
--- a/libs/core/src/lib/popover/popover-body/popover-body.component.html
+++ b/libs/core/src/lib/popover/popover-body/popover-body.component.html
@@ -9,7 +9,7 @@
     [style.min-width.px]="_popoverBodyMinWidth"
     [style.width.px]="_popoverBodyWidth"
 >
-    <div fd-scrollbar style="max-height: 100%; max-width: 100%; width: 100%">
+    <div fd-scrollbar>
         <ng-content select="[fd-popover-body-header]"></ng-content>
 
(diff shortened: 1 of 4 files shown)


libs/core/src/lib/popover/popover-body/popover-body.component.html
```diff
@@ -10,17 +9,19 @@
     [style.min-width.px]="_popoverBodyMinWidth"
     [style.width.px]="_popoverBodyWidth"
 >
-    <ng-content select="[fd-popover-body-header]"></ng-content>
+    <div fd-scrollbar style="max-height: 100%; max-width: 100%; width: 100%">
```
platon-rov, line 12: Please don't use inline styles.

platon-rov (commented):
Nice work!

libs/core/src/lib/popover/popover-body/popover-body.component.html
InnaAtanasova: +1 on this. There shouldn't be inline css

InnaAtanasova (changes requested):

platon-rov (approved):

InnaAtanasova (approved):

PR #8926, opened 2022-11-01
fix(core): remove initial-focus directive from menu

shrvr:
## Related Issue(s)

closes none

Menu uses `initial-focus` directive to focus first menu-item. It doesn't allow user if he/she don't want to focus first-item and keep focus somewhere else. 

## Description

Instead of using `initial-focus`, added `focusAutoCapture` from `popover` to focus first menu-item.  

Ultimately, it's same behavior (Dropdown opens - focus moves to its first item),  but in addition, it enables users to prevent focus on first item by making `focusAutoCapture` false. It's same as in Action Sheet component.

### DIFF
--- a/libs/core/src/lib/menu/menu.component.ts
+++ b/libs/core/src/lib/menu/menu.component.ts
@@ -216,7 +216,7 @@ export class MenuComponent
         }
 
         this._changeDetectorRef.markForCheck();
-        this.menuItems.first.elementRef.nativeElement.focus();
+        this._menuItems.first.elementRef.nativeElement.focus();
     }
 
     /** Closes the menu */
--- a/libs/core/src/lib/menu/menu.component.ts
+++ b/libs/core/src/lib/menu/menu.component.ts
@@ -216,6 +216,7 @@ export class MenuComponent
         }
 
         this._changeDetectorRef.markForCheck();
+        this.menuItems.first.elementRef.nativeElement.focus();
     }
 
     /** Closes the menu */
--- a/libs/core/src/lib/menu/menu.component.ts
+++ b/libs/core/src/lib/menu/menu.component.ts
@@ -72,6 +72,13 @@ export class MenuComponent
     @Input()
     focusTrapped = true;
 
+    /**
+     * Whether the popover should automatically move focus into the trapped region upon
+     * initialization and return focus to the previous activeElement upon destruction.
+     */
+    @Input()
+    focusAutoCapture = true;
+
     /** Open submenu on hover after given milliseconds */
     @Input()
     openOnHoverTime = 0;
--- a/libs/core/src/lib/menu/menu.component.html
+++ b/libs/core/src/lib/menu/menu.component.html
@@ -3,8 +3,6 @@
 <ng-template #menuRootTemplate>
     <nav
         class="fd-menu"
-        fdInitialFocus
-        [enabled]="focusTrapped"
         [id]="id"
         [attr.aria-labelledby]="ariaLabelledby"
         [attr.aria-label]="ariaLabel"
--- a/libs/core/src/lib/popover/popover-body/popover-body.component.ts
+++ b/libs/core/src/lib/popover/popover-body/popover-body.component.ts
@@ -26,7 +26,6 @@ import {
 } from '@fundamental-ngx/core/shared';
 import { KeyUtil } from '@fundamental-ngx/core/utils';
 import { ContentDensityObserver, contentDensityObserverProviders } from '@fundamental-ngx/core/content-density';
-import { ScrollbarComponent } from '@fundamental-ngx/core/scrollbar';
 
 /**
  * A component used to enforce a certain layout for the popover.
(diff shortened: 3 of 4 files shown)


platon-rov (changes requested):
I don't understand the issue that been fixed, sorry. We're selecting the first menu item after menu list popover got opened, that's correct I think. Then we can use ESC button to close menu and get focus back to menu button. What's wrong in this behavior?

InnaAtanasova:
> I don't understand the issue that been fixed, sorry. We're selecting the first menu item after menu list popover got opened, that's correct I think. Then we can use ESC button to close menu and get focus back to menu button. What's wrong in this behavior?

I am also confused about this issue. @shrvr let's discuss it during the scrum.

shrvr:
> 

Problem is when it focuses the first item. It uses `initial-focus` directive that focuses first element when zone gets stable. This was causing this issue https://github.com/SAP/fundamental-ngx/pull/8752#issuecomment-1287151248. So, when we come back to main menu it focuses 'Fruits' even go to 'Milk Products' and come back. This is the one issue I found using `initial-focus` directive.

Now, also that user don't want to `focus `first menu item, so this `initial-focus` don't allow it. But, if we use `focusAutoCapture`, user can make it `false ` and can keep `focus` wherever he/she wants. It gives more control to user. Moreover, we already extending `BasePopoverClass` to `MenuComponent`, so why don't we just use its property directly.

shrvr:
Reverted last commit for removing `tabindex` from scrollbar. So, currently it won't focus first menu item. @InnaAtanasova @platon-rov

InnaAtanasova (approved):

libs/core/src/lib/menu/menu.component.ts
```diff
@@ -69,6 +69,13 @@ export class MenuComponent
     @Input()
     focusTrapped = true;
 
+    /**
+     * Whether the popover should automatically move focus into the trapped region upon
+     * initialization and return focus to the previous activeElement upon destruction.
+     */
+    @Input()
+    focusAutoCapture = true;
```
platon-rov, line 77: Why we need this? 

We should not be able to change the correct behavior. Dropdown opened - focus moved to its first item, it's a rule of thumb on my opinion.

libs/core/src/lib/menu/menu.component.ts
```diff
@@ -205,6 +212,7 @@ export class MenuComponent
             this._popoverService.open();
         }
         this._changeDetectorRef.markForCheck();
+        this.menuItems.first.elementRef.nativeElement.focus();
```
mikerodonnell89, line 215: This doesn't seem to be working


https://user-images.githubusercontent.com/2471874/200667299-52207cbc-bef0-4668-a471-2af26eab1908.mov

mikerodonnell89 (changes requested):
See vid below

libs/core/src/lib/menu/menu.component.ts
shrvr: Yes, ultimately, it's same behavior, but in addition, it enables users to prevent focus on first item by making `focusAutoCapture` false. It's same as in `Action Sheet` component. 
@platon-rov

libs/core/src/lib/menu/menu.component.ts
shrvr: oh yes, sorry, I forgot to mention here. 

So, the issue was, when `focusAutoCapture` is `true`, and `popover` opens, it focuses first `tabbable` element which is scrollbar (so not focusing first item). To solve this, I had added this commit (5ed73078a9cf25024fc22daf430cb33a74613b2c) but now to handle `tabindex` of scrollbar we have separate issue https://github.com/SAP/fundamental-ngx/issues/8929, so I reverted this commit.

But all these still were causing the test fail (`focus first element on open`), so I added this line to just pass the test.

Let me know your thought on this.
@mikerodonnell89 @InnaAtanasova @platon-rov

libs/core/src/lib/menu/menu.component.ts
platon-rov: Okay, but where it is used?

platon-rov (changes requested):

libs/core/src/lib/menu/menu.component.ts
```diff
@@ -209,6 +216,7 @@ export class MenuComponent
         }
 
         this._changeDetectorRef.markForCheck();
+        this._menuItems.first.elementRef.nativeElement.focus();
```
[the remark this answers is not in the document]
shrvr: @platon-rov https://github.com/SAP/fundamental-ngx/pull/8926#discussion_r1017144596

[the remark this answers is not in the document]
shrvr: Replaying to https://github.com/SAP/fundamental-ngx/pull/8926#discussion_r1018910943 :

`focusAutoCapture` is a derived property of popover(`BasePopoverClass`) so it is handled by popover service.

platon-rov (commented):
Still my opinion is that we should not give users ability to change the default browser's behavior. Let me raise this on daily.

platon-rov (approved):

shrvr:
Merging the logic to PR - https://github.com/SAP/fundamental-ngx/pull/8964

PR #8967, opened 2022-11-16
fix(core,platform): fund-styles popover's changes adoption

platon-rov:
## Related Issue(s)

Closes none.

## Description

This PR isn't a complete refactoring but only contains changes to adopt respective ones from https://github.com/SAP/fundamental-styles/pull/4026. Also, some minor things were automatically solved.

### Before:

CleanShot 2022-11-16 at 18 04 15@2x

### After:

CleanShot 2022-11-16 at 18 03 42@2x


g-cheishvili (approved):

mikerodonnell89 (commented):
I think the popover w/ arrow now overlapping its control is incorrect:

Before:
Screen Shot 2022-11-29 at 11 15 44 AM

After:
Screen Shot 2022-11-29 at 11 15 51 AM

libs/core/src/lib/popover/popover-body/popover-body.component.ts
```diff
@@ -82,7 +75,7 @@ export class PopoverBodyComponent {
     _closeOnEscapeKey = false;
 
     /** Classes added to arrow element */
-    _arrowClasses: string[] = [];
+    _arrowClasses: string;
```
mikerodonnell89, line 78: Should be private as well as quite a few other properties in this class

platon-rov:
> I think the popover w/ arrow now overlapping its control is incorrect:

https://github.com/SAP/fundamental-styles/pull/4057 Here is the fix.

libs/core/src/lib/popover/popover-body/popover-body.component.ts
platon-rov: This property should be public as it's used in template.

Issue #8982, opened 2022-11-22
feat(docs): unify approach to show examples in compact condensity mode


platon-rov:
#### Is this a bug, enhancement, or feature request?

Enhancement.

#### Briefly describe your proposal.

We don't have a unified approach to show components which support various content densities, somewhere we have separate examples with hard-coded density somewhere else we don't and switch in header should be used to switch density globally and some components just don't support densities. 

I think it can be unified easily

1. Remove all the hardcoded density examples.
2. Add labels to the components which support various content densities

Additionally, review where content density support should be added, but it's out of scope of this issue.

PR #8983, opened 2022-11-22
fix(docs): removing of hardcoded density

platon-rov:
## Related Issue(s)

Refers to https://github.com/SAP/fundamental-ngx/issues/8982

## Description

Removed hard-coded content density from some examples where it's not supposed.

### DIFF
--- a/libs/docs/core/calendar/examples/calendar-mobile-example/calendar-mobile-example.component.html
+++ b/libs/docs/core/calendar/examples/calendar-mobile-example/calendar-mobile-example.component.html
@@ -14,7 +14,7 @@
             <fd-button-bar
                 fdType="emphasized"
                 label="Ok"
-                [fdCompact]="false"
+                fdCozy
                 (click)="dialog.close('Continue'); dateChanged(landscapeCalendar.selectedDate)"
             >
             </fd-button-bar>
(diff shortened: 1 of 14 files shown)


libs/docs/core/calendar/examples/calendar-mobile-example/calendar-mobile-example.component.html
```diff
@@ -14,7 +14,7 @@
             <fd-button-bar
                 fdType="emphasized"
                 label="Ok"
-                [fdCompact]="false"
+                fdCozy
```
InnaAtanasova, line 17: isn't this also "hardcoding" the density?

InnaAtanasova (commented):
Just one comment for fdCozy, the rest looks great

g-cheishvili (approved):

g-cheishvili:
e2e fail, because there are tests for compacts

mikerodonnell89 (commented):
Should we be removing entirely any examples that solely exist to show the hardcoded density? It is odd that some have components have this and some do not

droshev:
> Should we be removing entirely any examples that solely exist to show the hardcoded density? It is odd that some have components have this and some do not

I see @mikerodonnell89 's point. There is no point have separate examples for cozy and compact.

platon-rov:
> I see @mikerodonnell89 's point. There is no point have separate examples for cozy and compact.

yea, we have ticket for it https://github.com/SAP/fundamental-ngx/issues/8982, you said that would be good task for new intern

PR #9010, opened 2022-11-28
fix(core,platform): button toggled state

platon-rov:
## Related Issue(s)

Closes none.
Refers to https://github.tools.sap/dxp/backlog/issues/812

## Description

Toggled button state refactoring after changes in fundamental-styles.

## Screenshots

No visual changes.

## BREAKING CHANGE:
* using of `toggled` input property instead of `aria-selected` attribute is now required


InnaAtanasova (approved):

libs/platform/src/lib/button/button.component.ts
```diff
@@ -43,13 +43,18 @@ export class ButtonComponent extends BaseComponent implements AfterViewInit {
...
     /** arialabel, tooltip for truncated text
      * for acccesiblity of the element */
     @Input()
     title: Nullable<string>;
 
-    /** aria-selected for acccesiblity to
-     *  the native HTML button*/
+    /** @deprecated use toggled input property instead
+     * aria-selected for acccesiblity to the native HTML button
+     */
```
g-cheishvili, line 57: Please add documentation entry like it is for Content Density, so that user sees it right when they arrive at the docs page

libs/platform/src/lib/button/button.component.ts
```diff
@@ -72,9 +77,8 @@ export class ButtonComponent extends BaseComponent implements AfterViewInit {
     @Input()
     ariaControlsId: Nullable<string>;
 
-    /**
-     * propagate aria-pressed for accessiblity to
-     * the native HTML button
+    /** @deprecated use toggled input property instead
+     * propagate aria-pressed for accessiblity to the native HTML button
```
g-cheishvili, line 81: Documentation entry

g-cheishvili (changes requested):

mikerodonnell89 (approved):

g-cheishvili (approved):

PR #9079, opened 2022-12-11
chore(docs): got rid of unnecessary logs

g-cheishvili:
## Related Issue(s)

closes probably none

## Description

- added commonJS dependencies to allowed list
- disabled `strictNullChecks`. It was always warning, even if object property was optional because it was strictly checking if it was `undefined` or `null` and as a result, annoying and unnecessary logs were produced

### DIFF
--- a/apps/docs/project.json
+++ b/apps/docs/project.json
@@ -163,7 +163,16 @@
                     }
                 ],
                 "scripts": ["node_modules/marked/lib/marked.js"],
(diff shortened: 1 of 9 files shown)


droshev (approved):

g-cheishvili:
After discussion with @N1XUS we came to conclusion that data types need changes, rather than tsconfig

Issue #9011, opened 2022-11-29
fd-action-sheet does not allow to use custom icons


tomsawdayee:
#### Is this a bug, enhancement, or feature request?
Feature request

#### Briefly describe your proposal.

I'm trying to use an external icon library, and while most of the components allow me to use custom icons, I found that the Action-sheet control does not - since theres not control over the button that is rendered within **fd-action-sheet-item**
I would like to be able to gain control over the button that is rendered so I can replace the icon with a custom one.

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Current version - 0.36.2

platon-rov:
@tomsawdayee you can update to 0.38.2 or higher and use custom icons like so

```html
<fd-action-sheet-item label="label">
  <my-custom-icon></my-custom-icon>
</fd-action-sheet-item>
```

tomsawdayee:
@platon-rov Great news! That's exactly what I was looking for. Thanks a bunch!

PR #9092, opened 2022-12-14
feat(core): custom icon placement on core button

platon-rov:
## Related Issue(s)

Relates to https://github.com/SAP/fundamental-ngx/issues/9011

## Description

Feature to have custom icon or any content on button **before** the text.

## Screenshots

CleanShot 2022-12-14 at 18 51 20@2x

### DIFF
--- a/libs/core/src/lib/button/button.component.html
+++ b/libs/core/src/lib/button/button.component.html
@@ -1,15 +1,19 @@
-<ng-container *ngIf="glyphPosition === 'before' && glyph">
-    <fd-icon [glyph]="glyph"></fd-icon>
+<ng-container *ngIf="glyphPosition === 'before'">
+    <ng-container [ngTemplateOutlet]="iconTemplate"></ng-container>
 </ng-container>
 
 <span class="fd-button__text" *ngIf="label">
     {{ label }}
 </span>
 
-<ng-content></ng-content>
-
-<ng-container *ngIf="glyphPosition === 'after' && glyph">
-    <fd-icon [glyph]="glyph"></fd-icon>
+<ng-container *ngIf="glyphPosition === 'after'">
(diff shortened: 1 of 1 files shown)


libs/core/src/lib/button/button.component.html
```diff
@@ -1,15 +1,19 @@
...
+<ng-container *ngIf="glyphPosition === 'after'">
+    <ng-container [ngTemplateOutlet]="iconTemplate"></ng-container>
 </ng-container>
 
 <fd-icon glyph="slim-arrow-down" *ngIf="fdMenu"></fd-icon>
+
+<ng-template #iconTemplate>
+    <fd-icon *ngIf="glyph" [glyph]="glyph"></fd-icon>
+
+    <ng-content></ng-content>
```
N1XUS, line 18: Maybe keep original ng-content, and use this one with [select]?
There might be cases when button is used like so:
```
<button fd-button>
   Some inner text
   <fd-icon...></fd-icon>
</buton>```

N1XUS (changes requested):

libs/core/src/lib/button/button.component.html
platon-rov: Adding `[select]` is fine, though it adds one more detail to think of. 

Your example won't be affected with current changes, but still no any new details added.

platon-rov: `glyphPosition` is `before` by default, so we show projected content even if nothing (like in your example) specified, neither glyph, label or glyphPosition.

InnaAtanasova:
Correct me if I am wrong, but our buttons have the option to add an icon. The problem the user is facing is not having access to the button in Action Sheet, which prevents them from adding an icon.
Screenshot 2022-12-15 at 11 36 07 AM

mikerodonnell89 (approved):

PR #9154, opened 2022-12-29
test: removed build targets from small projects

g-cheishvili:
## Description
Removed `build` targets from small projects, things like `core/button` and etc. Let's see the potential speed improvements


g-cheishvili:
Result: It is technically able to build, but linting does not like the fact that some libraries are not buildable... Even though it will not have any issues

g-cheishvili:
This is not needed

Issue #8897, opened 2022-10-24
Defect Hunting Sprint 103


InnaAtanasova:
Please list the issues you've found during the defect hunting.

shrvr:
- [ ] popover arrow is missing

![image](https://user-images.githubusercontent.com/65063487/208968538-bc13550e-e2f4-4109-ba07-c63dafc36a9a.png)

shrvr:
- [ ] [Submenu](https://fundamental-ngx.netlify.app/#/core/menu#with-submenu) overflows

![image](https://user-images.githubusercontent.com/65063487/208982875-a5559972-fa2a-4410-9473-afec354ea9ca.png)

platon-rov:
* [ ] Core Feed List Item "more" is not focusable 
<img width="621" alt="CleanShot 2022-12-27 at 16 26 27@2x" src="https://user-images.githubusercontent.com/20265336/209691301-d3e22c9a-925a-4c6b-ad58-8cd70aa4957d.png">

platon-rov:
* [ ] [imho] Grid List Card not needed focus

<img width="224" alt="CleanShot 2022-12-28 at 11 33 05@2x" src="https://user-images.githubusercontent.com/20265336/209798791-f2be4275-4534-43a3-98fb-8570760cc7f0.png">

platon-rov:
* [ ] Core Input Group incorrect hover/active state of button, HC themes

<img width="106" alt="CleanShot 2022-12-28 at 11 42 58@2x" src="https://user-images.githubusercontent.com/20265336/209799920-f4a99836-efa1-4c9a-8feb-8cb466bd5277.png">
<img width="95" alt="CleanShot 2022-12-28 at 11 43 23@2x" src="https://user-images.githubusercontent.com/20265336/209799969-c03433cb-4113-491b-97ce-49f75b2d3b05.png">

Specs:
<img width="283" alt="CleanShot 2022-12-28 at 11 43 36@2x" src="https://user-images.githubusercontent.com/20265336/209800002-dfbf72b2-c17b-4ac2-b35d-2493e12c8fb0.png">

platon-rov:
* [ ] Seems like input focus/button active state is incorrect

<img width="719" alt="CleanShot 2022-12-28 at 11 49 38@2x" src="https://user-images.githubusercontent.com/20265336/209800736-07887626-3845-4204-9d9d-b0541506c635.png">
[...]

PR #9157, opened 2022-12-29
fix(core): remove timeline node tab index

mikerodonnell89:
part of #8897

### DIFF
--- a/libs/core/src/lib/dynamic-page/dynamic-page.component.scss
+++ b/libs/core/src/lib/dynamic-page/dynamic-page.component.scss
@@ -27,7 +27,7 @@
     height: 100%;
 
     .fd-bar--floating-footer {
-        z-index: 2;
+        z-index: 6;
     }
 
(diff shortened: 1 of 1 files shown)


libs/core/src/lib/dynamic-page/dynamic-page.component.scss
```diff
@@ -27,7 +27,7 @@
     height: 100%;
 
     .fd-bar--floating-footer {
-        z-index: 2;
+        z-index: 6;
```
platon-rov, line 30: Looks like it's not related to timeline.

platon-rov (changes requested):

libs/core/src/lib/dynamic-page/dynamic-page.component.scss
mikerodonnell89: Wrong branch, thanks!

PR #9158, opened 2022-12-30
fix(core): feed list item more el tabindex

platon-rov:
## Related Issue(s)

Refers to https://github.com/SAP/fundamental-ngx/issues/8897#issuecomment-1366010523

## Description

Feed List Item more element tab index issue fix.

### Before:

Not tabbable.

### After:

Tabbable.


g-cheishvili (approved):
This is strange, it should have been tabbable without explicit tabindex

platon-rov:
> This is strange, it should have been tabbable without explicit tabindex

`a` without `href` is not tabbable :)

PR #9160, opened 2022-12-30
fix(core,platform,fn): project architechture & ci optimizations

platon-rov:


g-cheishvili:
If you delete all the small projects, you will not have that granular control and understanding of the changes that we have right now.
The situation right now:
- change tabs component
- All the components depending on `tabs` and also documentation depending on that will be affected and retested

Situation without `tabs` project:
- change tabs component
- everything is retested and rebuilt because `tabs` is not changed, but `core` is and on core depends most of the things

platon-rov:
> If you delete all the small projects, you will not have that granular control and understanding of the changes that we have right now. The situation right now:
> 
> * change tabs component
> * All the components depending on `tabs` and also documentation depending on that will be affected and retested
> 
> Situation without `tabs` project:
> 
> * change tabs component
> * everything is retested and rebuilt because `tabs` is not changed, but `core` is and on core depends most of the things

I'm just fooling around :D

PR #9195, opened 2023-01-09
fix(core): date-picker respects closeOnDateChoose input now

### LINKED ISSUE #9060: Close popup on Date range selection
Bug

We are using date range component and we found out that the popup closes on escape or on outside click. Ideally, it should allow the user to select start-end date on enter and close the popup with values selected.

0.37.1-rc.0

Choose date range component.
open the date range popup;
select start & end date ;
Press enter on end date;
Popup stays open, ideally it should close;

g-cheishvili:
## Related Issue(s)

closes #9060 

## Description

The calendar would alter date-range values, as a result, there was no way of knowing whether or not the user had finished selecting dates, and also date-picker was not respecting `closeOnDateChoose` input.

### DIFF
--- a/libs/core/src/lib/calendar/calendar-views/calendar-day-view/calendar-day-view.component.ts
+++ b/libs/core/src/lib/calendar/calendar-views/calendar-day-view/calendar-day-view.component.ts
@@ -51,9 +51,11 @@ export class CalendarDayViewComponent<D> implements OnInit, OnChanges, OnDestroy
             this._buildDayViewGrid();
         }
     }
+
     get currentlyDisplayed(): CalendarCurrent {
         return this._currentlyDisplayed;
     }
+
     /** @hidden */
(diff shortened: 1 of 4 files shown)


N1XUS (approved):

mikerodonnell89 (changes requested):
I think we should try to mirror ui5's behavior with how the input is populated when selecting dates in a range. Here, after opening a range datepicker and clicking a date, the input displays "Invalid Date" for the end date, and if the user clicks outside the popover before selecting an end date, the end date is "null". UI5 does not update the selected date range until the user has selected a valid start and end date in the popover

g-cheishvili:
> 

We can do that in nearest future, now it just requires a lot of changes. Let's use it this way, especially if you consider the fact that input was there before, just it was not respected by the component itself, so it was a bug

mikerodonnell89 (approved):
We'll address my previous comment in another PR

PR #9204, opened 2023-01-10
fix(cx): fix prod build and add schematics

mikerodonnell89:
fixes none


droshev (approved):

libs/cx/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,169 @@
+import { Rule, SchematicContext, Tree, chain, SchematicsException } from '@angular-devkit/schematics';
+import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
+import { getPackageVersionFromPackageJson, hasPackage } from '../utils/package-utils';
+import { addPackageJsonDependency, NodeDependency, NodeDependencyType } from '@schematics/angular/utility/dependencies';
+import { hasModuleImport } from '../utils/ng-module-utils';
+import { WorkspaceSchema } from '@schematics/angular/utility/workspace-models';
+
+import { defaultFontStyle } from './styles';
+import { addModuleImportToModule, findModuleFromOptions } from '@angular/cdk/schematics';
```
N1XUS, line 9: If user installed @fundamental-ngx/cx without installing @fundamental-ngx/core, @angular/cdk package might not be installed on the users application. Check how adding dependencies is done in core. We've splitted schematic steps into separate independent and hidden schematics, which allows us to first install dependencies, and then use code from those dependencies.

PR #9206, opened 2023-01-11
fix: update release badge for correct workflow

N1XUS:
Tiny fix for readme files for correct build workflow badge


g-cheishvili (approved):
Oh great, I wanted to do that and constantly forget about it :D

PR #9211, opened 2023-01-12
fix(core,cx): nested list icon should extend fd-icon

### LINKED ISSUE #8902: Support BusinessSuiteInAppSymbols and TNT icons in Side Navigation with Icons
Feature request

We want to use icons from BusinessSuiteInAppSymbols in the glyph of the Side Navigation with Icons(https://sap.github.io/fundamental-ngx/#/core/side-navigation#icons). Currently, only icons from SAP-icons are supported. Similar to the [glyph](https://sap.github.io/fundamental-ngx/#/core/icon#icon-business-suite), a font property would be needed.

FD 0.36.0 (All versions)

[https://github.com/SAP/fundamental-ngx/blob/459f4420034af8b39ab0853da5ff6c604883b26b/libs/docs/core/side-navigation/examples/side-navigation-icons-example.component.html#L6](https://github.com/SAP/fundamental-ngx/blob/459f4420034af8b39ab0853da5ff6c604883b26b/libs/docs/core/side-navigation/examples/side-navigation-icons-example.component.html#L6)

mikerodonnell89:
fixes #8902 

<code>NestedListIconDirective</code> becomes <code>NestedListIconComponent</code> and extends <code>IconComponent</code>, allowing devs to pass a <code>[font]</code> input to change the icon set. No breaking changes to existing templates

### DIFF
--- a/libs/core/src/lib/nested-list/nested-list-directives.ts
+++ b/libs/core/src/lib/nested-list/nested-list-directives.ts
@@ -7,17 +7,14 @@ import {
     HostBinding,
     HostListener,
     Input,
-    OnChanges,
-    OnInit,
     Optional,
     ViewEncapsulation
 } from '@angular/core';
 import { NestedItemService } from './nested-item/nested-item.service';
-import { applyCssClass } from '@fundamental-ngx/cdk/utils';
-import { CssClassBuilder } from '@fundamental-ngx/cdk/utils';
 import { RtlService } from '@fundamental-ngx/cdk/utils';
 import { map } from 'rxjs/operators';
 import { Observable, of } from 'rxjs';
+import { IconComponent } from '@fundamental-ngx/core/icon';
 
 let uniqueId = 0;
 
@@ -43,20 +40,12 @@ export class NestedListHeaderDirective {
     }
 }
 
-@Directive({
-    selector: '[fdNestedDirectivesIcon], [fd-nested-list-icon]'
+@Component({
+    // eslint-disable-next-line @angular-eslint/component-selector
+    selector: '[fdNestedDirectivesIcon], [fd-nested-list-icon]',
+    template: `<ng-content></ng-content>`
 })
-export class NestedListIconDirective implements CssClassBuilder, OnChanges, OnInit {
-    /** The property allows user to pass additional css classes */
-    @Input()
-    class = '';
-
-    /**
-     * The icon name to display. See the icon page for the list of icons
-     * here: https://sap.github.io/fundamental-ngx/icon
-     */
-    @Input() glyph: string;
-
+export class NestedListIconComponent extends IconComponent {
     /** Role attribute */
     @Input()
     @HostBinding('attr.role')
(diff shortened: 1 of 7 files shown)


libs/core/src/lib/nested-list/nested-list-directives.ts
```diff
@@ -65,30 +54,6 @@ export class NestedListIconDirective implements CssClassBuilder, OnChanges, OnIn
     /** @hidden */
     @HostBinding('class.fd-nested-list__icon')
     fdNestedListIconClass = true;
```
N1XUS, line 56: private readonly

N1XUS:
> > @mikerodonnell89 question: since icon component doesn't have any template, should it be still done as a component? I was thinking that we can move basic icon applying stuff into a standalone directive and simply put it to the hostDirectives. This can actually extend current syntax for the cases when it's better to write `<i glyph="decline" />` rather than full `fd-icon glyph="decline" />` what do you think?
> 
> Yes actually I had a few ideas for improvements - one would be that we'd actually just use `fd-icon` for these icons and apply the side nav class from the parent using ContentChildren, so we wouldn't need new components. But I went this route so I wouldn't introduce breaking changes. Also, I opted to change this from directive to component because directives can't extend components (fd-icon). So this solution isn't perfect but seemed like the best option without introducing breaking changes

I was meaning, maybe we can change fd-icon to directive itself?

mikerodonnell89:
> > > @mikerodonnell89 question: since icon component doesn't have any template, should it be still done as a component? I was thinking that we can move basic icon applying stuff into a standalone directive and simply put it to the hostDirectives. This can actually extend current syntax for the cases when it's better to write `<i glyph="decline" />` rather than full `fd-icon glyph="decline" />` what do you think?
> > 
> > 
> > Yes actually I had a few ideas for improvements - one would be that we'd actually just use `fd-icon` for these icons and apply the side nav class from the parent using ContentChildren, so we wouldn't need new components. But I went this route so I wouldn't introduce breaking changes. Also, I opted to change this from directive to component because directives can't extend components (fd-icon). So this solution isn't perfect but seemed like the best option without introducing breaking changes
> 
> I was meaning, maybe we can change fd-icon to directive itself?

Correct me if I'm wrong but I think the icon module needs a component to import styles, since a directive can't have styleUrls.

N1XUS:
> > > > @mikerodonnell89 question: since icon component doesn't have any template, should it be still done as a component? I was thinking that we can move basic icon applying stuff into a standalone directive and simply put it to the hostDirectives. This can actually extend current syntax for the cases when it's better to write `<i glyph="decline" />` rather than full `fd-icon glyph="decline" />` what do you think?
> > > 
> > > 
> > > Yes actually I had a few ideas for improvements - one would be that we'd actually just use `fd-icon` for these icons and apply the side nav class from the parent using ContentChildren, so we wouldn't need new components. But I went this route so I wouldn't introduce breaking changes. Also, I opted to change this from directive to component because directives can't extend components (fd-icon). So this solution isn't perfect but seemed like the best option without introducing breaking changes
> > 
> > 
> > I was meaning, maybe we can change fd-icon to directive itself?
> 
> Correct me if I'm wrong but I think the icon module needs a component to import styles, since a directive can't have styleUrls.

With pure angular approach, yes. Only component can load the styles. But constructed stylesheets can be loaded by anything ;)

libs/core/src/lib/nested-list/nested-list-directives.ts
platon-rov: > Decorated component class members must be public. You cannot make an @[Input](https://angular.io/api/core/Input)() property private or protected.

https://angular.io/guide/aot-compiler#phase-2-code-generation

mikerodonnell89:
As discussed with @N1XUS the CSS/component issue will be addressed in a later PR

Issue #9199, opened 2023-01-10
Search Field suggestion items overlay search field (if many suggestions are found)


MichaelBaySAP:
![image](https://user-images.githubusercontent.com/40032066/211578013-574b15dd-24e8-4a34-847a-5c2a3972d30a.png)
Here is a Stackblitz that demonstrates the issue: https://stackblitz.com/edit/angular-9qybvz?file=src%2Fapp%2Fplatform-search-field-basic-example.component.ts

To reproduce:

- open the stackblitz  https://stackblitz.com/edit/angular-9qybvz?file=src%2Fapp%2Fplatform-search-field-basic-example.component.ts
- type 't' into the first search field

I have tried with CSS to manually not overlay the search field.

InnaAtanasova:
Hi @MichaelBaySAP, a PR with the fix is now open and once it's merged you can test it on your side. Please let us know if you still face issue after the update. Thanks

littleamigo:
Hi @InnaAtanasova, thanks for the fix. I've tested it and it works basically okay.

However, the suggestions do not disappear when I scroll the whole page content up and down, neither it scrolls along with the rest of the page. It actually keeps its exact location while the other page content is moving as I scroll. I've observed the same weird visual appearance also with the opened value list of a select control. Is this something you already know and more like a general problem with these open value lists that you already tackle in another issue? Or would you like me to open another issue covering that?

PR #9237, opened 2023-01-16
fix(platform): add fix for Search Field suggestions menu overflow

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/9199

## Description
The Search field suggestions menu was not handling properly the case where we have many items. 
Note: the css code is done here as this component is specific for Platform

### Before:
Screenshot 2023-01-16 at 6 09 38 PM

### After:

Screenshot 2023-01-16 at 6 08 57 PM
Screenshot 2023-01-16 at 6 09 08 PM

### DIFF
--- a/libs/docs/platform/search-field/examples/platform-search-field-basic-example.component.ts
+++ b/libs/docs/platform/search-field/examples/platform-search-field-basic-example.component.ts
@@ -9,45 +9,9 @@ import { SuggestionItem, SearchInput } from '@fundamental-ngx/platform/search-fi
     encapsulation: ViewEncapsulation.None
 })
 export class PlatformSearchFieldBasicExampleComponent {
-    suggestions: SuggestionItem[] = [
-        {
-            value: 'Apple'
-        },
-        {
-            value: 'Banana'
-        },
-        {
-            value: 'Blueberry'
-        },
-        {
-            value: 'Cherry'
-        },
-        {
-            value: 'Grape'
-        },
-        {
-            value: 'Lemon'
-        },
-        {
-            value: 'Lime'
-        },
-        {
-            value: 'Orange'
-        },
-        {
-            value: 'Peach'
-        },
-        {
-            value: 'Pineapple'
-        },
-        {
-            value: 'Plum'
-        },
-        {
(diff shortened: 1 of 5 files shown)


libs/docs/platform/search-field/examples/platform-search-field-basic-example.component.ts
```diff
@@ -45,6 +45,69 @@ export class PlatformSearchFieldBasicExampleComponent {
         },
         {
             value: 'Raspberry'
+        },
```
N1XUS, line 48: Maybe instead of hardcoded variants use generated with the help of new Array(n)?

N1XUS (changes requested):
image
Stumbled across such behaviour

mikerodonnell89:
> image Stumbled across such behaviour

To clarify - seeing this happen when you scroll up while the popover is open

libs/docs/platform/search-field/examples/platform-search-field-basic-example.component.ts
InnaAtanasova: done

mikerodonnell89 (changes requested):
While this is a great improvement (so I'd be happy to merge it as-is), I'd like to see the popover fit into the visible area of the window as soon as it's opened, to avoid this situation where there are options below the bottom of the screen

Screenshot 2023-01-17 at 11 38 57 AM

mikerodonnell89 (approved):

platon-rov (approved):

N1XUS (approved):

g-cheishvili:
@InnaAtanasova this fails https://cloud.nx.app/runs/iv6BgvzKEE/task/docs-platform-search-field:e2e

InnaAtanasova:
> @InnaAtanasova this fails https://cloud.nx.app/runs/iv6BgvzKEE/task/docs-platform-search-field:e2e

@g-cheishvili for me it doesn't load the actual problem. All I get is:
Screenshot 2023-01-23 at 4 49 00 PM

Issue #9202, opened 2023-01-10
enhancement: CDK focusable list


N1XUS:
#### Is this a bug, enhancement, or feature request?
enhancement
#### Briefly describe your proposal.
With focusable list we have vertical or horizontal navigation available.
In some components (grid, table, etc) navigation can be bi-directional (up-down and left-right at the same time).
I would suggest improving focusable list component and add focusable grid functionality which can be used in table and grid

g-cheishvili:
And calendar, that one is big too

PR #9242, opened 2023-01-18
feat(cdk): focusable list & item upgrade, focusable grid feature and docs

platon-rov:
## Related Issue(s)

Closes https://github.com/SAP/fundamental-ngx/issues/9202

## Description

* Focusable Grid feature for CDK and docs for it.
* Focusable Item enhancements, refactoring
  * Functionality to focus inner tabbable elements.
  * Announcing info when in grid.
* Focusable List enhancements, refactoring
  * Keeping only one tab stop per list.
  * Input to manipulate wrapping mode.

### After:

image


libs/docs/cdk/focusable-grid/focusable-grid-header/focusable-grid-header.component.html
```diff
@@ -0,0 +1,11 @@
+<header>Focusable Grid</header>
+<description>
+    You can quickly build out focusable grid collection for your components. For that you will need to use a combination
+    of <code>[fdkFocusableGrid]</code>, <code>[fdkFocusableList]</code> and <code>[fdkFocusableItem]</code> directives.
+    It also respects <code>[fdkDisabled]</code> and <code>[fdkReadonly]</code> directives, so that those elements will
```
mikerodonnell89, line 5: Typical behavior for readonly elements (such as inputs) still allows them to be focused and reached via tab keys - so maybe they should not be excluded from the arrow keys as well?

libs/docs/cdk/focusable-grid/examples/short-rows-example/short-rows-example.component.html
```diff
@@ -0,0 +1,13 @@
+<div fdkFocusableGrid shortRowFocus="first">
+    <div *ngFor="let list of [].constructor(5); let rowIndex = index" fdkFocusableList navigationDirection="horizontal">
```
mikerodonnell89, line 2: Confused at the purpose of `navigationDirection="horizontal"` here, isn't the direction both horizontal and vertical?

libs/cdk/src/lib/utils/directives/focusable-list/focusable-list.directive.ts
```diff
@@ -26,22 +60,54 @@ export class DeprecatedFocusableListDirective extends DeprecatedSelector {}
...
+        {
+            provide: FDK_FOCUSABLE_LIST_DIRECTIVE,
+            useExisting: FocusableListDirective
+        },
+        DestroyedService
+    ]
 })
 export class FocusableListDirective implements AfterViewInit {
-    /** Direction of navigation. */
+    /** Direction of navigation. Always horizontal when in grid. */
```
mikerodonnell89, line 72: In my opinion a third option here like `'grid'` would make more sense

libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
```diff
@@ -38,64 +43,152 @@ export class DeprecatedFocusableItemDirective extends DeprecatedSelector {}
...
+            this._disableTabbableElements();
+        }
     }
 
     /** @hidden */
-    ngAfterViewInit(): void {
-        this._viewInit$.next(true);
+    @HostListener('focusin')
+    async _onFocusin(): Promise<void> {
+        if (this._timerId != null) {
```
mikerodonnell89, line 138: What is the purpose of `_timerId`? I'm sure there's a reason I just don't know. Change detection related?

mikerodonnell89 (commented):
This is great! Couple comments/questions

libs/cdk/src/lib/utils/directives/focusable-list/focusable-list.directive.ts
platon-rov: Good idea!

libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
platon-rov: That's for supporting multiple focusable elements inside a cell, when we come to the cell we should focus first focusable child and read a11y text but then when tabbing between focusable elements inside a single cell we should not manipulate focus neither read a11y text again.

libs/docs/cdk/focusable-grid/focusable-grid-header/focusable-grid-header.component.html
platon-rov: Yea, I agree with you, let me change it here and in focusable list.

mikerodonnell89 (changes requested):
build is failing: https://app.netlify.com/sites/fundamental-ngx/deploys/63ca8cd1b7845b00091e8970#L1063

libs/cdk/src/lib/utils/directives/focusable-list/focusable-list.directive.ts
```diff
@@ -117,12 +117,12 @@ export class FocusableListDirective implements OnChanges, AfterViewInit {
             return;
         }
 
-        if (changes.wrap) {
+        if (changes['wrap']) {
             this._keyManager = this._keyManager.withWrap(changes.wrap.currentValue);
```
g-cheishvili, line 121:
```ts
this._keyManager = this._keyManager.withWrap(changes['wrap'].currentValue);
```

g-cheishvili (approved):
Looks good to me

mikerodonnell89 (changes requested):
This looks and works great but just one last thing, after mouse clicking a focusable item, the item is focused but the focus outline doesn't initially appear, I can only see it once I start using the keyboard

platon-rov:
> This looks and works great but just one last thing, after mouse clicking a focusable item, the item is focused but the focus outline doesn't initially appear, I can only see it once I start using the keyboard

Nice catch, so I've updated docs and now it looks better, although this feature is not about styling so when using it - styles for focus state should be implemented on consumer side.

PR #9253, opened 2023-01-20
fix(platform): add property expandOnInit to platform tree table

### LINKED ISSUE #9241: [fdp-table] [Sourcing][Dev blocker]: Expand all / collapse all on page load
Enhancement

1) There should be a way using which programatically one should be able to decide - whether the table should be expanded or collapsed by default.

Angular 13 
core / platform - 0.35.x

This affects guided sourcing customers, please reach out in case of any further queries.

mikerodonnell89:
fixes #9241 

Adds a boolean input allowing the tree table rows to all be expanded when the table is loaded


libs/platform/src/lib/table/table.component.ts
```diff
@@ -849,6 +853,10 @@ export class TableComponent<T = any> extends Table<T> implements AfterViewInit,
         this._listenToTableContainerMouseLeave();
 
         this._cdr.detectChanges();
+
+        if (this.expandOnInit === true) {
```
g-cheishvili, line 857:
```suggestion
        if (this.expandOnInit) {
```
Why bother with strict `true` comparison though

g-cheishvili (approved):

InnaAtanasova (approved):

platon-rov (approved):

PR #9274, opened 2023-01-24
fix(core,cx): error when no linkItem provided

### LINKED ISSUE #9272: fd-nested-list error: Cannot set properties of undefined (setting 'ariaDescribedby')
Bug

We are using the fd-nested-list directive and after migrating from 0.33.4 to 0.36.2 we receive this error: `TypeError: Cannot set properties of undefined (setting 'ariaDescribedby')`. Our error handling system picks up this new error.

0.36.2 and newer

Reproduced in StackBlitz: https://stackblitz.com/edit/angular-x6bcfd?file

Stack trace from version 0.36.2
```
TypeError: Cannot set properties of undefined (setting 'ariaDescribedby')
    at fundamental-ngx-core-nested-list.mjs:901:13
    at Array.forEach (<anonymous>)
    at QueryList.forEach (core.mjs:22927:23)
    at NestedListDirective._setAriaAttributes (fundamental-ngx-core-nested-list.mjs:899:26)
    at Object.next (fundamental-ngx-core-nested-list.mjs:867:18)
    at ConsumerObserver.next (Subscriber.js:91:33)
    at SafeSubscriber._next (Subscriber.js:60:26)
    at SafeSubscriber.next (Subscriber.js:31:18)
    at combineLatest.js:42:40
    at OperatorSubscriber._next (OperatorSubscriber.js:13:21)
```

mikerodonnell89:
fixes #9272 

Fixes a console error when a nested list item does not have a link


InnaAtanasova (approved):

libs/core/src/lib/nested-list/nested-list/nested-list.directive.ts
```diff
@@ -172,16 +172,18 @@ export class NestedListDirective implements AfterContentInit, NestedListInterfac
...
-            item.linkItem._ariaLabel = this._translationResolver.resolve(lang, 'coreNestedList.linkItemAriaLabel', {
-                itemDetails: item.linkItem.getTitle(),
-                index: i + 1,
-                total: this.nestedItems.length,
-                selectedDescription:
-                    !this._nestedListStateService.selectable && item.linkItem.selected
-                        ? ', ' + this.ariaLabelSelected
-                        : ''
-            });
+            if (item.linkItem) {
```
N1XUS, line 175: you can invert the condition so that if there's no link item, just return. It will be more readable I think.

libs/cx/src/lib/nested-list/nested-list/nested-list.component.ts
```diff
@@ -183,16 +183,18 @@ export class NestedListComponent implements AfterContentInit, NestedListInterfac
...
-            item.linkItem._ariaLabel = this._translationResolver.resolve(lang, 'coreNestedList.linkItemAriaLabel', {
-                itemDetails: item.linkItem.getTitle(),
-                index: i + 1,
-                total: this.nestedItems.length,
-                selectedDescription:
-                    !this._nestedListStateService.selectable && item.linkItem.selected
-                        ? ', ' + this.ariaLabelSelected
-                        : ''
-            });
+            if (item.linkItem) {
```
N1XUS, line 186: Same here

libs/core/src/lib/nested-list/nested-list/nested-list.directive.ts
```diff
@@ -172,16 +172,20 @@ export class NestedListDirective implements AfterContentInit, NestedListInterfac
...
-                index: i + 1,
-                total: this.nestedItems.length,
-                selectedDescription:
-                    !this._nestedListStateService.selectable && item.linkItem.selected
-                        ? ', ' + this.ariaLabelSelected
-                        : ''
-            });
+            if (!item.linkItem) {
+                return;
+            } else {
```
N1XUS, line 177: else is not needed.

mikerodonnell89: True, but I think that is more of a style preference

PR #9289, opened 2023-01-27
fix(platform): table column ngIf bug

### LINKED ISSUE #9212: [fdp-table][Sourcing][Dev blocker]: ngIf does not work in fundamental platform table
Bug
*ngIf for fdp table column fails to evaluate when 

- Initial value of variable is false and
- Variable value changes after table is rendered.

Fundamental ngx - 0.38.2

- Create a platform table with <fdp-column> containing an *ngIf checking for some variable value being true or not. 
- Initialise a variable with value as false
- create a method that causes the change in variable's value to true

Expected: The column that wasn't visible previously should be visible now because the variable's value is true.

Observed: The column is still not visible.

(refer to code and video)

```
<!-- component.html file -->
<fdp-table
  [dataSource]="source"
  [trackBy]="trackBy"
  emptyTableMessage="No data found"
>
  <fdp-table-toolbar title="Order Line Items" [hideItemCount]="false">
    <fdp-table-toolbar-actions>
      <fdp-button label="Hide column" (click)="hideCol()"></fdp-button>
    </fdp-table-toolbar-actions>
  </fdp-table-toolbar>

  <fdp-column name="name" key="name" label="Name" align="start"> </fdp-column>

  <fdp-column
    name="description"
    key="description"
    *ngIf="testBool"
    label="Description"
  >
  </fdp-column>

  <fdp-column name="price" key="price.value" label="Price" align="end">
  </fdp-column>

  <fdp-column name="status" key="status" label="Status" align="center">
  </fdp-column>
</fdp-table>

```

```
// component.ts file
import { Observable, of } from 'rxjs';

import { FdDate } from '@fundamental-ngx/core/datetime';
import { TableDataSource, TableDataProvider, TableState } from '@fundamental-ngx/platform/table';

@Component({
    selector: 'fdp-platform-table-default-example',
    templateUrl: './platform-table-default-example.component.html'
})
export class PlatformTableDefaultExampleComponent {
    source: TableDataSource<ExampleItem>;
 // test boolean value that is assigned false 
    testBool: boolean = false;

    constructor() {
        this.source = new TableDataSource(new TableDataProviderExample());
    }

    trackBy(index: number, item: ExampleItem): number {
        return item.id;
    }

    alert(message: string): void {
        alert(message);
    }

    hideCol () {
// method that toggles the value of the bool
        this.testBool = !this.testBool;
        console.log(this.testBool);
    }
}

// data provider and other boiler plate code

https://user-images.githubusercontent.com/54993789/212319429-c27b5544-167f-4f8b-a532-a6b750409ba0.mov


```

mikerodonnell89:
fixes #9212

### DIFF
--- a/libs/platform/src/lib/table/table.component.ts
+++ b/libs/platform/src/lib/table/table.component.ts
@@ -1810,6 +1810,7 @@ export class TableComponent<T = any> extends Table<T> implements AfterViewInit, 
     /** @hidden */
     private _listenToColumns(): void {
         this.columns.changes.pipe(startWith(null)).subscribe(() => {
+            this.setTableState({ ...this.getTableState(), columns: this.columns.map((column) => column.name) });
             const columns = this.getTableColumns();
             const prevColumns = this._tableColumnsSubject.getValue().map((column) => column.name);
             const currentColumns = columns.map((column) => column.name);
--- a/libs/platform/src/lib/table/table.component.spec.ts
+++ b/libs/platform/src/lib/table/table.component.spec.ts
@@ -210,6 +210,13 @@ describe('TableComponent internal', () => {
                 <fdp-column name="status" key="status" label="Status"></fdp-column>
 
                 <fdp-column name="verified" key="isVerified" label="Client Verified"></fdp-column>
+
+                <fdp-column
+                    *ngIf="showInvisibleColumn"
+                    name="invisible"
+                    key="isInvisible"
+                    label="Invisible Column"
+                ></fdp-column>
             </fdp-table>
         `
     })
@@ -221,6 +228,7 @@ describe('TableComponent internal', () => {
         /** So big so table column won't grow on any device */
         customColumnWidth = 10000;
         rowsClass: TableRowClass<any> = 'class';
+        showInvisibleColumn = false;
     }
 
     describe('TableComponent Host', async () => {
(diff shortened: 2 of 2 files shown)


libs/platform/src/lib/table/table.component.ts
```diff
@@ -1810,6 +1810,7 @@ export class TableComponent<T = any> extends Table<T> implements AfterViewInit,
     /** @hidden */
     private _listenToColumns(): void {
         this.columns.changes.pipe(startWith(null)).subscribe(() => {
+            this.setTableState({ ...this.getTableState(), columns: this.columns.map((column) => column.name) });
```
N1XUS, line 1813: Isn't this mean that all other columns (that were previously hidden with table state) will also appear?

g-cheishvili: Where should we set this state update function in your opinion ? @N1XUS

N1XUS: I think we need to compare previous array of columns with new one. If new array one has added columns, push them to the state and ignore updating previous one.
This would be better due to the fact that users can manually hide some columns via p13 dialog, and when this check triggers, that config will be useless.

mikerodonnell89:
Closing in favor of https://github.com/SAP/fundamental-ngx/pull/9329

PR #9290, opened 2023-01-27
fix(platform): add public function for checking of rows

### LINKED ISSUE #9277: [Sourcing][Dev-blocker][fdp-table] No way to set some rows as selected programmatically by default
bug

1) Add a fdp table , selection mode = multiple
2) programatically set some rows as checked 
3) So that user can see those rows as default selected

Not sure , if we can do it as of today. Please update.

Angular 13
core / platform : 0.35.x

1) Add a fdp table , selection mode = multiple
2) try to set programatically some rows as checked 
3) not able to achieve this

please update if this can be achieved somehow?

mikerodonnell89:
fixes none

Adding a public function to the platform table component called `setRowCheckedState(rowIndex: number, value: boolean)` which allows developers to check or uncheck a row by index. It can be used like so:

```
    @ViewChild(TableComponent)
    table: TableComponent;

    constructor() {
        this.source = new TableDataSource(new TableDataProviderExample());
    }

    ngAfterViewInit() {
        this.table.tableRows.forEach((row, index) => {
            if (this.someCondition) {
                this.table.setRowCheckedState(index, true);
            }
        });
    }
```


libs/platform/src/lib/table/table.component.ts
```diff
@@ -1075,6 +1075,23 @@ export class TableComponent<T = any> extends Table<T> implements AfterViewInit,
         }
     }
 
+    /**
+     * Sets the checked state of the row.
+     * @param rowIndex Index of the row.
+     * @param value Value of the checked state.
+     */
+    setRowCheckedState(rowIndex: number, value: boolean): void {
```
N1XUS, line 1083: I think we can also expend this functionality just like we did with `rowNavigatable`, where the data of the row itself has `checked` flag, property name of which is set via @Input property of the table. This way developers could use just a datasource to mark the checked/unchecked row instead of iterating over the rows

mikerodonnell89: @platon-rov is addressing that with #9304

mikerodonnell89:
Reopening as this public function may still be useful. But rolling back the commit with the rowChecked input property

platon-rov (approved):

Issue #9239, opened 2023-01-17
[fdp-table][Sourcing][Dev blocker] query regarding fdp-table selective enablement of checkbox and selected All checkbox in tree table form.


rayansailani:
#### Is this a bug, enhancement, or feature request?
Query/ Feature request

#### Briefly describe your proposal.

When using fdp-table as a treeTable there arises a need to selectively enable the checkboxes based on the type of row. For example each row could contain a param called `isRowSelectable`. Currently the way we have implement is by using the rowsClass property in which we check for the row property and add a class that disables the checkbox. Is there no way to do this by default currently in the component. 
![image](https://user-images.githubusercontent.com/54993789/212943917-0f71733f-e673-4bcf-a0a1-d602227947c4.png)
picture for reference. 

In the same use case when we have selective enablement of checkbox, the select all checkbox is of no use as it will select the elements that are semantically not to be selected. If there would be a param `hideSelectAllRowsCheckbox` that could hide the previously mentioned cb, it would be better as one has to hide the checkbox in styles as follows 
```
fdp-table tr:first-child fd-checkbox {
        display: none !important;
}
```
![image](https://user-images.githubusercontent.com/54993789/212952660-4148d9e3-ad1c-4a93-98f6-2d6325823177.png)
picture for reference

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
0.38.2
#### Please provide relevant source code if applicable.
```
// component.html file
<fdp-table
  [dataSource]="source"
  emptyTableMessage="No data found"
  [rowsClass]="rowsClass"
  selectionMode="multiple"
>
  <fdp-table-toolbar title="Order Line Items" [hideItemCount]="false">
    <fdp-table-toolbar-actions>
      <fdp-button
        label="Show All checkboxes"
        (click)="setRowClassString()"
      ></fdp-button>
      <fdp-button
        label="Hide checkboxes out of stock"
        (click)="setRowClassFunction()"
      ></fdp-button>
    </fdp-table-toolbar-actions>
  </fdp-table-toolbar>

  <fdp-column name="name" key="name" label="Name" align="start"> </fdp-column>

  <fdp-column name="description" key="description" label="Description">
  </fdp-column>

  <fdp-column name="price" key="price.value" label="Price" align="end">
  </fdp-column>

  <fdp-column name="status" key="status" label="Status" align="center">
  </fdp-column>
</fdp-table>

```

```
// component.ts
[...]

PR #9304, opened 2023-01-30
feat(platform): table not selectable and preselected rows

platon-rov:
## Related Issue(s)

Closes https://github.com/SAP/fundamental-ngx/issues/9239

## Description

* feat: table `selectedKey` input to preselect rows
* feat: table `selectableKey` input to disable selections for rows

### After:

image


mikerodonnell89 (changes requested):
This is great. One thing I noticed is that if I take the editable table example and make it multi-select, when I create a new row, and mark it checked, and press save, it is not checked after the save event

platon-rov:
> This is great. One thing I noticed is that if I take the editable table example and make it multi-select, when I create a new row, and mark it checked, and press save, it is not checked after the save event

I reproduced this on main so I think it's out of scope of the current pr, created ticket https://github.com/SAP/fundamental-ngx/issues/9323

mikerodonnell89 (approved):

Issue #9248, opened 2023-01-20
[fdp-table][Sourcing][Accessibility]: input box inside fdp-table-cell issues


I543348:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
1) Not able to add spaces inside a input box in a fdp-table-cell : <tr> has keydown event listener, when removed works fine
2)  Not able to traverse through characters using left and right arrow keys inside a input box inside a table cell
3) Accessibility: Not able to move between the input boxes in different rows using up and down arrow keys

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)

angular 13
core / platform : 0.35.x

#### If this is a bug, please provide steps for reproducing it.
Try to type something inside a input box in any table cell and add space - https://sap.github.io/fundamental-ngx/#/platform/table#custom-columns example
Notice that there is "." dot added but not space.

1) Not able to add spaces inside a input box in a fdp-table-cell : <tr> has keydown event listener, when removed works fine
2)  Not able to traverse through characters using left and right arrow keys inside a input box inside a table cell
3) Not able to move between the input boxes in different rows using up and down arrow keys

#### Please provide relevant source code if applicable.
pls refer to https://sap.github.io/fundamental-ngx/#/platform/table#custom-columns 

#### Is there anything else we should know?
works fine when keydown event listener needs to be removed

mikerodonnell89:
I will speak with the accessibility designers about how point #2 should work, since we will still want the left/right arrow keys to move cell focus.

mikerodonnell89:
Interaction guidelines state that table cells with inputs in them should default to focus navigation. When a user moves to a cell with an editable input, they can press f2 (windows) or return (mac) to move into edit mode, where the text cursor will move to the input and the user can then use the arrow keys to move the cursor around text in the input. The user can also start typing immediately after focusing the cell, which would clear all text in the cell and enter edit mode automatically.

mikerodonnell89:
https://github.com/SAP/fundamental-ngx/pull/9368

mikerodonnell89:
Fixed as part of https://github.com/SAP/fundamental-ngx/pull/9324

I543348:
This is fixed. Checked and tested on NGX website https://sap.github.io/fundamental-ngx/#/platform/table#custom-columns

PR #9306, opened 2023-01-30
fix(platform): can not enter space characters in inputs inside table cells

mikerodonnell89:
part of #9248

On our current build, go to the platform table custom columns example, which has a column where the cells have input components. Note that you can not press space to enter a space character in the input

### DIFF
--- a/libs/platform/src/lib/table/table.component.ts
+++ b/libs/platform/src/lib/table/table.component.ts
@@ -1404,7 +1404,12 @@ export class TableComponent<T = any> extends Table<T> implements AfterViewInit, 
             return;
         }
 
-        event?.preventDefault();
+        if (event) {
+            const eventTarget = event.target as HTMLElement;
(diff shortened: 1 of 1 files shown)


libs/platform/src/lib/table/table.component.ts
```diff
@@ -1404,7 +1404,15 @@ export class TableComponent<T = any> extends Table<T> implements AfterViewInit,
             return;
         }
 
-        event?.preventDefault();
+        if (event) {
```
N1XUS, line 1407: @mikerodonnell89 what if we check whether the row itself is currently focused, instead of checking the element type?
I mean, when user presses enter on focused row, that's when the navigation, or activation should happen. If user focuses inside the row, then he can press any button he likes. What do you think?

mikerodonnell89: The only time this function is ever called and an event is actually passed is when a space press is made. But I can be more explicit in this function here and check specifically for the space press

PR #9314, opened 2023-02-01
fix(platform): screenreader reading fdp-select selected value twice

### LINKED ISSUE #8710: Screenreader reading 'fdp-select' selected value twice when placeholder is not provided
bug

#### Briefly describe your proposal
When we select any value in fdp-select and use down/up arrows to change the selected value, without opening the popover, the screenreader reads the selected value twice if placeholder is not provided. Since, placeholder might not be needed in all implementations especially if there must be a pre-selected value for the fdp-select component, the values should not be read twice when navigating using arrows. This issue is reproduced for both JAWS and macOS voiceover.

angular version: 13.3.4
ngx-platform: 0.35.4

1. Remove placeholder or keep it as an empty string for the fdp-select component, if provided
2. Open any screenreader, Like JAWS
3. Set focus on fdp-select component, close the popover if opened
4. Change the selected value using up/down arrows, while the popover is closed, and observe the voiceover speech

skaquib123:
## Related Issue(s)

closes #8710 

## Description

Screenreader reading 'fdp-select' selected value twice when placeholder is not provided. 

### Before:
image

### After:

I have tested in local by removing placeholder that it is working as expected.

Screenshot 2023-02-01 at 4 16 01 PM

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [x] poor examples

4. Accessibility testing

##### PR Quality

-   [x] update `README.md`


libs/core/src/lib/select/select.component.ts
```diff
@@ -409,7 +409,11 @@ export class SelectComponent
 
     /** @hidden */
     ngOnInit(): void {
-        this.ariaLabel = this.ariaLabel || this.placeholder;
+        if (this.placeholder) {
+            this.ariaLabel = this.ariaLabel || this.placeholder;
+        } else {
+            this.ariaLabel = 'Select an option';
```
mikerodonnell89, line 415: Needs to utilize the i18n module

mikerodonnell89 (changes requested):

platon-rov (changes requested):
please discard changes to package-lock.json & yarn.lock.

libs/core/src/lib/select/select.component.html
```diff
@@ -47,7 +47,7 @@
         [attr.aria-haspopup]="!(this.readonly || this.disabled)"
         [attr.aria-required]="required"
         [attr.title]="triggerValue"
-        [attr.aria-label]="ariaLabel"
+        [attr.aria-label]="ariaLabel || ('platformSelect.selectOptionLabel' | fdTranslate)"
```
mikerodonnell89, line 50: Sorry for the confusion, I was referring to the operator here. Because you are setting ariaLabel in the component to the hardcoded english string, the translation will never be used here

mikerodonnell89 (changes requested):

libs/core/src/lib/select/select.component.ts
```diff
@@ -424,8 +424,7 @@ export class SelectComponent<T = any>
 
     /** @hidden */
     ngOnInit(): void {
-        this.ariaLabel = this.ariaLabel || this.placeholder;
-
+        this.ariaLabel = this.placeholder;
```
platon-rov, line 427: If we have aria-label provided it will be overwritten with placeholder value, is it expected?

platon-rov (changes requested):

libs/core/src/lib/select/select.component.ts
mikerodonnell89: I suppose this was sort of the previous behavior, when no ariaLabel was provided, so maybe technically a separate issue. Won't be overwritten by default given there is no default value

mikerodonnell89: I'll approve but also ask to remove this

mikerodonnell89 (changes requested):
Please remove this line entirely: https://github.com/SAP/fundamental-ngx/pull/9314/files#r1108378174

Then we can get this merged

libs/core/src/lib/select/select.component.ts
mikerodonnell89: I am actually thinking it would make sense if the developer provides both an ariaLabel and a placeholder, that they'd want both read by the screenreader. So we won't want this line

mikerodonnell89 (approved):

PR #9317, opened 2023-02-01
fix(core): dynamic page should not add scrollbar directive if there is already a tab list

mikerodonnell89:
fixes none


libs/core/src/lib/dynamic-page/dynamic-page.component.html
```diff
@@ -19,7 +19,12 @@
     </header>
 
     <ng-content select="fd-tab-list"></ng-content>
-    <div fd-scrollbar>
+
+    <div fd-scrollbar *ngIf="!_tabComponent">
+        <ng-content select="fd-dynamic-page-content"></ng-content>
+    </div>
+
+    <div *ngIf="_tabComponent">
```
platon-rov, line 27: do we need it? 

I see only two cases

1. TabList present, ng-content on line 21 catches it
2. TabList *not* present, we wrap ng-content on line 24 with container and show it, line 23

mikerodonnell89: Two ways of doing the same thing I think

InnaAtanasova (approved):

PR #9318, opened 2023-02-01
fix(core): give action sheet wrapper same z index as dialog

mikerodonnell89:
part of defect hunting

before:
Screenshot 2023-02-01 at 1 47 01 PM

after:
Screenshot 2023-02-01 at 1 46 41 PM


platon-rov (approved):
Approving but probably we have to get back to global z-index managing as we had in styles.

g-cheishvili (approved):

PR #9319, opened 2023-02-01
fix(ci): lint-staged

platon-rov:
fixes

```
Nx no longer supports using positional arguments for base and head
```


platon-rov:
> What's up with lockfile updates? Is that intentional?

reverted, thanks

g-cheishvili (approved):

Issue #9135, opened 2022-12-26
[Accessibility][Sourcing]: fdp-table accessibility issues


I543348:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
1) While announcing table cell value - Its corresponding column header is not output by screenreader
2) All tables have role=table, this should be changed as grid and treegrid accordingly: https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/
3) Enter and exit edit mode not available when there are more than one interactive elements inside a table cell

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 13.x.x
   "@fundamental-ngx/core": "0.35.4",
    "@fundamental-ngx/platform": "0.35.4",

#### If this is a bug, please provide steps for reproducing it.
Can be reproducible on https://sap.github.io/fundamental-ngx/#/platform/table 

#### Is there anything else we should know?
This is a blocker for guided sourcing customers

I543348:
This works fine , when there are more than 1 interactive elements in a table call , ESC key is used to come out of interactive elements.

mikerodonnell89:
Reopening for "description" being read twice

fkolar:
I think this is more a duplicate of #9527. I think this is the same problem. we should keep only one issue. 
@I543348

mikerodonnell89:
To elaborate - column header is read twice for jaws only when first accessing a cell in the grid for the first time ("Name" seen here twice after clicking "Astro Phone 6". The column heading is only read once after clicking "Beam Breaker B-1"
![Screenshot 2023-05-10 at 4 30 57 PM](https://github.com/SAP/fundamental-ngx/assets/2471874/79b6c1c3-6981-424e-87c0-53d545f5c115)

I543348:
@mikerodonnell89 , We checked with Accessibility Test Lab (ATL) team as eve for first time , the column header should not be read twice.
This has to be reopened

I543348:
Hi Team, i just checked with Central accessibility team and they are gave below inputs.
Please check and help.

<img width="846" alt="Screenshot 2023-06-01 at 10 45 51 PM" src="https://github.com/SAP/fundamental-ngx/assets/88182358/bd1fe4c4-8262-432e-92b2-e1eae6303e5e">

cc: @mikerodonnell89 , @fkolar

I543348:
This is working fine.
Thanks.

Issue #9134, opened 2022-12-26
[Accessibility][Sourcing]: arrow keys not functioning properly on table with frozen columns


I543348:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
Wrong behaviour of left or right arrow keys when on a table with frozen columns

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 13.x.x
   "@fundamental-ngx/core": "0.35.4",
    "@fundamental-ngx/platform": "0.35.4",

#### If this is a bug, please provide steps for reproducing it.
1) Reach a fdp-table with one or two frozen columns
2) Try using left or right arrow on any of the rows

Actual: The frozen columns get scrolled left or right
Expected: The user should be able to traverse through different cells of that row

#### Please provide relevant source code if applicable.
Replicable on https://sap.github.io/fundamental-ngx/#/platform/table :Column Freezing section

#### Is there anything else we should know?
This is a blocker for guided sourcing customers

I543348:
The visible focus is splitting across 2 columns , as per accessibility guidelines, the visible focus should be clearly visible .
This needs to be reopened
please see attached screenshot, 
<img width="1792" alt="Screenshot 2023-03-06 at 9 57 23 PM" src="https://user-images.githubusercontent.com/88182358/223171162-f8ac7b9b-6bbd-4f0f-9df7-827fbe9260cb.png">

Issue #9132, opened 2022-12-26
[Accessibility][Sourcing]: fdp tree table accessibility issues


I543348:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
1) Not able to navigate into child row(s) after expanding parent row using down arrow key in tree table
2) Tree level is not output by screenreader

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 13.x.x
   "@fundamental-ngx/core": "0.35.4",
    "@fundamental-ngx/platform": "0.35.4",

#### If this is a bug, please provide steps for reproducing it.
1) Not able to navigate into child row(s) after expanding parent row using down arrow key in tree table
2) Tree level is not output by screenreader
3) Position of the cell should be read out by screenreader eg : column 3 row 4 both in voiceover and jaws (forms mode and linear reading mode)

#### Please provide relevant source code if applicable.
Reproducible on fdp tree table on https://sap.github.io/fundamental-ngx/#/platform/table 

#### Is there anything else we should know?
This is a blocker for Deutsche Telecom customer for guided sourcing

I543348:
This is fixed

Issue #9131, opened 2022-12-26
[Acccessibility][Sourcing]: fdp-table navigation issues using keyboard


I543348:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
There are some navigation issues with regards to <fdp-table>

1. Reach using keyboard on a blank or empty table cell - Notice that up/ down arrow doesn’t work and only left / right arrow key works on this type of cell
2. Reach the table first time using keyboard (tab key press)- Now once the focus is on the first column of table - Notice that left / right arrow keys does not work to traverse to other column headers
3. Non interactive elements (text) do no show visible focus - so user is not aware which cell he / she is currently on but screen reader reads the cell values 
4.  Not able to easily skip the table cells : Jump between first and last cell in table not working: 
for windows refer: https://www.w3.org/WAI/ARIA/apg/example-index/grid/dataGrids.html 
for mac: fn + option + right arrow
5. Come out of table : On press of tab key inside table anywhere, the focus should come out of table and move to next interactive element

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 13.x.x
   "@fundamental-ngx/core": "0.35.4",
    "@fundamental-ngx/platform": "0.35.4",

#### If this is a bug, please provide steps for reproducing it.
replicable on ngx site

#### Please provide relevant source code if applicable.
These are reproducible on ngx website : https://sap.github.io/fundamental-ngx/#/platform/table 

#### Is there anything else we should know?
Please contact us for any further information. These issues are blocker for Deutsche Telecom customer for guided sourcing.

mikerodonnell89:
@platon-rov assigning this to you since you have a better understanding of fdkFocusableGrid

I543348:
This is fixed

PR #9324, opened 2023-02-02
fix(cdk,core,platform): table keyboard

platon-rov:
## Related Issue(s)

Closes https://github.com/SAP/fundamental-ngx/issues/9131
Closes https://github.com/SAP/fundamental-ngx/issues/9132
Closes https://github.com/SAP/fundamental-ngx/issues/9134
Closes https://github.com/SAP/fundamental-ngx/issues/9135

## Description

* feat(cdk): focusable grid extended to support focusable rows and cells at the same time
* feat(cdk): focusable grid extended to support page-down, page-up keys
* feat(core): cdk focusable grid used in core table
* feat(platofrm): cdk focusable grid used in platform table


libs/docs/cdk/focusable-grid/examples/default-example/default-example.component.html
```diff
@@ -1,5 +1,10 @@
-<div fdkFocusableGrid wrapHorizontally (itemFocused)="itemFocused($event)">
-    <div *ngFor="let list of [].constructor(3); let rowIndex = index" fdkFocusableList navigationDirection="grid">
+<div fdkFocusableGrid wrapHorizontally (itemFocused)="itemFocused($event)" (rowFocused)="rowFocused($event)">
```
mikerodonnell89, line 1: Now when I focus a cell in the first row, then move focus to a cell in the next row, the whole row is focused, and I cannot see a way to go back to just focusing an individual cell. Is this intentional?

libs/docs/core/table/examples/table-focusable-example/table-focusable-example.component.html
```diff
@@ -10,11 +10,11 @@
...
-            <td fd-table-cell>Whole</td>
-            <td fd-table-cell>Row</td>
-            <td fd-table-cell>Can also</td>
-            <td fd-table-cell>Be</td>
-            <td fd-table-cell>Focused</td>
+            <td fd-table-cell [focusable]="true">Whole</td>
+            <td fd-table-cell [focusable]="true">Row</td>
+            <td fd-table-cell [focusable]="true">Can also</td>
+            <td fd-table-cell [focusable]="true">Be</td>
+            <td fd-table-cell [focusable]="true">Focused</td>
```
mikerodonnell89, line 17: Just wondering, any reason for this instead of using the allCellsFocusable input on the table?

mikerodonnell89 (commented):
Looking good - couple comments above. Hard to read the diff, maybe due to linting?

libs/docs/cdk/focusable-grid/examples/default-example/default-example.component.html
platon-rov: Yea, it's intended and to focus cell inside a focusable row use F2 key, I mentioned it in the docs.

mikerodonnell89 (approved):

libs/cdk/src/lib/utils/directives/focusable-list/focusable-list.directive.ts
```diff
@@ -239,15 +375,27 @@ export class FocusableListDirective implements OnChanges, AfterViewInit {
...
                     this._keydown$.next({ list, event, activeItemIndex: this._keyManager?.activeItemIndex });
 
+                    // Already handled
                     if (event.defaultPrevented) {
                         return;
                     }
 
+                    // Prevent scrolling and other default actions
+                    // But allow tabbing in/out and F2 to jump into list
+                    if (!KeyUtil.isKeyCode(event, [TAB, F2])) {
```
mikerodonnell89, line 385: We still need to be able to click buttons via keyboard, enter text into inputs, etc when these are all in focusable lists

mikerodonnell89 (changes requested):
see comment about catch-all preventDefault

mikerodonnell89:
add-on PR: https://github.com/SAP/fundamental-ngx/pull/9368

libs/cdk/src/lib/utils/directives/focusable-list/focusable-list.directive.ts
platon-rov: That's a nice catch, thanks. Fixed.

mikerodonnell89 (approved):

mikerodonnell89 (changes requested):
Just saw your comment here: https://github.com/SAP/fundamental-ngx/pull/9368#pullrequestreview-1295428127

We need to be able to move in to the focusable children with enter as well as f2, and need to be able to move back to the grid with escape

g-cheishvili:
Please mention in [this section](https://deploy-preview-9324--fundamental-ngx.netlify.app/#/core/table#focusable) that clicking `escape` will get you back to the focusable column when you are focused on it's content

libs/core/src/lib/table/directives/table-cell.directive.ts
```diff
@@ -8,29 +7,19 @@ import {
...
-    col: number;
-    totalRows: number;
-    totalCols: number;
-    origin: HTMLTableSectionElement;
-}
 
 @Directive({
     selector: '[fdTableCell], [fd-table-cell]',
-    providers: [DestroyedService, TabbableElementService]
+    hostDirectives: [FocusableItemDirective]
```
g-cheishvili, line 20: I think here you could alias the focusable item's `fdkFocusableItem` property to be `focusable`

g-cheishvili: [here is the example](https://angular.io/guide/directive-composition-api#:~:text=Furthermore%2C%20you%20can%20alias%20inputs%20and%20outputs%20from%20hostDirective%20to%20customize%20the%20API%20of%20your%20component%3A)

libs/core/src/lib/table/directives/table-cell.directive.ts
```diff
@@ -8,29 +7,19 @@ import {
...
     forwardRef,
-    Optional,
-    HostListener
+    inject,
+    Output,
+    EventEmitter
 } from '@angular/core';
 import { CheckboxComponent } from '@fundamental-ngx/core/checkbox';
-import { DestroyedService, TabbableElementService } from '@fundamental-ngx/cdk/utils';
+import { CellFocusedEventAnnouncer, FocusableItemDirective, FocusableItemPosition } from '@fundamental-ngx/cdk/utils';
```
g-cheishvili, line 15: optimize imports here?

libs/core/src/lib/table/directives/table-cell.directive.ts
g-cheishvili: Same goes for output

g-cheishvili (changes requested):
Minor stuff

libs/core/src/lib/table/directives/table-cell.directive.ts
platon-rov: Should be fine now.

g-cheishvili (approved):

mikerodonnell89 (approved):
approving, works great but I think we need to address the failing e2e, guessing there's a good chance it's related and not a fluke failure

Issue #9178, opened 2023-01-04
WizardGeneratorFormItem does not support properties to enable secondaryText


MichaelBaySAP:
To enable secondaryText on Select control, properties like secondaryKey, showSecondaryText and secondaryTextAlignment are required to be set on the WizardGeneratorFormItem. However, as of now, it is not possible to set them as the interface `WizardGeneratorFormItem` does not support them. Would it be possible to extend the interface by optional properties to enable secondaryTexts?

Here, I have built a Stackblitz for this. (https://stackblitz.com/edit/angular-4qb9ea?file=src%2Fapp%2Fwizard-generator-default-example.component.ts)

littleamigo:
From https://fundamental-ngx.netlify.app/#/platform/select/api:

> Optional secondaryKey
secondaryKey: string

> showSecondaryText
showSecondaryText: boolean = false

> secondaryTextAlignment
secondaryTextAlignment: TextAlignment = 'right'

littleamigo:
It works as expected - thanks, Mike!

PR #9326, opened 2023-02-02
fix(platform): dynamic form item should support select controlTemplate to improve two-column rendering

littleamigo:
fixes https://github.com/SAP/fundamental-ngx/issues/9178

Adds some properties from select to dynamic form item, allowing further customization of select in form generator/wizard generator

### DIFF
--- a/libs/platform/src/lib/form/form-generator/interfaces/dynamic-form-item.ts
+++ b/libs/platform/src/lib/form/form-generator/interfaces/dynamic-form-item.ts
@@ -1,6 +1,7 @@
 import { AsyncValidatorFn, ValidatorFn } from '@angular/forms';
 import { FormStates } from '@fundamental-ngx/cdk/forms';
 import { Observable } from 'rxjs';
+import { TemplateRef } from '@angular/core';
 
 import { ContentDensity } from '@fundamental-ngx/cdk/utils';
 import {
@@ -251,7 +252,7 @@ export interface DynamicFormFieldItem {
     secondaryTextAlignment?: TextAlignment;
 
     /** Custom template used to build control body. */
-    controlTemplate: TemplateRef<any>;
+    controlTemplate?: TemplateRef<any>;
 }
 
 type PreparedDynamicFormFieldItemFields = {
--- a/libs/platform/src/lib/form/form-generator/interfaces/dynamic-form-item.ts
+++ b/libs/platform/src/lib/form/form-generator/interfaces/dynamic-form-item.ts
@@ -249,6 +249,9 @@ export interface DynamicFormFieldItem {
 
     /** Horizontally align text inside the second column (Applicable for two columns layout) */
     secondaryTextAlignment?: TextAlignment;
+
+    /** Custom template used to build control body. */
+    controlTemplate: TemplateRef<any>;
 }
 
 type PreparedDynamicFormFieldItemFields = {
--- a/libs/core/src/lib/select/select.component.ts
+++ b/libs/core/src/lib/select/select.component.ts
@@ -172,7 +172,7 @@ export class SelectComponent<T = any>
 
     /** Custom template used to build control body. */
     @Input()
-    controlTemplate: TemplateRef<any>;
+    controlTemplate?: TemplateRef<any>;
 
     /** The element to which the popover should be appended. */
     @Input()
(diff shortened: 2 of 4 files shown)


libs/platform/src/lib/form/form-generator/interfaces/dynamic-form-item.ts
```diff
@@ -237,6 +237,9 @@ export interface DynamicFormFieldItem {
 
     /** Horizontally align text inside the second column (Applicable for two columns layout) */
     secondaryTextAlignment?: TextAlignment;
+
+    /** Custom template used to build control body. */
+    controlTemplate: TemplateRef<any>;
```
N1XUS, line 242: - [ ] TemplateRef is not imported;
- [ ] Property should be optional

N1XUS (changes requested):

libs/platform/src/lib/form/form-generator/interfaces/dynamic-form-item.ts
g-cheishvili: I do not like this. This interface is intended for use in every control, if you enhance select to use this, then other provided controls should use this as well.
For achieving custom stuff, I would go for [this approach](https://sap.github.io/fundamental-ngx/#/platform/form-generator#custom)

littleamigo: I do not understand your proposal, I'm afraid. Select is not a custom control act. https://sap.github.io/fundamental-ngx/#/platform/wizard-generator > Wizard Generator Elements: select.
In addition, already 3 properties from select were added: "secondaryKey, showSecondaryText, secondaryTextAlignment" recently by mikerodonnell89 and he asked me to add this one myself as we figured this was missing in the first place.

g-cheishvili: @littleamigo I created issue, which describes this. https://github.com/SAP/fundamental-ngx/issues/9349

It would be nice if we made this happen this way

platon-rov (approved):
As agreed here https://github.com/SAP/fundamental-ngx/issues/9349 we can merge this one to unblock customers and then think about refactoring.

g-cheishvili:
> As agreed here #9349 we can merge this one to unblock customers and then think about refactoring.

Yes, but without `TemplateRef` import and optional, this will not be buildable and we can not merge it

mikerodonnell89:
> > As agreed here #9349 we can merge this one to unblock customers and then think about refactoring.
> 
> Yes, but without `TemplateRef` import and optional, this will not be buildable and we can not merge it

Looks like these have already been addressed

littleamigo:
Let me know if there is anything missing by me!

I'd also accept a better default body template implementation for the 2-column select which renders the 2 values with a separating colon, for example. The default implementation is certainly for 99% of the use cases inappropriate:
image
image

littleamigo:
@g-cheishvili @mikerodonnell89 @N1XUS @platon-rov How are we going to merge this thing?

littleamigo:
it look like commit messages have to obey specific rules? why is that? looks strange to me because i'm used to have as many commit as i want in the way i want to achieve the goal of my pull request. i typically squash and merge them at the end. then, only the pull request title goes into the single commit message. this is the one worth to be checked for specific rules FMPOV.

littleamigo:
> > it looks like commit messages have to obey specific rules? why is that? looks strange to me because i'm used to have as many commit as i want in the way i want to achieve the goal of my pull request. i typically squash and merge them at the end. then, only the pull request title goes into the single commit message. this is the one worth to be checked for specific rules FMPOV.
> 
> Could you just fix the last commit and rebase on main? What you're saying makes sense and I think it is the better approach, would you mind creating an issue for that?

@g-cheishvili not sure what exactly you want me to do:
- rebase on main -> no problem, i can do
- fix the last commit -> what is wrong with the last [commit](https://github.com/SAP/fundamental-ngx/pull/9326/commits/0c52b48ab14ac77d4ec0cf8ff6c24e1fecd3db08)? or do you want me to revert the last 2 commits because they are not required?

g-cheishvili:
> > > it looks like commit messages have to obey specific rules? why is that? looks strange to me because i'm used to have as many commit as i want in the way i want to achieve the goal of my pull request. i typically squash and merge them at the end. then, only the pull request title goes into the single commit message. this is the one worth to be checked for specific rules FMPOV.
> > 
> > 
> > Could you just fix the last commit and rebase on main? What you're saying makes sense and I think it is the better approach, would you mind creating an issue for that?
> 
> @g-cheishvili not sure what exactly you want me to do:
> 
> * rebase on main -> no problem, i can do
> * fix the last commit -> what is wrong with the last [commit](https://github.com/SAP/fundamental-ngx/pull/9326/commits/0c52b48ab14ac77d4ec0cf8ff6c24e1fecd3db08)? or do you want me to revert the last 2 commits because they are not required?

just fix the name to be compliant with conventional commits, that's all

N1XUS:
Closing in favour of #9497

littleamigo:
N1XUS confirmed he did some refactorings to the form generator and applied my proposed changes to it

PR #9346, opened 2023-02-06
feat(platform): added id property to form generator items

### LINKED ISSUE #8729: Fundamentals Wizard Generator: Mark every control with data-testid
feature request

Problem: As there are no data-testids, in test automation tools like Cypress, we have to locate elements by id or label. The id is something like fdp-id-9 and changes over time. The label is not a unique identifier if there exist multiple labels with the same name.

Proposed solution:

Define an additional data-testid property for each wizard control that the developer can set

By default, set every data-testid to the "name" property control of the property as this is sort of the ID.

As select-items and radio-button items are also distinct html elements, they should also carry data-testids which could be a combination of the name of the select/radio-button and the text for the item.

`dxp/template-ui/issues/275`

g-cheishvili:
## Related Issue(s)

closes #8729 

## Description

Added `id` property to the `BaseDynamicFormGeneratorControlInterface` and made dynamic fields respect that. Previously `name` would have been used as an `id` for control fields, now users can use different id for their DOM elements.
As before, `fdp-form-label-${id || name}` is used for labels, `fdp-form-label-content-${id || name}` is used for label contents and additionally `id || name` is now used for the form controls themselves. Placement and usage of that passed id depend on the control itself, but provided controls are attaching those ids to the correct places.

Custom controls will have to implement that by themselves, examples that are given for custom controls are fixed

### DIFF
--- a/libs/platform/src/lib/form/form-generator/dynamic-form-generator-checkbox/dynamic-form-generator-checkbox.component.html
+++ b/libs/platform/src/lib/form/form-generator/dynamic-form-generator-checkbox/dynamic-form-generator-checkbox.component.html
@@ -1,6 +1,7 @@
 <ng-container [formGroup]="form">
     <ng-container [formGroupName]="formGroupName">
         <fdp-checkbox-group
+            [id]="id"
             [isInline]="formItem.guiOptions?.inline || false"
             [inlineLayout]="formItem.guiOptions?.inlineLayout"
             [name]="name"
--- a/libs/docs/platform/smart-filter-bar/examples/platform-smart-filter-bar-custom-filter-example.component.ts
+++ b/libs/docs/platform/smart-filter-bar/examples/platform-smart-filter-bar-custom-filter-example.component.ts
@@ -65,6 +65,7 @@ export class PlatformSmartFilterBarSliderComponent extends BaseDynamicFormGenera
         <ng-container [formGroup]="form">
             <ng-container [formGroupName]="formGroupName">
                 <fdp-date-picker
+                    [id]="id"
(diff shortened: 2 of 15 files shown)


libs/platform/src/lib/form/form-generator/dynamic-form-generator-checkbox/dynamic-form-generator-checkbox.component.html
```diff
@@ -1,6 +1,7 @@
 <ng-container [formGroup]="form">
     <ng-container [formGroupName]="formGroupName">
         <fdp-checkbox-group
+            [id]="id"
```
mikerodonnell89, line 4: I think we want an ID for each checkbox not just the whole group. Same for radio

mikerodonnell89 (commented):
Do we want to automatically apply a generated ID for each form control like this PR does for each form group?

libs/platform/src/lib/form/form-generator/dynamic-form-generator-checkbox/dynamic-form-generator-checkbox.component.html
g-cheishvili: That's a tough question. I think having ID on the main group should be enough

platon-rov (approved):

libs/platform/src/lib/form/form-generator/dynamic-form-generator-checkbox/dynamic-form-generator-checkbox.component.html
g-cheishvili: Resolved?

PR #9350, opened 2023-02-07
fix: overflow layout memory leaks

### LINKED ISSUE #9345: overflow layout memory leaks
bug
Currently, overflow layout does not completely destroys, thus keeping links to elements/components/directives inside.

How to reproduce:
Follow this instructions on how to collect detached elements:
https://learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/memory-problems/#discover-detached-dom-tree-memory-leaks-with-heap-snapshots

Go to overflow layout page, then go to home page, run Collect garbage and then click "Get Detached elements"

N1XUS:
## Related Issue(s)

closes #9345

## Description
- Fixed overflow layout memory leaks;
- Fixed memory leaks from focusKeyManager if used with QueryList;
- Fixed animations in dialog;
- Unit tests are now using NoopAnimations module instead of BrowserAnimationsModule;
- Fixed dialog and message box memory leaks;

### DIFF
--- a/libs/core/src/lib/overflow-layout/directives/overflow-layout-item.directive.ts
+++ b/libs/core/src/lib/overflow-layout/directives/overflow-layout-item.directive.ts
@@ -67,7 +67,7 @@ export class OverflowLayoutItemDirective implements OverflowItem, OnInit {
 
     /** @hidden */
     ngOnInit(): void {
-        this._overflowItemRef?.setElementRef(this.elmRef);
+        // this._overflowItemRef?.setElementRef(this.elmRef);
         this._overflowItemRef?.setOverflowItem(this);
     }
 }
--- a/libs/core/src/lib/overflow-layout/directives/overflow-layout-item.directive.ts
+++ b/libs/core/src/lib/overflow-layout/directives/overflow-layout-item.directive.ts
@@ -67,7 +67,6 @@ export class OverflowLayoutItemDirective implements OverflowItem, OnInit {
 
     /** @hidden */
     ngOnInit(): void {
-        // this._overflowItemRef?.setElementRef(this.elmRef);
         this._overflowItemRef?.setOverflowItem(this);
     }
 }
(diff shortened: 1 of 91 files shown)


g-cheishvili (approved):

libs/core/src/lib/overflow-layout/directives/overflow-layout-item.directive.ts
```diff
@@ -67,7 +67,7 @@ export class OverflowLayoutItemDirective implements OverflowItem, OnInit {
 
     /** @hidden */
     ngOnInit(): void {
-        this._overflowItemRef?.setElementRef(this.elmRef);
+        // this._overflowItemRef?.setElementRef(this.elmRef);
```
platon-rov, line 70: can be removed?

libs/core/src/lib/overflow-layout/overflow-layout.component.ts
```diff
@@ -309,17 +326,46 @@ export class OverflowLayoutComponent implements OnInit, AfterViewInit, OnDestroy
...
-                        this._overflowLayoutService._keyboardEventsManager.onKeydown(event);
+                        this._keyboardEventsManager?.onKeydown(event);
                     }
                 })
         );
     }
+
+    /** @hidden */
+    private _setFocusKeyManager(): void {
+        if (!this.enableKeyboardNavigation) {
```
platon-rov, line 345: can this be done with focusable-list cdk thing? probably not in scope of this pr

wdio_docs-core-button.conf.js
```diff
@@ -0,0 +1,5 @@
+module.exports.config = require('./e2e/wdio.base-config.js')({
```
platon-rov, line 1: probably should be removed.

libs/core/src/lib/overflow-layout/overflow-layout.component.ts
N1XUS: I would suggest other PR for that

PR #8903, opened 2022-10-26
fix(platform): implementation of expand and collapse all feature in table

skaquib123:
## Related Issue(s)

closes #8901 

## Description

Implemented expand and collapse all feature in table where we can expand and collapse all the rows of the table.

### Before:
image

##### During Implementation

1. Visual Testing:

-   [x] visual misalignments/updates
-   [x] check Light/Dark/HCB/HCW themes
-   [x] RTL/LTR - proper rendering and labeling
-   [x] responsiveness(resize)
-   [x] Content Density (Cozy/Compact/(Condensed))
-   [x] States - hover/disabled/focused/active/on click/selected/selected hover/press state
-   [x] Interaction/Animation - open/close, expand/collapse, add/remove, check/uncheck
-   [x] Mouse vs. Keyboard support
-   [x] Text Truncation

2. API and functional correctness

-   [x] check for console logs (warnings, errors)
-   [x] API boundary values
-   [x] different combinations of components - free style
-   [x] change the API values during testing

3. Documentation and Example validations

-   [x] missing API documentation or it is not understandable
-   [x] poor examples
-   [x] Stackblitz works for all examples

4. Accessibility testing
5. Browser Testing - Edge, Safari, Chrome, Firefox

##### PR Quality

-   [x] the commit message(s) follows the guideline:
        https://github.com/SAP/fundamental-ngx/blob/main/CONTRIBUTING.md
-   [x] tests for the changes that have been done
-   [x] all items on the PR Review Checklist are addressed :
        https://github.com/SAP/fundamental-ngx/wiki/PR-Review-Checklist
-   [x] Run npm run build-pack-library and test in external application
-   [x] update `README.md`
-   [x] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)

PR #9351, opened 2023-02-07
fix(platform): table drag and drop bugs

### LINKED ISSUE #9309: bug(platform): dnd in tree table is broken

Bug.

1. "Other 3" not being placed instead of "Laptops",
2. "Laptops" remain in active state.
3. "Other 2" lost his nested rows when "Screens" being placed on it.

https://user-images.githubusercontent.com/20265336/215847165-9ee7cfc2-8930-4900-98d2-3b436f5ac133.mov

Latest.

mikerodonnell89:
fixes #9309 

Couple things in this PR:

1. new CDK dragEnd function returning too early due to [0 being falsy](https://github.com/SAP/fundamental-ngx/compare/fix/platform-table-dnd-9309?expand=1#diff-3fc261727cd9abdb353a666bbb833911634addc69ff0d2611031ae94ede7bcd3L146)
2. Updated documentation to not have duplicate rows
3. Reverting changes from [this commit](https://github.com/SAP/fundamental-ngx/pull/8903/commits/6a4df64dd428900140fb17f806e0287b0aa41416) - must've slipped under the radar during the review process (@platon-rov noticed but still the PR got merged). This PR was for expand/collapse all, not related to drag and drop, this commit changes the original behavior of the drag and drop. I think what we really need is a way to not only drag and drop child rows into parent rows but also a way to drag them out of parent rows, or rearrange them.


platon-rov (approved):
This pr fixes the mentioned issue, although drag'n'drop still isn't ideal. It's not possible to drop "inside" or "instead" so we end up in a situation when after you placed a row (1) inside another row you cannot pull out that row (1), can we create a ticket to fix that?

mikerodonnell89:
> 

I agree, mentioned the same thing in point #3. I think for this we will also need additional styles. There needs to be a visual distinction between dropping into an item vs. dropping next to an item

InnaAtanasova (approved):

PR #9366, opened 2023-02-10
fix(core): use static element ref property

N1XUS:
Changed view child references to element ref for some components to reduce the level of inclusion.


mikerodonnell89 (dismissed):
Something with platform radio is not working, in the first example I can no longer use the arrow keys to navigate through the radio buttons in the group

g-cheishvili (approved):
I already like where this is going. Approve

InnaAtanasova (approved):

PR #9368, opened 2023-02-10
[WIP/blocked] fix(cdk,platform): table input arrow keys

mikerodonnell89:
According to the accessibility guidelines, table cells that are navigable via keyboard should not automatically focus their inner focusable children when navigated to. User should press enter/f2 to change to "edit mode" where the inner child will be focused. They can press enter/f2 again or escape to leave edit mode. 

Note that this PR does not have a base of main but instead the table keyboard branch by @platon-rov


platon-rov (changes requested):
Hey, when you noticed this https://github.com/SAP/fundamental-ngx/pull/9324#discussion_r1102135587 i realized that the only way to fix it by switching to this f2-based solution. So i applied it and this pr should be closed thou.

mikerodonnell89:
Closing in favor of #9324

Issue #8985, opened 2022-11-22
multi input vs. multi combo


droshev:
#### Is this a bug, enhancement, or feature request?
bug and enhancement (maybe even a breaking change)

#### Briefly describe your proposal.
The current implementations of `core` `multi-input` and `platform` `multi-input` and `multi-combo` do not follow the design guidelines at 100%. The goal of this ticket is to:
- [x] introduce `multi-combo` in `core` package
- [ ] review the existing implementations and apply the proper styling and behaviour, especially, the more items and reviewing the tokens

Design Guidelines
- https://experience.sap.com/fiori-design-web/multi-combobox/
- https://experience.sap.com/fiori-design-web/multiinput/
- https://experience.sap.com/fiori-design-web/multiinput/#reviewing-tokens

additional items to be considered transferred from #7611 

- [x] Multi-input - revisit the keyboard - TBD
- [x] Multi-input: [this PR](https://github.com/SAP/fundamental-ngx/pull/7588) allows the user to tab to the expand key to open the dropdown of selectable options. However the user then must shift+tab back to the input part of the multi-input, then press tab, to focus the first item in the list. Instead we should have an additional tab keypress, or down arrow press, focus the first item in the dropdown
- [x] Multi-input: both the checkbox and the entire list item receive tab focus
- [x] Multi-input (core): adding new tokens example. Type a new token and submit it with the enter key. The token is not added to the tokenizer and when opening the dropdown, is not visible in the list. However you will notice the token has been added to the total count (the "View All" count as increased) and the new item has been added next to the "Selected:" text. - was already fixed 
![image](https://user-images.githubusercontent.com/6586561/219009237-665f3d90-eeb6-456f-ace7-f7b64f007f61.png)

droshev:
Maybe we can combine this task with bringing all the platform features into core and deprecating the platform ones?

N1XUS:
Additionally. in platform multi-input component it's impossible to unset the selected item by clicking on the checkbox, or on the item itself

PR #9390, opened 2023-02-15
fix(platform): multi-input bugfixes

N1XUS:
## Related Issue(s)

relates #8985

## Description
- Fixed `reset` functionality with reactive forms example;
- Fixed focused item when dropdown is opened;
- Fixed issue with focusing the checkbox;

### DIFF
--- a/libs/platform/src/lib/form/multi-input/multi-input.component.ts
+++ b/libs/platform/src/lib/form/multi-input/multi-input.component.ts
@@ -373,14 +373,14 @@ export class PlatformMultiInputComponent extends BaseMultiInput implements OnIni
     setAsSelected(item: MultiInputOption[]): void {
         this._selected = item;
         this.inputText = '';
-        this.markListItemsAsSelected();
+        this._markListItemsAsSelected();
     }
 
     /**
      * @hidden
      * Mathod for marking items in dropdown as selected.
      */
-    markListItemsAsSelected(): void {
+    _markListItemsAsSelected(): void {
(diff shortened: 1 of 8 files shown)


InnaAtanasova:
Regarding this part of the fix: `Fixed focused item when dropdown is opened;`
Is it an expected behaviour? When I type the letter `a`, the first item is focussed but I can no longer type in the input field. What if there's a very long list of items and the user needs to scroll through 100+ items to reach the desired item when they can write 2nd or 3rd letter and get there faster?

Screenshot 2023-02-15 at 3 22 36 PM

InnaAtanasova:
First item focus issue steps to reproduce:
type the letter `a` => the first item of the list is selected=>press the arrow down to go to the item below=> you need to press the arrow key twice to move to the other item.

Screenshot 2023-02-15 at 3 28 13 PM

libs/platform/src/lib/form/multi-input/multi-input.component.ts
```diff
@@ -366,6 +373,18 @@ export class PlatformMultiInputComponent extends BaseMultiInput implements OnIni
...
         this._selected = item;
         this.inputText = '';
+        this.markListItemsAsSelected();
+    }
+
+    /**
+     * @hidden
+     * Mathod for marking items in dropdown as selected.
+     */
+    markListItemsAsSelected(): void {
```
platon-rov, line 383: if hidden let's add `_` prefix

platon-rov (approved):

PR #9391, opened 2023-02-15
chore(ci,deps): latest typescript, nx plugins refactoring, linting issues

platon-rov:
## Related Issue(s)

Closes none.

## Description

Fixes broken "sap component" nx generator by bumping up typescript version. Additionally, minor refactoring was done.

## Screenshots

No visual changes.

PR #9403, opened 2023-02-17
fix: type error

mikerodonnell89:
fixes none

Type error, possibly related to new dependencies in #9391 but this fixes it ahead of time

### DIFF
--- a/libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
+++ b/libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
@@ -84,7 +84,7 @@ export class FocusableItemDirective implements HasElementRef {
     private _tabbable = true;
 
     /** @hidden */
-    private _timerId: number | null;
+    private _timerId: any | null;
 
     /** @hidden */
     @HostBinding('attr.tabindex')
--- a/libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
+++ b/libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
@@ -84,7 +84,7 @@ export class FocusableItemDirective implements HasElementRef {
     private _tabbable = true;
 
     /** @hidden */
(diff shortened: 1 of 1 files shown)


libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
```diff
@@ -84,7 +84,7 @@ export class FocusableItemDirective implements HasElementRef {
     private _tabbable = true;
 
     /** @hidden */
-    private _timerId: number | null;
+    private _timerId: any | null;
```
N1XUS, line 87: You can use https://www.typescriptlang.org/docs/handbook/utility-types.html#returntypetype:
```
private _timerId: ReturnType<typeof setTimeout> | null;
```

N1XUS (approved):

libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
```diff
@@ -84,7 +84,7 @@ export class FocusableItemDirective implements HasElementRef {
     private _tabbable = true;
 
     /** @hidden */
-    private _timerId: number | null;
+    private _timerId: ReturnType<typeof setTimeout> | null;
```
g-cheishvili, line 87:
```suggestion
    private _timerId: ReturnType<typeof global.setTimeout> | null;
```
This should fix it. Basically ts confuses node and web environments

libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
mikerodonnell89: thanks!

libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
mikerodonnell89: sounds good

PR #9420, opened 2023-02-20
fix: platform table keyboard dnd

### LINKED ISSUE #9137: [Accessibility][Sourcing]: Drag drop table rows not possible using keyboard
Bug

User is not able to perform drag and drop of rows using keyboard
This is currently possible only using mouse and is blocker for disabled users

Angular 13.x.x
   "@fundamental-ngx/core": "0.35.4",
    "@fundamental-ngx/platform": "0.35.4",

Try to drag and drop a fdp-table row using keyboard

Blocker for guided sourcing customers

mikerodonnell89:
fixes #9137 

Focus a cell, press and hold alt/option key and use the up/down arrow keys to move it. Currently has the same behavior as mouse drag and drop


platon-rov (changes requested):
Behavior doesn't look predictable, some when moving up/down doesn't take effect, here is a short video

https://user-images.githubusercontent.com/20265336/220942854-5cb92d3f-76f0-491e-990c-6307a747aab2.mov

mikerodonnell89:
> Behavior doesn't look predictable, some when moving up/down doesn't take effect, here is a short video

This is simply working the same way as if you were to click and drag the item one row up or down, so I think this can be considered the intended behavior. We have been talking about needing various platform table dnd improvements, I've created an issue here https://github.com/SAP/fundamental-ngx/issues/9421

g-cheishvili (changes requested):
After one successful drag with keyboard, focus was lost and it is kinda hard to guess where next action will go 😀

g-cheishvili:
After you group something with `alt/option+arrow down` you can not ungroup it. Focus is fixed though.
The same thing happens with mouse interaction, if you drag one item into another, there is no way to ungroup it, dragging child item above parent does not do anything

mikerodonnell89:
> After you group something with `alt/option+arrow down` you can not ungroup it. Focus is fixed though. The same thing happens with mouse interaction, if you drag one item into another, there is no way to ungroup it, dragging child item above parent does not do anything.
> 
> Expectation probably is that after arrow down action, arrow up will set child on a level of a parent

Yes, we'll worry about that here: #9421

Issue #9228, opened 2023-01-16
Defect Hunting Sprint 107


InnaAtanasova:
Please list the issues you've found during the defect hunting.

mikerodonnell89:
A handful of https://github.com/SAP/fundamental-ngx/issues/9225 still need some work

droshev:
- [ ] platform table with toolbar - https://fundamental-ngx.netlify.app/#/platform/table make the toolbar with overflow

mikerodonnell89:
All on Core:
- [ ] Action sheet - clicking the control button when the action sheet is already open does not close it. This is not the case with menu, popover, etc

- [ ] Avatar group, group type example, there are certain parts of the overflow button you can click that won't open the popover
<img width="216" alt="Screenshot 2023-02-20 at 5 50 31 PM" src="https://user-images.githubusercontent.com/2471874/220220912-c09a3b5e-bc0a-4356-b0ae-39776113d121.png">

- [ ] List - sometimes pressing shift+tab to focus the element before the list doesn't work. Doesn't always happen

- [ ] List - mouse click the 5th item in the list to focus it. Then press shift+tab. This moves the focus to the first item in the list. Press shift - this moves the focus back to the 5th item

- [ ] Multi-combobox - cannot navigate to or select items from the popover list via keyboard only

- [ ] Multi-input - when I first open the complex popover, only the state text displays. The secondary text displays on subsequent popover closes/opens. Before:
<img width="858" alt="Screenshot 2023-02-20 at 6 09 04 PM" src="https://user-images.githubusercontent.com/2471874/220222582-7a8c506f-2ce1-4140-9202-da3775f44d3a.png">
After:
<img width="592" alt="Screenshot 2023-02-20 at 6 09 42 PM" src="https://user-images.githubusercontent.com/2471874/220222643-a4a55e6a-8531-4882-bbb3-7e68f926491d.png">

- [ ] Search field popover items missing focus radius
<img width="229" alt="Screenshot 2023-02-20 at 6 12 41 PM" src="https://user-images.githubusercontent.com/2471874/220222958-638e7119-6f37-41ec-a620-c56ee3d52052.png">

- [ ] Compact select, control button focus - cannot see the top and bottom sides of the focus box.
compact:
[...]

PR #9435, opened 2023-02-22
fix(core): fixed dialog body redundant elements

g-cheishvili:
## Related Issue(s)

part of #9228 

## Description

Updated scrollbar component to be the directive, instead of component


libs/core/src/lib/scrollbar/scrollbar.directive.ts
```diff
@@ -96,7 +100,24 @@ export class ScrollbarComponent {
     /**
      * @hidden
      */
-    constructor(private _elementRef: ElementRef<HTMLElement>) {}
+    constructor(private _elementRef: ElementRef<HTMLElement>) {
+        scrollbarElementsQuantity++;
+        const platform = inject(PLATFORM_ID);
+        if (!styleSheet && isPlatformBrowser(platform)) {
+            styleSheet = new CSSStyleSheet();
+            styleSheet.replace(scrollbarStyles.cssSource).then(() => {
```
N1XUS, line 108: You can use `replaceSync` so you don't have to work with promises

libs/core/src/lib/scrollbar/scrollbar.module.ts
```diff
@@ -1,11 +1,11 @@
...
 
-import { ScrollbarComponent } from './scrollbar.component';
+import { ScrollbarDirective } from './scrollbar.directive';
 
 @NgModule({
-    declarations: [ScrollbarComponent],
-    imports: [CommonModule],
-    exports: [ScrollbarComponent]
+    declarations: [],
+    imports: [CommonModule, ScrollbarDirective],
```
N1XUS, line 8: Do you need `CommonModule` at all?

N1XUS (approved):

Issue #9030, opened 2022-12-02
Is there a way I can freeze the last one or several columns of <fdp-table>?


Menoen:
#### Is this a bug, enhancement, or feature request?
Feature request
#### Briefly describe your proposal.
I notice that the fdp-table has a input "freezeColumnsTo", which can freeze several columns from left, but if there a way I can only freeze one column on the right but not to affect others?

PR #9467, opened 2023-03-01
feat(core,platform): update fundamental-styles version, adopt breaking changes

mikerodonnell89:
fixes none

### DIFF
--- a/libs/core/src/lib/table/table.component.ts
+++ b/libs/core/src/lib/table/table.component.ts
@@ -62,6 +62,16 @@ export class TableComponent implements AfterContentInit, AfterViewInit {
     @Input()
     noBorderY = false;
 
+    /** Whether or not to show the table's outer border */
+    @HostBinding('class.fd-table--no-outer-border')
+    @Input()
+    noOuterBorder = false;
+
+    /** Whether or not to show the table's top border */
+    @HostBinding('class.fd-table--top-border')
+    @Input()
+    topBorder = false;
+
     /** Whether or not to display the table in pop in mode, it also require change of markup */
     @HostBinding('class.fd-table--pop-in')
     @Input()
--- a/libs/docs/core/table/examples/table-fixed-example.component.html
+++ b/libs/docs/core/table/examples/table-fixed-example.component.html
@@ -1,0 +1,41 @@
+<div fd-table-fixed style="height: 300px">
+    <table fd-table aria-label="Default">
+        <thead fd-table-header>
+            <tr fd-table-row>
+                <th fd-table-cell>Column Header</th>
+                <th fd-table-cell>Column Header</th>
+                <th fd-table-cell>Column Header</th>
+                <th fd-table-cell>Column Header</th>
+                <th fd-table-cell>Column Header</th>
+            </tr>
+        </thead>
+        <tbody fd-table-body>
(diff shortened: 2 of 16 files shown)


libs/docs/core/table/examples/table-fixed-example.component.html
```diff
@@ -0,0 +1,41 @@
+<div fd-table-fixed style="height: 300px">
```
N1XUS, line 1: This actually breaks BEM. We use fd-table--fixed on the element that is not fd-table.
On my opinion, thats should be somehow resolved.

libs/platform/src/lib/table/table.component.scss
```diff
@@ -194,9 +194,12 @@ $fd-table-scrollbar-width-with-border: calc(#{$fd-table-scrollbar-width} + var(-
         .fd-table__cell {
             margin-top: 0;
             max-height: 100%;
-            position: sticky;
             top: 0;
 
+            &--fixed {
```
N1XUS, line 199: Platform docs example?

libs/docs/core/table/table-docs.component.html
```diff
@@ -185,13 +185,30 @@
 </component-example>
 <code-example [exampleFiles]="tablePagination"></code-example>
 
+<separator></separator>
+
+<fd-docs-section-title id="fixed" componentName="table"> Table with Header and Footer </fd-docs-section-title>
```
N1XUS, line 190: Missing `sticky` notice

N1XUS (changes requested):

libs/docs/core/table/examples/table-fixed-example.component.html
mikerodonnell89: We need to address this in styles (should have been part of https://github.com/SAP/fundamental-styles/pull/4214), and I guess may require substantial refactoring so not going to worry about this now.

libs/platform/src/lib/table/table.component.scss
mikerodonnell89: Coming in a later PR when addressing #9030

N1XUS (approved):

libs/core/src/lib/table/table.component.ts
```diff
@@ -62,6 +62,16 @@ export class TableComponent implements AfterContentInit, AfterViewInit {
     @Input()
     noBorderY = false;
 
+    /** Whether or not to show the table's outer border */
+    @HostBinding('class.fd-table--no-outer-border')
+    @Input()
+    noOuterBorder = false;
+
+    /** Whether or not to show the table's top border */
```
InnaAtanasova, line 70: The typedocs should mention the default value (false in this case)

libs/platform/src/lib/table/table.component.ts
```diff
@@ -212,6 +212,10 @@ export class TableComponent<T = any> extends Table<T> implements AfterViewInit,
     @Input()
     enableDragResize = true;
 
+    /** Whether to fix the table header and footer. Will be automatically set to true if therer are freezable columns.*/
```
InnaAtanasova, line 215: typo for `therer` and shouldn't it be "Will be automatically set to `false` if there are freezable columns.", since `fixed = false;` ?

InnaAtanasova (approved):

PR #9472, opened 2023-03-03
fix(platform): align date picker writeValue behavior with datetime picker

weihaolu:
## Description

when the date picker is supposed to write a value that is not pickable then it will set an empty string. Contrary to this the date time picker sets the value. Since the disable function should only be disabling what the user can pick, programatically set values (e.g.: through the form control) should be set and visible while validators handle what to do with such a value


mikerodonnell89:
Agreed that this should be the functionality. I think we'll need to verify this behavior for all input components in reactive forms

mikerodonnell89 (approved):

mikerodonnell89 (approved):

InnaAtanasova (approved):

InnaAtanasova:
Hi @weihaolu could you take a look at the failing checks? Thank you!

PR #9497, opened 2023-03-09
feat(platform): settings generator

### LINKED ISSUE #9432: User Menu and dialog in the shellbar
feature request

Create user menu and settings dialog component which will be added to the shellbar. It should look like [this](https://fundamental-styles.netlify.app/?path=/docs/components-user-menu--default).

- [x] Can we create it in platform and make it driven based on data (like wizard generator)?
- [x] Can we offer/include the theming selector (maybe integrate with the existing Theming functionality). The application should be able to specify the subset of the supported themes.
- [x] Document it with some examples

Out of scope: storing runtime the selected theme and being used in other applications.

N1XUS:
## Related Issue(s)

closes #9432

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [x] poor examples

4. Accessibility testing

##### PR Quality

-   [x] update `README.md`


libs/cdk/src/lib/utils/pipes/async-or-sync.pipe.ts
```diff
@@ -0,0 +1,135 @@
+import { ChangeDetectorRef, EventEmitter, OnDestroy, Pipe, PipeTransform, inject } from '@angular/core';
```
g-cheishvili, line 1: Can't we reuse `async` pipe here? Inject the pipe and invoke their transform function only in case of the non-static values. We would not need to support everything on our end

N1XUS: That might bring some memory leaks, so I would keep it like that for now.

N1XUS: I have another idea. Instead of loading pipe into pipe, I'll create `makeAsync` pipe which will transform any non-awaitable/subscribable variable into subscribable one, and then apply async pipe.

g-cheishvili: That way developer will need to use two pipes on one place

libs/core/src/lib/theming/theming.service.ts
```diff
@@ -18,13 +18,15 @@ import { THEMING_CONFIG_TOKEN } from './tokens';
...
+     * Theming configuration.
+     */
+    readonly config = new BaseThemingConfig();
 
     /**
-     * Current theme applied to the application.
+     * Observable of the current theme, applied to the application.
      */
-    get currentTheme(): Observable<CompleteThemeDefinition | null> {
+    get currentTheme$(): Observable<CompleteThemeDefinition | null> {
```
g-cheishvili, line 29: Add another getter, with name `currentTheme` and log in that getter that this is deprecated, and return `currentTheme$`

libs/cdk/src/lib/utils/pipes/async-or-sync.pipe.ts
N1XUS: Yea, OR he can make it async in his service/component. Since I'm against modifying passed initial data, pipe I think this is the best option here. Calling asyncPipe.transform directly may have unintentional side effects.

InnaAtanasova:
Screenshot 2023-03-15 at 1 13 15 PM
 I guess this issue comes from the Forms?

InnaAtanasova:
Screenshot 2023-03-15 at 1 08 25 PM
some overflow issues with the forms (inline help)

libs/platform/src/lib/smart-filter-bar/interfaces/smart-filter-dynamic-form-item.ts
```diff
@@ -3,9 +3,13 @@ import { SelectItem } from '@fundamental-ngx/platform/shared';
...
 
-export interface SmartFilterBarDynamicFormFieldItem extends DynamicFormFieldItem {
+export type SmartFilterBarDynamicFormFieldItem = DynamicFormFieldItem<{
     guiOptions: SmartFilterBarDynamicFormFieldGuiOptions;
-}
+}>;
+
+// export interface SmartFilterBarDynamicFormFieldItem extends DynamicFormFieldItem {
+//     guiOptions: SmartFilterBarDynamicFormFieldGuiOptions;
+// }
```
InnaAtanasova, line 12: maybe delete these if not needed

libs/platform/src/lib/wizard-generator/interfaces/wizard-generator-form-group.interface.ts
```diff
@@ -61,9 +61,11 @@ export interface WizardGeneratorFormGroupItem extends DynamicFormFieldGroup {
...
-    dependencyFields?: WizardGeneratorDependencyFields;
-}
+export type WizardGeneratorFormFieldItem = DynamicFormFieldItem<{ dependencyFields?: WizardGeneratorDependencyFields }>;
+
+// export type WizardGeneratorFormFieldItem extends DynamicFormFieldItem<{dependencyFields: WizardGeneratorDependencyFields}> {
+//     /**
+//      * @description Object of dependency fields that are used with `when` function
+//      */
+//     dependencyFields?: WizardGeneratorDependencyFields;
+// }
```
InnaAtanasova, line 71: needed?

N1XUS:
@InnaAtanasova Your comments has been addressed.
Now user can configure not just the width of the sidebar, but also mobile breakpoint when sidebar becomes 100%, and content is hidden by default, until user clicks on some sidebar item.

InnaAtanasova (approved):

mikerodonnell89:
Some visual things i noticed:

- No padding on the skeleton items here:

Screenshot 2023-03-20 at 12 57 26 PM


- Feels odd to me that the form is so far right aligned here, I think with screens at this width we should be using multiple columns such that these inputs are more left aligned

Screenshot 2023-03-20 at 12 51 57 PM

N1XUS:
> Some visual things i noticed:
> 
> * No padding on the skeleton items here:
> 
> Screenshot 2023-03-20 at 12 57 26 PM
> 
> * Feels odd to me that the form is so far right aligned here, I think with screens at this width we should be using multiple columns such that these inputs are more left aligned
> 
> Screenshot 2023-03-20 at 12 51 57 PM

Hello Mike. Thank you for your feedback.
As to skeleton padding - it's not the issue of settings generator, but more of form-generator with skeleton, since form generator uses grid with paddings inside, and skeleton completely ignores it. So it should be addressed separately.

As to the form alignment, it's just an example of the grid positioning of the items. Users can change the responsive grid layout as they want to.

PR #9502, opened 2023-03-10
fix(platform): table with freezing columns hidden cell issues

### LINKED ISSUE #9492: [Sourcing][Accessibility] fdp-table frozen column focus issue
Bug

1) When moving between left and right table cells in a frozen columns , the focus appears to be broken and looks like 2 cells are focussed at once , which will confuse the disabled user

2) Reach the table using keyboard , press right arrow after description column , notice that visible focus is on "price" column but its not actually visible to users , next time we press right arrow key - the focus is visible on "status" column. When the hidden column receives focus, it should be scrolled in to view

Angular 15
core / platform: 0.39.0

This is replicable on [9082](https://sap.github.io/fundamental-ngx/#/platform/table#freezable) 

223171162-f8ac7b9b-6bbd-4f0f-9df7-827fbe9260cb

This is needed for guided sourcing customer commitment

mikerodonnell89:
fixes #9492 

Previously if a non-frozen table cell was partially obstructed by a frozen one, the focus for the entire non-frozen cell was still visible. Also, if a non-frozen cell was entirely obstructed by a frozen one and was navigated via keyboard, the cell would not be scrolled to, and it would appear as though focus was lost (when it was simply on the hidden element). This fixes both issues

https://user-images.githubusercontent.com/2471874/224383490-f335beac-be47-4394-a032-203b12b0d898.mov

### DIFF
--- a/libs/docs/platform/table/e2e/table-contents.ts
+++ b/libs/docs/platform/table/e2e/table-contents.ts
@@ -7,6 +7,14 @@ export const tableCellArr = [
     '66.04',
     'Stocked on demand'
 ];
+export const freezeTableCellArr = [
+    '10 Portable DVD player',
+    'diam neque vestibulum eget vulputate',
+    '66.04',
+    'Stocked on demand',
+    '2020',
+    'true'
+];
 export const groupTableCellArr = [
     '10 Portable DVD player',
     'diam neque vestibulum eget vulputate',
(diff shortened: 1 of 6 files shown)


g-cheishvili (approved):
It works

InnaAtanasova (approved):

PR #9529, opened 2023-03-16
feat(platform): implement end column freezing

### LINKED ISSUE #9030: Is there a way I can freeze the last one or several columns of <fdp-table>?

mikerodonnell89:
fixes #9030 

Adds a new input `freezeToEndColumns` that allows the freezing of columns at the end of the platform table.


libs/i18n/src/lib/languages/bulgarian.ts
```diff
@@ -423,6 +423,7 @@ export const FD_LANGUAGE_BULGARIAN: FdLanguage = {
         headerMenuSortDesc: 'Сортирай Нисходящо',
         headerMenuGroup: 'Групирай',
         headerMenuFreeze: 'Замрази',
+        headerMenuEndFreeze: 'Freeze to End',
```
InnaAtanasova, line 426: should use the proper language

libs/i18n/src/lib/languages/chinese.ts
```diff
@@ -352,6 +352,7 @@ export const FD_LANGUAGE_CHINESE: FdLanguage = {
         headerMenuSortDesc: '降序排序',
         headerMenuGroup: '分组',
         headerMenuFreeze: '冻结',
+        headerMenuEndFreeze: 'Freeze to End',
```
InnaAtanasova, line 355: should use the proper language

InnaAtanasova (approved):

N1XUS (approved):

PR #9553, opened 2023-03-20
fix(core): various bugfixes

### LINKED ISSUE #9552: Bugs before release
- [x] Dialog without specified width or height always opens at 100% width and height:
image
image

- [x] Form message: example Message With InputGroup - Hover. If hover input field and clicking on addon button, form message will disappear, even though the hover is still in input, and when cursor is moved outside input element, form message is displayed. This behaviour repeats until user clicks outside input element;
- [ ] Form message: Formatted text does not have space before text: 
image

- [x] Illustrated message: inline svg does not shown if promise used to retrieve the svg:
image

- [ ] Inline help: If place cursor at a certain location, inline help starts appearing/disappearing very quickly: https://user-images.githubusercontent.com/6586561/226288994-1d86a9f9-b41f-4328-a5b6-cb5f27ac65d5.mov
- [ ] Inline help: some positions have transitions, some dont
- [x] Link with icon: Icon is not vertically centered: 
image

- [x] List: Unread indicator missing styling: 
image

- [ ] Multi Combobox: Selection ordered by item index. Should be ordered by user selection order;

- [ ] Rating indicator: Examples use native input elements. Should be using fundamentals controls instead;
- [x] https://fundamental-ngx.netlify.app/#/core/segmented-button#form - segmented button with reactive forms does not highlight initial selected value;
- [x] Select: List looks ugly
image

- [x] https://fundamental-ngx.netlify.app/#/core/switch - playground is missing semantic option;
- [x] Timeline: icon is overflowed by the vertical line: 
image

- [x] Complex dialog example: two scrollbars due to resize handle: 
image

N1XUS:
## Related Issue(s)

fixes #9552

## Description
- Fixed dialog 100% width and height;
- Fixed dialog double scrollbar when resizer is visible;
- Fixed illustrated message with async svg config;
- Fixed core list item and unread indicator;
- Fixed switch docs schema;
- Fixed timeline vertical line overflowing the icon;
- Fixed segmented button initial state and highlighting selected button;

### DIFF
--- a/libs/core/src/lib/link/link.component.scss
+++ b/libs/core/src/lib/link/link.component.scss
@@ -1,1 +1,5 @@
 @import 'fundamental-styles/dist/link';
+
+.fd-link__content--icon-line-height {
+    line-height: var(--sapContent_LineHeight);
+}
--- a/libs/core/src/lib/list/list-item/list-item.component.ts
+++ b/libs/core/src/lib/list/list-item/list-item.component.ts
@@ -10,6 +10,7 @@ import {
     forwardRef,
     HostBinding,
     HostListener,
+    inject,
     Inject,
     Input,
     OnChanges,
@@ -198,6 +199,11 @@ export class ListItemComponent
 
     /** @hidden */
     readonly _uniqueId = 'fd-list-item-' + ++listItemUniqueId;
+
+    /** @hidden */
+    readonly _list = inject(FD_LIST_UNREAD_INDICATOR, {
+        optional: true
+    });
 
     /** @hidden */
     constructor(
--- a/libs/cdk/src/lib/utils/functions/index.ts
+++ b/libs/cdk/src/lib/utils/functions/index.ts
@@ -10,3 +10,4 @@ export * from './is-compact-density';
 export * from './module-deprecations-provider';
 export * from './consumer-provider-factory';
 export * from './get-document-font-size';
+export * from './is-odd';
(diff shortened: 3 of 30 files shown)


InnaAtanasova (approved):

g-cheishvili (approved):

libs/core/src/lib/link/link.component.scss
```diff
@@ -1 +1,5 @@
 @import 'fundamental-styles/dist/link';
+
+.fd-link__content--icon-line-height {
+    line-height: var(--sapContent_LineHeight);
```
mikerodonnell89, line 4: do we need this in styles too?

libs/core/src/lib/list/list-item/list-item.component.ts
```diff
@@ -199,6 +200,11 @@ export class ListItemComponent
     /** @hidden */
     readonly _uniqueId = 'fd-list-item-' + ++listItemUniqueId;
 
+    /** @hidden */
+    readonly _list = inject(FD_LIST_UNREAD_INDICATOR, {
```
mikerodonnell89, line 204: Can you explain what's going on here? Just curious, I don't understand

mikerodonnell89 (approved):
Good stuff - couple questions above

libs/core/src/lib/list/list-item/list-item.component.ts
N1XUS: We inject list here to get it's byline state (https://github.com/SAP/fundamental-ngx/pull/9553/files#diff-dc5ece276511bdd17a8ed4a0d3b869f73564d6b6dc759e79d64df46e2edc09b1R8) since unread indicator is available only for the byline list

PR #9554, opened 2023-03-20
fix(core): dynamic collapsible subheader z-index must be higher

### LINKED ISSUE #9357: "collapse" and "pin" buttons get half hidden when clicked inside the fdp-dynamic-page-content

looks like a bug

leave the buttons fully visible

current (our project is using angular 15.1.0 and fundamental 0.38.2)

Open for example the https://sap.github.io/fundamental-ngx/#/platform/dynamic-page#non-collapsible example in Stackblitz and set the 'collapsible' and 'pinnable' to "true". Then click somewhere inside the 'fdp-dynamic-page-content' like this:

image

See that the "buttons" get half hidden.

**Note:** not sure (can't judge) if this was already reported in https://github.com/SAP/fundamental-ngx/issues/9152

cc @droshev (thanks!)

bpan4:
## Related Issue(s)

closes #9357 

## Description

Used inline styling to modify z-index of ```fd-dynamic-page__collapsible-header-visibility-container--button-group```. This keeps the pin, collapse, and expand buttons above ```fd-dynamic-page__content``` when other parts of the page are clicked.

### Before:
image

image

Buttons are hidden when ```fd-dynamic-page__content``` is clicked

### After:
image

image

Buttons remain visible when ```fd-dynamic-page__content``` is clicked


bpan4:
> I think the style should be in the css file, not inline

The styling is from fundamental-styles! It explicitly states the z-index of the subheader, and changing it would fix this issue. However, I'm not sure if it was defined for a specific implementation of the subheader or if it can be changed without repercussions.

InnaAtanasova (approved):

PR #9560, opened 2023-03-21
feat(platform): form generator inline help support for templates

### LINKED ISSUE #9382: inline help enhancements
enhancement

allow the form inline help to support template(not only simple text)

dxp `dxp/extension-manager/issues/456`

g-cheishvili:
## Related Issue(s)

closes #9382 

## Description

These changes bring inline help features for supporting templates via form generator configuration.

This update also deprecates passing templates explicitly to the inline helps(now one input takes care of both), and deprecates `hint.text` in favor of `hint.content` in hint config, adds support for passing the content to the `fd-label`, instead of title and template separately


libs/docs/platform/form-generator/examples/platform-form-generator-inline-help-example.component.html
```diff
@@ -13,10 +13,16 @@
...
+
 <fdp-form-generator
     [formItems]="formItemsHintOnGroupHeader"
     [hint]="formGeneratorMainHint"
     mainTitle="Form generator with hint"
 >
     <fdp-button type="submit" label="Submit form"></fdp-button>
 </fdp-form-generator>
+
+<ng-template #hintTemplate> Hint template! <fd-icon glyph="activate"></fd-icon> </ng-template>
```
N1XUS, line 28: Would it be difficult to also pass configuration of the current form item? It might be useful to render text from label or value

N1XUS (approved):

libs/docs/platform/form-generator/examples/platform-form-generator-inline-help-example.component.html
g-cheishvili: It is problematic because the template is for inline help and inline help does nothing to do with control in general. If someone requests it, we can do that as a separate feature

g-cheishvili: Although, idea is cool

PR #9590, opened 2023-03-27
fix(platform): exclude columns from p13n dialog

### LINKED ISSUE #9520: [Sourcing] Column with empty label appears on p13 dialog
enhancement
There is a use case when we have a column with empty label to perform multi-actions (represented as three dots, buttons, or symbol...). However, it is shown in a column panel with empty title like this:

Screenshot 2023-03-14 at 8 31 22 PM

We have [filterable]="false" or [sortable]="false" to make the column not visible on filter or sort panel. Can we have something similar to make that not visible in column panel? Or can we have some property to act as alt label in column panel?

Sample: https://angular-wtufa9.stackblitz.io

Angular 15.2.0
fundamental-ngx/core: 0.39.1-rc.4 or latest
fundamental-ngx/platform": 0.39.1-rc.4 or latest

Please refer: https://stackblitz.com/edit/angular-4hdtve?file=src%2Fapp%2Fplatform-table-p13-columns-example.component.html
UX member does not recommend to have a label for "action" type column, so we don't have workaround for this case.

fkolar:
Introducing new input property which removes column from the P13n list

## Related Issue(s)

closes #9520

## Description

Introduced new `disableP13n` property on the `TableColum` that excludes columns from P13n Dialog. By default all columns are included.

### DIFF
--- a/libs/docs/platform/table/child-docs/p13-dialog/p13-dialog-docs.component.html
+++ b/libs/docs/platform/table/child-docs/p13-dialog/p13-dialog-docs.component.html
@@ -16,7 +16,8 @@
 <description>
     To make columns settings available connect
     <code>&lt;fdp-table-p13-dialog [table]="table"&gt;&lt;/fdp-table-p13-dialog&gt;</code> to a table. Once there is at
-    least one <code>fdp-column</code> definition the columns settings button will be available in the table toolbar.
+    least one <code>fdp-column</code> definition the columns settings button will be available in the table toolbar. You
+    can also exclude columns from the p13n dialog using the <code>disableP13n</code> input property.
 </description>
 <component-example>
     <fdp-platform-table-p13-columns-example></fdp-platform-table-p13-columns-example>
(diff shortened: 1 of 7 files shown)


g-cheishvili (approved):
Please restart pipeline

fkolar:
Putting on hold this PR, to see if we cant find better solution.

InnaAtanasova (approved):

mikerodonnell89 (approved):

mikerodonnell89:
> Putting on hold this PR, to see if we cant find better solution.

What is it you don't like about this approach?

fkolar:
> > Putting on hold this PR, to see if we cant find better solution.
> 
> What is it you don't like about this approach?

Even this  simple, I wanted to see if we can somehow use the p13 column component but for now I think this is the best we can get and unblock everybody

mikerodonnell89:
> > > Putting on hold this PR, to see if we cant find better solution.
> > 
> > 
> > What is it you don't like about this approach?
> 
> Even this simple, I wanted to see if we can somehow use the p13 column component but for now I think this is the best we can get and unblock everybody

I agree, I think let's get this merged

N1XUS:
@fkolar please resolve the conflicts so we could merge your branch. Thank you

fkolar:
updated from main

PR #9594, opened 2023-03-27
feat(cdk): created breakpoint directive

### LINKED ISSUE #9361: [Reopen][Sourcing][Responsiveness]: fdp table toolbar is not responsive
enhancement

1) With 200% zoom , If there are many action buttons , then the buttons are overflowing out of container
~~2) With 400% zoom, even if there are only 2 action buttons with search the buttons are overflowing out of container~~
**we only need to account for 200% zoom**

Expected result: The buttons should wrap to next line in the toolbar container

angular 13 
core / platform: 0.35.x

Replicable on https://sap.github.io/fundamental-ngx/#/platform/table 

try adding few more actions buttons say about 4-5 buttons and try zooming to 200%
MicrosoftTeams-image (4)

https://sap.github.io/fundamental-ngx/#/platform/table 

This is crticial for DTAG customer for guided sourcing
priority B

g-cheishvili:
## Related Issue(s)

closes #9361 

## Description

This PR introduces the Breakpoint directive, which is for easing responsiveness-related tasks on the template side. Also added example which showcases the usage of it in case of the table action bar


libs/cdk/src/lib/utils/tokens/viewport-size.observable.ts
```diff
@@ -0,0 +1,20 @@
...
+ * from @angular/cdk/overlay.
+ */
+export const ViewportSizeObservable = new InjectionToken<Observable<number>>('ViewportSizeObservable', {
+    providedIn: 'root',
+    factory: () => {
+        const viewportRuler = inject(ViewportRuler);
+        return viewportRuler
+            .change(50)
+            .pipe(map((e) => (e.target as Window).innerWidth))
+            .pipe(startWith(window.innerWidth));
```
N1XUS, line 18: In most cases I see this directive should be used, width of the parent container is much more preferable rather than window width

N1XUS (approved):

mikerodonnell89 (dismissed):
I am not seeing the `<div *fdkBreakpointS>Will be visible on S</div>` el

Screenshot 2023-03-28 at 12 22 33 PM

libs/docs/cdk/breakpoint/breakpoint-header/breakpoint-header.component.html
```diff
@@ -0,0 +1,12 @@
+<header>Breakpoint</header>
+<description>
+    <p>
+        During the development of a web application or a component, we often need a logic for determining visibility of
```
mikerodonnell89, line 4: I think the docs need to give some more detail for when using the default values, i.e. pixel amounts, whether the breakpoints work as minimums or maximums, etc

g-cheishvili:
> I am not seeing the `<div *fdkBreakpointS>Will be visible on S</div>` el
> 
> Screenshot 2023-03-28 at 12 22 33 PM

oops, wrong example code

mikerodonnell89 (changes requested):
Seeing some odd behavior on certain resizes, I can occasionally set the screen to a certain size and I'll see this, which corrects itself when I click what should be the action buttons

responsivebuttons

g-cheishvili:
Fixed

libs/cdk/src/lib/utils/tokens/viewport-size.observable.ts
g-cheishvili: Added that @N1XUS

InnaAtanasova (approved):

mikerodonnell89 (approved):

PR #9604, opened 2023-03-29
fix(platform): enforce css specificity in Value Help Dialog label

InnaAtanasova:
## Description
enforce the specificity so that the css coming from the Platform is not dependant on the css files orger


mikerodonnell89 (changes requested):
This should take care of it: https://github.com/SAP/fundamental-ngx/pull/9605

PR #9612, opened 2023-03-31
fix(platform): tooltip for column size shrink

### LINKED ISSUE #9533: [Sourcing][Reopen] fdp-table : Tooltip for column size shrink with lengthy name
Bug

Upon column shrinking  resizing - if the column name is lengthy, a tooltip should be shown to the user.

angular 15
core / platform : 0.39.1-rc-5

This is replicable in https://sap.github.io/fundamental-ngx/#/platform/table#default

1) shrink the column
2) text is truncated with elipsis
3) hover on the column 
4) no tooltip displayed

This is  blocker for guided sourcing.

fkolar:
## Related Issue(s)

closes #9533

## Description

Adds tooltip for the column. Also adds ellipsis for the header text. 
I am not happy with the fix. It adds title for every column and I think what we want is to show the tooltip only if the ellipsis are shown.  Should we add it to the PR or another? 

Still not sure about the best solution for this. 

- After resize is finished lookup the header label element and check if the text is overflowing and only then add the title. 

```ts
// ellipsis visible? 
_isOverflowVisible(el: HTMLElement): boolean {      
        return el.scrollWidth > e.clientWidth
    }
```

Native solution would be just to add 

```html
<span  [attr.title]="_isOverflowVisible (,,) > column.label : null"
```

I think we should do this after resize is finished and then store this informationon the `TableColumn`

##### PR Quality

-   [X ] the commit message(s) follows the guideline:
-   [X ] all items on the PR Review Checklist are addressed :
-   [x] update `README.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,33 @@
+# [0.40.0-rc.12](https://github.com/SAP/fundamental-ngx/compare/v0.40.0-rc.11...v0.40.0-rc.12) (2023-04-02)
+
+
+
+# [0.40.0-rc.11](https://github.com/SAP/fundamental-ngx/compare/v0.40.0-rc.10...v0.40.0-rc.11) (2023-04-02)
+
+
+
+# [0.40.0-rc.10](https://github.com/SAP/fundamental-ngx/compare/v0.40.0-rc.9...v0.40.0-rc.10) (2023-04-02)
+
+
+### Bug Fixes
+
+* **platform:** no tooltip for breadcrumb when overflow dots are shown ([#9610](https://github.com/SAP/fundamental-ngx/issues/9610)) ([9bf5758](https://github.com/SAP/fundamental-ngx/commit/9bf5758155cc46c017654d99cc5e7bb49bf3372a))
+
+
+
+# [0.40.0-rc.9](https://github.com/SAP/fundamental-ngx/compare/v0.40.0-rc.8...v0.40.0-rc.9) (2023-04-02)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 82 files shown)


g-cheishvili:
Maybe we should add an option to disable truncation? Someone might want to wrap or something

fkolar:
> Maybe we should add an option to disable truncation? Someone might want to wrap or something

Good point and just leave default behavior.

g-cheishvili:
> > Maybe we should add an option to disable truncation? Someone might want to wrap or something
> 
> Good point. It is just by spec. header should [never wrap ](https://experience.sap.com/internal/fiori-design-web/grid-table/#column-headers-best-practices).
> 
> I think it does not make much sense to show tooltip for every column, but is it such problem? For performance reason, you need to do this check only once after you resize, which means:
> 
> * save this information to the tableColumn
> * extends `ResizeService` to perform this additional check
> * Extends the  `ResizeService` to access visibleColumns
> * Extends TableRef to also expose `ID`
> * etc..

Yeah, showing tooltip always is not appealing but from performance perspective it makes sense to have all the tooltips visible. @droshev what do you think?
As for the wrap behavior, I think since this is platform component, we can forget about the wrapping here

fkolar:
> After pulling this locally, I'm seeing neither ellipses nor tooltip
> Screenshot 2023-04-03 at 11 00 54 AM

@mikerodonnell89  But None of the Description are shorten on this picture, Tooltips is visible only for the columns that are smaller than text. Can you test on the API Docs locally and see what is the difference between your version and e..g default platform Table?

fkolar:
> Ok I see now, it is working well. Do we want to perform this check on page init as well?

I think it would be too much. Assuming, the hor. scrollbar will always appear so initially columns won't be shrinked. Unless you see some case or way where we can set some fixed width and columns will overlaps their heading labels. 

I was trying to simulate this, but there was always a scrollbar.

mikerodonnell89 (approved):

g-cheishvili (approved):

PR #9630, opened 2023-04-04
fix(platform): custom status for approval flow

### LINKED ISSUE #9518: [Sourcing][Dev blocker] <fdp-approval-flow> Unable to set custom status
feature request
Back then, for platform approval flow, we can freely set our own status for approval node. However, with current adoption, we notice that these statuses do not show anymore in the approval node:

Previous version:

Screenshot 2023-03-14 at 11 05 00 AM

Now:

Screenshot 2023-03-14 at 11 06 54 AM

Looking at https://github.com/SAP/fundamental-ngx/blob/fefac1929380f482a9786b09513779548281256c/libs/platform/src/lib/approval-flow/approval-flow-node/approval-flow-node.component.html, we see that we can only use 4 types of statuses now if it is not duedate compared to previous version where we can freely use our own status. This is a blocker for our Guided Sourcing project. Could you revert or provide any feature for us to use our own status like we had for previous version?

Angular 15.2.0
fundamental-ngx/core: 0.39.1-rc.4
fundamental-ngx/platform": 0.39.1-rc.4

Screenshot 2023-03-14 at 11 18 29 AM

fkolar:
## Related Issue(s)

closes #9518

## Description

Introduced custom status labels and custom color mapping for the approval flow. 

##### PR Quality

-   [X] the commit message(s) follows the guideline:
-   [X] tests for the changes that have been done
-   [X] all items on the PR Review Checklist are addressed :
-   [x] update `README.md`

### DIFF
--- a/libs/platform/src/lib/approval-flow/approval-flow-node/approval-flow-node.component.ts
+++ b/libs/platform/src/lib/approval-flow/approval-flow-node/approval-flow-node.component.ts
@@ -80,6 +80,14 @@ export class ApprovalFlowNodeComponent implements OnInit, OnChanges, OnDestroy {
      */
     @Input() allNodesInColumnApproved = false;
 
+    /** A reference to the app custom statuses */
+    @Input()
+    approvalStatusTemplate: TemplateRef<any>;
+
+    /** Custom status to color mapping  */
+    @Input()
+    statusColorMapping: Record<ApprovalStatus, ObjectStatus>;
+
     /** Whether the node is in edit mode */
     @Input()
     @HostBinding('class.fdp-approval-flow-node--edit-mode')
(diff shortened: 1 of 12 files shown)


libs/platform/src/lib/approval-flow/approval-flow-node/approval-flow-node.component.ts
```diff
@@ -80,6 +80,14 @@ export class ApprovalFlowNodeComponent implements OnInit, OnChanges, OnDestroy {
      */
     @Input() allNodesInColumnApproved = false;
 
+    /** A reference to the app custom statuses */
+    @Input()
+    approvalStatusTemplate: TemplateRef<any>;
```
mikerodonnell89, line 85: This works, but why not just accept string inputs? Would be much simpler

mikerodonnell89 (commented):
See my comment above

libs/platform/src/lib/approval-flow/approval-flow-node/approval-flow-node.component.ts
fkolar: @Mike O'Donnell, the 9630
>This works, but why not just accept string inputs? Would be much simpler
>Do you mean to pass an array? The reason for the TemplateRef is 

The idea behind this was

1.  you have this template there and you need to provide also internationalisation and  using a pipe it might be easier than just to maintain different versions of strings in your dataset, and I wanted to keep it in sync to have the same behaviour what we have inside

```html
  <ng-container [ngSwitch]="node.status">
            <ng-container *ngSwitchCase="'approved'">{{
                'platformApprovalFlow.statusApproved' | fdTranslate
            }}</ng-container>
            <ng-container *ngSwitchCase="'rejected'">{{
                'platformApprovalFlow.statusRejected' | fdTranslate
            }}</ng-container>
            <ng-container *ngSwitchCase="'in progress'">{{
                'platformApprovalFlow.statusInProgress' | fdTranslate
            }}</ng-container>
            <ng-container *ngSwitchCase="'not started'">{{
                'platformApprovalFlow.statusNotStarted' | fdTranslate
            }}</ng-container>
        </ng-container>

```
2. You want to map your STATUS -> ObjectStatus colors. 

3. You have a `ApprovalNode` instance, which is passed from `ApprovalFlowComponent` (when rendering every node), you want to pass this object to the app to decide how they want to render this status.

4. Extensibility. If they want to provide really custom behavior for the status, like they want to concatenate something or use icons maybe they can here. 


This way you are leaving it completly on the application and you dont need to add additional functionality into the `ApprovalFlowComponent`.

I can imagine you can pass something like 

```ts
  @Input
  approvalStatusMapping: Record< ApprovalStatus, strings>
```
And internally you would have to make sure when rendering a node, you check the strings + add additional piece of code that handles this mappings. The problem I see if you pass already translated string literals into it like `日期范围输入 | approved`, treating i18n data already as part of the app data instead of static strings I am not sure if this would not lead to problems. So I believe this is more saver and flexible. 

Let me know your thought...

g-cheishvili: This is much more flexible and robust, and I agree with Frank

g-cheishvili (approved):
I like these changes. Except for the fact that this component is deprecated, I do not see any other reason why we should not merge this

PR #9634, opened 2023-04-05
fix(core): attach popover to different container via append to

### LINKED ISSUE #9624: [Bug]: Core Popover appendTo input doesn't work

Bug

"appendTo" input parameter for the Fundamental Core Popover component doesn't work.  

Angular: 15.0.4
Fundamental: 38.0.0

It can be seen in the demo (https://sap.github.io/fundamental-ngx/#/core/popover#append-to). If you open the StackBlitz and make some changes, you will see that the component behaves the same if you pass or not pass the "appendTo" input.

You can also add a debugger and see that `appendTo` is always undefined in the popover service.

The appendTo is not assigned at any moment to the service.

I will open a PR with a proposal to fix it

aflalves:
## Related Issue(s)

closes #9624

## Description

The `appendTo` is never assigned to the service scope, so it is always `undefined` when the `_getPositionStrategy` checks for it, meaning it will always attach it to the trigger element.

### Before:

image

There is some padding around the trigger button which makes it a bit confusing to noticed. The screenshot below is without the padding and without the `display: block`:

image

### After:

image

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [x] poor examples

4. Accessibility testing

##### PR Quality

-   [x] update `README.md`

### DIFF
--- a/libs/core/src/lib/popover/popover-service/popover.service.ts
+++ b/libs/core/src/lib/popover/popover-service/popover.service.ts
@@ -208,6 +208,7 @@ export class PopoverService extends BasePopoverClass {
      */
     refreshConfiguration(config: BasePopoverClass): void {
         const onlyChanged = Object.keys(new BasePopoverClass()).filter((key) => this[key] !== config[key]);
+        this.appendTo = config.appendTo;
 
         if (onlyChanged.includes('isOpen')) {
             if (config.isOpen) {
--- a/libs/core/src/lib/popover/popover-service/popover.service.ts
+++ b/libs/core/src/lib/popover/popover-service/popover.service.ts
@@ -208,7 +208,6 @@ export class PopoverService extends BasePopoverClass {
      */
     refreshConfiguration(config: BasePopoverClass): void {
         const onlyChanged = Object.keys(new BasePopoverClass()).filter((key) => this[key] !== config[key]);
-        this.appendTo = config.appendTo;
 
         if (onlyChanged.includes('isOpen')) {
             if (config.isOpen) {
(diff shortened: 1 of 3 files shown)


libs/core/src/lib/popover/popover-service/popover.service.ts
```diff
@@ -208,6 +208,7 @@ export class PopoverService extends BasePopoverClass {
      */
     refreshConfiguration(config: BasePopoverClass): void {
         const onlyChanged = Object.keys(new BasePopoverClass()).filter((key) => this[key] !== config[key]);
+        this.appendTo = config.appendTo;
```
g-cheishvili, line 211: New BasePopoverClass.appendTo will be new input, if new input is different from the one which is in PopoverService. Then on line 225, of this file, that different is going to be assigned to `this`, So I do not see any reason for this specific explicit assignment

g-cheishvili (changes requested):

libs/core/src/lib/popover/popover-service/popover.service.ts
aflalves: Exactly, but it is not getting assigned on line 225 which causes the bug.

So I went deeper and found out that when we do `new BasePopoverClass()`, the `appendTo` key is not there, because it is `undefined`. We can see in the screenshot below that `config.appendTo` exists but `basePopoverClass` doesn't have this key, so its never added to the `onlyChanged`:

image

Instead of having the explicit assignment, we could instead assign `null` as a default value to the `BasePopoverClass.appendTo`, so the key will be there. I've tested and it worked as expected, so maybe we could go with this approach instead.

g-cheishvili: That's better, yes

aflalves: cool, its done :)

mikerodonnell89 (approved):

g-cheishvili (approved):

PR #9636, opened 2023-04-05
wip: tweaks to list screenreader behavior

### LINKED ISSUE #9556: [Sourcing][Accessibility] fd-list accessibility issues
Bug

1) Redundant word "group" is being read when traversing through fd-list items
2) Also checkbox checked state is not read by screenreader 
~~instead "?" is read~~ [fixed](https://github.com/SAP/fundamental-styles/pull/4518), needs fundamental-styles version update
3) please make sure the word "list item" is also read for every li in ul

Angular 15
core / platform: 0.39.1-rc-5

This is replicable on https://sap.github.io/fundamental-ngx/#/core/list#selection 
Screenshot 2023-03-21 at 2 48 27 PM

mikerodonnell89:
fixes #9556 

Some improvements to list accessibility. More verbose on VoiceOver but some information read twice on Chrome screenreader

### DIFF
--- a/libs/core/src/lib/list/list-item/list-item.component.html
+++ b/libs/core/src/lib/list/list-item/list-item.component.html
@@ -1,5 +1,5 @@
 <ng-container *ngIf="checkbox || radio">
-    <div fd-form-item class="fd-list__form-item">
+    <div fd-form-item class="fd-list__form-item" aria-hidden="true">
         <ng-content select="fd-checkbox"></ng-content>
         <ng-content select="fd-radio-button"></ng-content>
     </div>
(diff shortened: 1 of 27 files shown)


libs/core/src/lib/list/list-item/list-item.component.html
```diff
@@ -1,5 +1,5 @@
 <ng-container *ngIf="checkbox || radio">
-    <div fd-form-item class="fd-list__form-item">
+    <div fd-form-item class="fd-list__form-item" aria-hidden="true">
```
InnaAtanasova, line 2: If you add aria-hidden true to an element, the screen reader won't be able to read the element and it's children. Is it an expected behaviour to hide the checkbox and the radio from the SR?

Screenshot 2023-04-12 at 6 03 58 PM

InnaAtanasova (changes requested):

Issue #9637, opened 2023-04-05
[Sourcing] Unable to scroll to ok /cancel buttons when choosing time


ogudekar:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
Scrollbars appear on Date-Time picker when you scroll up-down outside the datepicker popup. Ok and Cancel button hide behind the scrollbar and can not be clicked

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
"@fundamental-ngx/core": "0.39.1-rc.20",
"@fundamental-ngx/platform": "0.39.1-rc.20"
"@angular/core": "15.2.0"

#### If this is a bug, please provide steps for reproducing it.
Open datetime picker. Click on minute. Scroll up and down outside.

https://user-images.githubusercontent.com/130004670/230230616-5e350779-8bce-44f5-bc51-d7607921f0da.mp4

oscargong:
This issue seems to occur only when the "Content Density" is set to "Cozy". It works fine with "Condensed" and "Compact" settings.

I543348:
This is a blocker for our current release

droshev:
Hello, we would need some more steps how to reproduce it. I tried the same link and i didn't see scrollbars
<img width="961" alt="Screenshot 2023-04-06 at 9 49 22 AM" src="https://user-images.githubusercontent.com/4380815/230398127-e220fb6c-0a21-41a4-8036-3616fd26110f.png">

oscargong:
> Hello, we would need some more steps how to reproduce it. I tried the same link and i didn't see scrollbars
> 
> <img width="961" alt="Screenshot 2023-04-06 at 9 49 22 AM" src="https://user-images.githubusercontent.com/4380815/230398127-e220fb6c-0a21-41a4-8036-3616fd26110f.png">
> 
> 

Hi Deno, need to scroll down until the DatePicker is out of view, the scroll back.

PR #9641, opened 2023-04-07
fix: temp fix for popover scrolling issues

mikerodonnell89:
temporary fix for #9637 

fd-scrollbar seems to be conflicting with the scroll position strategy logic from angular cdk, often making popovers unscrollable. This is a temporary workaround


InnaAtanasova (approved):

g-cheishvili:
This leads to something else. I can not recall to what, but there will be another issue if you remove fd-scrollbar 😆

PR #9657, opened 2023-04-10
fix(platform): only scroll freezable cells when using l/r keyboard buttons, freezable column width not updating on table data load

### LINKED ISSUE #9642: [Sourcing][Dev blocker] <fdp-table> Some issues with freeze columns
bug
There are two issues:
- Issue 1: if we have a inline link (or even button) in a table with scrollbar, first click always does not trigger and set the scrollbar to the left. After a few clicks on the same link, it is fine, but it will be replicated in another link:

https://user-images.githubusercontent.com/22842279/230691629-341759ee-471b-4385-88aa-98a5d20f50d9.mov

In more complex scenario, when we have inline link to open a popover (if successfully), any event triggered to close popover will also give us the same behavior like above (set scrollbar to left).

It is working fine with table without freeze column.

- Issue 2: even though we set width for every column, depending on window size, when we try to for example update table from empty state, it will create a white space in freeze column position until we hover the cursor somewhere on the table. In this case, setting fixed width for parent element will solve the issue, but we believe this has a huge limitation. Can you fix it?

https://user-images.githubusercontent.com/22842279/230692034-eb7e086c-27cb-47ec-9796-cf0abfd0ef27.mov

In our application, we sometimes is switching from empty to having data in table by adding/deleting item, and we could not set fixed width for parent element.

Angular 15.2.0
"@fundamental-ngx/core": "0.39.1-rc.26",
"@fundamental-ngx/platform": "0.39.1-rc.26"

Please refer the link below:

https://stackblitz.com/edit/angular-udar4q?file=src%2Fapp%2Fplatform-table-freezable-example.component.html,src%2Fapp%2Fplatform-table-freezable-example.component.ts,src%2Fapp%2Fapp.module.ts

mikerodonnell89:
fixes #9642

### DIFF
--- a/libs/platform/src/lib/table/table.component.html
+++ b/libs/platform/src/lib/table/table.component.html
@@ -108,6 +108,8 @@
                             [class.fd-table__item--clickable]="_isColumnHasHeaderMenu(column)"
                             [cellFocusedEventAnnouncer]="null"
                             (cellFocused)="_onCellFocused($event, column.label, null)"
+                            (keydown.arrowLeft)="_scrollToOverlappedCell()"
+                            (keydown.arrowRight)="_scrollToOverlappedCell()"
                         >
                             <ng-container
                                 *ngTemplateOutlet="tableHeaderCellTemplate; context: { column: column }"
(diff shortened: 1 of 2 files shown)


N1XUS (approved):
@mikerodonnell89 I was thinking that platform component class has a lot of helper methods in it. Maybe we can extract some of them into a table service? Like this one: `_scrollToOverlappedCell ` can be done as a method that returns scroll coordinates, and in component we will have just a one-two liner method

PR #9672, opened 2023-04-12
Check state for tree table

### LINKED ISSUE #9355: [Sourcing]: fdp tree table on check of parent checkbox , child checkboxes should also be checked
Bug

On check of parent row's checkbox , the child rows should also automatically get checked until and unless user unchecks few of children rows

angular 13
core / platform : 0.35.x

reproducible on https://sap.github.io/fundamental-ngx/#/platform/table#tree 

https://sap.github.io/fundamental-ngx/#/platform/table#tree

This is a development blocker for guided sourcing 
priority is A
Expecting a property from fdp-table which can be used to decide whether on check of "parent" row child rows should be checked or not - may be a boolean field.

fkolar:
## Related Issue(s)

closes #9355

## Description

There are for main changes: 
1) Expose TableRow as interface
2) Update init logic so it can work both with custom object as well as TableRow so app can work with Table row already on the level of DataProvider
3) To be able to propagate the selection to parents and child new properly is introduced eneableTristateMode. Default behavior is false, so the table works as before
4) Updated docs and tests

##### PR Quality

-   [X ] the commit message(s) follows the guideline:
-   [X ] tests for the changes that have been done
-   [ X] all items on the PR Review Checklist are addressed :
-   [ ] update `README.md`


mikerodonnell89 (changes requested):
This is looking great. I did notice this when toggling a tree parent node. The "added" array in the vent displays the "laptops" object three times rather than laptops once with the two children
Screenshot 2023-04-18 at 12 19 31 PM

fkolar:
> This is looking great. I did notice this when toggling a tree parent node. The "added" array in the event displays the "laptops" object three times rather than laptops once with the two children
> Screenshot 2023-04-18 at 12 19 31 PM

@mikerodonnell89 , This is good catch and great question. What do you guys think how it should work ? :

* When checking a parent node and **tristatemode = false,** then maybe it should emit only 1 node
* When checking a parent node and **tristatemode = TRUE,**, maybe it should emit all the checked nodes that are marked implicitly 

Or do we expect => I literally click on 1 node, so it should emit 1 node?

N1XUS:
Hello @fkolar. One more thing here. `children` array of the row is coming from the DataSource itself, so it may confuse developers who would use this functionality. So in my opinion, we should go with the second option where we include any changed row checkbox state

mikerodonnell89:
@fkolar I think this way you described is best

> When checking a parent node and tristatemode = false, then maybe it should emit only 1 node
When checking a parent node and tristatemode = TRUE,, maybe it should emit all the checked nodes that are marked implicitly

fkolar:
> Hello @fkolar. One more thing here. `children` array of the row is coming from the DataSource itself, so it may confuse developers who would use this functionality. So in my opinion, we should go with the second option where we include any changed row checkbox state

The whole idea is that on the app level you want to work with the tree not a flatten list, and I also we should include everything. But again only if TRISTATE is ON. actually when its false it is going to work as before  -> you can one item you emit one item

mikerodonnell89:
> > Hello @fkolar. One more thing here. `children` array of the row is coming from the DataSource itself, so it may confuse developers who would use this functionality. So in my opinion, we should go with the second option where we include any changed row checkbox state
> 
> So I guess it can be as it is now. Right @mikerodonnell89
> 
> Now it it emits all changed as you screenshots shows, when tristate is false, then it will work as before, when whatever you check (one item) it emits one item.

I mean, if tristate is enabled and I check "Laptops", I think the "added" array should display "Laptops", "Astro Laptops 1516", and "Benda Laptop 1408", rather than "Laptops" 3 times, no?

fkolar:
> > > Hello @fkolar. One more thing here. `children` array of the row is coming from the DataSource itself, so it may confuse developers who would use this functionality. So in my opinion, we should go with the second option where we include any changed row checkbox state
> > 
> > 
> > So I guess it can be as it is now. Right @mikerodonnell89
> > Now it it emits all changed as you screenshots shows, when tristate is false, then it will work as before, when whatever you check (one item) it emits one item.
> 
> I mean, if tristate is enabled and I check "Laptops", I think the "added" array should display "Laptops", "Astro Laptops 1516", and "Benda Laptop 1408", rather than "Laptops" 3 times, no?

Let me check..

mikerodonnell89 (commented):
Ok I'm liking the added/removed/etc. properties on the selection event now. But I am confused about what's going on with this "index" array - I thought I understood it before, it was the index of the changed item (or items), but now I'm struggling to make sense of what's going on here. Can you explain? Screenshot attached:
Screenshot 2023-04-24 at 12 30 47 PM

fkolar:
> Ok I'm liking the added/removed/etc. properties on the selection event now. But I am confused about what's going on with this "index" array - I thought I understood it before, it was the index of the changed item (or items), but now I'm struggling to make sense of what's going on here. Can you explain? Screenshot attached:

It seems this is existing behavior. As you can see, when I select first child the index is ZERO, reflecting first item in the tree. And when we select all child it simply does the same what was set before.  If we want to change this I would address this in different PR. 

Screenshot 2023-04-26 at 8 15 19

mikerodonnell89 (approved):

fkolar:
@N1XUS , @mikerodonnell89  - can be merge?

PR #9713, opened 2023-04-18
fix(platform): minor table visual issues

mikerodonnell89:
fixes none

Change detection issue with the the expanded/collapsed icon not updating and the `[class.fd-table__row--draggable]="enableRowReordering && _rowsDraggable"` addition fixes the mouse cursor on draggable child items

### DIFF
--- a/libs/platform/src/lib/table/components/table-cell-content/table-cell-content.component.ts
+++ b/libs/platform/src/lib/table/components/table-cell-content/table-cell-content.component.ts
@@ -1,12 +1,11 @@
-import { ChangeDetectionStrategy, Component, Input, ViewEncapsulation } from '@angular/core';
+import { Component, Input, ViewEncapsulation } from '@angular/core';
 import { TableRow } from '../../models/table-row.model';
 import { TableColumn } from '../table-column/table-column';
 
 @Component({
     selector: 'fdp-table-cell-content',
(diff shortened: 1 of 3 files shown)


libs/platform/src/lib/table/components/table-cell-content/table-cell-content.component.ts
```diff
@@ -1,12 +1,11 @@
-import { ChangeDetectionStrategy, Component, Input, ViewEncapsulation } from '@angular/core';
+import { Component, Input, ViewEncapsulation } from '@angular/core';
 import { TableRow } from '../../models/table-row.model';
 import { TableColumn } from '../table-column/table-column';
 
 @Component({
     selector: 'fdp-table-cell-content',
     templateUrl: './table-cell-content.component.html',
-    encapsulation: ViewEncapsulation.None,
-    changeDetection: ChangeDetectionStrategy.OnPush
```
N1XUS, line 9: Default change detection may have performance impact, and since this is a cell component, it's very possible that the table will have a lots of these component instances. I would suggest keeping the OnPush strategy, and try using ngOnChanges to notify the component that important data has been changed

N1XUS (changes requested):

InnaAtanasova (approved):

PR #9757, opened 2023-04-26
fix(platform): include hidden columns also with the column selection

### LINKED ISSUE #9520: [Sourcing] Column with empty label appears on p13 dialog

fkolar:
## Related Issue(s)

closes #9520

## Description

This is additional fix to existing issue where we did not include hidden columns with the final selection. Updated `getTableColumns` to return conditional either all disableP13n = true or false, so we can easily merge column from dialog. 

##### PR Quality

### DIFF
--- a/libs/platform/src/lib/table/components/table-p13-dialog/table-p13-dialog.component.ts
+++ b/libs/platform/src/lib/table/components/table-p13-dialog/table-p13-dialog.component.ts
@@ -224,7 +224,8 @@ export class TableP13DialogComponent implements OnDestroy {
             dialogRef.afterClosed
                 .pipe(filter((result) => !!result))
                 .subscribe(({ visibleColumns: result }: ColumnsDialogResultData) => {
-                    this._applyColumns(result);
+                    const allColumns = this._mergeWithHiddenColumns(result);
+                    this._applyColumns(allColumns);
                 })
         );
     }
@@ -277,9 +278,9 @@ export class TableP13DialogComponent implements OnDestroy {
     }
 
     /** @hidden */
-    private _getTableColumns(): TableColumn[] {
+    private _getTableColumns(revertSelection = false): TableColumn[] {
         const cols = this._table?.getTableColumns() || [];
-        return cols.filter(({ disableP13n }) => !disableP13n);
+        return cols.filter(({ disableP13n }) => disableP13n === revertSelection);
     }
 
     /** @hidden */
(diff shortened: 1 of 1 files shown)


mikerodonnell89 (approved):
good catch

InnaAtanasova (approved):

fkolar:
@mikerodonnell89 @InnaAtanasova , can we merge? Or @droshev  can you give me merging privileges for the things that are approved

mikerodonnell89 (changes requested):
e2e tests are failing, looks relevant this time

g-cheishvili (approved):

fkolar:
I will check.

I543348:
> @mikerodonnell89: I need to wait for @Anusha, as its not just e2e, but we opened up a box with more problems. So before fixing this we need to clarify a few things e.g.:
> 
> Columns: ` [Name, Description, Price, Status, Action ]`
> 
> * Mark **Action** column as **disableP13n**,  and open up a p13n dialog and you want to move Column **Name** all the way to bottom (last column)
>   
>   * What should happen ? Because by putting the disableP13n column back, it will placed as last  as it originally was.  We have no way to understand that column **Name** needs to be really last column
> * Mark **Price** as **disableP13n**, the same as above ? you move any columns before or after the **Price**
>   
>   * How do you want to maintained the original position of this **Price** columns?
> 
> I don't think there is a solution, besides rolling back the change.

Thanks @fkolar , I think the columns positions should not be changed. please let me know if there is any clarifications required from my end. Thanks.

fkolar:
Closing PR, as we are going to roll back this requirements to its original state.

Issue #9246, opened 2023-01-19
[Accessibility] [Sourcing]: fdp-table infinite scroll accessibility issue


I543348:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Briefly describe your proposal.
The infinite scroll table example is not supported with keyboard.
User is not able to traverse through table cells instead only scroll works with up / down arrows keys

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
Angular 13
core / platform : 0.35.x

#### If this is a bug, please provide steps for reproducing it.
1) Reach the table using keyboard
2) The focus comes to the table body 
3) Now try to reach the table column cell or row cell
4) There is no way to reach any row using keyboard

#### Please provide relevant source code if applicable.
https://sap.github.io/fundamental-ngx/#/platform/table
#### Is there anything else we should know?
This is a critical issue for guided sourcing customers

droshev:
in the example [here](https://fundamental-ngx.netlify.app/#/platform/table#page-scrolling) we offer the scrolling. the only missing part is preserving the focus.

fkolar:
@skaquib123, have you checked?

skaquib123:
Hey @fkolar, yes I have checked this. There was a issue in this. The  guidelines they are following for keyboard accessibility does not apply to this particular tree table. It had a fixed header so thats what created the accessibility problem. I even took it up to Mehfooz, Havva of our accessibility team. I guess they have now changed the fixed header issue. So this should be working now.

@I543348  can you please confirm weather it is working as per expectation?

droshev:
@I543348 any update?

I543348:
The issue is still replicable on NGX https://sap.github.io/fundamental-ngx/#/platform/table/basic#freezable
Steps to replicate :
1) traverse from search box to header checkbox and try to press left arrow key  - doesnt work instead frozen columns get scrolled left / right
2) traverse from search box to header checkbox - Press down arrow to move into the table rows - now reach header - This works now .
Point 1 should also work like point 2. please refer recording for both the cases.

https://user-images.githubusercontent.com/88182358/233543558-92a5a99a-b44d-4396-a2cb-7b32e00ee721.mov

PR #9762, opened 2023-04-27
fix(platform): checkbox focus/keyboard nav issue in platform table

mikerodonnell89:
fixes #9146 not the original issue but the one described here: https://github.com/SAP/fundamental-ngx/issues/9246#issuecomment-1517255240


libs/platform/src/lib/table/table.component.html
```diff
@@ -260,13 +262,13 @@
                                             | async
                                             | selectionCellStyles : _rtl : _semanticHighlightingColumnWidth
                                     "
+                                    (click)="_toggleMultiSelectRow(row, $event)"
+                                    (keydown.enter)="_toggleMultiSelectRow(row, $event)"
+                                    (keydown.space)="_toggleMultiSelectRow(row, $event)"
```
g-cheishvili, line 267: We have a directive called `fdkClick` which does exactly this, maybe it is worth using that instead?

InnaAtanasova (approved):

PR #9781, opened 2023-04-30
feat(core): tree component

N1XUS:
## Related Issue(s)

closes none

## Description
Added tree component implementation;
Warning - fundamental-styles package version is updated, may affect other components.

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [x] poor examples

4. Accessibility testing


g-cheishvili:
Screenshot 2023-05-08 at 15 33 32
Item focus outline overlaps with indicator color and it is hard to digest which color is in indicator

When you click on a expand button, for a split second outline on button appears and then focus if moved to the item. I think button should not have focus outlines, otherwise it creates this awkward moment

InnaAtanasova:
8. Once a tree item is focussed, how do you expand/collapse the item (which key)?
9. 
```
<button tabindex="-1" aria-haspopup="true" class="fd-tree__expander ng-star-inserted" aria-expanded="false"><fd-icon role="presentation" class="sap-icon--navigation-right-arrow"></fd-icon></button>
``` 
should be a span with role="presentation" and aria-hidden="true". Also some label that should change to "expand/collapse"

10. Apart from aria-level, List Item should have aria-expanded="true/false", [aria-setsize](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-setsize), [aria-posinset](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-posinset). 
11. The icons in the List item should have either role="presentation" and aria-hidden="true"

libs/core/src/lib/tree/components/tree-item/tree-item.component.html
```diff
@@ -0,0 +1,86 @@
...
+            class="fd-tree__expander"
+            [class.is-expanded]="expanded"
+            aria-haspopup="true"
+            [attr.aria-expanded]="expanded"
+        >
+            <fd-icon [glyph]="expanded ? 'navigation-down-arrow' : 'navigation-right-arrow'"></fd-icon>
+        </button>
+        <div class="fd-tree__expander fd-tree__expander--loading" *ngIf="!_childrenLoaded">
+            <fd-skeleton type="rectangle" width="1rem" height="1rem"></fd-skeleton>
+        </div>
```
InnaAtanasova, line 34: why only the expander has a skeleton? Shouldn't the whole item have it?

InnaAtanasova:
You need RTL service for the expand/collapse icon
Screenshot 2023-05-08 at 5 34 56 PM

InnaAtanasova:
Great work with the code! 🥇

libs/core/src/lib/tree/components/tree-item/tree-item.component.html
N1XUS: @InnaAtanasova Item skeleton is located here: https://github.com/SAP/fundamental-ngx/pull/9781/files#diff-6527bb81695f8c236de4e9413a47f48fd6eddda06a95fbe0927743b91f5c71e9R7 since it rendered before we initiate tree item

N1XUS:
> 8. Once a tree item is focussed, how do you expand/collapse the item (which key)? (figured this out :D)
> 
> ```
> <button tabindex="-1" aria-haspopup="true" class="fd-tree__expander ng-star-inserted" aria-expanded="false"><fd-icon role="presentation" class="sap-icon--navigation-right-arrow"></fd-icon></button>
> ```
> 
> should be a span with role="presentation" and aria-hidden="true". Also some label that should change to "expand/collapse"
> 
> 10. Apart from aria-level, List Item should have aria-expanded="true/false", [aria-setsize](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-setsize), [aria-posinset](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-posinset).
> 11. The icons in the List item should have either role="presentation" and aria-hidden="true"

for the 8:
image

N1XUS:
@InnaAtanasova as to number 5: it's navigatable, but it does not have navigation indicator. Those two properties are separate

InnaAtanasova:
> @InnaAtanasova as to number 5: it's navigatable, but it does not have navigation indicator. Those two properties are separate

I know they are 2 separate concepts, but are used in a wrong way in this case. 
If you consider the List specs, you have "Navigation List", where all items are links and there are no navigation indicators as the entire list is navigatable. 
You have navigation indicator when the list is mixed, consists of items with links and items with no links. In this case, the items without the indicator are not links. 
So you can't have 2 items that are links and one is with indicator.

libs/core/src/lib/tree/components/tree-item/tree-item.component.html
```diff
@@ -0,0 +1,95 @@
+<div
+    class="fd-tree__item-container"
+    [attr.tabindex]="_containerTabIndex"
+    (click)="_onClick($event)"
+    (keyup.enter)="_onClick($any($event))"
+    (keyup.space)="_onClick($any($event))"
```
g-cheishvili, line 6:
```suggestion
    (fdkClicked)="_onClick($event)"
```

libs/core/src/lib/tree/components/tree-item/tree-item.component.html
```diff
@@ -0,0 +1,95 @@
...
+            [(ngModel)]="_selectionState"
+            [ngModelOptions]="{ standalone: true }"
+            [selectedValue]="_selectionState"
+            [value]="true"
+            tabIndex="-1"
+            [standalone]="true"
+            (click)="_onClick($event)"
+            (keyup.enter)="_onClick($any($event))"
+            (keyup.space)="_onClick($any($event))"
+            tabIndexValue="-1"
```
g-cheishvili, line 91: Where is this coming from?

libs/core/src/lib/tree/components/tree-item/tree-item.component.html
```diff
@@ -0,0 +1,95 @@
...
+        <fd-radio-button
+            [(ngModel)]="_selectionState"
+            [ngModelOptions]="{ standalone: true }"
+            [selectedValue]="_selectionState"
+            [value]="true"
+            tabIndex="-1"
+            [standalone]="true"
+            (click)="_onClick($event)"
+            (keyup.enter)="_onClick($any($event))"
+            (keyup.space)="_onClick($any($event))"
```
g-cheishvili, line 90: If it is not focusable, can keyboard events come from it?

libs/core/src/lib/tree/components/tree-item/tree-item.component.html
N1XUS: @g-cheishvili fdkClicked does not prevent default, thus, space scrolls the page

libs/core/src/lib/tree/components/tree-item/tree-item.component.html
N1XUS: Nah, forgot to remove when added tabindex

libs/core/src/lib/tree/components/tree-item/tree-item.component.html
mikerodonnell89: Could we have fdkClicked catch events and provide options for preventDefault/stopPropagation/stopImmediatePropagation?

PR #9818, opened 2023-05-10
fix(core,platform,cx): navigation components had missing/incorrect roles

### LINKED ISSUE #9608: [Sourcing][Accessibility]:  fd-vertical-navigation accessibility issues
Bug

1) when focus is on "deliveries" the screenreader doesnt read UI control type "list item"
2) even for nested list items , the UI control type is not read by screenreader
3) when the parent node is expanded and the focus is on the parent node, the focus is appearing for entire list , rather it should be moved to parent node and the levels have to be read out by screen reader

Angular 15
core / platform: 0.39.2

Please find the screenshots for issue 1, 2 and 3 

Issue 1
Screenshot 2023-03-30 at 5 00 07 PM

Issue 2
Screenshot 2023-03-30 at 5 00 29 PM

Issue 3
Screenshot 2023-03-30 at 5 00 38 PM

mikerodonnell89:
fixes #9608


InnaAtanasova (approved):

libs/i18n/src/lib/languages/georgian.ts
```diff
@@ -42,6 +42,10 @@ export const FD_LANGUAGE_GEORGIAN: FdLanguage = {
     coreMessageStrip: {
         dismissLabel: 'გაუქმება'
     },
+    coreNavigation: {
+        mainNavigation: 'Main Navigation',
+        navigationPath: 'Navigation Path'
+    },
```
g-cheishvili, line 48:
```suggestion
    coreNavigation: {
        mainNavigation: 'მთავარი ნავიგაცია',
        navigationPath: 'ნავიგაციის გზა'
    },
```

libs/i18n/src/lib/languages/russian.ts
```diff
@@ -56,6 +56,10 @@ export const FD_LANGUAGE_RUSSIAN: FdLanguage = {
     coreMessageStrip: {
         dismissLabel: 'Закрыть'
     },
+    coreNavigation: {
+        mainNavigation: 'Main Navigation',
+        navigationPath: 'Navigation Path'
+    },
```
g-cheishvili, line 62:
```suggestion
    coreNavigation: {
        mainNavigation: 'Главная Навигация',
        navigationPath: 'Путь навигации'
    },
```

libs/core/src/lib/list/list-navigation-item/list-navigation-item.component.ts
```diff
@@ -99,11 +117,18 @@ export class ListNavigationItemComponent implements AfterContentInit, AfterViewI
             this._iconComponent._navigationItemIcon = true;
         }
         this._innerText = this._text.elementRef.nativeElement.textContent ?? '';
+        this._ariaLevel = 1;
+        this._childItems?.forEach((item) => {
+            item._ariaLevel = 2;
+        });
```
g-cheishvili, line 123: If there is a child inside the navigation item, will it be ariaLevel 1 always? What I am saying is that maybe ariaLevel can be by default 0 and in afterContentInit somehow we could always increase it by 1?

g-cheishvili: Or it can inject parent `ListNavigationItemComponent` and then increase it's value by 1 and set that value for itself?

PR #9876, opened 2023-05-16
feat(core): added select all toggler to the multi combo box

### LINKED ISSUE #9711: Select all function for multi combo box
**Feature Request**
The Fiori Guidelines describe there's the option to have a "Select All" checkbox at top of the multi combo box option list:
https://experience.sap.com/fiori-design-web/multi-combobox/#select-a-value

Quote: "If selecting all items is a common use case, you can also show a Select All [checkbox]." (https://experience.sap.com/fiori-design-web/checkbox/) at the top of the list (property: showSelectAll)

Styles and NGX.

Please provided an example for Select All in Styles and NGX library.

g-cheishvili:
## Related Issue(s)

closes #9711

## Description

- Added input for "select all"
- Fixed visual detection for selection

### DIFF
--- a/libs/core/src/lib/multi-combobox/select-all-toggler.component.ts
+++ b/libs/core/src/lib/multi-combobox/select-all-toggler.component.ts
@@ -23,7 +23,7 @@ import { Observable, takeUntil } from 'rxjs';
     `,
     styles: [
         `
-            fd-multi-combobox-select-all-toggler.fd-list__item {
+            li[fd-multi-combobox-select-all-toggler].fd-list__item {
                 display: block;
                 background-color: var(--sapBackgroundColor);
             }
--- a/libs/core/src/lib/multi-combobox/select-all-toggler.component.ts
+++ b/libs/core/src/lib/multi-combobox/select-all-toggler.component.ts
@@ -11,7 +11,8 @@ import { DestroyedService } from '@fundamental-ngx/cdk/utils';
 import { Observable, takeUntil } from 'rxjs';
 
 @Component({
-    selector: 'fd-multi-combobox-select-all-toggler',
+    // eslint-disable-next-line @angular-eslint/component-selector
+    selector: 'li[fd-multi-combobox-select-all-toggler]',
     template: `
         <fd-checkbox
             [label]="'coreMultiComboBox.selectAllLabel' | fdTranslate"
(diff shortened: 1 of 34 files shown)


mikerodonnell89 (changes requested):
I think that issue with the selection in the popover needs to be fixed as part of this pr

libs/core/src/lib/multi-combobox/select-all-toggler.component.ts
```diff
@@ -0,0 +1,72 @@
...
+        <fd-checkbox
+            [label]="'Select All'"
+            [ngModel]="checkboxValue"
+            (ngModelChange)="change($event)"
+            [tristate]="true"
+        ></fd-checkbox>
+    `,
+    styles: [
+        `
+            fd-multi-combobox-select-all-toggler.fd-list__item {
```
mikerodonnell89, line 18: Should be a fundamental-styles thing too I think?

g-cheishvili: I could not find anything in wiki about this, only at experience sap, so I took and did it only here because of that

g-cheishvili:
> I think that issue with the selection in the popover needs to be fixed as part of this pr

Ok, fixed

mikerodonnell89 (changes requested):
Can it be an `<li>` and a direct child of the `ul` like the other items in the popover? Because with this current template, we are seeing this border radius issue
Screenshot 2023-05-17 at 10 52 35 AM
Also in the current state, as far as i can tell, there is no way to access the select all button via keyboard

mikerodonnell89 (changes requested):
Two small things:

1. For every other item in the popover, I can click anywhere along the entire width of the item to select it and close the popover simultaneously. For the "select all", i need to click the text in the label explicitly, and this discrepancy between the two is a little odd UX-wise in my opinion
2. Still unable to focus the "select all" option via the arrow keys, which we wild need for accessibility

InnaAtanasova:
Just the focus on Select All item, like we discussed. The rest looks good!

g-cheishvili:
@mikerodonnell89 @InnaAtanasova Done

mikerodonnell89 (approved):

InnaAtanasova:
https://github.com/SAP/fundamental-ngx/assets/39598672/354b77ef-36f6-428f-b41b-07b2043e0bb3

When I use the keyboard to open the menu, for a few seconds I get "Invalid entry" message which then disappears. 
The focus for Select All is different when accessed by keyboard and by mouse.

Keyboard:
Screenshot 2023-06-21 at 4 32 08 PM


Mouse:
Screenshot 2023-06-21 at 4 31 47 PM

The focus should be like the one in the keyboard case (the whole width and no border)

InnaAtanasova (approved):

PR #9971, opened 2023-06-02
fix(core): update 72 Bold font name

### LINKED ISSUE #9814: Title component font-weight issues in different browsers
Bug
The Title component visually looks different (font-weight) in Chrome (left) and Safari (right) browsers.
Here's a comparison:
Screenshot 2023-05-09 at 4 56 05 PM

For a reference, here's how the component looks like in Fund-styles in the same browsers:
Screenshot 2023-05-09 at 4 57 50 PM

latest

droshev:
## Related Issue(s)

fixes #9814 

## Description

fix the right name of 72-Bold font name


libs/fn/schematics/ng-add/styles.ts
```diff
@@ -8,14 +8,14 @@ export const defaultFontStyle = `
 }
 
 @font-face {
-    font-family: '72';
+    font-family: ''72-Light';
```
g-cheishvili, line 11:
```suggestion
    font-family: '72-Light';
```

mikerodonnell89 (approved):

g-cheishvili (approved):
Looks like we do not have an option here

InnaAtanasova (approved):

PR #9985, opened 2023-06-05
fix(core): template based inline help was not setting screenreader element

### LINKED ISSUE #9276: [Sourcing][Accessibility] Tooltip not being read for linear reading mode in JAWS
Bug

Tooltip not being read for linear reading mode in JAWS screen reader

Steps to replicate the issue:
 1. Go to https://sap.github.io/fundamental-ngx/#/core/inline-help
 2. Under 'Inline Help' example, use JAWS linear reading mode to read the tooltips
 3. Observe the issue

mikerodonnell89:
fixes #9276


libs/core/src/lib/inline-help/inline-help.directive.ts
```diff
@@ -139,4 +132,19 @@ export class InlineHelpDirective extends BasePopoverClass implements OnInit, OnC
             this.additionalBodyClass += ' ' + INLINE_HELP_ICON_CLASS;
         }
     }
+
+    /** @hidden */
+    private _setupScreenreaderElement(content: string | Nullable<TemplateRef<any>>): void {
+        let srElement = document.createElement('span');
```
g-cheishvili, line 138: Since this folk is called from the input and it is very likely that it will be very important for this component to be rendered before serving to the client (SSR use case), I think this should be modified a little bit:
- Component should have `Renderer2` injected and code will look like this `let srElement = this.renderer2.createElement('span');`
- Optionally you can append using the `renderer2.appendChild`

g-cheishvili (changes requested):

libs/core/src/lib/inline-help/inline-help.directive.ts
```diff
@@ -139,4 +134,19 @@ export class InlineHelpDirective extends BasePopoverClass implements OnInit, OnC
...
+        if (typeof content === 'string') {
+            srElement.innerText = content;
+        } else if (content) {
+            srElement = content.createEmbeddedView(content).rootNodes[0];
+        }
+
+        if (srElement.style) {
+            srElement.style.cssText = `position: absolute !important; height: 1px; width: 1px; overflow: hidden; clip: rect(1px, 1px, 1px, 1px);`;
+        }
+        this._renderer.appendChild(this._elementRef.nativeElement, srElement);
```
g-cheishvili, line 150: I had a suspicion and I tested one thing. If you write in that template some Angular binding, it will not work. It will render once and will not pay attention to any further changes. Also, if content is changed then the DOM will be polluted with unnecessary things.
Also, the `createEmbeddedView` parameter is the context for the template, in this case, we do not have need for a context, especially the template, which would be unusable, just pass `null`.

What I propose is that:
1. Create `viewRef` as class property, when `_setupScreenreaderElement` is called, check if there is a `viewRef` already, if there is then you can call `destroy()` on that and it will be cleared. Same for Element type I suppose.

2. Instead of this `srElement = content.createEmbeddedView(content).rootNodes[0];` do 
```ts
this.srViewRef = content.createEmbeddedView(content);
this._viewContainerRef.insert(this.srViewRef); // This will fix issue with the binding
srElement = this.srViewRef.rootNodes[0];
```

g-cheishvili (approved):

PR #9991, opened 2023-06-07
feat(platform): table optimisation

### LINKED ISSUE #9746: [Sourcing] <fdp-table> Perf with virtualScroll and pageScrolling in certain window size

Bug and/or enhancement

Issue 1: (maybe enhancement)

In certain window size with platform table on dialog, when we have virtualScroll and pageScrolling enabled and when we show some large amount of rows on UI, if we scroll up or down very fast, whitespace will show. 

https://user-images.githubusercontent.com/22842279/234118058-9cecac04-64db-454d-a9d8-58fec0f3abe3.mov

Issue 2: (bug)

In certain window size with platform table on dialog, when we have virtualScroll and pageScrolling enabled and when we show some large amount of rows on UI, if we resize one column and then scroll up/down fast, in additional to whitespace, we see the scrollbar jump up and down for a while.

https://user-images.githubusercontent.com/22842279/234118907-ac69fb41-4dc3-49b2-8e13-7a7e4a9163c3.mov

After jumping is done, when we try to scroll to next page, we see that we still stay on the same page by looking at the first column.

https://user-images.githubusercontent.com/22842279/234118675-90c497bc-218a-468c-a558-173f230bfd06.mov

40.2.rc-12 or previous

Please refer the link below. Step to reproduce:

- Create a platform table with large amount of items (more than 1000) on dialog.
- Resize the width of window size like opening dev tool.
- Scroll down to fetch some amount of data several times.
- Try to scroll up and down fast.
- Then adjust one or more column size.
- Try to scroll up and down fast some times.

https://stackblitz.com/edit/angular-acnvmy?file=src%2Fapp%2Fplatform-table-page-scrolling-example.component.ts,src%2Fapp%2Fplatform-table-page-scrolling-example.component.html,package.json,node_modules%2F%40fundamental-ngx%2Fcore%2Fdialog%2Fdialog-service%2Fdialog.service.d.ts

### LINKED ISSUE #9926: [Sourcing] Tree table: ngDoCheck() and ChangeDetection gets triggered whenever we hover on table UI container or perform row expansion or collapse
A bug

We'd like to avoid unnecessary ChangeDetection (CD) being triggered since it will cause bad performance issues while loading large data set into table, or during expand/collapse table rows.

Current release with Angular 15.

- Create any view with the `<fdp-table>` component, and add `ngDoCheck()` to the component class
<pre>
    ngDoCheck() {
        console.warn('ngDoCheck() invoked...');
    }
</pre>
- Write a custom function as follows, inside HTML where the table component resides, and bind it to the `[hideItemCount]` property in `<fdp-table-toolbar>` like this `[hideItemCount]=onHideItemCount()`
<pre>
    onHideItemCount() {
        console.warn('onHideItemCount invoked...');
        return false;
    }
</pre>
- When hover over any area inside the table, noticed that in browser console it logs hundreds of `ngDoCheck() invoked....`, along with the `onHideItemCount invoked...` message invoked in HTML from the custom function `onHideItemCount()`

Screenshot 2023-05-25 at 6 19 53 PM

Right now in our component where we use `<fdp-table>`, we had quite a few function calls being invoked from HTML template. Due to the CD being triggered so often from the `<fdp-table>`,  those functions will be called over 10k of times.
We're already in progress replacing those function calls with custom pipes instead, but still would like the CD issue to be fixed from F-NGX side.

### LINKED ISSUE #9932: Platform Table Improvements
enhancement

The current platform table is hard to be maintained. We need to come up with improvements (including performance).
- [ ] we need to have an overview of all the current features so we don't break something after refactoring
- [ ] too big component => think how to split in smaller components like table body, table row, table cells(diff type cells), table header, table header cells
2w

### LINKED ISSUE #9955: [Sourcing] Unable to expand/collapse row after DND

bug

Following the enhancement for https://github.com/SAP/fundamental-ngx/issues/9718, after I try drag item on row which is not droppable, unable expand/collapse any row.

https://github.com/SAP/fundamental-ngx/assets/22842279/f06d2567-db2e-4d8d-99bb-f681f440ac7f

v0.42.0-rc.0

Try to dnd one row onto row that is not droppable -> then try to expand any row

Fundamental NGX playground

N1XUS:
## Related Issue(s)

closes #9932
closes #9926
closes #9955
closes #9746

## Description
Main idea behind this refactoring was:
- Reduce amount of recompiled code during development (Hot Module Reloading). This is achieved by ejecting helper classes, directives, models from `table` package into a separate `table-helpers`.
- Reduce amount of html code inside table template. this is achieved by ejecting table row related code into a separate components;
- Reduce amount of business logic inside table component itself. This is achieved by ejecting logical parts of code into a separate directives that. For example, virtual scroll directive would be initiated only if table element has `virtualScroll` property. Otherwise it will not be loaded at all, same goes to dnd support. Initial state and dataSource are always applied since it's a core functionality of the table;
- Reduce amount of change detection cycles:
  - Reduce triggers during event emission: now event listeners run outside zone, and triggering the change only when previous calculation result is different than current (for example, resize functionality: previously checked table and it's related components on every `mousemove`, now it reacts only if resize is started or finished. It will not trigger detection during the resize process).
  - Reduce triggers during table property changes: previously we ran change detection for whole table to recalculate everything. Now we can call it per each row if needed. Each row reacts on it's own data, so we shouldn't have 100+ detections at once when something changes for a particular row or cell inside it.
  - Rely more on async objects, such as visible rows array, table loading state, row checked state, etc.
  - Reduce method calls. Previously we had function evaluation for some properties (for example, getting element ref, checking whether the cell is the first in the row, etc.). That approach added tiny delays in function evaluation, thus, slowing down the detection itself. Now mostly we rely on calculated properties, which are being calculated only when the related properties are changed.
  - Reduce component re-initialization: Platform table now tries to reuse existing table row component instance and pass new data to it rather than re-create new component instance with it's view (thus reducing createEmbedView calls for row, inner cells, etc.);

Platform table refactoring includes following breaking changes:

**Components that rely on `HasElementRef` interface:**

- `HasElementRef` interface has been changed. Now instead of `elementRef()` method, components should have `elementRef: ElementRef` property. This can include getters too;

**Core table:**

- `TableCellDirective` now directly extends from `FocusableItemDirective`. Previously it had `FocusableItemDirective` as a host directive applied to itself;

**Platform table:**

- All directives, services, models, data source classes are moved from `@fundamental-ngx/platform/table` into a separate package `@fundamental-ngx/platform/table-helpers`. To keep backwards compatibility, `@fundamental-ngx/platform/table` re-exports `@fundamental-ngx/platform/table-helpers` so that the import paths are the same as before;
- Platform table directives are standalone now;
- Platform table data-source related functionality has been moved to `TableDataSourceDirective` standalone directive. Input properties are preserved;
- Platform table drag&drop related functionality has been moved to `TableDraggableDirective` standalone directive. Input properties are preserved;
- Platform table initial state related functionality has been moved to `TableInitialStateDirective` standalone directive. Input properties are preserved;
- Platform table virtual scroll related functionality has been moved to `TableVirtualScrollDirective` standalone directive. Input properties are preserved;
- Platform table data source has been refactored to utilize `@fundamental-ngx/cdk/data-source` implementation. Backwards compatibility is preserved for classes that extend from data-source/data-provider. More complex classes may need to be refactored according to new class signature;
- Platform table now uses `TableRowImpl` class as implementation of `TableRow` interface instead of simple object;
- Helper methods from TableComponent has been moved to a standalone functions;
- `SearchInput` type is now not re-exported from `@fundamental-ngx/platform/table`. Instead use `@fundamental-ngx/platform/search-field`;
- Table component html is now split between logical parts: table, table header rows, table content rows. Resulting markup is not changed, but the business logic parts are being moved to appropriate components;
- Platform table rows now try to reuse the html element they bounded to. This means that if developers use reference to elementRef of the row, and rows array changes, they need to manually recalculate the reference to the element;
- Platform table group row: text pattern changed from {{ column.name }} : {{ column.value }} to {{ column.name }}: {{ column.value }}. Space before colon has been removed;
- Platform table input properties recalculation now more efficient. Previously we checked property value on every `mousemove`, now, only when something inside the table itself is changed. This may require additional detectChanges calls from developers who are using Platform Table.
- Such input properties has been removed from platform table class:
  - `dataSource `- setter available via `fdp-table[dataSource]` but class property now accessible from `_dataSourceDirective` property of Table component;
  -  `state`, `initialVisibleColumns`, `initialSortBy`, `initialFilterBy`, `initialGroupBy`, `initialPage` setters available via `fdp-table[property]` but the class property now accessible from `initialState` property of Table component;
  - `isTreeTable`, `enableRowReordering`, `dropMode` setters available via `fdp-table[property]` but the class property now accessible via `_dndTableDirective` property of Table component;
  - `virtualScroll`, `renderAhead` setter available via `fdp-table[virtualScroll]`, but the class property now accessible via `_virtualScrollDirective` property of Table component;
- Such events has been removed from Platform Table class:
  - `rowsRearrange` - listener still available via `fdp-table(rowsRearrange)`, but actual eventEmitter is now located in `_dndTableDirective` property of Table component;
  - `onDataRequested`, `onDataReceived` - listeners still available via `fdp-table(event)`, but actual eventEmitters are now located in `_dataSourceDirective` property of Table component;
- Private API changed. For more information refer to [this PR](https://github.com/SAP/fundamental-ngx/pull/9991)

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [x] poor examples

4. Accessibility testing

##### PR Quality

-   [x] update `README.md`


InnaAtanasova:
A few issues I noticed with the focus. 
This one is for focusable table. The styling for the row focus looks off and also I can't go back and focus the table header cells:

https://github.com/SAP/fundamental-ngx/assets/39598672/d1f3f182-b775-488f-888d-ac2e0f7e0608

InnaAtanasova:
In tables with checkboxes I should be able to navigate through the focusable elements with the keyboard. I noticed this issue for all examples with checkboxes and email links. The user should be able to use the keyboard to reach out to every interactive element.
https://github.com/SAP/fundamental-ngx/assets/39598672/bcb67b9d-b83d-4390-b561-d388b1b81998

mikerodonnell89 (commented):
Why are so many files having their constructor parameters changed from private to public?

N1XUS:
> Why are so many files having their constructor parameters changed from private to public?

`HasElementRef` interface has been changed. Now instead of `elementRef()` method, components should have `elementRef: ElementRef` property.  The reason is that previously we had a function without arguments that just returned elementRef. Function evaluation may take some insufficient amount, yet it may block the thread for that time.

N1XUS:
@InnaAtanasova checked your cases on main. If I'm not mistaken, it has same behaviour, no?

libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
```diff
@@ -127,8 +149,8 @@ export class FocusableItemDirective implements HasElementRef {
     }
 
     /** @hidden */
-    @HostListener('focusin')
-    async _onFocusin(): Promise<void> {
+    // @HostListener('focusin')
```
mikerodonnell89, line 152: If we leave these comments maybe make a note that it's really handled in `runOutsideAngular`

libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
```diff
@@ -92,27 +92,49 @@ export class FocusableItemDirective implements HasElementRef {
...
-        private _tabbableElementService: TabbableElementService,
-        private _liveAnnouncer: LiveAnnouncer
-    ) {
+    protected readonly _zone = inject(NgZone);
+
+    /** @hidden */
+    constructor() {
         this._focusableObserver
-            .observe(this._elementRef, false)
+            .observe(this.elementRef, false)
```
fkolar, line 111: What is the reason we are changing from private to public? Maybe comment would help .

libs/cdk/src/lib/utils/directives/selectable-list/selectable-item.directive.ts
```diff
@@ -81,7 +81,7 @@ export class SelectableItemDirective<ElementType extends Element = HTMLElement,
         private disabled$: FdkDisabledProvider,
         private readonly$: FdkReadonlyProvider,
         private selectionService: SelectionService<ElementType, ValueType>,
-        private _elementRef: ElementRef<ElementType>,
+        public readonly elementRef: ElementRef<ElementType>,
```
fkolar, line 84: The same comment as above. just missing reason / connection why we change this private / public modifier - comment would be helpful

libs/cdk/src/lib/utils/helpers/get-native-element.ts
```diff
@@ -19,6 +19,6 @@ export function getNativeElement<T extends Element = HTMLElement>(
  */
 export function isHasElementRef<T extends Element = HTMLElement>(
```
fkolar, line 20: Maybe when we are refactoring we can renaming this confusing function too. **isHas** and just leave **has**

libs/core/src/lib/tabs/tab-item/tab-item.directive.ts
```diff
@@ -48,7 +48,7 @@ export class TabItemDirective implements CssClassBuilder, OnChanges, OnInit {
     linkItem: TabLinkDirective;
 
     /** @hidden */
-    constructor(private _elementRef: ElementRef) {}
+    constructor(public readonly elementRef: ElementRef) {}
```
fkolar, line 51: Do we want to keep all these files in sync? I see above we refactored these from construction to properties used `inject()` function and why not in these classes?

fkolar (changes requested):
In general

fkolar:
Anything is going to change for the app developer - There any breaking changes?

fkolar (changes requested):
Can you also check the https://github.com/SAP/fundamental-ngx/pull/9947, it does not work in this branch. Last time it was overriden by changes that were marged the same day and now I see it stopped working again.

libs/cdk/src/lib/utils/directives/selectable-list/selectable-item.directive.ts
N1XUS: Hey @fkolar. Thank you for your review. The reason why this part got changes is that main interface `HasElementRef` was changed (it was used somethere for table directives/components), and that affected majority of other classes. The reason to move from method call to a direct property access is to reduce amount of function evaluations (even though it had only `return this._elementRef` line, VM would still need to evaluate it), and to simplify the components itself.

libs/core/src/lib/tabs/tab-item/tab-item.directive.ts
N1XUS: The main task is about platform table components. These are the side effects and can be updated separately.

N1XUS:
> Anything is going to change for the app developer - There any breaking changes?

Breaking changes are documented in the description of the PR and will be added during merge to generate breaking change log.

N1XUS:
> I like the idea splitting the code into separate standalone component. I am fine with the answers, the only things are my general comments that some functionality stopped working . I see Inna also reported some regressions.

I've checked Inna's comments, and the functionality is the same as on main branch.

N1XUS:
> Can you also check the #9947, it does not work in this branch. Last time it was overriden by changes that were marged the same day and now I see it stopped working again.

Fixed

InnaAtanasova (approved):

InnaAtanasova:
The problems I mentioned are not a regression, they are also present in main. I created an issue for these cases: https://github.com/SAP/fundamental-ngx/issues/10021

mikerodonnell89 (dismissed):
Loading state boolean displaying as "true" even when the table is not loading: 
Screenshot 2023-06-13 at 10 23 21 AM

mikerodonnell89:
Initial state example - can no longer focus the true/false group by rows via keyboard

https://github.com/SAP/fundamental-ngx/assets/2471874/b0c347e7-44a6-4d6f-b6fe-a8792f7d7e14

libs/cdk/src/lib/utils/directives/selectable-list/selection.service.spec.ts
```diff
@@ -36,7 +36,7 @@ class TestDirective implements SelectableItemToken<HTMLElement, string> {
         this.clicked.emit();
     }
 
-    elementRef(): ElementRef<HTMLElement> {
+    get elementRef(): ElementRef<HTMLElement> {
```
mikerodonnell89, line 39: I am noticing here and in a handful of components that they are still using private element refs and public getter functions. Why not use public elementref on these?

g-cheishvili:
libs/platform/src/lib/table-helpers/directives/table-cell-resizable.directive.ts _cdr is not used

libs/platform/src/lib/table/components/table-p13-dialog/filtering/filter-rule.component.html *ngSwitchCaseDefault -> *ngSwitchDefault

https://github.com/SAP/fundamental-ngx/blob/0d898c849a88869fd6a88cd6a8b3a01546e0eb33/libs/platform/src/lib/table/components/table-row/table-row.component.ts#L119 - not used anywhere

https://github.com/SAP/fundamental-ngx/blob/043d1735754bb890fc5f0bace7d09b19ba2dfb54/libs/platform/src/lib/table/components/table-group-row/table-group-row.component.html#L45 - Here we have a type mismatch thing happening, I would say that we should have a small helper function or used `$any` to fight these warnings

https://github.com/SAP/fundamental-ngx/blob/0d898c849a88869fd6a88cd6a8b3a01546e0eb33/libs/platform/src/lib/table/components/table-header-row/table-header-row.component.ts#L153-L158 - we do not use this and seems like it is a private function, so why not remove this?

https://github.com/SAP/fundamental-ngx/blob/180d05b95725ed07957d416a1dd75f17a3d5655c/libs/platform/src/lib/table/table.component.html#L73 - unused rowIdx

https://github.com/SAP/fundamental-ngx/blob/0ae7728069bf2c874efcd2031680b1d1acf4fea6/libs/platform/src/lib/table/table.component.ts#L561 - unused property

https://github.com/SAP/fundamental-ngx/blob/92e1f432c87257a813ad51c8dfbf0a58df0f5be9/libs/platform/src/lib/table-helpers/directives/table-cell-resizable.directive.ts#L90-L94 - this can be done in distinctUntilChanged pipe callback function. First, map to the desired structure, then just distinctUntilChanged. This way you can even avoid saving prevData

Could not find any circular dependencies, which is awesome, considering the functionality over here, nice work

libs/cdk/src/lib/utils/directives/selectable-list/selection.service.spec.ts
N1XUS: @mikerodonnell89 in some cases elementRef is refefred to inner viewChild

fkolar (approved):

N1XUS:
@g-cheishvili on the last item of the comment - yes, it's can't be done differently since it will bring circular dependency (I've tried splitting, and this is the best aproach at the moment)

g-cheishvili:
> @g-cheishvili on the last item of the comment - yes, it's can't be done differently since it will bring circular dependency (I've tried splitting, and this is the best aproach at the moment)

Even if you use InjectionToken it still recognizes circularity?

N1XUS:
> > @g-cheishvili on the last item of the comment - yes, it's can't be done differently since it will bring circular dependency (I've tried splitting, and this is the best aproach at the moment)
> 
> Even if you use InjectionToken it still recognizes circularity?

Yes.

N1XUS:
> Initial state example - can no longer focus the true/false group by rows via keyboard
> 
>  Screen.Recording.2023-06-13.at.10.26.05.AM.mov

fixed

N1XUS:
@g-cheishvili:
`- Here we have a type mismatch thing happening, I would say that we should have a small helper function or used $any to fight these warnings` -> I wouldn't say that calling helper function for each row simply to remove IDE warning is a performance improvement.

g-cheishvili (approved):
Besides this
https://github.com/SAP/fundamental-ngx/blob/4263d33dd919828f92992c97adaf8291be6a64ba/libs/platform/src/lib/table-helpers/directives/table-cell-resizable.directive.ts#L80

I am more than happy to include this in the main. Amazing job

PR #10038, opened 2023-06-16
fix(core): new checkbox hidden element for a11y

mikerodonnell89:
will be needed for https://github.com/SAP/fundamental-styles/pull/4608

I am fairly sure this can be merged right away without needing hardcoded css from the aforementioned pr


g-cheishvili (approved):
Yeah, it will work without the CSS as before

InnaAtanasova (approved):

Issue #9236, opened 2023-01-16
Defect Hunting Sprint 115


InnaAtanasova:
Please list the issues you've found during the defect hunting.

mikerodonnell89:
- [ ] Core combobox: use keyboard arrows to focus items in the popover. Press escape. Focus does not return to combobox

mikerodonnell89:
**Core:**
- [ ] Core carousel - need to replace placeholder images
<img width="370" alt="Screenshot 2023-06-15 at 10 40 55 AM" src="https://github.com/SAP/fundamental-ngx/assets/2471874/55e9bddf-3608-44b7-a311-4d6c2d8b44f7">

- [ ] Console error thrown in illustrated message doc page
- [ ] input with inline help alignment looks off
- [ ] reactive form input example - no visual indicator for what fields are required
- [ ] list item with buttons - cutoff border on hover
<img width="1291" alt="Screenshot 2023-06-15 at 12 22 15 PM" src="https://github.com/SAP/fundamental-ngx/assets/2471874/4614fb14-88d9-4184-a53a-adb84a53155e">


**Platform:**
- [ ] combobox - focus items in the list using keyboard arrow keys. Press escape. Focus does not return to combobox

mikerodonnell89:
Carousel loading indicator is off center
<img width="402" alt="Screenshot 2023-06-21 at 2 45 33 PM" src="https://github.com/SAP/fundamental-ngx/assets/2471874/7019b0f7-dc62-4702-989d-1c9c8c04d0e6">

PR #10061, opened 2023-06-21
fix(docs): placeimg is shutting down so new placeholder site

mikerodonnell89:
part of #9236 

Placeimg is shutting down at the end of the month, so I'm moving all of our doc placeholder images to picsum.photos which we were already using in some other places

I left "nature", "people" etc in the urls if we find a service that supports this. Leaving a query param with a random number ensures that the browser doesn't cache the same image repeatedly, which is important on docs like the carousel etc


g-cheishvili:
> part of #9236
> 
> Placeimg is shutting down at the end of the month, so I'm moving all of our doc placeholder images to picsum.photos which we were already using in some other places
> 
> I left "nature", "people" etc in the urls if we find a service that supports this. Leaving a query param with a random number ensures that the browser doesn't cache the same image repeatedly, which is important on docs like the carousel etc

Why not just download and serve locally from documentation assets? We would not rely on 3rd parties

mikerodonnell89:
> > part of #9236
> > Placeimg is shutting down at the end of the month, so I'm moving all of our doc placeholder images to picsum.photos which we were already using in some other places
> > I left "nature", "people" etc in the urls if we find a service that supports this. Leaving a query param with a random number ensures that the browser doesn't cache the same image repeatedly, which is important on docs like the carousel etc
> 
> Why not just download and serve locally from documentation assets? We would not rely on 3rd parties

Sure that's an option, I am not sure if we initially went with these placeholder sites to keep the repo size down or something. @InnaAtanasova @droshev ?

g-cheishvili:
> > > part of #9236
> > > Placeimg is shutting down at the end of the month, so I'm moving all of our doc placeholder images to picsum.photos which we were already using in some other places
> > > I left "nature", "people" etc in the urls if we find a service that supports this. Leaving a query param with a random number ensures that the browser doesn't cache the same image repeatedly, which is important on docs like the carousel etc
> > 
> > 
> > Why not just download and serve locally from documentation assets? We would not rely on 3rd parties
> 
> Sure that's an option, I am not sure if we initially went with these placeholder sites to keep the repo size down or something. @InnaAtanasova @droshev ?

We have removed so much from repo that even if you add now 100 of these images total will still not be near the size of previous one 😄 But in all seriousness reliance on 3rd party sometimes shoots in our leg because sometimes e2e fails

InnaAtanasova (approved):

PR #10063, opened 2023-06-21
fix(core): redundant popover open/close/event things

mikerodonnell89:
part of #9236 

Both `PopoverComponent` and `PopoverService` extend `BasePopoverClass`. Both were emitting the `isOpenChange` event, causing mobile popovers to be opened twice. This is an immediate fix for that problem but I think we need to restructure these components, in my opinion it is not great to have the component and service extending the same base class

This solution of ascending the element parent node tree feels a bit code smelly but I think it is the best option for two reasons

1. We can solve the problem of ascending the parent tree by requiring the application developer to add `fd-popover-body` wrapper around the contents of the dialog. I'd like to avoid this since it is a breaking change
2. I thought why not just add `fd-popover-body` to the `popover-mobile.component.html` but to do this we need to import the popover module, which introduces a circular dependency between popover mobile module and popover module, because popover component uses `createDynamicModule`

So I think this is the best option, barring a more significant refactor


g-cheishvili (approved):

InnaAtanasova (approved):

mikerodonnell89:
failing on unrelated e2e, merging

PR #10089, opened 2023-06-27
chore: tsconfig standartisation

N1XUS:
What's done:
- Standartised tsconfig files across secondary entries;
- Added missing `angularCompilerOptions` config object for some entries;
- Refactored imports to use `index.ts` file directly instead of `public_api.ts`


g-cheishvili (approved):
I especially like removal of public_api files

Issue #9823, opened 2023-05-11
Horizon 2023 Project


droshev:
List of components to be created/updated:
- [x] **bring Angular 16** - This is in separate ticket
- [x] Avatar #9825 - https://github.com/SAP/fundamental-ngx/pull/10553
- [x] Avatar Group #9826
- [x] Busy Indicator #9827
- [x] Buttons (Standard, Menu Button, Split Button, Segmented Button, Toggle Button) #9828
- [x] Calendar/Date picker/Date Time picker #9829 - https://github.com/SAP/fundamental-ngx/pull/10324
- [x] Card #9831
- [x] Carousel  #9821 - https://github.com/SAP/fundamental-ngx/pull/10274
- [x] Checkbox #9832 - https://github.com/SAP/fundamental-ngx/pull/10315
- [x] Combobox #9833 - no changes needed
- [x] Dialog #9834 - https://github.com/SAP/fundamental-ngx/pull/10569
- [x] Dynamic Page #9835 - no changes needed
- [x] Dynamic Side Content #9836 - no changes needed
- [x] Flexible Column Layout #9838 
- [x] Icon Tab Bar #9839
- [x] Input #9840 
- [x] Label #9841 - no changes required
- [x] Link #9842
- [x] List #9843
- [x] Menu #9844
- [x] Message Strip #9845
- [x] Multi Combobox #9846
- [x] Multi Input #9847 - no changes needed
- [x] Page #9848 - no changes needed
- [x] Popover #9849
- [x] Progress Indicator #9850
- [x] Radio Button #9851
- [x] Shell #9852 - https://github.com/SAP/fundamental-ngx/pull/10275
- [x] Select #9853
- [x] Sllider and Range Slider #9854 - Denis
- [x] Step Input #9855
- [x] Switch #9856
- [x] Table #9857 - no changes needed
- [x] Tag #9858
- [x] Textarea #9859
- [x] Tile #9860 - no changes needed
- [X] Title #9861 - no changes needed
- [x] Toolbar #9824 - https://github.com/SAP/fundamental-ngx/pull/10314

New Design Changes:
- [x] Splitter #9890 - no changes needed
- [x] Message Popover #9889 - https://github.com/SAP/fundamental-ngx/pull/10317
- [x] Panel #9896
- [x] Token and Tokenizer #9920
- [x] Object Status #10070 - https://github.com/SAP/fundamental-ngx/pull/10539
- [x] Bar

BTP 
[...]

PR #10091, opened 2023-06-27
feat(core): added wrapping to the checkbox label

### LINKED ISSUE #9832: Component Design Update: Checkbox
- [ ] review markup changes from fd-styles
- [ ] Horizon specs(visual and interaction) updates
- [ ] Delta theming
- [ ] identify and document breaking changes

g-cheishvili:
## Related Issue(s)

closes #9832
part of #9823 

## Description

Not much has changed, but I noticed that label wrapping was missing from the API

### DIFF
--- a/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
+++ b/libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
@@ -61,10 +61,10 @@ export class CheckboxComponent implements ControlValueAccessor, AfterViewInit, O
     @ViewChild('labelElement')
     labelElement: ElementRef;
 
-    /** @hidden Whether input label should be wrapped */
+    /** Whether input label should be wrapped */
     @Input() wrapLabel: boolean;
 
-    /** @hidden Whether input label should be wrapped */
+    /** Vertical position of the label compared to the checkbox box */
     @Input() valignLabel: 'top' | 'middle' = 'middle';
 
     /** Sets the `aria-label` attribute to the element. */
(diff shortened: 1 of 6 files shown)


libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
```diff
@@ -61,6 +61,12 @@ export class CheckboxComponent implements ControlValueAccessor, AfterViewInit, O
     @ViewChild('labelElement')
     labelElement: ElementRef;
 
+    /** @hidden Whether input label should be wrapped */
+    @Input() wrapLabel: boolean;
+
+    /** @hidden Whether input label should be wrapped */
```
InnaAtanasova, line 67: same text as on line 64

libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
```diff
@@ -61,6 +61,12 @@ export class CheckboxComponent implements ControlValueAccessor, AfterViewInit, O
...
+    @Input() valignLabel: 'top' | 'middle' = 'middle';
```
InnaAtanasova, line 68: I am not sure about `valignLabel` name :D `alignLabel`?

g-cheishvili: It's a reminiscent of valign attribute from table cells 😁 I think it describes purpose best 😁

InnaAtanasova: let's hope the user makes the same association :D

mikerodonnell89 (commented):
lgtm other than @InnaAtanasova 's comment about the same api doc text

libs/core/src/lib/checkbox/checkbox/checkbox.component.ts
g-cheishvili: Hope never dies

PR #10170, opened 2023-07-14
fix(core): remove manual change detection from dialog container

### LINKED ISSUE #9682: Busy Indicator in Component dialog not closing after value change.
Bug

In component based dialog, any data change after initial load is not reflected until a click event or button press event is triggered.

0.39.0 and 0.40.0

Open a component based dialog.
Change any value of a variable or set busyindicator to false after sometime or after completion of any API call.
The changes are not reflected until you click or press any button.

Stackblitz code provided as example.
https://stackblitz.com/edit/angular-mylhjb?file=src%2Fapp%2Fdialog-example.component.ts,src%2Fapp%2Fcomponent-based-dialog-example.component.ts

mikerodonnell89:
fixes #9682

### DIFF
--- a/libs/core/src/lib/dialog/dialog-container/dialog-container.component.ts
+++ b/libs/core/src/lib/dialog/dialog-container/dialog-container.component.ts
@@ -1,5 +1,7 @@
 import {
+    AfterContentChecked,
     AfterViewInit,
+    ChangeDetectionStrategy,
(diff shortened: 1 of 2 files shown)


g-cheishvili:
I would suggest to check deeper why this issue happens, otherwise removing on push is like that flex seal ad

Issue #10171, opened 2023-07-15
[Sourcing] `fd-switch` keeps toggling every scroll in `virtualScroll`


CSENoni:
#### Is this a bug, enhancement, or feature request?

bug (and questions)

#### Briefly describe your proposal.

Please help us with this issue when `fd-switch` is jumping every scroll. For this case, it should be replicable at the page after `renderAhead` (100 in my link) and `[(checked)]` is set. When only on-way binding `[checked]` is set, the toggle is not updated when we scroll down or up. This can be replicable with `virtualScroll`.

https://github.com/SAP/fundamental-ngx/assets/22842279/7f2e2206-7af0-4ef9-8737-8ebed1160108

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)

current or version after `0.42.0-rc.14`

#### If this is a bug, please provide steps for reproducing it.

Please refer to link:

#### Please provide relevant source code if applicable.

`fd-switch` (case 2)
https://stackblitz.com/edit/angular-ttxut2-cklqbt?file=src%2Fapp%2Fdisplay-text.component.ts,src%2Fapp%2Fapp.module.ts,src%2Fapp%2Fplatform-table-page-scrolling-example.component.html,src%2Fapp%2Fplatform-table-page-scrolling-example.component.ts

#### Is there anything else we should know?
`fdp-switch` may also be affected so please take a look at it also.

CSENoni:
This issue has been modified to focus on one issue. Please refer to https://github.com/SAP/fundamental-ngx/issues/10183 for _case 3_.

fkolar:
I can confirm if you (on provided stackblitz example) e.g. on the row 43 switch those toggles ON, and you start scrolling up and down, you can see the flickering and those switches refreshes. 
On does not happen with line 30 for example

fkolar:
@CSENoni , Do you think  you can quickly check this agian. Today I was not able to reproduce it.

CSENoni:
> @CSENoni , Do you think you can quickly check this agian. Today I was not able to reproduce it.

Hi Fran, you can use this link as baseline: https://stackblitz.com/edit/angular-ttxut2-ujafdu?file=src%2Fapp%2Fdisplay-text.component.ts,src%2Fapp%2Fapp.module.ts,src%2Fapp%2Fplatform-table-page-scrolling-example.component.html,package.json 

[...]

PR #10200, opened 2023-07-21
fix(core): only apply the animation/transition to the switch when the…

mikerodonnell89:
… switch is in the process of being toggled

part of #10171 

Adds a class to the fd-switch that removes any animation/transition. Removes this class while the checking/unchecking process is happening, allowing for the transition. This removes the flickering effect when the switch component is first rendered


libs/core/src/lib/switch/switch.component.scss
```diff
@@ -7,3 +7,11 @@
 .fd-switch__invisible-text {
     display: none !important;
 }
+
+// to prevent the transition when the switch is first rendered
+.fd-switch-no-animate * {
+    -webkit-transition: none !important;
+    -moz-transition: none !important;
+    -ms-transition: none !important;
+    -o-transition: none !important;
```
N1XUS, line 16: I think the browsers that we support already can use `transition` property without vendor prefixes

fkolar (approved):
Looks good to me. I am not sure if vendor prefixes are dead

libs/core/src/lib/switch/switch.component.scss
fkolar: I am not sure if there is already 100% support across all the browsers, desktop and mobile

N1XUS: https://caniuse.com/?search=transition

N1XUS: @fkolar we are supporting evergreen browsers, and they perfectly support transition property without vendor prefixes

fkolar: Cool. so we just vendor prefixes and we should be good to go right ?

PR #10262, opened 2023-08-01
chore(core,platform,cdk): replaced DestroyService usages with DestroyRef

g-cheishvili:
## Related Issue(s)

closes None

## Description

As of Angular 16, there is no need for our DestroyService anymore, so this PR replaces that with built-in alternative, and deprecates existing service

### DIFF
--- a/libs/cdk/src/lib/forms/cva/cva-control.class.ts
+++ b/libs/cdk/src/lib/forms/cva/cva-control.class.ts
@@ -1,7 +1,6 @@
-import { ChangeDetectorRef, inject, Injectable } from '@angular/core';
-import { takeUntil } from 'rxjs';
+import { ChangeDetectorRef, DestroyRef, inject, Injectable } from '@angular/core';
 import { CvaDirective } from './cva.directive';
-import { DestroyedService } from '@fundamental-ngx/cdk/utils';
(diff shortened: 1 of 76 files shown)


libs/cdk/src/lib/forms/cva/cva-control.class.ts
```diff
@@ -25,15 +24,15 @@ export class CvaControl<T> {
     protected _changeDetector = inject(ChangeDetectorRef);
 
     /** @Hidden */
-    protected _destroy$ = inject(DestroyedService);
+    protected _destroyRef = inject(DestroyRef);
 
     /** @hidden */
     listenToChanges(): void {
-        this.cvaDirective?.markForCheck.pipe(takeUntil(this._destroy$)).subscribe(() => {
+        this.cvaDirective?.markForCheck.pipe(takeUntilDestroyed(this._destroyRef)).subscribe(() => {
```
N1XUS, line 31: Isn't takeUntilDestroyed can inject destroyRef if none provided?

g-cheishvili:
> Code looks good. @g-cheishvili have you checked memory leaks by any chance?

No but everything is exactly as before, just less additional dependencies

Issue #7226, opened 2021-11-23
Feed Input: Avatar shape


daisy-yovcheva:
#### Is this a bug, enhancement, or feature request?
Small Enhancement

#### Which versions of Angular and Fundamental Library for Angular are affected? (If this is a feature request, use current version.)
0.33.0-rc.156

#### If this is a bug, please provide steps for reproducing it.
The avatar should be in a "circular" shape.

Fiori ref: https://experience.sap.com/internal/fiori-design-web/feedinput/#feed-input1

![Screen Shot 2021-11-22 at 4 17 05 PM](https://user-images.githubusercontent.com/72403541/142953879-07fe8e89-7c71-4c43-b127-a45c417a211e.png)

UI5 Demo: https://sapui5.hana.ondemand.com/#/entity/sap.m.FeedInput/sample/sap.m.sample.FeedInput

droshev:
you can achieve that by setting that input to the avatar

IB002:
Address by [ ](https://github.com/SAP/fundamental-ngx/pull/10273)

PR #10273, opened 2023-08-02
docs(core): added feed input example

IB002:
## Related Issue(s)

Related to request https://github.com/SAP/fundamental-ngx/issues/7226

closes

## Description

Adds a feed input example with a circular avatar icon


mikerodonnell89 (changes requested):
An e2e test is failing, on account of the new example not containing an image. Here's the test:

```
    it('should have example with avatar image assigned', async () => {
        const inputAvatarLength = await getElementArrayLength(feedInputAvatar);
        for (let i = 0; i < inputAvatarLength; i++) {
            if (i === 1) {
                continue;
            }
            await waitForPresent(feedInputAvatar, i);
            await expect(await getAttributeByName(feedInputAvatar, 'image', i)).not.toBe('');
            await expect(await getAttributeByName(feedInputAvatar, 'image', i)).not.toBeNull();
        }
    });
```

Failing on that last expectation. My suggestion is to use an image in the circle avatar for this example rather than changing the test

droshev (approved):

PR #10314, opened 2023-08-07
feat(core): toolbar horizon update

### LINKED ISSUE #9824: Component Design Update: Toolbar
- [ ] review markup changes from fd-styles
- [ ] Horizon specs(visual and interaction) updates
- [ ] Delta theming
- [ ] identify and document breaking changes

N1XUS:
closes #9824
BREAKING CHANGE:
Removed inner `<div class="fd-toolbar">...</div>`.
Now classes and other attributes applied to the root `fd-toolbar` element.

### DIFF
--- a/libs/core/src/lib/toolbar/toolbar.component.ts
+++ b/libs/core/src/lib/toolbar/toolbar.component.ts
@@ -10,9 +10,10 @@ import {
     DestroyRef,
     ElementRef,
     forwardRef,
+    HostBinding,
+    inject,
     Inject,
     Input,
-    NgZone,
     Optional,
     QueryList,
     SkipSelf,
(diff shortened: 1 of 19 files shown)


libs/core/src/lib/toolbar/toolbar.component.ts
```diff
@@ -286,9 +311,9 @@ export class ToolbarComponent implements AfterViewInit, AfterViewChecked, CssCla
     }
 
     /** @hidden */
-    get elementRef(): ElementRef {
-        return this.toolbar;
-    }
+    // get elementRef(): ElementRef {
```
mikerodonnell89, line 314: Can just delete this i think

mikerodonnell89 (changes requested):
The popover body has lost its padding


Before:
Screenshot 2023-08-08 at 11 00 59 AM

After:
Screenshot 2023-08-08 at 11 00 12 AM

mikerodonnell89 (changes requested):
Popover has some padding now but is missing any sort of border. Looking at other menus, it's not unique to the toolbar overflow
Screenshot 2023-08-09 at 9 38 54 AM

N1XUS:
> Popover has some padding now but is missing any sort of border. Looking at other menus, it's not unique to the toolbar overflow Screenshot 2023-08-09 at 9 38 54 AM

That will be fixed with popover update

Issue #6495, opened 2021-09-10
Update the guide how to create new component in the library


InnaAtanasova:
This guide needs to include detailed steps of how to create a new component in the library, assuming it's a new contributor who has never worked with our library before. 
It should include everything from where to register the new module, how to include the new component in the documentation, where to add the module so it's properly exported, what files are needed for the documentation, Stackblitz, etc.
And last but not least all the steps to test the component, add/update existing tests, etc. 

Please use this page: https://github.com/SAP/fundamental-ngx/wiki/Guide-on-how-to-create-new-component-in-Fundamental-ngx

Once it's finalized we will update the existing readme file.

IB002:
Updated with [ #10322](https://github.com/SAP/fundamental-ngx/pull/10322)

PR #10322, opened 2023-08-10
docs(core): updated new component readme

IB002:
## Related Issue(s)

Related to issue https://github.com/SAP/fundamental-ngx/issues/6495

### DIFF
--- a/libs/docs/shared/NEW_COMPONENT.md
+++ b/libs/docs/shared/NEW_COMPONENT.md
@@ -1,99 +1,257 @@
-# Implementing a new component for Fundamental Library for Angular
+# Implementing A New Component For Fundamental Library For Angular
 
-In this guide, we'll explore the library and documentation code base and create a new component of our own.
+## Content
 
-### Code Base Structure
+-   [1. Code Structure](#1)
+-   [2.Prerequisites ](#2)
+-   [3. Generating A New Component](#3)
+-   [4. Component Example](#4)
+-   [5. Updating Documentation](#5)
+-   [6. Publishing](#6)
 
-This repository contains two separate projects - the component library and the documentation application. Different build tasks are used, depending on which project we're building. However, the documentation app consumes the library source directly, so there's no need to compile the library to test changes you're making to a component - simply running the documentation app locally will serve changes to the library immediately.
+## <a name="1"></a>1. Code Structure
 
-## Generating new component
+The fundamental-ngx repository is split into two projects
 
-Creating a new component is pretty straightforward. If you have the `NX Console` plugin, simply open it and run the `sap-component` schematic. This can also be done manually using the command line.
+-   the component library: a centralized modular repository made from components. It allow users to easily insert/reuse a large part of the code improving the quality and time required for their application.
+-   the documentation app: used to catalog the various different components, as well as present use cases for them.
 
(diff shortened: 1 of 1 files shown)


libs/docs/shared/NEW_COMPONENT.md
```diff
@@ -1,99 +1,249 @@
...
+`npx nx generate sap-component -–name=[component name] –-project=[package name]`
 
-## Updating documentation
+The following command takes two inputs; the first one is the name of the new component, and is located after “name=”. Any name can be chosen, however it is recommended that the name used, is relevant to the functionality of the component. Additionally, the name needs to be in kebab-case/dash-case, if you intend to contribute to the ngx library.The second input relates to which component library the newly generated component will belong to. Currently, there are four, present, in fundamental-ngx, which are: “Core”, “Platform Docs”, “CX”, “CDK”.  To select your preferred library, you just need to add its name in lowercase after “project=”. For example, to access “Core” you need to use “core”. For “Platform Docs”, you can use either “platform”, or “fn”; the rest of the libraries follow the same logic. If you are new to ngx, or are unsure which library is best suited for your needs, it is recommended that you use “core”.
```
g-cheishvili, line 32:
```suggestion
The following command takes two inputs; the first one is the name of the new component, and is located after “name=”. Any name can be chosen, however it is recommended that the name used, is relevant to the functionality of the component. Additionally, the name needs to be in kebab-case/dash-case, if you intend to contribute to the ngx library.The second input relates to which component library the newly generated component will belong to. Currently, there are four, present, in fundamental-ngx, which are: “Core”, “Platform Docs”, “CX”, “CDK”.  To select your preferred library, you just need to add its name in lowercase after “project=”. For example, to access “Core” you need to use “core”. For “Platform”, you can use “platform”; the rest of the libraries follow the same logic. If you are new to ngx, or are unsure which library is best suited for your needs, it is recommended that you use “core”.
```

no more `fn`

g-cheishvili (approved):

libs/docs/shared/NEW_COMPONENT.md
```diff
@@ -1,99 +1,249 @@
...
+
+    <fd-header-tabs></fd-header-tabs>
+</fd-doc-page>
+```
+
+We will now look at the example folder ,`examples`, which is used to contain all the created examples for the functionality of the component. You will notice that the folder is not empty, as a default example was created when generating the component. To demonstrate how to add more examples, we will create a new one, that will contain two copies of the button created previously. Firstly, we will create a folder that has a relevant name to the example. The most utilized naming conventions here are  `[component name]-[example name]-example`, and `[example-name]-example,where all the words are separated by a dashed line. As for the "example name", the most discriptive, and short one, should be used, for example since our example will contain two instances of our button component, we can just call it `double-button`, so that it clearly communicates to other users what it is demonstrating. For the naming conventions, while both are equally popular, a common rule of thumb is to use the former for components with shorter names, while the latter for components with longer or more complex names.
```
mikerodonnell89, line 145: Some formatting oddities in here like spaces before commas, which I think should be fixed by linting. Also I think "where all the words are separated by a dashed line. As for the "example name", the most discriptive, and short one, should be used, for example since our example will contain two instances of our button component, we can just call it" is formatted as if it is code, missing a back quotation mark somewhere

mikerodonnell89 (commented):
Looks great! Some minor comments

droshev (approved):

PR #10325, opened 2023-08-11
don't merge: demonstrate how to add new items using the form generator

mikerodonnell89:
fixes none


g-cheishvili (approved):

g-cheishvili:
I think this is not adding, but just showing existing items. Adding would be if you pushed to some array new item with exact same type as before

N1XUS:
Agree with @g-cheishvili. This example is not something I would consider as "adding", rather just conditionally "showing".

N1XUS (changes requested):

mikerodonnell89:
> I think this is not adding, but just showing existing items. Adding would be if you pushed to some array new item with exact same type as before

That's correct, but from the user's perspective, same thing I think. I'll need to speak to the client if this satisfies their requirements

droshev:
They would need to add fields on the fly.

droshev:
We have the features being implemented [here](https://github.com/SAP/fundamental-ngx/pull/10368)

Issue #6047, opened 2021-07-20
extend dialog examples


droshev:
#### Is this a bug, enhancement, or feature request?
documentation enhancement

#### Briefly describe your proposal.

Create a dialog example 
- where the content can swipe between 2(or more `pages`)
- provide back button on top so it can give a cue to the user she is in a second page

dkrishnan-01:
@droshev . here are the screens:
[Screen 1 ](https://www.figma.com/file/izfDuxvyGDWwpCsLUuoi6C/CJ-Experiences-Preview-Builder?node-id=714%3A15246)- when modal pops ups it will have content that looks like this.

[Screen 2](https://www.figma.com/file/izfDuxvyGDWwpCsLUuoi6C/CJ-Experiences-Preview-Builder?node-id=714%3A15246) - when the user clicks on a Link it will take them to the next page with a back arrow on the top left of the modal header

droshev:
@hanjunt do you think you can explore developing those examples as part of the documentation?

PR #10343, opened 2023-08-14
docs(core): added swipe dialog example

IB002:
## Related Issue(s)

https://github.com/SAP/fundamental-ngx/issues/6047

## Description

Adds a swipe dialog example


mikerodonnell89 (commented):
It looks like lint was skipped for this, I'd recommend resetting the commit and attempting again without `--no-verify`

droshev (changes requested):
I don't see swiping here
not-swiping

Also can you make the first and second level list to be with the same size? The different sizes brings even more confusion.

g-cheishvili:
I do not think that "swipe" is the term we are looking for here. It is more of an multi page dialog, similar to what we have in mobile menu with children

droshev (changes requested):
I still don't see the swiping effect on the item clicking.

droshev:
It can be reopened when there is someone who can bring the swipping

PR #10350, opened 2023-08-16
chore: documenting technical parts of the library

g-cheishvili:
## Related Issue(s)

closes

## Description

Documented few important key moments in our automation, structure and plugin code


mikerodonnell89 (approved):
very helpful, thanks @g-cheishvili !

CREATE_RELEASE.md
```diff
@@ -25,24 +25,42 @@ Or if you want to use the conventional commits to determine next version automat
...
+include all the changes from the 0.40.1 release to the latest RC release, which is 0.40.1-rc.\*. In this case you need
+to do the following:
+
+-   `git checkout v0.40.0`
+-   `git cherry-pick <commit-hash>` or make changes you want to make and commit them
+-   `npm run hotfix-release`
+
+After this, lerna determines that the next version should be `0.40.1`, and CI will create a release from the HEAD.
+CI also finds out that the `0.40.1` is newer than the latest RC release, which is `0.40.1-rc.*`, and updates the main
+branch version to `0.40.1`. Also it must be noted that the npm will mark this release as the `latest`.
```
mikerodonnell89, line 58: I think this isn't always true, not the case for the ng15 hotfixes I've been doing

mikerodonnell89 (commented):
see comment

CREATE_RELEASE.md
g-cheishvili: Basically after lerna step, it compares whether the given result is bigger than the latest `latest` and chooses bigger version for the `latest` candidate

VERSIONING.md
```diff
@@ -0,0 +1,24 @@
+# Versioning in fundamental-ngx
```
droshev, line 1: This can be shortened and added to [`CONTRIBUTING.md`](https://github.com/SAP/fundamental-ngx/blob/main/CONTRIBUTING.md)

droshev:
@mikerodonnell89 @InnaAtanasova @g-cheishvili shall we keep those `.md` files in the root or we put them in a subfolder?

g-cheishvili:
> @mikerodonnell89 @InnaAtanasova @g-cheishvili shall we keep those `.md` files in the root or we put them in a subfolder?

I think they are relatively small for now and it's ok to keep them in root, but yeah, we could move them into `docs` or something like that

InnaAtanasova (approved):

PR #10409, opened 2023-08-28
feat(docs): examples standalone migration

N1XUS:
closes none;

- Migrated `@fundamental-ngx/docs` to standalone components;
- Refactored Stackblitz example service to reflect new project structure and initialisation;
- Fixed issue with always loading old version of fundamental-ngx libraries in Stackblitz examples;
- Stackblitz examples now use Horizon theme by defaultж
- Added prettier plugin to organize imports


libs/docs/cdk/breakpoint/breakpoint-header/breakpoint-header.component.ts
```diff
@@ -1,7 +1,14 @@
 import { Component } from '@angular/core';
+import { RouterOutlet } from '@angular/router';
+import { HeaderTabsComponent } from '../../../shared/src/lib/core-helpers/header-tabs/header-tabs.component';
+import { ImportComponent } from '../../../shared/src/lib/core-helpers/import/import.component';
+import { DescriptionComponent } from '../../../shared/src/lib/core-helpers/description/description';
+import { HeaderComponent } from '../../../shared/src/lib/core-helpers/header/header.component';
```
g-cheishvili, line 6: Paths to shared is not correct

g-cheishvili (changes requested):
we should have `@fundamental-ngx/docs/shared` instead of relative calls

g-cheishvili:
> I believe that there is no need for the `*Module` stuff anymore in `*-docs.module.ts` files anymore. Things like `CommonModule` or `SharedDocumentationPageModule` and etc

Actually, none of the things are needed in it at all. Only thing needed is the RouterModule.routes, rest is standalone, so no need for them in the modules.

Just SharedDocumentationPageModule also provides CopyService and the ApiService and those two should be moved to the root route's providers array, along with the currentComponentProvider

g-cheishvili (approved):

PR #10415, opened 2023-08-29
feat(core): multi input option item

### LINKED ISSUE #10341: [Sourcing][P2]: MutliInput - allow to work directly with OptionItem
enhancement

Even internally MultiInput is using and converting everything into `OptionItem`, but it does not consider if `OptionItem` is passed and used as type application use. If we do, then we dont need to provide `valueFn` or the `displayFn` and everything will be streamlined + we get some performance gain, if components does not need to call those callback functions

It just takes to check if item is already `OptionItem` if it is we dont need to do any conversion. 

```
private _getOptionItem(item: any): OptionItem {
        const { label, value } = this._getValueAndLabel(item);
        return {
            item,
            label,
            value,
            isSelected: false
        };
    }
```

g-cheishvili:
## Related Issue(s)

closes #10341

## Description

Multi-input now allows a OptionItem interfaces as dropdown values and also typings have been improved slightly

### DIFF
--- a/libs/cdk/src/lib/utils/directives/auto-complete/auto-complete.directive.ts
+++ b/libs/cdk/src/lib/utils/directives/auto-complete/auto-complete.directive.ts
@@ -27,6 +27,10 @@ export class AutoCompleteDirective {
     @Input()
     enable = true;
 
+    /** Matcher function for testing the str for a search term */
+    @Input()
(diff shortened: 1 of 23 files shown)


fkolar (approved):
Thanks for adding in.

PR #10419, opened 2023-08-29
fix(platform): console error when opening platform select

mikerodonnell89:
fixes a console error when opening platform select

### DIFF
--- a/libs/platform/src/lib/form/select/select/select.component.ts
+++ b/libs/platform/src/lib/form/select/select/select.component.ts
@@ -129,10 +129,12 @@ export class SelectComponent extends BaseSelect implements AfterViewInit, AfterV
 
     /** @hidden */
     _onOpenChange(isOpen: boolean): void {
-        if (isOpen) {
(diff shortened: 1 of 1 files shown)


libs/platform/src/lib/form/select/select/select.component.ts
```diff
@@ -129,10 +129,12 @@ export class SelectComponent extends BaseSelect implements AfterViewInit, AfterV
...
     _onOpenChange(isOpen: boolean): void {
-        if (isOpen) {
-            this.formMessage._popover.close();
+        if (this.formMessage) {
+            if (isOpen) {
+                this.formMessage._popover.close();
+            }
+            this.formMessage._popover.setIgnoreTriggers(isOpen);
         }
-        this.formMessage._popover.setIgnoreTriggers(isOpen);
```
N1XUS, line 135: you can just use optional chaining (`?.`) to reduce folding

N1XUS (approved):

PR #10434, opened 2023-08-30
fix(platform): table header cells without popovers should have aria-haspopup false

### LINKED ISSUE #10284: [P1][Sourcing][Accessibility][Table][a11y help] The column headers are marked with aria-haspopup. But no menu can be opened over them. They also refer to an empty element via aria-controls.
Bug

The column headers are marked with aria-haspopup. But no menu can be opened over them. They also refer to an empty element via aria-controls.

This case is not applicable for table which dont have sort , filter etc, hence the aria property has to be updated for such cases.
And enabled only when sorting is true for those columns.

Screenshot 2023-08-03 at 9 38 24 PM

Reported by customer

mikerodonnell89:
fixes #10284

### DIFF
--- a/libs/platform/src/lib/table/components/table-header-row/table-header-row.component.html
+++ b/libs/platform/src/lib/table/components/table-header-row/table-header-row.component.html
@@ -66,7 +66,7 @@
             column._endFreezed ? 'fd-table__cell--fixed-end' : '',
             column.name === freezeEndColumnsTo ? 'fd-table__cell--fixed-end-last' : '',
             isLast && (_fdpTableService._isShownNavigationColumn$ | async) !== true ? 'is-last-child' : '',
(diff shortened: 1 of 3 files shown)


libs/platform/src/lib/table/components/table-header-row/table-header-row.component.html
```diff
@@ -80,7 +80,7 @@
                     : _tableColumnResizeService.getColumnWidthStyle(column.name)
                     : _tableColumnResizeService.getNextColumnsWidth(column.name)
         "
-        [fdPopoverTrigger]="tablePopover.popover"
+        [fdPopoverTrigger]="_isColumnHasHeaderMenu(column) ? tablePopover.popover : null"
```
N1XUS, line 83: avoid using functions in template. Better to introduce pipe (possibly `pure` will work too)

Issue #10197, opened 2023-07-20
Implement Generic Tag component


InnaAtanasova:
#### Is this a bug, enhancement, or feature request?
feature request

#### Briefly describe your proposal.
 Generic Tag is a component that exists in Fund-styles but not in Fund-ngx.

PR #10435, opened 2023-08-30
feat(core): introduce Generic Tag new component

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/10197

## Description
adds new component to Core lib


libs/core/src/lib/generic-tag/generic-tag.component.ts
```diff
@@ -0,0 +1,100 @@
...
+import { FD_GENERIC_TAG_COMPONENT } from './tokens';
+
+export type GenericTagType = 'error' | 'success' | 'warning' | 'information';
+
+@Component({
+    // eslint-disable-next-line @angular-eslint/component-selector
+    selector: '[fd-generic-tag]',
+    templateUrl: './generic-tag.component.html',
+    styleUrls: ['./generic-tag.component.scss'],
+    encapsulation: ViewEncapsulation.None,
```
N1XUS, line 20: Please make this component standalone

libs/core/src/lib/generic-tag/generic-tag.component.ts
```diff
@@ -0,0 +1,100 @@
...
+    value: Nullable<string>;
+
+    /**
+     *  Aria defines role description for the Generic Tag
+     */
+    @Input()
+    ariaRoleDescription: Nullable<string> = 'Generic Tag';
+
+    /** @hidden */
+    constructor(public readonly elementRef: ElementRef) {}
```
N1XUS, line 68: You can use `inject` function instead of DI in constructor. This way it will be easier if some component would inherit this class as a base one.

g-cheishvili (approved):

PR #10456, opened 2023-09-01
fix(platform): fixed multi combobox async value selection token

### LINKED ISSUE #10377: [P1][Sourcing]MultiComboBox Component not getting tokenized for selected data.
Bug

We've have certain use case where we provide the values for MultiComboBox from backend as the user types into the search input.

To achieve this, we wrote a custom class that inherits the `BaseDataProvider` and overrides the `fetch` function. So when the user types in the input field, this function is invoked to get the data from server thru API if the data was not found in the values passed during the MultiComboBoxDataSource initialization in constructor.

_**The issue is that when the user selects the value passed from `fetch` function, the selection is not getting tokenized and the value is not getting emitted in the `selectionChange` Emitter**_

https://github.com/SAP/fundamental-ngx/assets/136386704/3c537d05-3b96-461c-aa82-fcbf12243fa9

As in the above recording, values like "Apple", "Pineapple" were passed in the initial list to the MultiComboBox whereas the values like "Turkey", "Tesla" were passed as the user types in the search from `fetch` function which are not getting tokenized.

Latest

1. Create a component to use MultiComboBox and have dataset of 10 values of type Object.
2. Have a custom class that inherits the BaseDataProvider and pass only first 5 values in the constructor.
3. Override the `fetch` function, that will return the remaining values as the use types in search. This is to mimic server call.
4. Search for the value from the remaining dataset and select the value. You'll see it will not get tokenized as shown in the above recording.   

You can also just clone the following repo which has the above setup.
https://github.com/vidworks/table-destroy

https://github.com/vidworks/table-destroy/tree/main/src/app/multi-combo-example

http://localhost:4200/multi-combo

This is an issue for both Core and Platform MultiComboBox. In fact Platform is more stable than Core.

g-cheishvili:
## Related Issue(s)

closes #10377

## Description

This should stitch it. Problem is that it never updates the full list of the items, it needs more attention in nearest future, but for now it should work

### DIFF
--- a/libs/platform/src/lib/form/multi-combobox/multi-combobox/multi-combobox.component.html
+++ b/libs/platform/src/lib/form/multi-combobox/multi-combobox/multi-combobox.component.html
@@ -151,16 +151,15 @@
 <ng-template #listTemplate>
     <ul
         fd-list
-        [hasMessage]="!!stateMessage"
         (focusEscapeList)="handleListFocusEscape($event)"
         [dropdownMode]="true"
         class="fdp-multi-combobox__list fd-list--multi-input"
         [hasMessage]="!!stateMessage || (!!advancedStateMessage?.hasErrors && !!advancedStateMessage?.template)"
         [id]="id + '-result'"
         role="listbox"
-        [style.maxHeight]="!mobile && maxHeight"
-        [style.minWidth]="!mobile && minWidth + 'px'"
-        [style.maxWidth]="autoResize && maxWidth + 'px'"
+        [style.max-height]="!mobile && maxHeight"
+        [style.min-width]="!mobile && minWidth + 'px'"
+        [style.max-width]="autoResize && maxWidth + 'px'"
         [attr.aria-labelledby]="id + '-search'"
         aria-multiselectable="true"
     >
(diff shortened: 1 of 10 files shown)


libs/platform/src/lib/form/multi-combobox/multi-combobox/multi-combobox.component.html
```diff
@@ -151,16 +151,15 @@
...
         [dropdownMode]="true"
         class="fdp-multi-combobox__list fd-list--multi-input"
         [hasMessage]="!!stateMessage || (!!advancedStateMessage?.hasErrors && !!advancedStateMessage?.template)"
         [id]="id + '-result'"
         role="listbox"
-        [style.maxHeight]="!mobile && maxHeight"
-        [style.minWidth]="!mobile && minWidth + 'px'"
-        [style.maxWidth]="autoResize && maxWidth + 'px'"
+        [style.max-height]="!mobile && maxHeight"
+        [style.min-width]="!mobile && minWidth + 'px'"
```
N1XUS, line 161: We can also do `[style.min-width.px]` right?

N1XUS (approved):

PR #10474, opened 2023-09-04
feat(core): added template based popover body

### LINKED ISSUE #10392: [P2][Sourcing][Sep18] `fd-popover` loads content from body even when it is closed

enhancement

It can be reproduced with one or more `fd-popover`. In initial load, even when it is closed, the content in `fd-popover-body` is still loaded. To reproduce this issue, I have multiple `fd-popover` and one customized component with two lifecycle hooks to showcase the issue:

https://github.com/SAP/fundamental-ngx/assets/22842279/049a4ef8-6221-47f7-bae9-86d40326c3dd

As observed here from log, all the customized components are initialized for initial load. Not sure if this is expected behavior, but from what I think, the content should be loaded only when we open the popover. It can be further observed with the following fix on the `datepicker` issue in https://github.com/SAP/fundamental-ngx/issues/10242 since `datepicker` is also using `fd-popover`.

This will cause major perf issue in place we have many usage of popover, especially affecting `fdp-combobox`, `fdp-select`, or `fdp-multi-combobox`...

latest

- Create a fd-popover with two-way bindings [(isOpen)]
- Set isOpen initially closed.
- Create a customized component having `ngOnInit` to show log.
- Add multiple `fd-popover`
- Loading browser and observe the logs

https://stackblitz.com/edit/angular-39lqgd?file=src%2Fapp%2Fpopover-programmatic-open-example.component.html

g-cheishvili:
## Related Issue(s)

closes #10392

## Description

Downport of the #10455


g-cheishvili:
Github actions are glitching, but these changes have passed successfully on main branch

PR #10483, opened 2023-09-05
fix(platform): use pipe for table-row-header popup evaluation

mikerodonnell89:
fixes none. PR comment from @N1XUS that did not make it in to the original PR to main that was merged


libs/platform/src/lib/table/components/table-header-row/table-header-row.component.ts
```diff
@@ -141,9 +143,16 @@ export class TableHeaderRowComponent extends TableRowDirective implements OnInit
     _columnTrackBy(index: number, column: TableColumn): string {
         return column.name;
     }
+}
+
+/** @hidden */
+@Pipe({ name: 'isColumnHasHeaderMenu' })
+export class IsColumnHasHeaderMenuPipe implements PipeTransform {
```
g-cheishvili, line 150: This sounds weird :D

N1XUS (approved):

libs/platform/src/lib/table/components/table-header-row/table-header-row.component.ts
mikerodonnell89: I agree, it was the name of the previous function

PR #10484, opened 2023-09-05
fix(platform): hiding selection column when data is not visible

### LINKED ISSUE #10201: [P2][Sourcing] Header checkbox can still be selectable if NGX package imported in child module

bug

Following the fix from https://github.com/SAP/fundamental-ngx/issues/10134, if we have NGX package imported in child module, when table is empty or has no visible columns, the header checkbox is still selectable.

Screenshot 2023-07-21 at 12 35 55 PM

Expectation: it should work even the NGX packages are imported in child module.

Baseline: https://fundamental-ngx.netlify.app/#/platform/table/scrolling when we search to return empty data.

latest

- Have NGX core/platform packages imported in child module.
- Create platform table in multi selectionMode
- Search or set columns so that no data is displayed.
- Try to se/delect header checkbox

https://stackblitz.com/edit/angular-sufq3o-jvuhry?file=src%2Fapp%2Fplatform-table-page-scrolling-example.component.ts

g-cheishvili:
## Related Issue(s)

closes #10201

## Description

Downport of the #10428

### DIFF
--- a/libs/platform/src/lib/table/table.component.ts
+++ b/libs/platform/src/lib/table/table.component.ts
@@ -339,7 +339,7 @@ export class TableComponent<T = any>
     @Input()
     set selectionMode(value: SelectionModeValue) {
         this._selectionMode = value;
-        this._isShownSelectionColumn = this.selectionMode !== SelectionMode.NONE;
+        this._isSelectionColumnShown = this.selectionMode !== SelectionMode.NONE;
         this._setSelectionColumnWidth();
     }
 
@@ -553,6 +553,7 @@ export class TableComponent<T = any>
      * Contains all rows including group rows.
      */
     _tableRows: TableRow<T>[] = [];
+
     /**
      * @hidden
      * Representation of table rows that came from dataSource.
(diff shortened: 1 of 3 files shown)


libs/platform/src/lib/table/table.component.ts
```diff
@@ -539,6 +539,19 @@ export class TableComponent<T = any>
         );
     }
 
+    /**
+     *  Whether the selection column should be shown,
+     *  If there are no visible rows or visible columns
+     *  or selection mode is none, then selection column should not be shown.
+     **/
+    get isSelectionColumnShown(): boolean {
```
mikerodonnell89, line 547: Personally I'd rather use a function call in the template than a getter because calling a getter in a template appears to be simply accessing a value, but is actually calling a function anyways, so it'll be harder to catch as a potential performance issue. Leaving a function call in a template will be easily to identify for any performance issues in the future

g-cheishvili: This thing is OnPush, this function will be called only on change detection cycles

g-cheishvili: I get your point, but I do not see function benefiting us at the moment. With the ng16, signal instead of this will be on point

g-cheishvili: **Moved to the function**

PR #10503, opened 2023-09-07
fix(core): vertical nav keyboard navigation

N1XUS:
## Related Issue(s)

closes #10388

## Description
- Fixed issue with focusing button instead of whole vertical nav item;
- Fixed issue with focus outline of the parent expanded item;
- Added role="presentation" for icon and button

### Before:
image

### After:
image

PR #10504, opened 2023-09-07
fix(core): vertical nav keyboard navigation

N1XUS:
Downport of https://github.com/SAP/fundamental-ngx/pull/10503


mikerodonnell89 (changes requested):
This is an improvement when the cell is collapsed, however, a bit confusing when focusing the parent, as it reads the children as well then says expanded after reading the children. Not sure if that's the intent. Probably shouldn't read the children at all? 


Screenshot 2023-09-07 at 12 23 08 PM

Also, when focusing cells without children, "collapsed" is still read, shouldn't mention collapsed or expanded in that case

g-cheishvili (approved):

Issue #10295, opened 2023-08-03
[P2][Sourcing][Accessibility][Multi input tokens]The horizontal alignment of the list is not perceptible with the screen reader. Selection lists are usually vertically oriented and are operated with the up and down arrow keys. Here, however, only right and left arrow keys work, which is not perceptible with JAWS


I543348:
#### Is this a bug, enhancement, or feature request?
Bug

#### Briefly describe your proposal.
The horizontal alignment of the list is not perceptible with the screen reader. Selection lists are usually vertically oriented and are operated with the up and down arrow keys. Here, however, only right and left arrow keys work, which is not perceptible with JAWS

#### If this is a bug, please provide steps for reproducing it.
The horizontal alignment of the list is not perceptible with the screen reader. Selection lists are usually vertically oriented and are operated with the up and down arrow keys. Here, however, only right and left arrow keys work, which is not perceptible with JAWS
tested on multi input mobile mode after selecting few tokens

#### Is there anything else we should know?
reported by customers

droshev:
@I543348 Can you provide more context and screenshots

droshev:
@I543348 The agreement is that you provide us with all the input so we can work on the issue.

nikolay-kolarov:
Hi,

I found a gap in the implementation which I think is the issue for with the customer complains.
When you navigate in the tokenizer in the multiinput, you should be able to use [arrow left] and [arrow up] to go to the previous token on the left. But in the MultiInput implementation only [arrow left] works and [arrow up] goes in the list, which is wrong.
You can have a look at the UI5 implementation as a reference: https://ui5.sap.com/#/entity/sap.m.MultiInput/sample/sap.m.sample.MultiInput

Best Regards,
Nikolay

fkolar:
@I543348, I will keep the WIP, until we are 100% sure and until we update this with some screenshot or some more details. Unless you can say, that you agree @nikolay-kolarov and you guys are talking about the same problem. In this case please remove WIP

sagar7259:
Hello,

On accessing the tokens of the multi value input, the reading is made has
"list with 7 items
Blueberry Deletable token
7 of 7
To move to an item press the Arrow keys."

Here to move to an next token, tutor message is read as to use the arrow keys, but its failed to conveyed which arrow keys to be used, weather is a up/down arrow key or right/left arrow key.

When this issue is compared with UI5 component, here in UI5 demo kit the same tokens can be accessed using any arrow keys.

[...]

PR #10534, opened 2023-09-11
feat(core): tokenizer vertical arrow keys navigation

g-cheishvili:
## Related Issue(s)

Part of #10295

## Description

Added support for navigation between tokens using vertical arrow keys

### DIFF
--- a/libs/core/src/lib/token/tokenizer.component.ts
+++ b/libs/core/src/lib/token/tokenizer.component.ts
@@ -433,10 +433,8 @@ export class TokenizerComponent implements AfterViewInit, OnDestroy, CssClassBui
      * Check whether user intends to go back to the previous token
      **/
     private _goBackRequested($event: KeyboardEvent, rtl: boolean): boolean {
-        if (!rtl) {
-            return KeyUtil.isKeyCode($event, LEFT_ARROW) || KeyUtil.isKeyCode($event, UP_ARROW);
-        }
-        return KeyUtil.isKeyCode($event, RIGHT_ARROW) || KeyUtil.isKeyCode($event, DOWN_ARROW);
+        const backKeys = rtl ? [RIGHT_ARROW, DOWN_ARROW] : [LEFT_ARROW, UP_ARROW];
+        return KeyUtil.isKeyCode($event, backKeys);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -419,6 +428,26 @@ export class TokenizerComponent implements AfterViewInit, OnDestroy, CssClassBui
         return totalTokenWidth;
     }
 
+    /**
+     * Check whether user intends to go back to the previous token
+     **/
+    private _goBackRequested($event: KeyboardEvent, rtl: boolean): boolean {
```
N1XUS, line 434: `@hidden`

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -419,6 +428,26 @@ export class TokenizerComponent implements AfterViewInit, OnDestroy, CssClassBui
...
+        if (!rtl) {
+            return KeyUtil.isKeyCode($event, LEFT_ARROW) || KeyUtil.isKeyCode($event, UP_ARROW);
+        }
+        return KeyUtil.isKeyCode($event, RIGHT_ARROW) || KeyUtil.isKeyCode($event, DOWN_ARROW);
+    }
+
+    /**
+     * Check whether user intends to go forward to the next token
+     **/
+    private _goForwardRequested($event: KeyboardEvent, rtl: boolean): boolean {
```
N1XUS, line 444: `@hidden`

libs/core/src/lib/token/tokenizer.component.ts
```diff
@@ -419,6 +428,28 @@ export class TokenizerComponent implements AfterViewInit, OnDestroy, CssClassBui
         return totalTokenWidth;
     }
 
+    /**
+     * @hidden
+     * Check whether user intends to go back to the previous token
+     **/
+    private _goBackRequested($event: KeyboardEvent, rtl: boolean): boolean {
+        if (!rtl) {
```
N1XUS, line 436: Can be simplified like:
```typescript
const isBackButton = isRtl ? [RIGHT_ARROW, DOWN_ARROW] : [LEFT_ARROW, UP_ARROW];

return KeyUtil.isKeyCode($event, isBackButton);
```

N1XUS: Same logic can be applied for `_goForwardRequested`

N1XUS: And maybe rename methods to `_isGoBackRequested` and `_isGoForwardRequested` since these methods check on whether the action being requested and not fire any action by itself.

N1XUS (approved):

PR #10549, opened 2023-09-14
fix(platform): table initial selection hydration

### LINKED ISSUE #10454: [Sourcing][P2] Apply Tristate also during table initialization, not only during check/uncheck
 bug

Currently, when we initialize a TREE Table with `TableRow` data type and set of the children `checked`, it is not propagated all the way up to parent, when` [enableTristate] is true`. 

It only works during `_toggleMultiSelectRow` / `_toggleSingleSelectRow`

latest

We could do: 

1. There is a method `_applyTristateSelection` called only during toggle row action

```
toggleMultiSelectRow(rowToToggle: TableRow, event?: Event): void {
     ..
        this._rangeSelector.applyValueToEachInRange((idx) => {
            const row = rows[idx];
            if (row.type === TableRowType.ITEM || row.isTree) {
               ..
                this._applyTristateSelection(row, added, removed);
            }
        });
```

2. we need to call the same method even during `_setTableRows()`
 - something like `this.initializeTristate() `
 - if tristate is enabled, then go recursively over `this._tableRows ` items and when `checked row` is found, call  
`this._applyTristateSelection(checkRow, [], []);
`

g-cheishvili:
## Related Issue(s)

closes #10454

## Description

Downport of the #10544

### DIFF
--- a/libs/cdk/src/lib/utils/helpers/dfs.ts
+++ b/libs/cdk/src/lib/utils/helpers/dfs.ts
@@ -1,0 +1,59 @@
+import { get } from 'lodash-es';
+
+/** @hidden */
+interface DfsOptions<T> {
+    children: string | ((item: T) => Array<T>);
+    shouldVisit: (item: T) => boolean;
+    postOrder: boolean;
+}
+
+/** @hidden */
+const _defaultDfsOptions: DfsOptions<any> = {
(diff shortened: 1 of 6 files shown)


g-cheishvili:
Main PR is approved

Issue #10137, opened 2023-07-07
[P1][Sourcing][translation] `f-ngx/i18n` missing some lang locales


yixuan-liu:
#### Is this a bug, enhancement, or feature request?
Enhancement

#### Briefly describe your proposal.
`F-NGX/i18n` is missing some locales that Guided Sourcing currently supports. We will need complete i18n locale support from `F-NGX/i18n` to cover all of our supported languages/locales. 

Our complete locale list:
Languages supported by Guided Sourcing

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/I036737/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/I036737/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

Lang   (SS) | Lang   (SS) | Lang   Description(SS) | F-NGx   supported
-- | -- | -- | --
bg | bg_BG | Bulgarian |  v1 |
cs | cs_CZ | Czech |  v1 |
da | da_DK | Danish |  |
de | de_DE | German | ✅ | 
el | el_GR | Greek |  |
en | en_US | English | ✅ |  
es | es_ES | Spanish |  |
fi | fi_FI | Finnish |  |
fr | fr_FR | French |  v1 |
hu | hu_HU | Hungarian |  |
hr | hr_HR | Croatian |  |
it | it_IT | Italian |  v1 |
ja | ja_JP | Japanese | |  
ko | ko_KR | Korean |  |
nl | nl_NL | Dutch |  |
no | no_NO | Norwegian | | 
pl | pl_PL | Polish |  v1 |
pt | pt_BR | Brazilian Portuguese | ✅ |
ro | ro_RO | Romanian |  |
ru | ru_RU | Russian |  v1 |
sv | sv_SE | Swedish |  |
tr | tr_TR | Turkish |  v1 |
zh_CN | zh_CN | Simplified Chinese |  tbd |
[...]

PR #10571, opened 2023-09-19
feat(i18n): using json formats as a source for tew and added resolve utilities

g-cheishvili:
## Related Issue(s)

Part of #10137

## Description

Using json language files as sources now.
Added helper functions to make translations easier now
Added and updated i18n documentation sections

BREAKING CHANGE:
Placeholders should be wrapped inside one pair of the curly braces from now on.
`{{ something }}` with context `{something: 'value'}` would become "value" before, now the same output will require `{ something }` to be the input
Also `{{{ something }}}` would result in `{value}`, now it will throw, to have that same output, the user will have to type `'{'{something}'}'`


libs/docs/i18n/writing-translations.component.ts
```diff
@@ -0,0 +1,141 @@
...
+    CodeSnippetComponent,
+    DescriptionComponent,
+    DocsSectionTitleComponent,
+    ExampleFile,
+    HeaderComponent
+} from '@fundamental-ngx/docs/shared';
+
+@Component({
+    template: `
+        <header>Writing translations</header>
```
N1XUS, line 16: Maybe move to separate file for better readability?

libs/i18n/src/lib/utils/load-json.ts
```diff
@@ -0,0 +1,44 @@
+import { FdLanguage, FdLanguageKeyIdentifier, FlatFdLanguage } from '../models/lang';
+
+interface RecursiveRecord<V = string> {
+    [key: string]: V | RecursiveRecord<V>;
+}
+
+function set<V = string>(obj: RecursiveRecord<V>, path: string, value: V): void {
```
N1XUS, line 7: Is there a difference with lodash's `set` function?

libs/i18n/src/lib/utils/translation-resolver.ts
```diff
@@ -1,54 +1,103 @@
...
+            return expression;
+        }
+        const result = new IntlMessageFormat(expression, locale || 'en-US').format(args || {});
+        return Array.isArray(result) ? result.join('') : result.toString();
+    }
+
+    /**
+     * Returns the raw ICU string for the provided language by key and args
+     **/
+    private getRaw(lang: FdLanguage, key: FdLanguageKeyIdentifier, args?: FdLanguageKeyArgs): string {
```
N1XUS, line 67: `_` prefix ;)

N1XUS (approved):
Awesome job! Just couple of minor comments

PR #10576, opened 2023-09-21
fix: fd tab list tabindex accessibility

ZsDeak-SAP:
## Description

- Only interactive elements should have a tab index. Tab content is not guaranteed to be an interactive element, so the forced default tabindex is the wrong a11y, causing us to fail the audit.
- No guide on how to get started, so it's easy to miss things, like the project uses yarn, getting version conflicts on `npm i`
- Top level package version is redundant for monorepos such as this (it being outdated for months proves that point, but also recommended to be removed [here](https://github.com/lerna/lerna/issues/2879#issuecomment-1471606448))

### Before:
tabs_before

### After:
tabs_after


g-cheishvili:
You're updating quite a few different topics.
- private: true - I agree, we should move to that
- locally installing the dependencies with `npm` will always result in failure, that is ensured by `"preinstall": "npx only-allow yarn"`
- overrideTabindex - Ok, lets make it false by default

All the topics need their PRs and the PR titles should be in a conventional commits format

ZsDeak-SAP:
> locally installing the dependencies with npm will always result in failure, that is ensured by "preinstall": "npx only-allow yarn"

only allow yarn will only work if you have already run `yarn` at least once, that install `only-allow` as well. a brand new clone will try to install with npm and fail with version problems (i've just had this issue myself that's how i know 😆 )

commits are in conventional commit form, why do you need a separate PR per modified line? lerna can handle different commit types in the same PR afaik, and updates according to the highest effect.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fundamental-ngx",
-  "version": "0.43.0-rc.4",
+  "private": true,
```
droshev, line 3: why did you have to change that?

ZsDeak-SAP: Top level package version is redundant for monorepos such as this (it being outdated for months proves that point, but also recommended to be removed https://github.com/lerna/lerna/issues/2879#issuecomment-1471606448)

g-cheishvili:
> > locally installing the dependencies with npm will always result in failure, that is ensured by "preinstall": "npx only-allow yarn"
> 
> only allow yarn will only work if you have already run `yarn` at least once, that install `only-allow` as well. a brand new clone will try to install with npm and fail with version problems (i've just had this issue myself that's how i know 😆 )

Strange, never had that problem, but in any case maybe it is better to add `"npm": "please use yarn instead of npm"` in engines of package.json? We have `strict-engine` set to true in .npmrc, so it should fail then

ZsDeak-SAP:
I did some digging, since it is indeed weird, turns out this `strict-engine` is not exactly what we were led to believe it was [link](http://www.marcusoft.net/2015/03/packagejson-and-engines-and-enginestrict.html) \
tried `"npm": 'please dont'` as well, did not stop my npm commands i think for the same reason 😕  \
also preinstall isn't perfect (it still doesn't work for `npm install` although i already have allow-only), because the install can fail at building the [ideal tree](https://github.com/npm/cli/blob/650021838ad0fe565f925b47dfd2ea65beefa4c5/lib/commands/install.js#L151) which happens before any of the scripts. My `npm install` also failed with `ERESOLVE` \
image

ZsDeak-SAP:
the package.json version commit is now obsolete as was merged in #10633

droshev:
#10671 and #10672 are addressing those changes.

ZsDeak-SAP:
i appreciate that you guys take the time to create all those branches and PRs so my small fixes can be merged, but don't you think it would be better if this open source lib would allow for actual outside contribution?
@droshev

PR #10578, opened 2023-09-21
feat(core): complete standalone migration

N1XUS:
Migrated all components and directives to standalone


libs/core/src/lib/breadcrumb/breadcrumb.module.ts
```diff
@@ -1,18 +1,18 @@
...
-import { PopoverModule } from '@fundamental-ngx/core/popover';
-import { MenuModule } from '@fundamental-ngx/core/menu';
-import { LinkModule } from '@fundamental-ngx/core/link';
 import { PortalModule } from '@angular/cdk/portal';
 import { PipeModule } from '@fundamental-ngx/cdk/utils';
 import { ContentDensityModule } from '@fundamental-ngx/core/content-density';
+import { IconModule } from '@fundamental-ngx/core/icon';
+import { LinkModule } from '@fundamental-ngx/core/link';
+import { MenuModule } from '@fundamental-ngx/core/menu';
+import { PopoverModule } from '@fundamental-ngx/core/popover';
```
g-cheishvili, line 12: still modules

g-cheishvili (approved):

N1XUS:
> This should be a BREAKING CHANGE. libs/core/src/lib/combobox/index.ts does note export mobile module anymore. That's one thing I noticed, but if there are more, those should be stated as well

Mobile combobox module was used in a specific use case, and not by the users. It was exported mostly to not break the lib itself.

PR #10601, opened 2023-09-26
feat(core): multi input value identifier

### LINKED ISSUE #10341: [Sourcing][P2]: MutliInput - allow to work directly with OptionItem

### LINKED ISSUE #10378: Observable Async - Multi Input field value is not displayed properly
Bug
In Observable Async - Multi Input field, 

- on searching for an option by typing in keyboard the value appears in the drop down and after selection, the selected option is shown as well as the text I searched is also shown in the multi-input field. Only the selection option should be shown in this case.
image

- on searching for an option by typing in keyboard the value doesn't show properly, sometime I can see the valueFn content is shown in the field instead of displayFn content. Only the content in the displayFn should be shown.
image

0.46.0-rc5

https://fundamental-ngx.netlify.app/#/core/multi-input#observable-async
Use keyboard for searching the options.

https://fundamental-ngx.netlify.app/#/core/multi-input#observable-async

g-cheishvili:
## Related Issue(s)

closes #10341
closes #10378

## Description

This adds ability to have `identifier` on OptionItem. Also adds more type-safety to the multi-input

### DIFF
--- a/libs/cdk/src/lib/utils/datatypes/index.ts
+++ b/libs/cdk/src/lib/utils/datatypes/index.ts
@@ -1,3 +1,4 @@
-export * from './hash.datatype';
-export * from './size.datatype';
 export * from './color-accent.datatype';
+export * from './hash.datatype';
+export * from './object-nesting.types';
+export * from './size.datatype';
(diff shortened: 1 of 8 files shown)


fkolar (approved):
Looks good. since we discussed this before ..

fkolar:
We need to get it in but updated docs site woudl be good as well.

InnaAtanasova (approved):

PR #10634, opened 2023-09-29
fix(core): some improvements to using labels with block busy indicator

### LINKED ISSUE #10412: [Sourcing][Accessibility][Busy indicator] [P3]The label's color contrast doesnt meet standard ratio
Bug

The label of busy indicator should be visible properly to the user , just like indicator.
But the color of label is too light and unnoticeable, looks like disabled state
Expectation : the label should be near to the busy indicator animation with a little darker color , at the center of the the content

Angular 15

https://stackblitz.com/edit/angular-5l37ya?file=src%2Fapp%2Fbusy-indicator-wrapper-example.component.html

Screenshot 2023-08-29 at 11 39 09 AM

https://stackblitz.com/edit/angular-5l37ya?file=src%2Fapp%2Fbusy-indicator-wrapper-example.component.html

mikerodonnell89:
fixes #10412 

before:
Screenshot 2023-09-29 at 12 23 19 PM

after:
Screenshot 2023-09-29 at 12 22 50 PM


g-cheishvili (changes requested):
Why involve JS into the equasion at all? `* + .fs-busy-indicator` will do the same trick

libs/core/src/lib/busy-indicator/busy-indicator.component.scss
```diff
@@ -1,21 +1,25 @@
 @import 'fundamental-styles/dist/busy-indicator';
 
-.fd-busy-indicator {
+* + .fd-busy-indicator,
+* + .fd-busy-indicator-extended__label {
```
g-cheishvili, line 4:
```suggestion
#{$siblingBusyIndicator},
#{$siblingBusyIndicator} + .fd-busy-indicator-extended__label {
```
I think this is a bug. It does not do as previously did. to cover previous behavior you can save `* + .fd-busy-indicator` in some variable and then use it like this here

libs/core/src/lib/busy-indicator/busy-indicator.component.scss
```diff
@@ -1,21 +1,25 @@
...
+    white-space: nowrap;
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+}
+
+* + .fd-busy-indicator-extended__label {
```
g-cheishvili, line 12: Same thing with this. This style will always be applied because it will always have a sibling before

g-cheishvili (changes requested):

libs/core/src/lib/busy-indicator/busy-indicator.component.scss
```diff
@@ -1,15 +1,28 @@
...
+#{$siblingBusyIndicator},
+#{$siblingBusyIndicator} + .fd-busy-indicator-extended__label {
+    white-space: nowrap;
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+}
+
+#{$siblingBusyIndicator} + .fd-busy-indicator-extended__label {
```
g-cheishvili, line 15:
```suggestion
#{$siblingBusyIndicator} + .fd-busy-indicator-extended__label {
    white-space: nowrap;
    position: absolute;
    top: 50%;
    left: 50%;
```
Sorry just noticed

g-cheishvili (approved):

PR #10663, opened 2023-10-04
chore(docs): popover examples

droshev:
Updated popover examples(proper markup for the header and footer), simplified some examples


g-cheishvili (approved):

mikerodonnell89 (commented):
Any reason some of the e2e testing is simply deleted with no replacement?

droshev:
> Any reason some of the e2e testing is simply deleted with no replacement?

@mikerodonnell89 they weren't testing the popover. they were testing whether if you put H5 will be there. It doesn't look e2e either

PR #10678, opened 2023-10-05
fix(core): add aria-label to fd-tokenizer

### LINKED ISSUE #10293: [P3][Sourcing][Accessibility][Multi input tokens]Values are not noticeable even when read linearly with the screen reader, because they are in an unlabeled selection list (role=listbox).
Bug

Values are not noticeable even when read linearly with the screen reader, because they are in an unlabeled selection list (role=listbox).

Values are not noticeable even when read linearly with the screen reader, because they are in an unlabeled selection list (role=listbox).
When reading linearly with the screen reader's virtual cursor, JAWS outputs a selection list before the empty input field, but it is unclear how the selection list relates to the input field, and the values of the selection list are also not output because output is only made in forms mode.
tested on multi input mobile mode

reported by customer

mikerodonnell89:
fixes #10293


g-cheishvili (approved):
Please try to fill the languages

Issue #10649, opened 2023-10-02
Sprint 123 Defect hunting - Horizon Testing


g-cheishvili:
Issues that are found during the deffect hunt

g-cheishvili:
Core grid list keyboard navigation is not bound to the grid list container. If you arrive at the last item in any of the grid list containers and click "arrow right" it will jump to the next example component. Additionally, grid list item does not have ability to have navigation vertically. Only `arrow right` and `arrow left` work

g-cheishvili:
<img width="738" alt="DH Autocomplete is enabled for combobox" src="https://github.com/SAP/fundamental-ngx/assets/28543391/43924f6a-d33a-411c-8007-0348fea32f91">
Autocomplete is still enabled in Combobox example

g-cheishvili:
<img width="1578" alt="DH breadcrumbs incorrect routes" src="https://github.com/SAP/fundamental-ngx/assets/28543391/e2ada899-4c20-4650-b31f-f3d93fa22c86">
Breadcrumbs RouterLink example has incorrect routes attached to it and as a result user will not be able to see correct navigation when clicked on a example

g-cheishvili:
Core illustrated message inline SVG example calls `inject` outside of the Injection Context

g-cheishvili:
<img width="1251" alt="Screenshot 2023-10-03 at 13 17 05" src="https://github.com/SAP/fundamental-ngx/assets/28543391/16f039e5-231d-4597-9042-0bd7ce660327">
Arrow is in a wrong place. Popover placement.

g-cheishvili:
<img width="708" alt="Screenshot 2023-10-03 at 13 20 50" src="https://github.com/SAP/fundamental-ngx/assets/28543391/cf086314-1e1d-45ad-8a24-5f8240dc633d">
Micro process flow shows incorrectly labels. They are cut out

g-cheishvili:
<img width="1260" alt="Screenshot 2023-10-03 at 13 26 06" src="https://github.com/SAP/fundamental-ngx/assets/28543391/bcadf319-23c0-4e7e-8620-70ca7383f827">

Does not look compact to me, plus example description mentions removed property `compact`

g-cheishvili:
<img width="1181" alt="Screenshot 2023-10-03 at 13 28 06" src="https://github.com/SAP/fundamental-ngx/assets/28543391/ff94d945-2e7e-4cde-9a78-dfe284f15a10">

[...]

PR #10712, opened 2023-10-12
fix(core): some multi-combobox fixes

mikerodonnell89:
part of defect hunting. fixes this issue: https://github.com/SAP/fundamental-ngx/issues/10649#issuecomment-1758420556
and this issue: https://github.com/SAP/fundamental-ngx/issues/10649#issuecomment-1758422342
and this issue: https://github.com/SAP/fundamental-ngx/issues/10649#issuecomment-1758454457

need to test on chrome browser on an actual phone. not reproducible using chrome developer tools


g-cheishvili:
It must be iphone thing. My samsung opens up correctly

g-cheishvili:
> 
> 
> It must be iphone thing. My samsung opens up correctly

Whoops, sorry. PR changes work, but main is broken on my phone too

g-cheishvili (approved):

mikerodonnell89:
> > It must be iphone thing. My samsung opens up correctly
> 
> Whoops, sorry. PR changes work, but main is broken on my phone too

So I've just learned that iOS will automatically zoom in a little bit when an input is focused and the font size is smaller than 16px. So what I'm seeing on my phone is actually a zoomed-in results list. Might need to take this up with the designers...

PR #10731, opened 2023-10-16
fix(core,btp): various dh fixes

N1XUS:
## Related Issue(s)

closes none
- fixed popover resize capturing text;
- changed theming initialization mechanizm;
- added tablet and mobile examples for btp vertical navigation
- Changed overflow mechanizm calculation for btp vertical navigation

### DIFF
--- a/libs/btp/navigation/navigation-component.token.ts
+++ b/libs/btp/navigation/navigation-component.token.ts
@@ -1,12 +1,13 @@
 import { computed, Signal, TemplateRef } from '@angular/core';
+import { FdbViewMode } from '@fundamental-ngx/btp/shared';
 import { Nullable } from '@fundamental-ngx/cdk/utils';
 import { FdbNavigationListItemComponent } from './navigation-list-item-component.token';
-import { FdbNavigationMode, FdbNavigationState, FdbNavigationType } from './navigation.types';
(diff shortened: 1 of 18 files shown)


libs/btp/navigation/navigation-component.token.ts
```diff
@@ -1,11 +1,12 @@
...
 import { FdbNavigationListItemComponent } from './navigation-list-item-component.token';
+import { FdbNavigationMode, FdbNavigationState, FdbNavigationType } from './navigation.types';
 
 export abstract class FdbNavigationComponent {
-    abstract state: Signal<'expanded' | 'snapped' | 'popup'>;
-    abstract type: Signal<'horizontal' | 'vertical'>;
-    abstract mode: Signal<'desktop' | 'tablet' | 'phone'>;
+    abstract state: Signal<FdbNavigationState>;
+    abstract type: Signal<FdbNavigationType>;
+    abstract mode: Signal<FdbNavigationMode>;
```
g-cheishvili, line 9: Can you use the mode from the shared types? Me and Inna used that together and would be nice to have it here as well

g-cheishvili (changes requested):
Just one small change

g-cheishvili (approved):

PR #10744, opened 2023-10-17
fix(btp): custom vertical nav home

N1XUS:
closes none;

- Added ability to set routerLink for home item and highlight when it's active;
- Added ability to remove separator between home item and the rest of the items.

### DIFF
--- a/libs/btp/navigation/navigation-list/navigation-list-item.component.ts
+++ b/libs/btp/navigation/navigation-list/navigation-list-item.component.ts
@@ -26,7 +26,7 @@ import {
     inject,
     signal
 } from '@angular/core';
-import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+import { toSignal } from '@angular/core/rxjs-interop';
 import { RouterLink, RouterLinkActive } from '@angular/router';
 import {
     CssClassBuilder,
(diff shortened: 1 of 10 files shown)


libs/btp/navigation/navigation-list/navigation-list-item.component.ts
```diff
@@ -253,6 +259,21 @@ export class NavigationListItemComponent
...
+                activeRoute?.isActiveChange
+                    .pipe(startWith(activeRoute.isActive), takeUntilDestroyed(this._destroyRef))
+                    .subscribe((isActive) => {
+                        this.isRouterLinkActive.set(isActive);
+                    });
+            },
+            {
+                allowSignalWrites: true
+            }
+        );
```
g-cheishvili, line 276: you could do `toSignal` with `Injector` passed to it and not require `allowSignalWrites`

libs/btp/navigation/navigation.component.ts
```diff
@@ -94,6 +95,12 @@ export class NavigationComponent
         this.type.set(type);
     }
 
+    /** @hidden */
+    @Input({ alias: 'homeSeparator', transform: (value: any) => coerceBooleanProperty(value) })
+    set _homeSeparator(value: boolean) {
+        this.homeSeparator.set(value);
+    }
```
g-cheishvili, line 102: Use `BooleanInput` instead, otherwise it will not work with the string values like `"" | "true"` and etc

g-cheishvili (approved):

libs/btp/navigation/navigation-list/navigation-list-item.component.ts
N1XUS: Yeah, tried that, didn't work actually

PR #10746, opened 2023-10-18
fix(core): dynamic stacked tabs

N1XUS:
Downport of #10745


libs/core/src/lib/scroll-spy/scroll-spy.directive.ts
```diff
@@ -1,12 +1,25 @@
...
+    Directive,
+    ElementRef,
+    EventEmitter,
+    HostListener,
+    Input,
+    OnInit,
+    Output,
+    inject
+} from '@angular/core';
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
```
g-cheishvili, line 12: It's ng15 :/

g-cheishvili (changes requested):
Not compatible with ng15

g-cheishvili (approved):

Issue #10713, opened 2023-10-13
[Sourcing][i18n] Missing some string translations inside `f-ngx/i18n`


yixuan-liu:
#### Is this a bug, enhancement, or feature request?
A bug

#### Describe your proposal.
We'd like to get the strings translated for supported languages.

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
- Angular 15
- F-NGX `0.43.12`

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;

1. DateTime Picker (in both `Core` and `Platform`)  are missing i18n for `Previous month` and `Next month` tooltips

![date-time-1](https://github.com/SAP/fundamental-ngx/assets/22454655/7f1196ab-3ae2-4fc7-ad62-c5a0648d1b39)
![date-time-2](https://github.com/SAP/fundamental-ngx/assets/22454655/772e9526-eb38-498c-8e7e-bea9109c6789)

2. `<fdp-approval-flow>`

- `Menu` and `Add node` tooltips are not translated
![approval-4](https://github.com/SAP/fundamental-ngx/assets/22454655/73371389-94fd-41e6-bce1-3064fb6c4e17)
![approval-1](https://github.com/SAP/fundamental-ngx/assets/22454655/1093048c-7a2b-412a-9d71-aa5b687592b4)

- Approval statues: `not started`, `in progress`, `approved`, `rejected` are not translated in version `0.43.12`

![approval-6](https://github.com/SAP/fundamental-ngx/assets/22454655/2c405586-a06a-4671-a72c-cd2e94ee8678)

However, I see those are present in Angular 16 release src code: 
https://github.com/SAP/fundamental-ngx/blob/ng-15-downport/libs/i18n/src/lib/languages/english.ts
``
    "platformApprovalFlow.statusApproved": "approved",
    "platformApprovalFlow.statusRejected": "rejected",
    "platformApprovalFlow.statusInProgress": "in progress",
    "platformApprovalFlow.statusNotStarted": "not started",
``

#### Please provide relevant source code (if applicable).
- `0.43.12` translation
https://github.com/SAP/fundamental-ngx/blob/ng-15-downport/libs/i18n/src/lib/languages/english.ts

- `0.46.0` translation
https://github.com/SAP/fundamental-ngx/blob/ng-15-downport/libs/i18n/src/lib/languages/english.ts

PR #10748, opened 2023-10-18
feat(core): calendar now uses i18n package for translating labels

g-cheishvili:
## Related Issue(s)

Part of the #10713

## Description

`CalendarI18nLabels` has been deprecated and instead, now i18n provides key `coreCalendar` which is responsible for translating the labels.
Added translations for all of the languages


libs/core/src/lib/calendar/patch-deprecated-i18n-labels.ts
```diff
@@ -0,0 +1,29 @@
...
+import { FD_LANGUAGE, FdLanguage } from '@fundamental-ngx/i18n';
+import { Observable, map, startWith, switchMap } from 'rxjs';
+import { CalendarI18nLabels } from './i18n/calendar-i18n-labels';
+
+/**
+ * Use deprecated `CalendarI18nLabels` to patch `FD_LANGUAGE` observable
+ */
+export function patchDeprecatedI18nLabels(): Observable<FdLanguage> {
+    const deprecatedSymbol = inject(CalendarI18nLabels, { optional: true });
+    if (deprecatedSymbol) {
```
mikerodonnell89, line 11:
```suggestion
    if (deprecatedSymbol && deprecatedSymbol.labelsChange) {
```

I reverted the core calendar i18n example and the page would not load due to console errors. Need this check to allow backwards compatibility

mikerodonnell89 (changes requested):
see comment

libs/core/src/lib/calendar/patch-deprecated-i18n-labels.ts
g-cheishvili: I will fix that, but differently. Thanks

mikerodonnell89 (approved):

PR #10754, opened 2023-10-18
fix: table cell focus issue

### LINKED ISSUE #10626: [P1][Sourcing][Accessibility][ATL testing][Table]: Selected checkbox has no visible focus in table
Bug

#### Describe your proposal.
Selected checkbox has no visible focus in table, instead some white border appears which is hardly visible to users.
Pls correct the focus color and make it look prominent.

https://github.com/SAP/fundamental-ngx/assets/88182358/a57e4df7-c910-468a-8375-78e8eb50bc2d

No

ng: 15
platform: 0.43.x

Multi-select table

reported by ATL

mikerodonnell89:
fixes #10626

### DIFF
--- a/libs/core/src/lib/table/table.component.scss
+++ b/libs/core/src/lib/table/table.component.scss
@@ -19,6 +19,6 @@
 }
 
 // TODO: remove hardcoded style after fundamental-styles version bump
-.fd-table__cell--focusable[aria-selected=true]:after {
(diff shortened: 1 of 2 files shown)


N1XUS (approved):

g-cheishvili (approved):
I guess these hardcoded stuff will not be removed for ng15

PR #10759, opened 2023-10-19
fix(core): multi input new token

N1XUS:
## Related Issue(s)

closes none
- Fixes the issue with adding custom token to core multi input


g-cheishvili (approved):

mikerodonnell89 (changes requested):
What issue exactly is this addressing? I'm still able to add tokens with only spaces and seeing the tokens misaligned
Screenshot 2023-10-19 at 10 57 06 AM

PR #10761, opened 2023-10-19
fix(core): static height scenario handler

### LINKED ISSUE #10741: [P1][Sourcing][Responsiveness] Multiple tabs are shown selected in tab-list
Bug

#### Describe your proposal.
On rendering the fd-tab asynchronously, multiple fd-tabs are shown selected (blue underline) as shown in the below screenshot.
Screenshot 2023-10-17 at 4 27 52 PM
 
But only the first tab should be selected (by default) and others should not be selected.

No, it cannot be handled from application end. Tried multiple things.

Angular: "15.2.0"
Fundamental Library: "0.43.11"

1. fd-tab-list is being used
2. Render fd-tabs asynchronously( using setTimeout, api calls etc.) inside fd-tab-list

Editor Link : https://stackblitz.com/edit/angular-ugca8x?file=src%2Fapp%2Fdynamic-page-tabs-example.component.ts
Application Link: https://angular-ugca8x.stackblitz.io

Yes

N1XUS:
## Related Issue(s)

closes #10741
- This PR handles the case when the scrollSpy container height isn't changed, but new items are being added.


mikerodonnell89 (approved):
works well 👍

PR #10766, opened 2023-10-20
fix(core): reactive condensed state

### LINKED ISSUE #10765: [Bug] `fd-list-navigation-item-text` does not honour `condensed` directive of parent `fd-vertical-navigation`

Bug

#### Describe your proposal.

`fd-list-navigation-item-text` does not honour `condensed` directive of parent `fd-vertical-navigation` when changing value at runtime.

Angular: 16.1.0
Fundamental-ngx: 0.46.0

```html
<fd-shellbar [sideNav]="true">
  <button
    fd-shellbar-side-nav
    fd-button
    fdType="transparent"
    glyph="menu2"
    style="width: 2.75rem"
    (click)="condensed = !condensed"
  ></button>
</fd-shellbar>
<fd-vertical-navigation [condensed]="condensed">
  <li fd-list-navigation-item>
    <fd-icon glyph="home"></fd-icon>
    <span fd-list-navigation-item-text> Overview </span>
  </li>
</fd-vertical-navigation>
```

```typescript

@Component({
  selector: 'app-nav',
  templateUrl: './nav.component.html',
  styleUrls: ['./nav.component.css']
})
export class NavComponent {
  condensed: boolean = false;
}

```

image
image

Full source code here: https://github.com/samcarswell/fundamental-ngx-nav-item-text-bug

N1XUS:
## Related Issue(s)

closes #10765

## Description
Previously, we set `condensed` state for the vert nav items only when vert nav component is initialized.
Now vert nav component reacts to the changes of the `condensed` input property and updates the list accordingly taking into consideration possible changes in the list itself too.

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [x] poor examples

4. Accessibility testing

##### PR Quality

-   [x] update `README.md`

### DIFF
--- a/libs/core/src/lib/list/list-navigation-item/list-navigation-item.component.ts
+++ b/libs/core/src/lib/list/list-navigation-item/list-navigation-item.component.ts
@@ -13,7 +13,11 @@ import {
     Input,
     Optional,
     QueryList,
-    forwardRef
+    Renderer2,
+    effect,
+    forwardRef,
+    inject,
+    signal
 } from '@angular/core';
 import { KeyUtil, RtlService } from '@fundamental-ngx/cdk/utils';
 import { FD_ICON_COMPONENT, IconComponent } from '@fundamental-ngx/core/icon';
@@ -62,10 +66,6 @@ export class ListNavigationItemComponent implements AfterContentInit, AfterViewI
     _ariaLevel: number;
 
     /** @hidden */
-    @HostBinding('class.fd-list__navigation-item--condensed')
-    _condensed = false;
-
-    /** @hidden */
     @ContentChild(FD_LIST_COMPONENT)
     _listComponent: ListComponentInterface;
 
@@ -92,6 +92,9 @@ export class ListNavigationItemComponent implements AfterContentInit, AfterViewI
     }
 
     /** @hidden */
+    _condensed = signal(false);
+
+    /** @hidden */
     _expanded = false;
 
     /** @hidden */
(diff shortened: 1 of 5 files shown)


libs/core/src/lib/vertical-navigation/vertical-navigation.component.ts
```diff
@@ -26,38 +28,72 @@ import { VerticalNavigationMainNavigationComponent } from './vertical-navigation
...
+                const isCondensed = this._condensed();
+
+                this._mainNavItems().forEach((item) => {
+                    item._condensed.set(isCondensed);
+                });
+            },
+            {
+                allowSignalWrites: true
+            }
+        );
```
g-cheishvili, line 86: Can we make a injection from mainNavItem to navbar and have calculation from there? I just do not like signal writes in effects

libs/core/src/lib/list/list-navigation-item/list-navigation-item.component.ts
```diff
@@ -113,7 +116,18 @@ export class ListNavigationItemComponent implements AfterContentInit, AfterViewI
...
+    private readonly _renderer2 = inject(Renderer2);
+
+    /** @hidden */
+    constructor(private _elementRef: ElementRef, @Optional() private _rtlService: RtlService) {
+        effect(() => {
+            if (this._condensed()) {
+                this._renderer2.addClass(this._elementRef.nativeElement, 'fd-list__navigation-item--condensed');
+            } else {
+                this._renderer2.removeClass(this._elementRef.nativeElement, 'fd-list__navigation-item--condensed');
+            }
```
g-cheishvili, line 128: host metadata would be preferable

libs/core/src/lib/vertical-navigation/vertical-navigation.component.ts
N1XUS: It should be done on the root level only, so we use contentCHildren with descendants false in this case. As to writting signals in effects, it still ok to do so

libs/core/src/lib/list/list-navigation-item/list-navigation-item.component.ts
N1XUS: Nope. It will not react on the changes because host stuff is not from this cdr context rather than parent one (which is not vertical nav component but example component)

N1XUS: Yes

g-cheishvili (approved):

PR #10767, opened 2023-10-20
fix(core): various carousel fixes

g-cheishvili:
## Related Issue(s)

DH issues

## Description

- Fixed carousel vertical keyboard navigation
- Fixed focus lost on keyboard navigation
- Fixed carousel busy indicator misplacement

### DIFF
--- a/libs/core/src/lib/carousel/carousel.component.ts
+++ b/libs/core/src/lib/carousel/carousel.component.ts
@@ -1,3 +1,5 @@
+import { Direction } from '@angular/cdk/bidi';
+import { NgClass, NgFor, NgIf, NgTemplateOutlet } from '@angular/common';
 import {
     AfterContentInit,
     AfterViewChecked,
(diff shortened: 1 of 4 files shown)


libs/core/src/lib/carousel/carousel.component.ts
```diff
@@ -373,13 +373,35 @@ export class CarouselComponent
     onKeydownArrowRight(event: KeyboardEvent): void {
         event.preventDefault();
         this._isRtl() ? this.previous() : this.next();
+        this.carouselContainer.nativeElement.focus();
     }
 
     /** @hidden */
     @HostListener('keydown.arrowleft', ['$event'])
     onKeydownArrowLeft(event: KeyboardEvent): void {
```
N1XUS, line 381: Maybe combine these methods into one that defines which action to take based on rtl and vertical input?

g-cheishvili: Yeap, makes sense

N1XUS (approved):

PR #10768, opened 2023-10-20
fix(core): wizard scroll and focus fixes

### LINKED ISSUE #10391: Wizard focus and scrolling issues
When navigating to the next step, the first focusable item in the new step should be focused. Also there seems to be an issue with scrolling with the stacking wizards. When navigating to a new step, the wizard no longer automatically scrolls to the step. And if the user has scrolled down, the wizard will scroll itself to the top

mikerodonnell89:
fixes #10391


libs/core/src/lib/wizard/wizard.component.ts
```diff
@@ -412,15 +428,27 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
             });
             this._setUpScrollListener();
         }
-        this.scrollbar.scroll({ top: 0, behavior: 'smooth' });
+    }
+
+    /** @hidden */
+    private _focusFirstTabbableElement(index = 0): void {
+        const contentContainer = this._elRef.nativeElement.querySelectorAll('.fd-wizard__content')[index];
+        const elToFocus = contentContainer?.querySelector('button, input, select, [tabindex]') as HTMLElement;
```
g-cheishvili, line 450: https://github.com/SAP/fundamental-ngx/blob/fefac1929380f482a9786b09513779548281256c/libs/cdk/src/lib/utils/services/tabbable-element.service.ts#L14
You need something like this here?

g-cheishvili (approved):

PR #10802, opened 2023-10-25
fix(core): deprecate tabbableScrollbar

### LINKED ISSUE #10615: [P3.1][Sourcing][Accessibility][ATL testing][Table]: Sort popover in table has wrong key usage for focus
Bug

#### Describe your proposal.
As soon as user opens sort column popover for a column , automatically the first menu of the popover "Sort Ascending" has to be focussed. But currently focus is not trapped inside and user has to press extra tab key to move the focus explicitly to this popover and then use up or down arrow keys to move between the menu items which is wrong.
Expectation: User opens the sort popover and automatically the first menu item has to receive the focus and user can then only use up / down arrow keys to operate

Screenshot 2023-09-29 at 11 17 41 AM

No

ng: 15
platform: 0.43.x

replicable on https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/table/basic

https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/table/basic

mikerodonnell89:
fixes #10615

### DIFF
--- a/libs/core/src/lib/popover/popover-body/popover-body.component.ts
+++ b/libs/core/src/lib/popover/popover-body/popover-body.component.ts
@@ -45,9 +45,6 @@ export class PopoverBodyComponent implements AfterViewInit {
     /** Whether to wrap content with fd-scrollbar directive. */
     _disableScrollbar = false;
 
-    /** Should fd-scrollbar have tabindex*/
-    _tabbableScrollbar = true;
-
     /** @hidden */
     @ViewChild(CdkTrapFocus)
     _cdkTrapFocus: CdkTrapFocus;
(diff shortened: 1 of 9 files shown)


libs/core/src/lib/popover/popover-body/popover-body.component.ts
```diff
@@ -178,5 +175,6 @@ export class PopoverBodyComponent implements AfterViewInit {
         if (this._focusAutoCapture) {
             this._cdkTrapFocus.focusTrap.focusFirstTabbableElement();
         }
+        console.log(document.activeElement);
```
mikerodonnell89, line 178:
```suggestion
```

PR #10846, opened 2023-10-30
feat: using properties instead of the json

g-cheishvili:
## Related Issue(s)

## Description


libs/docs/shared/src/lib/core-helpers/toolbar/toolbar.component.ts
```diff
@@ -186,6 +180,7 @@ export class ToolbarDocsComponent implements OnInit, OnDestroy {
         private _docsService: DocsService,
         private _http: HttpClient
     ) {
+        langSubject$.subscribe((lang) => console.log({ lang }));
```
N1XUS, line 183: Is it needed?

g-cheishvili: No, I forgot

mikerodonnell89 (commented):
Unit tests are still looking for the old json files

mikerodonnell89 (commented):
I'm nitpicking here, but this doc page needs a little updating for the new *.properties files: https://deploy-preview-10846--fundamental-ngx.netlify.app/#/i18n/writing-translations

g-cheishvili:
Closing this, because I accidentally froze the CI, need fresh one

PR #10852, opened 2023-10-31
fix(platform): multi combobox ignore tab key

### LINKED ISSUE #10824: [P3.2][Sourcing][Accessibility][Pretest][Multi combobox]The list entries receive the focus not only with arrow keys, but also individually with TAB.
Bug

#### Describe your proposal.
The list entries receive the focus not only with arrow keys, but also individually with TAB. 

list should be traversed only using arrow keys 

No

Angular: 15
core platform: 0.43.14

https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/multi-combobox#data-source

N1XUS:
closes #10824

- Previously we had an event handler for tab key, which is not by the guidelines. This PR removes that handler to enable native tab traverse.


mikerodonnell89 (approved):
Approving but I do have one thought. If the user focuses an option with the arrow keys and presses tab, the focus moves to the start of the browser. I think it'd be best to move to the next focusable (via tab) item on the page. Probably more of a question for ux

N1XUS:
@droshev @mikerodonnell89 Maybe we just activate focus trap for the dropdown? So while dropdown is opened, tab key is not operable?

mikerodonnell89:
@N1XUS two things:

- I think we may want this behavior in core as well
- On platform combobox, tabbing to a new combobox will leave the popover for the previous combobox still open

droshev:
> @N1XUS two things:
> 
> * I think we may want this behavior in core as well
> * On platform combobox, tabbing to a new combobox will leave the popover for the previous combobox still open

yes, core and platform should be done in the same way

N1XUS:
> core and platform should be done in the same way

They are now.

g-cheishvili:
If you're focused on the input, and click the `tab`, the focus is moved to the next element, but the dropdown is not closed

N1XUS:
@g-cheishvili please re-check

g-cheishvili (approved):
I like this behavior

droshev (approved):

PR #10853, opened 2023-10-31
fix(core,platform): inputs with dropdowns ignore tab key

N1XUS:
Downport of #10852


droshev (changes requested):
see my comment [here](https://github.com/SAP/fundamental-ngx/pull/10852#pullrequestreview-1707392495)

droshev (approved):

PR #10856, opened 2023-10-31
feat: using properties instead of the json

g-cheishvili:
## Related Issue(s)

Replica of the #10846

## Description


g-cheishvili:
Again, broke the CI

PR #10857, opened 2023-10-31
feat(i18n): using properties instead of the json in i18n

g-cheishvili:
## Related Issue(s)

Replica of the #10846

## Description

- if you have properties staged, they will generate the appropriate jsons
- I had to fix the CI agents issue


N1XUS:
@g-cheishvili two questions:
- If we have `.json` files, why do we load `.properties` files to docs?
- If translations are loaded via http now, do we have schematics that would update customers application to automatically include translations in resulting bundle?

g-cheishvili:
> @g-cheishvili two questions:
> 
> * If we have `.json` files, why do we load `.properties` files to docs?
> * If translations are loaded via http now, do we have schematics that would update customers application to automatically include translations in resulting bundle?

using json in prod builds and .properties in local as discussed. For clients nothing changes

PR #10866, opened 2023-11-01
feat(i18n): using properties format instead of the json

g-cheishvili:
## Related Issue(s)

## Description

- Source of truth now is the `.properties` files.
- if you have properties staged, they will generate the appropriate jsons


libs/nx-plugin/src/executors/transform-translations/duplicates.ts
```diff
@@ -0,0 +1,39 @@
...
+function set<V = string>(obj: RecursiveRecord<V>, path: string, value: V): void {
+    const pathParts = path.split('.');
+    let pathValue = obj;
+    for (let i = 0; i < pathParts.length - 1; i++) {
+        const part = pathParts[i];
+        pathValue[part] = pathValue[part] || {};
+        pathValue = pathValue[part] as RecursiveRecord<V>;
+    }
+    pathValue[pathParts[pathParts.length - 1]] = value;
+}
```
N1XUS, line 14: lodash `set` wont work?

N1XUS (approved):

N1XUS:
@g-cheishvili. Great job! As an idea: we could deprecate FD_LANGUAGE implementation where we have to load every used language into our bundle, and introduce FD_LAZY_LANGUAGE that can be a promise with `import(...).then()` function what would allow i18n service to load language JSON files lazily

g-cheishvili:
> @g-cheishvili. Great job! As an idea: we could deprecate `FD_LANGUAGE` implementation where we have to load every used language into our bundle, and introduce `FD_LAZY_LANGUAGE` that can be a promise with `import(...).then()` function what would allow i18n service to load language JSON files lazily.
> 
> For the end user not much would change since we still would support old `FD_LANGUAGE` approach, but it would be preferable to use lazy implementation

I'd say we should get rid of it and have a service instead

PR #10904, opened 2023-11-06
fix(btp): tool header styles discrepancy and updated overflows

g-cheishvili:
## Related Issue(s)

closes

## Description

- Updated the syntax for the action buttons. There are breaking changes
- Fixed overflow menu's keyboard navigation
- Fixed spacing between the items in actions bar
- updated examples accordingly
- added two new directives in focusable list feature, indirect children and indirect focusable lists, so that now it can works without the need for direct descendancy
- Added ability for more flexible "custom overflow button" in the overflow list.
- Deprecated the CDK's Disabled directive's event manager plugin

All the CSS additions are needed to be in ngx, because they are here because of the overflow layout

BREAKING CHANGE:
The `fdb-tool-header-actions` has been removed, now actions should be used directly


InnaAtanasova:
Great work! The overflow works so smoothly now! Just a few minor stuff:
Screenshot 2023-11-06 at 2 17 39 PM
Screenshot 2023-11-06 at 2 17 55 PM
Screenshot 2023-11-06 at 2 13 47 PM

InnaAtanasova (approved):

PR #10913, opened 2023-11-07
fix(core): select aria attributes

### LINKED ISSUE #10806: [P2][Sourcing][Accessibility][Pretest] [FDP-Select] Incorrect role "listbox" with the misleading additional information "collapsed" and "has popup"
Bug

#### Describe your proposal.
Incorrect role "listbox" with the misleading additional information "reduced" and "has popup"https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/
This is not as per w3 standards.

No

Angular 15
core/ platform: 0.43.14
When focus is on select box, correct role of the component is not announced and only listbox is getting announced

https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/select#mode

N1XUS:
## Related Issue(s)

closes #10806

## Description
- Changed role from `listbox` to `combobox`;
- Added `aria-roledescription="listbox"` attribute;
- Added `aria-hidden="true"` for dropdown icon so it wouldn't be read by the screenreader while reading select value.

### DIFF
--- a/libs/core/src/lib/select/select.component.html
+++ b/libs/core/src/lib/select/select.component.html
@@ -30,7 +30,6 @@
 <ng-template #selectInputControlTemplate>
     <div
         class="fd-select__control"
-        role="listbox"
         #selectControl
         [class]="_selectControlClass"
         [class.is-active]="_isOpen"
@@ -41,6 +40,8 @@
         [attr.id]="controlId"
         [attr.aria-active]="_isOpen"
         aria-live="polite"
+        role="combobox"
+        aria-roledescription="listbox"
         [attr.aria-expanded]="_isOpen"
         [attr.aria-disabled]="disabled"
         [attr.aria-controls]="controlId + '-list-box'"
@@ -70,7 +71,7 @@
             [class.is-disabled]="disabled"
             *ngIf="!readonly"
         >
-            <fd-icon [glyph]="glyph"></fd-icon>
+            <fd-icon [ariaHidden]="true" [glyph]="glyph"></fd-icon>
         </span>
     </div>
 </ng-template>


libs/core/src/lib/select/select.component.html
```diff
@@ -41,6 +40,9 @@
         [attr.id]="controlId"
         [attr.aria-active]="_isOpen"
         aria-live="polite"
+        role="combobox"
+        aria-autocomplete="list"
```
InnaAtanasova, line 44: why do you have `aria-autocomplete`? 
The aria-autocomplete attribute indicates whether inputting text could trigger display of one or more predictions of the user's intended value for a combobox, searchbox or  textbox and specifies how predictions will be presented if they are made. In Select we don't type. 
For more information: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-autocomplete

libs/core/src/lib/select/select.component.html
```diff
@@ -70,7 +72,7 @@
             [class.is-disabled]="disabled"
             *ngIf="!readonly"
         >
-            <fd-icon [glyph]="glyph"></fd-icon>
+            <fd-icon aria-hidden="true" [glyph]="glyph"></fd-icon>
```
InnaAtanasova, line 75: I don't think this is applied in the markup. 
Screenshot 2023-11-07 at 11 50 51 AM

libs/core/src/lib/select/select.component.html
N1XUS: @InnaAtanasova yeah, my bad. Will remove this one

InnaAtanasova (approved):

PR #10929, opened 2023-11-09
[WIP] feat(core,platform): replace empty components with directives

N1XUS:
Attempt to replace empty-templated components with directives.

Goal of this experiment is to reduce amount of view creation calls and rely more on the native DOM tree


g-cheishvili:
1 note . Rename classes with `Directive` suffix and then `export as PreviousName` and add deprecation to that previous name

N1XUS:
> 1 note . Rename classes with `Directive` suffix and then `export as PreviousName` and add deprecation to that previous name

I don't think it's going to work with DI since Angular's DI maps dependency to its classname, so we may find ourselves in situations when user tries to inject `VariantManagementWrapperComponent`, even though the injector has `VariantManagementWrapperDirective`, but could not resolve the dependency because of the different classnames

PR #10930, opened 2023-11-09
fix(platform): list accessibility issues

### LINKED ISSUE #10664: [P3.2][Sourcing][Accessibility][ATL testing][List with group header]: Header is included in total list items count
Bug

#### Describe your proposal.
The platform list with group header is having some problem when screenreader is reading the total count , the header is also counted as list item.
https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/list#group-header

Expectation: screenreader should not include the header of the list as list item or option.
Few hints: "li" has been added with role=option under "fdp-list-group-header"
I believe this might done to make header also traverse using up / down arrow keys for reachability, but this might be a root cause for screenreader reading it wrongly

Screenshot 2023-10-05 at 4 20 49 PM

Not really, as fdp-list and its structure depends on F-NGX library we cannot access child elements

angular 15
core / platform: 0.43.9
https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/list#group-header

```
<fdp-list [role]="'listbox'" id="recentsearchid">
 <fdp-list-group-header   class="recentSearches_color" [groupHeaderTitle]="recentSearchLbl"></fdp-list-group-header> 
  
  <fdp-standard-list-item 
      *ngFor="let item of items"
      [role]="'option'"
      [title]="item.title || item.name"
      class="recentSearches_color_value"
      (click)="actionedRow($event,item)"
      (keydown.enter)="actionedRow($event,item)">
  </fdp-standard-list-item >
</fdp-list>
```
This issue has been reported by ATL testing for sourcing.

N1XUS:
## Related Issue(s)

closes #10664

## Description
- Changed algorithm of `aria-posinset` assignment;
- Changed role of list item from `option` to `listitem` for standard list items;
- Changed role of list item from `option` to `group` for group/header list items;
- Added `aria-owns` attribute for group/header list items;


InnaAtanasova:
**List - Load data on button click** example:
The list items should have `aria-setsize`.
`aria-setsize` is used when not all items in the set are present in the DOM.
Here's more info: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-setsize

InnaAtanasova:
Screenshot 2023-11-09 at 11 17 27 AM

In these 2 cases, when you need to have `aria-selected` you need role `option`. 
Aria-selected can be used in this case:
Screenshot 2023-11-09 at 11 19 06 AM

If you have checkboxes you will need `aria-multiselectable="true"`
The markup would look like:
```
<ul role="listbox" aria-multiselctable="true">
    <li role="option" aria-selected="true"></li>
    <li role="option" aria-selected="false"></li>
    <li role="option" aria-selected="false"></li>
</ul>
```

N1XUS:
@InnaAtanasova please re-check

InnaAtanasova:
where the list elements are with role `option`, the `ul` role should be `listbox`.

mikerodonnell89 (changes requested):
- Should it say "and 3 more items" when the list has 6? Just three until the next header
- On first focus we see this "list 2 items" at the end

Screenshot 2023-11-13 at 10 50 21 AM

mikerodonnell89 (dismissed):
Working a lot better now but still does not address the original issue of the headers being included in the total list count:

Screenshot 2023-11-22 at 2 03 51 PM

N1XUS:
> Working a lot better now but still does not address the original issue of the headers being included in the total list count:
> 
> Screenshot 2023-11-22 at 2 03 51 PM

Group is a part of the list. That's why it's being read.
UI5 also have similar behaviour:
image

InnaAtanasova (approved):

PR #10932, opened 2023-11-09
fix(core): segmented button accessibility improvements

### LINKED ISSUE #10805: [P3.1][Sourcing][Accessibility][Pretest] Segmented button has wrong role as toggle not as per W3 standards
bug

#### Describe your proposal.
the role toggle is not correct as the pressed button cannot be operated but receives focus with TAB
Usually the toggle button should change the selection using SPACE OR ENTER key only , but currently the user has to move between states using tab and then press space or enter which is not as per w3 standards if role = toggle is used.
https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch/

No

Angular: 15
core/ platform: 0.43.14

replicable on https://ng-15-downport--fundamental-ngx.netlify.app/#/core/segmented-button#toggle

mikerodonnell89:
fixes #10805

### DIFF
--- a/libs/core/src/lib/button/button.component.ts
+++ b/libs/core/src/lib/button/button.component.ts
@@ -12,7 +12,7 @@ import {
 } from '@angular/core';
 import { BaseButton } from './base-button';
 import { Subscription } from 'rxjs';
-import { applyCssClass, CssClassBuilder } from '@fundamental-ngx/cdk/utils';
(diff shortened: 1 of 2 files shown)


InnaAtanasova (approved):

InnaAtanasova:
https://www.w3.org/WAI/ARIA/apg/patterns/listbox/

PR #10933, opened 2023-11-09
fix(core): segmented button a11y improvements

### LINKED ISSUE #10805: [P3.1][Sourcing][Accessibility][Pretest] Segmented button has wrong role as toggle not as per W3 standards

mikerodonnell89:
fixes #10805


InnaAtanasova (approved):

g-cheishvili:
Do not merge this yet

libs/core/src/lib/segmented-button/segmented-button.component.ts
```diff
@@ -90,8 +93,15 @@ export class SegmentedButtonComponent implements AfterViewInit, ControlValueAcce
     constructor(
         private readonly _changeDetRef: ChangeDetectorRef,
         private readonly _elementRef: ElementRef,
-        private readonly _destroyRef: DestroyRef
-    ) {}
+        private readonly _destroyRef: DestroyRef,
+        @Optional() @Host() private _focusableList: FocusableListDirective,
+        @Optional() private _rtlService: RtlService
+    ) {
+        this._focusableList.navigationDirection = this.vertical ? 'vertical' : 'horizontal';
```
g-cheishvili, line 100: This will need to be updated reactively, what if in the runtime user replaced vertical with horizontal?

libs/docs/core/segmented-button/examples/segmented-button-complex-example/segmented-button-complex-example.component.ts
```diff
@@ -1,14 +1,15 @@
 import { NgFor } from '@angular/common';
 import { Component } from '@angular/core';
 import { FormsModule } from '@angular/forms';
+import { FocusableListModule } from '@fundamental-ngx/cdk';
```
g-cheishvili, line 4: do not import anything from the `cdk` directly, use `cdk/utils` secondary entrypoint and do you need `Module`?

libs/docs/core/segmented-button/examples/segmented-button-default-example.component.ts
```diff
@@ -1,5 +1,6 @@
 import { Component } from '@angular/core';
 import { FormsModule } from '@angular/forms';
+import { FocusableListModule } from '@fundamental-ngx/cdk';
```
g-cheishvili, line 3: same here as above

libs/docs/core/segmented-button/examples/segmented-button-form-example/segmented-button-form-example.component.ts
```diff
@@ -1,13 +1,14 @@
 import { Component } from '@angular/core';
 import { FormControl, FormGroup, FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { FocusableListModule } from '@fundamental-ngx/cdk';
```
g-cheishvili, line 3: `cdk`

libs/docs/core/segmented-button/examples/segmented-button-toggle-example.component.ts
```diff
@@ -1,13 +1,14 @@
 import { Component } from '@angular/core';
 import { FormsModule } from '@angular/forms';
+import { FocusableListModule } from '@fundamental-ngx/cdk';
```
g-cheishvili, line 3: `cdk` and the `module` usage

PR #10953, opened 2023-11-14
Translation Delivery

service-tip-git:
Translation Delivery


g-cheishvili:
The ISO 639-1 code "sh" is not a language code on its own. It was formerly used for Serbo-Croatian, but the usage of "sh" has been deprecated in favor of the individual language codes for Serbian (sr) and Croatian (hr)

Which one does this PR refer to? 
@droshev

g-cheishvili:
Already in main

PR #10996, opened 2023-11-23
fix(platform): remove platform components action-bar, file-uploader, …

droshev:
BREAKING CHANGE:
remove platform components: action-bar, file-uploader, thumbnail, and upload collection


droshev:
addressed in https://github.com/SAP/fundamental-ngx/pull/11024

PR #11048, opened 2023-12-04
chore: updated workspace layout and added unit tests of schematics

g-cheishvili:
## Related Issue(s)

closes none as far as I'm aware

## Description

Modified structure, previously the roots for the libraries `platform`, `core`, `cx` and `cdk` were deep inside the real root folders. instead of `libs/*` they were `libs/*/src/lib`, which was not the most beautiful thing in the world and that layout added additional problems related to the schematics. Previously schematics were not actually the executable, usable, parts of the monorepo, they were there just as the sources, without ability to properly test them. Now, schematics are part of the `nx` monorepo, with their own distinct project, which are properly used by their respective libraries, because paths are now fixed. e.g. core's package.json references actual, existing `collection.json` in the file system and that gives us ability to directly launch `ng-add` and other schematics directly from our monorepo. additionally those ts files were transformed and moved to the dist via some custom executors, now, since they are proper nx projects, they use `@nx/js` builders to build.
Additionally previously we did not have ability to write unit tests on the schematics, but now, there is ability for that. we can and should write tests for the schematics. As a first step, covered with tests ng-add schematics of the core


g-cheishvili:
Since I updated almost every file, I have to `format:write` many of them

PR #11054, opened 2023-12-05
fix(core): segmented button accessibility improvements

droshev:
fixes https://github.com/SAP/fundamental-ngx/issues/10805

a clean PR for https://github.com/SAP/fundamental-ngx/pull/10932


InnaAtanasova:
https://www.w3.org/WAI/ARIA/apg/patterns/listbox/

PR #11058, opened 2023-12-07
fix(platform): reset table rows before new rows emitted

### LINKED ISSUE #11057: [P1][Sourcing][Regression] Table : Table is not rendering rows after updating datasource
Bug - Regression

#### Describe your proposal.
After rebinding the datasource, table in UI shows no data found.

No

Fundamental Library - v0.43.18
Angular - v15

1. Use stackblitz - [click here](https://stackblitz.com/edit/angular-9qxxca-mpyvnk?file=src%2Fapp%2Fplatform-table-preserved-state-example.component.html,src%2Fapp%2Fplatform-table-preserved-state-example.component.ts,src%2Fapp%2Fapp.module.ts)
2. Click on test data button
3. Notice that table becomes empty in the UI even though datasource is having data.

https://github.com/SAP/fundamental-ngx/assets/71999626/28bfedce-d7ea-45ea-a2fe-f88642d21bc9

Refer above stackblitz link.

May be the fix for this issue https://github.com/SAP/fundamental-ngx/issues/11002 is causing this regression.

N1XUS:
## Related Issue(s)

closes #11057

## Description
The issue was with late callback function call when data source changed. This fix calls cleanup of the table rows right when the data source is being set.

### DIFF
--- a/libs/platform/table/table.component.ts
+++ b/libs/platform/table/table.component.ts
@@ -1338,6 +1338,17 @@ export class TableComponent<T = any>
         tableEl.rows[position.rowIndex].cells[position.colIndex].focus();
     }
 
+    /**
+     * Clears the array of table rows. Triggered when new data source is applied.
+     */
+    clearTableRows(): void {
+        this._tableRows = [];
+        this._tableRowsVisible = [];
+        this._tableRowsInViewPortPlaceholder = [];
+        this._newTableRows = [];
+        this._dataSourceTableRows = [];
+    }
+
     // Private API
 
     /** @hidden */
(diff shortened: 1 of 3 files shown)


libs/platform/table/table.component.ts
```diff
@@ -1836,6 +1838,7 @@ export class TableComponent<T = any>
 
     /** @hidden */
     private _setTableRows(rows = this._dataSourceTableRows): void {
+        console.log(rows);
```
g-cheishvili, line 1841: Logs

g-cheishvili (approved):

PR #11063, opened 2023-12-08
fix(core): allow for programmatic change of pane size

mikerodonnell89:
fixes none. Issue raised to me via slack. Programmatically updating pane "size" property after init did not adjust pane size

### DIFF
--- a/libs/core/splitter/splitter-split-pane/splitter-split-pane.component.ts
+++ b/libs/core/splitter/splitter-split-pane/splitter-split-pane.component.ts
@@ -38,7 +38,10 @@ export class SplitterSplitPaneComponent implements OnInit, AfterViewInit, OnDest
     @Input()
     set size(size: string) {
         this._size = size;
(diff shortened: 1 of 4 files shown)


libs/core/splitter/splitter-split-pane/splitter-split-pane.component.ts
```diff
@@ -38,7 +38,10 @@ export class SplitterSplitPaneComponent implements OnInit, AfterViewInit, OnDest
     @Input()
     set size(size: string) {
         this._size = size;
-        this._actualSize = this._size;
+        if (this._actualSize !== this._size) {
+            this._actualSize = this._size;
+            this.sizeChange.emit();
+        }
```
N1XUS, line 44: I would rewrite it with signals, and in `splitter-pane-container` I would use effect that reacts on every pane.size$() update

N1XUS (approved):

PR #11082, opened 2023-12-12
fix(platform): fdp-combobox selection issues

### LINKED ISSUE #11067: [P2][Sourcing][fdp-combobox]: Combobox not showing correct selected value when lookupKey is unique and label is same
Bug

#### Describe your proposal.
There are 2 issues with respect to fdp-combobox
**Issue 1:** 
 When using custom dataSource that is passed to combobox with group=true which has name , type and id, then the selection value is not shown properly even though  lookupKey="id" is being passed and getting console error.

Expectation: among groups with same label but with different lookupKey - the selected value has to show that matches lookupKey unique one selected.

Getting below console error

Screenshot 2023-12-09 at 1 44 02 PM

**Issue 2:** 
When using fdp-combobox without group=true, with just lookupKey and displayName then the records having same label but different lookupKey , both are shown as selected 
Expectation: Both records should not be shown as selected but only the one based on lookupKey needs to be shown as selected 

Screenshot 2023-12-09 at 1 44 37 PM

No
Angular: 15
Core / platform: 0.43.20
For issue 1:
1) Use stackblitz [link](https://stackblitz.com/edit/angular-9qxxca-qhylee?file=src%2Fapp%2Fplatform-table-preserved-state-example.component.html,src%2Fapp%2Fplatform-table-preserved-state-example.component.ts,package.json)
2) Notice that from ts file the default selected item is given as datasource[0] but it doesnt show it as selected.
3) Open combobox select first record "Apple"
4) Check the console error.

https://github.com/SAP/fundamental-ngx/assets/88182358/4de90bda-8fb1-4cec-8e76-faf503516e67

For issue 2:
1) remove properties   [group]="true"and groupKey="type" from html file
2) Open combobox
3) Observe that both the Apple having different lookupKey are shown as selected.

https://github.com/SAP/fundamental-ngx/assets/88182358/c28f291e-c321-45a8-8d75-2ac4c161a98d

Pls use [link](https://stackblitz.com/edit/angular-9qxxca-qhylee?file=src%2Fapp%2Fplatform-table-preserved-state-example.component.html,src%2Fapp%2Fplatform-table-preserved-state-example.component.ts,package.json) 

This is a customer issue for us , and we need this fix to further address the customer issue.

mikerodonnell89:
fixes #11067

### DIFF
--- a/libs/platform/src/lib/form/combobox/combobox/combobox.component.ts
+++ b/libs/platform/src/lib/form/combobox/combobox/combobox.component.ts
@@ -192,7 +192,7 @@ export class ComboboxComponent extends BaseCombobox implements ComboboxInterface
             if (this.lookupKey) {
                 return this.lookupValue(this._selectedElement) === this.lookupValue(selectedItem);
             } else {
-                return selectedItem === this._selectedElement;
+                return selectedItem.value === this._selectedElement.value;
             }
         } else {
             return false;
--- a/libs/platform/src/lib/form/combobox/combobox/combobox.component.ts
+++ b/libs/platform/src/lib/form/combobox/combobox/combobox.component.ts
@@ -188,10 +188,15 @@ export class ComboboxComponent extends BaseCombobox implements ComboboxInterface
 
     /** @hidden Define is selected item selected */
     isSelectedOptionItem(selectedItem: any): boolean {
-        return (
-            (this.lookupKey && this.lookupValue(this._selectedElement) === this.lookupValue(selectedItem)) ||
-            this.displayValue(this._selectedElement) === this.displayValue(selectedItem)
-        );
+        if (!!this._selectedElement && !!selectedItem) {
+            if (this.lookupKey) {
+                return this.lookupValue(this._selectedElement) === this.lookupValue(selectedItem);
+            } else {
+                return selectedItem === this._selectedElement;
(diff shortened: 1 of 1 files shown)


InnaAtanasova (approved):

libs/platform/src/lib/form/combobox/combobox/combobox.component.ts
```diff
@@ -188,10 +188,15 @@ export class ComboboxComponent extends BaseCombobox implements ComboboxInterface
...
-        return (
-            (this.lookupKey && this.lookupValue(this._selectedElement) === this.lookupValue(selectedItem)) ||
-            this.displayValue(this._selectedElement) === this.displayValue(selectedItem)
-        );
+        if (!!this._selectedElement && !!selectedItem) {
+            if (this.lookupKey) {
+                return this.lookupValue(this._selectedElement) === this.lookupValue(selectedItem);
+            } else {
+                return selectedItem.value === this._selectedElement.value;
+            }
```
InnaAtanasova, line 196: did you see G's comment?

mikerodonnell89: Yea, seems trivial. More than one way to skin a cat

PR #11094, opened 2023-12-14
[WIP] feat: btp user menu

mikerodonnell89:
Adds the btp user menu. Some things need work:
- [ ] adopt the a11y spec once it is finalized
- [ ] needs testing
- [ ] Keyboard support does not work great when the user menu is in "phone" mode - not sure if we consider this an actual issue or not


droshev:
closing it for now, we will reopen it once we have all the input

PR #11113, opened 2023-12-18
chore(docs): splitter docs fixes

mikerodonnell89:


g-cheishvili (approved):

InnaAtanasova (approved):

migrations.json
```diff
@@ -0,0 +1,158 @@
+{
+    "migrations": [
```
g-cheishvili, line 2: When you update nx, you also need to run the nx migrations

PR #11114, opened 2023-12-18
fix(core): do not emit scroll event if no active element

mikerodonnell89:
11096 for main


mikerodonnell89:
No longer relevant, closing

Issue #10849, opened 2023-10-31
Full name is not seen on hover in multi-input.


AlpanaSamanta:
#### Is this a bug, enhancement, or feature request?
Bug

#### Describe your proposal.
if any value in the multi-input is long, when we hover on it, we do not see the full name in tooltip

#### Can you handle that on the application side
NO

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
fundamental-ngx = 0.46.1
angular = 16.2.8

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
We use Multi- Input component

#### Please provide relevant source code (if applicable).
<label fd-label for="dropdownExampleInput1">Default Multi Input</label>
<fd-multi-input
  inputId="dropdownExampleInput1"
  [dropdownValues]="[
    'SomeLongTextToShowThatWidthWillGrowWithTheContentSomeLongTextToShowThatWidthWillGrowWithTheContentSomeLongTextToShowThatWidthWillGrowWithTheContent',
    'Banana',
    'Pineapple',
    'Tomato',
    'Kiwi',
    'Strawberry',
    'Blueberry',
    'Orange'
  ]"
  placeholder="Search here..."
  [(ngModel)]="selected"
  style="width: 200px"
></fd-multi-input>

#### Please provide stackblitz example(s).
https://stackblitz.com/edit/aipjbn?file=src%2Fapp%2Fmulti-input-dropdown-width-example.component.html

#### In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations. 
No

#### Did you check the documentation and the API?
Yes

#### Did you search for similar issues?
Yes

#### Is there anything else we should know?
No

droshev:
Actually we should wrap the text on multiple lines

skaquib123:
Hello @droshev, Should we word-wrap list item text? Is that what you are referring to?

droshev:
maybe duplicate of https://github.com/SAP/fundamental-ngx/issues/11025?

rajeev-9bit:
Hi team, please review this proposed fix: https://github.com/SAP/fundamental-ngx/pull/11121

PR #11121, opened 2023-12-20
fix(core): hover tooltip for multi input

rajeev-9bit:
closes: https://github.com/SAP/fundamental-ngx/issues/10849

## Description

Added the [title] attribute to <fd-token> & <ng-template #list>

Tested the fix on Windows edge browser. Can safely assume that the fix should work on all modern browsers.

## Screenshots

10849_hover

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

### DIFF
--- a/libs/core/multi-input/multi-input.component.html
+++ b/libs/core/multi-input/multi-input.component.html
@@ -69,6 +69,7 @@
                         [disabled]="disabled"
                         (onCloseClick)="_onTokenClick(option, false, $event)"
                         (onRemove)="_onTokenClick(option, false)"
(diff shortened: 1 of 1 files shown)


mikerodonnell89 (approved):

rajeev-9bit:
Hi @mikerodonnell89 thanks for approving this. Although it is not ready to be merged because of the two pending checks marked as required - e2e and build-lint-test. I don't exactly know how to re-trigger them or how to even check if anything went wrong with those checks. Please guide. Thanks

PR #11142, opened 2023-12-25
feat(core,platform,cx): partial signals migration

N1XUS:
## Related Issue(s)

closes none

## Description
Signals migration PR includes:
- Breadcrumbs component signal migration;
- Message Popover component signal migration;
- Core table component signal migration;
- Core dynamic page signal migration;
- DestroyRef migration;
- Platform table resettable signal migration;
- Rtl service signal migration;
- Reduced `async` pipe usage in favor of signals;
- Services are now relying on `ngOnDestroy` rather than custom destroy methods that should be called from the component.


libs/cdk/utils/directives/readonly/fdk-readonly-provider.service.ts
```diff
@@ -32,17 +33,23 @@ export class FdkReadonlyProvider extends ReplaySubject<boolean> implements Reado
...
+        this._destroy$.onDestroy(() => {
+            this.complete();
+            this.readonlyObserver.unobserve(this.elementRef);
+        });
+
         combineLatest([this._readonlyChange$, this._viewModifiers$])
             .pipe(
                 tap(([isReadonly]) => this.setReadonlyState(isReadonly)),
-                takeUntil(this._destroy$)
+                takeUntilDestroyed(this._destroy$)
```
g-cheishvili, line 45: is this._destroy$ needed here at all? It's injection context

libs/cdk/utils/directives/resize/resize.directive.ts
```diff
@@ -82,7 +83,7 @@ export class ResizeDirective implements OnChanges, AfterViewInit, OnDestroy {
     private _resizeSubscriptions = new Subscription();
 
     /** @hidden */
-    private _isRtl = false;
+    private readonly _direction$ = computed(() => (this._rtlService?.rtlSignal() ? -1 : 1));
```
g-cheishvili, line 86: Having `$` for both signal and observable might be confusing code style. maybe we should ommit the `$` in case of signal and treat signals as functions?

libs/cdk/utils/directives/selectable-list/selection.service.ts
```diff
@@ -39,7 +41,7 @@ export class SelectionService<ElementType extends Element, ValueType = ElementTy
             map((value) => (this._isMultipleMode ? value : [value[0]])),
             map((coerced: ValueType[]) => coerced.filter(Boolean))
         );
-        this._normalizedValue$.pipe(takeUntil(this._destroy$)).subscribe((val) => (this._value = val));
+        this._normalizedValue$.pipe(takeUntilDestroyed(this._destroy$)).subscribe((val) => (this._value = val));
```
g-cheishvili, line 44: It should be injection context, so no need for destroyRef here I think

libs/cdk/utils/drag-and-drop/dnd-list/dnd-list.directive.ts
```diff
@@ -122,7 +126,7 @@ export class DndListDirective<T> implements AfterContentInit, OnDestroy {
     private readonly _refresh$ = new Subject<void>();
 
     /** @hidden */
-    private readonly _onDestroy$ = new Subject<void>();
+    private readonly _onDestroy$ = inject(DestroyRef);
```
g-cheishvili, line 129: When I moved away from DestroyService to using DestroyRef, I used name `_destroyRef` for class properties, which I believe works better, because it is not neither event, nor the observable, simply a ref

libs/cdk/utils/drag-and-drop/dnd-list/dnd-list.directive.ts
```diff
@@ -149,21 +153,21 @@ export class DndListDirective<T> implements AfterContentInit, OnDestroy {
...
+        this._refresh$.next();
+        this._refresh$.complete();
     }
 
     /**
      * Refreshes the indexes of the items.
      */
     refreshQueryList(): void {
-        const refresh$ = merge(this._refresh$, this._onDestroy$);
+        const refresh$ = merge(this._refresh$, destroyObservable(this._onDestroy$));
```
g-cheishvili, line 170: Yeah, I think I messed up with `destroyObservable` function, takeUntilDestroyed(this._destroyRef)` does the same exact thing, but less maintenance from our end

libs/cdk/utils/services/keyboard-support/keyboard-support.service.ts
```diff
@@ -45,7 +47,7 @@ export class KeyboardSupportService<T> {
         this._tabKeyNavigation = tabKeyNavigation;
         this._keyManager = new FocusKeyManager(queryList).withWrap(wrap).withHomeAndEnd();
```
g-cheishvili, line 48: If this function can be called multiple times, should not we destroy the old _keyManager first before creating new one? like we do in destroy hook?

libs/cdk/utils/services/rtl.service.ts
```diff
@@ -14,12 +15,17 @@ export class RtlService {
...
+
     /** @hidden */
     constructor(@Optional() @Inject(RTL_LANGUAGE) injectedRtlLanguages: string[]) {
         injectedRtlLanguages = injectedRtlLanguages || DefaultRtlLanguages;
 
         const filtered = injectedRtlLanguages.filter((language) => navigator.language.includes(language));
 
         this.rtl = new BehaviorSubject(filtered.length > 0);
+
+        this.rtlSignal = toSignal(this.rtl);
```
g-cheishvili, line 29: Why have `undefined` in type ? It's either rtl or not. If it is because of the toSignal signature, then it has a second parameter for options where you can request to sync and have type `Signal<boolean>`

g-cheishvili (approved):

g-cheishvili:
Left few comments here and there

libs/cdk/utils/directives/readonly/fdk-readonly-provider.service.ts
N1XUS: Since we already injected DestroyRef to use it's hook, why not pass it to the `takeUntilDestroyed`? It will save us time so that `takeUntilDestroyed` does not try to inject DestroyRef again.

libs/cdk/utils/directives/resize/resize.directive.ts
N1XUS: Usualy we use `$` for reactive properties. Signals are reactive, so I don't think it's an issue. Using `$` will flag property as something that is reactive (same as Observables).

libs/cdk/utils/directives/selectable-list/selection.service.ts
N1XUS: It is, but if we don't pass DestroyRef directly, `takeUntilDestroyed` will try to inject it again

PR #11145, opened 2023-12-27
fix: 11096 for ng-16-downport branch

mikerodonnell89:


libs/core/src/lib/tabs/tab-list.component.ts
```diff
@@ -301,13 +301,15 @@ export class TabListComponent implements TabListComponentInterface, AfterContent
...
     /** @hidden */
-    _highlightActiveTab({ id }: HTMLElement): void {
-        const tab = this._tabArray.find((_tab) => _tab.panel._panelId === id);
-        if (tab) {
-            const _tabWasActive = tab.active;
-            this._activateStackedTab(tab.panel, false);
-            if (!_tabWasActive) {
-                this._selectedTabChange(tab.panel);
+    _highlightActiveTab(element: HTMLElement): void {
+        if (element) {
```
N1XUS, line 305:
```typescript
if (!element) {
    return
}
```

libs/core/src/lib/tabs/tab-list.component.ts
```diff
@@ -301,13 +301,15 @@ export class TabListComponent implements TabListComponentInterface, AfterContent
...
+            const tab = this._tabArray.find((_tab) => _tab.panel._panelId === element.id);
+            if (tab) {
```
N1XUS, line 307:
```typescript
if (!tab) {
    return;
}
```

libs/core/src/lib/tabs/tab-list.component.ts
mikerodonnell89: what's the advantage of this?

N1XUS: Less nesting and more code readability

PR #11150, opened 2023-12-29
fix(platform): table p13 selection a11y fixes

### LINKED ISSUE #10868: [P3.1][Sourcing][Accessibility][p13n dialog] Checkbox is not being read by screenreader while browsing the 'Columns' inside <fdp-table>'s p13n dialog
Bug

#### Describe your proposal.
The checkbox and its status should be announced correctly to the screenreader when accessed.

No

- Angular 15
- F-NGX Platform: 0.43.14

Please see the `Personalization Dialog` section under this example:
https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/table/p13-dialog-table

Screenshot 2023-11-01 at 5 09 32 PM

UI type, information, properties of all UI elements should be conveyed correctly once to the screen reader users when accessed.

mikerodonnell89:
fixes #10868


g-cheishvili (approved):

libs/platform/table/components/table-p13-dialog/columns/columns.component.html
```diff
@@ -75,8 +75,8 @@ <h4 fd-title [headerSize]="4">
                 (click)="_moveActiveToBottom()"
             ></button>
         </fd-toolbar>
-        <ul fd-list>
-            <li fd-list-item>
+        <ul fd-list [selection]="true" aria-multiselectable="true">
```
InnaAtanasova, line 78: `aria-multiselectable` is used in roles grid, listbox, tablist and tree. Your `ul` element should have one of these roles. Additionally, the list elements should be with role `option` and `aria-selected` set to true or false.

InnaAtanasova (changes requested):

libs/platform/table/components/table-p13-dialog/columns/columns.component.html
InnaAtanasova: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-multiselectable

InnaAtanasova: https://www.digitala11y.com/aria-multiselectable-properties/

mikerodonnell89: @InnaAtanasova the correct roles and aria-selected are being applied to the elements via the directives. If you inspect the html:
Screenshot 2024-01-02 at 5 51 54 PM

InnaAtanasova (approved):

libs/platform/table/components/table-p13-dialog/columns/columns.component.html
InnaAtanasova: maybe we should make `multiselectable` a prop of the directive in this case. Could be in a separate PR. I will approve this one.

PR #11196, opened 2024-01-08
fix(cdk): tabbable observer out of the zone

N1XUS:
## Related Issue(s)

closes none

## Description
Original issue can be observed here: https://ng-15-downport--fundamental-ngx.netlify.app/#/core/popover
The cause of the issue was with `MutationObserverFactory` which triggered zone detection, which caused `ngAfterContentChecked` callback function of popover control component to be triggered infinitely.
This fix moves the mutation observer detection out of the Angular CDR cycle allowing us to manipulate native attributes without having `detectChanges` side-effects.

### DIFF
--- a/libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
+++ b/libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
@@ -139,14 +139,17 @@ export class FocusableItemDirective implements HasElementRef {
 
     /** Set tabbable state */
     setTabbable(state: boolean): void {
-        this._tabbable = state;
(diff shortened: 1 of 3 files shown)


libs/cdk/src/lib/utils/directives/focusable-item/focusable-item.directive.ts
```diff
@@ -139,14 +139,17 @@ export class FocusableItemDirective implements HasElementRef {
 
     /** Set tabbable state */
     setTabbable(state: boolean): void {
-        this._tabbable = state;
-        this._renderer2.setAttribute(this.elementRef.nativeElement, 'tabindex', this._tabbable ? '0' : '-1');
+        return;
```
g-cheishvili, line 142: return ?

InnaAtanasova (approved):

PR #11298, opened 2024-01-22
feat(platform): vhd validator

### LINKED ISSUE #11225: [Sourcing][P1] How do we limit the selection of rows in value-help-dialog component
 Bug

#### Describe your proposal.
How do we limit the selection of rows in value-help-dialog component. Lets say user has 100 rows in value help dialog component with multiple choice enabled. but they choose to restrict the user to select only 20 maximum rows. For this to achieve we do not have any implementation from NGX component.

We can show error only after they confirm the selection and not during check box selection.

angular- 15.2
NGX - 0.43.24

N1XUS:
## Related Issue(s)

closes #11225

## Description
- Added `tokenValidator` input function that is being triggered during submission of VHD values.
- Updated default example with usage of  `tokenValidator`

### After:
image

### DIFF
--- a/libs/docs/platform/vhd/examples/platform-vhd-basic-example.component.html
+++ b/libs/docs/platform/vhd/examples/platform-vhd-basic-example.component.html
@@ -24,6 +24,7 @@
     tokenViewField="name"
     [formatToken]="formatTokenFn"
     [dataSource]="dataSource"
+    [tokenValidator]="validator"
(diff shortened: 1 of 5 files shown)


droshev (approved):

oscargong:
Hi @droshev @N1XUS,
Could you bring this change to `ng-15-downport` too? 

Thanks

PR #11304, opened 2024-01-23
fix(core): dynamic page accessibility bugs

### LINKED ISSUE #10351: Dynamic Page Header Issues
Bugs in component

1. Component needs to be verify from accessibility perspective. I found some issues:
Title/Header name should be provided as a level 2 heading (H2) like we have in UI5 Library example:
Screenshot 2023-08-17 at 10 18 57
now in example with facets there is simple text.

2. Other thing which I found is that Heading from the facets are indicated as H1 level which shouldn’t be there - and it is reserved for whole Page Title. If we want them they should be below the H2 as a H3 level.
Screenshot 2023-08-17 at 10 21 51

3. Chevron  in header pointed always the same direction in your library  but in fact it should changes depends it is collapsing or expanding should show the possible action that user can trigger - same like in the spec
https://wiki.one.int.sap/wiki/pages/viewpage.action?pageId=2039387288 
Screenshot 2023-08-17 at 10 32 57
check the UI5 implementation:
https://sapui5.hana.ondemand.com/sdk/#/entity/sap.f.DynamicPage/sample/sap.f.sample.DynamicPageFreeStyle 
Tooltips are good.

4. In UI5 library we see the Focus on a whole group the same with hover state, after clicking it we can collapse or expand as well the section below:
Check here: https://sapui5.hana.ondemand.com/sdk/#/entity/sap.f.DynamicPage/sample/sap.f.sample.DynamicPageFreeStyle 

Newest 
https://fundamental-ngx.netlify.app/#/core/dynamic-page 

This whole component needs to be reviewed by the A11y experts

mikerodonnell89:
fixes #10351 

Regarding the `[style.transform]` line, I am not sure why the arrow was always rendering pointing down previously. If you changed the previous lines to 

```
                    <i *ngIf="!collapsed" class="sap-icon--customer"></i>
                    <i *ngIf="collapsed" class="sap-icon--calendar"></i>
```

You would see the customer when the page was not collapsed, and the calendar when it is. Unclear as to why the icon would not change for `"sap-icon--slim-arrow-up"` vs `"sap-icon--slim-arrow-down"`

### DIFF
--- a/libs/core/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.html
+++ b/libs/core/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.html
@@ -14,7 +14,7 @@
             </ng-container>
 
             <div class="fd-dynamic-page__title-subtitle-container">
-                <div class="fd-dynamic-page__title-container">
+                <div class="fd-dynamic-page__title-container" role="heading" aria-level="2">
                     <span
                         class="fd-dynamic-page__title"
                         [class.fd-dynamic-page__title--wrap]="titleWrap"
--- a/libs/core/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.html
+++ b/libs/core/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.html
@@ -14,7 +14,7 @@
             </ng-container>
 
             <div class="fd-dynamic-page__title-subtitle-container">
-                <div class="fd-dynamic-page__title-container" role="heading" aria-level="2">
+                <h2 class="fd-dynamic-page__title-container">
                     <span
                         class="fd-dynamic-page__title"
                         [class.fd-dynamic-page__title--wrap]="titleWrap"
@@ -29,7 +29,7 @@
                             <ng-container *ngTemplateOutlet="dynamicPageLayoutActionsRef"></ng-container>
                         </ng-container>
                     </div>
-                </div>
+                </h2>
                 <div
                     *ngIf="_collapsed && (subtitle || _subtitleTemplate)"
                     class="fd-dynamic-page__subtitle"
--- a/libs/core/src/lib/facets/facet/facet.component.html
+++ b/libs/core/src/lib/facets/facet/facet.component.html
@@ -3,9 +3,17 @@
 </ng-container>
 
 <ng-container *ngIf="type !== 'image'">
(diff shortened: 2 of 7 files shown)


N1XUS (approved):

libs/core/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.html
```diff
@@ -14,7 +14,7 @@
             </ng-container>
 
             <div class="fd-dynamic-page__title-subtitle-container">
-                <div class="fd-dynamic-page__title-container">
+                <div class="fd-dynamic-page__title-container" role="heading" aria-level="2">
```
InnaAtanasova, line 17: Just a question: why not `h2` directly, but a div with role and aria-level?

libs/core/src/lib/facets/facet/facet.component.html
```diff
@@ -3,7 +3,9 @@
 </ng-container>
 
 <ng-container *ngIf="type !== 'image'">
-    <h1 *ngIf="facetTitle" fd-title [headerSize]="5" class="fd-margin-bottom--sm" [id]="titleId">{{ facetTitle }}</h1>
+    <span *ngIf="facetTitle" fd-title [headerSize]="5" class="fd-margin-bottom--sm" [id]="titleId">{{
+        facetTitle
+    }}</span>
```
InnaAtanasova, line 8: That's actually not a good solution. In my opinion, what Nikolay suggested is the best option:

```
For 2 you need to take the heading level hierarchy and nesting into account. If you do not know which heading level is expected according to a floorplan and the usage/nesting of the control is not clear then you need to expose an API, so that the app devs can adjust this accordingly.
```

The facet title should be something the developers can specify based on their headings order in the application.

InnaAtanasova (changes requested):

libs/core/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.html
mikerodonnell89: This was Nikolay's suggestion

InnaAtanasova: He suggested: `you need to use either a heading htmlo element or role heading with aria-level`. role heading and aria-level are used when for some reason you can't use h1, h2, etc. 
Screenshot 2024-01-25 at 9 58 50 AM


https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role

InnaAtanasova (approved):

PR #11314, opened 2024-01-26
fix(platform): live announcer should say whether column is fixed

### LINKED ISSUE #11202: [P2][Sourcing][Accessibility]Table: Frozen column status is not being conveyed by screenreader
Bug

#### Describe your proposal.
In UI5 , frozen columns are read as "Fixed" but the same is not the case with NGX table columns.
No information is mentioned whether the current column is frozen or not by the screenreader for disabled users

No

Angular: 15
core/ platform: 0.43.24

1) use the example https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/table/basic#freezable
2) Traverse through frozen columns and notice the screenreader

In UI5 example https://sapui5.hana.ondemand.com/#/entity/sap.ui.table.Table/sample/sap.ui.table.sample.TableFreeze the "Fixed" word is read when focus is on frozen column , which is helpful for disabled users 

Screenshot 2024-01-09 at 1 54 54 PM

https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/table/basic#freezable
https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/table/basic#freezable
Discussed with ATL team and reported

mikerodonnell89:
fixes #11202

### DIFF
--- a/libs/platform/src/lib/table/components/table-row/table-row.component.ts
+++ b/libs/platform/src/lib/table/components/table-row/table-row.component.ts
@@ -25,6 +25,7 @@ import {
     FDK_FOCUSABLE_ITEM_DIRECTIVE,
     FDK_FOCUSABLE_LIST_DIRECTIVE,
     FocusableItemDirective,
+    FocusableItemPosition,
     KeyUtil,
     RtlService,
     uuidv4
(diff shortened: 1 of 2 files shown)


libs/platform/src/lib/table/components/table-row/table-row.component.ts
```diff
@@ -319,4 +320,23 @@ export class TableRowComponent<T> extends TableRowDirective implements OnInit, A
     _columnTrackBy(index: number, column: TableColumn): string {
         return column.name;
     }
+
+    /** @hidden */
+    _itemFocusedEventAnnouncer(position: FocusableItemPosition): string {
```
g-cheishvili, line 325:
```suggestion
    _itemFocusedEventAnnouncer = (position: FocusableItemPosition): string => {
```

N1XUS (approved):

Issue #11328, opened 2024-01-29
Core Facets layout is broken


droshev:
#### Is this a bug, enhancement, or feature request?
bug

#### Describe your proposal.
The layout is a bit broken

<img width="1077" alt="Screenshot 2024-01-29 at 13 29 03" src="https://github.com/SAP/fundamental-ngx/assets/4380815/4ade2673-433b-45b0-b6cb-d3c6e0fe47e9">
<img width="1218" alt="Screenshot 2024-01-29 at 13 28 58" src="https://github.com/SAP/fundamental-ngx/assets/4380815/8bc4339d-88a7-4f71-a666-ebfbd2416ec9">

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)

`0.48`

PR #11329, opened 2024-01-29
fix(core): facet display

mikerodonnell89:
fixes https://github.com/SAP/fundamental-ngx/issues/11328


libs/core/facets/facet/facet.component.html
```diff
@@ -2,7 +2,9 @@
     <ng-template [ngTemplateOutlet]="customContentTemplate"></ng-template>
 } @else {
     @if (facetTitle) {
-        <span fd-title [headerSize]="5" class="fd-margin-bottom--sm" [id]="titleId">{{ facetTitle }}</span>
+        <span style="display: block" fd-title [headerSize]="5" class="fd-margin-bottom--sm" [id]="titleId">{{
+            facetTitle
+        }}</span>
```
InnaAtanasova, line 7: This alone won't fix the problem. The title of a facet needs to be a heading. It needs `role="heading"` and `arial-level="..."`.
Right now this span element is just some text that has no semantic meaning to the user, and it should be a heading.

InnaAtanasova (changes requested):

libs/core/facets/facet/facet.component.html
```diff
@@ -2,7 +2,16 @@
     <ng-template [ngTemplateOutlet]="customContentTemplate"></ng-template>
 } @else {
     @if (facetTitle) {
-        <span fd-title [headerSize]="5" class="fd-margin-bottom--sm" [id]="titleId">{{ facetTitle }}</span>
+        <span
+            [attr.role]="!!headingLevel ? 'heading' : null"
+            [attr.aria-level]="headingLevel"
+            style="display: block"
```
droshev, line 8: Can't we just use `H2` element? it will reduce the coding, no?

mikerodonnell89: No, the heading level needs to be configurable by the developer

droshev: Then can we fix this issue first in styles and then bring it here, instead of having inline styling?

droshev:
@mikerodonnell89 did you bring the fix from styles?

mikerodonnell89:
> @mikerodonnell89 did you bring the fix from styles?

Yes but unsure if I should bump the version yet

Issue #11430, opened 2024-02-19
Select Form Message


InnaAtanasova:
#### Is this a bug, enhancement, or feature request?
enhancement/feature request

#### Describe your proposal:

The message displayed in the dropdown list of the Select component is being truncated when it exceeds a certain length. This truncation poses usability issues, especially when users need to view longer messages or options.

#### Proposed Solution:
Introduce an input property that allows the user to display the message on multiple lines within the dropdown list of the Select component. This enhancement would improve the user experience by ensuring that longer messages are fully visible and comprehensible.

khotcholava:
No longer reproducible. @InnaAtanasova 

### Current Version.

![Image](https://github.com/user-attachments/assets/49886278-4f0b-4943-8488-8cad01bccf83)

khotcholava:
@do-oel Please provide a StackBlitz example to make debugging easier and clearly demonstrate the issue.

droshev:
@khotcholava did you have troubles reproducing the reported issue?

PR #11437, opened 2024-02-21
fix(core): message displayed in select component is being truncated

samyak-9bit:
#11430

Introduced text wrapping

White-space


mikerodonnell89 (changes requested):
From the issue proposed solution: Select component should have a boolean input property that makes this an option

droshev (changes requested):
@samyak-9bit there is no option to toggle on/off that option

PR #11439, opened 2024-02-21
fix(cdk): add truncate title directive to fd-combobox

### LINKED ISSUE #11346: Upon hovering over the body of combobox, no tooltip or title with selected value is displayed
Go to the standard combobox,
select a value from the dropdown,
hover on selected value in the combobox,
No tooltip or title containing selected value comes upon hovering on the body of combobox.

mikerodonnell89:
fixes #11346


libs/cdk/utils/directives/truncate/truncate.directive.ts
```diff
@@ -83,3 +83,20 @@ export class TruncateDirective implements OnChanges, AfterViewInit {
         this._truncateTarget.style.cssText = this.fdkTruncateState ? this._truncationStyle : this._defaultStyle;
     }
 }
+
+@Directive({
+    selector: '[fdkTruncateTitle], [fdTruncateTitle], [fd-truncate-title]',
+    standalone: true
+})
+export class TruncateTitleDirective implements AfterContentChecked {
```
N1XUS, line 91: I'm not sure about the directive name. It does not truncate anything, rather adds title attribute to the element if text IS truncated

N1XUS (approved):

PR #11493, opened 2024-02-29
fix(platform): idea for handling scroll when the table header overlapping focused cell

### LINKED ISSUE #11443: [P2][SOURCING] For scrollable tables, when focus is moved to cells out of view port, half of focus is visible.
Bug

#### Describe your proposal. 
For scrollable tables, when focus is moved to cells out of view port, half of focus is visible. As soon as focus goes to cells out of view port, table should scroll enough to completely get the focussed cell in the view port.

No

0.43.27

Go to virtual scroll example: https://sap.github.io/fundamental-ngx/#/platform/table/scrolling
Click on a cell inside the table
Turn on JAWS for windows or VoiceOver for mac and try navigating to cells half visible in the view port
Focus goes to such cells but only half of focus is visible.

Screenshot 2024-02-22 at 6 32 58 PM

mikerodonnell89:
fixes #11443


g-cheishvili (approved):

mikerodonnell89:
> Would [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) behave differently here?

Yes, it tries to adjust the scroll position of the whole doc page.

Issue #11208, opened 2024-01-11
Feature request: Table Scrolling Behavior with Sticky Table Toolbar and Column Headers


dominik-oelfin:
#### Is this a bug, enhancement, or feature request?
Feature request
#### Describe your proposal.
We would like to request the following table scrolling behavior for tables in CORE and PLATFORM:

_Initial situation:_
A page contains a table that has more rows than can be displayed on the viewport of the page, so the table would scroll with the page scrollbar to reveal the remaining table rows below the bottom edge of the current page-viewport.

_Issue:_
When scrolling, the table toolbar and column headers scroll out of the viewport of the page together with the upper table rows and are not visible for the user anymore.

_What we want to request:_
An option to set the table toolbar and the column headers to "sticky" to the top of the parent container (most likely the content part of a dynamic page below the dynamic page header). When the table scrolls with the page scrollbar the table toolbar and column headers would then still be visible.

_Reference from openUI5:_
This openUI5 sample shows the exact behavior that we are looking for: https://sdk.openui5.org/entity/sap.m.Table/sample/sap.m.sample.TableScrollToIndex

_Current Fundamental NGX alternative:_
Fundamental NGX offers the page scrolling behavior for tables. However, this doesn't fully fulfill our requirements since the table height has to be set to be lower than the page height to ensure that there is only one scrollbar (table level) visible and not a second scrollbar on page level, which would lead to critical UX issues. Furthermore, when displaying multiple tables on the same page, this behavior is not suitable.
#### Can you handle that on the application side
Not without a lot of customizing. Out of the box support from Fundamental NGX is desired.
#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
latest, v0.43.24
#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
N/A
#### Please provide relevant source code (if applicable).
N/A
#### Please provide stackblitz example(s).
https://stackblitz.com/edit/8tbzhh?file=src%2Fapp%2Fdynamic-page-basic-example.component.html
#### In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations. 
N/A
#### Did you check the documentation and the API?
yes
#### Did you search for similar issues?
yes
#### Is there anything else we should know?
Visualization of the feature request:
![Feature Request](https://github.com/SAP/fundamental-ngx/assets/143730992/d22233d8-3b42-40a8-b091-c61f12da5ebd)

PR #11510, opened 2024-03-05
[WIP]feat(core,platform): add page scroll to Table component

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/11208


droshev (approved):

g-cheishvili (approved):

libs/core/table/directives/table-container.directive.ts
```diff
@@ -0,0 +1,73 @@
...
+import { ResizeObserverService } from '@fundamental-ngx/cdk/utils';
+import { FD_TOOLBAR } from '@fundamental-ngx/core/toolbar';
+import { Subscription } from 'rxjs';
+import { TableHeaderDirective } from './table-header.directive';
+
+@Directive({
+    selector: '[fdTableContainer], [fd-table-container]',
+    standalone: true
+})
+export class TableContainerDirective implements OnDestroy {
```
mikerodonnell89, line 20: I think if this is only used for the edge case of stickying the header when scrolling the outside container, the name should reflect that somehow, to avoid confusion for anyone that might stumble on this directive. With this name, you would think the table is always needing this container

libs/core/table/directives/table-container.directive.ts
```diff
@@ -0,0 +1,73 @@
...
+    /** Whether the scroll is applied on the page and not on the table */
+    @Input({ transform: booleanAttribute })
```
mikerodonnell89, line 22: The boolean input is probably not necessary if this is only used with this specific edge case of sticky header/scrolling outside container. But removing the input should accompany the change of directive name i mentioned in another comment

libs/core/table/directives/table-container.directive.ts
```diff
@@ -0,0 +1,73 @@
...
+    set toolbar(toolbarElRef: ElementRef) {
+        if (this.outerScroll && toolbarElRef) {
+            const toolbarEl = toolbarElRef.nativeElement;
+
+            if (this._resizeSubscription) {
+                this._resizeSubscription.unsubscribe();
+            }
+
+            this._resizeSubscription = this.resizeObserverService.observe(toolbarEl).subscribe(() => {
+                this._renderer.setStyle(toolbarEl, 'position', 'sticky');
```
mikerodonnell89, line 36: Why do we want to set all of these styles on a resize? A resize should be irrelevant to this in my opinion

InnaAtanasova: the position top of the Table header depends on the height of the Toolbar. In cozy mode the height of the toolbar is 2.75rem so the top of the table header should be 2.75rem. When we change the mode to compact, the height becomes 2rem so respectively the top of the Table header becomes 2rem. Initially I made a new subscription that was tracking the change of the toolbar height but saw that we already have one so I am using it to detect changes there and set the proper positions.

mikerodonnell89 (approved):
A basic unit test for the new container would be nice to have but because the directive is so simple, happy to merge this now and come back to that later

Issue #11409, opened 2024-02-14
fd-form-input-message-group block the Space key click event in rich text editor 


wei368:
#### Is this a bug, enhancement, or feature request?
Bug 
#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)

  "@fundamental-ngx/cdk": "0.43.21",

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;

1. enter some text in the rich text editor
2. make the text as bold by clicking the "B" tool bar
3. when tabbing to the 'B' toolbar and pressing the space key on the keyboard, the toolbar isn't toggled, and the bolded text remains unchanged.

The problem is that we have input-message-group in the out side of sap-rich-text-editor.

If we remove input-message-group, the accessibility works fine. but
If we have input-message-group, the accessibility does not work

I updated the sample code, please check out the code and try again. thanks

Here is the url for the sample code https://github.tools.sap/I851026/editor-rich-text-editor

<fd-form-input-message-group fillControlMode="equal" [triggers]="['mouseover','mouseout']">

#### Please provide relevant source code (if applicable).

Here is the url for the sample code https://github.tools.sap/I851026/editor-rich-text-editor

#### Is there anything else we should know?

rich text editor team has looked the issue. and they think the problem in in ngx

droshev:
@g-cheishvili Can you check this issue?

g-cheishvili:
I tested it. Provided fix. you can add input `[preventSpaceKeyScroll]="false"` and it should work

wei368:
thanks @g-cheishvili, I tested by using the above same code and our app code, it did not fix the problem by adding [preventSpaceKeyScroll]="false". Can you provide some more detail guidance. thanks a lot.

![Screenshot 2024-03-05 at 3 16 03 PM](https://github.com/SAP/fundamental-ngx/assets/5350405/7ed8e058-687e-4f5e-878b-428b1662123b)

g-cheishvili:
> thanks @g-cheishvili, I tested by using the above sample code and our app code, it did not fix the problem by adding [preventSpaceKeyScroll]="false". Can you provide some more detail guidance. thanks a lot.  Here is where I put 
> 
> ![Screenshot 2024-03-05 at 3 17 27 PM](https://github.com/SAP/fundamental-ngx/assets/5350405/25808cef-072e-4957-97fc-a59196f19cf4)
> 
> 

[...]

PR #11514, opened 2024-03-05
fix(core): disable form message group prevent defaults

g-cheishvili:
## Related Issue(s)

part of #11409


mikerodonnell89 (commented):
Looks like this works fine but the linter is upset

InnaAtanasova (approved):

PR #11520, opened 2024-03-06
fix(platform): add validator input for p13 filter component

### LINKED ISSUE #11400: [P2][Sourcing]Need validation handling for filters in fdp-table
Feature request

#### Describe your proposal.

Would like to have validation emit or options for filters in exclude and include section of fdp-table, currently conflicting conditions can be added to include and exclude.
in case of error filter pop should have error message.
e.g. INCLUDE :  include users with mobile numbers, EXCLUDE users with mobile numbers

No

V15.x

mikerodonnell89:
fixes #11400 

Provides the application developer with the means of handling their own validation for the personalization dialog filter rules. Developer provides a function that returns a boolean for valid/invalid. Example code has been provided in the personalization dialog filter example


droshev (changes requested):
Can you add an example where the user has the choice (like in value-help dialog) where they can return back to the dialog and correct the input or cancel the change from the dialog?

droshev:
> https://deploy-preview-11520--fundamental-ngx.netlify.app

@mikerodonnell89 I meant something like in the value help with 2 options
Screenshot 2024-03-15 at 07 57 34

mikerodonnell89:
> > https://deploy-preview-11520--fundamental-ngx.netlify.app
> 
> @mikerodonnell89 I meant something like in the value help with 2 options Screenshot 2024-03-15 at 07 57 34

Well that is up to the application developer, and it doesn't even make sense from a UX perspective to have these two different buttons, they do the same thing which is simply dismissing the dialog

droshev (approved):

PR #11535, opened 2024-03-11
fix(platform): add validator input for p13 filter component

### LINKED ISSUE #11400: [P2][Sourcing]Need validation handling for filters in fdp-table

mikerodonnell89:
fixes #11400


droshev (changes requested):
see https://github.com/SAP/fundamental-ngx/pull/11520#pullrequestreview-1932161268

droshev (approved):

PR #11563, opened 2024-03-15
fix(platform): vhd keyboard improvements

### LINKED ISSUE #11408: [Sourcing][P2][Accessibility] Value Help Dialog: Selected token is not read by screen reader
bug / enhancement

#### Describe your proposal.

Selected token is not read by screen reader	when accessing the multi-input field 
 
- eg. we should have one stop here, and reading out "Selected Items and Conditions 2 Name 1  Name 2 "
  - Oscar ScreenCapture 2024-03-12 at 02 43

Only the side blank can be focus
- The token should be focusable too.

Token cannot be deselected from keyboard
- we also cannot deselected a row from table

image

- this is happening on V0.48.4 
- but we are using 0.43 on our application

mikerodonnell89:
fixes #11408


droshev (changes requested):
I don't see a bug to be fixed

PR #11606, opened 2024-03-20
chore(deps,deps-dev): restructuring and updates

g-cheishvili:
## Related Issue(s)

Continuation of #11048

## Description

- Removed targets `lint` and `test` from sublibraries and deleted related files. Moved them into host libraries
- Fixed some cases of the circular deps and the incorrect not initialized variable usages
- Shortened the tsconfig.base.json pathes
- Updated Angular and Nx to the latest versions
- Updated browserlist to the latest
- Angular 17.0.0 based application is not compatible with Angular 17.3.0 features, so min version of our lib becomes 17.3.0


package.json
```diff
@@ -1,176 +1,177 @@
...
+        "nx": "nx",
+        "postinstall": "node ./decorate-angular-cli.js",
+        "prepare": "husky install",
+        "pre-commit": "lint-staged"
+    },
+    "publishConfig": {
+        "registry": "https://repository.hybris.com/api/npm/npm-repository/"
+    },
+    "dependencies": {
+        "@angular/animations": "17.3.0",
```
droshev, line 41: would that be a problem for applications staying on version 17.0.x, 17.1.x, 17.2.x?

droshev (approved):

package.json
g-cheishvili: Hm, I think I have to test it out

g-cheishvili: Checked it, although at the moment it will work without any issues, when we will be using new query signals or other features that are coming with 17.3, users who use 17.0-17.2 will not be able to build their applications

PR #11608, opened 2024-03-20
fix(platform): table empty cell screenreader text

### LINKED ISSUE #11596: [P1][Sourcing][Accessibility]Table: The empty cell information is not conveyed through screenreader
Bug,
Since the issue #10320 is no addressed.
Reopening as new issue.
the word"empty" is not being conveyed when user is on blank or empty cell without any value.
This is happening both in JAWS 2023 and JAWS 2024

#### Describe your proposal.
When user is using screenreader either jaws 2023 or 2024 version and traversing through empty cells, then the screenreader is not conveying empty cell information.
This should work both in VPC mode on and off of jaws.

No

Angular 15
Core/ Platform: 0.43.29

can be replicable on NGX website : https://sap.github.io/fundamental-ngx/#/platform/table/basic
can be replicable on NGX website : https://sap.github.io/fundamental-ngx/#/platform/table/basic

mikerodonnell89:
fixes #11596

### DIFF
--- a/libs/platform/src/lib/table/components/table-row/table-row.component.html
+++ b/libs/platform/src/lib/table/components/table-row/table-row.component.html
@@ -156,7 +156,7 @@
 
         <ng-container *ngIf="row.state === 'readonly'; else editModeCell">
             <span
-                *ngIf="tableTextContainer.innerText.trim() === ''"
+                *ngIf="tableTextContainer?.innerText?.trim() === ''"
                 style="
                     position: absolute !important;
                     height: 1px;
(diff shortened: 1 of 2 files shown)


libs/platform/src/lib/table/components/table-row/table-row.component.html
```diff
@@ -156,7 +156,7 @@
 
         <ng-container *ngIf="row.state === 'readonly'; else editModeCell">
             <span
-                *ngIf="(row.value | valueByPath : column.key) === '' || (row.value | valueByPath : column.key) === null"
+                *ngIf="tableTextContainer.innerText.trim() === ''"
```
g-cheishvili, line 159: `innerText` is `undefined` in JSDom, hence issues with unit tests, you either need to fix unit tests with mocking element innerText or do this differently

mikerodonnell89: > "There might be glitches between the detection cycles" 

Agreed but this will not be an issue for screenreader users which is what this issue is addressing

droshev (changes requested):
Let's take the time and come up with a fix that doesn't affect change detection

PR #11609, opened 2024-03-20
[WIP]test: Calendar, Input Group, Vartical Nav

kalo2711:
Calendar :Programmatic Date Change, Input Group Search and Vertical Navigation "Condensed Mode" have been fixed of compile issues.

### DIFF
--- a/libs/docs/core/input-group/examples/input-group-search-example/input-group-search-example.component.ts
+++ b/libs/docs/core/input-group/examples/input-group-search-example/input-group-search-example.component.ts
@@ -3,13 +3,18 @@ import { FormsModule } from '@angular/forms';
 import { FormItemComponent, FormLabelComponent } from '@fundamental-ngx/core/form';
 import { InputGroupModule } from '@fundamental-ngx/core/input-group';
 
-@Component({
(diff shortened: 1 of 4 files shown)


libs/docs/core/input-group/examples/input-group-search-example/input-group-search-example.component.ts
```diff
@@ -3,13 +3,18 @@ import { FormsModule } from '@angular/forms';
...
-@Component({
-    selector: 'fd-input-group-search-example',
-    templateUrl: './input-group-search-example.component.html',
-    standalone: true,
-    imports: [FormItemComponent, FormLabelComponent, InputGroupModule, FormsModule]
-})
-export class InputGroupSearchExampleComponent {
-    searchTerm = 'Search Term';
-    searchTermSecond = 'Search Term';
+
```
droshev, line 6: nothing was changed, you need to revert it

kalo2711:
The fixes are the same for Input Group, the code just doesn't show in the input-group-search-example-component.ts, this is why the stackblitz isn't wokring, but it is not an issue with the code itself

PR #11643, opened 2024-03-26
fix: issue with scrolling when topEl would be the cell instead of row

mikerodonnell89:
fixes none


g-cheishvili:
It still does the same

droshev:
guys, can you summarize the issue and why we are fixing it?

g-cheishvili:
> guys, can you summarize the issue and why we are fixing it?

Entire page scrolls when you navigate the focusable list. It is visible on core table focusable example

droshev:
then we should have a bug reporting that. I would like to avoid getting PRs without real bugs, which revert after.

g-cheishvili (approved):

PR #11644, opened 2024-03-26
fix: issue with scrolling when topEl would be the cell instead of row

mikerodonnell89:
fixes none


g-cheishvili (dismissed):

droshev:
guys, can you summarize the issue and why we are downporting?

mikerodonnell89:
> guys, can you summarize the issue and why we are downporting?

There was an issue where a table cell's focus was out of view due to the sticky header overlapping a cell. I opened a PR to fix this by scrolling the table when the header overlapped the focus cell. This introduced a bug where the whole page would scroll when using focus on the entire table row. This fixes that. Downporting to ng-15 because the bug was introduced there as well.

Issue #10698, opened 2023-10-11
[P3.2][Sourcing][Accessibility][ATL testing] Unnecessary chrome tab name is read on focus of dialog


I543348:
#### Is this a bug, enhancement, or feature request?
Bug

#### Describe your proposal.
On open of value help dialog https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/value-help-dialog, when the focus is at the first interactive element inside dialog, the browser tab name is being read unnecessarily.

Expectation: The browser's tab name should not be read when focus is inside value help dialog as its out of context for users

#### Can you handle that on the application side
No

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
Angular 15
core/ platform: 0.43.9

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
1) open link https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/value-help-dialog
2) access any example
3) turn on JAWS
4) trigger the value help button
5) hit enter to open dialog
6) notice the jaws speech history reading "Fundamental Library for Angular -Google chrome dialog"
![MicrosoftTeams-image (5)](https://github.com/SAP/fundamental-ngx/assets/88182358/7dcf405d-0213-4e63-b492-761071b16912)

#### In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations. 
Yes as it is reported by ATL team. the issue is discussed with experts and reported here.

droshev:
@I543348 This is nothing to do with the application. The documentation is not meant to be fully validated against your expectation

I543348:
Hi @droshev , Its not only happening in NGX website, but also in guided sourcing.
We would not raise any such issues which are not happening in our Guided sourcing application.
Could you pls reopen? 
This issue is tested and reported in Guided sourcing first by ATL team , after cross checking the same in NGX website , I have reported the issue.

Thanks.
cc: @sagar7259, @nikolay-kolarov

droshev:
@I543348 you need to inspect and see where that string is coming from.

sagar7259:
Hi Team,

When i analyzed this using forms mode and reading mode of JAWS, I got different results.

In reading mode:
The parent screen title "Fundamental library for angular - google chrome" is being read as an additional information. along with the dialog title.

[...]

PR #11650, opened 2024-03-27
fix(platform): dialog aria-labelledBy for vhd

### LINKED ISSUE #11377: [P3.2][Sourcing][Accessibility][Dialog] Reopening of 10698
Bug

#### Describe your proposal.
This issue is created as https://github.com/SAP/fundamental-ngx/issues/10698 was not being addressed for dialogs such as fd-dialog and value help dialog are reading chrome tab name

No

Angular: 15
core/platform: 0.43.26

Fd-dialog and value help dialog pls refer https://github.com/SAP/fundamental-ngx/issues/10698 for more info

mikerodonnell89:
fixes #11377 and #10698


libs/platform/value-help-dialog/value-help-dialog/value-help-dialog.component.ts
```diff
@@ -193,7 +195,7 @@ export class PlatformValueHelpDialogComponent<T = any> extends VhdComponent impl
 
     /** Id attribute for dialog title */
     @Input()
-    headerId: string | null = null;
+    headerId = `fd-vhd-header-unique-id-${vhdHeaderUniqueId++}`;
```
InnaAtanasova, line 198: not a big deal, but in my opinion the prefix should be `dfp-`.
Do we need the word `unique`? Can we shorten the id to `fd-vhd-header-id-${vhdHeaderUniqueId++}`?

g-cheishvili (approved):

PR #11662, opened 2024-03-29
fix(platform): focus menu button trigger when closed

### LINKED ISSUE #11398: [Sourcing][P3.1][Accessibility] `fdp-menu-button` lost focus after close the menu
Bug

#### Describe your proposal.
Use keyboard, after close `fdp-menu-button`, the focus back to browser URL bar

What is expected:
the visible focus should always follow exact keyboard commands and the keyboard focus should be always present on the same UI from where it was activated or closed and never be lost.

No

```
"@fundamental-ngx/cdk": "0.48.4",
  "@fundamental-ngx/platform": "0.48.4",
  "@fundamental-ngx/core": "0.48.4",
  "@fundamental-ngx/i18n": "0.48.4",
```

https://github.com/SAP/fundamental-ngx/assets/7035787/0d6d008b-e825-400d-a9ac-206e2fa42534

https://sap.github.io/fundamental-ngx/#/platform/menu-button

https://sap.github.io/fundamental-ngx/#/platform/menu-button

Yes I did

mikerodonnell89:
fixes #11398


g-cheishvili:
I would do this a little differently, because it is easy to forget removal of this patch. This is what I would do:
- inject(MenuButtonComponent, {host: true, optional: true}) in MenuTriggerDirective
- in MenuTriggerDirective._destroyMenu instead of doing what you do, would just do this._menuButtonComponent?.focus()
- and in MenuButtonComponent would add a function, called `focus` and in that function would focus on button whenever it's called

The benefit of this is not only that you no longer have selector hardcoded, but also that when the MenuButton is removed from the code, your compilation fails and you will not have dead/unused code anywhere

InnaAtanasova (approved):

Issue #11640, opened 2024-03-26
Floating Footer in Flexible Column Layout


julia-siegl:
#### Is this a bug, enhancement, or feature request?
Enhancement/ Bug

#### Describe your proposal.
When using the floating footer bar design in a dynamic page inside column layout, it would be great if each column gets is own footer toolbar instead of a overall footer toolbar spanning several columns to be in alignment with guidelines. Currently the code, e.g. used   in example https://sap.github.io/fundamental-ngx/#/core/dynamic-page#layout:
```
<fd-dynamic-page-footer>
    <!-- page footer content goes here -->
        <div fd-bar barDesign="floating-footer">
            ....
       </div>
</fd-dynamic-page-footer>
```

leads to an footer that is spanned over multiple columns:
![image](https://github.com/SAP/fundamental-ngx/assets/143191726/55396ec0-be44-4e4d-a275-f9f232b3c6bd)
([source](https://sap.github.io/fundamental-ngx/#/core/dynamic-page#layout) is library example)

Guideline:

<img width="539" alt="image" src="https://github.com/SAP/fundamental-ngx/assets/143191726/968b43bc-52f3-4f88-8db3-e6bc54a4c6d4">

([source](https://experience.sap.com/internal/fiori-design-web/flexible-column-layout/#footer-toolbar))

#### Can you handle that on the application side
Implementing dynamic resizing/positioning on application side seems like a workaround.
Can use the 'normal' footer ([here](https://sap.github.io/fundamental-ngx/#/core/bar#footer)), which is placed as expected, but not ideal.

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)

"@angular/core": "^17.1.1",
"@fundamental-ngx/core": "^0.49.0-rc.24",

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
- Check out library example:
https://sap.github.io/fundamental-ngx/#/core/dynamic-page#layout

#### Did you check the documentation and the API?
yes, issue also in the documentation, API in bar did not change floating footer ( e.g. tried _inPage_, _inHomePage_, _size_)...

#### Did you search for similar issues?
yes, nothing found

droshev:
example https://stackblitz.com/edit/1hqnwx

InnaAtanasova:
Hi @julia-siegl you can add a very quick fix to your project by adding position relative to the dynamic page. 
If you add 
```
[...]

PR #11678, opened 2024-04-01
fix(core): relative position property for Dynamic Page

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/11640

## Description
The floating footer of Dynamic page occupies the entire width of the page. When Dynamic page is placed inside Flexible Column Layout the footer needs to span within the column that Dynamic page occupies. For this to work we need to set the position of Dynamic page to relative. I added an input property that could allow this

### Before:
Screenshot 2024-04-01 at 3 15 57 PM

### After:
Screenshot 2024-04-01 at 2 35 23 PM

### DIFF
--- a/libs/core/dynamic-page/dynamic-page.component.html
+++ b/libs/core/dynamic-page/dynamic-page.component.html
@@ -8,6 +8,7 @@
     [class.fd-dynamic-page--transparent-bg]="background === 'transparent'"
     [class.fd-dynamic-page--list-bg]="background === 'list'"
     [attr.aria-label]="ariaLabel"
+    [style.position]="positionRelative ? 'relative' : null"
 >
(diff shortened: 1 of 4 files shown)


mikerodonnell89 (commented):
Works well - but I wonder if this is something that we can handle via ContentChildren without needing the developer to add an input

InnaAtanasova:
> Works well - but I wonder if this is something that we can handle via ContentChildren without needing the developer to add an input

Open to hear your ideas. Dynamic Page can be placed in so many different containers, with scrolling, without scrolling, flexible column layout, etc. We also need to take into account all the existing apps already using Dynamic Page. Adding position relative can mess up so many things. At least this is my opinion. Happy to hear more elegant solutions to this problem.

droshev (approved):

PR #11680, opened 2024-04-02
fix(core): fixed missing role on focusable dynamic page element

### LINKED ISSUE #11253: [Sourcing-2][Accessibility: ACC-264.1] Role attribute not provided in FD-DYNAMIC-PAGE
Bug

#### Describe your proposal.
Component having issue: https://sap.github.io/fundamental-ngx/#/platform/dynamic-page

At a basic level, all custom controls within a web application should provide information about the component name, state and type. The name of the component is the principal piece of information and should come first in the textual representation of the element. When custom elements do not expose textual equivalents, users who can not see the element or users who can not see color may not be able to fully interact with the element. For example, without knowing the role or state of an element, the user of a screen reader may not know what actions can be performed on an element even if the element exposes a textual name.

This FD-DYNAMIC-PAGE-HEADER is focusable but lacks a role attribute.

Code Snippet: <div tabindex="0" class="fd-dynamic-page__title-area" title="Shared Keys"><div class="fd-dynamic-page__main-container"><div class="fd-dynamic-page__breadcrumb-title-container"><div class="fd-dynamic-page__breadcrumb-container"><nav class="fd-dynamic-page__breadcrumb-wrapper" aria-label="Breadcrumbs"></nav></div><div class="fd-dynamic-page__title-container"><div class="fd-dynamic-page__title-subtitle-container"><div class="fd-dynamic-page__title-container"><span class="fd-dynamic-page__title" fdignoreclickonselection="">Shared Keys</span><div class="fd-dynamic-page__toolbar-container"></div></div></div></div></div></div></div>

No

All latest version

Using Access assistant easily can check.
1.  Login with the URL Provided.
2. Activate 'shared keys' app.
•	Download Access assistant tool from level access.
•	For registered AMP user, login with your SAP email ID into SAP AMP instance: sap.levelaccess.net

Yes

g-cheishvili:
## Related Issue(s)

closes #11253

## Description

- Removed tabindex from the header
- Added tabindex to the title element and the role of heading to it. When it gets focus through mouse or the tab it will activate the focus design to not break the styling
- Removed h2 and used div instead, because aria-level and heading is now the actual title that is in it

### DIFF
--- a/libs/core/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.html
+++ b/libs/core/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.html
@@ -13,10 +13,15 @@
                 <ng-content select="fd-facet"></ng-content>
             }
             <div class="fd-dynamic-page__title-subtitle-container">
-                <h2 class="fd-dynamic-page__title-container">
+                <div class="fd-dynamic-page__title-container">
                     <span
+                        tabindex="0"
+                        role="heading"
+                        aria-level="2"
                         class="fd-dynamic-page__title"
                         [class.fd-dynamic-page__title--wrap]="titleWrap"
+                        (focus)="headingFocused$.set(true)"
+                        (blur)="headingFocused$.set(false)"
                         fdkIgnoreClickOnSelection
                     >
                         <ng-template [ngTemplateOutlet]="_titleTemplate ? titleRef : defaultTitle"></ng-template>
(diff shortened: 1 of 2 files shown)


mikerodonnell89 (approved):

libs/core/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.html
```diff
@@ -13,10 +13,15 @@
                 <ng-content select="fd-facet"></ng-content>
             }
             <div class="fd-dynamic-page__title-subtitle-container">
-                <h2 class="fd-dynamic-page__title-container">
+                <div class="fd-dynamic-page__title-container">
                     <span
+                        tabindex="0"
+                        role="heading"
+                        aria-level="2"
```
mikerodonnell89, line 20: I think this probably needs to be either aria-level 1 (assuming the page will be fullscreen), or, configurable via input by the dev

mikerodonnell89 (changes requested):
See my comment about aria-level, thanks!

droshev:
We need to handle the markup/styling changes in styles first.

Issue #11685, opened 2024-04-03
[P2][Sourcing] Breadcrumb does not pass click when collapse in the menu


oscargong:
#### Is this a bug, enhancement, or feature request?
enhancement or feature request

#### Describe your proposal.
In `breadcrumb-item.component.ts` :

```typescript
get _needsClickProxy(): boolean {
        return  !!this.breadcrumbLink?.elementRef.nativeElement.getAttribute('href') || !!this.breadcrumbLink.routerLink;
    }
```

But we cannot use simply use <code>href</code> attribute or <code>routerLink</code> directive. We are handling click event manually in the `fd-breadcrumb-item`

However, If breadcrumb is in a menu, `(click)` event won't be trigger

#### Can you handle that on the application side
No workaround from my application

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
V0.49.3

#### Please provide relevant source code (if applicable).
```html
<fd-breadcrumb class="breadcrumb-style" *ngIf="containsCurrentInBreadcrumbs() ? breadcrumbs.length > 1 :  breadcrumbs.length >0">
    <ng-container *ngFor="let bc of breadcrumbs">
        <fd-breadcrumb-item>
            <a *ngIf="bc.routerLink" fd-breadcrumb-link (click)="breadscrumbClick(bc)">{{bc.name}}</a>
        </fd-breadcrumb-item>
    </ng-container>
</fd-breadcrumb>
```

#### In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations.
Not an ADA issue

#### Did you check the documentation and the API?
Yes

#### Did you search for similar issues?
Yes

#### Is there anything else we should know?

**_IMPORTANT: Please refrain from providing links or screenshots of SAP's internal information, as this project is open-source, and its contents are accessible to anyone._**

droshev:
@oscargong We couldn't not understand the problem and the use case. Do you think you can elaborate more?

oscargong:
Hi @droshev 

I made a stackblitz example:

https://stackblitz.com/edit/e2njgi?file=src%2Fapp%2Fbreadcrumb-routerLink-example.component.html

Also a screen recording:

https://github.com/SAP/fundamental-ngx/assets/7035787/72474bab-9cc1-4b7f-a0ae-10a1148a2379

[...]

PR #11686, opened 2024-04-03
[WIP]fix(core): add `needsClickProxy` input for fd-breadcrumb-item to pass click event

oscargong:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/11685

## Description

When view size is not enough, breadcrumb will be pushed into menu. If breadcrumb is in a menu, before, we only pass click event if there is a <code>href</code> attribute or <code>routerLink</code> directive.

Changes:
- Add new `needsClickProxy` Input to pass the click event
- Add a new document example


droshev:
We need to discuss the issue and see whether there is another option?

PR #11691, opened 2024-04-04
fix(core): remove parent row tabindex when cell is focused to address screenreader issue

### LINKED ISSUE #11376: [P2][Sourcing][Accessibility][Tree table] Levels are not being read correctly and creates confusion for users as it displays only numbers 1 of 4 and so on
Bug

#### Describe your proposal.
This is in continuation of issue https://github.com/SAP/fundamental-ngx/issues/10618 , here the reading "1 of 4" , "2 of 4" creates confusion for users as users wont know 1 means first parent node and 4 means total children in that parent node.
Expectation is to similar to value help dialog's tree table reading where we get a word "level" 
for eg:
Row 1 of 6, level 2 and so on... 

No

Angular : 15
core/ platform: 0.43.26

Its replicable on https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/table/basic#tree with 0.43.26

https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/table/basic#tree

pls refer tree table in 0.43.26 example https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/table/basic#tree

This is reported by ATL team

mikerodonnell89:
fixes #11376 

Focusable grid/list uses tabindex to handle navigation via keyboard, not only for tab presses but also arrow keys. When a parent element has a tab index, and its children have tab indexes, the screenreader will read "x of y" when focusing a child, where x is the index and y is the number of children. This PR removes the parent row's tabindex when a cell receives focus, and puts it back when the cell is blurred.

### DIFF
--- a/libs/core/src/lib/table/directives/table-cell.directive.spec.ts
+++ b/libs/core/src/lib/table/directives/table-cell.directive.spec.ts
@@ -7,10 +7,12 @@ import { TableCellDirective } from './table-cell.directive';
 
 @Component({
     template: `
-        <td fd-table-cell>
-            <fd-checkbox></fd-checkbox>
-        </td>
-        <td fd-table-cell [key]="key">{{ key }}</td>
+        <tr tabindex="-1">
+            <td fd-table-cell>
+                <fd-checkbox></fd-checkbox>
+            </td>
+            <td fd-table-cell [key]="key">{{ key }}</td>
+        </tr>
     `
 })
 class TestComponent {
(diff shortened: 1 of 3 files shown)


droshev (changes requested):
@mikerodonnell89 Can you check the changes in JAWS, verify whether they match what the expected announcement is and document it so we can test it after?

droshev (approved):

Issue #10961, opened 2023-11-15
SelectComponent ignores reactive `isDisabled` changes


ZsDeak-SAP:
### Is this a bug, enhancement, or feature request?
bug

### Describe your proposal.
When the reactive form uses `enable()`, or `disable` on a FormControl that's tied to a SelectComponent, the `isDisabled` handler is correctly called in the component, but the state does not change automatically on the view.

Also your StackBlitz pages don't seem to work at the moment, they have a compile/install error.

### Can you handle that on the application side
Only if I recreate the component with a different default disabled state, but i wouldn't call that even a workaround. I can also use the `disabled` input on the component, but that'll throw Angular warnings about using both reactive and non-reactive methods together.

### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.) \
Checked out fundamental-ngx 0.47.1, latest master at the moment of writing to be precise.

### If this is a bug, please provide steps for reproducing it; the exact components you are using;
Click the toggle
![toggle-disabled](https://github.com/SAP/fundamental-ngx/assets/115458873/622dec40-7a37-456b-9ab9-88a455a224ee)

### Please provide relevant source code (if applicable).
```typescript
@Component({
    selector: 'fd-select-forms',
    templateUrl: './select-forms.component.html',
    styleUrls: ['select-forms.component.scss'],
    standalone: true,
    imports: [FormsModule, ReactiveFormsModule, SelectModule, NgFor]
})
export class SelectFormsComponent {
    options: string[] = ['Apple', 'Pineapple', 'Tomato', 'Strawberry'];

    disabledForm = new FormGroup({
        disabledControl: new FormControl({ value: 'Apple', disabled: true }, Validators.required)
    });

    toggleDisabled() {
        const control = this.disabledForm.get('disabledControl');
        if(!control) return;
        if(control?.disabled) {
            control.enable();
            return;
        }
        control.disable();
    }
}
```
```html
<h4>Disabled select</h4>
<form [formGroup]="disabledForm">
    <fd-select formControlName="disabledControl">
        <li fd-option *ngFor="let option of options" [value]="option">{{ option }}</li>
    </fd-select>
</form>
<button (click)="toggleDisabled()">Toggle disabled</button>
<small style="display: block">Value: {{ disabledForm.controls.disabledControl.value }}</small>
<small style="display: block">Status: {{ disabledForm.status }}</small>
```

### Did you check the documentation and the API?
Y
[...]

PR #11693, opened 2024-04-05
fix(core): step input disabled state

ZsDeak-SAP:
## Related Issue(s)

Same problem as #10961  only with the step-input

Same fix as #10994 

Note: using `detectChanges` is more performant in such localised changes for low level components. [`markForCheck`](https://github.com/angular/angular/blob/8cd1b2323ac32ad09ec1a7f97a6413c93b7efe6f/packages/core/src/render3/instructions/mark_view_dirty.ts#L15) would run change detection for all ancestors (eventually), while `detectChanges` runs only for this component and its children, which are very few.


ZsDeak-SAP:
closed in favor of https://github.com/SAP/fundamental-ngx/pull/11724

PR #11738, opened 2024-04-11
fix(core): keyboard enhancements for grid list

### LINKED ISSUE #10062: Grid List Keyboard support
Improve the Grid List Support following the keyboard spec for the component.
`Up`/`Left` - move focus to the previous item (except if the focus is on the first, then do nothing)
`Down`/`Right` - move focus to the next item (except of the focus is on the last, then do nothing)
`Shift` + `Left`/`Up` - Starts or extends a range selection to the bottom/end of the list
`Shift` + `Right`/`Down` - Starts or extends a range selection to the top/beginning of the list.
`Shift` + `Home` - Start or extend range selection to the first item of the Grid List.
`Shift` + `End` - Start or extend range selection to the last item of the Grid List.

p. 246

FIORITECHP1-27741

mikerodonnell89:
fixes #10062


mikerodonnell89:
No need for this on ng-15

PR #11878, opened 2024-05-03
fix(platform): check validator before open Filtering Settings Dialog

oscargong:
PR https://github.com/SAP/fundamental-ngx/pull/11535 added validator input for p13 filter component. 

But it caused a regression, when there is no  `<fdp-table-p13n-filter [validator]="validator" />` inside 
```html
<fdp-table-p13-dialog [table]="table"> </fdp-table-p13-dialog>
``` 
Table filter cannot be opened.

This issue can be reproduced on both 0.49 and 0.43
- StackBlitiz https://stackblitz.com/edit/p6cmyh?file=src%2Fapp%2Fplatform-table-p13-filter-example.component.html


mikerodonnell89 (approved):

mikerodonnell89 (approved):
@oscargong Did you commit with `--no-verify`? The linter was throwing an error. I have pushed a fix. Thanks

Issue #11921, opened 2024-05-16
Show Clear button on Combobox without having it as a search


abhijith-darshan:
#### Is this a bug, enhancement, or feature request?

**Enhancement**

#### Describe your proposal.

In `Combobox` the clear button is only shown if both `isSearch` and `showClearButton` are set to `true`

In `UI5` it is possible to show clear button on a non `isSearch` mode as well as shown below

https://sapui5.hana.ondemand.com/#/entity/sap.m.ComboBox/sample/sap.m.sample.ComboBoxClearIcon

<img width="598" alt="image" src="https://github.com/SAP/fundamental-ngx/assets/137736216/6047777e-0e55-4b70-801e-495ec3e5dc7f">

#### Can you handle that on the application side

No due to `&&` condition in `combobox.component.html`

https://github.com/SAP/fundamental-ngx/blob/494e77f349bb84b0975f3a58400d19d5a22b64fa/libs/core/combobox/combobox.component.html#L92

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)

**Current version**

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;

**N/A**

#### Please provide relevant source code (if applicable).

**N/A**

#### Please provide stackblitz example(s).

**N/A**

#### In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations.

**N/A**

#### Did you check the documentation and the API?

**Yes**

#### Did you search for similar issues?

**Yes**

#### Is there anything else we should know?

**No**

**_IMPORTANT: Please refrain from providing links or screenshots of SAP's internal information, as this project is open-source, and its contents are accessible to anyone._**

abhijith-darshan:
I think setting it as an `OR ||` condition would fix this -

```typescript 
....
@if ((isSearch || showClearButton) && inputText && inputText.length > 0)
....
```

Consequence would be that when **Combobox** is used in non search mode OR in search mode, by default the clear button is shown. But can be overridden with `showClearButton` flag with `false`

PR #11924, opened 2024-05-18
fix(core): show clear button in any combobox

droshev:
BREAKING CHANGE:
showClearButton set default to false

fixes: #11921

before: showClearButton only when the combobox is used as search field
after: showClearButton can be used independently and can be set to any type


InnaAtanasova (changes requested):
Need an example in the documentation

droshev:
> Need an example in the documentation

@InnaAtanasova Thank you for the review: having a documentation example is a great idea. I just added it.

InnaAtanasova:
The width of the input changes when the clear button is shown. 
Expectation: the width of the control should stay consistent:
Screenshot 2024-05-22 at 9 12 40 AM
Screenshot 2024-05-22 at 9 12 31 AM

mikerodonnell89 (commented):
Does platform need this too?

InnaAtanasova (approved):

PR #12015, opened 2024-06-19
fix(platform): quick fix performance for platform table with many columns

CSEHoangV:
## Related Issue(s)

None (Requested by Sourcing)

## Description

Currently, even with `virtualScroll`, which handles large amount of items, but platform table has issue with large amount of columns which causes some slowness and even page freeze. The idea in this quick fix is to render table cell data only in viewport and reduce DOMNodes, js event listener...

- Implement IntersectionObserver to control the rendering in viewport.
- Introduce property input `onlyRenderVisibleCells` which can be used with any platform table feature (`virtualScroll`, `pageScrolling`, etc...). This toggle is to enable this change.
- Introduce property input `useCellPlaceholder` to use placeholder [fd-busy-indicator] when cell is not within viewport.
- Change affects scrolling behavior and experience, for example, it affects `renderAhead` of `virtualScroll` as it only render data in viewport. It should be better enabled if there are many columns in table for now. 

## Screenshots

https://github.com/SAP/fundamental-ngx/assets/162270184/fb667b81-89c6-40c9-b375-3bf3c59f9737

### DIFF
--- a/libs/docs/platform/table/examples/advanced-scrolling/advanced-scrolling-example.component.html
+++ b/libs/docs/platform/table/examples/advanced-scrolling/advanced-scrolling-example.component.html
@@ -8,7 +8,8 @@
     emptyTableMessage="No data found"
     bodyHeight="300px"
     [virtualScroll]="true"
-    [enhancedRendering]="enhancedRendering"
+    [onlyRenderVisibleCells]="onlyRenderVisibleCells"
+    [useCellPlaceholder]="cellPlaceholder"
     [pageScrolling]="true"
     [pageSize]="20"
     [renderAhead]="20"
@@ -17,8 +18,9 @@
     [expandOnInit]="true"
 >
     <fdp-table-toolbar title="Order Line Items" [hideItemCount]="false" [hideSearchInput]="true">
-        <fdp-table-toolbar-actions *ngIf="enhancedRendering">
+        <fdp-table-toolbar-actions *ngIf="onlyRenderVisibleCells">
             <fdp-button label="Add column" (click)="addColumn()"></fdp-button>
+            <fdp-button label="Toggle cell placeholder" (click)="toggleCellPlaceholder()"></fdp-button>
         </fdp-table-toolbar-actions>
     </fdp-table-toolbar>
 
@@ -57,7 +59,7 @@
     >
     </fdp-column>
 
-    <ng-container *ngIf="enhancedRendering">
+    <ng-container *ngIf="onlyRenderVisibleCells">
         <fdp-column
             width="100px"
             *ngFor="let column of columns; let i = index"
--- a/libs/docs/platform/table/examples/advanced-scrolling/advanced-scrolling-example.component.html
+++ b/libs/docs/platform/table/examples/advanced-scrolling/advanced-scrolling-example.component.html
@@ -8,14 +8,19 @@
     emptyTableMessage="No data found"
     bodyHeight="300px"
     [virtualScroll]="true"
+    [enhancedRendering]="enhancedRendering"
     [pageScrolling]="true"
-    [pageSize]="50"
-    [renderAhead]="50"
+    [pageSize]="20"
+    [renderAhead]="20"
     [pageScrollingThreshold]="80"
     hasChildrenKey="hasChildren"
     [expandOnInit]="true"
 >
-    <fdp-table-toolbar title="Order Line Items" [hideItemCount]="false" [hideSearchInput]="true"> </fdp-table-toolbar>
+    <fdp-table-toolbar title="Order Line Items" [hideItemCount]="false" [hideSearchInput]="true">
+        <fdp-table-toolbar-actions *ngIf="enhancedRendering">
+            <fdp-button label="Add column" (click)="addColumn()"></fdp-button>
+        </fdp-table-toolbar-actions>
+    </fdp-table-toolbar>
 
     <fdp-column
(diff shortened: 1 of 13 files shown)


libs/docs/platform/table/examples/advanced-scrolling/advanced-scrolling-example.component.html
```diff
@@ -8,14 +8,19 @@
...
-    [renderAhead]="50"
+    [pageSize]="20"
+    [renderAhead]="20"
     [pageScrollingThreshold]="80"
     hasChildrenKey="hasChildren"
     [expandOnInit]="true"
 >
-    <fdp-table-toolbar title="Order Line Items" [hideItemCount]="false" [hideSearchInput]="true"> </fdp-table-toolbar>
+    <fdp-table-toolbar title="Order Line Items" [hideItemCount]="false" [hideSearchInput]="true">
+        <fdp-table-toolbar-actions *ngIf="enhancedRendering">
```
oscargong, line 20: `enhancedRendering` is default `false` in the `advanced-scrolling-example.component.ts` . Do we need add a button to toggle `enhancedRendering` in this example?

CSEHoangV: I think there is no need to toggle as it may cause issue (page freeze) if there are many columns in the example. The example should show-case how it works with new change as there is another table on the same page for comparison.

oscargong (approved):

libs/platform/src/lib/table/components/table-row/table-row.component.html
```diff
@@ -1,228 +1,258 @@
...
-            overflow: 'hidden',
-            clip: 'rect(1px, 1px, 1px, 1px)',
-            color: 'transparent',
-            display: 'block'
-        }"
-    >
-        {{ (row.checked ? 'platformTable.deselectSingleRow' : 'platformTable.selectSingleRow') | fdTranslate }}
-    </span>
+<ng-template #fallbackContent>
+    <fd-busy-indicator [loading]="true" size="s" [block]="true"> </fd-busy-indicator>
```
mikerodonnell89, line 2: I think it would be best to render nothing here, seeing the busy indicator to the left of the table for a fraction of a second is a bit weird. I think rendering nothing will make the enhanced rendering less perceptible

libs/platform/src/lib/table/components/table-row/table-row.component.ts
```diff
@@ -117,6 +117,10 @@ export class TableRowComponent<T> extends TableRowDirective implements OnInit, A
     @Input()
     isTreeRow: boolean;
 
+    /** Whether the row is only rendering within viewport */
+    @Input()
+    enhancedRendering: boolean;
```
mikerodonnell89, line 122: Need a more specific name here, maybe "onlyRenderVisibleCells" or something like that?

libs/platform/src/lib/table-helpers/directives/table-advanced-viewport.directive.ts
```diff
@@ -0,0 +1,128 @@
...
+import { Observable, take, takeUntil, Subscriber, Subject } from 'rxjs';
+import { DestroyedService } from '@fundamental-ngx/cdk/utils';
+import { ViewportRootService } from '../services/table-advanced-viewport-root.service';
+
+@Directive({
+    selector: '[fdpTableViewport]',
+    standalone: true,
+    providers: [DestroyedService]
+})
+export class ViewportDirective implements OnInit, OnDestroy {
```
mikerodonnell89, line 22: Need a more specific name here and for the filename as well. It should give us an idea as to what the directive does

mikerodonnell89 (changes requested):

libs/platform/src/lib/table/components/table-row/table-row.component.html
CSEHoangV: Agree, but our UX wants to see a placeholder. You can see it fast there in ngx playground, but application may be different (ex: api, other computations..., which ofc should be addressed by our side) Initially, I want to use `fd-skeleton`, but some implementation in that component make the CPU spikes up...
I guess I can let it decided by application to pass the placeholder.

libs/platform/src/lib/table-helpers/directives/table-advanced-viewport.directive.ts
```diff
@@ -0,0 +1,128 @@
...
+        } else {
+            this.checkViewport();
+        }
+    }
+
+    /** @hidden */
+    checkViewport(): void {
+        const options: IntersectionObserverInit = {
+            root: this._viewportRootService.getRootNode(),
+            rootMargin: '50px',
```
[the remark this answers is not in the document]
CSEHoangV: It is like adding css around the element. The margin is also considered to decide when the intersection happens. This helps to decide when to render data.

libs/platform/src/lib/table-helpers/directives/table-advanced-viewport.directive.ts
CSEHoangV: Updated

libs/platform/src/lib/table/components/table-row/table-row.component.ts
CSEHoangV: Updated

mikerodonnell89 (approved):

Issue #11979, opened 2024-06-04
Bug: Incorrect font weight for column header in sorted state (core and platform table)


do-oel:
#### Is this a bug, enhancement, or feature request?
Bug
#### Describe your proposal.
When a table is sorted by a specific column, the respective column header changes the text font weight from 700 (semibold) to 400 (regular). It should stay consistent with the other column headers at 700.
![image](https://github.com/SAP/fundamental-ngx/assets/143730992/be4f9564-8b7d-41aa-95bd-aa2ded1106f8)

#### Can you handle that on the application side
No
#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
Current
#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
fd-table, fdp-table, column sorting

PR #12018, opened 2024-06-19
fix(platform): add fix for incorrect font weight for column header in sorted state

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/11979

## Description

When a table is sorted by a specific column, the respective column header changes the text font weight from 700 (semibold) to 400 (regular). It should stay consistent with the other column headers at 700. Added a CSS fix.

### Before:
Screenshot 2024-06-19 at 9 45 53 AM

### After:
Screenshot 2024-06-19 at 9 46 02 AM


mikerodonnell89 (approved):
Is this not a fundamental-styles thing?

Issue #11899, opened 2024-05-08
Performance issue with a large number of columns


Asbarn:
BUG
There's a performance issue when using a lot of custom columns inside fdp-table component

Proposal:
Remove unused listeners and observers to free up the javascript thread. Right now there are too many subscriptions which can't be turned off if we don't need them.
Also even with virtualScroll enabled, rows are being re-rendered even if they are not visible:
note in the example that translate pipe is being triggered = page number * page size, when with virtual scroll it should be triggered just page size times. So when we scroll down 5-10 pages the table gets reloaded fully and page just freezes or takes >1 minute to refresh
Also please take a second look at how you use "will-change" property: IMO you use it too much and it not only does not optimise table but slows it down

#### Can you handle that on the application side
We tried to switch to DataProvider and use virtual scroll but it only improved our performance slightly
We tried to switch to signals and also use deferrable views but it didn't help

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
Angular: >17.1
Fundamental: >0.49.2

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
This bug is reproducible if you have a lot of columns and if you have custom columns, pipes and additional customisations in SCSS and HTML

Stackblitz example containing some of the source code:
https://stackblitz.com/edit/wza6e3

#### Did you check the documentation and the API?
We checked the source code. Proposal provided above

#### Did you search for similar issues?
Yes, similar issues have been reported before but no solution has been provided

#### Is there anything else we should know?
I implemented the same table using core fd-table and it works much better. It still freezes on >5-10 page as it re-renders the whole table because it does not virtual scroll but it still works better than platform's version of table

IMPORTANT NOTE:
Removing the check for innerText drastically improved the performance for us:
![Bild](https://github.com/SAP/fundamental-ngx/assets/47718251/12955386-0303-4342-bbd8-56ad92d2616f)
Can we do something about this line?

**_IMPORTANT: Please refrain from providing links or screenshots of SAP's internal information, as this project is open-source, and its contents are accessible to anyone._**

droshev:
We are working a new virtual scroll with #11443. Once it is done, we shall analyze this scenario. @Asbarn thanks for reporting the issue.

mikerodonnell89:
Need to take fix from #12015 and apply it to main

mikerodonnell89:
[...]

PR #12021, opened 2024-06-20
fix(platform): fix performance by using another markup properties

CSEHoangV:
## Related Issue(s)

partial fix https://github.com/SAP/fundamental-ngx/issues/11899

## Description

innerText triggers reflow issue which cause performance as we have more contents in platform table.

As per doc: https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext

```
Moreover, since innerText takes CSS styles into account, reading the value of innerText triggers a [reflow](https://developer.mozilla.org/en-US/docs/Glossary/Reflow) to ensure up-to-date computed styles. (Reflows can be computationally expensive, and thus should be avoided when possible.)
```

There is another difference based on doc. Given the context from the table row component, this should not have impact, but this will need confirmation.

### DIFF
--- a/libs/platform/src/lib/table/components/table-row/table-row.component.html
+++ b/libs/platform/src/lib/table/components/table-row/table-row.component.html
@@ -156,7 +156,7 @@
 
         <ng-container *ngIf="row.state === 'readonly'; else editModeCell">
             <span
-                *ngIf="tableTextContainer?.innerText?.trim() === ''"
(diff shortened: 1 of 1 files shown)


mikerodonnell89 (approved):

InnaAtanasova (approved):

CSEHoangV:
I will add this fix to another pr: https://github.com/SAP/fundamental-ngx/pull/12015.

PR #12049, opened 2024-06-28
feat(platform): virtual scroll whole rows enhancement for ng15

### LINKED ISSUE #11491: [P2] fdp-table scrolling updates
The table scrolling behavior must be reworked to more closely match that of the ui5 grid table: https://sapui5.hana.ondemand.com/#/entity/sap.ui.table.Table/sample/sap.ui.table.sample.Basic

Notable differences are:

- All rows must have a matching fixed height. The table cannot display partial rows.
- In UI5, when a scroll event is performed, rather than scrolling the table container by 1px as is typical when performing a scroll, the scroll event is caught and prevented, the `<tr>` elements are fixed in place, and the data from each row is moved up or down to the previous or next row(s) depending on the scroll direction. This provides a snappier feel from a UX perspective and also eliminates some edge case problems such as #11443 because there is no way to have the table only showing a partial row (also due to the fixed height of rows)
- If a cell is focused, and the user scrolls, the focused cell position stays the same, effectively making a different cell focused:

https://github.com/SAP/fundamental-ngx/assets/2471874/6495c774-6840-4d3c-b8c9-2cba0f6b799b

This is tricky to explain, because from the developer's perspective, it's the same `<td>` that is focused. But from the user's perspective, the focus has changed cells.

- If a table cell has some child element that is actionable by the user such as a text input, a scroll event on the table should be considered as a "submit" action by the user

This will be released as a significant breaking change to the existing virtual scrolling functionality of the platform table. If these changes are downported to a previous version of fundamental-ngx, it will need to be released as a separate feature, so as to not break the existing functionality for users on fundamental-ngx using ng15 etc

mikerodonnell89:
fixes #11491 

Adds a new input to the platform table `[scrollWholeRows]`, intended to be used in conjunction with `[virtualScroll]="true"`. Modifies the behavior of (vertical) virtual scrolling for the table. When the user performs a scroll (via mousewheel, scrollbar, keyboard arrows, etc), rather than scrolling pixel by pixel, an entire row will be scrolled. This changes the UX to more closely match that of UI5. It also comes with some inherent performance improvements such as not rendering hidden `<tr>`s.

This is added as a new feature. If the developer does not add `[scrollWholeRows]="true"`, they'll see the previous virtual scrolling behavior

### DIFF
--- a/libs/platform/src/lib/table-helpers/directives/table-virtual-scroll.directive.ts
+++ b/libs/platform/src/lib/table-helpers/directives/table-virtual-scroll.directive.ts
@@ -1,12 +1,13 @@
-import { Directive, inject, Input, OnChanges, SimpleChanges } from '@angular/core';
-import { DestroyedService } from '@fundamental-ngx/cdk/utils';
+import { Directive, HostListener, inject, Input, OnChanges, OnDestroy, SimpleChanges } from '@angular/core';
+import { DestroyedService, FocusableItemPosition, KeyUtil } from '@fundamental-ngx/cdk/utils';
 import { ContentDensityMode } from '@fundamental-ngx/core/content-density';
-import { BehaviorSubject, filter } from 'rxjs';
+import { BehaviorSubject, filter, Subscription } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 import { FDP_TABLE_VIRTUAL_SCROLL_DIRECTIVE, ROW_HEIGHT } from '../constants';
 import { TableVirtualScroll } from '../models';
 import { TableScrollDispatcherService } from '../services/table-scroll-dispatcher.service';
 import { Table } from '../table';
+import { DOWN_ARROW, UP_ARROW } from '@angular/cdk/keycodes';
 
 @Directive({
     // eslint-disable-next-line @angular-eslint/directive-selector
@@ -20,12 +21,20 @@ import { Table } from '../table';
         }
     ]
 })
-export class TableVirtualScrollDirective extends TableVirtualScroll implements OnChanges {
+export class TableVirtualScrollDirective extends TableVirtualScroll implements OnChanges, OnDestroy {
     /** Whether to show only visible rows in matter of performance
      * false by default, when true setting bodyHeight and rowHeight is required.
      */
     @Input()
     virtualScroll = false;
+
+    /**
+     * Whether to scroll whole rows rather than pixel by pixel when the user performs a scroll.
+     * This gives greater performance improvements but also lends to some potentially undesirable UX changes.
+     * A "rowHeight" must be provided when using this feature.
+     */
+    @Input()
+    scrollWholeRows = false;
 
     /** Cache size for the virtualScroll, default is 40 in each direction */
     @Input()
@@ -61,6 +70,48 @@ export class TableVirtualScrollDirective extends TableVirtualScroll implements O
     private readonly _destroy$ = inject(DestroyedService);
 
     /** @hidden */
+    private _wheelTimeout: any;
+
+    /** @hidden */
+    private _scrollMockTimeout: any;
+
+    /** @hidden */
+    private _lastMockScrollPosition = 0;
+
+    /** @hidden */
+    private _focusableGridSubscription: Subscription;
+
+    /** @hidden */
+    private _focusedCell: FocusableItemPosition;
+
+    /** @hidden */
+    private _previousStartNodeIndex: number | null = null;
+
+    /** @hidden */
+    @HostListener('keydown', ['$event'])
+    _keydownHandler(event: KeyboardEvent): void {
+        if (this.scrollWholeRows) {
+            if (this._focusedCell?.rowIndex === 0 && KeyUtil.isKeyCode(event, UP_ARROW)) {
+                if (this._table.tableContainer.nativeElement.querySelectorAll('tr')[1].ariaRowIndex !== '0') {
+                    const cellToFocus = {
+                        colIndex: this._focusedCell.colIndex,
+                        rowIndex: 1,
+                        totalCols: this._focusedCell.totalCols,
+                        totalRows: this._focusedCell.totalRows
+                    };
+                    this._table._focusableGrid.focusCell(cellToFocus);
+                }
+                this._scrollRow('up', -1);
+            } else if (
+                this._focusedCell?.rowIndex === this._getNumberOfRowsToDisplay() &&
+                KeyUtil.isKeyCode(event, DOWN_ARROW)
+            ) {
+                this._scrollRow('down', 1);
+            }
+        }
+    }
+
+    /** @hidden */
     ngOnChanges(changes: SimpleChanges): void {
         if (!this._table.tableScrollable) {
             return;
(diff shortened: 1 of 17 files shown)


mikerodonnell89:
Two things to fix:

- When resizing a column to make it smaller, it won't immediately shrink the column when releasing the mouse button. User must click outside the table and the column then shrinks
- Resizing the window to shrink the table does not show horizontal scrollbar

libs/platform/src/lib/table-helpers/directives/table-virtual-scroll.directive.ts
```diff
@@ -79,62 +135,174 @@ export class TableVirtualScrollDirective extends TableVirtualScroll implements O
...
 
-        let startNodeIndex = Math.floor(scrollTop / rowHeight) - this.renderAhead;
-        startNodeIndex = Math.max(0, startNodeIndex);
+            this._setRows(startNodeIndex);
 
-        let visibleNodeCount =
-            Math.ceil(this._table.tableContainer.nativeElement.clientHeight / rowHeight) + 2 * this.renderAhead;
-        visibleNodeCount = Math.min(totalNodeCount - startNodeIndex, visibleNodeCount);
+            this._table.tableScrollMockContainer.nativeElement.style.width = '1rem';
+            this._table.tableScrollMockContainer.nativeElement.style.maxHeight = this.bodyHeight;
```
CSEHoangV, line 164: This is duplicated DOM query. This should be locally cached.
Ex: `const scrollMockContainer = this._table.tableScrollMockContainer.nativeElement`.

libs/platform/src/lib/table-helpers/directives/table-virtual-scroll.directive.ts
```diff
@@ -79,62 +135,174 @@ export class TableVirtualScrollDirective extends TableVirtualScroll implements O
...
+                        scrollTopPosition: scrollable.getScrollTop()
+                    });
+                });
+        }
+    }
+
+    /** @hidden */
+    private _getNumberOfRowsToDisplay(): number {
+        let tableHeight = this._table.tableContainer.nativeElement.clientHeight;
+        tableHeight = tableHeight - this._table.tableContainer.nativeElement.querySelector('thead').clientHeight;
```
CSEHoangV, line 240: duplicated DOM query

libs/platform/src/lib/table-helpers/directives/table-virtual-scroll.directive.ts
```diff
@@ -79,62 +135,174 @@ export class TableVirtualScrollDirective extends TableVirtualScroll implements O
...
+            event.preventDefault();
+            event.stopImmediatePropagation();
+            clearTimeout(this._wheelTimeout);
+            const deltaY = event.deltaY;
+            if (deltaY) {
+                this._wheelTimeout = setTimeout(() => {
+                    this._table.tableScrollMockContainer.nativeElement.scrollBy({ top: deltaY });
+                    this._scrollRow(deltaY > 0 ? 'down' : 'up', deltaY > 0 ? 1 : -1);
+                    this._lastMockScrollPosition = this._table.tableScrollMockContainer.nativeElement.scrollTop;
+                }, 5);
```
CSEHoangV, line 256: duplicated DOM query. Also is it better to use https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame instead of setTimeout?

libs/platform/src/lib/table-helpers/services/table-column-resize.service.ts
```diff
@@ -176,8 +176,15 @@ export class TableColumnResizeService implements OnDestroy {
 
     /** Retrieves custom column value or returns `unset` */
     getColumnWidthStyle(columnName: string): string {
-        const calculatedWidth = this._fixedColumnsWidthMap.get(columnName);
-        return calculatedWidth || 'unset';
+        if (this._tableRef._virtualScrollDirective?.scrollWholeRows) {
+            return (
+                this._fixedColumnsWidthMap.get(columnName) ||
+                this._tableRef._tableWidthPx / this._tableRef.getVisibleTableColumns().length + 'px'
+            );
```
CSEHoangV, line 183: I think this `this._tableRef._tableWidthPx` needs optimization if we need this. I observe removing this new condition contributes performance improvement.

libs/platform/src/lib/table/components/table-row/table-row.component.html
```diff
@@ -163,7 +163,13 @@
         <ng-template #primaryTextContainer>
             <div
                 [class.fd-table__text]="column.applyText"
-                [class.fd-table__text--no-wrap]="column.noWrap"
+                [class.fd-table__text--no-wrap]="column.applyText"
+                [class.fd-table__cell--truncate-txt]="column.applyText"
+                [attr.title]="
+                    tableTextContainer.offsetWidth < tableTextContainer.scrollWidth
```
CSEHoangV, line 169: Can you also optimize this? `offsetWidth`, `innerText` and `scrollWidth` can potentially cause forced reflow especially this is changed frequently.

libs/platform/src/lib/table-helpers/services/table-column-resize.service.ts
mikerodonnell89: This should only need to be called on init

khotcholava (approved):
Looks good to me

libs/platform/src/lib/table-helpers/services/table-column-resize.service.ts
CSEHoangV: This is at `td` level. We basically do DOM query on potential reflow properties for every cell. And it is used as function call on table-row html (which is recommended to be avoided). May be converted to pipe also or avoid using DOM query here.

libs/platform/src/lib/table-helpers/directives/table-virtual-scroll.directive.ts
```diff
@@ -68,6 +128,14 @@ export class TableVirtualScrollDirective extends TableVirtualScroll implements O
         if (this.virtualScroll && (changes['rowHeight'] || changes['virtualScroll'] || changes['renderAhead'])) {
             this.calculateVirtualScrollRows();
         }
+        if (changes?.scrollWholeRows && changes.scrollWholeRows.currentValue && !this.virtualScroll) {
```
dpavlenishvili, line 131: i think if we are going to exclude ngOnChanges, it will be better if you use input set or input signal instead of handling this logic inside ngOnchanges.

libs/platform/src/lib/table-helpers/directives/table-virtual-scroll.directive.ts
```diff
@@ -79,62 +147,195 @@ export class TableVirtualScrollDirective extends TableVirtualScroll implements O
...
+                tableScrollMockContainer.scrollHeight
+            ) {
+                const startingNodeIndex = this._table._tableRowsVisible.length - this._getNumberOfRowsToDisplay();
+                this._setRows(startingNodeIndex);
+            }
+        });
+    };
+
+    /** @hidden */
+    private _scrollRow(count: number, fromMockScrollbarListener = false): void {
```
dpavlenishvili, line 302: it's little bit non understandable how this method is working, maybe you should add some description before *hidden

libs/platform/src/lib/table-helpers/directives/table-virtual-scroll.directive.ts
mikerodonnell89: No signals in ng15 but yes an input setter will work

dpavlenishvili (approved):

libs/platform/src/lib/table/components/table-row/table-row.component.html
mikerodonnell89: we'll address this in a later PR

Issue #11848, opened 2024-04-29
Throw error for invalid values in combobox


pavandk:
Enhancement:

This is created on follow up of [ticket](https://github.com/SAP/fundamental-ngx/issues/11618)
We understand combo box should allow invalid values as well. But we have scenario where we need to show error if user enters values other than dropdown values. We are unable to do because in case of invalid values the control still holds previous selected dropdown value. Can you please enhance control to handle such scenarios?

ZsDeak-SAP:
@pavandk for the invalid values and state you should use the `state` input. there was indeed an issue with the selection after the input has been modified.
@droshev i fixed the issue, see the PR please 😸

mikerodonnell89:
fixed with https://github.com/SAP/fundamental-ngx/pull/12062

PR #12062, opened 2024-07-02
fix(core): control value not updating properly on user input

ZsDeak-SAP:
## Related Issue(s)

https://github.com/SAP/fundamental-ngx/issues/11848

## Description

Keeping input and form control value in sync (unless using objects only).
One exceptional scenario is when the user do not use `communicateByObject`, but does use `displayFn`. `displayFn` expects one of the `item` objects, that is why in `inputText` setter we need to set `_value`, not call `setValue`.

[this example](https://fundamental-ngx.netlify.app/#/core/combobox#display-object) and many others can be used to validate. select one of the items, modify the text input and open the select again by the chevron button. there should be no selected highlighting when `communicateByObject` is false, since the new valid value is what the user typed.

### DIFF
--- a/libs/core/combobox/combobox.component.spec.ts
+++ b/libs/core/combobox/combobox.component.spec.ts
@@ -72,7 +72,7 @@ describe('ComboboxComponent', () => {
     it('should handle input entry on dropdown mode', () => {
         jest.spyOn(component, 'onChange');
         component.communicateByObject = true;
-        component.displayFn = (item: any): string => item.displayedValue;
+        component.displayFn = (item: any): string => item?.displayedValue;
         component.onMenuClickHandler(component.dropdownValues[1]);
         expect(component.onChange).toHaveBeenCalledWith(component.dropdownValues[1]);
     });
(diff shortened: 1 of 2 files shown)


InnaAtanasova (approved):

InnaAtanasova:
Thank you for taking care of this issue @ZsDeak-SAP 🙇‍♀️

mikerodonnell89 (changes requested):
We will need this for the platform combobox as well

ZsDeak-SAP:
> We will need this for the platform combobox as well

i cannot reproduce, nor find this issue in the code of that combobox. the implementation is vastly different

mikerodonnell89:
Ok - one of my understandings of this issue is that previously, when the user would select an item in the list and then erase a character from the input, the popover would still display the previously selected item as selected with a blue background like so:

Screenshot 2024-07-08 at 10 58 58 AM

And this PR changes that behavior so the previously selected item is no longer displayed as selected, like so: 

Screenshot 2024-07-08 at 11 55 36 AM

So, we'd expect this behavior in platform as well, correct? Because on your PR, the platform combobox is still working like my first screenshot above

ZsDeak-SAP:
@mikerodonnell89 that's a good point, but Core and Platform comboboxes are very different, even in how they function. Core updates all comboboxes by default on typing, Platform only sends the changes when you click out or hit enter. Try after selecting an item deleting a character, hit enter/or click out, and then open the list again. The previously selected item is deselected as expected, and the texts showing the state under the combobox examples show the correct value.

mikerodonnell89 (approved):

Issue #12097, opened 2024-07-11
Combobox - Reactive form not working properly with [communicateObject]


saurabhagarwal7755:
#### Is this a bug, enhancement, or feature request?
Bug

#### Describe your proposal.
For COMBOBOX
We are using fd-combobox with [communicateByObject] as we are using custom objects from the dropdownValues (trigger new div when the value changes) and the combobox is "Required" field for us.
In fundamentals version 0.43, upon removal of the data from combobox it was clearing the data like this: https://docs-0-43--fundamental-ngx.netlify.app/#/core/combobox
Please check the image below:
![Screenshot 2024-07-11 at 12 27 08 PM](https://github.com/SAP/fundamental-ngx/assets/52731148/282e1a52-8f8b-4990-b8cc-ab169c680453)
In fundamentals version 0.47 to 0.50 it got updated and now upon clearing the field in combobox, its taking the previously selected value, due to which we are not able to do invalid check and it is allowing to save the information even if the data is empty, as it is taking the last selected value and saving it: https://docs-0-47--fundamental-ngx.netlify.app/#/core/combobox
Please check the image below:
![image](https://github.com/SAP/fundamental-ngx/assets/52731148/bc164cd9-ed58-4597-a46b-533e0d7fd74f)

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
1. Go to version 0.47 of fundamentals.
2. Navigate to combobox, in which  [communicateByObject]="true" is used.
3. Select an option and then clear the value. 
4. Upon clearing it should clear the Json Value also, but here it is having the value selected in step 3
5. In version 0.43, this behaviour is working as expected.

InnaAtanasova:
Hi @saurabhagarwal7755 thanks for opening the issue. Could you please verify if this solution is what you expect as behaviour. Thanks

saurabhagarwal7755:
Hello @InnaAtanasova ,
Checked the preview in provided link. This solution would work for me, thank you for the fix.
Thanks

PR #12100, opened 2024-07-11
fix(core): clear the json value if input is empty in Combobox

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/12097

## Description
Step 1: select a value, the json value is updated
Screenshot 2024-07-11 at 12 45 45 PM

Step 2: clear the input, the json value is set to null
Screenshot 2024-07-11 at 12 45 53 PM

Before the fix the json value was keeping the previous selection when the input was cleared.

### DIFF
--- a/libs/core/combobox/combobox.component.ts
+++ b/libs/core/combobox/combobox.component.ts
@@ -821,6 +821,8 @@ export class ComboboxComponent<T = any>
             // Do not set new value if theres multiple items that have same label.
             if (values.length === 1 && this.displayFn(values[0]) !== this.displayFn(this.getValue())) {
                 this.setValue(values[0]);
+            } else if (values.length === 0) {
+                this.setValue(null);
             }
             this.onChange(this.getValue());
         } else {


mikerodonnell89 (approved):

khotcholava (approved):

dreamweiver:
@InnaAtanasova shouldn't the actual search term( "inputText") be returned instead of null always?

Because Imagine you are trying to search for an item inside a combo box, if there are no matching items for the given search term, then actual search text should be returned (in case of **communicateByObject** "true"), similar to when **communicateByObject** is set to false.

This helps in the following ways,
- The consuming application can use this invalid search text to show it in the error text to the user like
 "**invalid selection 'seachText'**" when using a custom validator, which is better than showing null.
- returning actual search text can be used to process custom actions. for eg if a user types the word like "create", "new", etc.,  then custom action can be carried on the consuming application side(this is just custom action)

I can also create a PR with this change, if you don't have any objection :)

Issue #12093, opened 2024-07-09
[P2][Sourcing] Unable to select multiple users if using search in approval flow


CSEHoangV:
#### Is this a bug, enhancement, or feature request?

bug

#### Describe your proposal.

- When we try to add multiple users in approval flow and if we use search and select another user, the previous selected users will be clear:

https://github.com/SAP/fundamental-ngx/assets/162270184/cfca8254-e4c8-439c-9fc3-bff497803f7b

#### Can you handle that on the application side

No

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)

Angular 15 (I think even in main branch)
0.43.35

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;

bug

#### Please provide relevant source code (if applicable).

https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/approval-flow

#### Please provide stackblitz example(s).

https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/approval-flow

#### Did you check the documentation and the API?

Yes

#### Did you search for similar issues?

Yes

#### Is there anything else we should know?

We have multiple reports from customers about this issue. This is not good for them especially when they have a lot of approvers so they need to use search to select user.

**_IMPORTANT: Please refrain from providing links or screenshots of SAP's internal information, as this project is open-source, and its contents are accessible to anyone._**

mikerodonnell89:
Fix has been merged for ng15 but we'll need this on the main branch as well

droshev:
closing it as we don't have spec

PR #12107, opened 2024-07-13
fix(platform): cache user selection after searching for new user list for approval flow

CSEHoangV:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/12093

## Description

In approval flow, when we first selected users, the previous selected user(s) will be deselected after we select new users from the search result. The reason is that the selected list only includes users currently displayed on the dialog and all selection will be checked against current displayed list items before emitting to cache current selection. 

After fix:

https://github.com/user-attachments/assets/b5ec29d2-c820-4d58-b702-fd329e98ee09


libs/platform/src/lib/approval-flow/approval-flow-user-list/approval-flow-user-list.component.ts
```diff
@@ -122,15 +130,28 @@ export class ApprovalFlowUserListComponent implements AfterViewInit, OnChanges,
...
+            this._cdr.detectChanges();
+        }
     }
 
     /** @hidden */
     private _getUsersFromSelectedItems(items: BaseListItem[]): ApprovalUser[] {
         return items
-            .map((item) => this.users.find((user) => `${this._idPrefix + user.id}` === item.itemEl.nativeElement.id))
+            .map((item) => this.users.find((user) => `${this._idPrefix + user.id}` === item.id))
             .filter((u): u is ApprovalUser => !!u);
```
CSEHoangV, line 155: Change it due to DOM query on potential large user display on dialog which can cause bottleneck on rendering.

libs/platform/src/lib/list/base-list-item.ts
```diff
@@ -534,7 +534,9 @@ export class BaseListItem extends BaseComponent implements OnInit, AfterViewInit
      * Show navigation for single list
      */
     ngOnInit(): void {
-        this.id = `fdp-list-item-${nextListItemId++}`;
+        if (!this.id || this.id === this.defaultId) {
+            this.id = `fdp-list-item-${nextListItemId++}`;
```
CSEHoangV, line 538: By default, id is `defaultId` which is overridden by this implementation. We provide custom input parameter for id but it always gets overridden.

libs/platform/src/lib/approval-flow/approval-flow-user-list/approval-flow-user-list.component.ts
```diff
@@ -84,27 +88,31 @@ export class ApprovalFlowUserListComponent implements AfterViewInit, OnChanges,
...
     constructor(private _cdr: ChangeDetectorRef) {}
 
     /** @hidden */
     ngAfterViewInit(): void {
         if (this.selectedUsers.length) {
-            const selectedApproversNames = this.selectedUsers.map((approver) => approver.name);
-
-            this._selectedItems = this.listItems.filter(
-                (item) => !!item.avatar?.ariaLabel && selectedApproversNames.includes(item.avatar.ariaLabel)
-            );
```
CSEHoangV, line 100: This is not unique which can result in multiple selection on users having same name.

mikerodonnell89 (commented):
Looks good, one minor thing I noticed after searching for someone to add, is that the "cancel" button basically works the same as the "add" button. I'm guessing it should revert whoever has been newly checked. The dialog then says it's going to be adding 3 people, but then (correctly) only adds 2. See below: 

https://github.com/user-attachments/assets/aa0fe106-5a5c-44d3-9944-39bf1e6e86b8

mikerodonnell89 (approved):

Issue #12035, opened 2024-06-24
[P2][Sourcing][Performance] Event triggers when platform table is inside dynamic page component


CSEHoangV:
#### Is this a bug, enhancement, or feature request?

bug

#### Describe your proposal.

Dynamic header is using scroll spy directive which determine content in current view when it navigate between tabs or we scroll the dynamic content. However, some event triggers which is not related to the dynamic page component like table scroll, table focus, or mousemove... triggers onScroll of this scroll-spy which has DOM manipulation.

Table scroll:

https://github.com/SAP/fundamental-ngx/assets/162270184/1b54d644-5d08-4533-bb9f-682630bb2fb2

Table focus or table click:

https://github.com/SAP/fundamental-ngx/assets/162270184/ba13e6c1-f635-4557-9001-e6ce868e6202

I believe this is a wrong behavior because table scroll or click should not determine the content position in current view of dynamic content. And with the onScroll implementation of scroll-spy directive, it can affect the performance of table especially when we scroll the table and render the next rows.

#### Can you handle that on the application side

Remove dynamic header but that is not the solution.

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)

Angular 15
0.43.35

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;

- Add core dynamic header with tabs.
- Add platform table.
- Put breakpoint on onScroll function of scroll-by directive or run performance profile to detect the issue.
- Perform table scroll, click, mousemove...

#### Please provide relevant source code (if applicable).

https://stackblitz.com/edit/angular-9qxxca-476tct?file=package.json

#### Please provide stackblitz example(s).

https://stackblitz.com/edit/angular-9qxxca-476tct?file=package.json

#### Did you check the documentation and the API?

Yes

#### Did you search for similar issues?

Yes

#### Is there anything else we should know?

The redundant call gets triggered by https://github.com/SAP/fundamental-ngx/blob/ng-15-downport/libs/core/src/lib/tabs/tab-list.component.ts#L310

**_IMPORTANT: Please refrain from providing links or screenshots of SAP's internal information, as this project is open-source, and its contents are accessible to anyone._**

CSEHoangV:
Another explanation is that from the record, tab2 is highlighted and we don't do anything to change position of dynamic body (actions on table), so it should not trigger onScroll from scrollspy directive.

PR #12125, opened 2024-07-16
fix(core): Remove redundant api calls on dynamic page

CSEHoangV:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/12035

## Description

The fix from https://github.com/SAP/fundamental-ngx/pull/10762 handles content body with static height and when there is a new item being added asynchronously. However, it created a side effect which triggers many calls on `onScroll` from scroll-spy directive even when there is no change to dynamic content body.

The scenario mentioned above has only handled if there is a new change to tab list. Therefore, this fix will add a check if there is a change to the size of tab list and will only execute `onScroll` from scroll-spy directive if that is a case.

After fix:

Remove redundant calls and regression check on previous fix:

 

https://github.com/user-attachments/assets/3507b243-6a3d-408a-801f-9330654b6a6c

### DIFF
--- a/libs/core/src/lib/tabs/tab-list.component.ts
+++ b/libs/core/src/lib/tabs/tab-list.component.ts
@@ -26,7 +26,7 @@ import { MenuComponent } from '@fundamental-ngx/core/menu';
 import { OverflowLayoutComponent } from '@fundamental-ngx/core/overflow-layout';
 import { ScrollSpyDirective } from '@fundamental-ngx/core/scroll-spy';
 import { ScrollbarDirective } from '@fundamental-ngx/core/scrollbar';
-import { Observable, Subject, Subscription, fromEvent, merge } from 'rxjs';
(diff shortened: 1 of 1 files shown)


yixuan-liu (approved):

dpavlenishvili (approved):

mikerodonnell89 (changes requested):
I think this is more of a band-aid solution, the fix should be coming from the scroll spy directive itself. A scroll on a scrollable child shouldn't trigger the scroll spy logic on the parent component that actually has the directive applied to it. And the scroll event listener definitely shouldn't be called when a child item is clicked, I have no clue why that is happening here but we need to investigate 

**EDIT** - The more I look in to this it does make sense to have a check for this in the tab-list. But in the future - may be good to address what I've mentioned above

mikerodonnell89 (approved):

PR #12132, opened 2024-07-17
fix(core): remove breadcrumb click proxy, allowing events on original breadcrumb item to be passed to overflow item

### LINKED ISSUE #11685: [P2][Sourcing] Breadcrumb does not pass click when collapse in the menu

mikerodonnell89:
fixes #11685

### DIFF
--- a/libs/core/src/lib/breadcrumb/breadcrumb.component.ts
+++ b/libs/core/src/lib/breadcrumb/breadcrumb.component.ts
@@ -135,10 +135,8 @@ export class BreadcrumbComponent implements OnInit, AfterViewInit {
      * if original element had router link we are proxying click to that element
      * */
     itemClicked(breadcrumbItem: BreadcrumbItemComponent, $event: any): void {
-        if (breadcrumbItem._needsClickProxy) {
-            $event.preventDefault();
-            breadcrumbItem.breadcrumbLink.elementRef.nativeElement.click();
-        }
+        $event.preventDefault();
+        breadcrumbItem.breadcrumbLink.elementRef.nativeElement.click();
     }
 
     /** @hidden */
--- a/libs/core/src/lib/breadcrumb/breadcrumb.component.ts
+++ b/libs/core/src/lib/breadcrumb/breadcrumb.component.ts
@@ -131,9 +131,8 @@ export class BreadcrumbComponent implements OnInit, AfterViewInit {
     }
 
     /**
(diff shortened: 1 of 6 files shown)


libs/docs/core/breadcrumb/examples/breadcrumb-examples.component.ts
```diff
@@ -36,4 +36,8 @@ export class BreadcrumbRouterLinkExampleComponent {
         `
     ]
 })
-export class BreadcrumbHrefExampleComponent {}
+export class BreadcrumbHrefExampleComponent {
+    onClick(value: string): void {
+        console.log(value);
```
InnaAtanasova, line 41: In my opinion an alert would be better

libs/core/src/lib/breadcrumb/breadcrumb.component.ts
```diff
@@ -135,10 +135,8 @@ export class BreadcrumbComponent implements OnInit, AfterViewInit {
      * if original element had router link we are proxying click to that element
      * */
     itemClicked(breadcrumbItem: BreadcrumbItemComponent, $event: any): void {
```
dpavlenishvili, line 137: if conditional check is no longer needed, please update function doc descriptions too.

dpavlenishvili (approved):

Issue #12036, opened 2024-06-25
Combobox with asyc dataSource leads to: Cannot read properties of undefined (reading 'length') 


irsapirsap:
#### Is this a bug, enhancement, or feature request?
Bug

#### Describe your proposal.
Stackblitz:
https://stackblitz.com/edit/vutgl2-n2yi5x?file=src%2Fapp%2Fcombobox-loading-example.component.ts,src%2Fapp%2Fcombobox-loading-example.component.html

fails with (open developer tools to see the console):
vendor.js:57447 ERROR TypeError: Cannot read properties of undefined (reading 'length')
    at ComboboxComponent_ng_template_1_Template (vendor.js:155437:92)
    at executeTemplate (vendor.js:58830:7)
    at refreshView (vendor.js:60254:7)
    at detectChangesInView (vendor.js:60416:5)
    at detectChangesInEmbeddedViews (vendor.js:60363:7)
    at refreshView (vendor.js:60277:5)
    at detectChangesInView (vendor.js:60416:5)
    at detectChangesInComponent (vendor.js:60394:3)
    at detectChangesInChildComponents (vendor.js:60428:5)
    at refreshView (vendor.js:60304:7)

#### Can you handle that on the application side
no

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
See the stackbliz example

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
See the stackbliz example

#### Please provide relevant source code (if applicable).
https://stackblitz.com/edit/vutgl2-n2yi5x?file=src%2Fapp%2Fcombobox-loading-example.component.ts,src%2Fapp%2Fcombobox-loading-example.component.html

#### Please provide stackblitz example(s).
https://stackblitz.com/edit/vutgl2-n2yi5x?file=src%2Fapp%2Fcombobox-loading-example.component.ts,src%2Fapp%2Fcombobox-loading-example.component.html

#### In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations.
not accessibility related

#### Did you check the documentation and the API?
yes

#### Did you search for similar issues?
yes

#### Is there anything else we should know?
Yes, the probem appears because you declare the `dataSource` as opitional
https://github.com/SAP/fundamental-ngx/blob/main/libs/platform/form/combobox/commons/base-combobox.ts#L167
[...]

PR #12162, opened 2024-07-22
fix(platform): Combobox with asyc dataSource

khotcholava:
closes [#12036](https://github.com/SAP/fundamental-ngx/issues/12036)

fix(combobox): support null as `dataSource`

Description:
Updated base combobox component, now it supports null as `dataSource`

### DIFF
--- a/libs/platform/form/combobox/commons/base-combobox.ts
+++ b/libs/platform/form/combobox/commons/base-combobox.ts
@@ -165,7 +165,7 @@ export abstract class BaseCombobox
 
     /** Datasource for suggestion list */
     @Input()
-    set dataSource(value: FdpComboBoxDataSource<any> | null) {
+    set dataSource(value: FdpComboBoxDataSource<any>) {
         if (value) {
             this._initializeDataSource(value);
         }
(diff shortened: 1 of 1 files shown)


libs/platform/form/combobox/commons/base-combobox.ts
```diff
@@ -165,9 +165,12 @@ export abstract class BaseCombobox
 
     /** Datasource for suggestion list */
     @Input()
-    set dataSource(value: FdpComboBoxDataSource<any>) {
-        if (value) {
+    set dataSource(value: FdpComboBoxDataSource<any> | null) {
+        if (value !== null) {
             this._initializeDataSource(value);
```
InnaAtanasova, line 170: this is not the same as before. With `if (value !== null)` you are not checking if value is false, 0, "", undefined. If that's the case, ok.

InnaAtanasova:
The title of the PR should be `fix(platform): text...`

khotcholava:
I found a simple solution.
The problem was with the suggestions not getting values when dataSource was null.
I initially changed it to assign an empty array. When dataSource is null, it doesn't affect the suggestions, and the value will always be an empty array. However, if we pass an array, it will use the provided array.

InnaAtanasova (approved):

Issue #12160, opened 2024-07-22
[P1][Sourcing]: Lagging when scrolling when using `pageScrolling` and `virtualScroll`


CSEHoangV:
#### Is this a bug, enhancement, or feature request?

bug

#### Describe your proposal.

This is without `scrollWholeRows`. When we scroll a table with `pageScrolling` and `virtualScroll`, we experience some slight lagging. For example, when it scrolls to some certain positions, it cause a small delay and then jump to next scroll.

- 0.43.38 (lag)

https://github.com/user-attachments/assets/2daaa706-cf0c-4464-8e2e-326206824cf2

- 0.43.36 (non-lag and smoother scroll)

https://github.com/user-attachments/assets/4a6f8202-321e-45e3-a427-57a7eff3003b

I think `pageScrolling` causes issue since when I disable it and use only `virtualScroll`, it will not cause any lagging issue.

#### Can you handle that on the application side

no

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)

angular 15
0.43.38

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;

- Create platform table with large amount of items and columns (just make easier to see the issue).
- Enable `pageScrolling` and `virtualScroll`.
- Scroll the table.

Result: Notice some light delay (the page of table is stay and then jump to another page...)
Expected: Should not cause any lagging.

#### Please provide stackblitz example(s).

0.43.38: https://stackblitz.com/edit/angular-ccrvey?file=src%2Fapp%2Fplatform-table-virtual-scroll-example.component.html
0.43.36: https://stackblitz.com/edit/angular-rgjghz?file=src%2Fapp%2Fplatform-table-virtual-scroll-example.component.html

#### In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations.

no

#### Did you check the documentation and the API?

yes

#### Did you search for similar issues?

yes

#### Is there anything else we should know?

Possible caused by this commit: https://github.com/SAP/fundamental-ngx/pull/12049/commits/b5dcc20728e9bc9858d20475cca9e23c568d9d36 with some change on scss and table.component.html since I revert this change to previous one and the scroll works fine. To be more specific, this block of change: https://github.com/SAP/fundamental-ngx/pull/12049/files#diff-0562f5d038c3f170150e90bf1ea0967017e5e026269740c41e68a89512ead50bL507

[...]

PR #12164, opened 2024-07-22
fix(platform): pageScroll/virtualScroll lag, and bug where new page of data would not be fetched after scroll

### LINKED ISSUE #12157: [P2][Sourcing] `scrollWholeRows`: Unable to fetch next page of rows if expand and collapse one row

bug

#### Describe your proposal.

Using `scrollWholeRows` and `pageScrolling`, if we expand and collapse one row item, we are unable to load next page of rows when scrollbar hits bottom.

https://github.com/user-attachments/assets/06eaab04-f903-4c3d-b327-de4034dbfcb9

Angular 15
0.43.38

Add platform table a lot of items and enable `scrollWholeRows`, `pageScrolling`, `virtualScroll`.
Expand first row.
Collapse the same row.
Scroll till scrollbar hitting bottom.

Result: it does not load next page of rows.
Expected it should load next page of rows.

If we disable `scrollWholeRows`, it will work fine.

https://stackblitz.com/edit/angular-rvez99-btza5n?file=src%2Fapp%2Fadvanced-scrolling-example.component.html

No

Yes

Yes

Please also consider expand/collapse all.

mikerodonnell89:
Fixes #12157 and #12160

### DIFF
--- a/libs/platform/src/lib/table/table.component.scss
+++ b/libs/platform/src/lib/table/table.component.scss
@@ -508,6 +508,7 @@ fdk-dynamic-portal {
 
     &__body--virtual-scroll {
         background-color: var(--sapList_Background, #fff);
+        will-change: scroll-position;
 
         .fd-table__body:not(:last-child) .fd-table__row:last-child .fd-table__cell {
             border-bottom-color: var(--sapList_BorderColor) !important;
@@ -515,6 +516,7 @@ fdk-dynamic-portal {
 
         .fd-table__body {
             position: static;
+            will-change: transform;
 
             &--scroll-mock {
                 top: 0;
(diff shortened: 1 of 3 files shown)


libs/platform/src/lib/table/table.component.scss
```diff
@@ -508,13 +508,15 @@ fdk-dynamic-portal {
 
     &__body--virtual-scroll {
         background-color: var(--sapList_Background, #fff);
```
CSEHoangV, line 510: Still lagging. I think from https://github.com/SAP/fundamental-ngx/pull/12049/files#diff-0562f5d038c3f170150e90bf1ea0967017e5e026269740c41e68a89512ead50bL507, you need also see if you can put back `position: relative;` since this resolve issue in my local.

dpavlenishvili (approved):

libs/platform/src/lib/table/table.component.scss
CSEHoangV: Issue is resolved

InnaAtanasova (approved):

Issue #11968, opened 2024-05-29
Grid List Keyboard support bugs


droshev:
#### Is this a bug, enhancement, or feature request?
bug

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)

latest

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;

Go to Grid List and: 
- [ ] If I tab, I navigate inside the item to the first interactive element in the item.  Further tabs lead me to the next interactive element within the item. If there are no more interactive elements within an item, the next tab leads to the first interactive element in the next item (instead of leaving the grid list)
- [ ] When using tab to navigate to the grid list, the item level is not in the tab chain. When using keyboard, there seems to be no way to get to item level and invoke arrow key navigation

PR #12173, opened 2024-07-24
[WIP]fix(core): Grid List Keyboard support bugs

khotcholava:
closes [#11968](https://github.com/SAP/fundamental-ngx/issues/11968)

fix(grid-list): ensure item level is in tab chain and enable arrow key navigation

## Description

- Resolved issue where tabbing inside the Grid List navigated only through interactive elements within items.
- Enabled arrow key navigation at the item level.

### DIFF
--- a/libs/core/grid-list/components/grid-list-item/grid-list-item.component.html
+++ b/libs/core/grid-list/components/grid-list-item/grid-list-item.component.html
@@ -1,7 +1,7 @@
 <div
     #gridListItem
     [id]="id"
-    tabindex="-1"
+    tabindex="0"
     class="fd-grid-list__item"
     role="option"
     [class.is-navigated]="isNavigated"


InnaAtanasova:
the title of the PR should be `fix(core): text...`
it's fix/feat/chore(core/platform/btp): short description. It's the same for the commits. Here's some info:
https://www.conventionalcommits.org/en/v1.0.0/

khotcholava:
I found a simple solution.
The problem was with the suggestions not getting values when dataSource was null.
I initially changed it to assign an empty array. When dataSource is null, it doesn't affect the suggestions, and the value will always be an empty array. However, if we pass an array, it will use the provided array.

droshev:
@khotcholava There are some more keyboard support things to be done. We will discuss it next week and come back to you.

Issue #12139, opened 2024-07-18
Message Strip Alerts aren't showing.


droshev:
#### Is this a bug, enhancement, or feature request?
bug?

#### Describe your proposal.

https://fundamental-ngx.netlify.app/#/core/message-strip -> Message Strip Alerts aren't showing

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
latest

khotcholava:
We need to update the documentation example. This is not a bug; it works correctly on the [StackBlitz example](https://stackblitz.com/run?file=package.json).

PR #12177, opened 2024-07-25
docs(core): resolve issue with alerts not displaying

khotcholava:
closes  [#12139](https://github.com/SAP/fundamental-ngx/issues/12139)

## Description
Message strip alerts are now shown depending on its positions and  placements

### DIFF
--- a/libs/core/message-strip/alert/message-strip-alert-container/message-strip-alert-container.component.scss
+++ b/libs/core/message-strip/alert/message-strip-alert-container/message-strip-alert-container.component.scss
@@ -1,6 +1,8 @@
 .fd-message-strip-alert-overlay {
     max-width: 33.33vw;
     padding: 0.5rem;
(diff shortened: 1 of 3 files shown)


mikerodonnell89 (approved):

libs/core/message-strip/alert/message-strip-alert-container/message-strip-alert-container.component.scss
```diff
@@ -1,6 +1,8 @@
 .fd-message-strip-alert-overlay {
     max-width: 33.33vw;
     padding: 0.5rem;
+    position: fixed;
```
droshev, line 4: I would prefer to move that in styles

droshev (changes requested):

InnaAtanasova (approved):

Issue #11854, opened 2024-04-30
Missing documentation for RtlService


LiorCapsiSAP:
We needed to use a control (menu in our case) in RTL. In the FD documentation, there is a button to "simulate RTL" but there is no documentation about how to implement it in our own application (using RtlService). I had to go to the documentation source code to see how the button was implemented there.
It would be a great help to have a section in the documentation about RtlService and how/when to use it. Maybe even add an RTL section to the doc of every control

PR #12198, opened 2024-07-30
docs(core): implement RTL service documentation

khotcholava:
closes [#11854](https://github.com/SAP/fundamental-ngx/issues/11854)

## Description
Added documentation for using RtlService to implement RTL in applications. Included examples and guidance on how and when to use RtlService, and added an RTL section to each control's documentation.


droshev:
the switch you add is interfering with the other one:
Screenshot 2024-07-30 at 09 15 19

InnaAtanasova (commented):
Screenshot 2024-07-30 at 9 35 05 AM
 why do you need 2 switches? When the top one is used, the bottom one seems broken.
I think it would make more sense to use as an example one of the languages that are mentioned above.

alexhristov14:
RtlService docs should be in CDK utilities instead of Core since the service is in CDK.

libs/docs/core/rtl-service/rtl-service-docs.component.html
```diff
@@ -0,0 +1,38 @@
...
+    <li>Hebrew (he)</li>
+    <li>Khowar (khw)</li>
+    <li>Kashmiri (ks)</li>
+    <li>Kurdish (ku)</li>
+    <li>Pashto (ps)</li>
+    <li>Urdu (ur)</li>
+    <li>Yiddish (yi)</li>
+</ul>
+<description>
+    You can override this list by providing a custom list view the <code>RTL_LANGUAGE</code> injection token.
```
mikerodonnell89, line 32:
```suggestion
    You can override this list by providing a custom list to the <code>RTL_LANGUAGE</code> injection token.
```

libs/docs/core/rtl-service/examples/rtl-service-basic-example/rtl-service-basic-example.component.ts
```diff
@@ -0,0 +1,50 @@
...
+    // Signal to track RTL state
+    rtl$ = computed(() => !!this._rtlService?.rtlSignal());
+    // Signal to track a direction
+    _dir$ = computed<Direction>(() => {
+        return this.rtl$() ? 'rtl' : 'ltr';
+    });
+
+    constructor() {
+        effect(() => {
+            this._updateTextContainerDirection();
```
mikerodonnell89, line 36: I am new to the signals API (still doing most of my work in Angular 15), so correct me if wrong but doesn't an effect need to have a reference to a signal inside of it? What is the purpose of having "effect" here?

libs/docs/core/rtl-service/rtl-service-docs.component.html
```diff
@@ -0,0 +1,38 @@
+<fd-docs-section-title id="rtl-service-usage" componentName="rtl-service"> RTL Service Usage </fd-docs-section-title>
+<description>
+    The <code>RtlService</code> service can be used to manage the Right-To-Left (RTL) settings in your application.
+</description>
+
+<description>
+    It uses a <code>BehaviorSubject</code> to track the RTL status and provides a <code>Signal</code> wrapper for
+    reactive programming. The service can be configured with a custom list of RTL languages using the
```
mikerodonnell89, line 8: An example of this would be cool too

mikerodonnell89 (changes requested):
Looking good! Couple comments above

libs/docs/core/rtl-service/examples/rtl-service-basic-example/rtl-service-basic-example.component.ts
mikerodonnell89, line 36: I am new to signals so forgive me (still doing most of my work in Angular 15). I am confused why there would be an effect here, what are you demonstrating?

khotcholava: If you look at the `_updateTextContainerDirection` function, it uses a signal within the function. I'm not directly using the signals in the effect; instead, `_updateTextContainerDirection` handles it. When the `_dir$()` signal changes, the effect will work.

libs/docs/core/rtl-service/rtl-service-docs.component.html
khotcholava: Added example.

libs/docs/core/rtl-service/examples/rtl-service-basic-example/rtl-service-basic-example.component.ts
mikerodonnell89: So doesn't clicking the toggle button call `_updateTextContainerDirection()`, which triggers the effect, which calls `_updateTextContainerDirection()` again? Is that intended?

khotcholava: You're absolutely correct. There's no need to call `_updateTextContainerDirection` explicitly on button click. The effect will automatically trigger whenever `_dir$()` changes.

mikerodonnell89 (approved):

dpavlenishvili (approved):

package.json
```diff
@@ -1,177 +1,178 @@
...
-      "yarn nx format:write --uncommitted"
-    ],
-    "*.{js,ts,html,md,json,yml}": [
-      "yarn eslint --fix"
-    ],
-    "*.properties": [
-      "yarn nx run i18n:transform-translations"
-    ]
-  }
+    "name": "fundamental-ngx",
```
droshev, line 2: why did you change this?

Issue #12212, opened 2024-08-01
Documentation navigation on home page


khotcholava:
#### Is this a bug, enhancement, or feature request?
bug
#### Describe your proposal.
Navigation does not work when clicking a content list item on the documentation's home page.

#### Can you handle that on the application side
no

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
latest
#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
Go to core docs home page.
Go to "Content" section.
Click some of the content list item. 
Navigation does not work.

#### Please provide stackblitz example(s).
https://sap.github.io/fundamental-ngx/#/core/home
#### In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations.
Yes
#### Is there anything else we should know?

**_IMPORTANT: Please refrain from providing links or screenshots of SAP's internal information, as this project is open-source, and its contents are accessible to anyone._**

PR #12213, opened 2024-08-02
fix(core): fixed link navigation on main page

khotcholava:
closes [#12212](https://github.com/SAP/fundamental-ngx/issues/12212)

## Description
Fixed the bug where navigation did not work when clicking on a content list item on the documentation's home page. This ensures that users can navigate correctly by selecting items from the content list.


mikerodonnell89 (changes requested):
I don't think turning off the sanitizer is the fix we want for this

mikerodonnell89:
I think we can basically do what this little thing is doing in the rest of our docs:
Screenshot 2024-08-02 at 10 25 53 AM

and set up routes for these sections, take a look at https://github.com/SAP/fundamental-ngx/blob/main/libs/docs/shared/src/lib/core-helpers/docs-section-title/docs-section-title.component.ts

mikerodonnell89:
I'd say just get rid of the links altogether

Issue #12206, opened 2024-07-31
Grid List documentation example optimization


InnaAtanasova:
#### Is this a bug, enhancement, or feature request?
Enhancement

#### Describe your proposal.
Instead of individual examples for every selection mode in Grid List create one example and use Segmented Button to switch the modes. 
![Screenshot 2024-07-31 at 10 40 15 AM](https://github.com/user-attachments/assets/85e333a9-e704-4d02-8f9a-edbdee2df1e5)

PR #12217, opened 2024-08-02
docs(core): grid list docs optimization

alexhristov14:
closes https://github.com/SAP/fundamental-ngx/issues/12206

## Description
Created one example of all grid list modes and switching between modes with segmented button.


libs/docs/core/grid-list/examples/combo-select/grid-list-combo-select-example.component.html
```diff
@@ -0,0 +1,241 @@
...
+                        <ng-template fd-grid-list-item-body>
+                            <p>781 Main Street</p>
+                            <p>Anytown, SD 57401</p>
+                            <p>USA</p>
+                            <a [routerLink]="['./']" fd-link tabindex="0">john_li&#64;example.com</a>
+                        </ng-template>
+                    </fd-grid-list-item>
+                }
+            </fd-grid-list>
+            <span aria-hidden="true" [style.display]="'none'" id="grid-list-button-1">Sort</span>
```
mikerodonnell89, line 69: No fault of your PR, but I'm not sure why we had this set up with these hidden spans, can just use data binding on the button's aria-label

droshev (approved):

libs/docs/core/grid-list/examples/index.ts
```diff
@@ -29,5 +30,6 @@ export const COMPONENTS = [
     GridListDndExampleComponent,
     GridListLayoutExampleComponent,
     GridListFocusingItemExampleComponent,
-    GridListAutoHeightExampleComponent
+    GridListAutoHeightExampleComponent,
```
InnaAtanasova, line 33: You need to remove the imports of the components that are not used as well as their registration.

dpavlenishvili (approved):

Issue #12218, opened 2024-08-02
Grid list item with multi selection and navigation - can't select item via mouse click


mikerodonnell89:
When trying to select a grid list item that has navigation in multiselect mode with a mouse click, only the navigation event is fired, and the select does not go through.  See vid

https://github.com/user-attachments/assets/aeb3559b-2c21-4dd2-8885-563a6646a539

droshev:
@khotcholava Can you check this one after you fix the grid list keyboard? thanks

PR #12229, opened 2024-08-05
fix(core): ensure grid list item selection works

khotcholava:
fix(core): ensure grid list item selection works with navigation in multiselect mode.

closes [#12218](https://github.com/SAP/fundamental-ngx/issues/12218)

## Description
When trying to select a grid list item that has navigation in multiselect mode with a mouse click, only the navigation event was fired, and the select did not go through. This fix ensures that both the navigation and selection events are handled correctly.

### DIFF
--- a/libs/core/grid-list/components/grid-list-item/grid-list-item.component.ts
+++ b/libs/core/grid-list/components/grid-list-item/grid-list-item.component.ts
@@ -407,6 +407,9 @@ export class GridListItemComponent<T> implements AfterViewInit, OnDestroy {
 
     /** @hidden */
     _onClick(event: MouseEvent): void {
+        console.log({
+            target: event.target
+        });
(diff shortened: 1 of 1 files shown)


libs/core/grid-list/components/grid-list-item/grid-list-item.component.ts
```diff
@@ -407,6 +407,9 @@ export class GridListItemComponent<T> implements AfterViewInit, OnDestroy {
 
     /** @hidden */
     _onClick(event: MouseEvent): void {
+        console.log({
+            target: event.target
+        });
```
InnaAtanasova, line 412: please remove the console log

dpavlenishvili (approved):

InnaAtanasova (approved):

PR #12231, opened 2024-08-05
fix(core,platform): added read only / non-editable state to checkboxes

### LINKED ISSUE #12204: checkbox readonly state

bug and enhancement

#### Describe your proposal.
the checkbox in NGX has issues with readonly state:
- it is missing in core
- there is an input `readonly` in platform but it is not working. ([example](https://stackblitz.com/edit/4ruxra?file=src%2Fapp%2Fplatform-binary-checkbox.component.html,src%2Fapp%2Fplatform-binary-checkbox.component.ts))

We need to bring it and implement it properly. You can find how it looks in [styles](https://fundamental-styles.netlify.app/?path=/docs/sap-fiori-components-forms-checkbox--docs)

latest

alexhristov14:
fixes #12204

## Description
- Added the readonly state 
- Updated the checkbox docs by adding both the readonly and displayOnly to the states section


droshev:
@alexhristov14 Can you write unit test for the feagture?

droshev:
@alexhristov14 Can you add also an example(read-only and display) with the interim state too?

droshev (approved):

mikerodonnell89 (approved):

PR #12233, opened 2024-08-06
fix(core): multi input tokens should not have title attr

### LINKED ISSUE #11675: [P4][Sourcing][Accessibility]Add Content From Sourcing Library-tooltip is not showing the exact text
Bug

#### Describe your proposal.
After selecting value from multi input, tooltip is not showing the exact text for the selected values.

No

Angular: 15
core / platform: 0.43.29

1. Select values from multi input.
2. Mouse over on the selected value.
3. Tooltip should show for the selected value.

https://github.com/SAP/fundamental-ngx/blob/main/libs/core/multi-input/multi-input.component.html

https://stackblitz.com/edit/6jb1hz?file=src%2Fapp%2Fmulti-input-example.component.ts,src%2Fapp%2Fmulti-input-example.component.html

MC

Screenshot 2024-06-14 at 12 53 15 PM

mikerodonnell89:
fixes #11675


InnaAtanasova (approved):

InnaAtanasova:
The issue is reported for Angular 15. You need to open a fix in the ng15 branch, too

mikerodonnell89:
> The issue is reported for Angular 15. You need to open a fix in the ng15 branch, too

So actually the title attribute should never have been set on the tokens. They were added to the main branch only, and the ng15 branch correctly doesn't have them. See my comment here: https://github.com/SAP/fundamental-ngx/issues/11675#issuecomment-2271858613

dpavlenishvili (approved):

droshev:
@mikerodonnell89 looks like it was added for a reason. The original issue is [this](https://github.com/SAP/fundamental-ngx/issues/10849). Can you double check?

mikerodonnell89:
Waiting for more info from designers before merging this

mikerodonnell89:
After speaking with the designers about this issue, here's what I've learned:

- the only instance in which a token should be truncated is when there is only one selected token. Otherwise, the multi-input will simply display the "_ more" text, and any other tokens that are short enough to fit in the multi-input along with the "_ more" text.
- when there is a truncated token displayed, rather than adding a title attribute, clicking the truncated token will display the multi-input's popover, displaying the full text of the token. Title attributes are not helpful for keyboard or touch screen users

PR #12240, opened 2024-08-09
fix(core): type incorrect assumption of input always existing

### LINKED ISSUE #12001: Console errors throwing in fd-multi-input
It is a bug

Component affected: https://sap.github.io/fundamental-ngx/#/core/multi-input

#### Describe your proposal.

Console errors throwing in production in this function 
xe._makeSearchTermChangeAnnouncements
with error as Cannot read properties of undefined (reading 'length') 

No

Angular Version: 15.2.9
ngx version: 0.43.21

ZsDeak-SAP:
## Related Issue(s)
closes #12001 

## Description

Input might be undefined, we can't assume it's not. Added an example, if you type `undefined` to that input, on `main` it'll reproduce the error of the linked issue. \
Type of filterFn is wrong on multi-input.component (this['dropdownValues']) as it contains a package private type. This is why in the example I used anys

### DIFF
--- a/libs/docs/core/multi-input/examples/multi-input-example/multi-input-example.component.ts
+++ b/libs/docs/core/multi-input/examples/multi-input-example/multi-input-example.component.ts
@@ -14,6 +14,7 @@ export class MultiInputExampleComponent {
     secondSelected = ['Pineapple'];
     thirdSelected = ['Kiwi'];
     selectedDisabledAutoComplete = ['Kiwi'];
+    filterSelected = [];
 
     objectValues = [
         { label: 'Apple', value: 'apple_value' },
@@ -31,5 +32,9 @@ export class MultiInputExampleComponent {
 
     searchTermChange(text: string): void {
         console.log(text);
+    }
+
+    customFilter(currentArray: (string | any)[], searchTerm: string): any {
+        return searchTerm === 'undefined' ? undefined : currentArray.filter((v) => v.startsWith(searchTerm));
     }
 }
--- a/libs/docs/core/multi-input/examples/multi-input-example/multi-input-example.component.ts
+++ b/libs/docs/core/multi-input/examples/multi-input-example/multi-input-example.component.ts
@@ -14,7 +14,6 @@ export class MultiInputExampleComponent {
     secondSelected = ['Pineapple'];
     thirdSelected = ['Kiwi'];
     selectedDisabledAutoComplete = ['Kiwi'];
-    filterSelected = [];
 
     objectValues = [
         { label: 'Apple', value: 'apple_value' },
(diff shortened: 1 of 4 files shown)


alexhristov14:
> @droshev [this change](https://github.com/SAP/fundamental-ngx/pull/12217) i suspect failed some e2e tests, so i tried to fix them as well

I've created a new PR for the e2e testing commit #12253 so you can revert it from this one and only keep the changes for the multi-input issue

libs/docs/core/multi-input/examples/multi-input-example/multi-input-example.component.ts
```diff
@@ -32,4 +33,8 @@ export class MultiInputExampleComponent {
     searchTermChange(text: string): void {
         console.log(text);
     }
+
+    customFilter(currentArray: (string | any)[], searchTerm: string): any {
```
dpavlenishvili, line 37: update return type

ZsDeak-SAP: > Type of filterFn is wrong on multi-input.component (this['dropdownValues']) as it contains a package private type. This is why in the example I used anys

mikerodonnell89 (changes requested):
The linked issue and the comment thread on that issue, and the comment thread here, don't make it clear exactly what this PR is trying to solve. I think a stackblitz demonstrating the problem in the linked issue would go a long way. But until we have a clear explanation, I don't think we can move forward with merging this

ZsDeak-SAP:
@mikerodonnell89 I hope you see how it's difficult to create a stackblitz to an wrongly working internal function. nevertheless here you can [find an example](https://stackblitz.com/edit/zdi7tq?file=src%2Fapp%2Fmulti-input-example.component.ts). type in 'undefined' and you'll get a console error. This example is partly incorrect, as `filterFn` would expect an array returning function, and does have a default value, but only through that can I reach (right now) the `multiAnnouncerOptions` input. 

`multiAnnouncerOptions` does not have a default value and is not a required input, and as such we cannot assume it to exist. this will cause the console error above.

libs/docs/core/multi-input/examples/multi-input-example/multi-input-example.component.ts
mikerodonnell89: I don't think we want to include an example of a deliberately broken filter function in our docs

mikerodonnell89 (changes requested):
Looks good but I think let's remove the new example you added. There's already the custom filter example, and I think the `return searchTerm === 'undefined' ? undefined` is going to be confusing to anyone who doesn't know the (very specific) context of this issue

droshev:
@ZsDeak-SAP did you see what @mikerodonnell89 suggested?

ZsDeak-SAP:
sorry, yes, fixed

droshev:
@mikerodonnell89 Can you review the latest changes?

PR #12288, opened 2024-08-21
fix(core): grid list keyboard support

khotcholava:
fix(core): Enhance keyboard navigation and accessibility

closes [#11968](https://github.com/SAP/fundamental-ngx/issues/11968) 

## Description
- Implemented keyboard navigation for the grid list using arrow keys, Tab, F2, and F7 keys.
- Added Tab navigation to move focus between grid items, similar to arrow key navigation.
- Ensured proper handling of child item selection within the grid list using Tab.
- Enabled exiting the grid list when a child item is selected by pressing the Tab key.


libs/core/grid-list/components/grid-list-item/grid-list-item.component.ts
```diff
@@ -279,9 +281,13 @@ export class GridListItemComponent<T> implements AfterViewInit, OnDestroy {
     get selectionMode(): GridListSelectionMode | undefined {
```
dpavlenishvili, line 281: create selectionMode enum too and use instead of strings

droshev:
1. focus on the first item from the grid list
2. press F2 (it moves the focus on the first focusable element)
3. press F2 - it doesn't go back to the focusing the whole grid list item

droshev (changes requested):
see my comments

mikerodonnell89 (changes requested):
Looking good, two things I noticed during testing: 

1. In the "delete" mode, if you focus the grid list, and then move focus from item to item, then go back and delete the first item, I cannot use tab to focus any items in the grid list. See video

https://github.com/user-attachments/assets/56dca8d1-ccc5-4e7d-9f52-427bd174f919

2. Would be nice if an Escape key press moved focus from child items back to the parent, similar to what we have for table, list etc

mikerodonnell89 (approved):

libs/core/grid-list/components/grid-list/grid-list.component.ts
dpavlenishvili, line 1: image


unit test is failing

dpavlenishvili (approved):

PR #12300, opened 2024-08-23
feat(platform): move the column reordering button in the list item representing the column

### LINKED ISSUE #12175: Platform table personalization dialog column ordering updates
The table column ordering p13 dialog needs to match the new design that we seen in ui5 etc, where the buttons to reorder a column are on the list item representing the column itself, rather than how we have it now, in the dialog toolbar.

Current: 
Screenshot 2024-07-24 at 12 08 24 PM

Desired:
image

shawchandankumar:
Currently, the column reorder button is in the personalization dialog toolbar, which doesn't clearly show which of the following list items is active. By moving the reordering buttons on an individual list item whichever list item is currently active, it shows the reordering buttons for that particular list item.

Closes #12175

### Before:
image

### After:
image


mikerodonnell89 (changes requested):
Works well with mouse clicks but keyboard support is broken - if you navigate to the up/down buttons using just keystrokes, then press enter while the up or down button is focused, it deselects the column rather than moving it

mikerodonnell89:
Looking better but why does the move-row-up button lose focus after pressing enter?

shawchandankumar:
> Looking better but why does the move-row-up button lose focus after pressing enter?

It's because of rendering after a change I believe. In move down button it's not happening

shawchandankumar:
> Looking better but why does the move-row-up button lose focus after pressing enter?

@mikerodonnell89 Is it good now?

mikerodonnell89 (approved):

PR #12305, opened 2024-08-23
fix(core): add fix for the missing padding in Calendar

InnaAtanasova:
## Related Issue(s)

closes none

## Description
for some reason the CSS in fund-ngx renders the padding overwrite in a wrong way. Will attempt to fix it in styles as well, but for now this solution will make sure the padding is rendered properly/


dpavlenishvili (approved):

droshev:
@InnaAtanasova Can't we have this from styles? I brought the latest styles btw.

PR #12324, opened 2024-08-26
feat(core): popover modal styling

### LINKED ISSUE #12210: [FEATURE REQUEST] Add modal property to Popover control

It is a feature request.

#### Describe your proposal.

Is it possible to add a "modal" property to the Popover control like it is working in sapui5? https://sapui5.hana.ondemand.com/#/api/sap.m.Popover%23constructor

image

No

- Angular 17.3.1
- FundamentalNGX 0.51.1 (current available version)

Yes, I didn't find it here - https://sap.github.io/fundamental-ngx/#/core/popover/api

alexhristov14:
closes #12210 
Styling for closeOnOutsideClick PR: https://github.com/SAP/fundamental-styles/pull/5595


InnaAtanasova (commented):
1. Please remove the empty scss file
2. Could you add an example in the documentation that illustrates this case? This way we can test in future the functionality if something goes wrong.

InnaAtanasova (commented):
Currently when `closeOnOutsideClick]="false"` you can't close the popover with the Esc key. 
You also need a case where you can't close the popover with both, outside click and esc key

InnaAtanasova:
Screenshot 2024-08-28 at 9 31 40 AM

You only show the HTML file, need to show the TS file, too

libs/docs/core/popover/popover-docs.component.ts
```diff
@@ -114,6 +118,16 @@ export class PopoverDocsComponent {
...
     ];
 
+    popoverClosingExample: ExampleFile[] = [
+        {
+            language: 'html',
+            code: getAssetFromModuleAssets(popoverClosingSrc),
+            fileName: 'popover-closing-example',
+            typescriptFileCode: getAssetFromModuleAssets(popoverClosingSrcTs),
+            component: 'PopoverClosingExampleComponent'
+        }
```
InnaAtanasova, line 128:
```
{
            language: 'html',
            code: getAssetFromModuleAssets(popoverClosingSrc),
            fileName: 'popover-closing-example'
        },
        {
            language: 'typescript',
            code: getAssetFromModuleAssets(popoverTriggerSrcTs),
            fileName: 'popover-closing-exampl',
            typescriptFileCode: getAssetFromModuleAssets(popoverClosingSrcTs),
            component: 'PopoverClosingExampleComponent'
        }
```

InnaAtanasova (approved):

khotcholava:
@alexhristov14 It seems that the popover close functionality is not working as expected. I tested it and found that pressing the ESC key or clicking outside the popover does not close it. Could you please look into this issue?

alexhristov14:
> @alexhristov14 It seems that the popover close functionality is not working as expected. I tested it and found that pressing the ESC key or clicking outside the popover does not close it. Could you please look into this issue?
> 
>  Screen.Recording.2024-09-03.at.15.26.36.mov

I've tested it again now and it still works as intended. The `closeOnOutsideClick` property prevents users to close the popover when clicking outside of the body (when set to false) but it's still closable with ESC. As for the `closeOnEscapeKey`, it prevents users to close the popover body with ESC (when set to false). And the third example is a combination of both where neither clicking on the outside nor the ESC button closes the body. The only way to close it back it to click the popover yet again.

The default behaviour is that it closes in both cases: whenever clicked outside or the ESC button is pressed.

InnaAtanasova:
> @alexhristov14 It seems that the popover close functionality is not working as expected. I tested it and found that pressing the ESC key or clicking outside the popover does not close it. Could you please look into this issue?
> 
>  Screen.Recording.2024-09-03.at.15.26.36.mov

@khotcholava it's the naming of the button that makes the confusion. `closeOnOutsideClick` button opens a popover that is not supposed to close on outside click. 
Screenshot 2024-09-03 at 1 53 20 PM

As you can see the property is set to false. But I agree with you. The way we have it now is confusing . It makes more sense if the button was named something like "preventCloseOnOutsideClick" and same for the other examples. @alexhristov14

Issue #12232, opened 2024-08-06
Issue In Combobox Field Delete Search Term With Chinese Characters


claudiuzsap:
#### Is this a bug, enhancement, or feature request?
Bug
#### Can you handle that on the application side
Unfortunately not
#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
Latest
#### If this is a bug, please provide steps for reproducing it; the exact components you are using;

1. Add **_Pinyin - Traditional Chinese_** language keyboard in your system
2. Navigate to Fundamental NGX documentation website on Combobox Field: https://sap.github.io/fundamental-ngx/#/core/combobox
3. Go to first Combobox field example which has 5 fruits as options: Apple, Pineapple, etc.
4. Select any option from the Combobox field  (ex: Apple)
5. Make sure to have activated the new Chinese language
6. In the Search Input after the selected option (ex: Apple) add any Chinese character without using copy paste of a Chinese character, it is needed to be typed from the keyboard (My guess is that here is the problem because a Chinese character if you type it from the keyboard, after you press a keyboard letter then you have to choose using a number what variant of that character you want to use. Also if you don't have any value selected the issue is not reproducible, this could be also a hint)
7. Then try to delete characters from search term of Combobox and you will see that the Combobox search term got into a broken state and is stuck (In the video attached after I made the same steps I was pressing Backspace continuously and couldn't delete the search term because it got stuck. Switching to another option of the Combobox didn't fix the state)

https://github.com/user-attachments/assets/d143c5bb-8b62-4e54-be96-fa5f72da7dab

PR #12340, opened 2024-08-30
fix(core): combobox handle IME input correctly to prevent accidental deletion of characters

khotcholava:
fix(core): handle IME input correctly to prevent accidental deletion of characters

closes [#12232](https://github.com/SAP/fundamental-ngx/issues/12232)

## Description
- Added handling for `compositionstart` and `compositionend` events to properly manage IME input (e.g., Chinese, Japanese).
- Introduced `isComposing` flag to bypass autocomplete logic during IME composition.
- Updated `inputText` synchronization after IME input ends to ensure it matches the finalized value.
- Prevented unwanted modifications of input value when using IME, addressing the issue where typing and deleting partially composed characters could remove the last English character.

### DIFF
--- a/libs/cdk/utils/directives/auto-complete/auto-complete.directive.ts
+++ b/libs/cdk/utils/directives/auto-complete/auto-complete.directive.ts
@@ -3,7 +3,7 @@ import { Directive, ElementRef, EventEmitter, Input, NgZone, Output, inject } fr
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { fromEvent, map, switchMap } from 'rxjs';
 import { first } from 'rxjs/operators';
-import { KeyUtil } from '../../functions/key-util';
+import { KeyUtil } from '../../functions';
 
 export interface AutoCompleteEvent {
     term: string;
@@ -58,6 +58,9 @@ export class AutoCompleteDirective {
     private lastKeyUpEvent: KeyboardEvent;
 
     /** @hidden */
+    private isComposing = false;
+
+    /** @hidden */
     private readonly _elementRef = inject(ElementRef);
 
     private readonly _zone = inject(NgZone);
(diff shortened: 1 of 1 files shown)


libs/cdk/utils/directives/auto-complete/auto-complete.directive.ts
```diff
@@ -57,6 +57,9 @@ export class AutoCompleteDirective {
     /** @hidden */
     private lastKeyUpEvent: KeyboardEvent;
 
+    /** @hidden */
+    private isComposing = false;
```
InnaAtanasova, line 61: We usually prefix the private members with underscore. And I know the property above is private and not prefixed. We missed it during the review.

libs/cdk/utils/directives/auto-complete/auto-complete.directive.ts
```diff
@@ -72,6 +75,12 @@ export class AutoCompleteDirective {
          */
         this._zone.runOutsideAngular(() => {
             const keyupEvent = fromEvent<KeyboardEvent>(this._elementRef.nativeElement, 'keyup');
+            const compositionStartEvent = fromEvent<CompositionEvent>(
+                this._elementRef.nativeElement,
+                'compositionstart'
```
InnaAtanasova, line 80: Screenshot 2024-09-03 at 1 47 09 PM
Screenshot 2024-09-03 at 1 47 00 PM

'compositionstart' and 'compositionend' are experimental technology and not supported in all browsers:
https://developer.mozilla.org/en-US/docs/Web/API/EditContext/compositionstart_event

khotcholava: Hi @InnaAtanasova, I found this [MDN article on the `compositionstart` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event), which mentions that it is supported by all major browsers. I also tested it on Safari, Edge, Firefox, and Chrome, and it worked without any issues.

dpavlenishvili (approved):

droshev:
@khotcholava can you investigate what is wrong with CI?

PR #12367, opened 2024-09-05
fix(platform): prevent filter input from losing focus when entering s…

### LINKED ISSUE #12285: [P4] [Sourcing] [Platform] Table column filtering does not allow multiple words
Bug in Angular 15

#### Describe your proposal.
The column filtering does not allow multiple words. When users attempt to enter spaces, the filter textbox loses focus and the space character is not entered. This can be a problem, since users may need to include spaces in their filter query.

Angular 15

Video attached. In this case, the attempted query is "Stocked on demand", but users can only get so far as to type "Stocked". Attempting to enter the space character makes it lose focus.

https://github.com/user-attachments/assets/520f1eec-8950-40e4-93d3-8f85fc027ab5

Yes

Yes

dpavlenishvili:
…paces

## Related Issue(s)

closes #12285 

### Before:
image

### After:
image

### DIFF
--- a/libs/platform/table/components/table-cell-header-popover/table-cell-header-popover.component.html
+++ b/libs/platform/table/components/table-cell-header-popover/table-cell-header-popover.component.html
@@ -99,6 +99,7 @@
                     fd-form-control
                     [id]="'fdp-table-column-filtering-' + column.name"
                     (keydown.enter)="_setColumnHeaderFilterBy(column.key, $any($event.target)?.value)"
+                    (keydown)="$event.stopPropagation()"
                 />
             </div>
         </div>


libs/platform/table/components/table-cell-header-popover/table-cell-header-popover.component.html
```diff
@@ -99,6 +99,7 @@
                     fd-form-control
                     [id]="'fdp-table-column-filtering-' + column.name"
                     (keydown.enter)="_setColumnHeaderFilterBy(column.key, $any($event.target)?.value)"
+                    (keydown)="$event.stopPropagation()"
```
mikerodonnell89, line 102: I am confused why this works but it does

dpavlenishvili: The second (keydown) event listener with stopPropagation() ensures that the keydown event does not bubble up the DOM tree. This prevents unintended behavior such as triggering other event listeners higher in the component hierarchy that could interfere with the expected behavior of input field. Without this, the keydown event might propagate to other components, such as the popover or list items, causing the input field to lose focus or trigger other actions.

mikerodonnell89: I think leave a comment here explaining the situation, I'd be confused if I stumbled upon this without contrext

dpavlenishvili: image

droshev:
@dpavlenishvili this change needs also to be donwported to ng-15. You can ask @mikerodonnell89 to downport it in a separate PR, thanks.

khotcholava (approved):

PR #12370, opened 2024-09-05
feat(core): implement ContentDensity to handle all type of themes on demand.

### LINKED ISSUE #11368: No way to set `fdCozy` to `fd-dialog` component on desktop viewport
Bug

#### Describe your proposal.
Make `cozy` / `compact` configurable for `fd-dialog` or at least make it possible to override the default content density

Yes, some workarounds - using `mobile: true` in dialog config even for desktop and overriding some CSS

I'm currently using `@angular/*: 17.1.2` and `@ngx-fundamental:0.48.4`

In the `fd-dialog` component setting `fdCozy` to any element doesn't change anything, elements are still `compact`. This is caused by setting `fdCozy` only for dialogs opened on `mobile` view in the source code, see https://github.com/SAP/fundamental-ngx/blob/v0.48.4/libs/core/dialog/dialog-header/dialog-header.component.html#L4. 

Sample markup (pretty standard) we use in our project:
```html
<fd-dialog>
  <fd-dialog-header>
    <ng-template fdkTemplate="header">
      <div fd-bar-left>
        <fd-bar-element>...</fd-bar-element>
      </div>
      <div fd-bar-right>
        <fd-button-bar></fd-button-bar>
      </div>
    </ng-template>

    <ng-template fdkTemplate="subheader">
      <div fd-bar-middle>
        <fd-bar-element>
          <fd-input-group></fd-input-group>
        </fd-bar-element>
      </div>
    </ng-template>
  </fd-dialog-header>

  <fd-dialog-body>...</fd-dialog-body>
  <fd-dialog-footer>...</fd-dialog-footer>
</fd-dialog>
```

`-`

`-`

Yes

Yes

`-`

dpavlenishvili:
…nsity to handle all type of themes on demand.

## Related Issue(s)

closes #11368


libs/docs/core/dialog/examples/dialog-mobile/dialog-mobile-example.component.ts
```diff
@@ -36,7 +37,8 @@ export class DialogMobileExampleComponent {
             mobile: true,
             responsivePadding: true,
             ariaLabelledBy: 'fd-dialog-header-5',
-            ariaDescribedBy: 'fd-dialog-body-5'
+            ariaDescribedBy: 'fd-dialog-body-5',
+            contentDensity: ContentDensityMode.COZY
```
mikerodonnell89, line 41: Doesn't appear to be overriding the app-level density:

Screenshot 2024-09-11 at 9 23 00 AM

mikerodonnell89 (changes requested):
see comment:

libs/docs/core/dialog/examples/dialog-mobile/dialog-mobile-example.component.ts
dpavlenishvili: I reviewed the requirements and implementation again, and I think the expectations align with the requested requirements. Everything is still working as it was before. The only addition is the ability to configure content density, which was a requested requirement. Previously, users encountered issues setting fdCozy manually and had to use a workaround by setting mobile: true in the dialog configuration. This workaround transformed the header, footer, and other components to a cozy style but was not the desired approach, because fdcozy only was available on mobile. Now, users can configure different content densities directly within any dialog configuration.

It's important to note that when content density is set manually, it remains fixed and does not update with global content density changes. This behavior is consistent with how it worked previously in mobile mode as well.

droshev:
@mikerodonnell89 Can you check the latest changes?

libs/docs/core/dialog/dialog-docs.component.ts
```diff
@@ -382,6 +384,7 @@ export class DialogDocsComponent {
         private _dialogService: DialogService
     ) {
         this.schema = this._schemaFactory.getComponent('dialog');
+        console.log(this.schema);
```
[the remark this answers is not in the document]
khotcholava: @dpavlenishvili ping

khotcholava (changes requested):

droshev:
@mikerodonnell89 Can you check the latest changes?

droshev:
@dpavlenishvili can you check the failing unit?

Issue #12336, opened 2024-08-30
[P2][Sourcing][Localisation][PLATFORM Multi-Combobox] : Invalid entry - error message in english not translated to Hungarian language


184801:
#### Is this a bug, enhancement, or feature request?
BUG

#### Describe your proposal.
**Invalid entry - error message EN**
Expectation :  Invalid Entry" error message is  not translated into the Hungarian  language.

#### Can you handle that on the application side
No

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
Angular: 15
core/ platform: 0.43.40

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
*"Invalid Entry" error translation in fdp-multi-combobox*
Open the fdp-multi-combobox Component
Enter Text Not Present in the List
Observe the "Invalid Entry" Error
Translate the "Invalid Entry" Error Message
"Invalid Entry" error message is  not translated into the Hungarian language.

https://github.com/user-attachments/assets/1a41fe97-43fe-499b-b2f6-13982595e7f9

#### Please provide stackblitz example(s).
https://stackblitz.com/edit/angular-6mvwz6?file=src%2Fapp%2Fmulti-combobox-columns-example.component.ts

#### Is there anything else we should know?

**_IMPORTANT: Please refrain from providing links or screenshots of SAP's internal information, as this project is open-source, and its contents are accessible to anyone._**

mikerodonnell89:
fixed and will be published in 0.43.44

PR #12410, opened 2024-09-12
fix(core,platform): provide translation for multi-combobox Invalid En…

mikerodonnell89:
part of #12336

### DIFF
--- a/libs/core/src/lib/multi-combobox/multi-combobox.component.ts
+++ b/libs/core/src/lib/multi-combobox/multi-combobox.component.ts
@@ -7,9 +7,11 @@ import {
     ContentChildren,
     ElementRef,
     EventEmitter,
+    Inject,
     InjectionToken,
     Injector,
     Input,
+    OnDestroy,
     OnInit,
     Output,
     QueryList,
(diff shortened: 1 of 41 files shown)


libs/core/src/lib/multi-combobox/multi-combobox.component.ts
```diff
@@ -359,6 +379,11 @@ export class MultiComboboxComponent<T = any> extends BaseMultiCombobox<T> implem
         this._openDataStream();
     }
 
+    /** @hidden */
+    ngOnDestroy(): void {
+        this._langSubscription.unsubscribe();
+    }
+
```
dpavlenishvili, line 386: maybe you remove _langSubscription, directly subscribe _lang$ and use takeUntilDestory. with that way you remove langSub variable and onDestory hook.

dpavlenishvili: @mikerodonnell89 everything looks good, i just left only one comment related on onDestory

mikerodonnell89: Unfortunately for this PR, takeUntilDestroyed was introduced in Angular 16

dpavlenishvili (approved):

PR #12442, opened 2024-09-18
feat(core): add avatar Indication colors

droshev:
add avatar Indication colors


droshev:
successor https://github.com/SAP/fundamental-ngx/pull/12444

PR #12453, opened 2024-09-20
fix(core): toolbar render spacer before overflow conditionally

### LINKED ISSUE #12239: Bug: Overflow of fdp table toolbar leads to changing spacing when resizing the viewport
Bug
#### Describe your proposal

- When using the fdp-table with a table toolbar and adding additional buttons to the toolbar actions, resizing the browser window leads to a growing and shrinking spacing between the sort button and the overflow button (see stackblitz example and video below). This is not intended.
- **Expected behavior: all actions should always be aligned to the right with fixed spacing in between**
- Additionally the automatically appearing separators (grey lines) in between the search bar, filter button, the additional actions and the overflow button are not needed and can be removed

No
Latest
see description above
https://stackblitz.com/edit/775juw?file=src%2Fapp%2Fplatform-table-sortable-example.component.html
#### Screen Recording
https://github.com/user-attachments/assets/3a95d7ae-3085-4fd1-9d5b-f0c0465b5c9a

droshev:
## Related Issue(s)

closes #12239 

## Description

render spacer before the overflow button conditionally (only if the application doesn't have non-fixed spacer)


dpavlenishvili (approved):

libs/core/toolbar/toolbar.component.ts
```diff
@@ -168,6 +168,9 @@ export class ToolbarComponent implements AfterViewInit, AfterViewChecked, CssCla
         return this._titleComponent$.value;
     }
 
+    /** @hidden */
+    @ContentChildren(ToolbarSpacerDirective) spacer: QueryList<ToolbarSpacerDirective>;
```
InnaAtanasova, line 172: you can use signals for this as well: https://angular.dev/guide/signals/queries#contentchildren
spacers = contentChildren(ToolbarSpacerDirective);

libs/core/toolbar/toolbar.component.ts
```diff
@@ -178,6 +181,9 @@ export class ToolbarComponent implements AfterViewInit, AfterViewChecked, CssCla
     /** @hidden */
     overflownItems: ToolbarItem[] = [];
 
+    /** @hidden */
+    spacerUsed = false;
```
InnaAtanasova, line 185: if you use `contentChildren` signal, this can also be a computed signal and then its value will change when the spacers are updated.

libs/core/toolbar/toolbar.component.ts
droshev: check out the changes I have pushed

libs/core/toolbar/toolbar.component.ts
droshev: check out the changes I have pushed

mikerodonnell89 (changes requested):
The overflow priority example doesn't look great on small screens:

this pr:
Screenshot 2024-09-24 at 2 04 01 PM

main branch:
Screenshot 2024-09-24 at 2 04 06 PM

mikerodonnell89 (approved):
Approving as the issue I pointed out was pre-existing, but not apparent in the example due to the example previously not having the spacer. We can address that in another issue

droshev:
> The overflow priority example doesn't look great on small screens:
> 
> this pr: Screenshot 2024-09-24 at 2 04 01 PM
> 
> main branch: Screenshot 2024-09-24 at 2 04 06 PM

should be fixed now:

PR #12467, opened 2024-09-23
fix(platform): virtual scroll should account for popping columns

### LINKED ISSUE #12263: [BUG] Performance issues for Table fundamental-ngx component
Bug

Describe your proposal.
In the [attached example](https://stackblitz.com/edit/sjxljz-u5wgpm), the performance of the table decreases more and more after loading over 700 elements, especially when opening the dialog with the action button. It takes several seconds to load more elements or to open and close the dialog.

No

Fundamental: v0.51
Angular: 18.0.3

Steps to reproduce:

1. Open attached example of table.
3. Click the growing button until over 900 elements are loaded.
4. Click the delete icon on any of the items to open dialog.
5. Observe how long it takes to open the dialog.

Expected behavior: 
Items load faster when the growing button is used, and the dialogs open immediately with a large number of items in the table.

https://stackblitz.com/edit/sjxljz-u5wgpm

Did you check the documentation and the API?
Yes

Did you search for similar issues?
Yes

mikerodonnell89:
fixes #12263 

Updates virtual scroll logic to cooperate with "popping" columns on small screens. Also adds a new input `[secondaryRowHeight]` which behaves similarly to `[rowHeight]`, but for the popping columns which are moved into the first column when the screen is set to a small width. Additionally, this PR updates the docs to explain that `[rowHeight]` (as well as `[secondaryRowHeight]`) behaves more like a minimum height, rather than a height. More on that [here](https://github.com/SAP/fundamental-ngx/issues/12475). Lastly, some more updates to the virtual scrolling docs stressing the importance that each row must be of the same height for virtual scrolling to behave smoothly.

### DIFF
--- a/libs/docs/platform/table/child-docs/scrolling/table-scrolling-docs.component.html
+++ b/libs/docs/platform/table/child-docs/scrolling/table-scrolling-docs.component.html
@@ -32,8 +32,8 @@
     <li><code>[virtualScroll]="true"</code> - enables virtual scroll</li>
     <li><code>[bodyHeight]="value"</code> - sets the height of the table body (viewport)</li>
     <li>
-        <code>[rowHeight]="value"</code> - sets the height of the table row if it is different from default for the
-        current content density
+        <code>[rowHeight]="value"</code> - sets the minimum height of the table row if it is different from default for
+        the current content density
     </li>
 </ul>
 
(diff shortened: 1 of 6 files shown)


dpavlenishvili (approved):
looks good.

PR #12476, opened 2024-09-26
feat(core): update Notifications to latest design

InnaAtanasova:
## Related Issue(s)
closes none

## Description

bring latest fundamental-styles for Notifications
NOTE: missing keyboard interaction for the Notification List. To be implemented soon

BREAKING CHANGE:
update in the markup for Notification and Notification Group


droshev (approved):

libs/core/notification/notification-group/notification-group.component.ts
```diff
@@ -1,68 +1,65 @@
...
-        public readonly elementRef: ElementRef,
-        @Optional() @Inject(FD_POPOVER_COMPONENT) private _popover: BasePopoverClass
-    ) {
-        if (this._popover) {
-            this._popover.focusTrapped = true;
-            this._popover.focusAutoCapture = true;
-        }
-    }
+    readonly ariaDescription = computed(
+        () => `Notification group ${this.groupHeader()?.expanded() ? 'expanded' : 'collapsed'}`
```
mikerodonnell89, line 45: another in need of translation

mikerodonnell89 (changes requested):
Looks good, couple of comments on things that need to be prepared for translations. Looks like most of the directives/components leave it to the application developer to handle all the logic so there is not much to test here. But it is a lot of breaking changes so I hope we don't have too many devs using the previous implementation

dpavlenishvili (approved):
everything looks good, just need localization changes as Mike mentioned and thats it. Just in case, i approve it.

Issue #11877, opened 2024-05-03
Table header is not sticky when used with Dynamic page


InnaAtanasova:
#### Is this a bug, enhancement, or feature request?
bug, reported in Slack
#### Describe your proposal.

When table is in the content of the dynamic page, the header of the table is not sticky and hides behind the dynamic page header

https://github.com/SAP/fundamental-ngx/assets/39598672/f54b8a42-817c-42aa-94d5-4120b31f148a

dreamweiver:
@InnaAtanasova , do we have a reproducible version on stackblitz ? i can look into this bug

dreamweiver:
i think i got it, when i changed `bodyheight` from fixed height to `tableBodyHeight`, the table overflows  and vertical scroll bar is shown and as mentioned above the header does not get fixed.

i will try to look into this :)

karolgromacki:
@droshev The issue still occurs
here is link to stackblitz:
https://stackblitz.com/edit/sjxljz-u5wgpm

dreamweiver:
@karolgromacki , the basic idea behind the fdp-table  components sticky header logic is based on the
1.  bodyHeight defined on the table component or
2. the default height(which is set to 100%)

in your case, the second option would apply but since you have enclosed it within `fd-dynamic-page-content` and it doesn't have a default height defined over it, which means the child element `fdp-table` which has height set to `100%`, won't have any effect.

so conclusion, you can just add a `height:100%` to `fd-dynamic-page-content` and everything will work like a charm 😄 

![Screenshot 2024-09-30 at 14 33 29](https://github.com/user-attachments/assets/05a4fb4e-b445-4396-ad30-88350afe5348)

dreamweiver:
@droshev , this is not an issue from `fdp-table` component, but rather  `fd-dynamic-page-content` component didn't have a height defined on it and due to this, the child elements couldn't use relative sizing 

do you think we. can add `height:100%` to `fd-dynamic-page-content` component ? if yes, then i can create a PR for this

dreamweiver:
> @droshev , this is not an issue from `fdp-table` component, but rather the issue on `fd-dynamic-page-content` component didn't have a height defined on it and due to this, the child elements couldn't use relative sizing
> 
> do you think we, can add `height:100%` to `fd-dynamic-page-content` component ? if yes, then I can create a PR for this

looking at this issue, i think we need to have a PR for this 😄 

ill create one soon

dreamweiver:
[...]

PR #12489, opened 2024-09-30
fix(core): default height missing for dynamic page content component

dreamweiver:
## Related Issue(s)

closes #https://github.com/SAP/fundamental-ngx/issues/11877

## Description

Dynamic Page Content component doesn't have a default height associated with it and hence the child elements which depend on relative sizing like "fdp-table" component is not able to compute the dynamic height for the sticky headers on the `fdp-table`

here is an example of the stackblitz with the fix(explicitly setting height 100% on `<fd-dynamic-page__content`>`

## Screenshots

Before the fix,

https://github.com/user-attachments/assets/f8bb3896-a9ef-419c-8c55-ed98632019bb

After setting the default height on the `fd-dynamic-page__content` component,

https://github.com/user-attachments/assets/472cea2c-f6bd-459a-8f2f-bee1bc80599c

##### During Implementation

1. Visual Testing:

-   [x] Text Truncation

-   [x] API boundary values

-   [x] poor examples

4. Accessibility testing

##### PR Quality

-   [x] update `README.md`


mikerodonnell89 (dismissed):
I understand the issue is that we need to get the header fixed, but we don't want to be seeing the table scrolling up above the header. I'm not sure if there is a css only fix for this problem, may require some js. But this PR fixes one problem and introduces another

Screenshot 2024-10-01 at 3 10 24 PM

dpavlenishvili:
> I understand the issue is that we need to get the header fixed, but we don't want to be seeing the table scrolling up above the header. I'm not sure if there is a css only fix for this problem, may require some js. But this PR fixes one problem and introduces another
> 
> Screenshot 2024-10-01 at 3 10 24 PM

i think same, if header should be sticky then it should be aligned in absolute top of the container.

dpavlenishvili (dismissed):
see comment related on sticky header issue

dreamweiver:
> I understand the issue is that we need to get the header fixed, but we don't want to be seeing the table scrolling up above the header. I'm not sure if there is a css only fix for this problem, may require some js. But this PR fixes one problem and introduces another
> 
> Screenshot 2024-10-01 at 3 10 24 PM

That is because the `fd-dynamic-page-content` component which contains the `fdp-table` as child element has a padding at the top (16px) and this is a free space for the `fdp-table` to occupy and hence the scrolling content of the `fdp-table` is visible.

Screenshot 2024-10-06 at 14 14 49

so I don't think it is a problem with the `fdp-table` component's sticky header implementation, but rather an issue with `fd-dynamic-page-content` and how it renders its child element.
if we remove the top padding, then you won't see any table content upon scrolling

dreamweiver:
@dpavlenishvili  @mikerodonnell89 , there is a way to fix this issue via CSS(remove top padding on `fd-dynamic-page-content` and add margin-top for `fdp-table`) 
But i don think its a good idea to do this on the component level, because all these components are suppose to be generic, we shouldn't add logic or code specific to other components behaviour.

I believe this should be handled by the consuming application,by introducing custom CSS styles, something like this https://stackblitz.com/edit/sjxljz-xxe577?file=src%2Fstyles.scss

let me know what you guys think about this.

mikerodonnell89:
> But i don think its a good idea to do this on the component level, because all these components are suppose to be generic, we shouldn't add logic or code specific to other components behaviour.

We do try to practice this, but sometimes need to make exceptions to the rule, and i think this would be an ok example of when to do that. I don't think we want to add a change in to the library that breaks the visuals (see my earlier comment), and then additionally require that the dev needs to add some custom css on their end. So I think for this specific feature, it's best to add a custom class that bends the separation-of-concerns rule a bit.

dreamweiver:
> https://stackblitz.com/edit/sjxljz-xxe577?file=src%2Fstyles.scss
 @mikerodonnell89, as agreed I have added the custom CSS styles to support an FDp-table with a sticky header inside a dynamic page content component.
 
 could you review it when you are free :)

dreamweiver:
> > But i don think its a good idea to do this on the component level, because all these components are suppose to be generic, we shouldn't add logic or code specific to other components behaviour.
> 
> We do try to practice this, but sometimes need to make exceptions to the rule, and i think this would be an ok example of when to do that. I don't think we want to add a change in to the library that breaks the visuals (see my earlier comment), and then additionally require that the dev needs to add some custom css on their end. So I think for this specific feature, it's best to add a custom class that bends the separation-of-concerns rule a bit.

@mikerodonnell89, as agreed I have added the custom CSS styles to support an FDp-table with a sticky header inside a dynamic page content component.

could you review it when you are free :)

mikerodonnell89 (approved):
Looks good!

dpavlenishvili (approved):

dreamweiver:
@droshev could you merge the PR, because i don't have the necessary 
permission

PR #12491, opened 2024-09-30
fix(core): form label white space

### LINKED ISSUE #12490: Colon attribute in Core Input Label produces whitespace before the colon

Bug.

#### Describe your proposal.

Using the `colon` attribute in an Input Label, produces a colon which has a whitespace before and after and it should only have whitespace after the colon.

No.

All of them.

Use a label like this: `<label fd-form-label colon="true">My label</label>`

https://stackblitz.com/edit/wfcaop?file=src%2Fapp%2Finput-example.component.html

Yes.

Yes (https://github.com/SAP/fundamental-ngx/issues?q=is%3Aissue+colon).

droshev:
fixes #12490

### DIFF
--- a/libs/core/form/form-label/form-label.component.html
+++ b/libs/core/form/form-label/form-label.component.html
@@ -17,11 +17,8 @@
 @if (inlineHelpPlacement === 'before' && inlineHelpContent) {
     <ng-template [ngTemplateOutlet]="inlineHelpRef"></ng-template>
 }
-<span class="fd-form-label" [class.fd-form-label--required]="required">
+<span class="fd-form-label" [class.fd-form-label--required]="required" [class.fd-form-label--colon]="colon">
     <ng-content></ng-content>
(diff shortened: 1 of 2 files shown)


khotcholava (approved):

littleamigo:
before this fix we've seen an unwanted artifical horizontal gap between the label and the colon and now the gap is completely vanished and it still doesn't look right after this fix:
image
I also see the bad rendering in official netlify example at https://docs-0-53--fundamental-ngx.netlify.app/#/core/input#basic

droshev:
@littleamigo can you check with one of the following versions: `0.50.6`, `0.51.6`, or `0.52.4`?

littleamigo:
@droshev yes, I checked with 0.50.6 as we are on Angular 17, and it looks better but not completely right:
image

Apparently, the colon is still displaced in combination with the required red asterisk.

PR #12496, opened 2024-10-01
fix(core): datepicker addon button should not be tabindex, popover opens with f4 key

### LINKED ISSUE #11282: [Sourcing-2][Accessibility: ACC-264.1][WIP] Icon: Toggle read twice.
Bug

#### Describe your proposal.
Component having issue:https://sap.github.io/fundamental-ngx/#/core/datetime-picker
Datetime Picker: 
When user access the icon 'Display calendar toggle' , the word toggle is mentioned twice.
The word toggle should not be read and UI should for the icon should be changed to button.

No

All latest version

Using Screen reader can check by implementing similar case. Just use Datetime picker and access the icon 'Display calendar toggle' , the word toggle is mentioned twice.

Yes

Please refer attachement:

[ACC-264.1.docx](https://github.com/SAP/fundamental-ngx/files/14001050/ACC-264.1.docx)

mikerodonnell89:
fixes #11282


libs/core/date-picker/date-picker.component.html
```diff
@@ -37,7 +37,7 @@
         [disabled]="disabled"
         [isControl]="true"
         [button]="true"
-        [buttonFocusable]="buttonFocusable"
+        [buttonFocusable]="false"
```
dpavlenishvili, line 40: i think you can set buttonFocusable variable as false and use it instead of set buttonFocusable input directly as false.

libs/core/datetime-picker/datetime-picker.component.html
```diff
@@ -33,7 +33,7 @@
         [disabled]="disabled"
         [isControl]="true"
         [button]="true"
-        [buttonFocusable]="buttonFocusable"
+        [buttonFocusable]="false"
```
dpavlenishvili, line 36: same thing here

dpavlenishvili (changes requested):
everything looks good, i left two similar comment that could be my misunderstanding.

libs/core/date-picker/date-picker.component.html
mikerodonnell89: Yes marking it as deprecated does signal our plan to remove it, but I do like your idea that we should keep the variable here, just make false the default value, so anyone who does still expect the button to be focusable will see that we have deprecated it, but its removal won't be such a surprise

PR #12498, opened 2024-10-02
fix(core,platform): radio a11y improvements

### LINKED ISSUE #11281: [Sourcing-2][Accessibility: ACC-264.1][WIP] Radio Button: Label and group label not read.
Bug

#### Describe your proposal.
Component having issue:https://sap.github.io/fundamental-ngx/#/core/radio
radio button:
The radio buttons should always be associated to its individual and group label.

No

All latest version

Using Screen reader can check by implementing similar case. Just use any Radio Button. Radio buttons will not be associated to its label both label  and also its group label header by screen reader

Yes

mikerodonnell89:
fixes #11281 

Various improvements to radio button accessibility, including some changes to examples. I think some of the screenshots from the linked issue are dated, so this PR addresses things a little differently


libs/platform/table/components/table-view-settings-dialog/grouping/grouping.component.html
```diff
@@ -17,9 +17,15 @@ <h4 fd-title [headerSize]="4">{{ 'platformTable.groupDialogHeader' | fdTranslate
     <fd-dialog-body disablePaddings>
         <ul fd-list [selection]="true">
             <li fd-list-group-header>
-                <span fd-list-title>{{ 'platformTable.groupDialogGroupOrderHeader' | fdTranslate }}</span>
+                <span fd-list-title id="group-dialog-group-order-header">{{
```
InnaAtanasova, line 20: in my opinion, this id should be auto generated and then assigned as aria-describedby to the list items (also automatically).

libs/docs/core/list-byline/examples/list-byline-selection-example/list-byline-selection-example.component.html
```diff
@@ -11,26 +11,44 @@ <h4>With Checkboxes</h4>
...
         </div>
     </li>
     <li fd-list-item [selected]="selectionValue === 'radio3'">
-        <fd-radio-button name="radio1" id="radio3" value="radio3" [(ngModel)]="selectionValue"></fd-radio-button>
+        <fd-radio-button
+            name="radio1"
+            id="radio3"
+            value="radio3"
+            [(ngModel)]="selectionValue"
+            aria-describedby="with-radio-buttons"
```
InnaAtanasova, line 50: I think it's more appropriate in this case to have `aria-labelledby` then `aria-describedby`
Screenshot 2024-10-03 at 12 30 19 PM

libs/platform/table/components/table-view-settings-dialog/grouping/grouping.component.html
```diff
@@ -43,9 +49,15 @@ <h4 fd-title [headerSize]="4">{{ 'platformTable.groupDialogHeader' | fdTranslate
 
         <ul fd-list [selection]="true">
             <li fd-list-group-header>
-                <span fd-list-title>{{ 'platformTable.groupDialogGroupByHeader' | fdTranslate }}</span>
+                <span fd-list-title id="group-dialog-group-by-header">{{
```
InnaAtanasova, line 52: I think this should also be `[attr.id]="groupOrderHeaderId"`

libs/platform/table/components/table-view-settings-dialog/sorting/sorting.component.html
```diff
@@ -16,9 +16,15 @@ <h4 fd-title [headerSize]="4">{{ 'platformTable.sortDialogHeader' | fdTranslate
     <fd-dialog-body disablePaddings>
         <ul fd-list [selection]="true">
             <li fd-list-group-header>
-                <span fd-list-title>{{ 'platformTable.sortDialogSortOrderHeader' | fdTranslate }}</span>
+                <span fd-list-title id="sort-dialog-sort-order-header">{{
```
InnaAtanasova, line 19: This is still hard-coded

libs/platform/table/components/table-view-settings-dialog/sorting/sorting.component.html
```diff
@@ -16,9 +16,15 @@ <h4 fd-title [headerSize]="4">{{ 'platformTable.sortDialogHeader' | fdTranslate
...
+                    'platformTable.sortDialogSortOrderHeader' | fdTranslate
+                }}</span>
             </li>
-            <li fd-list-item [selected]="direction === SORT_DIRECTION.ASC">
+            <li
+                fd-list-item
+                [selected]="direction === SORT_DIRECTION.ASC"
+                aria-describedby="sort-dialog-sort-order-header"
```
InnaAtanasova, line 26: same as above

InnaAtanasova (changes requested):

libs/platform/table/components/table-view-settings-dialog/grouping/grouping.component.html
mikerodonnell89: good catch

InnaAtanasova (approved):

PR #12499, opened 2024-10-02
docs(core): textarea and input state form message a11y improvements

### LINKED ISSUE #11288: [Sourcing-2][Accessibility: ACC-264.4][WIP] Error message not conveyed.
Bug

#### Describe your proposal.
Component having issue: https://sap.github.io/fundamental-ngx/#/core/textarea and https://sap.github.io/fundamental-ngx/#/core/input
Textarea State: Error and Input State: Error
On Error , We get error message alert only on clicking on iput/textarea field. error message should conveyed with clicking also. Due to this reason error message isn't readable by screen reader

No

All latest version

Using Screen reader can check by implementing similar case. Just use input or textarea with error state. Screen reader will not able to read any error message without click on input field

Yes

mikerodonnell89:
fixes #11288

### DIFF
--- a/libs/docs/core/input/examples/input-state-example.component.html
+++ b/libs/docs/core/input/examples/input-state-example.component.html
@@ -7,7 +7,7 @@
             id="input-52"
             placeholder="Field placeholder text"
             state="success"
-            aria-labelledby="input-state-form-message-52"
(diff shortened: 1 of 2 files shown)


InnaAtanasova (approved):
this PR only changes doc files and should be `docs(core): ...`

PR #12505, opened 2024-10-06
fix(platform): cache user selection after searching for new user list for approval flow

CSEHoangV:
closes https://github.com/SAP/fundamental-ngx/issues/12093
ng-15 https://github.com/SAP/fundamental-ngx/pull/12107

### DIFF
--- a/libs/platform/list/list.component.ts
+++ b/libs/platform/list/list.component.ts
@@ -69,6 +69,10 @@ export class SelectionChangeEvent {
     selectedItems: BaseListItem[];
     /** Index */
     index: number;
+    /** Selected item */
+    added?: BaseListItem;
+    /** Deselected item */
+    removed?: BaseListItem;
 }
 
 let nextListId = 0;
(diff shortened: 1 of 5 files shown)


libs/platform/list/list.component.ts
```diff
@@ -513,10 +517,16 @@ export class ListComponent<T>
 
         this._selectionModel = new SelectionModel<BaseListItem>(this._multiSelect, this.selectedItems);
 
-        this._selectionModel.changed.pipe(takeUntilDestroyed(this._destroyed)).subscribe(() => {
+        this._selectionModel.changed.pipe(takeUntilDestroyed(this._destroyed)).subscribe((selection) => {
             this.selectedItems = this._selectionModel.selected;
             const event = new SelectionChangeEvent();
             event.selectedItems = this.selectedItems;
+            if (selection.added?.length) {
+                event.added = selection.added[0];
```
mikerodonnell89, line 525: So, this change does fix the issue we are seeing for the approval flow component. However, I think this code could pose a problem in the future for the list component, agnostic to the approval flow. It is interesting that `added` on a selectionChangeEvent is always just one `BaseListItem` but selectionModel emits an array of them.  Which makes me think, instead of doing `event.added = selection.added[0];` we need to either:
- Iterate over `selection.added` and emit a `selectedItemChange` event for each `selection.item`
- Or, update the `selectedItemChange` event to be able to pass either a single item (as it is now), or emit an array of them

same idea goes for `selection.removed` as well

mikerodonnell89 (changes requested):
comment above

droshev:
@CSEHoangV what do we do with this PR? Are you going to update and prepare for merging? Or we close it?

CSEHoangV:
> @CSEHoangV what do we do with this PR? Are you going to update and prepare for merging? Or we close it?

I will spend some time working on it later this month.

PR #12523, opened 2024-10-07
fix(core): form label required * should be aria-hidden

### LINKED ISSUE #11284: [Sourcing-2][Accessibility: ACC-264.1] Required information read as star.
Bug

#### Describe your proposal.
Component having issue: https://sap.github.io/fundamental-ngx/#/core/input
Required information for all mandatory input field is read out as ‘star’. Information as required for all the mandatory UIs should be communicated to the user

No

All latest version

Using Screen reader can check by implementing similar case. Just use any input mandatory field. Required information for all mandatory input field is read out as ‘star’.

Yes

mikerodonnell89:
fixes #11284

### DIFF
--- a/libs/core/form/form-label/form-label.component.html
+++ b/libs/core/form/form-label/form-label.component.html
@@ -17,7 +17,12 @@
 @if (inlineHelpPlacement === 'before' && inlineHelpContent) {
     <ng-template [ngTemplateOutlet]="inlineHelpRef"></ng-template>
 }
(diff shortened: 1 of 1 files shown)


InnaAtanasova (changes requested):
Screenshot 2024-10-08 at 10 22 47 AM

Have you tested the output with SR? `aria-hidden` will remove the content and the children from the a11y tree and the label shouldn't be read.

mikerodonnell89:
Closing, fix will come in styles

Issue #12369, opened 2024-09-05
[BUG] fd-avatar-group focus, overflow bugs


egorwow70:
#### Is this a bug, enhancement, or feature request?

Couple of bugs connected to fd-avatar-group component from fundamental-ngx/core library.

Control - https://docs-0-49--fundamental-ngx.netlify.app/#/core/avatar-group.

#### Describe your proposal.

I think it should be fixed on the library side.

#### Can you handle that on the application side

No.

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)

- "@fundamental-ngx/core": "0.49.0",
- "@sap-theming/theming-base-content": "11.13.0",
- "fundamental-styles": "0.34.0",
- "@angular/core": "17.3.0",

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;

**a) Bug with focus:**

1. Open stackblitz example - https://stackblitz.com/edit/uy2qbh?file=src%2Fapp%2Fdynamic-page-tabs-example.component.html
2. Open default tabs example.
3. When there is more than one avatar group control on the page, firstly click on the first one, you should see a blue border, that highlights a focus state.
4. After click on the second control, focus will still remain on the first one.
![image](https://github.com/user-attachments/assets/f7d645c1-d59f-47e6-88c0-125653f96ad1)

**b) Bug with switching tabs and avatar group control rendering:**

1. Open stackblitz example - https://stackblitz.com/edit/uy2qbh?file=src%2Fapp%2Fdynamic-page-tabs-example.component.html
2. Open default tabs example.
3. Select the second tab.
4. Go back to the first tab.
5. You see only "+n more button", but you should see the whole avatar group row with images as well.
![image](https://github.com/user-attachments/assets/3a49a44f-bcce-41a9-b6c4-651dc938232f)

**c) Not a bug, but a feature request**

Is it possible to hide a standard avatar group item details popover header? But then we need access to navigate between group and item views
![image](https://github.com/user-attachments/assets/457849a5-0d14-4963-8240-83b90e636397)

I found the way how to custom the body of the popover with <ng-template #details>, but didn't find the way how to hide this header
![image](https://github.com/user-attachments/assets/7e29993e-6fd8-4bb3-98d8-54031226db0b)

**d) Not a bug, but a feature request**
 Is it possible to open group popover from the center of the avatar group?
 
 Like this:
 
[...]

PR #12530, opened 2024-10-10
fix(core): address Avatar Group issues

khotcholava:
fix(core): Resolve focus retention & rendering issues, add feature enhancements

Closes [#12369](https://github.com/SAP/fundamental-ngx/issues/12369)

## Description
- Fixed focus bug where focus remained on the first avatar group control after interacting with the second one.
- Fixed rendering issue where only "+n more" button displayed after switching tabs instead of the full avatar group row.
- Added feature to open group popover from the center of the avatar group.
- Added configuration option to make "+n more" button circular to match avatars typically being in circle shapes.
- Introduced property to set the number of visible avatars before displaying the "+n more" button.


libs/docs/core/avatar-group/avatar-group-docs.component.html
```diff
@@ -1,14 +1,14 @@
...
-    <p>
-        In this example, which is called <code>individual</code> type avatar group, which means that every item is
-        displayed separately, every item can be focused separately and details template will be displayed for that item.
-    </p>
-</description>
-<component-example>
-    <fd-avatar-group-default-example></fd-avatar-group-default-example>
-</component-example>
-<code-example [exampleFiles]="avatarGroupDefaultExample"></code-example>
+<!--<fd-docs-section-title id="default" componentName="AvatarGroup"> Avatar Group Standard Usage </fd-docs-section-title>-->
```
mikerodonnell89, line 1: Accidental commit?

mikerodonnell89 (changes requested):
Looks good but what is going on with all of the commented out html?

mikerodonnell89 (changes requested):
In the "group type standard usage" examples, after opening the popover, i can not navigate through the avatars with the arrow keys. Because each avatar in the popover is clickable, we do need this functionality with keyboard control as well

khotcholava:
@mikerodonnell89  In this [example](https://sap.github.io/ui5-webcomponents/components/AvatarGroup/#grouping), the web components avatar group features tab navigation but lacks arrow navigation.

I'm currently working on implementing tab navigation.

khotcholava:
@mikerodonnell89 Implemented arrow navigation.

mikerodonnell89 (changes requested):
- For the "Group Type" examples, I don't think each avatar should be focusable/tabbable. It's just supposed to be the one focus around the whole group. The user shouldn't be able to do this:
https://github.com/user-attachments/assets/b5cc768f-c88f-4164-9fa1-798331accd90

mikerodonnell89 (changes requested):
If I open the group type with the space bar, I'd expect the first avatar in the popover to receive focus. It doesn't, and also pressing tab still goes to focus each individual avatar in the non-popover part of the component:

https://github.com/user-attachments/assets/59721035-1f2e-4b9f-960b-5b4ed92020ff

libs/core/avatar-group/components/default-avatar-group-overflow-body/default-avatar-group-overflow-body.component.ts
```diff
@@ -73,6 +74,16 @@ export class DefaultAvatarGroupOverflowBodyComponent implements AfterViewInit, O
     get isRtl(): boolean {
         return !!this._rtlService?.rtl.value;
     }
+
+    /**
+     * List of avatars to be rendered in the overflow popover.
+     **/
+    avatars = input<AvatarGroupItemRendererDirective[]>([]);
+
+    /**
```
dpavlenishvili, line 83: missing declaration or wrong place for this jsDoc.

droshev:
@mikerodonnell89 @dpavlenishvili Can you review it again?

droshev:
@dpavlenishvili any update?

mikerodonnell89 (changes requested):
Does not look like the first issue (regarding the focus staying on the first avatar group after interacting with a second) was fixed. In this example I've used the mouse to click both, then the escape key to close the second group

https://github.com/user-attachments/assets/63ca5cf0-de58-4073-a94a-a4bf99f31221

droshev:
@khotcholava Can you check the e2e?

libs/core/avatar-group/avatar-group.component.ts
```diff
@@ -88,6 +94,10 @@ export class AvatarGroupComponent implements AvatarGroupHostConfig {
     @Input()
     size: AvatarGroupHostConfig['size'] = 'l';
 
+    /** Popover placement */
+    @Input()
+    popoverPlacement: Placement | null = null;
```
dpavlenishvili, line 99: use Nullable type.

libs/core/avatar-group/components/avatar-group-host.component.ts
dpavlenishvili, line 1: add or update relevant unit tests for maxVisibleItems

libs/core/popover/popover-service/popover.service.ts
```diff
@@ -127,6 +131,13 @@ export class PopoverService extends BasePopoverClass {
                 this._removeOverlay(this._modalBodyClass, this._modalTriggerClass);
             }
         });
+
+        // Add keydown listener for space key
+        this._renderer.listen('document', 'keydown', (event: KeyboardEvent) => {
```
dpavlenishvili, line 136: Listening for keydown events globally on the document level can lead some performance issues, if it is possible scope the event listener to the popover element directly. 
for example:
`
const popoverElement = this._elementRef.nativeElement;
this._renderer.listen(popoverElement, 'keydown', (event: KeyboardEvent) => {
    if (event.code === 'Space' && this.isOpen) {
        this._storeAndFocusFirstTabbable();
    }
});
`

libs/core/avatar-group/components/avatar-group-host.component.ts
khotcholava: There is not e2e test for avatar group host

libs/core/avatar-group/avatar-group.component.ts
khotcholava: PopoverPlacement type is Placement | null. It's not Nullable<Placement>. If I change it I'm afraid it will broke already existing functionality

droshev:
@khotcholava needs to update with the comments from UX, check e2e/unit and make it as PR. Needs to be reviewed afterwards

droshev:
@khotcholava needs to update with the comments from UX, check e2e/unit and make it as PR. Needs to be reviewed afterwards

droshev:
@khotcholava Can you summarize what you did and what needs to be implemented?

mikerodonnell89:
Closing as this is quite dated - we will need to readdress #12369

Issue #10729, opened 2023-10-16
Fundamental Platform Table: can't hide border for platform table 


snowqu:
#### Is this a bug, enhancement, or feature request?
Bug
#### Describe your proposal.
should have option to hide all the borders.

#### Can you handle that on the application side
No.

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
Angular: 15.2.9
Fundamental: 0.43.11

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
It's a bug

1.https://sap.github.io/fundamental-ngx/#/platform/table/basic
2.try playground example
3.enable noHorizontalBorders, noVerticalBorders, noBorders, noOuterBorders, noBodyBorders
4.There is still a vertical border before the last column, and there is a horizontal border under header. 
It shouldn't display any border now. 

<img width="2529" alt="image" src="https://github.com/SAP/fundamental-ngx/assets/6847549/b74e05e1-5e4f-4bfb-831f-33af9a57ea39">

#### Please provide relevant source code (if applicable).
it could reproduce from official playground example. 

#### Please provide stackblitz example(s).
it could reproduce from official playground example. 

#### Did you check the documentation and the API?
Yes

#### Did you search for similar issues?
Yes

#### Is there anything else we should know?
No

droshev:
@kalo2711 Can you try to reproduce the issue?

kalo2711:
The issue is still there, the last vertical and first horizontal borders still appear in this configuration.

PR #12555, opened 2024-10-16
fix(platform): platform table border fix

khotcholava:
fix(platform): table ensure all borders are hidden when border options are enabled

closes [#10729](https://github.com/SAP/fundamental-ngx/issues/10729)

## Description

This fix addresses the issue where vertical and horizontal borders are still displayed in the Platform Table component, despite enabling the `noHorizontalBorders`, `noVerticalBorders`, `noBorders`, `noOuterBorders`, and `noBodyBorders` options. Specifically:

- Removed the vertical border before the last column.
- Removed the horizontal border below the table header when border-hiding options are applied.

### Before 
image

### After 
image

### DIFF
--- a/libs/docs/platform/table/examples/platform-table-default-example.component.html
+++ b/libs/docs/platform/table/examples/platform-table-default-example.component.html
@@ -1,15 +1,4 @@
-<!-- @TODO redo the changes-->
-<fdp-table
-    [noBorders]="true"
-    [noBodyBorders]="true"
-    [noHorizontalBorders]="true"
-    [noVerticalBorders]="true"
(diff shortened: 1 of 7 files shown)


mikerodonnell89 (commented):
What about these ones?

Screenshot 2024-10-16 at 10 22 20 AM

mikerodonnell89:
> Regarding the header toolbar, its bottom border is part of the toolbar design. Would you like us to handle that on our end as well?

Nope, I think if that is coming from the toolbar then no class on the table should affect it, so this is all looking good

mikerodonnell89 (approved):

PR #12581, opened 2024-10-17
fix(platform): icon-tab-bar _generateTabBarItems function improvement

### LINKED ISSUE #12474: Icon Tab Bar overflow bug
bug

#### Describe your proposal.
Icon Tab Bar overflow should not be there if there is no tabs inside:

Screenshot 2024-09-26 at 10 55 33

### LINKED ISSUE #12554: [BUG] More button can't work as expected in fdp-icon-tab-bar
Angular version: 17.3.12
**Fundamental version: 0.50.3 & 0.49.2**

After More button is clicked, some items are not in the list, and sometimes, even the popover is not visible.
Please see the screenshot below, there are 10 items in all, but only 2 items display in the list. 
image

And I found that because other items are hidden. Why are they hidden?
image

Please fix from your side. And all the versions should be covered.

Note: It's an urgent bug from customer. Please take high priority to this issue.

You can easily reproduce it in stackblitz which is mentioned in the doc: https://docs-0-50--fundamental-ngx.netlify.app/#/platform/icon-tab-bar

dpavlenishvili:
## Related Issue(s)

closes #12474
closes #12554

### DIFF
--- a/libs/platform/icon-tab-bar/icon-tab-bar.component.ts
+++ b/libs/platform/icon-tab-bar/icon-tab-bar.component.ts
@@ -353,13 +353,11 @@ export class IconTabBarComponent implements OnInit, TabList {
         flatIndexRef = flatIndexRef || { value: 0 };
         return config.map((item, index) => {
             const uId = `${indexPrefix}${index}`;
-            if (!('color' in item)) {
-                item.color = 'default';
-            }
+            item.color = item.color || 'default';
             const result: IconTabBarItem = {
                 ...item,
                 index,
-                cssClasses: [],
+                cssClasses: item.color ? [`fd-icon-tab-bar__item--${item.color}`] : [],
                 uId,
                 hidden: false,
                 parentUId,
@@ -371,9 +369,6 @@ export class IconTabBarComponent implements OnInit, TabList {
                     uId
                 )
             };
-            if (item.color) {
-                result.cssClasses = [`fd-icon-tab-bar__item--${item.color}`];
-            }
             return result;
         });
     }


libs/platform/icon-tab-bar/icon-tab-bar.component.ts
```diff
@@ -353,13 +353,11 @@ export class IconTabBarComponent implements OnInit, TabList {
         flatIndexRef = flatIndexRef || { value: 0 };
         return config.map((item, index) => {
             const uId = `${indexPrefix}${index}`;
-            if (!('color' in item)) {
-                item.color = 'default';
-            }
+            item.color = item.color || 'default';
```
khotcholava, line 356: What if you store `item.color || 'default'` in a variable instead of modifying the value directly?
For example: `const color = item.color || 'default'`
You can then use this variable in cssClasses.

dpavlenishvili: Mutating item.color directly ensures that the item consistently has a valid color property in all cases inside return. By updating item.color early, it prevents potential future issues when item.color is referenced elsewhere in the result object or other parts of the code. Essentially, it is necessary for item.color to always have a valid value, but there is no better way to set it by default without similar mutation.

khotcholava (approved):
LGTM

mikerodonnell89 (commented):
I'm a bit confused about how the code changes in this PR fix the linked issues, but, it appears to do the trick. @dpavlenishvili care to give an explanation? I'm lost as to why the `--hidden` class is no longer applied

InnaAtanasova (approved):

PR #12596, opened 2024-10-17
fix(core) Add Support for Multi-line Messages in Select Component Dropdown

khotcholava:
fix(core): Add Support for Multi-line Messages in Select Component Dropdown

closes [#11430](https://github.com/SAP/fundamental-ngx/issues/11430)

## Description
- Refactored the select component to manage text display mode.
- Adjusted padding in the select component message to eliminate unnecessary space.
- Introduced a new input property, textOverflow, to control how text is displayed in the dropdown list, allowing for either ellipsis or full text.
***

### After 

Screenshot 2024-10-18 at 02 39 57

### DIFF
--- a/libs/core/select/select.component.html
+++ b/libs/core/select/select.component.html
@@ -96,8 +96,11 @@
                 tabindex="-1"
                 role="presentation"
                 [type]="state || 'default'"
-                [innerHtml]="stateMessage"
-            ></li>
+            >
+                <span fd-list-title>
+                    {{ stateMessage }}
+                </span>
+            </li>
         }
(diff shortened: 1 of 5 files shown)


droshev (changes requested):
the PR title is not well written and the PR fields are not set

mikerodonnell89 (approved):

LukasHeimann:
Is it intentional that this change introduces additional whitespace when there is no value state text at all? https://github.com/SAP/fundamental-ngx/issues/12810

droshev:
there should not be any extra space

Issue #12518, opened 2024-10-07
[BUG] fd-combobox - dynamic groupFn change causes NG0956


egorwow70:
#### Is this a bug, enhancement, or feature request?

BUG

#### Can you handle that on the application side

No

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)

- "@angular/core": "18.0.3",
- "fundamental-styles": "0.38.0",
- "@sap-theming/theming-base-content": "11.18.0",
- "@fundamental-ngx/cdk": "0.53.0",
- "@fundamental-ngx/platform": "0.53.0",
- "@fundamental-ngx/core": "0.53.0",

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;

We are using dynamic groupFn, because the data in the combobox should be grouped dynamically, cause in the app a user can change data and it can affect the grouping inside the combobox.

1. Open stackblitz example - https://stackblitz.com/edit/nytpv8?file=package.json
2. Open console
3. Search in console by "tracking"
4. Favorite/unfavorite fruits
5. You should see such warnings in console
![image](https://github.com/user-attachments/assets/6ec520d9-76d2-4e13-ad3c-2dcd51333383)

Expected result: no warnings and errors in console

NG0956: The configured tracking expression (track by identity) caused re-creation of the entire collection of size 2. This is an expensive operation requiring destruction and subsequent creation of DOM nodes, directives, components etc. Please review the "track expression" and make sure that it uniquely identifies items in a collection. Find more at https://angular.dev/errors/NG0956

We are worried, that the angular says **This is an expensive operation requiring destruction and subsequent creation of DOM nodes, directives, components etc**

#### Please provide stackblitz example(s).

Stackblitz - https://stackblitz.com/edit/nytpv8?file=package.json

#### Did you check the documentation and the API?

Yes

#### Did you search for similar issues?

Yes

PR #12612, opened 2024-10-18
fix(core): combobox dynamic groupFn

khotcholava:
fix(core): combobox dynamic groupFn

closes [#12518](https://github.com/SAP/fundamental-ngx/issues/12518)

## Description

Fixed dynamic function re-creation issue.

### Before 

[old.webm](https://github.com/user-attachments/assets/ed2eacfa-e867-43e4-8e6c-f69216a2aaa4)

***

### After

[new.webm](https://github.com/user-attachments/assets/422ed22d-d6b0-4f85-bd91-85dfa6d648cd)

### DIFF
--- a/libs/core/combobox/combobox.component.html
+++ b/libs/core/combobox/combobox.component.html
@@ -132,11 +132,11 @@
     >
         <ng-content></ng-content>
         @if (groupFn) {
-            @for (group of displayedValues | listGroupPipe: groupFn; track group) {
+            @for (group of displayedValues | listGroupPipe: groupFn; track group.key) {
                 <li role="group" fd-list-group-header [tabindex]="0">
                     <span fd-list-title>{{ group.key }}</span>
                 </li>
-                @for (term of group.value; track term) {
+                @for (term of group.value; track term.name) {
                     <li
                         role="option"
                         fd-list-item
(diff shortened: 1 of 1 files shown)


libs/core/combobox/combobox.component.html
```diff
@@ -133,11 +133,11 @@
     >
         <ng-content></ng-content>
         @if (groupFn) {
-            @for (group of displayedValues | listGroupPipe: groupFn; track group) {
+            @for (group of displayedValues | listGroupPipe: groupFn; track group.key) {
                 <li role="group" fd-list-group-header [tabindex]="0">
                     <span fd-list-title>{{ group.key }}</span>
                 </li>
-                @for (term of group.value; track term) {
+                @for (term of group.value; track term.name) {
```
mikerodonnell89, line 140: This only works with the given stackblitz example because the items in the example have a `name` property. If the example code were instead
```
export type ComboboxItem = { testName: string; type: string };

    dropdownValues: ComboboxItem[] = [
        { testName: 'Apple', type: 'Fruits' },
        { testName: 'Banana', type: 'Fruits' },
        { testName: 'Pineapple', type: 'Fruits' }
```

it will still throw the warning

mikerodonnell89 (changes requested):

libs/core/combobox/combobox.component.html
khotcholava: There's no need to add tracking to this, as you're right—we don’t know what type the combobox can accept. We could use indexes, but I’m not sure if that’s the best approach. Without the tracking function, it works as expected, and we’re not seeing any warning messages. @mikerodonnell89

mikerodonnell89 (approved):

Issue #11783, opened 2024-04-19
[Feature Request] Infobar showing applied filters in platform table


do-oel:
#### Is this a bug, enhancement, or feature request?
Feature Request
#### Describe your proposal.
According to the Fiori Guidelines, filtered tables display an info bar (toolbar in active mode) below the table toolbar to indicate the set filter. "Beneath the toolbar, display a filter infobar (which itself is a special toolbar) if the responsive table is filtered." (https://experience.sap.com/fiori-design-web/responsive-table/#components)

The infobar should use the following component from Fundamental NGX: <fd-toolbar fdType="info" [active]="true"> with
Text: "Filtered By: Column Name (Filter Parameters)", e.g., "Filtered By: Size (XS, M, L)"
Icon: "decline"

The infobar should only appear once a filter is set and disappear once the filter is reset.

Example wire:
![image](https://github.com/SAP/fundamental-ngx/assets/143730992/58f87239-eaf2-4e76-a3de-46ec2082ad23)
#### Can you handle that on the application side
No
#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
All, current version (v0.49), if possible downport to v0.48
#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
N/A
#### Please provide relevant source code (if applicable).
N/A
#### Please provide stackblitz example(s).
N/A
#### In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations.
N/A
#### Did you check the documentation and the API?
Yes
#### Did you search for similar issues?
Yes
#### Is there anything else we should know?
N/A

rebecagm:
Hi @droshev and @dpavlenishvili Could we prioritise this feature in the upcoming sprints? With our product release approaching, we want to ensure the table filters function correctly to enhance the user experience.

droshev:
@rebecagm We are doing our best. We will keep you posted.

droshev:
@khotcholava there are 2 issues reported in the [PR](https://github.com/SAP/fundamental-ngx/pull/12634#issuecomment-2461634464): 
Thanks for this fix! Highly appreciated. 

Two smaller things for improvement:
1. The filter info bar shows an index number in front of the filter term. This should be hidden (e.g, Filtered by: StatusColor (**0:** positive)
2. The category name is using no spaces, this seems a bit weird (e.g, **StatusColor** instead of Status Color)


[...]

PR #12634, opened 2024-10-22
fix(platform): Update info bar table to reflect selected filters

khotcholava:
fix(platform): table showing info bar with selected filters

closes [#11783](https://github.com/SAP/fundamental-ngx/issues/11783)

- Implemented an info bar beneath the table toolbar to display active filters, following SAP Fiori design recommendations.
- Used <fd-toolbar fdType="info" [active]="true"> from Fundamental NGX.
- Displayed filter information with the correct format
- Included a "decline" icon in the info bar.
- Info bar appears only when a filter is applied and hides once the filter is reset.
- Added feature to close the popover when the dismiss button is clicked. The popover will reappear upon changing the filters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.53.1-rc.16](https://github.com/SAP/fundamental-ngx/compare/v0.53.1-rc.15...v0.53.1-rc.16) (2024-10-18)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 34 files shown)


khotcholava:
NOTE: 

**!!! Not yet ready for review !!!**

mikerodonnell89 (changes requested):
Two things in this screenshot:

- [object Object]
- I think we need the compact button

Screenshot 2024-10-24 at 12 58 24 PM

InnaAtanasova:
Still shows [object Object]
Screenshot 2024-10-28 at 9 45 38 AM

mikerodonnell89 (approved):

do-oel:
Thanks for this fix! Highly appreciated. 

Two smaller things for improvement:
1. The filter info bar shows an index number in front of the filter term. This should be hidden (e.g, Filtered by: StatusColor (**0:** positive)
2. The category name is using no spaces, this seems a bit weird (e.g, **StatusColor** instead of Status Color)


image

Issue #12577, opened 2024-10-17
[BUG] Link in checkbox lable is not clickable


akatyabg:
The links in a checkbox component cannot be opened (is not clickable).
The issue can be seen also in the documentation: https://sap.github.io/fundamental-ngx/#/core/checkbox#custom-label

To Reproduce

Go to [checkbox with custom label example](https://sap.github.io/fundamental-ngx/#/core/checkbox#custom-label)
Click on the link
The link doesn't open

droshev:
@khotcholava Can you check this issue as it is important for one team? thanks

PR #12636, opened 2024-10-23
fix(core): Checkbox link in  label  is not clickable

khotcholava:
fix(core): Checkbox link in  label  is not clickable

closes [#12577](https://github.com/SAP/fundamental-ngx/issues/12577)

## Description
Updated the checkbox component to ensure that links within custom labels are clickable and function as intended. This resolves the issue observed in the documentation example.

### Before
[Screencast from 2024-10-23 16-34-31.webm](https://github.com/user-attachments/assets/449361b0-1f49-444d-aba0-bcbed78f6c98)

### After
[new.webm](https://github.com/user-attachments/assets/b82613d0-a3d6-44e3-a98e-8ab89934aaa0)

### DIFF
--- a/libs/core/checkbox/checkbox/checkbox.component.scss
+++ b/libs/core/checkbox/checkbox/checkbox.component.scss
@@ -7,6 +7,10 @@ fd-checkbox {
 
     .fd-checkbox__label {
         user-select: none;
+
+        .fd-checkbox__label-container {
+            pointer-events: auto;
+        }
     }
 
     &.fd-checkbox--standalone {


libs/core/checkbox/checkbox/checkbox.component.scss
```diff
@@ -7,6 +7,10 @@ fd-checkbox {
 
     .fd-checkbox__label {
         user-select: none;
+
+        .fd-checkbox__label-container {
+            pointer-events: auto;
+        }
```
InnaAtanasova, line 13: you can use:
```
 &-container {
            pointer-events: auto;
}
```

InnaAtanasova: Also, you need to bring this fix to fundamental-styles

InnaAtanasova (approved):

mikerodonnell89 (approved):

droshev:
@khotcholava e2e of checkbox are failing. Can you check them?

mikerodonnell89:
> @khotcholava e2e of checkbox are failing. Can you check them?

this PR only adds some styling to an existing class, I think the failing test must not come from this PR, and we're just seeing the error maybe from the CI only looking to test affected components?

dpavlenishvili (approved):

droshev:
@khotcholava Can you follow up?

Issue #11004, opened 2023-11-24
[enhancement] fdp table -  make responsive pop-in areas navigatable 


lpgarzonr:
#### enhancement
Given I have a fdp-table with rowNavigatable
And some columns use the fdpTableColumnResponsive directive
And I resize the table so that some columns are moved to the pop-in mode
When I click on the pop-in areas
Then I should be able to navigate to what was specified for the table `rowNavigate`
But now I can navigate only if I over and click on the columns that do not have the `fdpTableColumnResponsive` 

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
0.47.4-rc.6, [Table Responsive Columns](https://fundamental-ngx.netlify.app/#/platform/table/basic#responsive-columns)
#### Please provide stackblitz example(s).

https://stackblitz.com/edit/nlvqwb?file=src%2Fapp%2Fplatform-table-responsive-columns-example.component.html

mikerodonnell89:
I think we need to refactor the whole approach here, we are using an additional `<tr>` for the pop-ins which is confusing to screenreader users, who are told there are double the amount of rows

droshev:
@kalo2711 Can you try to reproduce the issue with the latest version of fundamental-ngx?

kalo2711:
The area isn't clickable in pop-in mode, but the rest seems to be working good.

PR #12673, opened 2024-10-30
feat(platform): fdp table - make responsive pop-in areas navigable

khotcholava:
feat(platform): fdp table - make responsive pop-in areas navigable

closes [#11004](https://github.com/SAP/fundamental-ngx/issues/11004)

## Description 
- Ensure row navigation works correctly when interacting with pop-in columns using `fdpTableColumnResponsive`.
- Malfunction code: `showItemsCount` input was duplicated.


libs/platform/table/table.component.html
```diff
@@ -18,7 +18,7 @@
         }
     </div>
     @if (showGrowingButton) {
-        <fdp-table-growing-button [showItemsCount]="showItemsCount" [showItemsCount]="true"></fdp-table-growing-button>
```
khotcholava, line 21: Removed a duplicate `showItemsCount` input. One instance was hardcoded to `true`, while the other took its value from the component input. Now, `showItemsCount` consistently derives its value from the component input.
@mikerodonnell89 @InnaAtanasova

libs/platform/table/table.component.scss
```diff
@@ -492,6 +492,7 @@ fdk-dynamic-portal {
     &__body-table {
         &.fd-table--pop-in {
             .fd-table__row--secondary {
+                cursor: pointer;
```
khotcholava, line 495: Made the secondary row clickable and added a pointer cursor to indicate interactivity.
 @mikerodonnell89 @InnaAtanasova

mikerodonnell89: I think it also needs the active/hover state as well and should be applied to the entire cell

mikerodonnell89 (dismissed):

droshev:
@khotcholava to follow up

libs/platform/table/table.component.scss
```diff
@@ -648,3 +653,7 @@ fdk-dynamic-portal {
 th.fd-table__cell .fd-table__inner {
     font-weight: 700;
 }
+
+.tr-hovered {
```
dpavlenishvili, line 657: inappropriate naming, it should be something like fd-classNameExample, or it should be styled directly in template

mikerodonnell89:
Couple things:

- Border here is now missing
Screenshot 2024-12-05 at 11 14 12 AM

- When an item is selected and then i hover the description area, hover state is only applied to the title area. You can see the title area has a grey background and the description area has a blue background in the screenshot above
- Description header text ("Description (popping < 900px)") is still receiving text selection cursor, implying clicking the area is not actionable

droshev:
@dpavlenishvili can you review it? thanks

droshev:
@mikerodonnell89 Can you take it from here. According to Akaki the work is done.

mikerodonnell89 (approved):

Issue #12674, opened 2024-10-30
Accessibility issue in wizard - The section in wizard content does not have a role attribute 


MonishaK28:
![image](https://github.com/user-attachments/assets/e7b25e4a-3075-448f-a15f-257e421f7784)

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.) 
Latest

#### Is there anything else we should know?

**_IMPORTANT: Please refrain from providing links or screenshots of SAP's internal information, as this project is open-source, and its contents are accessible to anyone._**

PR #12734, opened 2024-11-18
fix(core): wizard section add role attribute

khotcholava:
fix(core): wizard section add role attribute

closes [#12674](https://github.com/SAP/fundamental-ngx/issues/12674)

- Added `role="tabpanel"` role to wizard section

Screenshot 2024-11-18 180855

### DIFF
--- a/libs/core/wizard/wizard-content/wizard-content.component.html
+++ b/libs/core/wizard/wizard-content/wizard-content.component.html
@@ -1,5 +1,6 @@
 <ng-template #contentTemplate>
     <section
+        role="tabpanel"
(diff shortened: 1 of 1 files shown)


mikerodonnell89 (changes requested):
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tabpanel_role states that tabpanels must be associated with a tab, that's part of a tablist. The steps in ours have role list/listitem, same as ui5. But ui5 is using `role="region"` for the sections, maybe we go with that?

dpavlenishvili (approved):

Issue #12641, opened 2024-10-25
UXC: Settings (Generator) design updates


droshev:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Describe your proposal.
Update Settings Generator to the latest design. (bring the missing styles in Fundamental Styles)

Timeline:
- PR opened **Nov 11**
- PR merged **Nov 14**

droshev:
closing as it is done with the latest spec

PR #12739, opened 2024-11-20
fix(platform): update settings generator docs and component to update theme/language automatically, add readonly property to form generator

mikerodonnell89:
This is part of #12641, and not intended to be an entire update to the new design. We'll do this incrementally

### DIFF
--- a/libs/core/input-group/input-group.component.html
+++ b/libs/core/input-group/input-group.component.html
@@ -81,5 +81,6 @@
         [attr.placeholder]="_cvaControl.cvaDirective?.placeholder || null"
         [attr.aria-required]="required"
         (search)="_onSearchEvent($event)"
+        autocomplete="off"
     />
 </ng-template>
--- a/libs/core/input-group/input-group.component.html
+++ b/libs/core/input-group/input-group.component.html
@@ -81,6 +81,5 @@
         [attr.placeholder]="_cvaControl.cvaDirective?.placeholder || null"
         [attr.aria-required]="required"
         (search)="_onSearchEvent($event)"
-        autocomplete="off"
     />
 </ng-template>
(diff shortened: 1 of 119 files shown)


droshev:
@dpavlenishvili @khotcholava Can you review the PR?

libs/core/input-group/input-group.component.html
```diff
@@ -81,5 +81,6 @@
         [attr.placeholder]="_cvaControl.cvaDirective?.placeholder || null"
         [attr.aria-required]="required"
         (search)="_onSearchEvent($event)"
+        autocomplete="off"
```
khotcholava, line 84: Is it necessary for the input group component to always have autocomplete set to "off"? @mikerodonnell89

khotcholava (changes requested):

libs/core/input-group/input-group.component.html
mikerodonnell89: You're right, probably best to just do this specifically for combobox/multi-input etc

libs/platform/form/form-generator/base-dynamic-form-generator-control.ts
```diff
@@ -65,4 +65,13 @@ export abstract class BaseDynamicFormGeneratorControl<T extends BaseDynamicFormF
         }
         return (this.formItem?.useMessageAsPlaceholder && this.formItem?.message) || '';
     }
+
+    /** @hidden */
+    get readonly(): boolean {
```
dpavlenishvili, line 70: you can simplify it:

`
get readonly(): boolean {
    return this.formItem?.readonly ?? false;
}
`

libs/platform/settings-generator/layouts/settings-generator-sidebar-layout/settings-generator-sidebar-layout.component.html
```diff
@@ -5,50 +5,72 @@
...
-                                    width="40%"
-                                    textLines="1"
-                                    [style.margin]="'auto 0'"
-                                ></fd-skeleton>
-                            </div>
+    <div class="fdp-settings-generator__search-header">
+        <h3 fd-title class="fdp-settings-generator__title">
+            {{ 'platformSettingsGenerator.settingsLabel' | fdTranslate }}
+        </h3>
+        <!-- TODO: implement search input
```
khotcholava, line 12: Do we need commented code here? @mikerodonnell89

droshev (approved):

PR #12771, opened 2024-11-25
fix(platform): table filter enhancement

khotcholava:
feat(core):
closes [#11783](https://github.com/SAP/fundamental-ngx/issues/11783)

- Add support for formatting array values in table component
- Fix format field name

## Before

Screenshot 2024-11-25 at 17 49 05

## After

Screenshot 2024-11-25 at 17 49 53


khotcholava:
@mikerodonnell89 @dpavlenishvili Ping

InnaAtanasova:
Please use proper naming for the PRs. If this gets merged without the proper title we would need to manually fix the release notes

mikerodonnell89 (changes requested):
Two things I noticed:

- There's a space before the comma
- no "Status Color" text, just "(critical)"

Screenshot 2024-12-05 at 10 49 43 AM

karolgromacki:
Hi @khotcholava , I've noticed one problem - so far the info bar is displayed after changing the filter through UI or from code, but I have a case when I load the whole tableState from the session storage and it doesn't apply the filter info bar - it seems that the problem is related to
`this._tableService.stateChange$.subscribe(({ type, state }) => {
                switch (type) {`

because when we use `this.table.setTableState()`, the state is changed, but the type is undefined and 
 `this._setAppliedFilterNames(state.current);` is not executed.

khotcholava:
@karolgromacki Please include the valid code example. It's easier for me to debug.

droshev:
@khotcholava fix the linting issues, make sure e2e/unit are not failing. Then we should review it.

PR #12808, opened 2024-12-03
fix(core): enable segmented button

### LINKED ISSUE #12675: Segmented-button Accessibility when switching between disabled/enabled

bug

#### Describe your proposal.

When the segmented-button becomes disabled, it is not possible to focus on it anymore - this is expected
But when the button becomes enabled, it is still not possible to focus on it.

We are using reactive forms, but the same happened with template driven

We couldn't find a way to hack it. Is there a component method or service we could call to refresh things up?

ng 17 + fd 48 
ng 18 + fd 53

 There are 2 ways to reproduce:

1. Component is initialized as disabled and becomes enabled - focus not possible
2. Component is initialized as enabled, becomes disabled and enabled again -> focus not possible

the stackblitz crashes when I try to open it. It is simple to reproduce by just taking the last example in the docs with reactive forms and do:

```typescript
setTimeout(() => {
  this. customForm.controls['basic'].disable()
  this. customForm.controls['basic'].enable()
}, 2000)
```

yes

yes

droshev:
fixes #12675

fix enable segmented button
delete e2e as the unit tests cover the testing enough


mikerodonnell89 (changes requested):
This does allow the segmented button to be focused again, however I can no longer use the arrow keys to change the selected segment. Still working for the other examples, I'm just seeing this problem on the newly modified example

InnaAtanasova:
> This does allow the segmented button to be focused again, however I can no longer use the arrow keys to change the selected segment. Still working for the other examples, I'm just seeing this problem on the newly modified example

I can confirm the same thing. The arrow keys are not working. Also, I would suggest an improvement for the example. Add a button that programatically disables and enables the Segmented button. Additionally, explain in a 2-3 short sentences what this example is doing. Right now, with the short timeout, for people that don't have an idea what the issue is, it's not very clear why we have this example.

droshev:
> so, I would suggest an improvement for the example. Add a button that programatically disables and enables the Segmented button. Additionally, e

fixed!

mikerodonnell89:
> > so, I would suggest an improvement for the example. Add a button that programatically disables and enables the Segmented button. Additionally, e
> 
> fixed!

I am not seeing an update regarding Inna's documentation suggestions

Also, while I am seeing the arrow key focus problem is fixed, I can no longer use enter/space to change the selection on this example

alexhristov14:
like @mikerodonnell89 said, enter/space does not change the selection in the reactive form. Also, clicking it with a mouse only focuses it but doesn't change the value/selection.

InnaAtanasova (changes requested):
Complex example doesn't work as expected. 
To reproduce the problem:
- click on the first button. It gets selected state, the value is updated to `value: first`
- click on the second button. The value is updated to `value: first,second`, but the segmented button doesn't show selection.
Screenshot 2025-02-05 at 9 55 21 AM

libs/core/segmented-button/segmented-button.component.spec.ts
```diff
@@ -23,17 +22,10 @@ const isSelectedClass = 'fd-button--toggled';
...
-
-    @ViewChild('third', { read: ElementRef })
-    thirdButton: ElementRef;
-
-    @ViewChild(SegmentedButtonComponent)
-    segmentedButton: SegmentedButtonComponent;
+    @ViewChild('first', { read: ElementRef }) firstButton: ElementRef;
+    @ViewChild('second', { read: ButtonComponent }) secondButton: ButtonComponent;
+    @ViewChild('third', { read: ElementRef }) thirdButton: ElementRef;
+    @ViewChild(SegmentedButtonComponent) segmentedButton: SegmentedButtonComponent;
```
InnaAtanasova, line 28: these can be signal based

InnaAtanasova (changes requested):
https://github.com/user-attachments/assets/a6fdcf5c-08a4-4890-acc1-76cb515fa82c

There's hover state on the selected button and using arrow keys focus the entire segmented button.

mikerodonnell89 (approved):

Issue #12793, opened 2024-11-29
Enhancement: Change growing button state when all items are loaded


do-oel:
#### Is this a bug, enhancement, or feature request?
Enhancement
#### Describe your proposal.
For platform tables with growing button: Currently, when all items in a table are loaded, the growing button still remains active and clickable although no more items can be loaded.

In this case, the growing button should be replaced by a non-interactive but focusable text stating "Showing n of n"

Current state:
![image](https://github.com/user-attachments/assets/4d426fe3-c50a-45d7-ac5a-a012ddf39096)

Proposed changes:
![image](https://github.com/user-attachments/assets/a4b9fa6b-b2d8-470f-aa5e-115b52563f80)

#### Can you handle that on the application side
No

PR #12817, opened 2024-12-06
feat(platform): table growing button state

khotcholava:
closes [#12793](https://github.com/SAP/fundamental-ngx/issues/12793)

Add conditional rendering to growing button component Implemented logic to display the growing button only if there are more items to load, using a conditional check. This ensures that the button is hidden once all items have been loaded.

## Snapshots

https://github.com/user-attachments/assets/b3eb215c-9185-47af-a33b-cbfaeafa8f96


droshev:
@dpavlenishvili Can you review it.

droshev:
@mikerodonnell89 Can you take over? Akaki said he did the fix.

alexhristov14 (dismissed):
There should be an else statement where showing "n of n" is replacing the load more button as per issue

mikerodonnell89 (approved):

PR #12857, opened 2024-12-19
fix(platform): match filter strategies for string data type columns

mohammad-nmeri:
## Description

This PR addresses an issue where filter strategies for string columns were incorrectly suggesting strategies for both string and number data types. Previously, when a column was of type string, the displayed filter strategies included both string-related and number-related options.

With this fix, the filter strategies for string columns now show only those relevant to string data types, ensuring more accurate and appropriate suggestions for users when filtering string-based columns.

### Before:
before

### After:
after


mohammad-nmeri:
selected wrong branch, i will reopen new one

PR #12859, opened 2024-12-19
feat(filtering): add dropdown menu for predefined filter options

mohammad-nmeri:
## Description
This PR enhances the filtering dialog in the Platform App by adding a dropdown for selecting predefined filter options. It introduces a new filterValuesOptions attribute in the columns component, which expects a string array of predefined filter values to be sent when defining a column. These values will be displayed in the filter dropdown, improving the filtering experience. Additionally, a new "Array" data type for columns is implemented to support this functionality.

## Screenshots
main

### Before:
before

### After:
after

### DIFF
--- a/libs/docs/platform/table/examples/advanced-scrolling/advanced-scrolling-example.component.html
+++ b/libs/docs/platform/table/examples/advanced-scrolling/advanced-scrolling-example.component.html
@@ -17,14 +17,16 @@
 >
     <fdp-table-toolbar title="Order Line Items" [hideItemCount]="false" [hideSearchInput]="true"> </fdp-table-toolbar>
 
+<!--    Example of Array Filter-->
     <fdp-column
         name="name"
         key="name"
         label="Name"
         align="start"
+        [filterValuesOptions]="options_demo"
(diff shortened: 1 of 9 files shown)


mikerodonnell89 (changes requested):
@mohammad-nmeri Sorry, but before we can review, commit messages must adhere to these rules for release log generation purposes: https://github.com/SAP/fundamental-ngx/blob/main/CONTRIBUTING.md#commit

mohammad-nmeri:
@mikerodonnell89, apologies for the oversight. I have updated the commit message

mikerodonnell89 (changes requested):
@mohammad-nmeri It looks like maybe you did not push your changes, as I'm seeing your commit message is still missing the prefix. It should be looking like your PR title. Also, I think your formatting would not pass our pre-commit lint rules. I think you may have somehow bypassed our pre-commit hooks. If you run `yarn install` on the repository again, and soft reset your changes and commit again, it should run the checks

mikerodonnell89:
Reopening with a new PR (not forked)

PR #12860, opened 2024-12-19
fix(platform): match filter strategies for string data type columns

mohammad-nmeri:
## Description

Matched the strategy filter to display only strategies relevant to the string data type when the column is of type string. Previously, the filter dialog was showing strategies for both number and string data types for columns with a string data type. This update ensures that only the appropriate strategies for string data types are shown when the column is of type string.

### Before:
before

### After:
after


mikerodonnell89:
We had previously discussed this and because greater than/less than/etc are considered valid operations when applied to strings in native JS, we decided to keep it. @droshev have I missed some update here?

mikerodonnell89:
Closing, as I've confirmed with the team that we have discussed this previously and decided to keep all operators for strings as well.

PR #12861, opened 2024-12-19
chore: update to eslint 9 and the flat configuration

robertIsaac:
## Description

currently the repo is using the old `.eslintrc` and the unsupport eslint 8 versions
this PR migrate to the new flat configuration and eslint 9

I also updated nx since the old version was buggy, you may want to do the upgrade in a separate branch first which is understandable

note that I didn't add all the old plugins, this can be done in a future PR, or if mandatory I can work on it in the future


robertIsaac:
closed in favor of https://github.com/SAP/fundamental-ngx/pull/12877

PR #12877, opened 2024-12-23
chore: update to eslint 9 and the flat configuration

robertIsaac:
## Description

currently the repo is using the old `.eslintrc` and the unsupport eslint 8 versions
this PR migrate to the new flat configuration and eslint 9

I also updated nx since the old version was buggy, you may want to do the upgrade in a separate branch first which is understandable

note that I didn't add all the old plugins, this can be done in a future PR, or if mandatory I can work on it in the future


mikerodonnell89 (changes requested):
@robertIsaac this appears to introduce some problems with committing. For example, i just opened [this file](https://github.com/SAP/fundamental-ngx/blob/main/libs/cdk/utils/directives/auto-complete/auto-complete.directive.ts) and entered a tab before a variable definition, to test the formatting. On commit, I see the error: `42:5  error  Definition for rule '@angular-eslint/no-output-on-prefix' was not found  @angular-eslint/no-output-on-prefix`

robertIsaac:
@mikerodonnell89 
this is due to how the new flat config works it reads from the directory it's executed, not from the closest config file to the file

to illustrate the problem try these command
```shell
eslint libs\platform\form\auto-complete\auto-complete.directive.ts # will fail
cd .\libs\platform\
eslint form\auto-complete\auto-complete.directive.ts # will pass
```

the problem is that husky is triggering eslint directly which will be using the root-level eslint.config.js file not the `libs\platform\eslint.config.js` one

while `nx lint platform` will use the correct file

now how to fix this problem? to be honest I'm not sure
I need to research it

robertIsaac:
I used the flag `unstable_config_lookup_from_file` to fix it, it looks like the only way now
read more about it here https://eslint.org/blog/2024/10/eslint-v9.12.0-released/#experimental-configuration-file-resolution

mikerodonnell89 (approved):

PR #12915, opened 2025-01-10
feat(core): created a new legend section

### LINKED ISSUE #12403: display calendar legend

bug

#### Describe your proposal.

Hello, fundamental-ngx team. We are implementing the app using angular and fundamental-ngx/core and have a requirement to display a date picker with special dates and legend displayed. As I can see, special dates can be displayed with current library functionality, but the legend is missing. **Is it possible to fix it on the library side?**

_**Legend should be supported by both fd-calendar and fd-date-picker from fundamental-ngx/core**_

As for the technical side, I think it will be easy for devs, who are using the lib, to configure that via those objects (SpecialDayRule), for example via new field - title or label or legendLabel (naming up to you):

image

I am not a designer, but this is an example what we need:
image

Sapui5 example - https://sapui5.hana.ondemand.com/#/entity/sap.ui.unified.Calendar/sample/sap.ui.unified.sample.CalendarSpecialDaysLegend

No.

- "@fundamental-ngx/core": "0.49.0",
- "@sap-theming/theming-base-content": "11.13.0",
- "fundamental-styles": "0.34.0",
- "@angular/core": "17.3.0",

Yes.

Yes

alexhristov14:
closes #12403


libs/core/calendar/calendar-legend/calendar-legend.component.ts
```diff
@@ -0,0 +1,93 @@
...
+        private elementRef: ElementRef,
+        private viewContainer: ViewContainerRef,
+        private focusingService: CalendarLegendFocusingService
+    ) {
+        this.calIndex = this.focusingService.getCalIndex() - 1;
+    }
+
+    /** @hidden */
+    ngOnInit(): void {
+        this._addCalendarLegend();
```
mikerodonnell89, line 57: This logic being placed in the init lifecycle method will lead to problems if the special days are updated after the calendar is created. You could watch for changes to the specialDaysRules. But I think a better way of handling this would be to avoid typescript-side logic wherever possible, instead opting to render as much as possible with double curly brace interpolation in `calendar.component.html`

mikerodonnell89 (changes requested):
This is looking good, but I left a few comments on things to consider. Also this PR seems to just add this to calendar, I think to consider the original issue completed, we'll need to add this to datepicker/datetimepicker as well. Though I would be in favor of considering that a separate task, I prefer several small pr's over one large one when possible.

PR #12957, opened 2025-01-21
fix(core): popover escape key to close should not bubble to dialog

### LINKED ISSUE #12605: [Accessibility: ACC-270.7] List box: Dialog closes on closing the list box
Bug

#### Describe your proposal.
On closing the list box with ‘esc key’ the list box must only get closed and not the dialog.

No

All

Component used: [Dialog](https://sap.github.io/fundamental-ngx/#/core/dialog) with [Select](https://sap.github.io/fundamental-ngx/#/core/select)
On closing the list box ‘Application type’ with esc key, it is observed that the dialog ‘Add Application’ is also getting closed automatically.
image

You can try ngx examples, issue is reproducible there.

Yes, issue is raised by an accessibility expert.

Yes

Yes

mikerodonnell89:
fixes #12605

The entire popover should have a keydown escape handler to close the popover body. Also, multi-input/multi-combobox/combobox all had their own escape key handlers for closing their popovers, resulting in the popover service close function being called twice due to the popover now handling this itself


alexhristov14:
The popover works well but I've noticed that the multi-input and the multi-combobox now act differently on escape. After navigating in the multi-combobox items with the arrows, you have to press **escape** twice before closing it completely. Was that intentional?

InnaAtanasova (approved):

PR #13050, opened 2025-02-17
chore: migrate all control flow

robertIsaac:
- remove all usage of CommonModule
- force the new control flow using eslint rule


droshev:
the core and platform unit tests are failing

robertIsaac:
I fixed them
had to revert one CommonModule import, no idea why it fails when I remove it, it's unused

droshev (approved):

PR #12955, opened 2025-01-21
feat(core): lazy load popover body

robertIsaac:
## Description

we have weird behavior with web components when the popover body is eagerly loaded that it break other popover bodies (only one popover body should be rendered from a web component at one point)
we are using these two components inside our web components, and thus we need their popover body to be lazily loaded
 
### Before:

image

### After:

image

##### During Implementation

1. Visual Testing:

-   [x] visual misalignments/updates
-   [ ] check Light/Dark/HCB/HCW themes
-   [ ] RTL/LTR - proper rendering and labeling
-   [ ] responsiveness(resize)
-   [ ] Content Density (Cozy/Compact/(Condensed))
-   [ ] States - hover/disabled/focused/active/on click/selected/selected hover/press state
-   [ ] Interaction/Animation - open/close, expand/collapse, add/remove, check/uncheck
-   [ ] Mouse vs. Keyboard support
-   [ ] Text Truncation

2. API and functional correctness

-   [ ] check for console logs (warnings, errors)
-   [ ] API boundary values
-   [ ] different combinations of components - free style
-   [ ] change the API values during testing

3. Documentation and Example validations

-   [ ] missing API documentation or it is not understandable
-   [ ] poor examples
-   [ ] Stackblitz works for all examples

4. Accessibility testing
5. Browser Testing - Edge, Safari, Chrome, Firefox

##### PR Quality

-   [ ] the commit message(s) follows the guideline:
        https://github.com/SAP/fundamental-ngx/blob/main/CONTRIBUTING.md
-   [ ] tests for the changes that have been done
-   [ ] all items on the PR Review Checklist are addressed :
        https://github.com/SAP/fundamental-ngx/wiki/PR-Review-Checklist
-   [ ] Run npm run build-pack-library and test in external application
-   [ ] update `README.md`
-   [ ] [Breaking Changes Wiki](https://github.com/SAP/fundamental-ngx/wiki/Breaking-Changes)

PR #13090, opened 2025-03-06
feat: lazy load popover body in all cases for the avatar group

robertIsaac:
I missed two cases in https://github.com/SAP/fundamental-ngx/pull/12955 
and they are covered now

### DIFF
--- a/libs/core/avatar-group/avatar-group.component.html
+++ b/libs/core/avatar-group/avatar-group.component.html
@@ -43,12 +43,12 @@
                     [ngTemplateOutletContext]="{ $implicit: hiddenItems }"
                 ></ng-template>
             </fd-popover-control>
-            <fd-popover-body>
+            <ng-template fdPopoverBody>
                 <ng-template
                     [ngTemplateOutlet]="avatarGroupOverflowBodyTemplate"
                     [ngTemplateOutletContext]="{ $implicit: hiddenItems }"
                 ></ng-template>
-            </fd-popover-body>
+            </ng-template>
         </fd-popover>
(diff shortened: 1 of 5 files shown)


mikerodonnell89 (changes requested):
I'm not sure how, but it appears to be causing some problems with the "standard usage" example overflow:

https://github.com/user-attachments/assets/01d700d4-792f-469c-aa07-16cb4f3d97de

robertIsaac:
that's interesting, I will try to debug it tomorrow

robertIsaac:
image
this style was being loaded with the popover body which was affecting the fd-popver width
now that we lazy load it this style wasn't loaded till the popover is opened
the fix is to add this style to the group avatar itself

robertIsaac:
BTW this should fix another issue we are facing currently where there is a space for the avatars but they are collapsed and we weren't sure why

lpgarzonr:
Thanks @robertIsaac!.

mikerodonnell89:
Seeing this focus outline surrounding the whole page when clicking the group type example:
Screenshot 2025-03-10 at 10 47 52 AM

robertIsaac:
okay I will check it tomorrow

robertIsaac:
@mikerodonnell89 can you check it now

robertIsaac:
@mikerodonnell89 is it possible to backport it to 0.54?

droshev:
@robertIsaac done! `0.54.3` is [out](https://github.com/SAP/fundamental-ngx/releases/tag/v0.54.3)

robertIsaac:
Thank you

PR #13103, opened 2025-03-11
fix(core): wizard step are clickable even without permissions

### LINKED ISSUE #13102: Wizard requiredField can be overpass because of fd-wizard-progress-bar
**Is this a bug, enhancement, or feature request?**
Bug

**Describe your proposal.**
I would like to use a WIzard with some steps that are required, like the doc. 
However, if I fill the required field > go to the next step > go back by using the fd-wizard-progress-bar > delete data in required field 
I can still go to the next step with the fd-wizard-progress-bar  but it's not possible because required field is empty and the button next is disabled. 

**Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)**
Angular : 18.2.0
Fundamental : v0.54.1

**If this is a bug, please provide steps for reproducing it; the exact components you are using;**
Testable directly in documentation : Fundamental documentation > Wizard > Dialog Wizard
https://sap.github.io/fundamental-ngx/#/core/wizard#dialog
VIdeo for example

https://github.com/user-attachments/assets/5676a48f-d1af-4bc7-a29b-ae7cac639d6c

**Did you check the documentation and the API?**
Yes

**Did you search for similar issues?**
Yes

alexhristov14:
closes #13102


libs/docs/core/wizard/examples/wizard-visible-summary-example.component.ts
```diff
@@ -80,6 +80,14 @@ export class WizardVisibleSummaryExampleComponent {
         });
     }
 
+    validateStep = (): boolean | Promise<boolean> => {
```
mikerodonnell89, line 83: Why the promise?

mikerodonnell89 (changes requested):
This appears to be working as we'd like for the "next step" button, but not for the steps up at the top of the wizard:

https://github.com/user-attachments/assets/98a9e13b-5193-4b33-b77e-236447347272

mikerodonnell89 (approved):

Issue #13118, opened 2025-03-17
[P2][Sourcing] Japanese IME is not working in platform combobox on window


CSEHoangV:
**Is this a bug, enhancement, or feature request?**

Bug

**Describe your proposal.**

Following up from https://github.com/SAP/fundamental-ngx/issues/12983, Japanese IME is working fine on MAC. However, for window machine, the input conversion does not occur.

This is due to the input parameter from KeyboardEvent gives different information when user presses spacebar between MAC and window.

<img width="1264" alt="Image" src="https://github.com/user-attachments/assets/9a5835c1-5828-4b73-b4b3-aab1e38be45f" />

In Window, the key of "Space" when using IME will be "Process" and gives different keycode which is not currently handled in combobox keydown handler.

Relevant Information: https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#ime_and_composition_keys

**If this is a bug, please provide steps for reproducing it; the exact components you are using;**

On window,
Have Japanese IME enabled and choose Hiragana
Have platform combobox with list of some Japanese values
Type Romanji to convert it to Hiragana.

For example:

- Type ni(に)ho(ほ)nn(ん) and press spacebar on platform combobox.

Expected:

- It is converted to 日本 and 日本 is also highlighted from autocomplete after pressing enter.

This can be reproducible on fundamental playground when we use Japanese IME on window machine.

**Can you handle that on the application side**

No

**Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)**

Angular: 15
core/ platform: 0.43.51

**Did you check the documentation and the API?**
Yes

PR #13123, opened 2025-03-17
fix(platform): do not set focus on list when there is IME processing

CSEHoangV:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13118

## Description

When IME is used, on window, for spacebar or any keydown that IME detects that needs conversion, KeyboardEvent has properties of 'Process' as key and 229 as keyCode. On combobox keydown handler, this is not checked and will set focus on list which stops the processing input conversion.

When input method is processing, we should not interrupt the ime process and set focus on list in this case. Therefore, add check against it.

Relevant information: https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#ime_and_composition_keys

```
"Process" [3] | The Process key. Instructs the IME to process the conversion. | VK_PROCESSKEY (0xE5) |   |  
-- | -- | -- | -- | --
```
https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event#keydown_events_with_ime

### DIFF
--- a/libs/cdk/src/lib/utils/functions/key-util.spec.ts
+++ b/libs/cdk/src/lib/utils/functions/key-util.spec.ts
@@ -113,7 +113,7 @@ describe('KeyUtil', () => {
     describe('isKeyType', () => {
         interface TestValue {
             event: KeyboardEvent;
-            keyType: 'numeric' | 'alphabetical' | null | undefined;
+            keyType: 'numeric' | 'alphabetical' | 'ime' | null | undefined;
         }
 
         const positiveTestValues: TestValue[] = [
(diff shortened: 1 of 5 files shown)


mikerodonnell89 (changes requested):
Looks good for this specific component, but I am thinking, if possible, this would be better handled in the fdkAutoComplete directive so that it applies to any input with the autocomplete popover. If not possible there for wahtever reason, I think we'll also need this logic in all those other components with autocomplete popovers (multi-input, multi-combobox etc. on both core and platform)

CSEHoangV:
> Looks good for this specific component, but I am thinking, if possible, this would be better handled in the fdkAutoComplete directive so that it applies to any input with the autocomplete popover. If not possible there for wahtever reason, I think we'll also need this logic in all those other components with autocomplete popovers (multi-input, multi-combobox etc. on both core and platform)

I only add same implementation on multi-combobox (core and platform) as it has about similar focus check on keydown handler. 
For multi-input, they don't do focus when using spacebar and only allow certain keycode which is not 229 from ime then does not cause any issue during input conversion.
For core combobox, they only allow "control" type to perform focus which also only allows specific keycode and therefore will not cause issue.

mikerodonnell89 (approved):

PR #13133, opened 2025-03-23
chore: update to nx 20.6

robertIsaac:


package.json
```diff
@@ -155,16 +155,13 @@
...
         "typedoc-plugin-merge-modules": "5.1.0",
-        "typescript": "5.6.3",
+        "typescript": "5.8.2",
         "typescript-eslint": "^8.22.0",
         "wait-on": "6.0.1",
         "wdio-axe": "2.2.3",
         "wdio-chromedriver-service": "6.0.4",
         "webdriverio": "7.27.0"
     },
-    "resolutions": {
```
droshev, line 165: we need to keep that section, otherwise nx@npm gets multiple versions installed and CI fails. Maybe you just update it with 20.6.2

robertIsaac: but the pipeline has passed

droshev: it wasn't failing all the time
before either

robertIsaac: can we merge it like that and ping me if it fails again, I want to investigate it and find why

droshev (approved):

package.json
robertIsaac: I took a look at the yarn.lock file and saw the nx was mentioned twice
https://github.com/SAP/fundamental-ngx/blob/d27e32708c4784785ac74bd51d83db0f121a28b6/yarn.lock
at lines 19202 and 19268
but at this PR nx is mentioned only once, so there shouldn't be a problem now, I'm not sure what caused the other one to have it twice and it will be useless now to investigate
if this happened again in the main branch I will investigate and resolve it

robertIsaac: why people in angular 19.0 and 19.1 won't be able to use if we merged this PR
angular shouldn't have any breaking changes between minor versions, so a library built with 19.2 should be fully compatible with 19.0 and 19.1

robertIsaac: I created the PR and I wasn't sure if I could solve all the problems that's why I created it draft so I don't waste someone's time in reviewing it because there were multiple issues with typescript
I will mark it ready now

PR #13173, opened 2025-04-02
chore: remove unused lerna dependency 

robertIsaac:


droshev:
we use lerna for the versioning

droshev:
@robertIsaac I will check nx releases and hopefully migrate to using it. Thanks for pointing out, i will close this PR for now.

Issue #13181, opened 2025-04-04
More/Less button for Notifications content


YuliiaPanova-sap:
Is this a bug, enhancement, or feature request?
Feature request/enhancement

Describe your proposal.
The new component Notification doesn't have More/Less button for show/truncate long description. 
But the SAP's Standart has it as mandatory feature. 
Now description is truncated by default and no ways to see it fully.

Standart - [**UXC-024** ](https://wiki.one.int.sap/wiki/display/uxc/UXC-024)

![Image](https://github.com/user-attachments/assets/c4418b2b-7177-41ad-a36e-2e5b08e98157)

PR #13183, opened 2025-04-07
feat(core): add more and less expander for Notifications

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/13181

## Description
Long Title and Description in Notifications should truncate after 2 lines and an expander should be shown in the Footer (meta) with configurable text.

### DIFF
--- a/libs/core/notification/notification-footer/notification-footer.component.ts
+++ b/libs/core/notification/notification-footer/notification-footer.component.ts
@@ -1,13 +1,71 @@
-import { ChangeDetectionStrategy, Component, ViewEncapsulation } from '@angular/core';
+import { ChangeDetectionStrategy, Component, input, signal, ViewEncapsulation } from '@angular/core';
+import { resolveTranslationSignal } from '@fundamental-ngx/i18n';
+import { FD_NOTIFICATION_FOOTER } from '../token';
 
 @Component({
     selector: 'fd-notification-footer',
-    template: `<ng-content></ng-content>`,
+    template: `<ng-content></ng-content>
+        @if (showTrigger()) {
+            <a role="button" class="fd-link fd-notification__link" tabindex="0" (click)="onTriggerClick()">
+                <span class="fd-link__content">{{ triggerLabel }}</span>
+            </a>
+        } `,
     encapsulation: ViewEncapsulation.None,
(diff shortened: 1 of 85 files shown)


libs/core/notification/notification-footer/notification-footer.component.ts
```diff
@@ -1,13 +1,54 @@
...
+     * When `true`, the full title and description are shown;
+     * when `false`, they are truncated after the second line.
+     */
+    expanded = signal(false);
+
+    /**
+     * Custom label for the expand button.
+     * Defaults to `'More'`.
+     */
+    moreLabel = input('More');
```
mikerodonnell89, line 43: I think the built in default will need to be translated. This is now done a bit more easily with `resolveTranslationSignalFn()`

mikerodonnell89 (changes requested):

libs/docs/core/notification/examples/notification-options/notification-options-example.component.html
```diff
@@ -126,14 +138,17 @@ <h4>Notification with avatar, priority indicator, unread title and object status
                 <h4 fd-notification-title [unread]="true">
                     Notification Title (unread) lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
                     tempor incididunt ut labore et dolore magna aliqua. Lorem ipsum dolor sit amet, consectetur
-                    adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+                    adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Lorem ipsum
+                    dolor sit amet consectetur adipisicing elit. Inventore cupiditate expedita at quaerat voluptatum
+                    fugit dolore ut, similique obcaecati perspiciatis praesentium quos repellendus laboriosam quidem
+                    iure? Natus quia ex praesentium.
```
YuliiaPanova-sap, line 144: The request was to create a More/Less option for notification content (description), not for the title. 
Maybe it makes sense to create a tooltip for long and truncated titles.

mikerodonnell89 (approved):

PR #13185, opened 2025-04-08
[WIP] fix(platform): search field buttons must be focusable

### LINKED ISSUE #12748: Keyboard and VoiceOver support for search in platform table

bug

angular: 18.2.4
fundamental: 0.53.1-rc.23

Copy example from [documentation page](https://sap.github.io/fundamental-ngx/#/platform/table/basic).
Navigate only with keyboard or only with VoiceOver. Found issues:

1. missing outline on search focus
2. when navigating only with keyboard it is not possible to clear search input 
3. when navigating only with VoiceOver (MacOS 14.7) it is not possible to search based on provided text or clear search input

In can be also reproduced using only keyboard or voiceOver on [documentation page](https://sap.github.io/fundamental-ngx/#/platform/table/basic)

Yes, accessibility expert confirms this is an issue.

mikerodonnell89:
fixes #12748

### DIFF
--- a/libs/platform/search-field/search-field.component.html
+++ b/libs/platform/search-field/search-field.component.html
@@ -87,7 +87,6 @@
             @if (inputText && inputText.length > 0) {
                 <span class="fd-input-group__addon fd-input-group__addon--button {{ appearance?.addonClass }}">
                     <button
-                        tabindex="-1"
                         [attr.aria-hidden]="!mobile"
                         [attr.aria-label]="'platformSearchField.clearButtonTitle' | fdTranslate"
                         class="fdp-search-field__clear fd-button fd-button--transparent {{ appearance?.buttonClass }}"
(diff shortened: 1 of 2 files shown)


mikerodonnell89:
Putting this PR on hold - the current behavior may be by design, need to speak with UX

mikerodonnell89:
> Now you focus the search button but the focus is not visible to the user. 

It is visible, just immediately touching the border of the entire search field with no spacing: 
Screenshot 2025-04-11 at 11 29 49 AM

> I checked UI5 and Ui5 wc and the clear and search buttons are not focusable. The Search button is activated on Enter. On Tab the focus goes outside of the Search component.

I will speak to UX and clarify the intended behavior, because yes while I agree that clear and search are performable by the keyboard without the need to focus the buttons, it does break the concept of having every clickable button also be keyboard-friendly

mikerodonnell89:
Closing, see my comment here: https://github.com/SAP/fundamental-ngx/issues/12748#issuecomment-2851940048

PR #13192, opened 2025-04-09
fix(core,platform): add input for dynamic page header title heading level

### LINKED ISSUE #13182: [P2][Sourcing][Dynamic page header] heading of the page should start with <h1>
is this a bug, enhancement, or feature request?
Bug

Describe your proposal.
The main dynamic page header is implemented as "h2" whereas per w3c standards https://www.w3.org/WAI/tutorials/page-structure/headings/ , the page heading should start from h1

Image

No

ng: 15
core/platform: 0.43.51

https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/dynamic-page

Please provide relevant source code (if applicable).
https://ng-15-downport--fundamental-ngx.netlify.app/#/platform/dynamic-page

In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations.
NA
Reported by DTAG customer

Did you check the documentation and the API?
Did you search for similar issues?
Is there anything else we should know?
IMPORTANT: Please refrain from providing links or screenshots of SAP's internal information, as this project is open-source, and its contents are accessible to anyone.

mikerodonnell89:
fixes #13182

### DIFF
--- a/libs/core/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.ts
+++ b/libs/core/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.ts
@@ -68,7 +68,7 @@ export class DynamicPageHeaderComponent
 
     /** Heading level of the dynamic page title. */
     @Input()
-    headingLevel = 2;
+    headingLevel: Nullable<1 | 2 | 3 | 4 | 5 | 6> = 2;
 
     /**
      * @hidden
--- a/libs/core/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.ts
+++ b/libs/core/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.ts
@@ -66,6 +66,10 @@ export class DynamicPageHeaderComponent
     @Input()
     subtitleWrap = false;
 
+    /** Heading level of the dynamic page title. */
+    @Input()
+    headingLevel = 2;
+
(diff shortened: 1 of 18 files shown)


libs/core/src/lib/dynamic-page/dynamic-page-header/header/dynamic-page-header.component.ts
```diff
@@ -66,6 +66,10 @@ export class DynamicPageHeaderComponent
     @Input()
     subtitleWrap = false;
 
+    /** Heading level of the dynamic page title. */
+    @Input()
+    headingLevel = 2;
```
InnaAtanasova, line 71: You need some type safety for this. The heading level can be a number from 1 to 6. The user should be notified if the number in not one of these.

libs/platform/src/lib/dynamic-page/dynamic-page.component.ts
```diff
@@ -102,6 +102,10 @@ export class DynamicPageComponent extends BaseComponent implements AfterContentI
     @Input()
     expandContent = true;
 
+    /** Heading level of the dynamic page title. */
+    @Input()
+    headingLevel: number;
```
InnaAtanasova, line 107: You need some type safety for this. The heading level can be a number from 1 to 6. The user should be notified if the number in not one of these.

InnaAtanasova (changes requested):

PR #13195, opened 2025-04-10
fix(platform): keyboard improvements for table filtering and header popover

### LINKED ISSUE #12770: VoiceOver and keybaord support for platform table filter does not work properly

bug

angular: 18.2.4
fundamental: 0.53.1-rc.23

Create platform table with filter (it can be done based on [this example](https://sap.github.io/fundamental-ngx/#/platform/table/basic#initial-state) .

1. Try to set some table filter using only keyboard. List of filterable columns should appear ('price', 'status', 'status color'). It is not possible to use only keyboard (space or enter key) to go inside 'price' or 'status' or 'status color' and set filter values.

2. Try to set some table filter using only VoiceOver (MacOs version 14.7). After selecting first filter property ('price'), the focus on filter dialog is lost and moved to main window.

Accessibility expert comment: _there is a trap for a user- it is not possible to go further with the reading because of the loop from filter to main window_

mikerodonnell89:
fixes #12770

### DIFF
--- a/libs/platform/table/components/table-cell-header-popover/table-cell-header-popover.component.ts
+++ b/libs/platform/table/components/table-cell-header-popover/table-cell-header-popover.component.ts
@@ -4,6 +4,7 @@ import {
     ChangeDetectionStrategy,
     Component,
     DestroyRef,
+    ElementRef,
     Input,
     QueryList,
     TemplateRef,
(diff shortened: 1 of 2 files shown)


InnaAtanasova (approved):

droshev:
@mikerodonnell89 unit tests are failing

droshev (approved):

PR #13227, opened 2025-04-23
fix(core): add User Menu to ShellBar actions, update examples

InnaAtanasova:
## Related Issue(s)
closes none, reported by email

## Description
ShellBar update: renders User Menu component inside ShellBar actions (the old way is kept for now to avoid breaking changes)
ShellBar example update: all examples use the new User Menu


mikerodonnell89 (approved):
LGTM, my only hesitation is that the popover is huge, maybe start with the "other accounts" closed?

droshev (approved):

PR #13235, opened 2025-04-25
fix(platform): fixed the overflow in icon tab bar component

### LINKED ISSUE #13136: Icon tab bar overflow exceeding container width
Icon tab bar overflow is not working properly - contents exceeding the container width

Image

alexhristov14:
fixes #13136


mikerodonnell89 (changes requested):
Looks like this works but it is causing a test to fail. Maybe something is wrong with the test rather than your changes?

mikerodonnell89 (approved):

PR #13236, opened 2025-04-25
fix(core): remove fdp from core object attribute

droshev:
remove fdp from core object attribute


InnaAtanasova (approved):

mikerodonnell89 (approved):
Does not look like this is used anywhere in the lib itself. I suppose this could potentially be a breaking change if anyone is using the fdp selector

PR #13249, opened 2025-04-29
fix(core): bring a11y improvements for Bar component

InnaAtanasova:
## Related Issue(s)
closes none

## Description
- add role to the parent element
- update documentation with aria-label

### DIFF
--- a/libs/core/bar/bar.component.ts
+++ b/libs/core/bar/bar.component.ts
@@ -36,6 +36,9 @@ export type BarDesignType = 'header' | 'subheader' | 'header-with-subheader' | '
             defaultContentDensity: ContentDensityMode.COMPACT
         })
     ],
+    host: {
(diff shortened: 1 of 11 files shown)


libs/core/bar/bar.component.ts
```diff
@@ -36,6 +36,9 @@ export type BarDesignType = 'header' | 'subheader' | 'header-with-subheader' | '
             defaultContentDensity: ContentDensityMode.COMPACT
         })
     ],
+    host: {
+        '[attr.role]': '"toolbar"'
```
mikerodonnell89, line 40: Could just do `'role': 'toolbar'` and avoid the data binding

mikerodonnell89 (approved):
Looks good - just one small thing, see above

droshev (approved):

PR #13267, opened 2025-05-01
fix(core): file uploader button a11y

### LINKED ISSUE #13230: [fd-file-uploader][Core]  buttonAriaLabel does not work and please support buttonAriaLabelledBy
Is it a bug ?

Yes, it's a bug.

Example(basic example in the document)
https://stackblitz.com/run?file=src%2Fapp%2Ffile-uploader-example.component.html

Image

Image

For support  buttonAriaLabelledBy, according to ACC-264.1(Level A)

ACC-264.1: Are all UI elements presented correctly using assistive technologies?
Follow the test case.
Identify all [elements on the screen] that are interactive or/and have a semantic meaning or serve only decoration purposes.
Check if [element role, name, state(s), values and other element attributes] for interactive and semantic/non-semantic elements can be determined by [assistive technology (screen reader)]when they have keyboard or virtual focus on desktop devices or been navigated using explore by touch navigation on mobile devices.

droshev:
fixes #13230

### DIFF
--- a/libs/core/button/base-button.ts
+++ b/libs/core/button/base-button.ts
@@ -61,7 +61,7 @@ export class BaseButton implements HasElementRef {
      * Text rendered inside button component
      */
     @Input()
-    label: string;
+    label: string | undefined;
 
     /** Whether to apply menu mode to the button.
      * Default value is set to false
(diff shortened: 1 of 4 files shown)


mikerodonnell89 (changes requested):
Looks like it works but five return statements in a single function is not my favorite thing in the world stylistically. Maybe some comments explaining what's happening?

mikerodonnell89 (approved):

PR #13276, opened 2025-05-08
fix(core): a11y updates for Carousel

InnaAtanasova:
## Related Issue(s)
closes none

## Description
- parent element: markup update, role, added aria-roledescription and aria-activedescendant + JS logic
- carousel item: updated role, host binding classes, added a11y attributes: aria-setsize, aria-posinset, aria-selected, aria-hidden and the necessary JS logic to generate them.
- carousel page indicators: markup update, removed `attr.data-slide-to`, added aria-label and i18n for it

### DIFF
--- a/libs/core/carousel/carousel.component.ts
+++ b/libs/core/carousel/carousel.component.ts
@@ -199,6 +199,15 @@ export class CarouselComponent implements AfterContentInit, AfterViewInit, After
     /** @hidden Start index of currently active items */
     currentActiveSlidesStartIndex = 0;
 
+    /** @hidden End index of currently active items */
+    currentActiveSlidesEndIndex = 0;
+
+    /** @hidden an array of id(s) of currently active item(s) */
+    currentActiveSlidesIds: string[] = [];
+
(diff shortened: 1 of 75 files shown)


libs/core/carousel/carousel.component.ts
```diff
@@ -304,6 +316,24 @@ export class CarouselComponent implements AfterContentInit, AfterViewInit, After
...
+            slide.ariaPosinset.set(index + 1);
+            slide.ariaHidden.set(!isActive);
+            slide.ariaSelected.set(isActive);
+
+            if (isActive) {
+                this.currentActiveSlidesIds.push(slide.id);
+            }
+        });
+
+        this.ariaActivedescendant = this.currentActiveSlidesIds.join(' ');
```
InnaAtanasova, line 336: this is necessary as the Carousel can have more than 1 visible slides at the same time and the string represents the ids of the visible items separated by a space: "item-1 item-2 item-3".

droshev (approved):

PR #13280, opened 2025-05-12
fix(platform): table sort/group dialog heading a11y improvements

### LINKED ISSUE #12789: Platform table filter and sort - 'Sort order', 'sort by' and filter by' should not be elements of the lists

bug

angular: 18.2.4
fundamental: 0.53.1-rc.23

1. Create some platform table with sorting and filtering. It may be copy of [example from documentation](https://sap.github.io/fundamental-ngx/#/platform/table/basic#initial-state).
2. open sort dialog
3. open browser's developer tools and check structure of the list. 'sort order' is implemented as a list item. In fact it should be list title and implemented example as a `<hn>` tag in sequentially descending order. 
Screenshot 2024-11-28 at 14 29 46

We can observe similar situation for 'sort by' and 'filter by'.

Because of this issue screen readers (tested with VoiceOver MacOs version 14.7) announce inappropriate items length and user can think that there is one more option than can be selected.

mikerodonnell89:
fixes #12789


libs/platform/table/components/table-view-settings-dialog/grouping/grouping.component.html
```diff
@@ -1,9 +1,9 @@
+<h3 fd-list-group-header>
+    <span fd-list-title [attr.id]="groupOrderHeaderId">
+        {{ 'platformTable.groupDialogGroupOrderHeader' | fdTranslate }}
+    </span>
+</h3>
```
InnaAtanasova, line 5: the heading level should be customizable. We can't know for sure the sequence of elements on the apps so hardcoding `h3` could cause problems. We can have it as default, but should add an option to modify it. I suggest you use a div with role heading and have the heading level be set by an input prop.

InnaAtanasova (changes requested):

InnaAtanasova (approved):

PR #13282, opened 2025-05-12
fix(platform): table pop in border issues

mikerodonnell89:
Fixes some small issues with pop-in tables - double border around rows, and the header should not have vertical borders.

before:
Screenshot 2025-05-12 at 5 44 01 PM

after:
Screenshot 2025-05-12 at 5 46 21 PM


libs/docs/platform/table/examples/platform-table-responsive-columns-example.component.html
```diff
@@ -16,7 +16,7 @@
     [rowsActivable]="true"
     emptyTableMessage="No data found"
     [noBorderX]="true"
```
InnaAtanasova, line 18: not related to the issue, but since we are changing `noBorderY` to `noVerticalBorders`, shouldn't we also change `noBorderX` to `noHorizontalBorders`?

mikerodonnell89: They actually do different things. Input names could use some improvement, but the API docs clarify it a bit.

```
    /** Table without horizontal borders. */
    @Input()
    noHorizontalBorders = false;
```

vs

```
    /** Table body without horizontal borders. */
    @Input()
    noBorderX = false;
```

noHorizontalBorders:
Screenshot 2025-05-20 at 5 44 20 PM


noBorderX:
Screenshot 2025-05-20 at 5 42 17 PM

InnaAtanasova (approved):

Issue #12692, opened 2024-11-05
additional information read out by jaws (x of x)


AlexanderSchygulaEviden:
#### Is this a bug, enhancement, or feature request?
probably a bug request
#### Describe your proposal.
When you enter the sidenav by clicking or navigating through the CX Sidenav in version 50. and beyond the position will be read out additionally by jaws screenreader:

"Level 1 item (3 of 5)"
"3 of 5"

the second counter comes sometimes with wrong values like 1 of 1 or 1 of 2 when for examples 9 sub treeitems are given in a group
#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
angular 17
fundamentla 50 and 53

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
Open the application
join the sidenav
open a submenu in the sidenav
navigate in and out the submenus with the arrow keys. 
when you enter or leave a submenu either way (up or down) it will read out numbers additionally.
#### In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations.
for a consistent read of the numbers you can use aria-posinset to specify in which position the item is in the given submenu and aria-setsize to describe the number of items in the subtree.
the numbers will be then read out depending on the users jaws settings correctly 
#### Did you check the documentation and the API?
yes
#### Did you search for similar issues?
no

mikerodonnell89:
@AlexanderSchygulaEviden can you give more specifics regarding how the current behavior is incorrect and how you'd expect it to work? Maybe using this example: https://stackblitz.com/edit/uez7kxxu?file=src%2Fapp%2Fside-navigation-cozy-wide-text-only-3-level-example.component.html

Explain what you'd expect the screenreader to say when focusing this item?

<img width="1139" alt="Image" src="https://github.com/user-attachments/assets/b74b1f0a-6aed-4d9c-bfc1-365ad2740c3f" />

mikerodonnell89:
Closing as this component has been deprecated

PR #13296, opened 2025-05-19
fix: remove transform-translations from precommit hook, prepare *.properties files for new translations

mikerodonnell89:
part of #12692 

**EDIT:**
after a team discussion, we've decided the best course of action is to remove the pre-commit hook that runs the `*.properties` to `*.ts translation` script. Moving forward with this PR. Original problem described below

Came across an interesting problem with a translation issue. In the issue linked above, we had to make a change to existing translated text, rather than adding a new translation. The text had been previously translated by the sap translation service. So in order to trigger the translation service that we need an updated translation, I need to modify each *.properties file. Since I'm not qualified to attempt translation myself, I'll need to change each *.properties file to English. Our pre-commit hook would automatically convert this to the *.ts for each language, and each (previously correctly translated) language would now have this string in English while we wait for the updated translations from the translation service. A potential alternative would be to remove the translation task that triggers automatically during the pre-commit hook, and developers for our library must be aware that they’ll need to run this script locally when necessary when adding new translations. So we have a decision to make with two options:

- potential short-term translation regressions and keep the pre-commit hook as-is, or
- remove the pre-commit hook as I’ve done in this PR, but we will need to understand that, when the translations are provided, we’ll need to run the script to convert the new *.properties files to *.ts.


g-cheishvili:
> * remove the pre-commit hook as I’ve done in this PR, but we will need to understand that, when the translations are provided, we’ll need to run the script to convert the new *.properties files to *.ts.

Not only that, when translation team creates a PR, the hook and the dependsOn ensure that the .ts files are updated along the .properties. They could not launch them manually.

What I'd suggest is having additional small library, which would be overriding the translation keys if there is a need for manually written texts

droshev:
closing for now and we can reopen it once it blocks us agian

PR #13299, opened 2025-05-22
fix(platform): rename TabDestinyMode to TabDensityMode for consistency

Anton-Schuster:
## Description

Fixed a typo in type name TabDestinyMode to TabDensityMode across several files in the platform/icon-tab-bar package.
This change improves code readability and consistency.
No logic or API changes, only the name was corrected.

Marked the old exported type as deprecated.


droshev (approved):

mikerodonnell89 (changes requested):
Good catch. It is an exported type, so potentially a breaking change when this gets merged, maybe we should leave the old as well and deprecate it

Anton-Schuster:
I added the old type back and marked it as deprecated. Is something else missing?

Issue #13302, opened 2025-05-25
BUG: SIGNIFICANT performance issue, suspected to be related to fd-button


LiorCapsiSAP:
Observing performance issues when having hundreds of fd-buttons on screen. Setting the button's ariaLabel seem to make a noticeable performance improvement.
Probably related to buttonArialabel getter running on **all the buttons** every tick.

to replicate-
have a page with hundreds of fd-buttons, all _WITHOUT ariaLabel_. run profiler and click a button.

using v0.56.1
ng 19.2.10

Thank you,
CDP Universe team.

PR #13304, opened 2025-05-26
fix(core): optimize Button performance with signals

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/13302

## Description
- made `ariaLabel` and `ariaDescription` input signals
- instead of getters used computed to get the value of `aria-label` and `aria-description`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,25 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.56.3-rc.2](https://github.com/SAP/fundamental-ngx/compare/v0.56.3-rc.1...v0.56.3-rc.2) (2025-05-27)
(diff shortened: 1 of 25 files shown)


droshev:
@LiorCapsiSAP we did some measures:
without the fix:
30000 non-accessible standard buttons:
main-30000btns
1000 non-accessible standard buttons:
main-1000btns
100 non-accessible standard buttons:
main-100btns

with the fix:
30000 non-accessible standard buttons:
fix-30000btns
1000 non-accessible standard buttons:
fix-1000btns
100 non-accessible standard buttons:
fix-100btns

You can see a visible difference when we have tens of thousands of non-accessible buttons. There are some tiny difference when we have like 1000 buttons.
Here is the fix
Nevertheless, my advise is to improve your buttons and make them more accessible

mikerodonnell89 (approved):

PR #13341, opened 2025-06-10
chore: update to angular 20

### LINKED ISSUE #13338: Support Angular 20
### Describe your proposal

Angular 20 released on May 28, therefore the library should also be updated to support it. 
Especially new Signal-based APIs make it interesting.

20

_No response_

### StackBlitz example(s)

_No response_

### Declaration

robertIsaac:
fixes #13338


robertIsaac:
@g-cheishvili can you check the changes for the `content-density-observer.service.ts` since you wrote it

g-cheishvili:
> @g-cheishvili can you check the changes for the `content-density-observer.service.ts` since you wrote it

Not much had changed in there to be afraid, just import path removals and the previously deprecated injection flags replacement. I'd say it is safe

robertIsaac:
@droshev can someone from the team take a look at the carousal component unit test?

libs/cdk/utils/pipes/async-or-sync.pipe.ts
```diff
@@ -14,7 +14,7 @@ export class AsyncOrSyncPipe implements OnDestroy, PipeTransform {
     /** @hidden */
     private _cdr: ChangeDetectorRef | null;
     /** @hidden */
-    private _asyncPipe: AsyncPipe | null;
+    private _asyncPipe: AsyncPipe;
```
g-cheishvili, line 17: I would say we do not even need the constructor at all, just delete constructor and do `new AsyncPipe(inject(ChangeDetectorRef)` or do `inject(AsyncPipe)` and get rid of ngOnDestroy code here

g-cheishvili (approved):
Good thing you removed `CommonModule`

PR #13343, opened 2025-06-11
fix(platform): disable scrolling during table row drag and drop

mikerodonnell89:
downport of https://github.com/SAP/fundamental-ngx/pull/13300

### DIFF
--- a/libs/cdk/src/lib/utils/directives/focusable-grid/focusable-grid.directive.ts
+++ b/libs/cdk/src/lib/utils/directives/focusable-grid/focusable-grid.directive.ts
@@ -63,6 +63,9 @@ export class FocusableGridDirective implements AfterViewInit {
     private readonly _focusableLists: QueryList<FocusableListDirective>;
 
     /** @hidden */
+    _preventKeydown = false;
+
+    /** @hidden */
     constructor(private readonly _destroy$: DestroyedService) {}
 
     /** @hidden */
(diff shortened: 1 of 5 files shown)


CSEHoangV:
I just checked the fix as requested. After drag and drop, the scrollbar is not enabled back and the table is no longer scrollable:


https://github.com/user-attachments/assets/fcaf1c3d-2d8e-45ba-af8b-d57a234d91d2


I think we should not disable it and may allow the action to be in some certain ways, but need to do the research on functional requirement. For example, I saw this from UI5 table which still allowed table scrollable while drag/drop.


https://github.com/user-attachments/assets/534aebbd-7d43-43b3-9a4d-7283819d0339

droshev:
@mikerodonnell89 will look once he is back

mikerodonnell89:
> I just checked the fix as requested. After drag and drop, the scrollbar is not enabled back and the table is no longer scrollable:

Just pushed a fix for this ^

> I think we should not disable it and may allow the action to be in some certain ways, but need to do the research on functional requirement. For example, I saw this from UI5 table which still allowed table scrollable while drag/drop.

Apologies, we are looking to implement this in a later PR - this PR just looks to prevent scrolling during a drag

droshev:
@I543348 It seems like your team doesn't need this so we will close it for now. We can always discuss it in the future if a new interest arises.

Issue #13297, opened 2025-05-20
Adopt Illustrated Messages updates from styles


droshev:
Adopt Illustrated Messages updates from [styles](https://github.com/SAP/fundamental-styles/issues/6062) and update the SVG show cases

Add a disclaimer on top of the doc page(as it is shown in the design)

`12`

PR #13353, opened 2025-06-16
fix(core): update the Illustrated messages

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13297
 
## Description
brings the new svgs


droshev (approved):
For me it is fine(as long as the designers are fine)

PR #13395, opened 2025-07-17
fix(platform): replace innerText with textContent on table row

CSEHoangV:
fix https://github.com/SAP/fundamental-ngx/issues/11899

Please refer to https://github.com/SAP/fundamental-ngx/pull/12021 for more information on the fix, it was later closed and change was included in https://github.com/SAP/fundamental-ngx/pull/12015. However, we only need change from https://github.com/SAP/fundamental-ngx/pull/12021 to fix performance since root cause came from the `innerText`.

Our application is in the process of migrating to Angular 20. We would like to migrate some of the fixes which only exist in Angular 15 downport to main branch.


droshev:
We will leave this to @mikerodonnell89 to review

libs/platform/table/components/table-row/table-row.component.ts
```diff
@@ -367,4 +384,30 @@ export class TableRowComponent<T> extends TableRowDirective implements OnInit, A
...
+        let value: string;
+        const isCellEmptyInNonColumnTemplate =
+            !column.columnCellTemplate &&
+            (!(value = get(row.value, column.key)) || (typeof value === 'string' && value.trim() === ''));
+
+        // Content comes from application template. Some element, like input field, value is not part of text node.
+        const isCellEmptyInColumnTemplate =
+            column.columnCellTemplate &&
+            tableTextContainer?.innerText?.trim() === '' &&
+            tableTextContainer?.querySelectorAll('input')?.length === 0;
```
mikerodonnell89: I think we need to check for any child element, not just input. For example if the developer has placed an image in the cell but no other text, I believe the screenreader would still read "Empty" for that cell in this case

CSEHoangV:
Still stuck with current work, I will revisit and make a suggested change later.

droshev:
@CSEHoangV Let's close it for now, you can reopen it whenever you have more time

libs/platform/table/components/table-row/table-row.component.ts
CSEHoangV: I think checking any child element is not enough as it is possible we have multiple layers of child element and visible element can be somewhere in nested levels. And you may also need to check other elements like radio, textarea,... not just input like you said. In addition, consumer can add custom aria-label in column template to have their own announcement so I don't think library needs to handle it.

I would say for the fix, I would keep the innerText as default fallback since we already implemented this. And since it does not always work and this comes from column template, consumer should decide when to do the announcement and customize their own ariba-label. It is easier for consumer to handle it and more effective.

libs/platform/table/components/table-row/table-row.component.html
```diff
@@ -136,7 +136,8 @@
         "
         [attr.aria-expanded]="_isTreeRowFirstCell($index, row) ? row.expanded : null"
         [attr.data-nesting-level]="$index === 0 ? row.level + 1 : null"
-        (cellFocused)="_tableRowService.cellFocused($event); _tableRowService.cellActivate($index, row)"
+        (cellFocused)="_handleCellFocused($event, $index, row, column, tableTextContainer)"
+        (cellFocusedOut)="_announceEmptyCell.set(false)"
```
mikerodonnell89: I think you can just use `(focusout)` and there's no need for this new event

libs/docs/platform/table/examples/platform-table-custom-column-example.component.html
```diff
@@ -48,16 +48,20 @@
...
         <fdp-table-cell *fdpCellDef="let item">
-            <fdp-input
-                type="text"
-                name="description"
-                placeholder="Enter the description"
-                [(ngModel)]="item.description"
-            >
-            </fdp-input>
+            @if (!item.description) {
+                <span [attr.aria-label]="'Empty Input'"></span>
```
mikerodonnell89:
```suggestion
                <span [attr.aria-label]="'Empty Cell'"></span>
```

CSEHoangV:
> Looks like it was a problem from before this PR, but screenreader is reading "Empty" on cells that aren't empty when using `[announceEmptyCell]="true"`: Screenshot 2025-10-21 at 4 11 24 PM

Yes, because `innerText` can only check textNode. It cannot check against other elements or shadow text. That was why I tried to add more implementations at the beginning to also check for those but realized there were many hassles to handle. So, I proposed consumer should provide their own aria-label for custom template by setting `announceEmptyCell` to false.

Current:

Screenshot 2025-10-21 at 1 27 23 PM

libs/docs/platform/table/examples/platform-table-custom-column-example.component.html
CSEHoangV: Looks you approved, let's me know if I still need to address this.

libs/platform/table/components/table-row/table-row.component.html
CSEHoangV: Looks you approved, let's me know if I still need to address this.

mikerodonnell89: @CSEHoangV sorry yes, we would like to see this change

libs/docs/platform/table/examples/platform-table-custom-column-example.component.html
mikerodonnell89: Sorry yes, we'd like to see this as well

PR #13397, opened 2025-07-18
fix(core): update inline help markup

MariaIDineva:
## Related Issue(s)

fixes: https://github.com/SAP/fundamental-ngx/issues/13440

## Description

**Changes implemented:**

- The tooltip element is no longer a `child` of the element that triggers tooltip visibility, it is now a `sibling` which allows the screen reader to read the content in the correct order. It used to read the tooltip text first and the trigger element second. Now it's the other way round.
  - Eg: For the "Inline Help" input example, JAWS used to read "Inline help tooltip. Inline help inside input edit. blank. Inline help tooltip." It now reads: "Inline help inside input edit. blank. Inline help tooltip."
  -  Eg: For the "Inline Help with Changed Trigger Event" example, JAWS used to read "Inline help tooltip. Click on me, to display inline help button." It now reads "Click on me, to display inline help button. Inline help tooltip."
- The tooltip element container, i.e. `fd-popover-body` now has `role="tooltip"`.
-  The tooltip element also has an `id` which matches the `aria-describedby` attribute of the element triggering the tooltip. For this to work, declaration of `_describedBy` field in `InlineHelpDirective` is moved to the constructor instead of the `ngOnInit` method as `_describedBy` field contains the value used in both attributes.

**NB!**
Screen readers do not announce the word "tooltip"even if an element has role="tooltip". The role="tooltip" is for assistive technology semantics, not for user-facing speech. The only way, as of now, to have the word "tooltip" read by a screen reader is to add that word to the accessible text.

### DIFF
--- a/libs/docs/core/inline-help/e2e/inline-help.e2e-spec.ts
+++ b/libs/docs/core/inline-help/e2e/inline-help.e2e-spec.ts
@@ -14,7 +14,7 @@ import {
 import { InlineHelpPo } from './inline-help.po';
 import { customMessage, defaultMessage } from './inline-help-contents';
 
-xdescribe('Inline help test suite', () => {
+describe('Inline help test suite', () => {
     const inlineHelpPage = new InlineHelpPo();
     const {
         inlineHelpIcons,


droshev:
We need to align this with the accessibility standard and expectations

droshev:
closing as it is not needed.

PR #13399, opened 2025-07-23
docs: add deprecation message

droshev:
add deprecation message

### DIFF
--- a/libs/docs/shared/src/lib/core-helpers/header/header.component.ts
+++ b/libs/docs/shared/src/lib/core-helpers/header/header.component.ts
@@ -1,11 +1,20 @@
 import { ChangeDetectionStrategy, Component } from '@angular/core';
+import { MessageStripComponent } from '@fundamental-ngx/core/message-strip';
 
 @Component({
     selector: 'header',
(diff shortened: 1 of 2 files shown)


libs/docs/shared/src/lib/core-helpers/header/header.component.ts
```diff
@@ -1,11 +1,20 @@
...
+import { MessageStripComponent } from '@fundamental-ngx/core/message-strip';
 
 @Component({
     selector: 'header',
+    imports: [MessageStripComponent],
     template: `
         <h1 class="header">
             <ng-content></ng-content>
         </h1>
+        <fd-message-strip type="warning" [dismissible]="false">
```
MariaIDineva: This warning will appear on top of all components, is that the goal of this change?

MariaIDineva (approved):

libs/docs/shared/src/lib/core-helpers/header/header.component.ts
droshev: In general yes. I added an exception for the `cdk` package.

Issue #12732, opened 2024-11-18
[Accessibility: ACC-270.5] Icon Tab Bar: Focus moves to the unselected tab


divyamalhotra488:
#### Is this a bug, enhancement, or feature request?
Bug

#### Describe your proposal.
Within the tab group , always the selected tab must receive the first tab focus.

#### Can you handle that on the application side
No

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
All

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
Component Used: [Icon Tab Bar](https://sap.github.io/fundamental-ngx/#/platform/icon-tab-bar)

**Steps to reproduce:**
1. Use Icon tab bar and select a tab that is not the first one.
2. Now, start navigating UI from screen start through keyboard.
Even though the ‘Item 1’ tab is in the selected state but when user use tab key to navigate all the UI elements always the ‘Item 0’ tab receives the first tab focus in the tab group.

![Screenshot 2024-11-18 at 6 02 25 PM](https://github.com/user-attachments/assets/f2d74963-4946-4687-8a37-30415f466103)

#### Please provide stackblitz example(s).
You can try ngx examples, issue is reproducible there.

#### In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations.
Yes, issue is raised by an accessibility expert.

#### Did you check the documentation and the API?
Yes

#### Is there anything else we should know?

**_IMPORTANT: Please refrain from providing links or screenshots of SAP's internal information, as this project is open-source, and its contents are accessible to anyone._**

PR #13401, opened 2025-07-25
Fix/focus moves to unselected tab in Icon Tab Bar

MariaIDineva:
## Related Issue(s)

fixes https://github.com/SAP/fundamental-ngx/issues/12732

## Description

When the user navigates using the tab button through [Icon Tab Bar](https://sap.github.io/fundamental-ngx/#/platform/icon-tab-bar), focus now correctly goes to the selected tab item, not always to the first item in the tab bar.

## Screenshots

Before:
before
After:
after


InnaAtanasova (approved):
Looks great!

droshev (approved):

PR #13403, opened 2025-07-27
chore(deps): migrate to nx v. 21.3 and angular v. 20.1

droshev:
migrate to nx v. 21.3 and angular v. 20.1

workaround for importing scss file for stackblitz due to a **breaking change** introduced by https://github.com/angular/angular-cli/pull/30781 due to Vite dev middleware

opened a bug to angular https://github.com/angular/angular-cli/issues/30792


droshev:
Angular fixed the issue => maybe it is a good idea to wait for the release with the fix.

Issue #13229, opened 2025-04-24
[d-message-strip][Core]   Screen Reader ACC-264.1(Level A) Message strip information is not read


Meggie886:
Is it a bug ?
Yes 

version v0.54.2

https://stackblitz.com/run?file=src%2Fapp%2Fmessage-strip-example.component.html

Screen reader cannot read the message tip(error ,warning, success..)

![Image](https://github.com/user-attachments/assets/51589ccf-1ae9-4269-8f5e-ed8af950add5)

droshev:
@Meggie886 Can you post what JAWS announcement was and what expectation is? BTW the stackblitz link is not correct.

Meggie886:
ACC-264.1: Are all UI elements presented correctly using assistive technologies?
Follow the test case.
Identify all [elements on the screen](https://wiki.one.int.sap/wiki/display/Acc/ACC-264+Details#ACC264Details-identifyelements) that are interactive or/and have a semantic meaning or serve only decoration purposes.
Check if [element role, name, state(s), values and other element attributes](https://wiki.one.int.sap/wiki/display/Acc/ACC-264+Details#ACC264Details-acc2641howto) for interactive and semantic/non-semantic elements can be determined by [assistive technology (screen reader)](https://wiki.one.int.sap/wiki/display/Acc/ACC-264+Details#ACC264Details-usingscreenreader) when they have keyboard or virtual focus on desktop devices or been navigated using explore by touch navigation on mobile devices.
Realized
(1) All UI elements are presented correctly using assistive technologies
Partly realized
(1) Some UI elements are not presented correctly using assistive technologies: <list of UI elements>
Not realized
(1) None of the UI elements is presented correctly using assistive technologies

Expectation : Screen reader can read the type pf  Message strip


And the link I don't know why you cannot open, if you cannot just check Message Strip Component, any example is ok

droshev:
@nikolay-kolarov Can you advise us what is exactly expected that we should update? thank you very much.

nikolay-kolarov:
Hi @droshev,
The icons are designed to be decorative, so you need to add the semantic of the Info bar, so that it can be reached with assistive technologies (in reading mode or with focus).
You can have a look at the message strip component: https://sap.github.io/ui5-webcomponents/components/MessageStrip/


For the reading mode:

![Image](https://github.com/user-attachments/assets/f4f7cf09-48ae-4d80-85c0-9e872800f634)

For the button on focus:

[...]

PR #13448, opened 2025-08-15
fix(core): message strip information is not read

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13229

## Description

When focus is on a closable message strip, JAWS used to only read the button text thus not providing enough information about the message strip type to disabled users. With this change:
- JAWS now reads hidden information about the message strip in the format: `{{messageStripType}}  message strip closable`. 
Eg: `Warning message strip closable`
- The role of the message strip is also updated to `note` (used to be `alert`) to match figma.
- Button default text is updated as well to reflect the latest figma. Instead of just `Dismiss button`, it now reads `{{messageStripType}} message strip close button`
Eg: `Warning message strip close button`

## Screenshots
Latest design:
Screenshot 2025-08-15 at 10 39 21


libs/core/message-strip/message-strip.component.html
```diff
@@ -1,25 +1,32 @@
...
-        glyph="decline"
-        fdType="transparent"
-        fdCompact
-        (click)="dismiss()"
-        [attr.aria-controls]="id"
-        [attr.aria-label]="dismissBtnTitle || ('coreMessageStrip.dismissLabel' | fdTranslate)"
-        [attr.title]="dismissBtnTitle || ('coreMessageStrip.dismissLabel' | fdTranslate)"
-    ></button>
-}
+<div role="note" [attr.aria-labelledby]="id + '__hidden-text ' + id + '__content-text'">
```
InnaAtanasova: You can add the role and the `aria-labelledby` directly to the host element and avoid having this extra `div` element in the markup. 

You don't need the underscores in the ids. You can have something like:
`"id + '-hidden-text ' + id + '-content-text'"`

libs/core/message-strip/message-strip.component.scss
```diff
@@ -11,3 +11,13 @@
 .fd-has-display-none {
     display: none;
 }
+
+.fd-message-strip__hidden-text {
```
InnaAtanasova: Could you rename this to `fd-message-strip__sr-only` as it's in Fundamental-styles? Once we adopt the latest version of Fundamental-styles we will remove this CSS and cnsume it directly from Fundamental-styles.

libs/core/message-strip/message-strip.component.html
```diff
@@ -1,25 +1,32 @@
...
+    @if (shouldShowIcon) {
+        <div class="fd-message-strip__icon-container">
+            @if (icon) {
+                <ng-template [ngTemplateOutlet]="icon.templateRef"></ng-template>
+            } @else {
+                <div class="fd-message-strip__icon-wrapper" aria-hidden="true">
+                    <fd-icon [glyph]="typeSpecificIconName"></fd-icon>
+                </div>
+                <span class="fd-message-strip__hidden-text" id="{{ id }}__hidden-text">{{
```
InnaAtanasova: Should be `fd-message-strip__sr-only`. See my other comment.

libs/core/message-strip/message-strip.component.html
```diff
@@ -1,25 +1,32 @@
...
-        [attr.title]="dismissBtnTitle || ('coreMessageStrip.dismissLabel' | fdTranslate)"
-    ></button>
-}
+<div role="note" [attr.aria-labelledby]="id + '__hidden-text ' + id + '__content-text'">
+    @if (shouldShowIcon) {
+        <div class="fd-message-strip__icon-container">
+            @if (icon) {
+                <ng-template [ngTemplateOutlet]="icon.templateRef"></ng-template>
+            } @else {
+                <div class="fd-message-strip__icon-wrapper" aria-hidden="true">
```
InnaAtanasova: Why do you need this additional `div` element around the icon? `role="presentation"` and `aria-hidden="true"` should be added to the icon component directly, without a wrapping element.

libs/core/message-strip/message-strip.component.html
```diff
@@ -1,25 +1,32 @@
...
-        fdCompact
-        (click)="dismiss()"
-        [attr.aria-controls]="id"
-        [attr.aria-label]="dismissBtnTitle || ('coreMessageStrip.dismissLabel' | fdTranslate)"
-        [attr.title]="dismissBtnTitle || ('coreMessageStrip.dismissLabel' | fdTranslate)"
-    ></button>
-}
+<div role="note" [attr.aria-labelledby]="id + '__hidden-text ' + id + '__content-text'">
+    @if (shouldShowIcon) {
+        <div class="fd-message-strip__icon-container">
```
InnaAtanasova: If we remove `aria-hidden="true"` from the wrapping icon element, in case of a custom icon (the if case), the icon won't be ignored by the screen reader.

libs/core/message-strip/message-strip.component.html
MariaIDineva: Thanks for the hint!

libs/core/message-strip/message-strip.component.scss
MariaIDineva: Oh, yes, thanks! That's what we discussed last week. I thought I had pushed it, but it seems like I missed it, sorry.

libs/core/message-strip/message-strip.component.html
MariaIDineva: Hm, I don't recall removing it. I'll bring it back, thanks for spotting it!

libs/core/message-strip/message-strip.component.html
MariaIDineva: I was trying to follow Niki Kolarov's suggestion and replicate what we have in webcomponents - https://github.com/SAP/fundamental-ngx/issues/13229#issuecomment-2886989967 
I will review your suggestion and refactor.

libs/core/message-strip/message-strip.component.ts
```diff
@@ -162,4 +201,59 @@ export class MessageStripComponent implements OnInit, OnChanges, CssClassBuilder
...
+                break;
+            case MessageStripTypeEnum.INFORMATION:
+                this.messageStripHiddenText$ = this._buildMessageStripHiddenText(MessageStripAnnouncement.INFORMATION);
+                this.defaultDismissButtonText$ = this._getDefaultHiddenButtonText(MessageStripAnnouncement.INFORMATION);
+                break;
+            default:
+                this.messageStripHiddenText$ = of(DEFAULT_HIDDEN_TEXT);
+                this.defaultDismissButtonText$ = of(DEFAULT_DISMISS_BUTTON_TEXT);
+        }
+    }
```
InnaAtanasova: Another solution to this would be use a map. Something like:

```
private setScreenReaderTexts(): void {
    if (!this.type) {
        return;
    }

    const announcementMap: Record<MessageStripTypeEnum, MessageStripAnnouncementType> = {
        [MessageStripTypeEnum.WARNING]: MessageStripAnnouncement.WARNING,
        [MessageStripTypeEnum.SUCCESS]: MessageStripAnnouncement.SUCCESS,
        [MessageStripTypeEnum.ERROR]: MessageStripAnnouncement.ERROR,
        [MessageStripTypeEnum.INFORMATION]: MessageStripAnnouncement.INFORMATION
    };

    const announcementType = announcementMap[this.type];

    if (announcementType) {
        this.messageStripHiddenText$ = this._buildMessageStripHiddenText(announcementType);
        this.defaultDismissButtonText$ = this._getDefaultHiddenButtonText(announcementType);
    } else {
        this.messageStripHiddenText$ = of(DEFAULT_HIDDEN_TEXT);
        this.defaultDismissButtonText$ = of(DEFAULT_DISMISS_BUTTON_TEXT);
    }
}
```

All mappings between type and announcement are in one object and adding a new type only needs updating the map, not two methods.


But this is only another approach I am sharing, I accept your solution.

InnaAtanasova (approved):

libs/core/message-strip/message-strip.component.ts
MariaIDineva: I actually like your suggestion and I will implement it. Also, I will improve the logic a bit so that I don't call the translation resolver twice in order to translate the announcement.

InnaAtanasova (approved):

Issue #13478, opened 2025-09-08
[P3][Sourcing][Platform toolbar][Accessibility] : Option for developer to set the heading level for fdp-toolbar


I543348:
### Describe your proposal

Describe your proposal
Please provide us an option to send the "headinglevel" for fdp-toolbar title, The issue happens when we have h1 on page for main heading at application side and suddenly we have h4 coming from toolbar heading title , 
hierarchy of title is not maintained and thats a problem for disabled users

Which versions of Angular and Fundamental Library for Angular are affected?
0.43.57

Steps to Reproduce (for bugs)
Inspect on toolbar title

<img width="1157" height="364" alt="Image" src="https://github.com/user-attachments/assets/fb8fa539-ec6e-4503-a386-52455703b343" />

StackBlitz example(s)

Declaration

I'm not disclosing any internal or sensitive information.

### Which versions of Angular and Fundamental Library for Angular are affected?

17

### Steps to Reproduce (for bugs)

Inspect on toolbar title element and notice heading level or h4 tag

### StackBlitz example(s)

Inspect on toolbar title element and notice heading level or h4 tag

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

I543348:
Thanks @InnaAtanasova , could you pls confirm which version of ng15 we can consume the fix?

PR #13479, opened 2025-09-08
fix(platform): add option to set heading level for Table Toolbar

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/13478
 
## Description
The Platform Table uses a Toolbar with a Title component, where the title is set through an input property. However, this implementation does not allow users to define the heading level.

This PR introduces a new input property that enables users to specify the heading level of the title. The visual appearance of the title remains unchanged; only the `aria-level` attribute is updated accordingly.

## Screenshots

Screenshot 2025-09-08 at 2 03 37 PM


I543348:
@InnaAtanasova , can we please get this released before Sep 30?

Issue #13445, opened 2025-08-14
Step input minimum not recognized


LEinfeldt:
### Describe your proposal

Using the step input, which provides a `min` value, it is possible to circumvent the min value. 

When setting `min=0` and loading the component initially, you can click on the decrement button and receive a '-1' value.

The reason is probably in this method: https://github.com/SAP/fundamental-ngx/blob/e3c3615ea9a0d1349ddf994ad5eeab731f34a3f6/libs/core/step-input/step-input.component.ts#L418

### Which versions of Angular and Fundamental Library for Angular are affected?

Angular 18, Fundamental: 0.54.2

### Steps to Reproduce (for bugs)

1. Create a step input
2. Set a `min` value to 0
3. Run the application
4. Click decrease button
5. See a value -1 appear in the input

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

MariaIDineva:
We seem to be unable to reproduce the reported bug. Whenever a `min` is set, the decrement button gets disabled and there is no way to go below the minimum value. Even if you manually set the value to be below the minimum, upon pressing enter, it gets set to the min. Please check out the screenshot below for reference:

<img width="1477" height="1125" alt="Image" src="https://github.com/user-attachments/assets/cc4e97bd-2233-4437-bc5c-7aa9c75a72ca" />

Please check that you are correctly setting the min value (`[min]="0"`, spot the data binding `[]`), test again and let us know if that worked.

MariaIDineva:
@LEinfeldt, were you able to test what I suggested?

LEinfeldt:
Hi @MariaIDineva,

thanks for your response and excuse my late answer please.

I just had a look into the suggestion and recognized, that my provided information was too little.
We use Reactive forms and have a `FormControl` which is initially null. This leads to the decrement button being pressable initially. However, I was also able to reproduce the behaviour in the Stackblitz linked in the Doc.

When setting a minimum, the decrement button is disabled correctly. When a user clears the input shown in the input component (mark it and delete the value with backspace) you enter a state, where no value is displayed, and the decrement button is enabled:

<img width="1600" height="749" alt="Image" src="https://github.com/user-attachments/assets/22e57267-554c-4ca3-a377-d1606b6bcf9b" />

After pressing the decrement button, you see a `-1` inside the form:

<img width="289" height="103" alt="Image" src="https://github.com/user-attachments/assets/750be916-558b-402c-812f-a5ca09aa506b" />

[...]

PR #13486, opened 2025-09-10
fix(core): Step Input minimum not recognized when input is cleared

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/13445

## Description
When the input is cleared, the increment and decrement button should start from the min value and max value to prevent going above/below the set max/min values


libs/core/step-input/step-input.component.ts
```diff
@@ -326,12 +326,20 @@ export class StepInputComponent implements OnInit, AfterViewInit, OnDestroy, Con
 
     /** @hidden */
     get canIncrement(): boolean {
-        return this.value == null || this.value + this.step <= this._max;
+        if (this.value == null) {
+            // If max is set, only allow increment if min < max
+            return isNaN(this.max) || (!isNaN(this.min) && this.min <= this._max);
```
MariaIDineva: I see the logic is slightly different for `canIncrement` and  `canDecrement` in case the value is invalid. Do we really want them to behave differently? I mean shouldn't both either:
-  return `isNaN(this.max)` / `isNaN(this.min)` in which case we will end up with both `+` and `-` disabled?
-  or both contain logic about the min/max values, i.e. `isNaN(this.max) || (!isNaN(this.min) && this.min <= this._max)` / `(!isNaN(this.max) && this.max >= this._min)` in which case we will have both `+` and `-` enabled and upon clicking on the `+`, we will start incrementing from the min value, when clicking on the `-` we will start decrementing from the max value?

Seeing the logic you've updated in `_incrementOrDecrement` method, maybe we want the latter?

InnaAtanasova: Good catch! Thank you!

Issue #12584, opened 2024-10-17
[Accessibility: ACC-264.1] List: Group label not read.


divyamalhotra488:
#### Is this a bug, enhancement, or feature request?
Bug

#### Describe your proposal.
Group labels within all expanded list boxes should be read associated to the first list item.

#### Can you handle that on the application side
No

#### Which versions of Angular and Fundamental Library for Angular are affected? Please, specify the exact version. (If this is a feature request, use current version.)
All

#### If this is a bug, please provide steps for reproducing it; the exact components you are using;
Component used: [Multi Combobox](https://sap.github.io/fundamental-ngx/#/core/multi-combobox)
Group labels within all expanded multi combobox are not read.
<img width="452" alt="image" src="https://github.com/user-attachments/assets/54e8c968-06b8-4342-802f-63c09d1a98b0">

#### Please provide stackblitz example(s).
You can try ngx examples, issue is reproducible there.

#### In case this is Accessibility related topic, did you consult with an accessibility expert? If not, please do so and share their recommendations.
Yes, issue is raised by an accessibility expert.

#### Did you check the documentation and the API?
Yes

#### Did you search for similar issues?
Yes

#### Is there anything else we should know?

**_IMPORTANT: Please refrain from providing links or screenshots of SAP's internal information, as this project is open-source, and its contents are accessible to anyone._**

droshev:
@nikolay-kolarov Can you advise us in this request?

PR #13494, opened 2025-09-16
fix(core): Multi Combobox group item not read

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/12584

## Description

Makes Multi Combobox group item focusable so that it is read by screen readers. Adds `Group Header` as `aria-roledescription` to the group header to improve description.

### DIFF
--- a/libs/core/list/directives/list-group-header.directive.ts
+++ b/libs/core/list/directives/list-group-header.directive.ts
@@ -1,4 +1,14 @@
-import { Directive, ElementRef, EventEmitter, HostBinding, HostListener, Input, Output } from '@angular/core';
+import {
+    Directive,
+    ElementRef,
+    EventEmitter,
+    forwardRef,
+    HostBinding,
+    HostListener,
+    Input,
+    Output
+} from '@angular/core';
+import { ListFocusItem } from '../list-focus-item.model';
 
 let uniqueId = 0;
 
@@ -7,11 +17,19 @@ let uniqueId = 0;
     host: {
         '[style.outline]': '"none"',
         '[class.fd-list__item]': 'true',
-        '[class.fd-list__group-header]': 'true'
+        '[class.fd-list__group-header]': 'true',
+        '[attr.tabindex]': '0',
+        '[attr.aria-roledescription]': '"Group Header"'
     },
-    standalone: true
+    standalone: true,
+    providers: [
+        {
+            provide: ListFocusItem,
+            useExisting: forwardRef(() => ListGroupHeaderDirective)
+        }
+    ]
 })
-export class ListGroupHeaderDirective {
+export class ListGroupHeaderDirective extends ListFocusItem {
     /** id of an element to be applied */
     @Input()
     @HostBinding('attr.id')
@@ -22,7 +40,9 @@ export class ListGroupHeaderDirective {
     keyDown: EventEmitter<KeyboardEvent> = new EventEmitter<KeyboardEvent>();
 
     /** @hidden */
-    constructor(readonly elementRef: ElementRef) {}
+    constructor(readonly elementRef: ElementRef) {
+        super();
+    }
 
     /** @hidden */
     @HostListener('keydown', ['$event'])


libs/core/combobox/combobox.component.html
```diff
@@ -127,22 +127,33 @@
...
     >
         <ng-content></ng-content>
         @if (groupFn) {
-            @for (group of displayedValues | listGroupPipe: groupFn; track group.key) {
-                <li role="group" fd-list-group-header [tabindex]="0">
+            @for (group of displayedValues | listGroupPipe: groupFn; track group.key; let groupIndex = $index) {
+                <li
+                    fd-list-item
+                    fd-list-group-header
+                    [ariaRole]="'group'"
```
InnaAtanasova: no need of binding here, can be just `ariaRole="group"`

libs/core/combobox/combobox.component.html
```diff
@@ -127,22 +127,33 @@
...
+                >
                     <span fd-list-title>{{ group.key }}</span>
                 </li>
-                @for (term of group.value; track term) {
+                @for (term of group.value; track term; let index = $index) {
                     <li
-                        role="option"
                         fd-list-item
+                        #item="fdListItem"
+                        [ariaRole]="'option'"
```
InnaAtanasova: same as above

libs/core/multi-combobox/multi-combobox.component.html
```diff
@@ -172,11 +173,18 @@
...
         <ng-content></ng-content>
         @if (isGroup) {
-            @for (group of _suggestions; track group) {
+            @for (group of _suggestions; track group; let groupIndex = $index) {
                 @if (!groupItemTemplate) {
-                    <label fd-list-group-header role="group">
+                    <li
+                        fd-list-item
+                        fd-list-group-header
+                        [ariaRole]="'group'"
```
InnaAtanasova: no need of binding

libs/core/multi-combobox/multi-combobox.component.html
```diff
@@ -172,11 +173,18 @@
...
+                        [attr.aria-roledescription]="'Group Header'"
```
InnaAtanasova: same

libs/core/combobox/combobox.component.html
```diff
@@ -127,22 +127,33 @@
...
         <ng-content></ng-content>
         @if (groupFn) {
-            @for (group of displayedValues | listGroupPipe: groupFn; track group.key) {
-                <li role="group" fd-list-group-header [tabindex]="0">
+            @for (group of displayedValues | listGroupPipe: groupFn; track group.key; let groupIndex = $index) {
+                <li
+                    fd-list-item
+                    fd-list-group-header
+                    [ariaRole]="'group'"
+                    [attr.aria-roledescription]="'Group Header'"
```
InnaAtanasova: same

InnaAtanasova (approved):

libs/core/combobox/combobox.component.html
```diff
@@ -127,22 +127,33 @@
...
         <ng-content></ng-content>
         @if (groupFn) {
-            @for (group of displayedValues | listGroupPipe: groupFn; track group.key) {
-                <li role="group" fd-list-group-header [tabindex]="0">
+            @for (group of displayedValues | listGroupPipe: groupFn; track group.key; let groupIndex = $index) {
+                <li
+                    fd-list-item
+                    fd-list-group-header
+                    ariaRole="group"
+                    aria-roledescription="Group Header"
```
mikerodonnell89: this needs translation

libs/core/combobox/combobox.component.html
```diff
@@ -127,22 +127,33 @@
         [style.maxHeight]="!mobile && maxHeight"
         [hasMessage]="listMessages && listMessages.length > 0"
         [byline]="byline"
+        [tabindex]="0"
```
mikerodonnell89, line 130: I'm surprised we need this - I would assume this would already be handled by the focusable list directive used internally by fd-list etc

libs/core/multi-combobox/multi-combobox.component.html
```diff
@@ -155,6 +155,7 @@
         [style.max-width.px]="autoResize && maxWidth"
         [attr.aria-labelledby]="_cva.id + '-search'"
         aria-multiselectable="true"
+        [tabindex]="0"
```
mikerodonnell89, line 158: same as above

libs/core/multi-combobox/multi-combobox.component.html
```diff
@@ -172,11 +173,18 @@
...
         @if (isGroup) {
-            @for (group of _suggestions; track group) {
+            @for (group of _suggestions; track group; let groupIndex = $index) {
                 @if (!groupItemTemplate) {
-                    <label fd-list-group-header role="group">
+                    <li
+                        fd-list-item
+                        fd-list-group-header
+                        ariaRole="group"
+                        aria-roledescription="Group Header"
```
mikerodonnell89: same as above

mikerodonnell89 (changes requested):
meant to mark as request changes

libs/core/combobox/combobox.component.html
MariaIDineva: There was a problem with keyboard navigation in JAWS which turned out to be related to those tabindexes. I'll test once more and see if this tabindex is redundant and simply removing the one on the `li`s below will do.

MariaIDineva: I tested it again and we do need this tabindex. JAWS would not only "get stuck" on the first item on the list and going up and down with the arrows would become impossible, but the speech output would also be different:

- Speech output **with** `[tabindex]="0"`:
```
expanded
blank
8 result list items
To set the value use the Arrow keys or type the value.
dialog
List box
Fruits Group Header
Apple
1 of 8
Banana
2 of 8
```
- Speech output **without** `[tabindex]="0"`:
```
expanded
blank
8 result list items
To set the value use the Arrow keys or type the value.
dialog
list
with 8 items
Fruits list box item
Apple
1 of 8
Apple list box item
1 of 8
```

Spot that if we remove the tabindex, the aria-label `Group Header` on the Fruit group header is not pronounced at all and pressing the down arrow would continuously produce `Apple list box item 1 of 8` instead of going to the next item in the list.

mikerodonnell89 (approved):

Issue #11092, opened 2023-12-14
Segmented Button a11y improvements


droshev:
#### Is this a bug, enhancement, or feature request?
enhancement

#### Describe your proposal.
It needs to go to `main` and to `ng-15-downport` branches
```
<ul 
    role="listbox" 
    aria-multiselectable="false" 
    aria-orientation="horizontal"
    aria-roledescription="Segmented button group" 
    aria-describedby="..." 
    class="..." 
    tabindex="-1">
    <li 
      tabindex="-1" 
      role="option"
      aria-roledescription="Segmented button" 
      aria-posinset="1" 
      aria-setsize="3" 
      aria-selected="false"
      class="..." 
      style="...">
    ...
    </li>
    <li 
      tabindex="0" 
      role="option"
      aria-roledescription="Segmented button" 
      aria-posinset="2" 
      aria-setsize="3" 
      aria-selected="true"
      class="..." 
      style="...">
      ...
    </li>
    <li 
      tabindex="-1" 
      role="option"
      aria-roledescription="Segmented button" 
      aria-posinset="3" 
      aria-setsize="3" 
      aria-selected="false"
      class="..." 
      style="...">
      ...
    </li>
  </ul>
```

PR #13505, opened 2025-09-24
fix(core): add some a11y improvements to segmented button

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/11092

## Description

1. Adds some missing a11y attributes as follow:
- to the segmented button listbox - `aria-multiselectable`, `aria-orientation`, `aria-roledescription`, `aria-activedescendant`
- to each of the options in the listbox - `aria-roledescription`, `aria-posinset`, `aria-setsize`, `aria-selected`

2. Removes `aria-pressed` attribute from the options in segmented button listbox as this attribute should only be present on elements with role `button` while the elements here have role `option`.

3. Adds an `id` input with a default value to `ButtonComponent`.

4. Updates `FocusableListDirective` to emit the `id` of the focused element which is used when setting `aria-activedescendant` on the segmented button listbox.

5. Adds translations of the `aria-roledescription` attributes on the segmented button listbox and on each individual option.

## Screenshots

Screenshot 2025-09-29 at 15 30 44

### DIFF
--- a/libs/cdk/utils/directives/focusable-list/focusable-list.directive.ts
+++ b/libs/cdk/utils/directives/focusable-list/focusable-list.directive.ts
@@ -40,6 +40,7 @@ import { ScrollPosition, scrollIntoView } from './scroll';
 
 export interface FocusableListItemFocusedEvent {
     index: number;
+    id: string | null;
     total: number;
 }
 
@@ -364,7 +365,8 @@ export class FocusableListDirective implements OnChanges, AfterViewInit, OnDestr
                     this._gridItemFocused$.next(directiveItem._position!);
                 }
 
-                this.itemFocused.next({ index, total: items.length });
+                const id = getItemElement(item)?.id ?? null;
+                this.itemFocused.next({ index, id, total: items.length });
                 this._focusableItems.forEach((i) => i.setTabbable(i === directiveItem));
                 this._keyManager?.setActiveItem(index);
             })
--- a/libs/cdk/utils/directives/focusable-list/focusable-list.directive.ts
+++ b/libs/cdk/utils/directives/focusable-list/focusable-list.directive.ts
@@ -366,7 +366,7 @@ export class FocusableListDirective implements OnChanges, AfterViewInit, OnDestr
                 }
 
                 const id = getItemElement(item)?.id ?? null;
-                this.itemFocused.next({ index, id, total: items.length });
+                this.itemFocused.next({ index, total: items.length, id });
                 this._focusableItems.forEach((i) => i.setTabbable(i === directiveItem));
                 this._keyManager?.setActiveItem(index);
             })
--- a/libs/cdk/utils/directives/focusable-list/focusable-list.directive.ts
+++ b/libs/cdk/utils/directives/focusable-list/focusable-list.directive.ts
@@ -40,8 +40,8 @@ import { ScrollPosition, scrollIntoView } from './scroll';
 
 export interface FocusableListItemFocusedEvent {
     index: number;
-    id: string | null;
     total: number;
+    id?: string | null;
 }
 
 export interface FocusableListKeydownEvent {
--- a/libs/core/segmented-button/segmented-button.component.ts
+++ b/libs/core/segmented-button/segmented-button.component.ts
@@ -248,10 +248,6 @@ export class SegmentedButtonComponent implements AfterViewInit, ControlValueAcce
         buttonComponent.elementRef.nativeElement.setAttribute('aria-roledescription', 'Segmented button');
         buttonComponent.elementRef.nativeElement.setAttribute('aria-posinset', String(index + 1));
         buttonComponent.elementRef.nativeElement.setAttribute('aria-setsize', String(this._buttons.length));
-        buttonComponent.elementRef.nativeElement.setAttribute(
-            'aria-selected',
-            String(this._isButtonSelected(buttonComponent))
-        );
     }
 
     /** @hidden */
--- a/libs/core/segmented-button/segmented-button.component.ts
+++ b/libs/core/segmented-button/segmented-button.component.ts
@@ -8,6 +8,7 @@ import {
     DestroyRef,
     ElementRef,
     Host,
+    HostBinding,
     HostListener,
     Input,
     NgZone,
@@ -38,6 +39,8 @@ import { EMPTY, Subject, asyncScheduler, fromEvent, merge } from 'rxjs';
 import { filter, observeOn, startWith, takeUntil, tap } from 'rxjs/operators';
 
 export type SegmentedButtonValue = string | (string | null)[] | null;
+
+let segmentedButtonUniqueId = 0;
 
 /**
  * Container for grouped buttons.
@@ -87,6 +90,12 @@ export class SegmentedButtonComponent implements OnInit, AfterViewInit, ControlV
     vertical = false;
 
     /** @hidden */
+    @HostBinding('attr.aria-activedescendant')
+    private get _ariaActiveDescendant(): string | null {
+        return this._focusedItemId;
+    }
+
+    /** @hidden */
     @ContentChildren(FD_BUTTON_COMPONENT)
     _buttons: QueryList<ButtonComponent>;
 
@@ -122,6 +131,9 @@ export class SegmentedButtonComponent implements OnInit, AfterViewInit, ControlV
     private _translationResolver = inject(TranslationResolver);
 
     /** @hidden */
+    private _focusedItemId: string | null = null;
+
+    /** @hidden */
     constructor(
         private readonly _changeDetRef: ChangeDetectorRef,
         private readonly _elementRef: ElementRef,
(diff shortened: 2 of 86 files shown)


libs/core/segmented-button/segmented-button.component.ts
```diff
@@ -54,7 +54,10 @@ export type SegmentedButtonValue = string | (string | null)[] | null;
     host: {
         role: 'listbox',
         '[class.fd-segmented-button]': 'true',
-        '[class.fd-segmented-button--vertical]': 'vertical'
+        '[class.fd-segmented-button--vertical]': 'vertical',
+        '[attr.aria-multiselectable]': 'toggle',
+        '[attr.aria-orientation]': 'vertical ? "vertical" : "horizontal"',
+        '[attr.aria-roledescription]': '"Segmented button group"'
```
mikerodonnell89: the "Segmented button group" string should be prepared for translation

MariaIDineva: Alright, good catch, thank you!

InnaAtanasova (commented):
Screenshot 2025-09-25 at 1 27 50 PM

1. The buttons that have only icons should have the `title` attribute and `aria-label` to provide some description. For example, the cart button should have something like "Add to cart" title (it will appear as a tooltip on hover). This should be updated in the examples.
2. The button elements shouldn't have `aria-pressed`. Although these are button elements, we overwrite their role and they have role="option", and `aria-pressed` can be applied only on button elements: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-pressed
3. Since the parent element has role listbox, you need to add aria-activedescendant to point to the active (focussed) option.

libs/core/segmented-button/segmented-button.component.ts
```diff
@@ -80,6 +89,12 @@ export class SegmentedButtonComponent implements AfterViewInit, ControlValueAcce
     @Input({ transform: booleanAttribute })
     vertical = false;
 
+    /** @hidden */
+    @HostBinding('attr.aria-activedescendant')
+    private get _ariaActiveDescendant(): string | null {
```
InnaAtanasova: I don't see `_ariaActiveDescendant` being used anywhere

libs/core/segmented-button/segmented-button.component.ts
```diff
@@ -103,6 +124,15 @@ export class SegmentedButtonComponent implements AfterViewInit, ControlValueAcce
...
     private readonly _onDestroy$ = new Subject<void>();
 
+    /** @hidden */
+    private readonly _lang$ = inject(FD_LANGUAGE);
+
+    /** @hidden */
+    private _translationResolver = inject(TranslationResolver);
+
+    /** @hidden */
+    private _focusedItemId: string | null = null;
```
InnaAtanasova: If you make this a signal you probably won't need to have a getter for the host binding. 
Your '[attr.aria-activedescendant]': '_focusedItemId()' can go in the 
```
host: {
        role: 'listbox',
        '[class.fd-segmented-button]': 'true',
        '[class.fd-segmented-button--vertical]': 'vertical',
        '[attr.aria-multiselectable]': 'toggle',
        '[attr.aria-orientation]': 'vertical ? "vertical" : "horizontal"',
        '[attr.aria-roledescription]': '_groupRoleDescription',
        '[attr.aria-activedescendant]': '_focusedItemId()'
    },
```

libs/core/segmented-button/segmented-button.component.ts
```diff
@@ -116,6 +146,10 @@ export class SegmentedButtonComponent implements AfterViewInit, ControlValueAcce
         effect(() => {
             this._focusableList.contentDirection = this._rtlService?.rtlSignal() ? 'rtl' : 'ltr';
         });
+        this._focusableList.itemFocused.pipe(takeUntil(this._onDestroy$)).subscribe((item) => {
+            this._focusedItemId = item.id;
+            this._changeDetRef.markForCheck();
```
InnaAtanasova: maybe this also won't be needed if you convert _focusedItemId to a signal.

libs/core/segmented-button/segmented-button.component.ts
```diff
@@ -195,6 +246,18 @@ export class SegmentedButtonComponent implements AfterViewInit, ControlValueAcce
...
+        if (!this._buttons) {
+            return;
+        }
+        this._buttons.forEach((button) => {
+            const element = button.elementRef.nativeElement;
+            if (!element.id) {
+                element.id = `fd-segmented-button-${segmentedButtonUniqueId++}`;
+            }
+        });
+    }
```
InnaAtanasova: While this solution works and looks ok, I would prefer to have the default assignment of the IDs in button component. Button can have an input for ID and if not provided, a default should be assigned. Then here we can just read the IDs. 
My concern is that when we do a11y improvements for other components we might face the same case where we need the button ID and then we will repeat the logic of accessing the native element and assign a default id.

libs/core/segmented-button/segmented-button.component.ts
```diff
@@ -239,6 +302,23 @@ export class SegmentedButtonComponent implements AfterViewInit, ControlValueAcce
         triggerEvents$.pipe(takeUntil(refresh$)).subscribe(() => this._handleTriggerOnButton(buttonComponent));
     }
 
+    /** @hidden */
+    private _setButtonAttributes(buttonComponent: ButtonComponent, index: number): void {
+        buttonComponent.elementRef.nativeElement.setAttribute('role', 'option');
+        buttonComponent.elementRef.nativeElement.setAttribute('aria-roledescription', this._buttonRoleDescription);
+        buttonComponent.elementRef.nativeElement.setAttribute('aria-posinset', String(index + 1));
+        buttonComponent.elementRef.nativeElement.setAttribute('aria-setsize', String(this._buttons.length));
```
InnaAtanasova: For setting the attributes I think it's better to use the Renderer2. Something like:
```
private _setButtonAttributes(buttonComponent: ButtonComponent, index: number): void {
    const el = buttonComponent.elementRef.nativeElement;

    this.renderer.setAttribute(el, 'role', 'option');
    this.renderer.setAttribute(el, 'aria-roledescription', this._buttonRoleDescription);
    this.renderer.setAttribute(el, 'aria-posinset', String(index + 1));
    this.renderer.setAttribute(el, 'aria-setsize', String(this._buttons.length));
}
```

OR you can extract the attributes in a Record and have something like:
```
 private _setButtonAttributes(buttonComponent: ButtonComponent, index: number): void {
        const el = buttonComponent.elementRef.nativeElement;

        const attributes: Record<string, string> = {
            role: 'option',
            'aria-roledescription': this._buttonRoleDescription,
            'aria-posinset': String(index + 1),
            'aria-setsize': String(this._buttons.length)
        };

        for (const [key, value] of Object.entries(attributes)) {
            this.renderer.setAttribute(el, key, value);
        }
    }
```

With Renderer2 it will work in all environments.

libs/core/segmented-button/segmented-button.component.ts
```diff
@@ -239,6 +302,23 @@ export class SegmentedButtonComponent implements AfterViewInit, ControlValueAcce
...
+    }
+
+    private _updateButtonRoleDescriptions(): void {
+        if (!this._buttons) {
+            return;
+        }
+        this._buttons.forEach((button) => {
+            button.elementRef.nativeElement.setAttribute('aria-roledescription', this._buttonRoleDescription);
```
InnaAtanasova: same here, should use Renderer2

libs/core/segmented-button/segmented-button.component.ts
MariaIDineva: I mean, it's a host binding with a getter so Angular will automatically call the getter upon each change detection cycle and the attribute on the host will automatically get updated in case of a change in this._focusedItemId.

libs/core/segmented-button/segmented-button.component.ts
MariaIDineva: That's a good idea, thanks!

libs/core/segmented-button/segmented-button.component.ts
MariaIDineva: Good point, I will refactor.

InnaAtanasova (approved):

libs/cdk/utils/directives/focusable-list/focusable-list.directive.ts
```diff
@@ -40,6 +40,7 @@ import { ScrollPosition, scrollIntoView } from './scroll';
 
 export interface FocusableListItemFocusedEvent {
     index: number;
+    id: string | null;
```
mikerodonnell89: While I doubt this event is commonly dealt with by our users, this is a breaking change, which I think can be avoided by making the id the final property of this event and also optional (`?`)

mikerodonnell89 (approved):

Issue #13490, opened 2025-09-11
Adopt the Breaking Changes from Fundamental-Styles


InnaAtanasova:
The following components have been updated in Fundamental-styles and some of them have breaking changes that need to be adopted in Fundamental-NGX. 

- [x] Action Bar - #13713
- [x] Avatar - #13812
- [ ] Avatar Group
- [ ] Calendar
- [ ] Checkbox
- [ ] Field Set
- [ ] Form Grid
- [ ] Form Group
- [ ] Form Header
- [x] Form Item - #14045
- [ ] Form Label
- [x] Form Message - #13616
- [x] Input - #13506
- [x] Input Group - #13351, #13385
- [ ] Radio
- [x] Textarea - #13506
- [x] Generic Tag - #13885
- [ ] Icons
- [x] Illustrated Messages - #13823, #13851, #13678
- [x] Info Label - #13811, #13832
- [x] Link - #13536
- [x] Message Strip - #13514
- [ ] Message Toast
- [x] Object Identifier - #13532, #13731
- [x] Object Marker - #13808
- [x] Object Number - #13529, #13719
- [x] Object Status - #13527
- [x] Text - #13696
- [ ] Title

droshev:
## Status Update — Analysis of Remaining Components

I've analyzed all unchecked components against the current fundamental-styles markup and the current state of fundamental-ngx. Here's the breakdown:

---

### Already Done (can be checked off)

These were completed as part of other PRs or require no code changes:

| Component | Status | Details |
|-----------|--------|---------|
| **Form Group** | Done in #14102 (`86304239bc`, Apr 8 2026) | `fd-form-group__header-text--wrap` class binding added, form-group-header migrated to signals |
| **Form Label** | Done in #14102 (`86304239bc`, Apr 8 2026) | Class name corrected from `fd-form-label--allow-wrap` to `fd-form-label--wrap` |
| **Form Header** | No code changes needed | Only CSS-only changes in fundamental-styles (height `2.75rem` → `var(--sapElement_LineHeight)`, font-size `var(--sapFontHeader4Size)` → `var(--sapGroup_Title_FontSize)`). These take effect automatically when fundamental-styles is upgraded. No template or TypeScript changes required. |
| **Field Set** | No code changes needed | No breaking changes in fundamental-styles for fieldset. The ngx component correctly applies `fd-fieldset` and `fd-fieldset__legend` classes. |
[...]

PR #13506, opened 2025-09-24
fix(core): adopt a11y breaking changes for Textarea and Input states

InnaAtanasova:
## Related Issue(s)

## Description
This changes concern ONLY textarea and input with directive `fd-form-control` (input[fd-form-control], textarea[fd-form-control])
Combobox, DatePicker, etc. belong to Input Group and there the logic for handling states will be different.

### DIFF
--- a/libs/core/form/form-control/form-control.component.ts
+++ b/libs/core/form/form-control/form-control.component.ts
@@ -1,85 +1,177 @@
 import {
-    Attribute,
     ChangeDetectionStrategy,
     Component,
+    computed,
+    effect,
     ElementRef,
-    HostBinding,
+    inject,
+    input,
     Input,
     OnChanges,
     OnDestroy,
     OnInit,
+    Renderer2,
+    signal,
     ViewEncapsulation
 } from '@angular/core';
 import { FormStates } from '@fundamental-ngx/cdk/forms';
-import { CssClassBuilder, Nullable, applyCssClass } from '@fundamental-ngx/cdk/utils';
+import { applyCssClass, CssClassBuilder, Nullable } from '@fundamental-ngx/cdk/utils';
 import { ContentDensityObserver, contentDensityObserverProviders } from '@fundamental-ngx/core/content-density';
+import { ValueStateAriaMessageService } from '@fundamental-ngx/core/shared';
 import { Subscription } from 'rxjs';
 import { FormItemControl, registerFormItemControl } from '../form-item-control/form-item-control';
 
-/**
- * Directive intended for use on form controls.
- *
- * ```html
- * <input type="text" fd-form-control />
- * ```
- */
+let formControlId = 0;
+
 @Component({
     // eslint-disable-next-line @angular-eslint/component-selector
     selector: 'input[fd-form-control], textarea[fd-form-control]',
-    template: ` <ng-content></ng-content>`,
+    template: `<ng-content></ng-content>`,
     styleUrl: './form-control.component.scss',
     encapsulation: ViewEncapsulation.None,
     changeDetection: ChangeDetectionStrategy.OnPush,
-    providers: [registerFormItemControl(FormControlComponent), contentDensityObserverProviders()]
+    providers: [registerFormItemControl(FormControlComponent), contentDensityObserverProviders()],
+    host: {
+        '[attr.type]': 'type()',
+        '[attr.aria-label]': 'ariaLabel',
+        '[attr.aria-labelledby]': 'ariaLabelledBy',
+        '[attr.aria-describedby]': 'combinedAriaDescribedBy()'
+    }
 })
 export class FormControlComponent implements CssClassBuilder, OnInit, OnChanges, OnDestroy, FormItemControl {
-    /**
-     *  The state of the form control - applies css classes.
-     *  Can be `success`, `error`, `warning`, `information` or blank for default.
-     */
+    /** aria-label for form-control. */
     @Input()
-    state: FormStates | null = null;
-
-    /** Type of the form control. */
-    @HostBinding('attr.type')
-    @Input()
-    type: string;
-
-    /** user's custom classes */
-    @Input()
-    class: string;
+    ariaLabelledBy: Nullable<string>;
 
     /** aria-label for form-control. */
     @Input()
     ariaLabel: Nullable<string>;
 
-    /** aria-label for form-control. */
-    @Input()
-    ariaLabelledBy: Nullable<string>;
+    /**
+     *  The state of the form control - applies css classes.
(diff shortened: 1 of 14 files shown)


libs/core/form/form-control/form-control.component.ts
```diff
@@ -1,85 +1,176 @@
...
+        success: this.valueStateSuccessMessage,
+        information: this.valueStateInformationMessage,
+        warning: this.valueStateWarningMessage,
+        error: this.valueStateErrorMessage
+    } as const;
+
+    /** @hidden */
+    private _currentValueStateMessage = computed(() => {
+        const st = this.state();
+        return st ? this._valueStateMessages[st]() : '';
```
MariaIDineva: Should those messages be translated?

libs/core/form/form-control/form-control.component.ts
```diff
@@ -1,85 +1,176 @@
...
+    host: {
+        '[attr.type]': 'type()',
+        '[attr.aria-label]': 'ariaLabel',
+        '[attr.aria-labelledby]': 'ariaLabelledBy',
+        '[attr.aria-describedby]': 'combinedAriaDescribedBy()'
+    }
 })
 export class FormControlComponent implements CssClassBuilder, OnInit, OnChanges, OnDestroy, FormItemControl {
+    /** aria-label for form-control. */
+    @Input()
```
mikerodonnell89, line 43: Why `@Input()` instead of signals for this?

libs/core/form/form-control/form-control.component.ts
```diff
@@ -1,85 +1,176 @@
...
-    @Input()
-    class: string;
+    /** User's custom classes */
+    class = input<string>();
 
-    /** aria-label for form-control. */
-    @Input()
-    ariaLabel: Nullable<string>;
+    /** Default ARIA message text for the "success" value state. */
+    valueStateSuccessMessage = input<string>(inject(ValueStateAriaMessageService).success);
```
mikerodonnell89, line 63: Not the responsibility of this PR, but I'm going to open an issue to move away from this weird way of getting these success/info/warning/error messages. These won't be translated automatically

mikerodonnell89: https://github.com/SAP/fundamental-ngx/issues/13516

libs/core/form/form-control/form-control.component.ts
```diff
@@ -1,85 +1,176 @@
...
     /** @hidden */
-    constructor(
-        public elementRef: ElementRef<HTMLInputElement | HTMLTextAreaElement>,
-        _contentDensityObserver: ContentDensityObserver,
-        @Attribute('aria-label') private ariaLabelAttr: string,
-        @Attribute('aria-labelledby') private ariaLabelledByAttr: string
-    ) {
-        _contentDensityObserver.subscribe();
+    constructor() {
+        this._subscriptions.add(this.contentDensityObserver.subscribe());
```
mikerodonnell89, line 164: Confused by this - I understand it's subscribing to content density changes but is it doing anything with that subscription?

libs/core/form/form-control/form-control.component.ts
```diff
@@ -92,15 +183,38 @@ export class FormControlComponent implements CssClassBuilder, OnInit, OnChanges,
...
         const tagName = this.elementRef.nativeElement.tagName.toLowerCase();
         return [
-            this.state ? 'is-' + this.state : '',
-            this.class,
-            tagName === 'textarea' ? 'fd-textarea' : tagName === 'input' ? 'fd-input' : ''
-        ];
+            tagName === 'textarea' ? 'fd-textarea' : tagName === 'input' ? 'fd-input' : '',
+            this.state() ? `is-${this.state()}` : '',
+            this.class()
+        ].filter(Boolean) as string[];
```
mikerodonnell89, line 190: what is this line doing?

mikerodonnell89 (commented):
some questions above

libs/core/form/form-control/form-control.component.ts
InnaAtanasova: sorry, I explained this during the scrum, but you prob were still on leave. These 2 inputs are used in 8+ other components and I would need to go refactor their logic if I migrate this to a signal. We discussed this to be done gradually as this PR will touch 10 components just for this.

libs/core/form/form-control/form-control.component.ts
InnaAtanasova: we had this:
```
[
  tagName === 'textarea' ? 'fd-textarea' : tagName === 'input' ? 'fd-input' : '',
  this.state() ? `is-${this.state()}` : '',
  this.class()
]
```
if a certain condition doesn't match, the fallback is an empty string ''. `.filter(Boolean)` will remove empty strings, null, undefined, false, 0. In this case, it mainly strips out the '' from when no class applies. Since Boolean isn’t a strict type guard we can get `(string | undefined | null)[]`, that's why we need to make sure we get an array of strings with `as string[]`.

Example:
```
['fd-input', '', 'extra-class'].filter(Boolean)
```

will result in:
```
['fd-input', 'extra-class']
```

libs/core/form/form-control/form-control.component.ts
InnaAtanasova: This is not in the scope of this PR. The translation should be coming from the service as these messages will be reused across multiple controls.

libs/core/form/form-control/form-control.component.ts
InnaAtanasova: I am not sure why it was added. I just noticed that we were not unsubscribing from it, so I added this logic.

mikerodonnell89 (changes requested):
build step failing on a unit test: https://github.com/SAP/fundamental-ngx/actions/runs/18200678351/job/51818480659?pr=13506#step:9:11602

Maybe doesn't matter since this is a temp base branch?

InnaAtanasova:
> build step failing on a unit test: https://github.com/SAP/fundamental-ngx/actions/runs/18200678351/job/51818480659?pr=13506#step:9:11602
> 
> Maybe doesn't matter since this is a temp base branch?

Fixed the failing tests. Turns out we were adding some CSS classes to the form label that were not needed and actually was wrong. 
https://github.com/SAP/fundamental-ngx/pull/13506/files#diff-237c9468131565c2056398fe6be4686df341306bc85b393cc726d83d88ffec0bL38
The classes `fd-form-label__wrapper--inline-help` and `fd-form-label__wrapper--inline-help--after` were added to Quick View Label even if the label is not used as a trigger for Inline Help.

mikerodonnell89 (approved):

PR #13510, opened 2025-09-26
fix(core): shellbar context area should allow ngIf

### LINKED ISSUE #13507: [fd-shellbar] While using *ngIf or @if inside of "fd-shellbar-context-area" errors thrown
### Describe your proposal

Whenever using angular *ngIf or @if syntax inside the fd-shellbar-context-area component, which leaves a comment in html, it will throw error and shellbar will not finish rendering.
Apparently it is trying to add display style on the comment as an element in the list of context area children.

Image
Image
Image
Image
Image
Image
Image

Angular: latest, Fundamental NGX: latest

Open any of the shellbar stackblitz examples and just add *ngIf="true" or use control flow syntax @if(true){} on one of child components inside of the fd-shellbar-context-area component.

### StackBlitz example(s)

https://stackblitz.com/edit/diqyyfcj

### Declaration

mikerodonnell89:
fixes #13507 

Also adds an event `contentItemVisibilityChange` that is fired when the context area has items that are either hidden or shown when compared to the previous state. The event passes the context area's hidden items as its parameter

### DIFF
--- a/libs/core/shellbar/shellbar-context-area/shellbar-context-area.component.ts
+++ b/libs/core/shellbar/shellbar-context-area/shellbar-context-area.component.ts
@@ -1,4 +1,6 @@
-import { ChangeDetectionStrategy, Component, ElementRef, inject } from '@angular/core';
+import { AfterViewInit, ChangeDetectionStrategy, Component, DestroyRef, ElementRef, inject } from '@angular/core';
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+import { ResizeObserverService } from '@fundamental-ngx/cdk/utils';
 import { FD_SHELLBAR_COMPONENT } from '../tokens';
 
 /**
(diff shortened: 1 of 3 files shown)


MariaIDineva (approved):

libs/core/shellbar/shellbar-context-area/shellbar-context-area.component.ts
```diff
@@ -48,8 +93,10 @@ export class ShellbarContextAreaComponent {
      * to exceed the end of th eshellbar.
      */
     showElements(): void {
-        this._getElementsWithPriority().forEach((el) => {
-            el.el.style.display = '';
+        this._getContextAreaItemsWithPriority().forEach((item) => {
+            if (item?.el?.style) {
```
MariaIDineva, line 95: Just one question, out of curiosity - this `if` clause was not present before the change, was it added with a particular purpose as adding optional chaining like so `item?.el?.style?.display = '';` should also fix the reported issue?

mikerodonnell89: That syntax throws the error `The left-hand side of an assignment expression may not be an optional property access`

libs/core/shellbar/shellbar-context-area/shellbar-context-area.component.ts
```diff
@@ -22,24 +34,57 @@ import { FD_SHELLBAR_COMPONENT } from '../tokens';
         `
     ]
 })
-export class ShellbarContextAreaComponent {
+export class ShellbarContextAreaComponent implements AfterViewInit {
+    /** Event emitted when items in the context area are hidden or shown. Event parameter is an array of all the items that are hidden. */
+    @Output()
+    contentItemVisibilityChange = new EventEmitter<HTMLElement[]>();
```
InnaAtanasova: Can you make this a signal output?

libs/core/shellbar/shellbar-context-area/shellbar-context-area.component.ts
```diff
@@ -22,24 +34,57 @@ import { FD_SHELLBAR_COMPONENT } from '../tokens';
...
+    private _resizeObserverService = inject(ResizeObserverService);
+
+    /** @hidden */
+    private readonly _destroyRef = inject(DestroyRef);
+
+    /** @hidden */
+    private _hiddenItems: HTMLElement[];
+
     /** @hidden */
     constructor(public el: ElementRef) {}
```
InnaAtanasova: Can you inject this with `inject`?

InnaAtanasova (approved):

PR #13511, opened 2025-09-29
[WIP] feat: bring ui5wc generator

droshev:
implementation of the ui5wc wrapper and the 4 packages


libs/ui5-webcomponents-base/theming/base-webcomponents-theming-provider.ts
```diff
@@ -0,0 +1,45 @@
...
+import supportedThemes from './supported-themes';
+import { setTheme } from '@ui5/webcomponents-base/dist/config/Theme.js';
+
+@Injectable()
+export abstract class WebcomponentsThemingProvider implements Ui5ThemingProvider, OnDestroy {
+  abstract name: string;
+
+  protected availableThemes = signal<string[]>(supportedThemes);
+
+  protected _globalThemingService: Ui5ThemingService | null = inject(Ui5ThemingService, { optional: true });
```
InnaAtanasova, line 49: Maybe add some typedocs for these

libs/ui5-webcomponents-base/theming/base-webcomponents-theming-provider.ts
```diff
@@ -0,0 +1,45 @@
...
+    return this.availableThemes();
+  }
+
+  async setTheme(theme: string): Promise<boolean> {
+    const supported = this.availableThemes().includes(theme);
+    if (supported) {
+      setTheme(theme);
+    }
+    return supported;
+  }
```
InnaAtanasova, line 100: same

nnaydenow:
- [x] Generated output (events) have `CustomEvent<any>` as a type. Not sure how the events will be handled in Angular world, but from web components side there are types that describe the event details.  (e.g. `ui5overflow = output<UI5CustomEvent<_AvatarGroup, 'overflow'>>();` and test it in the angular app_

- [x] When synchronizing outputs (events) you are striping the `ui5` world from event name `ui5selectionChange` => `selectionChange`. Since some events are conflicting with the native one (f.e `click`) we provide event in conflict and no conflict mode. This means that `click` is fired as `click` and `ui5-click`. I think it would be safest to convert the event from `pascalCase` to `kebab-case`. (e.g. `    const outputsToSync = [
      'ui5-click',
      'ui5-overflow',
    ];`

- [x] For enumarations I see that you generate `types.ts` file but I think that you can extract the type directly from the web components one. For example `UploadCollectionItem` has `uploadState` property. You could read the property from `UploadCollectionItem.uploadState`. I think it will reduce and simpify the import management. Investigate whether we can re-export them from ui5 webC directly

- [x] `libs/ui5-webcomponents-fiori/theming/index.ts` has `name = 'ui-5-webcomponents-theming-service';` ui-5 is written wrong. just to double check if it's expected. I'm not sure how the theming service works exactly but it's strange that fiori's one load the main one, but not its own.
image - `import('@ui5/webcomponents/dist/generated/json-imports/Themes.js'),` has to be `import('@ui5/webcomponents-fiori/dist/generated/json-imports/Themes.js'),` - add additional suffix with the package

- [x] Since you are consuming the final classes (you are not doing anything with the hierarchy of the classes of web components) of the web components I would suggest to filter the produced wrappers only to the classes of `CustomElement` type. These elements has `customElement` boolean property and `tagName` string property.
https://github.com/webcomponents/custom-elements-manifest/blob/main/schema.d.ts#L224

- We need to merge the manifests during the generation since the custom elements analyzer doesn't resolve cross package inheritance and components like `SideNavigationItem` have only 2 inputs, while in the reality they have more. - no issue here

droshev:
> * [x]  Generated output (events) have `CustomEvent<any>` as a type. Not sure how the events will be handled in Angular world, but from web components side there are types that describe the event details.  (e.g. `ui5overflow = output<UI5CustomEvent<_AvatarGroup, 'overflow'>>();` and test it in the angular app_
> * [x]  When synchronizing outputs (events) you are striping the `ui5` world from event name `ui5selectionChange` => `selectionChange`. Since some events are conflicting with the native one (f.e `click`) we provide event in conflict and no conflict mode. This means that `click` is fired as `click` and `ui5-click`. I think it would be safest to convert the event from `pascalCase` to `kebab-case`. (e.g. `    const outputsToSync = [ 'ui5-click', 'ui5-overflow', ];`
> * [x]  For enumarations I see that you generate `types.ts` file but I think that you can extract the type directly from the web components one. For example `UploadCollectionItem` has `uploadState` property. You could read the property from `UploadCollectionItem.uploadState`. I think it will reduce and simpify the import management. Investigate whether we can re-export them from ui5 webC directly
> * [x]  `libs/ui5-webcomponents-fiori/theming/index.ts` has `name = 'ui-5-webcomponents-theming-service';` ui-5 is written wrong. just to double check if it's expected. I'm not sure how the theming service works exactly but it's strange that fiori's one load the main one, but not its own.
>   image - `import('@ui5/webcomponents/dist/generated/json-imports/Themes.js'),` has to be `import('@ui5/webcomponents-fiori/dist/generated/json-imports/Themes.js'),` - add additional suffix with the package
> * [x]  Since you are consuming the final classes (you are not doing anything with the hierarchy of the classes of web components) of the web components I would suggest to filter the produced wrappers only to the classes of `CustomElement` type. These elements has `customElement` boolean property and `tagName` string property.
>   https://github.com/webcomponents/custom-elements-manifest/blob/main/schema.d.ts#L224
> * We need to merge the manifests during the generation since the custom elements analyzer doesn't resolve cross package inheritance and components like `SideNavigationItem` have only 2 inputs, while in the reality they have more. - no issue here

all comments were applied

libs/ui5-webcomponents-base/theming/base-webcomponents-theming-provider.ts
droshev: addressed

libs/ui5-webcomponents-base/theming/base-webcomponents-theming-provider.ts
droshev: addressed

libs/webc-generator/src/executors/generate/component-template.ts
```diff
@@ -0,0 +1,273 @@
...
+
+    return { componentImports, componentEnums: extractedEnums };
+}
+
+/** Helper function to generate input properties for the component. */
+function generateInputs(data: CEM.CustomElementDeclaration, enums: string[], className: string): string {
+    // className added
+    const inputs: string[] = [];
+    (data.members ?? []).filter(isField).forEach((member) => {
+        const typeText = member.type?.text?.replace(' | undefined', '');
```
MariaIDineva: Why remove the `undefined` type?

libs/webc-generator/src/executors/generate/component-template.ts
```diff
@@ -0,0 +1,273 @@
...
+   */
+  ui5${kebabToCamelCase(event.name)} = output<UI5CustomEvent<_${className}, '${event.name}'>>();`);
+    });
+    return outputs.join('\n');
+}
+
+/** Generate the Angular component wrapper. */
+export function componentTemplate(
+    data: CEM.CustomElementDeclaration,
+    cemPackage: CEM.Package,
```
MariaIDineva, line 219: Unused parameter `cemPackage`.

libs/webc-generator/src/executors/generate/component-template.ts
droshev: removed

MariaIDineva (approved):

PR #13514, opened 2025-10-01
fix(core): adopt a11y changes for message strip

MariaIDineva:
## Related Issue(s)

## Description

1. Adopts a11y changes for message strip component. Main changes:
 - A `default` message strip announcement is added - `Message Strip` (along with info, error, warning, success);
 - An announcement is always present regardless if the message strip has an icon or no (they used to be there only when an icon is present) and is placed as a sibling to `fd-message-strip__text` to follow the structure in fundamental styles;

2. Adds logic to check if an anchor tag is passed to the message strip in which case an additional `fd-message-strip--link` class is added to the host in order to apply the respective styles from fundamental styles.

3. Transforms all `@Input`s into signal `input`s.

4. Updates one of the message strip examples to depict allowing links in message strips.

## Screenshots
Screenshot 2025-10-01 at 16 54 11

### DIFF
--- a/libs/core/message-strip/message-strip.component.ts
+++ b/libs/core/message-strip/message-strip.component.ts
@@ -1,6 +1,6 @@
 import { AsyncPipe, NgTemplateOutlet } from '@angular/common';
 import {
-    AfterContentInit,
+    AfterContentChecked,
     booleanAttribute,
     ChangeDetectionStrategy,
     Component,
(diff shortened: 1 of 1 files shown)


libs/core/message-strip/message-strip.component.ts
```diff
@@ -139,7 +152,7 @@ export class MessageStripComponent implements OnInit, OnChanges, CssClassBuilder
     private _translationResolver = inject(TranslationResolver);
 
     /** @hidden */
-    constructor(public readonly elementRef: ElementRef) {}
+    private _hasProjectedLink = false;
```
InnaAtanasova: If you make this a signal, then in ngAfterContentInit you set it to true if there's a link inside, you won't need the getter in line 73. You can move the binding to he host in the decorator.

libs/core/message-strip/message-strip.component.ts
```diff
@@ -164,22 +158,28 @@ export class MessageStripComponent implements OnInit, OnChanges, CssClassBuilder
         this.setScreenReaderTexts();
     }
 
+    /** @hidden */
+    ngAfterContentInit(): void {
```
mikerodonnell89: @InnaAtanasova do you think this is safe to only check once on init rather than `ngAfterContentChecked`?

InnaAtanasova: Good point! `ngAfterContentInit` runs once after Angular projects external content into the component. This means that if the link is passed with conditions or dynamically will not be taken into consideration. `ngAfterContentChecked` runs on every change detection cycle, so it would catch cases where the projected content changes over time. But it’s much more expensive and usually unnecessary unless the DOM really can change dynamically.

PR #13518, opened 2025-10-06
fix(core): adopt latest changes for user menu

InnaAtanasova:
## Description
- modifier classes for text truncation
- popover header on scroll when user name is no longer visible
- updated tests
- examples update
IMPORTANT: brought some CSS from fund-styles as we can't release fund-styles not and can't use the RC version

## Screenshots
User name is visible (no heading)
Screenshot 2025-10-06 at 2 58 20 PM

User name is not visible (heading)
Screenshot 2025-10-06 at 2 58 10 PM


libs/core/list/directives/list-title.directive.ts
```diff
@@ -1,18 +1,20 @@
...
 export class ListTitleDirective implements OnInit {
-    /** @hidden */
-    @HostBinding('class.fd-list__title')
-    fdListTitleClass = true;
-
     /** Whether or not this should be wrapped, when too much text. */
-    @Input()
-    @HostBinding('class.fd-list__title--wrap')
-    wrap = false;
+    wrap = input(false, { transform: booleanAttribute });
```
mikerodonnell89, line 16: So I know you're just updating this to an input signal. And I know `wrap` defaults to false and doesn't apply a class. But the default behavior is to wrap. Changing the value of this input doesn't have any actual affect on the text as far as I can tell. Should we deprecate this input?

InnaAtanasova: good point! I added to typedocs the deprecated tag and removed the binding. Will leave the input for now to not break apps where it's used.

mikerodonnell89 (approved):

PR #13521, opened 2025-10-08
fix(core): update Avatar

InnaAtanasova:
## Related Issue(s)

## Description
- adopt a11y changes from styles: there's a default aria-label, label and alt attr passed by the user are reflected in the aria-label
- migrate Avatar implementation to signals
- refactored the implementation for the alternative options. 
Before: 
a) content options was not rendered (first avatar) 
b) alt option was not rendered

Screenshot 2025-10-08 at 11 16 09 AM

After:
Screenshot 2025-10-08 at 11 15 48 AM

### DIFF
--- a/libs/core/avatar/avatar.component.ts
+++ b/libs/core/avatar/avatar.component.ts
@@ -1,30 +1,36 @@
 import {
     Attribute,
+    booleanAttribute,
     ChangeDetectionStrategy,
     ChangeDetectorRef,
     Component,
+    computed,
+    DestroyRef,
+    effect,
     ElementRef,
-    EventEmitter,
-    HostBinding,
     HostListener,
-    Input,
+    inject,
+    input,
     OnChanges,
     OnInit,
-    Output,
+    output,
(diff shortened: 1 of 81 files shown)


libs/core/avatar/avatar.component.ts
```diff
@@ -53,191 +59,198 @@ export type IndicationColor = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10;
...
     /** @hidden */
     constructor(
         public readonly elementRef: ElementRef,
         private readonly _cdr: ChangeDetectorRef,
         private readonly _renderer: Renderer2,
         @Attribute('tabindex') private hostTabindex: number | null
-    ) {}
+    ) {
+        effect(() => {
+            this._glyph.set(this.glyph());
```
MariaIDineva, line 244: Just to confirm my understanding - is that the way to transform class fields into signals and that should be the pattern that we will follow going forward for the rest of the components?

PR #13524, opened 2025-10-09
fix(core): User Menu updates after design review

InnaAtanasova:
Fixes the following issues:
**Long string truncation:**
Screenshot 2025-10-10 at 4 50 42 PM

Screenshot 2025-10-10 at 4 50 48 PM

**Flickering on mobile mode**
Previously, when scrolling on mobile, the header with the title flickered due to being conditionally rendered with *ngIf.
The implementation has been refactored to keep the header always in the DOM and toggle its visibility using conditional opacity instead. This eliminates the flicker while maintaining the same visual behavior.

**Mobile mode header on scroll**
In mobile mode we have the same behaviour on scroll, when the title disappears, it should be shown in the header

**List focus bug**
Fixed an issue where pressing the arrow keys while focused on the first account item would jump directly to the menu items below.
Now, arrow key navigation correctly cycles through the account list items before moving to the next section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.57.4-rc.27](https://github.com/SAP/fundamental-ngx/compare/v0.57.4-rc.26...v0.57.4-rc.27) (2025-10-16)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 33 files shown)


libs/core/user-menu/components/user-menu-body.component.html
```diff
@@ -8,7 +8,7 @@
 }
 
 <ng-template #bodyHeader let-dialogRef="dialogRef">
-    <div fd-bar [class.fd-user-menu__main-view-bar]="!submenu()">
+    <div fd-bar style="position: sticky; top: 0; z-index: 6">
```
MariaIDineva, line 11: Is that style inline on purpose or shall we create a class for it instead? Same question for the nested div below?

droshev (approved):

PR #13527, opened 2025-10-14
fix(core): adopt a11y breaking changes for Object Status

MariaIDineva:
## Related Issue(s)

## Description
Adopts a11y breaking changes from [Fundamental Styles](https://github.com/SAP/fundamental-styles/pull/6066) for `Object Status`:
- Adds `role="button"` and `aria-roledescription` to Object Status when it's clickable.
- Adds a new `ariaRoleDescription` input which the user can use to override the default `aria-roledescription` for clickable items.
- Adds a screen reader text to `non-clickable` elements:
```
<span class="fd-object-status__sr-only">Object Status</span>
```
- Adds an indication color screen reader text when an `indication color` is provided:
```
<span class="fd-object-status__sr-only">Indication Color 2</span>
```
- Adds a status screen reader text when `status` is provided -  `'negative' | 'critical' | 'positive' | 'informative'`:
```
<span class="fd-object-status__sr-only">Informative entry</span>
```
- Adds translations for all screen reader texts.
- Transforms all inputs into signals.
- Updates core examples:
  - Removes `aria-label`.
  - Puts all clickable elements inside a `div` instead of an `a` tag.

### DIFF
--- a/libs/i18n/src/lib/models/fd-language.ts
+++ b/libs/i18n/src/lib/models/fd-language.ts
@@ -169,6 +169,10 @@ export interface FdLanguage {
         announcement: FdLanguageKey;
         ariaRoleDescription: FdLanguageKey;
         indicationColor: FdLanguageKey;
+        negative: FdLanguageKey;
+        critical: FdLanguageKey;
+        positive: FdLanguageKey;
+        informative: FdLanguageKey;
     };
     coreOverflowLayout: {
         /** @param count */


libs/core/object-status/object-status.component.ts
```diff
@@ -49,81 +43,104 @@ export type ObjectStatus = 'negative' | 'critical' | 'positive' | 'informative'
...
+        '[attr.aria-roledescription]': 'clickable() ? _ariaRoleDescription() : null'
     },
-    imports: [IconComponent, NgTemplateOutlet]
+    imports: [IconComponent, NgTemplateOutlet, FdTranslatePipe]
 })
 export class ObjectStatusComponent implements OnChanges, OnInit, CssClassBuilder {
     /** User's custom classes */
-    @Input()
-    class: string;
+    class = input<string | undefined>('');
```
InnaAtanasova: `input<Nullable<string>>`

libs/core/object-status/object-status.component.html
```diff
@@ -0,0 +1,29 @@
...
+@if (!clickable()) {
+    <span class="fd-object-status__sr-only">{{ 'coreObjectStatus.announcement' | fdTranslate }}</span>
+}
+@if (indicationColor()) {
+    <span class="fd-object-status__sr-only">{{
+        ('coreObjectStatus.indicationColor' | fdTranslate) + _indicationColorCode()
+    }}</span>
+}
+@if (status()) {
+    <span class="fd-object-status__sr-only">{{ _statusTranslateKey() | fdTranslate }}</span>
```
InnaAtanasova: We should have an input property that would allow the user to set their own status message and if they don't set it we default to these ones.

libs/core/object-status/object-status.component.ts
```diff
@@ -49,81 +43,104 @@ export type ObjectStatus = 'negative' | 'critical' | 'positive' | 'informative'
...
+    });
+
     /** Whether the Object status is icon-only. */
     get iconOnly(): boolean {
-        return !this.label && !this.textTemplate;
+        return !this.label() && !this.textTemplate();
     }
 
+    /** @hidden */
+    private readonly _ariaRoleDescription = signal<string>('');
```
InnaAtanasova, line 139: We should also have an input ariaRoleDescription in case the user wants to set a custom one. We should default to this one.

PR #13529, opened 2025-10-15
feat(core): adopt breaking changes for Object Number

MariaIDineva:
## Related Issue(s)

## Description

Adopts breaking changes coming from [Fundamental Styles](https://github.com/SAP/fundamental-styles/pull/6082) for `ObjectNumber`:
- Adds a screen reader text for the different states/statuses - `'negative' | 'critical' | 'positive' | 'informative'`:
```
<fd-object-number unit="EUR" status="negative" class="fd-object-number fd-object-number--negative fd-object-number--interactive" tabindex="0" role="button">
  <span class="fd-object-number__text> -2,000.00</span>
  <span class="fd-object-number__unit">EUR</span>
  <span class="fd-object-number__sr-only ng-star-inserted">Value State Error</span>
</fd-object-number>
```

- Adds a screen reader text when the Object Number is emphasized:
```
<fd-object-number unit="EUR" status="negative" class="fd-object-number fd-object-number--negative>
  <span class="fd-object-number__text fd-object-number__text--bold"> -2,000.00</span>
  <span class="fd-object-number__unit">EUR</span>
  <span class="fd-object-number__sr-only ng-star-inserted">Emphasized</span>
</fd-object-number>
```
- Adds two new states to Object Number - **interactive** and **inverted** which can also be combined. Updates the examples to display those.
- Adds `role="button"` and `tabindex=0` to interactive object numbers.
- Transforms all Object Number inputs to signals.

## Screenshots

Screenshot 2025-10-15 at 15 44 40

### DIFF
--- a/libs/core/object-number/object-number.component.html
+++ b/libs/core/object-number/object-number.component.html
@@ -6,5 +6,9 @@
     <span class="fd-object-number__sr-only">{{ 'coreObjectNumber.emphasized' | fdTranslate }}</span>
 }
 @if (status()) {
-    <span class="fd-object-number__sr-only">{{ statusKey() | fdTranslate }}</span>
+    @if (statusMessage()) {
+        <span class="fd-object-number__sr-only">{{ statusMessage() }}</span>
+    } @else {
+        <span class="fd-object-number__sr-only">{{ statusKey() | fdTranslate }}</span>
+    }
 }
--- a/libs/core/object-number/object-number.component.ts
+++ b/libs/core/object-number/object-number.component.ts
@@ -44,13 +44,16 @@ export class ObjectNumberComponent implements OnInit, OnChanges, CssClassBuilder
     unit = input<string>('');
 
     /** Set the value to true to display the object number in bold text */
-    emphasized = input<boolean>(false);
+    emphasized = input(false, { transform: booleanAttribute });
 
     /** Set the value to true to display the object number in large text */
-    large = input<boolean>(false);
+    large = input(false, { transform: booleanAttribute });
 
     /** Sets status/semantic color  'negative' / 'critical' / 'positive' / 'informative' */
     status = input<ObjectStatus | null>(null);
+
+    /** An optional status message for the object number */
+    statusMessage = input<string>();
 
     /** Status key to translate for screen readers */
     statusKey = computed<FdLanguageKeyIdentifier | null>(() => {
@@ -62,7 +65,7 @@ export class ObjectNumberComponent implements OnInit, OnChanges, CssClassBuilder
     });
 
     /** User's custom classes */
-    class = input<string | undefined>('');
+    class = input<string>();
 
     /** Id of the element that labels object number. */
     ariaLabelledBy = input<Nullable<string>>(null);
--- a/libs/core/object-number/object-number.component.spec.ts
+++ b/libs/core/object-number/object-number.component.spec.ts
@@ -49,12 +49,6 @@ describe('ObjectNumberComponent', () => {
         expect(component.elementRef.nativeElement.classList.contains('fd-object-number--large')).toBe(true);
     });
 
-    it('should add status', () => {
-        fixture.componentRef.setInput('status', 'positive');
-        fixture.detectChanges();
-        expect(component.elementRef.nativeElement.classList.contains('fd-object-number--positive')).toBe(true);
-    });
-
     it('should add custom class', () => {
         fixture.componentRef.setInput('class', 'custom-class');
         fixture.detectChanges();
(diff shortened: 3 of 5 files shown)


mikerodonnell89 (commented):
Honestly I don't know how to control this, and need to do some research. But I noticed when using Voiceover and control+option+arrow keys, when the sr-only element is focused, the border appears in the wrong place. Gives the appearance of highlighting the incorrect object status
Screenshot 2025-10-16 at 1 14 39 PM

MariaIDineva:
> Honestly I don't know how to control this, and need to do some research. But I noticed when using Voiceover and control+option+arrow keys, when the sr-only element is focused, the border appears in the wrong place. Gives the appearance of highlighting the incorrect object status

That's an interesting finding, I'll dig into it.

MariaIDineva:
So I played with the issue a bit:
- I tried it with JAWS on Chrome, using insert + arrows to navigate. In JAWS it does not give any focus to the elements and reads them altogether:
```
-2,000.00  EURValue State Error0.00  EURValue State Warning1,000.37  EURValue State Success100.88  EURValue State Information999.00  EUR
```
- I tried it with Voiceover on Safari. The behaviour is similar to the one that JAWS has on Chrome. It focuses the whole group of numbers and reads them all through, without allowing one-by-one navigation:
Screenshot 2025-10-17 at 11 51 13

- I also tried Voiceover on Edge, it works the same was as on Chrome.

- Back to Voiceover on Chrome:
  -  I also tried updating the styles of the screen reader element with what we have in SAPUI5:
```
.fd-object-number__sr-only {
    position: absolute !important;
    clip: rect(1px, 1px, 1px, 1px);
    user-select: none;
    left: 0;
    top: 0;
    font-size: 0;
}
```
But in this case focusing on the sr-element moves the visual focus to the page-content div.
Screenshot 2025-10-17 at 11 53 33

  - Another thing that I tried was adding `left: 0; right: 0`. Then the visual focus seems to be on correct element, but only if its width macthes the width of the sr-element (which depend on the particular text inside and it varies):
  
Screenshot 2025-10-17 at 13 30 00
Screenshot 2025-10-17 at 13 30 43

libs/core/object-number/object-number.component.ts
```diff
@@ -17,53 +20,61 @@ type ObjectStatus = 'negative' | 'critical' | 'positive' | 'informative';
...
 
     /** Sets unit of measure displayed. */
-    @Input()
-    unit: string;
+    unit = input<string>('');
 
     /** Set the value to true to display the object number in bold text */
-    @Input()
-    emphasized = false;
+    emphasized = input<boolean>(false);
```
InnaAtanasova: Can you make all boolean inputs to `{ transform: booleanAttribute }`?

libs/core/object-number/object-number.component.ts
```diff
@@ -17,53 +20,61 @@ type ObjectStatus = 'negative' | 'critical' | 'positive' | 'informative';
...
+        if (this.isValidObjectStatus(status)) {
+            return `coreObjectNumber.${status}`;
+        }
+        return null;
+    });
 
     /** User's custom classes */
-    @Input()
-    class: string;
+    class = input<string | undefined>('');
```
InnaAtanasova: `Nullable<string>`

libs/core/object-number/object-number.component.html
```diff
@@ -1,4 +1,10 @@
...
+    {{ number() | number: (decimal() ? _numberPipeConfig : '1.0-0') }}
 </span>
-<span class="fd-object-number__unit">{{ unit }}</span>
+<span class="fd-object-number__unit">{{ unit() }}</span>
+@if (emphasized()) {
+    <span class="fd-object-number__sr-only">{{ 'coreObjectNumber.emphasized' | fdTranslate }}</span>
+}
+@if (status()) {
+    <span class="fd-object-number__sr-only">{{ statusKey() | fdTranslate }}</span>
+}
```
InnaAtanasova, line 14: We should allow the user to set a custom message. They might not want the status to be "Value State Error", but something else. Maybe add an input property, something like `statusMessage` and if it's not set, default to the one we provide.

InnaAtanasova:
> I noticed when using Voiceover and control+option+arrow keys, when the sr-only element is focused, the border appears in the wrong place. Gives the appearance of highlighting the incorrect object status

Can you try the following options:
1. Completely hide the sr-only element when the SR visually highlights it.
```
.fd-object-number__sr-only:focus {
  display: none;
}
```

2. Change a bit the CSS. Try with clip-path. 
```
.fd-object-number__sr-only {
  position: absolute !important;
  width: 1px;
  height: 1px;
  padding: 0;
  margin: 0;
  overflow: hidden;
  clip: rect(0, 0, 0, 0);
  white-space: nowrap;
  border: 0;
  clip-path: inset(50%);
}
```

3. Your last option would be to make the sr-only element a sibling, not a child of the Object Status that gets the SR focus and link it via aria-labelledby or aria-describedby. The resulting markup should look like:
```
<span 
  class="fd-object-number fd-object-number--negative" 
  aria-describedby="object-status-desc">
    <span class="fd-object-number__text fd-object-number__text--bold">-2,000.00</span>
    <span class="fd-object-number__unit">EUR</span>
</span>
<span id="object-status-desc" class="fd-object-number__sr-only">
    Emphasized, Value State Error
</span>
```
This would require setting a default ID for the span with the status message so that you can link it.

libs/core/object-number/object-number.component.ts
MariaIDineva: About this one, if we use `Nullable<string>`, we get this error:
Screenshot 2025-10-21 at 11 04 01
which comes from the `CssClassBuilder` class. I could go ahead and modify the types there but that will lead to modifications in numerous other classes which extend `CssClassBuilder` .
So maybe I could just leave it as `class = input<string>();`, without the empty string as a default value.

libs/core/object-number/object-number.component.html
MariaIDineva: Shall that `statusMessage` have any effect if the user passes it, but does not pass `status`?

MariaIDineva:
Since none of the implementations discussed managed to fix the focus on the screen reader element when using Voiceover on Chrome, but the issue is not reproducible with JAWS, we are leaving the styles as they are.

libs/core/object-number/object-number.component.html
MariaIDineva: As discussed, `statusMessage` only has effect if `status` is passed.

PR #13532, opened 2025-10-16
feat(core): adopt breaking changes for Object Identifier

MariaIDineva:
## Related Issue(s)

## Description

Adopts breaking changes coming from [Fundamental Styles](https://github.com/SAP/fundamental-styles/pull/6067) for `Object Identifier`:
- Adds a screen reader element to read `Object Identifier` text.
- Adds a new `id` input with a default value.
- Adds `id`s to the screen reader text as well as to the description paragraph if a description is passed.
- Adds an `aria-describedby` attribute to `fd-link` when one is passed to the object identifier. Aria-described by points to the screen reader text and to the description paragraph (if present).
- Transforms all inputs into signals.

### DIFF
--- a/libs/core/object-identifier/object-identifier.component.ts
+++ b/libs/core/object-identifier/object-identifier.component.ts
@@ -7,6 +7,7 @@ import {
     ElementRef,
     QueryList,
     ViewEncapsulation,
+    booleanAttribute,
     computed,
     inject,
     input
@@ -64,10 +65,10 @@ export class ObjectIdentifierComponent implements AfterContentInit {
     description = input<Nullable<string>>();
 
     /** Whether the title should be bolded */
-    bold = input<boolean>(false);
+    bold = input(false, { transform: booleanAttribute });
 
     /** Whether the title is medium size */
-    medium = input<boolean>(false);
+    medium = input(false, { transform: booleanAttribute });
 
     /** @hidden */
     objectIdentifierClass = true;


libs/core/object-identifier/object-identifier.component.ts
```diff
@@ -5,56 +5,73 @@ import {
...
+
     /** Description text */
-    @Input()
-    description: Nullable<string>;
+    description = input<Nullable<string>>();
 
     /** Whether the title should be bolded */
-    @Input()
-    bold = false;
+    bold = input<boolean>(false);
```
InnaAtanasova: can you use `{ transform: booleanAttribute }`?

libs/core/object-identifier/object-identifier.component.ts
```diff
@@ -5,56 +5,73 @@ import {
...
 
     /** Whether the title is medium size */
-    @Input()
-    @HostBinding('class.fd-object-identifier--medium')
-    medium = false;
+    medium = input<boolean>(false);
```
InnaAtanasova: can you use `{ transform: booleanAttribute }`?

MariaIDineva:
> In the deployed version for some reason I don't see the id on the SR only element and the link doesn't have aria-describedby pointing to the id.

That's because the PR is opened against a feature branch and the deploy is not triggered after each change. I had mistakenly opened the PR against main and the deployed version is the version of my first commit. The attributes are present after the latest changes if you check locally:
Screenshot 2025-10-27 at 12 38 27

PR #13538, opened 2025-10-20
fix(cdk,core,platform): update focusable CDK to allow tabbing between child elements

### LINKED ISSUE #13515: [P2][Sourcing][Accessibility] Some elements are not focusable using TAB inside table
### Describe your proposal

If we have some elements like `input`, `date`, or `radio` fields inside table (elements with `tabindex` defaulting to be 0), using `TAB` on keyboard does not move to the next focusable element. It looks like some elements inside table cell has `tabindex` to be `-1` and when pressing `TAB`, it jumps to the button as below:

https://github.com/user-attachments/assets/4769c56a-94b3-4cb8-a76d-588396cdf3b8

I am not quite sure on accessibility requirement of the library but I see this works fine on UI5 table:

https://github.com/user-attachments/assets/d9f010dd-b8de-440c-a915-169240cfc909

Angular: 15, 20, F-ngx: Latest for each angular version

1. Create fdp-table with some elements having focusable by default.
2. Focus on one element inside table cell and use `TAB` to move to another focusable element.

### StackBlitz example(s)

Reproducible internally:

https://sap.github.io/fundamental-ngx/#/platform/table/basic#custom-columns

### Declaration

Angular: 15, 20, F-ngx: Latest for each angular version

_No response_

### StackBlitz example(s)

_No response_

### Declaration

mikerodonnell89:
fixes #13515 

needs a port to the main branch as well if this is approved

This PR allows the user to use the tab key to move focus around the focusable child elements of focusable items inside a focusable grid.


libs/cdk/src/lib/utils/directives/focusable-grid/focusable-grid.directive.ts
```diff
@@ -62,21 +71,36 @@ export class FocusableGridDirective implements AfterViewInit {
...
+    /** @hidden */
+    @ContentChildren(FDK_FOCUSABLE_ITEM_DIRECTIVE, { descendants: true })
+    private readonly _focusableItems: QueryList<FocusableItemDirective>;
+
     /** @hidden */
     _preventKeydown = false;
 
     /** @hidden */
-    constructor(private readonly _destroy$: DestroyedService) {}
+    constructor(private readonly _destroy$: DestroyedService, private readonly _elRef: ElementRef) {}
```
MariaIDineva: Unused ElementRef

MariaIDineva (approved):

InnaAtanasova (approved):

PR #13559, opened 2025-11-03
feat(btp, core): Vertical Navigation UXC updates

InnaAtanasova:
## Summary

This PR introduces the latest design updates for the Side Navigation / Vertical Navigation used in BTP. It includes significant improvements to functionality, accessibility, and documentation, along with cleanup and architecture updates.

## Key Changes

- Disabled State Support: Added handling and visual states for disabled navigation items.

- Improved Text Handling: Item text now wraps across multiple lines instead of truncating with an ellipsis.

- Overflow Promotion Logic: Items selected from the overflow menu are now promoted and displayed as the last visible item.

- Click-Based Selection: added click-based selection behavior.

- Removed Data-Object Rendering Logic: Deprecated the logic that builds the Side Navigation from a data object, as well as associated examples.

- New Usage Examples: Added examples demonstrating how to build the Side Navigation in:

     - Embedded mode
     - Overlay mode
     - These examples also include guidance for handling mobile mode.

- Accessibility Improvements: Added appropriate roles and ARIA attributes for better a11y support.

- Documentation Updates: Updated component descriptions and added i18n for ARIA labels and descriptions.

- Cross-Component Adjustments: Added Side Navigation–related logic and design alignment updates in Tool Layout (BTP), Dialog, Popover, and Shell Bar.

## Notes
This PR adopts the latest Angular features across most of the new and refactored code. Some directives and components still need to be migrated to signals, but expanding the scope further wasn’t feasible for this PR.


libs/core/popover/popover-mobile/popover-mobile.component.html
```diff
@@ -3,7 +3,7 @@
         <fd-dialog-header>
             <h1 fd-title [id]="titleId">{{ title }}</h1>
             @if (this.mobileConfig.hasCloseButton) {
-                <button fd-dialog-close-button (click)="close()"></button>
+                <button fd-dialog-close-button [mobile]="true" (click)="close()"></button>
```
InnaAtanasova, line 6: Author's note: The close button should always be in mobile mode as this is mobile popover, aka used only in mobile mode

libs/core/shellbar/shellbar.component.html
```diff
@@ -12,15 +12,21 @@
...
-                    fd-button
-                    fdType="transparent"
-                    glyph="menu2"
-                    (fdkClicked)="_navigationClicked($event)"
-                    [attr.title]="navigationButtonLabel || ('coreShellbar.navigationButtonLabel' | fdTranslate)"
-                    [attr.aria-label]="navigationButtonLabel || ('coreShellbar.navigationButtonLabel' | fdTranslate)"
-                ></button>
+                @if (navigationButtonTemplate) {
+                    <ng-container [ngTemplateOutlet]="navigationButtonTemplate"></ng-container>
+                } @else {
```
InnaAtanasova, line 17: Author's note: provide an option for the user to pass their own button or popover component, where the button is the trigger.

libs/btp/navigation/services/navigation.service.ts
```diff
@@ -1,9 +1,57 @@
...
 
 @Injectable()
 export class NavigationService {
     /** Currently active list item. */
     currentItem$ = new Subject<FdbNavigationListItem>();
+
+    /** Currently selected item (for click-based selection mode). */
+    selectedItem$ = signal<FdbNavigationListItem | null>(null);
+
+    /** Subject to notify when an overflow item is selected and should be promoted. */
```
InnaAtanasova, line 13: Author's note: the code that follows is for the overflow logic, when we have to show as last visible item an item from the overflow menu:
Screenshot 2025-11-17 at 1 58 45 PM

libs/btp/navigation/components/navigation-item/navigation-list-item.component.ts
```diff
@@ -225,18 +236,159 @@ export class NavigationListItemComponent extends FdbNavigationListItem implement
...
+            // In overflow menu context (More button popover), items should be menuitem
+            if (this.isOverflow$()) {
+                return 'menuitem';
+            }
+            return 'menuitemradio';
+        }
+        return 'treeitem';
+    });
+
+    /** aria-checked attribute value for menuitemradio role. */
```
InnaAtanasova, line 270: Author's note: treeitem has `aria-selected`, but menuitemradio has also aria-checked

libs/btp/navigation/components/navigation-item/navigation-list-item.component.ts
```diff
@@ -225,18 +236,159 @@ export class NavigationListItemComponent extends FdbNavigationListItem implement
...
+            if (link?.elementRef?.nativeElement) {
+                const textElement = link.elementRef.nativeElement.querySelector('.fd-navigation__text');
+                if (textElement) {
+                    return textElement.textContent?.trim() || undefined;
+                }
+            }
+        }
+        return undefined;
+    });
+
```
InnaAtanasova, line 319: Author's note: this is for a11y, to connect the parent item with the children items. The children's list should have as aria-label the content of the parent
Screenshot 2025-11-17 at 2 06 10 PM

libs/btp/navigation/components/navigation-item/navigation-list-item.component.ts
```diff
@@ -507,19 +759,92 @@ export class NavigationListItemComponent extends FdbNavigationListItem implement
...
+                        popover.popoverBody._focusFirstTabbableElement(true);
+                    }
+                }, 0);
+            });
+        } else {
+            // When popover closes, return focus to the parent link (for snapped state)
+            if (this.navigation.isSnapped$()) {
+                this._onZoneStable().subscribe(() => {
+                    this.focusLink();
+                });
```
InnaAtanasova, line 817: Author's note: This is to ensure keyboard navigation in overflow and snapped mode. Closing the popover with the children items the focus should go back to the parent. The focus should also go from parent to children items with left and right arrow keys.
Screenshot 2025-11-17 at 2 13 43 PM
Screenshot 2025-11-17 at 2 13 36 PM

libs/btp/navigation/components/navigation-list/navigation-list.component.ts
```diff
@@ -82,6 +110,51 @@ export class NavigationListComponent implements OnChanges, AfterViewInit, OnDest
...
+                if (parentLink?.elementRef?.nativeElement) {
+                    const textElement = parentLink.elementRef.nativeElement.querySelector('.fd-navigation__text');
+                    if (textElement) {
+                        return textElement.textContent?.trim() || null;
+                    }
+                }
+            }
+            return this.ariaLabel;
+        } catch {
+            // Fallback to default ariaLabel if DOM is not ready or there's any error
```
InnaAtanasova, line 157: Author's note: try and catch block is to prevent the "ExpressionChangedAfterItHasBeenCheckedError" error, which happens when the Side Nav is rendered inside a dialog. The navigation components are using computed properties and signals that get evaluated at different times, causing the tabindex and aria-label values to change after the initial change detection cycle.

libs/btp/navigation/components/navigation-list/navigation-list.component.ts
```diff
@@ -117,6 +198,21 @@ export class NavigationListComponent implements OnChanges, AfterViewInit, OnDest
...
+    private _elementRef = inject(ElementRef);
+
+    /** @hidden */
+    private _ariaLabelEffect = effect(() => {
+        // This effect runs whenever the signals change, but we need to ensure
+        // it doesn't run during the initial change detection cycle
+        setTimeout(() => {
+            this._updateAriaLabel();
+        }, 0);
+    });
```
InnaAtanasova, line 218: Author's note:This is also to prevent the "ExpressionChangedAfterItHasBeenCheckedError" error

libs/btp/navigation/components/navigation-list/navigation-list.component.ts
```diff
@@ -187,6 +299,25 @@ export class NavigationListComponent implements OnChanges, AfterViewInit, OnDest
...
+                this._renderer.setAttribute(this._elementRef.nativeElement, 'aria-label', ariaLabel);
+            } else {
+                this._renderer.removeAttribute(this._elementRef.nativeElement, 'aria-label');
+            }
+        } catch {
+            // Fallback to default ariaLabel if there's any issue during evaluation
+            if (this.ariaLabel) {
+                this._renderer.setAttribute(this._elementRef.nativeElement, 'aria-label', this.ariaLabel);
+            }
+        }
```
InnaAtanasova, line 316: Author's note:This is also to prevent the "ExpressionChangedAfterItHasBeenCheckedError" error

libs/btp/navigation/components/navigation-more-button/navigation-more-button.component.ts
```diff
@@ -98,25 +109,74 @@ export class NavigationMoreButtonComponent {
...
+                    const firstItem = this.listItems[0];
+                    const link = firstItem.link$();
+
+                    if (link?.elementRef?.nativeElement) {
+                        link.elementRef.nativeElement.focus();
+                    }
+                }
+            }, 150);
+        }
+    }
```
InnaAtanasova, line 200: Author's note: this is very ugly, but it's because of popover. We need to focus the first item inside the Side nav when the popover is open. 
Screenshot 2025-11-17 at 2 26 34 PM

libs/btp/navigation/components/navigation-start/navigation-content-start.component.ts
```diff
@@ -128,6 +140,13 @@ export class NavigationContentStartComponent extends FdbNavigationContentContain
             .subscribe(() => {
                 this._calculateVisibleItems();
             });
+
+        // Listen for overflow item selections to promote them to visible
+        this.navigation.service.overflowItemSelected$
+            .pipe(takeUntilDestroyed(this._destroyRef))
+            .subscribe((selectedItem) => {
+                this._promoteOverflowItem(selectedItem);
+            });
```
InnaAtanasova, line 149: Author's note: this is related to showing as last visible item the selected item from the overflow menu

mikerodonnell89 (commented):
In fundamental-styles the popovers don't have arrows. Looks like they're off vertically by a couple pixels here too

Screenshot 2025-11-18 at 11 22 38 AM

mikerodonnell89:
The width of the buttons here seems wonky to me. Like the popover needs more padding so the buttons are symmetrical left to right. It's apparent on the "All opportunities" button on hover or focus
Screenshot 2025-11-18 at 11 24 09 AM

mikerodonnell89:
Looks like too much top/bottom padding in this popover

Screenshot 2025-11-18 at 12 33 40 PM

droshev:
> The width of the buttons here seems wonky to me. Like the popover needs more padding so the buttons are symmetrical left to right. It's apparent on the "All opportunities" button on hover or focus Screenshot 2025-11-18 at 11 24 09 AM

@mikerodonnell89 what does wonky mean? Is it not against the design?

mikerodonnell89:
> @mikerodonnell89 what does wonky mean? Is it not against the design?

@droshev sorry for the slang 😆 

Screenshot 2025-11-18 at 1 18 46 PM

I don't think the designs get in to that specifically for this component, it's just a general popover thing. Appears to be coming from extra padding on the fd-navigation-link

Screenshot 2025-11-18 at 1 23 28 PM

Whereas if we look at action sheet, for example, we see the symmetry which surely is the intended look

Screenshot 2025-11-18 at 1 27 07 PM
Screenshot 2025-11-18 at 1 26 50 PM

mikerodonnell89:
The expanding arrows need RTL support 
Screenshot 2025-11-18 at 1 28 15 PM

libs/btp/navigation/components/navigation-more-button/navigation-more-button.component.ts
mikerodonnell89: This assumes that there's no other popovers with a 'a[fdb-navigation-link]' - risk we're willing to take? Can we tie an identifier to the popover or do something like this? https://github.com/SAP/fundamental-ngx/blob/982cca02cb920d9a82908fdd2283a2d3fde9e996/libs/btp/navigation/components/navigation-item/navigation-list-item.component.ts#L769

InnaAtanasova:
> > @mikerodonnell89 what does wonky mean? Is it not against the design?
> 
> @droshev sorry for the slang 😆
> 
> Screenshot 2025-11-18 at 1 18 46 PM
> I don't think the designs get in to that specifically for this component, it's just a general popover thing. Appears to be coming from extra padding on the fd-navigation-link
> 
> Screenshot 2025-11-18 at 1 23 28 PM
> Whereas if we look at action sheet, for example, we see the symmetry which surely is the intended look
> 
> Screenshot 2025-11-18 at 1 27 07 PM Screenshot 2025-11-18 at 1 26 50 PM

I agree, but it's per design 
Screenshot 2025-11-19 at 1 59 11 PM
Screenshot 2025-11-19 at 1 58 48 PM
Screenshot 2025-11-19 at 1 58 39 PM

InnaAtanasova:
> Looks like too much top/bottom padding in this popover
> 
> Screenshot 2025-11-18 at 12 33 40 PM

I mentioned it during the call, but just in case will note it here, there's an  invisible separator with groups so items that belong to a group are visually separated from the others.

InnaAtanasova:
> The expanding arrows need RTL support Screenshot 2025-11-18 at 1 28 15 PM

Fixed this. Thanks for noticing it
Screenshot 2025-11-19 at 2 31 56 PM
Screenshot 2025-11-19 at 2 31 50 PM

InnaAtanasova:
> The width of the buttons here seems wonky to me. Like the popover needs more padding so the buttons are symmetrical left to right. It's apparent on the "All opportunities" button on hover or focus Screenshot 2025-11-18 at 11 24 09 AM

This one and the comment that find-styles popovers don't have arrow are addressed. Now the popovers are without arrows
Screenshot 2025-11-19 at 2 38 07 PM
Screenshot 2025-11-19 at 2 38 01 PM

libs/btp/navigation/components/navigation-more-button/navigation-more-button.component.ts
```diff
@@ -135,7 +136,7 @@ export class NavigationMoreButtonComponent {
             map((lang: FdLanguage) => this._translationResolver.resolve(lang, 'btpNavigation.overflowMenuAriaLabel'))
         ),
         { initialValue: 'Additional Navigation Items' }
-    );
+    ) as Signal<string>;
```
MariaIDineva: You shouldn't need this type assertion.

libs/btp/navigation/components/navigation/navigation.component.scss
```diff
@@ -125,3 +144,91 @@ $block: fd-navigation;
...
+            right: auto;
+
+            &::after {
+                right: -0.25rem;
+                left: auto;
+            }
+        }
+
+        .fd-navigation__has-children-indicator:before {
+            --fdNavigation_Link_Has_Child_Indicator_Icon: '\e1ee';
```
MariaIDineva, line 235: It seems like the chevron does not rotate when expanding the children in rtl mode:

Image

libs/btp/navigation/components/navigation-link/navigation-link.component.ts
```diff
@@ -124,36 +142,174 @@ export class NavigationLinkComponent extends FdbNavigationItemLink implements On
     }
 
     /** @hidden */
-    @HostListener('click')
-    private _clickHandler(): void {
+    @HostListener('click', ['$event'])
+    _clickHandler(event: Event): void {
```
MariaIDineva: Why not keep those listeners private?

libs/btp/navigation/components/navigation/navigation.component.ts
MariaIDineva, line 55: [Those](https://github.com/SAP/fundamental-ngx/pull/13559/files/553357b2b6b464be4c183841a3bfe083a94659ce#diff-48acd89d7ef8033ae8a669a63654f74e89cfc1a3b96c572865545ec603401cd9R55) imports are unused now and can safely be removed.

MariaIDineva (commented):
It seems like in the design there is a slight difference between compact and cozy mode but I cannot spot any difference when I switch between those in the examples.

libs/btp/navigation/components/navigation/navigation.component.ts
MariaIDineva: OK, I don't know why but I cannot make github go to the correct line. I mean the imports on `line 55` in `libs/btp/navigation/components/navigation/navigation.component.ts`

MariaIDineva:
Just one more comment/question:
In the design it is stated that `"A click on a navigation item within the popover or anywhere outside the popover closes the popover."` but when clicking on the items in our examples the navigation stays open, is that desired?

InnaAtanasova:
> It seems like in the design there is a slight difference between compact and cozy mode but I cannot spot any difference when I switch between those in the examples.

yes, from the global switch (green rectangle) everything stays the same. The density for Side Nav is controlled manually on the component level (pink rectangle). 
Screenshot 2025-11-20 at 9 08 34 AM

libs/btp/navigation/components/navigation-link/navigation-link.component.ts
InnaAtanasova: When Angular compiles the template and the @HostListener decorator, it needs to access these methods to bind them to DOM events. If the methods are marked as private, TypeScript's compiler will warn that these methods are not accessible outside the class.

HostListener decorators create bindings that need to be accessible to Angular's change detection and event binding system. Even though the methods are logically "internal" to the component, Angular's runtime needs to access them. 

I will change them to `protected`.

libs/btp/navigation/components/navigation-more-button/navigation-more-button.component.ts
InnaAtanasova: removed! you were right, was not needed

libs/btp/navigation/components/navigation/navigation.component.scss
InnaAtanasova: done, broke it yesterday with the RTL changes. thanks for noticing!

libs/btp/navigation/components/navigation-more-button/navigation-more-button.component.ts
InnaAtanasova: yes, you are right! Found a better way to manage this.

libs/btp/navigation/components/navigation-link/navigation-link.component.ts
MariaIDineva: Interesting, thanks for explaining! If they are marked as protected, though, they should also not be accessible outside the class, still we don't get any warnings? Anyway, if it works, we're good.

MariaIDineva (approved):

Issue #13176, opened 2025-04-02
Unify fundamental-ngx and ui5-webcomponents-ngx


droshev:
The objective is to combine the 2 repositories (fundamental-ngx and ui5-webcomponents-ngx) into one(the current one). Here are the topics we need to cover:
- [ ] npm packages
- [ ] documentations
- [ ] playground and examples
- [ ] nx workspace
- [ ] releasing and hotfix

droshev:
All checked items have been completed. The remaining open topics (test automation, icon bundle size, localization imports, number-type input handling, checkbox indeterminate state) are tracked in #14078.

PR #13578, opened 2025-11-17
feat: add types for wrapper event emitters

MariaIDineva:
## Related Issue(s)

part of https://github.com/SAP/fundamental-ngx/issues/13176

## Description

1. Adds some wrapper generator logic that exports a custom `UI5WrapperCustomEvent` type from the `ui5-webcomponents-base` package. That type can be used to provide proper typing for all components outputs, thus avoiding imports of types and classes from the webcomponents dist folder.

Before:
```
// Import UI5 custom event types
import { UI5CustomEvent } from '@ui5/webcomponents-base';
import { default as _TextArea } from '@ui5/webcomponents/dist/TextArea.js';

onTextAreaChange(fieldName: string, event: UI5CustomEvent<_TextArea, 'input'>): void { }
```

Now:
```
import type { UI5WrapperCustomEvent } from '@fundamental-ngx/ui5-webcomponents-base';

onTextAreaChange(fieldName: string, event: UI5WrapperCustomEvent<TextArea, 'ui5Input'>): void {
```

2. Cleans up the generator logic a bit from unused parameters.

3. Adjusts the examples to use the new type for the outputs. Cleans up some unused methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,28 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [0.58.0-rc.23](https://github.com/SAP/fundamental-ngx/compare/v0.58.0-rc.22...v0.58.0-rc.23) (2025-11-26)
+
+
+### Bug Fixes
+
+* **platform:** show navigation indicator on Settings Generator items on small screens ([#13612](https://github.com/SAP/fundamental-ngx/issues/13612)) ([8cd8c5e](https://github.com/SAP/fundamental-ngx/commit/8cd8c5edfb23ef5bb68b32615302878c0b5f8938))
+
+
+
+
+
+# [0.58.0-rc.22](https://github.com/SAP/fundamental-ngx/compare/v0.58.0-rc.21...v0.58.0-rc.22) (2025-11-26)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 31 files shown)


libs/docs/ui5-webcomponents/carousel/examples/carousel-sample.ts
```diff
@@ -276,30 +269,30 @@ export class CarouselExample {
         return types[status] || '1';
     }
 
-    getImageStyle(image: any): string {
+    getImageStyle(image: { url: string; title: string; description: string; gradient: string }): string {
```
InnaAtanasova: Maybe create a type for { url: string; title: string; description: string; gradient: string }? Later if new props are added will be easier to modify just in one place.

netlify.toml
```diff
@@ -8,7 +8,7 @@
 [build]
   ignore = 'sh ./ignore.sh'
   publish = "dist/apps/docs"
-  command = "NODE_OPTIONS=--max_old_space_size=8192 npx nx run-many --target=generate --projects=ui5-webcomponents-base,ui5-webcomponents,ui5-webcomponents-ai,ui5-webcomponents-fiori && npx nx run docs:compile:production-unoptimized --base-href=\"/\""
+  command = "NODE_OPTIONS=--max_old_space_size=8192 npx nx run-many --target=generate --projects=ui5-webcomponents-base,ui5-webcomponents,ui5-webcomponents-ai,ui5-webcomponents-fiori --skip-nx-cache && npx nx run docs:compile:production-unoptimized --base-href=\"/\""
```
droshev, line 11: maybe we don't need skipping the cache anymore?

MariaIDineva: Yes, I will remove it.

MariaIDineva: Hm, now the build fails again. Maybe we should merge first and remove it another PR.

droshev: then keep it and log a ticket so we can debug it

PR #13586, opened 2025-11-18
chore(ci): move to npm trusted publishers

droshev:
move to npm trusted publishers

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -6,6 +6,9 @@ on:
         branches:
             - main
             - tmp_hotfix_branch
+permissions:
+    contents: write
+    id-token: write
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release.yml
```diff
@@ -52,8 +55,52 @@ jobs:
                   name: ${{ secrets.GH_NAME }}
                   email: ${{ secrets.GH_EMAIL }}
 
-            - name: Setup Node.js and Cache
-              uses: ./.github/actions/nodejs
+            - name: Use Node.js ${{ inputs.node-version }}
+              uses: actions/setup-node@v4.3.0
+              with:
+                  node-version: '22.x'
```
MariaIDineva, line 61: Do we intentionally use '22.x' instead of 'inputs.node-version'?

droshev: yes

PR #13599, opened 2025-11-20
chore: update to angular 21

robertIsaac:


droshev:
@robertIsaac you are super fast! thank you! We need to merge few updates, trigger a release and then we should be good to bring ng21.
btw the PR is failing

robertIsaac:
@droshev yesI see problems with import of `@fundamental-ngx/cdk/utils`
I think it's because nx doesn't support it yet, as it's not using a standard typescript paths
will try again when they release their support https://github.com/nrwl/nx/discussions/33557

robertIsaac:
for now I will make this one draft

droshev:
the ticket that does the migration in [nx](https://github.com/nrwl/nx/pull/33378)

robertIsaac:
I made [experimental branch with the nx 22.3](https://github.com/SAP/fundamental-ngx/tree/chore/nx-22.3) but still the build fails
so I created https://github.com/nrwl/nx/issues/33876

robertIsaac:
it's blocked by https://github.com/nrwl/nx/issues/34002

libs/cdk/utils/directives/readonly/fdk-readonly-provider.service.ts
```diff
@@ -25,26 +25,30 @@ export class FdkReadonlyProvider extends ReplaySubject<boolean> implements Reado
...
         private ngZone: NgZone,
         private elementRef: ElementRef<Element>,
-        private readonlyObserver: ReadonlyObserver,
         @Optional() @Self() @Inject(FDK_READONLY_DIRECTIVE) private selfReadonly$: ReadonlyBehavior,
         @Optional() @SkipSelf() @Inject(FDK_READONLY_DIRECTIVE) private parentReadonly$: ReadonlyBehavior
     ) {
         super(1);
 
+        // Initialize properties that depend on injected services
+        this._viewModifiers$ = new BehaviorSubject<ReadonlyViewModifier[]>(this._getInitialViewModifiers());
```
MariaIDineva, line 49: I guess this change is an attempt to fix the injection order, you mentioned that the other day @droshev. Could we instead solve the problem in a different way? Since we've discussed that our strategy going forward is to use dependency injection in field initializers,  I would suggest moving all injections outside the constructor. If we just order them properly, there will be no issues.
```
    /** @hidden */
    private readonly ngZone = inject(NgZone);
    
    /** @hidden */
    private readonly elementRef = inject(ElementRef<Element>);
    
    /** @hidden */
    private readonly selfReadonly$ = inject(FDK_READONLY_DIRECTIVE, { optional: true, self: true });
    
    /** @hidden */
    private readonly parentReadonly$ = inject(FDK_READONLY_DIRECTIVE, { optional: true, skipSelf: true });
    
    /** @hidden */
    private readonly _destroyRef = inject(DestroyRef);
    
    /** @hidden */
    private readonly readonlyObserver = inject(ReadonlyObserver);

    /** @hidden */
    private readonly _viewModifiers$ = new BehaviorSubject<ReadonlyViewModifier[]>(
        this._getInitialViewModifiers()
    );

    /** @hidden */
    private readonly _readonlyChange$ = this._getReadonlyChange$();
```

droshev: The goal of this PR is not to migrate the components/services. We will do that separately. I tried to keep the changes minimal

libs/ui5-webcomponents-ai/tsconfig.typedoc.json
```diff
@@ -11,7 +11,12 @@
         "jsx": "react",
         "jsxFactory": "JSX.createElement",
         "jsxFragmentFactory": "JSX.Fragment",
-        "moduleResolution": "bundler"
+        "moduleResolution": "bundler",
+        "baseUrl": ".",
+        "paths": {
+            "@fundamental-ngx/ui5-webcomponents-base": ["../ui5-webcomponents-base/index.ts"],
+            "@fundamental-ngx/ui5-webcomponents-base/*": ["../ui5-webcomponents-base/*/index.ts"]
```
MariaIDineva, line 18: Since we are in `ui5-webcomponents-ai` shouldn't these be:
```
            "@fundamental-ngx/ui5-webcomponents-ai": ["./index.ts"],
            "@fundamental-ngx/ui5-webcomponents-ai/*": ["./*/index.ts"]
```

droshev: It needs path to the base package

Issue #13546, opened 2025-10-27
User Search Limitations and Navigation Issues in approval flow


deepikajain1275:
### Describe your proposal

--Symptom--

When user is adding the user to the approval flow and going inside the details for the particular user within user search. After choosing a user from the search results and clicking the side arrow to view their details, closing the detail pane redirects the user back to the initial screen instead of returning to the previous search results.

Please fix to preserve the search state after viewing and closing a user's details?

Issue is replicated internally on our internal AMS Sandbox-T realm. Please see the attached video recording showing the steps with internal replication.

--Environment--

Source: portal
Type: Test System
System ID/Name: CLOUD / UCProcurement-T
Product version:
System managed by: SAP Cloud Tenant
Install base item: UCProcurement-T
Sold product: ARIBA
--Steps to reproduce--

As example on the customer realm:

Search for the e.g. document id: Doc2755324864

go to the task that can be edited, e.g. (edit task TSK2755324877)

Click on Edit Approval Flow

Click on plus icon to add the approver to the approval flow

Search for the particular user e.g. type in "andrea"

there is a search result for all findings contains "andrea"

go to the details for one user with the using the arrow on the right side next to the user name

then go back with the error on the left side

now you are not in the previous search result for entry "andrea" but you are at the initial screen.

Expected behavior: when you click arrow on the left side back to the user search result, you should be redirected to the previous search result, not to the beggining (whole list of the user).

 

--Business impact--

Business impact category: Go-live impacted
Is there a workaround: No

### Which versions of Angular and Fundamental Library for Angular are affected?

Angular :17.3.12 , Fundamental Library:0.43.60

### Steps to Reproduce (for bugs)

1. Navigate to /platform/approval-flow.
2. Click on the Edit button.
3. Click on Add Approval After.
4. In the search box, enter a name.
5. Select the desired item from the search results.
6. A detail dialog opens for the selected item.
7. Close the detail dialog.
8. List all data (i.e., verify that the full list of data items is visible again, not just the filtered/searched ones).

https://github.com/user-attachments/assets/2dc7b376-f09d-43f9-abde-2740f1974e67

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

droshev:
@deepikajain1275 Can you debug, fix and and open a PR

deepikajain1275:
@droshev I have debug and fix it but not able to create branch and PR can you please help me on it.

PR #13604, opened 2025-11-24
User Search Limitations and Navigation Issues in approval flow 

deepikajain1275:
## Related Issue(s)
When user is adding the user to the approval flow and going inside the details for the particular user within user search. After choosing a user from the search results and clicking the side arrow to view their details, closing the detail pane redirects the user back to the initial screen instead of returning to the previous search results.

closes https://github.com/SAP/fundamental-ngx/issues/13546

## Description

Save input search result when use retrun to their detail than we send store search result to search input field.

## Screenshots

Before fix 

https://github.com/user-attachments/assets/6298278c-53a1-402d-9743-bb59b3317d35

after fix

https://github.com/user-attachments/assets/3301b309-1912-43f2-b3e8-fb68a7a4e52f


droshev:
@deepikajain1275 you need to open a PR so we can run the CI, but before that you need to join https://github.com/SAP/

deepikajain1275:
Worng request

Issue #13609, opened 2025-11-25
Side Nav clean-up


InnaAtanasova:
Remove from the documentation and mark as deprecated all Side Nav variations except for the BTP one, which is the only one that is currently per UXC design.

PR #13610, opened 2025-11-25
docs(cx, core): add deprecation message for Side nav components in Core and CX 

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/13609

## Description

add deprecation message for Side nav components in Core and CX 
## Screenshots
Core Side Navigation
Screenshot 2025-11-25 at 1 04 22 PM

CX Side Navigation
Screenshot 2025-11-25 at 1 04 33 PM

### DIFF
--- a/libs/docs/core/side-navigation/side-navigation-header/side-navigation-header.component.ts
+++ b/libs/docs/core/side-navigation/side-navigation-header/side-navigation-header.component.ts
@@ -1,5 +1,4 @@
 import { Component } from '@angular/core';
-import { RouterLink } from '@angular/router';
 import { LinkComponent } from '@fundamental-ngx/core/link';
 import { MessageStripComponent } from '@fundamental-ngx/core/message-strip';
 import {
@@ -21,8 +20,8 @@ import {
         ImportComponent,
         HeaderTabsComponent,
         MessageStripComponent,
-        RouterLink,
-        LinkComponent
+        LinkComponent,
+        MessageStripComponent
     ]
 })
 export class SideNavigationHeaderComponent {}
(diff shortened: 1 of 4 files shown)


libs/docs/core/side-navigation/side-navigation-header/side-navigation-header.component.ts
```diff
@@ -22,7 +22,8 @@ import {
         HeaderTabsComponent,
         MessageStripComponent,
         RouterLink,
-        LinkComponent
+        LinkComponent,
+        MessageStripModule
```
MariaIDineva: Do we need this `MessageStripModule` since `MessageStripComponent` is already imported?
Also, it seems like  `LinkComponent` and `RouterLink` imports are not used and can be removed or maybe we would like to add a link to the **BTP Vertical Navigation component***?

libs/docs/cx/side-navigation/side-navigation-header/side-navigation-header.component.ts
```diff
@@ -11,6 +12,13 @@ import {
...
     templateUrl: './side-navigation-header.component.html',
     changeDetection: ChangeDetectionStrategy.OnPush,
-    imports: [DocPageComponent, HeaderComponent, DescriptionComponent, ImportComponent, HeaderTabsComponent]
+    imports: [
+        DocPageComponent,
+        HeaderComponent,
+        DescriptionComponent,
+        ImportComponent,
+        HeaderTabsComponent,
+        MessageStripModule
```
MariaIDineva: Do we need the whole module or can we import only `MessageStripComponent`?
Do we want to add a link here as well?

PR #13634, opened 2025-11-28
fix(platform): Approval flow search issue when add

deepikaSap1:
## Related Issue(s)

When user is adding the user to the approval flow and going inside the details for the particular user within user search. After choosing a user from the search results and clicking the side arrow to view their details, closing the detail pane redirects the user back to the initial screen instead of returning to the previous search results.

closes  https://github.com/SAP/fundamental-ngx/issues/13546

## Description

Save input search result when use retrun to their detail than we send store search result to search input field.

## Screenshots

Before fix

https://github.com/user-attachments/assets/d1c2fdd6-163e-41ed-892d-1937736ec929

after fix

https://github.com/user-attachments/assets/a316ee12-6d2f-4b83-bf3b-e6a0a045d3f3


droshev:
@deepikaSap1 do we need this PR?

PR #13645, opened 2025-12-02
chore: remove lodash

robertIsaac:
lodash is outdated took
it's also creating problems with vitest
so it's better not to have it

### DIFF
--- a/libs/btp/tsconfig.lib.json
+++ b/libs/btp/tsconfig.lib.json
@@ -6,7 +6,7 @@
         "declaration": true,
         "declarationMap": true,
         "types": [],
-        "lib": ["dom", "es2020"],
+        "lib": ["dom", "es2023"],
         "useDefineForClassFields": false,
         "moduleResolution": "bundler"
     },
--- a/libs/core/schematics/tsconfig.json
+++ b/libs/core/schematics/tsconfig.json
@@ -7,7 +7,7 @@
     "compilerOptions": {
         "allowSyntheticDefaultImports": true,
         "baseUrl": ".",
-        "lib": ["es2020", "dom"],
+        "lib": ["es2023", "dom"],
         "declaration": true,
         "module": "commonjs",
         "moduleResolution": "node",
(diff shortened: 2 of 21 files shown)


libs/cdk/utils/functions/lodash-utils.ts
```diff
@@ -0,0 +1,302 @@
+/**
```
droshev, line 1: do we need to name it lodash if it is our own functionality?

droshev:
@robertIsaac Thank you for the contribution. Could you give us some time to discuss and review it thoroughly?

robertIsaac:
Yes take your time to decide 
But if you don't want to merge this you need to give us a way to use fundamentals with vitest
Because right now it throws errors

droshev:
> Yes take your time to decide But if you don't want to merge this you need to give us a way to use fundamentals with vitest Because right now it throws errors

@robertIsaac if it is not something internal, do you think you can give us more context?

robertIsaac:
Creating a new Angular project (which uses vitest by default) and importing any of the components that uses lodash will break the unit test 
I have a reprod, I will push it to GitHub and share it with you

robertIsaac:
@droshev here is the reprod
https://github.com/robertIsaac/reprod-vitest-fundamental.git
and here is the problem
```
Error: Cannot find module '/Users/I755229/temp/reprod-vitest-fundamental/node_modules/lodash-es/get' imported from /Users/I755229/temp/reprod-vitest-fundamental/node_modules/@fundamental-ngx/core/fesm2022/fundamental-ngx-core-multi-input.mjs
Did you mean to import "lodash-es/get.js"?
```

libs/cdk/utils/functions/lodash-utils.ts
robertIsaac: no, we can named it anything
but I think for the short term it's good to be called lodash so if a bug raised and someone took a look they can compare how the new function behave compared to the lodash library

droshev: fair point 👍

InnaAtanasova (approved):

PR #13658, opened 2025-12-05
fix(core): Prevent event propagation on <span> element with class 'fd-time__item--collapsed'

yixuan-liu:

### DIFF
--- a/libs/core/time/time-column/time-column.component.ts
+++ b/libs/core/time/time-column/time-column.component.ts
@@ -253,8 +253,18 @@ export class TimeColumnComponent<K extends number, T extends SelectableViewItem<
     }
 
     /** @hidden */
-    @HostListener('click')
-    onItemClick(): void {
+    @HostListener('click', ['$event'])
+    onItemClick(event: MouseEvent): void {
+        const targetRef = event.target as HTMLElement;
+
+        // Check if the click happened on the collapsed time item span
+        if (
+            targetRef.classList.contains('fd-time__item--collapsed') ||
+            targetRef.closest('.fd-time__item--collapsed')
+        ) {
+            // Stop event propagation to prevent bubbling
+            event.stopPropagation();
+        }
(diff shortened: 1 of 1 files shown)


libs/core/time/time-column/time-column.component.ts
```diff
@@ -253,8 +253,18 @@ export class TimeColumnComponent<K extends number, T extends SelectableViewItem<
     }
 
     /** @hidden */
-    @HostListener('click')
-    onItemClick(): void {
+    @HostListener('click', ['$event'])
+    onItemClick(event: MouseEvent): void {
```
mikerodonnell89, line 257: We should make this event an optional parameter as a safeguard against breaking something else, and of course the subsequent code using it needs to check for it

libs/core/time/time-column/time-column.component.ts
```diff
@@ -253,8 +253,18 @@ export class TimeColumnComponent<K extends number, T extends SelectableViewItem<
...
+        const targetRef = event.target as HTMLElement;
+
+        // Check if the click happened on the collapsed time item span
+        if (
+            targetRef.classList.contains('fd-time__item--collapsed') ||
+            targetRef.closest('.fd-time__item--collapsed')
+        ) {
+            // Stop event propagation to prevent bubbling
+            event.stopPropagation();
+        }
```
mikerodonnell89, line 267: If we are to merge this PR as a temp hotfix, I think we need an input (something like `timeItemStopPropagation`) and this logic needs to be wrapped in a conditional for that input, to ensure we're not affecting behavior elsewhere

yixuan-liu: sure Mike.... will add in the conditional logic check based on a new input prop

droshev: `timeItemStopPropagationSourcing`

PR #13680, opened 2025-12-17
fix: remove project.json from secondary entry points

robertIsaac:
according to https://github.com/nrwl/nx/issues/33876#issuecomment-3664891086
this is the reason the project fail to build after updating to Angular 21


robertIsaac:
@droshev should we remove them from docs as well?

PR #13683, opened 2025-12-19
feat(ui5): add logic for superclass properties, slots and methods to wrapper components

MariaIDineva:
## Related Issue(s)

part of https://github.com/SAP/fundamental-ngx/issues/13176

## Description

Certain webcomponents would extend other component classes and would depend on their properties, slots and methods. For example, `SearchItemGroup` would extend `ListItemGroup` where it would inherit its `header-text` property (in the wrapper that is `headerText`) from. In the custom elements json file that is displayed as (note the "superclass" property in the "declarations" array):
```
    {
      "kind": "javascript-module",
      "path": "dist/SearchItemGroup.js",
      "declarations": [
        {
          "kind": "class",
          "description": "The `ui5-search-item-group` is type of suggestion item,\nthat can be used to split the `ui5-search-item` suggestions into groups.",
          "name": "SearchItemGroup",
          "superclass": {
            "name": "ListItemGroup",
            "package": "@ui5/webcomponents",
            "module": "dist/ListItemGroup.js"
          },
          "tagName": "ui5-search-item-group",
          "customElement": true,
          "_ui5experimental": true,
          "_ui5since": "2.9.0",
          "_ui5privacy": "public"
        }
      ],
   ....
   }
```

This PR adds logic to our gerenator script to add all inherited properties, slots and methods to the components so that they are exposed and available to use. It does so by:
- Adding a new function `collectInheritedMembers` which would recursively gather all that is inherited and pass it down to the `componentTemplate` generation logic.
- Adding a new `dependencyCemFiles` property to the generate options in `project.json` that would point to the CEM file that the particular library depends on as some components might inherit components from a different package, eg. a component from `ui5-webcomponents-fiori` could extend a component from `ui5-webcomponents`,


libs/webc-generator/src/executors/generate/executor.ts
```diff
@@ -1,5 +1,6 @@
 import { ExecutorContext, PromiseExecutor } from '@nx/devkit';
 import type * as CEM from '@ui5/webcomponents-tools/lib/cem/types-internal.d.ts';
+import type { CustomElementDeclaration } from '@ui5/webcomponents-tools/lib/cem/types-internal.d.ts';
```
droshev: don't we have that import from the previous row?

MariaIDineva: Correct, I'll fix that.

MariaIDineva:
As it turned out, the issue was particularly related to those inherited properties that come from a class from another library. For example, if a component from "webcomponents-fiori" extends a component from "webcomponents", the definitions for the inherited properties would be missing.

After a discussion with the webcomponents team, it was decided that they will start exposing all inherited properties, no matter which library they come from. That will happen when [this change](https://github.com/UI5/webcomponents/pull/12851) is released. We will only need to bump the version to use the new CEM file and the generation will work as expected. Therefore, this logic is redundant.

PR #13684, opened 2025-12-21
fix: ui5 executor wasn't generated the ui5 module

robertIsaac:
Before this PR the `libs/webc-generator/src/executors/generate/executor.ts` was not generating the `ui5.module.ts` file for the ui5 libraries
So when there is a change, or a new component is added it wouldn't reflect
This resulted in errors in compiling these libraries 
Now with this PR they will work correctly


droshev:
@robertIsaac Can you describe the problem so we can get the details?

droshev:
@robertIsaac actually we had libs/<package>fundamental-ngx.module.ts, but we weren't filling it

robertIsaac:
But in the index.ts it was exporting ui5.module.ts

robertIsaac:
It'll be deleted

PR #13689, opened 2025-12-23
docs: signalize stackblitz service

### LINKED ISSUE #13394: Angular 20 Stackblitz template
### Describe your proposal

Update the stackblitz template reflecting the Angular 20.

latest

_No response_

### StackBlitz example(s)

_No response_

### Declaration

droshev:
fixes #13394

### DIFF
--- a/libs/docs/core/avatar-group-legacy/examples/avatar-group-legacy-group-type-example.component.html
+++ b/libs/docs/core/avatar-group-legacy/examples/avatar-group-legacy-group-type-example.component.html
@@ -55,7 +55,7 @@
                 </div>
             }
             @if (avatarGroup_GroupType.overflowItemsCount > 0) {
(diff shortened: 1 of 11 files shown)


libs/docs/core/action-sheet/examples/action-sheet-compact/action-sheet-compact-example.component.html
```diff
@@ -1,4 +1,4 @@
-<fd-action-sheet fdCompact (isOpenChange)="isOpenChange($event)">
```
InnaAtanasova, line 1: I think we should keep this in the example. In mobile mode the Action Sheet has a completely different look (a dialog). Also, above this example we have another one (Default Action Sheet), which now basically is the same as this one.

PR #13696, opened 2026-01-01
fix(core): signalify text component

droshev:
signalify text component

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -247,6 +247,228 @@ export class MyComponent {
 }
 ```
 
+### Dependency Injection Patterns
+
+Angular's dependency injection system provides powerful patterns for component composition and configuration. This section covers two key patterns using `InjectionToken`.
+
+#### Pattern 1: Contextual Defaults with InjectionTokens
+
+Use `InjectionToken` to provide contextual defaults for child components. This pattern allows parent components to influence default values without directly manipulating child component inputs.
+
+**When to use:**
+
+- Parent components need to provide default configurations to unknown child components
+- Setting framework-level defaults (themes, sizes, behaviors)
+- Avoiding tight coupling between parent and child components
+- Providing optional configuration that children can override
+
+**Example - Providing default title size in dialogs:**
+
+```typescript
+// title.component.ts - Define the token
+import { InjectionToken } from '@angular/core';
+
+export type HeaderSizes = 1 | 2 | 3 | 4 | 5 | 6;
+
+export const DEFAULT_TITLE_SIZE = new InjectionToken<HeaderSizes>('DEFAULT_TITLE_SIZE');
+
+@Component({
+    selector: '[fd-title]'
+    /* ... */
+})
+export class TitleComponent {
+    readonly headerSize = input<HeaderSizes | null>(null);
+
+    private readonly _defaultHeaderSize = inject(DEFAULT_TITLE_SIZE, { optional: true });
+
+    constructor() {
+        effect(() => {
+            // Priority: explicit input > injected default > element tag name
+            const size = this.headerSize() ?? this._defaultHeaderSize ?? this._detectSize();
+            this._applySize(size);
+        });
+    }
+}
+
+// dialog-header.component.ts - Provide the default
+@Directive({
+    providers: [
+        {
+            provide: DEFAULT_TITLE_SIZE,
+            useValue: 5
+        }
+    ]
+})
+export class DialogHeaderBase {
+    // Any fd-title within this directive will default to size 5
+}
+```
+
+**Benefits:**
+
+- ✅ No need to query and manipulate child components
+- ✅ Type-safe dependency injection
+- ✅ Works naturally with signal inputs
+- ✅ Testable (mock the token in tests)
+- ✅ Loosely coupled (parent doesn't need to know about child implementation)
+- ✅ Children can opt-out by providing explicit values
+
+**Best practices:**
+
+- Always use `{ optional: true }` when injecting contextual defaults
+- Document the token with JSDoc describing its purpose
+- Export tokens from the component file for reusability
+- Use descriptive token names (e.g., `DEFAULT_TITLE_SIZE` not `TITLE_CONFIG`)
+
+**Why Not @ContentChild with Signal Inputs?**
+
+When migrating components to signals, you may encounter situations where a parent component needs to set default values for child component inputs. The old pattern of querying and manipulating children **no longer works with signal inputs**.
+
+**❌ This pattern breaks with signal inputs:**
+
+```typescript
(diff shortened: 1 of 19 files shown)


libs/core/text/text.component.ts
```diff
@@ -24,72 +35,88 @@ export type HyphenationType = 'none' | 'manual' | 'auto' | null;
...
+    readonly text = input<string>('');
 
     /**
-     * Max visible lines of text
+     * Maximum number of visible lines before text is clamped. Set to `null` to disable.
+     * @default null
      */
-    @Input()
-    maxLines: Nullable<number> = null;
+    readonly maxLines = input<Nullable<number>>(null);
```
robertIsaac: why null and not undefined?
because the default in angular is undefined not null

robertIsaac (approved):

libs/core/text/text.component.ts
```diff
@@ -24,72 +35,88 @@ export type HyphenationType = 'none' | 'manual' | 'auto' | null;
...
-    }
+    /**
+     * Computed signal determining if expand/collapse functionality is enabled.
+     * @hidden
+     */
+    protected readonly _expandable = computed(() => {
+        const expand = this.expandable();
+        const max = this.maxLines();
+        return expand && !!max && max > 0;
+    });
```
InnaAtanasova: in `_isCollapsed` and `_expandable` signals you repeat this logic twice:
```
const max = this.maxLines();
!!max && max > 0
```
You can make it a separate signal:
```
protected readonly _hasValidMaxLines = computed(() => {
    const max = this.maxLines();
    return max > 0;
});
```
I don't think you need !!max as you have max > 0, but check if this is the case. 

Then you can have:
```
protected readonly _isCollapsed = computed(() =>
    this.isCollapsed() && this._hasValidMaxLines()
);

protected readonly _expandable = computed(() =>
    this.expandable() && this._hasValidMaxLines()
);
```

libs/core/text/text.component.ts
```diff
@@ -24,72 +35,88 @@ export type HyphenationType = 'none' | 'manual' | 'auto' | null;
...
-        this._hasMore = !!this.maxLines && count > this.maxLines;
-        this._changeDetectorRef.detectChanges();
+    /**
+     * Updates `_hasMore` signal based on actual line count from line clamp directive.
+     * @hidden
+     */
+    protected checkLineCount(count: number): void {
+        const max = this.maxLines();
+        this._hasMore.set(!!max && count > max);
+        this._changeDetectorRef.markForCheck();
```
InnaAtanasova: You need to check if `this._changeDetectorRef.markForCheck();` is still needed since we are using signals and no change detection is needed. This will be needed if you have cases where text code runs outside of Angular's zone.

libs/core/text/text.component.ts
droshev: I will remove Nullable and use just null

libs/core/text/text.component.ts
droshev: thanks! updated

libs/core/text/text.component.ts
droshev: thanks! updated

libs/docs/core/text/examples/text-expandable.component.ts
```diff
@@ -8,20 +8,7 @@ import { FdPatchLanguageDirective } from '@fundamental-ngx/i18n';
...
-        Pellentesque a fermentum diam, eu fermentum enim. Phasellus faucibus risus id nisi faucibus, ac tempus nisi imperdiet.
-        Donec urna felis, ultricies vitae mauris at, convallis aliquet purus. Quisque vitae dapibus nisl, ultrices facilisis nisi.
-        Ut bibendum sed tortor eget pretium. Aenean dignissim nec sem non tempor.`;
+    protected readonly text = `Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam vestibulum justo non dui viverra mattis. Fusce venenatis tortor sit amet neque volutpat, eu mollis eros pulvinar. Duis sagittis aliquam lacus, quis tempor nisi vulputate et. Integer sagittis, metus ac hendrerit luctus, enim diam hendrerit nisl, et pretium turpis magna sed felis. Etiam volutpat condimentum justo at auctor. Sed imperdiet elementum ex, at semper metus vestibulum vitae. Quisque orci odio, tincidunt sed felis et, porta ultricies ligula. Etiam finibus, diam nec ultrices ultricies, odio nisi pretium augue, eget pharetra ipsum massa quis purus. Curabitur ante sapien, pharetra sed convallis non, vehicula nec leo. Fusce et elit aliquet, lobortis urna eu, porttitor eros.`;
```
mikerodonnell89, line 11: Why are these text variables all made readonly? Don't want the examples to give the impression that it's required

droshev: what is wrong If the text is not going to change?

libs/docs/core/text/examples/text-expandable.component.html
```diff
@@ -1,14 +1,15 @@
-<h3>Expandable text</h3>
+<h3>Expandable text with 3 lines (collapsed by default)</h3>
 <fd-text [text]="text" [maxLines]="3" [expandable]="true"></fd-text>
 
-<h3>Expandable text opened</h3>
-<fd-text [text]="text" [maxLines]="3" [expandable]="true" [isCollapsed]="false"></fd-text>
+<h3>Expandable text with two-way binding</h3>
+<fd-text [text]="text" [maxLines]="3" [expandable]="true" [(isCollapsed)]="isCollapsed"></fd-text>
+<p>Current state: {{ isCollapsed() ? 'Collapsed' : 'Expanded' }}</p>
```
MariaIDineva, line 6: Do we need this `p`? It looks a little bit odd to me since the text above and the text below don't have it.

Image

droshev: above is the component (`fd-text`) and below is the headline for the next example `h3`

libs/docs/core/text/examples/text-expandable.component.ts
mikerodonnell89: Nothing is really wrong with it, but putting myself in the shoes of someone reading code examples, I'd assume that's a necessary piece of the puzzle to get the same results as the example.

robertIsaac: it's actually recommended now to have all non-private properties of the component to be readonly

droshev: @mikerodonnell89 @robertIsaac is right - the [Angular Style Guide](https://angular.dev/style-guide#use-readonly-for-properties-that-shouldnt-change)

PR #13698, opened 2026-01-02
fix(core): signalify title component

droshev:
signalize and make zoneless title component, deprecate the module

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -247,6 +247,228 @@ export class MyComponent {
 }
 ```
 
+### Dependency Injection Patterns
+
+Angular's dependency injection system provides powerful patterns for component composition and configuration. This section covers two key patterns using `InjectionToken`.
+
+#### Pattern 1: Contextual Defaults with InjectionTokens
+
+Use `InjectionToken` to provide contextual defaults for child components. This pattern allows parent components to influence default values without directly manipulating child component inputs.
+
+**When to use:**
+
+- Parent components need to provide default configurations to unknown child components
+- Setting framework-level defaults (themes, sizes, behaviors)
+- Avoiding tight coupling between parent and child components
+- Providing optional configuration that children can override
+
+**Example - Providing default title size in dialogs:**
+
+```typescript
+// title.component.ts - Define the token
+import { InjectionToken } from '@angular/core';
+
+export type HeaderSizes = 1 | 2 | 3 | 4 | 5 | 6;
+
+export const DEFAULT_TITLE_SIZE = new InjectionToken<HeaderSizes>('DEFAULT_TITLE_SIZE');
+
+@Component({
+    selector: '[fd-title]'
+    /* ... */
+})
+export class TitleComponent {
+    readonly headerSize = input<HeaderSizes | null>(null);
+
+    private readonly _defaultHeaderSize = inject(DEFAULT_TITLE_SIZE, { optional: true });
+
+    constructor() {
+        effect(() => {
+            // Priority: explicit input > injected default > element tag name
+            const size = this.headerSize() ?? this._defaultHeaderSize ?? this._detectSize();
+            this._applySize(size);
+        });
+    }
+}
+
+// dialog-header.component.ts - Provide the default
+@Directive({
+    providers: [
+        {
+            provide: DEFAULT_TITLE_SIZE,
+            useValue: 5
+        }
+    ]
+})
+export class DialogHeaderBase {
+    // Any fd-title within this directive will default to size 5
+}
+```
+
+**Benefits:**
+
+- ✅ No need to query and manipulate child components
+- ✅ Type-safe dependency injection
+- ✅ Works naturally with signal inputs
+- ✅ Testable (mock the token in tests)
+- ✅ Loosely coupled (parent doesn't need to know about child implementation)
+- ✅ Children can opt-out by providing explicit values
+
+**Best practices:**
+
+- Always use `{ optional: true }` when injecting contextual defaults
+- Document the token with JSDoc describing its purpose
+- Export tokens from the component file for reusability
+- Use descriptive token names (e.g., `DEFAULT_TITLE_SIZE` not `TITLE_CONFIG`)
+
+**Why Not @ContentChild with Signal Inputs?**
+
+When migrating components to signals, you may encounter situations where a parent component needs to set default values for child component inputs. The old pattern of querying and manipulating children **no longer works with signal inputs**.
+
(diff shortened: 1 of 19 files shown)


libs/core/dialog/base/dialog-header-base.class.ts
```diff
@@ -17,9 +18,12 @@ export abstract class DialogHeaderBase implements AfterContentInit {
     /** @hidden */
     @ContentChild(TitleToken)
     set defaultTitleSize(title: TitleComponent) {
-        if (title && !title.headerSize) {
-            title.headerSize = 5;
-            this._changeDetectorRef.detectChanges();
+        if (title && !title.headerSize()) {
+            // Get the component instance's ComponentRef for setting signal inputs
+            const componentRef = (title as any)._componentRef as ComponentRef<TitleComponent>;
```
MariaIDineva: I am always a little worried when we have to do `as any` or any other type assertions.
Can we instead mark `headerSize` as `model` instead of `input` in the `TitleComponent` and do simply:
```
title.headerSize.set(5);
```
I, personally, find it easer to read. Also, that way we will preserve type safety. Do you see any drawbacks?

libs/core/dialog/base/dialog-header-base.class.ts
```diff
@@ -17,9 +18,12 @@ export abstract class DialogHeaderBase implements AfterContentInit {
     /** @hidden */
     @ContentChild(TitleToken)
```
InnaAtanasova: content child can also be a signal

libs/core/dialog/base/dialog-header-base.class.ts
```diff
@@ -53,6 +57,6 @@ export abstract class DialogHeaderBase implements AfterContentInit {
                     break;
             }
         });
-        this._changeDetectorRef.detectChanges();
+        this._changeDetectorRef.markForCheck();
```
InnaAtanasova, line 53: When you migrate to signal, please check if this is still needed?

MariaIDineva:
Just one more thing that I noticed - in the documentation we still encourage the user to use `TitleModule` instead of the component. Maybe we should update that as well, like you did for the text component.
Screenshot 2026-01-06 at 10 37 28

libs/core/dialog/base/dialog-header-base.class.ts
droshev: it could be, be i migrated only the text component. the dialog migration will be done separately

libs/core/dialog/base/dialog-header-base.class.ts
droshev: Can we migrate the dialog in a separate PR?

droshev:
> Just one more thing that I noticed - in the documentation we still encourage the user to use `TitleModule` instead of the component. Maybe we should update that as well, like you did for the text component. Screenshot 2026-01-06 at 10 37 28

had to introduce componentName in the import component so we can have it there

libs/core/dialog/base/dialog-header-base.class.ts
droshev: the logic was removed anyway, so no need of it at all

libs/core/title/title.component.ts
```diff
@@ -1,78 +1,137 @@
...
+ * })
+ * export class DialogHeaderComponent { }
+ *
+ * // Child receives the default automatically
+ * <fd-dialog-header>
+ *   <h1 fd-title>Title</h1>  <!-- Will be size 5 -->
+ * </fd-dialog-header>
+ * ```
+ */
+export const DEFAULT_TITLE_SIZE = new InjectionToken<HeaderSizes>('DEFAULT_TITLE_SIZE');
```
InnaAtanasova, line 40: This should go into a separate file called token.ts
There you can add an injection token for the Title component itself

libs/core/title/title.component.ts
```diff
@@ -1,78 +1,137 @@
...
+
+/**
+ * Abstract token for providing access to the title component's element reference.
+ * Used for dependency injection when components need to access the title element.
+ */
 export abstract class TitleToken {
     abstract elementRef: ElementRef;
 }
```
InnaAtanasova, line 49: This is not needed, in my opinion. It can be a simple token:
```
export const TITLE_TOKEN = new InjectionToken<{ elementRef: ElementRef }>('TITLE_TOKEN');
```
The elementRef can be injected inside the component logic:
```
@Component({
    selector: '[fd-title], [fdTitle]',
    // ...
    providers: [{ provide: TITLE_TOKEN, useExisting: TitleComponent }]
})
export class TitleComponent {
    readonly elementRef = inject(ElementRef<HTMLElement>);
    ...
    ...
}
```

In the Toolbar component (toolbar.component.ts), where the token is used, you can just use:
```
@ContentChild(TITLE_TOKEN)
set titleComponent(title: { elementRef: ElementRef } | null) {
    this._titleComponent$.next(title);
}
```

This way you won't need inheritance (line 70, `export class TitleComponent extends TitleToken`).

libs/core/message-box/message-box.module.ts
```diff
@@ -23,8 +23,8 @@ const declarations = [
 ];
 
 @NgModule({
-    imports: [TemplateModule, InitialFocusModule, TitleModule, declarations],
-    exports: [declarations, TemplateModule, InitialFocusModule, TitleModule],
+    imports: [TemplateModule, InitialFocusModule, TitleComponent, declarations],
+    exports: [declarations, TemplateModule, InitialFocusModule, TitleComponent],
```
InnaAtanasova, line 27: Why do you need to export TitleComponent?

droshev: It was legacy, i removed the export

libs/core/title/title.component.ts
droshev: updated

libs/core/title/title.component.ts
droshev: Both patterns exist: separate files for file multiple and shared tokens (like `libs/core/list/tokens.ts`) vs. in-component for single/coupled ones (`libs/core/multi-combobox/multi-combobox.component.ts`).
Splitting would separate token from JSDoc/examples and add overhead.
'DEFAULT_TITLE_SIZE' uses no FD_ prefix—it's a config token (like `THEMING_CONFIG_TOKEN`, `DYNAMIC_PAGE_HEADER_TOKEN`).
We can always introduce separate files if more tokens are needed.

InnaAtanasova (approved):

PR #13699, opened 2026-01-05
fix(core): signalify skeleton and update jsdoc

droshev:
signalify skeleton and update jsdoc

### DIFF
--- a/libs/core/skeleton/components/skeleton.component.ts
+++ b/libs/core/skeleton/components/skeleton.component.ts
@@ -1,110 +1,160 @@
 import { coerceNumberProperty, NumberInput } from '@angular/cdk/coercion';
+import { ChangeDetectionStrategy, Component, computed, input, ViewEncapsulation } from '@angular/core';
 
-import {
-    ChangeDetectionStrategy,
-    Component,
-    HostBinding,
-    Input,
-    OnChanges,
-    SimpleChanges,
-    ViewEncapsulation
-} from '@angular/core';
-
-export type SkeletonType = 'rectangle' | 'circle' | 'text';
+/**
+ * Type of skeleton visualization.
+ * - `'rectangle'` - Rectangular skeleton with rounded corners
+ * - `'circle'` - Circular skeleton
+ * - `'text'` - Multi-line text skeleton
+ * - `null` - No predefined shape, allows custom SVG projection
+ */
+export type SkeletonType = 'rectangle' | 'circle' | 'text' | null;
 
 let skeletonUniqueId = 0;
 
+/**
+ * Skeleton component for displaying loading placeholders with animated pulse effect.
+ */
 @Component({
     selector: 'fd-skeleton',
     templateUrl: './skeleton.component.html',
     styleUrl: './skeleton.component.scss',
     encapsulation: ViewEncapsulation.None,
     changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        class: 'fd-skeleton',
+        '[class.fd-skeleton--animated]': 'animated()',
+        '[style.width]': 'computedWidth()',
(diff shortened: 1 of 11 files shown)


libs/core/skeleton/components/skeleton.component.ts
```diff
@@ -1,110 +1,153 @@
...
+     * Computed height value from dimensions.
+     * @hidden
+     */
+    protected readonly _height = computed(() => this._dimensions().height);
+
+    /**
+     * Unique ID for SVG mask element.
+     * @hidden
+     */
+    protected readonly _id = `fd-skeleton-${skeletonUniqueId++}`;
```
InnaAtanasova: This is not a private attr and should not be prefixed with underscore (it's used in the template).

libs/core/skeleton/components/skeleton.component.ts
```diff
@@ -1,110 +1,153 @@
...
+
+    /**
+     * Computed array of line widths for text type.
+     * Last line is 60% width when there are multiple lines.
+     * @hidden
+     */
+    protected readonly _textLineWidths = computed(() => {
```
InnaAtanasova: same for this one. It's used in the template. We prefix with underscore only private properties.

libs/core/skeleton/components/skeleton.component.ts
```diff
@@ -1,110 +1,153 @@
...
+     * Computed dimensions (width and height) based on skeleton type.
+     * @hidden
+     */
+    protected readonly _dimensions = computed(() => {
+        const currentType = this.type();
+        const currentWidth = this.width();
+        const currentHeight = this.height();
+
+        if (currentType === 'text') {
+            const lines = this.textLines() || 1;
```
MariaIDineva: This fallback to 1 should never be needed since `textLines` has a default value set to `3`

libs/core/skeleton/components/skeleton.component.ts
```diff
@@ -1,110 +1,153 @@
...
+            return {
+                width: currentWidth ?? '100%',
+                height: currentHeight ?? (lines > 1 ? `${20 * lines}px` : '8px')
```
MariaIDineva: This magic number `20` can be a protected member of the class, say LINE_HEIGHT, which can also be used in the template (line 19).

MariaIDineva:
Same comment as the one in title component PR. Maybe we sould replace the import of the module with the component import since we should encourage standalone now and we have deprecated the module.
Screenshot 2026-01-06 at 10 40 14

droshev:
> Same comment as the one in title component PR. Maybe we sould replace the import of the module with the component import since we should encourage standalone now and we have deprecated the module. Screenshot 2026-01-06 at 10 40 14

That can be done once we merge the refactoring of `import.component`

libs/core/skeleton/components/skeleton.component.ts
droshev: what about cases (if someone passes) for example `<fd-skeleton type="text" [textLines]="0"></fd-skeleton>`, or `<fd-skeleton type="text" textLines=""></fd-skeleton>`?.

libs/core/skeleton/components/skeleton.component.ts
droshev: ✅ Done.

libs/core/skeleton/components/skeleton.component.ts
droshev: Refactored

libs/core/skeleton/components/skeleton.component.ts
droshev: refactored

InnaAtanasova (approved):

Issue #13702, opened 2026-01-05
Migration to Signals


InnaAtanasova:
List the component you are working on in the following pattern:
- [ ] Component name - Person - PR Link (when ready)
--------------------------------------------------------------------------------

- [ ] Text - Deno - [PR](https://github.com/SAP/fundamental-ngx/pull/13696)
- [ ] Title - Deno - [PR](https://github.com/SAP/fundamental-ngx/pull/13698)
- [ ] Skeleton - Deno - [PR](https://github.com/SAP/fundamental-ngx/pull/13699)

InnaAtanasova:
Closing this as there are new issues that place components in phases based on their dependency on each other.

PR #13719, opened 2026-01-13
fix(core): make object number use signals + adopt styles breaking changes

MariaIDineva:
## Related Issue(s)

## Description

1. Migrates `ObjectNumber` component to use latest Angular syntax with signals.

2. `ObjectNumber` implements `HasElementRef` interface (which it used to implement as inherited from `CssClassBuilder`) and keeps exposing `elementRef` for backwards compatiblity. 
3. Removes `class` input as it is no longer needed - any custom classes passed by the app get preserved with the new way that we add additional classes directly to the host, without `HostBinding`.

4. Adopts the Breaking changes from Fund-styles related to a11y.

**NB!** All styles from Fund styles were copy-pasted into the component's css and are to be deleted once we update to latest Fund styles. It was not possible to copy only the `sr-only` class as there were[ quite a few other changes](https://github.com/SAP/fundamental-styles/pull/6082/changes).

As part of adopting the style changes:
  - Adds a screen reader text for the different states/statuses - `'negative' | 'critical' | 'positive' | 'informative'`:
```
<fd-object-number unit="EUR" status="negative" class="fd-object-number fd-object-number--negative fd-object-number--interactive" tabindex="0" role="button">
  <span class="fd-object-number__text> -2,000.00</span>
  <span class="fd-object-number__unit">EUR</span>
  <span class="fd-object-number__sr-only ng-star-inserted">Value State Error</span>
</fd-object-number>
```

- Adds a screen reader text when the Object Number is emphasized:
```
<fd-object-number unit="EUR" status="negative" class="fd-object-number fd-object-number--negative>
  <span class="fd-object-number__text fd-object-number__text--bold"> -2,000.00</span>
  <span class="fd-object-number__unit">EUR</span>
  <span class="fd-object-number__sr-only ng-star-inserted">Emphasized</span>
</fd-object-number>
```
- Adds two new states to Object Number - **interactive** and **inverted** which can also be combined. Updates the examples to display those.
- Adds `role="button"` and `tabindex=0` to interactive object numbers.

- Adds translation for the SR span element content

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.59.0",
+    "version": "0.59.1-rc.0",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.59.0",
+    "version": "0.59.1-rc.0",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/cdk/package.json
+++ b/libs/cdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cdk",
-    "version": "0.59.0",
+    "version": "0.59.1-rc.0",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - CDK",
     "license": "Apache-2.0",
--- a/libs/core/package.json
+++ b/libs/core/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/core",
-    "version": "0.59.0",
+    "version": "0.59.1-rc.0",
     "schematics": "./schematics/collection.json",
     "ng-update": {
         "migrations": "./schematics/migrations.json"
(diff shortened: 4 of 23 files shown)


libs/core/object-number/object-number.component.ts
```diff
@@ -17,93 +16,105 @@ type ObjectStatus = 'negative' | 'critical' | 'positive' | 'informative';
...
 
     /** Id of the element that labels object number. */
-    @Input()
-    ariaLabelledBy: Nullable<string>;
+    readonly ariaLabelledBy = input<string | null>();
 
     /** Aria label for the object number. */
-    @Input()
-    ariaLabel: Nullable<string>;
+    readonly ariaLabel = input<string | null>();
```
InnaAtanasova: these 2 can be also undefined. If the user don't pass anything, the sting will be undefined.

libs/core/object-number/object-number.component.ts
```diff
@@ -17,93 +16,105 @@ type ObjectStatus = 'negative' | 'critical' | 'positive' | 'informative';
...
+        }
+
+        const customClass = this.class();
+        if (customClass) {
+            classes.push(customClass);
+        }
+
+        return classes.join(' ');
+    });
+
```
InnaAtanasova, line 97: I think you can simplify this and get rid of the if blocks with this:
```

/** @hidden Computed CSS classes */
    protected readonly _cssClass = computed(() =>
        [
            'fd-object-number',
            this.large() ? 'fd-object-number--large' : '',
            this.interactive() ? 'fd-object-number--interactive' : '',
            this.inverted() ? 'fd-object-number--inverted' : '',
            this.status() ? `fd-object-number--${this.status()}` : '',
            this.class()
        ]
            .filter(Boolean)
            .join(' ')
    );
```

PR #13731, opened 2026-01-14
fix(core): migrate Object Identifier to signals

InnaAtanasova:
## Related Issue(s)

## Description

- Migrates ObjectIdentifierComponent to signals
- Deprecates ObjectIdentifierModule
- Adopts the Breaking changes from Fund-styles related to a11y
- adds translation for the SR span element content

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,4 @@
 {
-    "editor.formatOnSave": true,
+    "editor.formatOnSave": false,
     "eslint.validate": ["json"]
 }
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.58.0-rc.98",
+    "version": "0.58.0-rc.100",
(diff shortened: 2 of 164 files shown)


libs/core/object-identifier/object-identifier.component.ts
```diff
@@ -1,77 +1,110 @@
...
-export class ObjectIdentifierComponent implements AfterContentInit {
-    /** Description text */
-    @Input()
-    description: Nullable<string>;
+export class ObjectIdentifierComponent {
+    /**
+     * Object Identifier ID
+     * If not set, a default value is provided
+     */
+    id = input('fd-object-identifier-' + ++objectIdentifierId);
```
MariaIDineva: Could this be `protected readonly` as well?

MariaIDineva: No, that's an input, of course it should be public. Please disregard.
But that means that the rest of the inputs should be public as well, right?

InnaAtanasova: yes, they (the inputs) should be public and readonly. Only the computed ones should be protected. I will update. Thanks!

PR #13735, opened 2026-01-16
fix(core): signalify layout grid

MariaIDineva:
## Related Issue(s)

## Description

Migrates to signals:
- `LayoutGridComponent`
- `LayoutGridColDirective`

Replaces all imports of deprecated `LayoutGridModule` with imports of `LayoutGridComponent`, `LayoutGridRowDirective` and `LayoutGridColDirective`.

Fixes unit tests and adds some more to cover the functionality.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,6 +1,6 @@
-name: Feature, Bug or Enhancement Report
-description: Report a bug, request an enhancement or feature.
-labels: []
+name: Bug Report
+description: Report a bug.
+labels: ['bug']
 body:
     - type: markdown
       attributes:
@@ -11,7 +11,7 @@ body:
       id: proposal-description
       attributes:
           label: Describe your proposal
-          description: Please provide a detailed description of your proposal, bug, or feature request.
+          description: Please provide a detailed description of your proposal, bug.
       validations:
           required: true
 
@@ -20,9 +20,16 @@ body:
       attributes:
           label: Which versions of Angular and Fundamental Library for Angular are affected?
           description: Specify the exact versions. (If requesting a feature, use current version.)
-          placeholder: 'Angular: x.y.z, Fundamental Library: x.y.z'
+          placeholder: 'Angular: x.y.z, Fundamental Library: x.y.z, UI5 Web Components: x.y.z'
       validations:
           required: true
+
+    - type: input
+      id: component
+      attributes:
+          label: Affected Component
+      validations:
+          required: false
 
     - type: textarea
       id: reproduction-steps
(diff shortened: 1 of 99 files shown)


libs/core/layout-grid/directives/layout-grid-col.directive.spec.ts
```diff
@@ -1,38 +1,47 @@
...
-    @ViewChild(LayoutGridColDirective, { static: true })
-    directiveElement: LayoutGridColDirective;
+    readonly fdLayoutGridCol = input<number | undefined>(undefined);
+    readonly colMd = input<number | undefined>(undefined);
+    readonly colLg = input<number | undefined>(undefined);
+    readonly colXl = input<number | undefined>(undefined);
+    readonly colOffset = input<number | undefined>(undefined);
+    readonly colOffsetMd = input<number | undefined>(undefined);
+    readonly colOffsetLg = input<number | undefined>(undefined);
+    readonly colOffsetXl = input<number | undefined>(undefined);
```
InnaAtanasova, line 40: why not
```
readonly fdLayoutGridCol = input<number | undefined>();
readonly colMd = input<number | undefined>();
readonly colLg = input<number | undefined>();
readonly colXl = input<number | undefined>();
readonly colOffset = input<number | undefined>();
readonly colOffsetMd = input<number | undefined>();
readonly colOffsetLg = input<number | undefined>();
readonly colOffsetXl = input<number | undefined>();
```
?

libs/core/layout-grid/directives/layout-grid-col.directive.ts
```diff
@@ -1,86 +1,79 @@
...
+     * @hidden
+     * Access to the host element's ElementRef.
+     *
+     * Use this for advanced scenarios like:
+     * - DOM measurements (dimensions, scroll position)
+     * - Focus management
+     * - Integration with third-party libraries requiring native elements
+     *
+     * For most use cases, prefer using directive inputs and outputs instead
+     * of direct DOM manipulation.
```
InnaAtanasova: you can remove this, it won't be shown to the user as we have @hidden. "Access to the host element's ElementRef." is enough in my opinion.

libs/core/layout-grid/layout-grid.component.ts
```diff
@@ -22,59 +11,50 @@ import { LayoutGridRowDirective } from './directives/layout-grid-row.directive';
...
+        }
 
-    /** @hidden */
-    ngOnChanges(): void {
-        this.buildComponentCssClass();
-    }
+        const customClass = this.class();
+        if (customClass) {
+            classes.push(customClass);
+        }
```
InnaAtanasova: Please double check, but I think this is not needed, which also means you can remove the input.

InnaAtanasova: This also would mean you can bind the rest of the classes in the host obj of the decorator.

libs/core/layout-grid/directives/layout-grid-col.directive.spec.ts
InnaAtanasova: ohm, this is the test, you can ignore my comment

libs/core/layout-grid/layout-grid.component.ts
InnaAtanasova: One hour later :D 
I tested and it turns out you really don't need the custom class. You can move all bindings to the host obj and get rid of the class in
layout-grid-2
layout-grid
put.

MariaIDineva: I tested and removing the class input seems to be fine. I applied the changes.

Issue #13384, opened 2025-07-09
[fd-combobox] Search term not showing the value User selects


emadaldin-baybars:
### Describe your proposal

for fd-combo, It should load the full text value after auto selecting the wanted item to well represent the selected value for the consumer 

### Which versions of Angular and Fundamental Library for Angular are affected?

"@angular/core": "^18.2.13","@fundamental-ngx/core": "~0.54.0",

### Steps to Reproduce (for bugs)

Assume we have array of objects [{id, value}]
- Write in the input field the search term you want to pick.
- Assuming you will find it and it's the first value
- Tap on `tab` key
Expected: the first value to be selected 
- remove one or two chars 
- Tap on `tab` key
Expected: to have the value selected and its full text loaded in the search term of the fd-combo 
Actual: the search term after removing chars will be there and Id of the value is selected 

### StackBlitz example(s)

[ Stackplitz ](https://stackblitz.com/edit/fhkhmmgv-m5csgl24)

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

emadaldin-baybars:
Hello Friends 👋🏻 

Any updates here regarding this ticket ?

droshev:
@emadaldin-baybars Do you think you can provide more background with the issue you are trying to address in your application? It is not clear what the expected behavior should be.

emadaldin-baybars:
> [@emadaldin-baybars](https://github.com/emadaldin-baybars) Do you think you can provide more background with the issue you are trying to address in your application? It is not clear what the expected behavior should be.

Hello @droshev 

Kindly check the example in [Here](https://stackblitz.com/edit/fhkhmmgv-m5csgl24) 

When you search for one item label, the result wanted item label should be presented in the input box which is already what happens when you search in the first time. But when you remove 2 chars from the result and tap on the `Tab` key, the result item label does not load 

Kindly check the record below 
https://github.com/user-attachments/assets/613f5597-5fcc-4d70-9203-ead7371693b2

PR #13762, opened 2026-01-20
feat(core,platform): search field complex results

### LINKED ISSUE #13563: Implement shell search with new search results list
### Describe your proposal

fundamental-styles has just added shell search bar improvements with new styles for the search results list. These need to be brought to fundamental-ngx

https://github.com/SAP/fundamental-styles/pull/6180

Angular 20, fundamental-ngx 0.57.*

_No response_

### StackBlitz example(s)

_No response_

### Declaration

mikerodonnell89:
fixes #13563 

Adds new features for the search field's results dropdown

**Breaking Changes:**
Search field now uses input signals, model signals, etc. This means breaking changes for developers when accessing any of the preexisting output events, inputs, viewChildren, etc.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.61.2-rc.8",
+    "version": "0.61.2-rc.9",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.61.2-rc.8",
+    "version": "0.61.2-rc.9",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/cdk/package.json
+++ b/libs/cdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cdk",
-    "version": "0.61.2-rc.8",
+    "version": "0.61.2-rc.9",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - CDK",
     "license": "Apache-2.0",
--- a/libs/cdk/utils/directives/template/template.directive.spec.ts
+++ b/libs/cdk/utils/directives/template/template.directive.spec.ts
@@ -1,14 +1,13 @@
-import { Component, ViewChild } from '@angular/core';
+import { Component, viewChild } from '@angular/core';
 import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
 import { TemplateDirective } from './template.directive';
 
 @Component({
-    standalone: true,
     imports: [TemplateDirective],
     template: '<ng-template fdkTemplate="Header">Template content</ng-template>'
 })
 class TestComponent {
-    @ViewChild(TemplateDirective) templateDirectiveRef: TemplateDirective;
+    readonly templateDirectiveRef = viewChild.required(TemplateDirective);
 }
 
 describe('TemplateDirective', () => {
@@ -29,10 +28,10 @@ describe('TemplateDirective', () => {
 
     it('should create an instance of the component and directive', () => {
         expect(component).toBeTruthy();
-        expect(component.templateDirectiveRef).toBeTruthy();
+        expect(component.templateDirectiveRef()).toBeTruthy();
     });
 
     it('should return the name of the template', () => {
-        expect(component.templateDirectiveRef.getName()).toBe('Header');
+        expect(component.templateDirectiveRef().getName()).toBe('Header');
     });
 });
--- a/libs/cdk/utils/directives/template/template.directive.ts
+++ b/libs/cdk/utils/directives/template/template.directive.ts
@@ -1,19 +1,17 @@
-import { Directive, Input, TemplateRef } from '@angular/core';
+import { Directive, TemplateRef, inject, input } from '@angular/core';
 
 @Directive({
-    selector: '[fdkTemplate]',
-    standalone: true
+    selector: '[fdkTemplate]'
 })
 export class TemplateDirective {
     /** Name of the template */
-    @Input('fdkTemplate')
-    name: string;
+    readonly name = input('', { alias: 'fdkTemplate' });
 
     /** @hidden */
-    constructor(public templateRef: TemplateRef<any>) {}
+    readonly templateRef = inject<TemplateRef<any>>(TemplateRef);
 
     /** @hidden */
     getName(): string {
-        return this.name;
+        return this.name();
     }
 }
--- a/libs/core/dialog/base/dialog-footer-base.class.ts
+++ b/libs/core/dialog/base/dialog-footer-base.class.ts
@@ -25,7 +25,7 @@ export abstract class DialogFooterBase {
      * This template uses the predefined Bar layout.
      */
     readonly footerTemplate = computed<TemplateRef<any> | undefined>(
-        () => this.customTemplates().find((t) => t.name === 'footer')?.templateRef
+        () => this.customTemplates().find((t) => t.name() === 'footer')?.templateRef
     );
 
     /**
@@ -35,7 +35,7 @@ export abstract class DialogFooterBase {
      * This template gives full control to the user without any predefined layout.
      */
     readonly customFooterTemplate = computed<TemplateRef<any> | undefined>(
-        () => this.customTemplates().find((t) => t.name === 'customFooter')?.templateRef
+        () => this.customTemplates().find((t) => t.name() === 'customFooter')?.templateRef
     );
 
     /** @hidden */
--- a/libs/core/dialog/base/dialog-header-base.class.ts
+++ b/libs/core/dialog/base/dialog-header-base.class.ts
@@ -24,7 +24,7 @@ export abstract class DialogHeaderBase {
      * Returns the TemplateRef when a template with name='header' is projected.
      */
     readonly headerTemplate = computed<TemplateRef<any> | undefined>(
-        () => this.customTemplates().find((t) => t.name === 'header')?.templateRef
+        () => this.customTemplates().find((t) => t.name() === 'header')?.templateRef
     );
 
     /**
@@ -33,6 +33,6 @@ export abstract class DialogHeaderBase {
      * Returns the TemplateRef when a template with name='subheader' is projected.
      */
     readonly subHeaderTemplate = computed<TemplateRef<any> | undefined>(
-        () => this.customTemplates().find((t) => t.name === 'subheader')?.templateRef
+        () => this.customTemplates().find((t) => t.name() === 'subheader')?.templateRef
     );
 }
--- a/libs/core/dialog/dialog.module.ts
+++ b/libs/core/dialog/dialog.module.ts
@@ -1,5 +1,5 @@
 import { NgModule } from '@angular/core';
-import { TemplateModule } from '@fundamental-ngx/cdk/utils';
+import { TemplateDirective } from '@fundamental-ngx/cdk/utils';
 import {
     BarComponent,
     BarElementDirective,
@@ -51,7 +51,7 @@ const declarations = [
     imports: [
         declarations,
         // Keeping these two items for backwards compatibility.
-        TemplateModule,
+        TemplateDirective,
         BarComponent,
         BarLeftDirective,
         BarMiddleDirective,
@@ -63,7 +63,7 @@ const declarations = [
     exports: [
         declarations,
         // Keeping these two items for backwards compatibility.
-        TemplateModule,
+        TemplateDirective,
         BarComponent,
         BarLeftDirective,
         BarMiddleDirective,
--- a/libs/core/menu/menu-mobile/menu-mobile.component.ts
+++ b/libs/core/menu/menu-mobile/menu-mobile.component.ts
@@ -16,7 +16,7 @@ import {
     ViewEncapsulation
 } from '@angular/core';
 import { outputToObservable, takeUntilDestroyed } from '@angular/core/rxjs-interop';
-import { InitialFocusDirective, RtlService, TemplateDirective, TemplateModule } from '@fundamental-ngx/cdk/utils';
+import { InitialFocusDirective, RtlService, TemplateDirective } from '@fundamental-ngx/cdk/utils';
 import {
     BarElementDirective,
     BarLeftDirective,
@@ -45,7 +45,6 @@ import { MenuService } from '../services/menu.service';
     changeDetection: ChangeDetectionStrategy.OnPush,
     encapsulation: ViewEncapsulation.None,
     imports: [
-        TemplateModule,
         BarLeftDirective,
         BarRightDirective,
         BarElementDirective,
--- a/libs/core/message-box/message-box.module.ts
+++ b/libs/core/message-box/message-box.module.ts
@@ -1,5 +1,5 @@
 import { NgModule } from '@angular/core';
-import { DynamicComponentService, InitialFocusModule, TemplateModule } from '@fundamental-ngx/cdk/utils';
+import { DynamicComponentService, InitialFocusModule, TemplateDirective } from '@fundamental-ngx/cdk/utils';
 import { TitleComponent } from '@fundamental-ngx/core/title';
 import { MessageBoxTemplateDirective } from './directives/message-box-template.directive';
 import { MessageBoxBodyComponent } from './message-box-body/message-box-body.component';
@@ -27,8 +27,8 @@ const declarations = [
  * Use direct imports of components and directives.
  */
 @NgModule({
-    imports: [TemplateModule, InitialFocusModule, TitleComponent, declarations],
-    exports: [declarations, TemplateModule, InitialFocusModule, TitleComponent],
+    imports: [TemplateDirective, InitialFocusModule, TitleComponent, declarations],
+    exports: [declarations, TemplateDirective, InitialFocusModule, TitleComponent],
     providers: [MessageBoxService, DynamicComponentService]
 })
 export class MessageBoxModule {}
--- a/libs/core/multi-combobox/mobile/mobile-multi-combobox.component.ts
+++ b/libs/core/multi-combobox/mobile/mobile-multi-combobox.component.ts
@@ -2,7 +2,7 @@ import { NgTemplateOutlet } from '@angular/common';
 import { Component, Inject, OnInit, TemplateRef, ViewChild } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { SelectableOptionItem } from '@fundamental-ngx/cdk/forms';
-import { Nullable, TemplateDirective, TemplateModule } from '@fundamental-ngx/cdk/utils';
+import { Nullable, TemplateDirective } from '@fundamental-ngx/cdk/utils';
 import { BarElementDirective, BarMiddleDirective, ButtonBarComponent } from '@fundamental-ngx/core/bar';
 import { ButtonComponent } from '@fundamental-ngx/core/button';
 import {
@@ -23,7 +23,6 @@ import { MULTI_COMBOBOX_COMPONENT } from '../multi-combobox.token';
     templateUrl: './mobile-multi-combobox.component.html',
     styleUrl: './mobile-multi-combobox.component.scss',
     imports: [
-        TemplateModule,
         BarMiddleDirective,
         BarElementDirective,
         ButtonBarComponent,
--- a/libs/core/multi-combobox/multi-combobox.component.ts
+++ b/libs/core/multi-combobox/multi-combobox.component.ts
@@ -1113,7 +1113,7 @@ export class MultiComboboxComponent<T = any> extends BaseMultiCombobox<T> implem
      */
     private _assignCustomTemplates(): void {
         this.customTemplates().forEach((template) => {
-            switch (template.name) {
+            switch (template.name()) {
                 case 'optionItemTemplate':
                     this.optionItemTemplate = template.templateRef;
                     break;
--- a/libs/core/multi-input/multi-input-mobile/multi-input-mobile.component.ts
+++ b/libs/core/multi-input/multi-input-mobile/multi-input-mobile.component.ts
@@ -9,7 +9,7 @@ import {
     ViewEncapsulation
 } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
-import { Nullable, TemplateDirective, TemplateModule } from '@fundamental-ngx/cdk/utils';
+import { Nullable, TemplateDirective } from '@fundamental-ngx/cdk/utils';
 import { BarElementDirective, BarMiddleDirective, ButtonBarComponent } from '@fundamental-ngx/core/bar';
 import { ButtonComponent } from '@fundamental-ngx/core/button';
 import {
@@ -30,13 +30,12 @@ import { MULTI_INPUT_COMPONENT, MultiInputInterface } from '../multi-input.inter
     changeDetection: ChangeDetectionStrategy.OnPush,
     encapsulation: ViewEncapsulation.None,
     imports: [
-        TemplateModule,
+        TemplateDirective,
         BarMiddleDirective,
         BarElementDirective,
         ButtonBarComponent,
         TitleComponent,
         TitleComponent,
-        TemplateDirective,
         BarMiddleDirective,
         BarElementDirective,
         ButtonBarComponent,
--- a/libs/core/package.json
+++ b/libs/core/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/core",
-    "version": "0.61.2-rc.8",
+    "version": "0.61.2-rc.9",
     "schematics": "./schematics/collection.json",
     "ng-update": {
         "migrations": "./schematics/migrations.json"
--- a/libs/core/popover/popover-mobile/popover-mobile.component.ts
+++ b/libs/core/popover/popover-mobile/popover-mobile.component.ts
@@ -11,7 +11,7 @@ import {
 } from '@angular/core';
 
 import { NgTemplateOutlet } from '@angular/common';
-import { TemplateModule } from '@fundamental-ngx/cdk/utils';
+import { TemplateDirective } from '@fundamental-ngx/cdk/utils';
 import {
     DialogBodyComponent,
     DialogCloseButtonComponent,
@@ -32,7 +32,7 @@ let mobilePopoverUniqueId = 0;
     changeDetection: ChangeDetectionStrategy.OnPush,
     encapsulation: ViewEncapsulation.None,
     imports: [
-        TemplateModule,
+        TemplateDirective,
         TitleComponent,
         NgTemplateOutlet,
         DialogComponent,
--- a/libs/cx/package.json
+++ b/libs/cx/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cx",
-    "version": "0.61.2-rc.8",
+    "version": "0.61.2-rc.9",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - cx",
     "license": "Apache-2.0",
--- a/libs/datetime-adapter/package.json
+++ b/libs/datetime-adapter/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/datetime-adapter",
-    "version": "0.61.2-rc.8",
+    "version": "0.61.2-rc.9",
     "description": "Datetime adapter for SAP Fundamentals, based on Day.js package",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/i18n/package.json
+++ b/libs/i18n/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/i18n",
-    "version": "0.61.2-rc.8",
+    "version": "0.61.2-rc.9",
     "description": "Fundamental Library for Angular - i18n",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/moment-adapter/package.json
+++ b/libs/moment-adapter/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/moment-adapter",
-    "version": "0.61.2-rc.8",
+    "version": "0.61.2-rc.9",
     "description": "Datetime adapter for SAP Fundamentals, based on Moment.js package",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/platform/approval-flow/approval-flow-node/approval-flow-node.component.ts
+++ b/libs/platform/approval-flow/approval-flow-node/approval-flow-node.component.ts
@@ -347,7 +347,7 @@ export class ApprovalFlowNodeComponent implements OnInit, OnChanges, OnDestroy {
 
     /** @hidden */
     _onMenuOpen(): void {
-        this._menu.refreshPosition();
+        this._menu?.refreshPosition();
     }
 
     /** @hidden */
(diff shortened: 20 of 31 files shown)


libs/core/title/title.component.ts
```diff
@@ -60,14 +60,18 @@ export abstract class TitleToken {
...
+        '[class.fd-title--two-line-clamp]': 'twoLineClamp()'
     },
     styleUrl: './title.component.scss',
     changeDetection: ChangeDetectionStrategy.OnPush,
     encapsulation: ViewEncapsulation.None,
     providers: [{ provide: TitleToken, useExisting: TitleComponent }]
 })
 export class TitleComponent extends TitleToken {
+    /** Whether this title should clamp at two lines of text. */
+    twoLineClamp = input(false);
```
InnaAtanasova: this has to be `readonly twoLineClamp = input(false, { transform: booleanAttribute });
`

libs/core/text/text.component.ts
```diff
@@ -60,6 +60,9 @@ export class TextComponent {
      */
     readonly isCollapsed = model(true);
 
+    /** @hidden */
+    subline = input(false);
```
InnaAtanasova: this has to be `readonly subline = input(false, { transform: booleanAttribute });`

libs/core/list/list-item/list-item.component.ts
```diff
@@ -181,6 +182,9 @@ export class ListItemComponent<T = any> extends ListFocusItem<T> implements Afte
     /** Template ref for Settings list item */
     settingsListTpl = input<TemplateRef<any>>();
 
+    /** @hidden Whether this is a suggestion type list item. */
+    suggestion = input(false);
```
InnaAtanasova: this has to be `readonly suggestion = input(false, { transform: booleanAttribute });`

libs/core/list/directives/list-title.directive.ts
```diff
@@ -15,6 +16,9 @@ export class ListTitleDirective implements OnInit {
     /** Whether the text should truncate with ellipsis. */
     readonly truncate = input(false, { transform: booleanAttribute });
 
+    /** Whether this title is used for scope in the shell search results. */
+    scope = input(false);
```
InnaAtanasova: this has to be `readonly scope = input(false, { transform: booleanAttribute });`

libs/core/list/directives/list-info-label.directive.ts
```diff
@@ -0,0 +1,11 @@
+import { Directive, HostBinding } from '@angular/core';
+
+@Directive({
+    selector: '[fd-list-info-label], [fdListInfoLabel]',
+    standalone: true
```
InnaAtanasova: please remove this, not needed as it's default now

libs/core/list/directives/list-info-label.directive.ts
```diff
@@ -0,0 +1,11 @@
...
+})
+export class ListInfoLabelDirective {
+    /** @hidden */
+    @HostBinding('class.fd-list__info-label')
+    fdListInfoLabel = true;
```
InnaAtanasova: please do the binding in the decorator

libs/core/bar/directives/bar-right.directive.ts
```diff
@@ -7,8 +7,12 @@ import { Directive } from '@angular/core';
...
-        class: 'fd-bar__right'
+        class: 'fd-bar__right',
+        '[class.fd-bar__right--stack-contents-vertically]': 'stackContentsVertically()'
     },
     standalone: true
 })
-export class BarRightDirective {}
+export class BarRightDirective {
+    /** Whether to stack the contents of this portion of the bar vertically. */
+    stackContentsVertically = input(false);
```
InnaAtanasova: this has to be `readonly stackContentsVertically = input(false, { transform: booleanAttribute });`

libs/core/bar/directives/bar-middle.directive.ts
```diff
@@ -7,8 +7,12 @@ import { Directive } from '@angular/core';
...
-        class: 'fd-bar__middle'
+        class: 'fd-bar__middle',
+        '[class.fd-bar__middle--stack-contents-vertically]': 'stackContentsVertically()'
     },
     standalone: true
 })
-export class BarMiddleDirective {}
+export class BarMiddleDirective {
+    /** Whether to stack the contents of this portion of the bar vertically. */
+    stackContentsVertically = input(false);
```
InnaAtanasova: needs `booleanAttribute` transform

libs/core/bar/directives/bar-left.directive.ts
```diff
@@ -7,8 +7,12 @@ import { Directive } from '@angular/core';
...
-        class: 'fd-bar__left'
+        class: 'fd-bar__left',
+        '[class.fd-bar__left--stack-contents-vertically]': 'stackContentsVertically()'
     },
     standalone: true
 })
-export class BarLeftDirective {}
+export class BarLeftDirective {
+    /** Whether to stack the contents of this portion of the bar vertically. */
+    stackContentsVertically = input(false);
```
InnaAtanasova: needs `booleanAttribute` transform

libs/core/bar/bar.component.ts
```diff
@@ -76,6 +78,12 @@ export class BarComponent implements OnChanges, OnInit, CssClassBuilder, OnDestr
      */
     role = input('toolbar');
 
+    /** Whether this bar is to be used for the search results initial suggestion title. */
+    initialSuggestionTitle = input(false);
+
+    /** Whether this bar is to be used for the search results initial suggestion subline. */
+    initialSuggestionSubline = input(false);
```
InnaAtanasova: needs `booleanAttribute` transform

InnaAtanasova:
The specs specify that there should be an autocomplete feature.
Screenshot 2026-01-20 at 1 33 46 PM

InnaAtanasova:
Screenshot 2026-01-20 at 1 28 23 PM

from design specs:
Screenshot 2026-01-20 at 1 28 33 PM

InnaAtanasova:
Screenshot 2026-01-20 at 1 42 11 PM
In the documentation we should show the user how to configure these cases.

InnaAtanasova:
Screenshot 2026-01-20 at 1 43 17 PM
need in the documentation Busy State case (maybe trigger it with "Show All Search Results").

InnaAtanasova:
Screenshot 2026-01-20 at 1 44 38 PM
Screenshot 2026-01-20 at 1 44 44 PM
When an item is selected it should be marked as such in the list.

InnaAtanasova:
https://github.com/user-attachments/assets/fe2ace7f-c978-4280-b4d2-4a604603abc8

Some keyboard event issues:
1. going from the input to the list, the focus goes to a random item, not the first focusable element
2. up and down arrow keys are inconsistent
3. The focus is lost after "Show All Search Results" button focus.

InnaAtanasova:
(from the specs) In mobile mode, when there's no search term in the input field, the list items show the delete button. It's not shown when an item is selected
Screenshot 2026-01-20 at 2 03 41 PM
Screenshot 2026-01-20 at 2 03 37 PM

InnaAtanasova:
Prob out of scope for this PR, but these need "See More/See Less" buttons:
Screenshot 2026-01-20 at 2 19 50 PM

Specs:
Screenshot 2026-01-20 at 2 20 20 PM

InnaAtanasova:
From the specs:
Screenshot 2026-01-20 at 2 24 23 PM
Steps 4 and 5 are missing. When the focus is on the input field, the list is not shown. It is shown only pressing the Space key.

InnaAtanasova:
https://github.com/user-attachments/assets/d66777ff-abc3-43a5-a83c-bf96b21123bc

The view is not updated when the categories change.

libs/core/text/text.component.html
```diff
@@ -1,4 +1,9 @@
-<p class="fd-text" [class.fd-text--pre-wrap]="whitespaces()" [style.hyphens]="hyphenation()">
+<p
+    class="fd-text"
+    [class.fd-text__subline]="subline"
```
MariaIDineva: That should be `subline()`

libs/docs/core/shellbar/examples/shellbar-complex-search-results-example.component.ts
```diff
@@ -0,0 +1,302 @@
...
+        FormsModule,
+        SegmentedButtonComponent,
+        IllustratedMessageModule
+    ]
+})
+export class ShellbarComplexSearchResultsExampleComponent implements OnInit {
+    @ViewChild(UserMenuComponent)
+    userMenuComponent: UserMenuComponent;
+
+    @ViewChild(SearchFieldComponent)
```
MariaIDineva: Just a suggestion, minor:
Maybe we can use signals for the fields in this example, eg. “viewChild” instead of “@ViewChild”.
Also, inject `MessageToastService` with a decorator instead of the constructor.

mikerodonnell89:
> The specs specify that there should be an autocomplete feature. 

Looking at our other components that use fdkAutoComplete, looks like this is a bug with that directive. I think to move things forward with this PR, let's consider that a separate issue

InnaAtanasova:
A11Y bug (keyboard): 
1. you can go from the input to the list with arrow down, but you can't go back to the input with keyboard.
2. In this scenario, when there are no results, you can't go to the list items with arrow down or tab
Screenshot 2026-01-23 at 12 06 19 PM

droshev:
> > The specs specify that there should be an autocomplete feature.
> 
> Looking at our other components that use fdkAutoComplete, looks like this is a bug with that directive. I think to move things forward with this PR, let's consider that a separate issue

UXC Lyra is an important delivery and we need to provide it 100%, also the designers will report it. Is it a problem addressing it in this PR?

droshev:
> > From what I can see in the design specs, in high contrast themes the shellbar search dropdown button must become blue upon selection and it stays yellow:
> 
> This would be out of the scope of this PR i think and needs to come from styles

That will be checked by the designers as well. And without their approval we won't be able to move forward.

InnaAtanasova:
Screenshot 2026-01-23 at 2 16 31 PM
In mobile mode the input should take the entire width and leave space only for the Cancel button.
The cancel button should not have the border.
There's no dialog footer aka the footer with the ok button should be removed.

mikerodonnell89:
> UXC Lyra is an important delivery and we need to provide it 100%, also the designers will report it. Is it a problem addressing it in this PR?

I am not sure how long this will take, but sure i can give it a go in this PR

mikerodonnell89:
> A11Y bug (keyboard):
> 
> 1. you can go from the input to the list with arrow down, but you can't go back to the input with keyboard.

You press escape which is how we've got most other similar components implemented with a few exceptions

mikerodonnell89:
> > > From what I can see in the design specs, in high contrast themes the shellbar search dropdown button must become blue upon selection and it stays yellow:
> > 
> > 
> > This would be out of the scope of this PR i think and needs to come from styles
> 
> That will be checked by the designers as well. And without their approval we won't be able to move forward.

Ok this actually just needed a small change in fd-ngx

libs/btp/jest.config.ts
```diff
@@ -1,4 +1,4 @@
-import baseConfig from '../../jest.config.base.ts';
+import baseConfig from '../../jest.config.base';
```
droshev: this will break the build. I changed it in [here](https://github.com/SAP/fundamental-ngx/pull/13599/changes/9891d0f4d8170812bf02711c32088cf1bbf48e87)

InnaAtanasova:
There are some console logs
Screenshot 2026-02-04 at 1 47 39 PM

InnaAtanasova:
Screenshot 2026-02-04 at 1 44 27 PM

In mobile mode there are a few problems:
1. When the mobile dialog is open, the input field should be on focus
2. the search input should take the entire width
3. missing the cancel button
4. Some css bug for selected state, the dark blue underline is missing
5. There are some errors in the console
6. Console.log 

The specs:
Screenshot 2026-02-04 at 1 43 58 PM

InnaAtanasova:
Keyboard navigation problem:
you can't go back to the search input using the keyboard

https://github.com/user-attachments/assets/2b992f1c-52ae-44ec-ab27-09de0bcfd18c

InnaAtanasova:
https://github.com/user-attachments/assets/39d4e7fd-7804-4eef-a035-adc714cdab4f

In mobile mode, once an item is selected you no longer can open the dropdown or go to the list items.

InnaAtanasova:
https://github.com/user-attachments/assets/453ab3e4-1010-4738-9c32-ceda53475a42

Tab or Enter should select Apple from the list of suggestions, instead Tab leaves the component. Enter just triggers the event, but does not mark the item as selected.

InnaAtanasova:
Screenshot 2026-02-04 at 2 05 35 PM
Screenshot 2026-02-04 at 2 04 36 PM
This element needs some a11y attributes.

InnaAtanasova:
Screenshot 2026-02-04 at 2 12 21 PM
Screenshot 2026-02-04 at 2 12 02 PM

The group is missing some a11y attributes.

libs/cdk/utils/directives/auto-complete/auto-complete.directive.ts
```diff
@@ -74,24 +73,14 @@ export class AutoCompleteDirective {
          * By ensuring that we set all properties we can proceed with stable data.
          */
         this._zone.runOutsideAngular(() => {
-            const keyupEvent = fromEvent<KeyboardEvent>(this._elementRef.nativeElement, 'keydown');
+            const keyupEvent = fromEvent<KeyboardEvent>(this._elementRef.nativeElement, 'keyup');
```
InnaAtanasova, line 79: This was recently changed because there was a bug. You can see this [PR](https://github.com/SAP/fundamental-ngx/pull/13483) and the corresponding issue:  https://github.com/SAP/fundamental-ngx/issues/13384

libs/platform/search-field/search-field.component.ts
```diff
@@ -259,10 +299,18 @@ export class SearchFieldComponent
...
+    advancedFilterButtonClick = new EventEmitter<void>();
+
     /** @hidden */
     @ViewChild('categoryDropdown', { static: false })
     categoryDropdown: PopoverComponent;
 
+    /** @hidden */
+    @ViewChild('categorySelectComponent', { static: false })
+    categorySelectComponent: SelectComponent;
+
```
InnaAtanasova: I know this component is getting very complicated and taking too long time but mixing old and new inputs, outputs, viewChild etc. is making the code very inconsistent. Also, migrating these to signals can lead to getting rid of ngOnInit, ngOnChanges by using computed signals.

libs/platform/search-field/search-field.component.ts
```diff
@@ -606,17 +825,110 @@ export class SearchFieldComponent
...
                 }
+                if (!suggestion?.children && suggestion?.value) {
+                    this._autoCompleteSuggestions.push(suggestion.value);
+                } else if (suggestion?.children?.length) {
+                    suggestion.children.forEach((child) => {
+                        this._autoCompleteSuggestions.push(child.value);
+                    });
+                }
             });
         });
```
InnaAtanasova, line 918: This method subscribes to observable synchronously inside a function, causing side effects and potential memory leaks. It creates a new subscription every time it's called, creates side effect by mutating `_autoCompleteSuggestions`. It can be changed to a computed signal. 
```
protected readonly suggestionsLength = computed(() => {
    const suggestions = this.dropdownValues();
    const input = this.inputText().trim().toLowerCase();
    
    let count = 0;
    const autoComplete: string[] = [];
    
    suggestions?.forEach((suggestion) => {
        const textToCheck = typeof suggestion === 'string' ? suggestion : suggestion?.value;
        if (input && textToCheck.toLowerCase().includes(input)) {
            count++;
        }
        if (!suggestion?.children && suggestion?.value) {
            autoComplete.push(suggestion.value);
        } else if (suggestion?.children?.length) {
            suggestion.children.forEach((child) => {
                autoComplete.push(child.value);
            });
        }
    });
    
    // Store autocomplete suggestions as signal too
    this.autoCompleteSuggestions.set(autoComplete);
    return count;
});

```

mikerodonnell89:
> Keyboard navigation problem: you can't go back to the search input using the keyboard
> 
>  Screen.Recording.2026-02-04.at.1.48.49.PM.mov

you can press escape

mikerodonnell89:
> Screen.Recording.2026-02-04.at.1.46.55.PM.mov 
> In mobile mode, once an item is selected you no longer can open the dropdown or go to the list items.

I'm not seeing this. Whenever I click the input, the mobile dialog pops up. Maybe there is a more specific set of steps to see what you're seeing in your video. Here's my experience:

https://github.com/user-attachments/assets/484736bb-5a1c-423d-97e5-7e381443767e

libs/platform/search-field/search-field.component.ts
mikerodonnell89: Unfortunately this component extends some components that are used throughout the entire library, so migrating this component to use signal inputs only means refactoring quite a few other components as well. I think we'll need to save that for a later date. But I've changed my signal inputs to old-school inputs for consistency. Should be easy to switch back

droshev:
I ran Claude review. Here are the top list of things for improvements:
## 🔴 Critical Issues

**1. Type Safety**
```typescript
// Replace any types
searchInScopeCallback?: () => any;  // ❌
searchInScopeCallback?: (item: SuggestionItem) => void | Promise<void>;  // ✅
```

**2. DOM Queries in Hot Path**
```
// ❌ Current - Expensive DOM queries
_inputValueMatchesFirstSuggestion(): boolean {
    return inputEl.value === 
        this.suggestionListEl.nativeElement.querySelector('.fd-list__item--suggestion')?.innerText;
}

private _setAriaOwns(): void {
    this.detectChanges(); // Manual CD!
    parentListEl?.querySelectorAll('.fd-list').forEach((list: HTMLElement) => {
        const visibleChildren = list.querySelectorAll('.fd-list__item--suggestion');
        // ...
    });
}

// ✅ Fix - Use data model
_inputValueMatchesFirstSuggestion(): boolean {
    const first = this.suggestionsFiltered[0] ?? null;
    return first ? this.inputText.toLowerCase() === first.value.toLowerCase() : false;
}

private _setAriaOwns(): void {
    const groups = this.suggestionsFiltered.filter(s => s.isGroupHeader);
    groups.forEach((group, gIndex) => {
        const childIds = group.children
            ?.map((_, cIndex) => `${this._menuId}-group-${gIndex}-item-${cIndex}`)
            .join(' ') ?? '';
        this.suggestionListEl.nativeElement.children[gIndex]?.setAttribute('aria-owns', childIds);
    });
}
```
**3. Anti-Patterns**

```
/// ❌ Fake events
const mockClick = new Event('click');
this._performButtonClick(mockClick, event.searchInScopeCallback);

// ❌ Side effects in pure functions
_autoCompleteDisplayFn = (event: SuggestionItem): string => {
    this._autoCompleteMostRecentSuggestionItem = event; // Side effect!
    return event.value;
};

// ✅ Fix - Direct invocation with error handling
if (event.searchInScopeCallback) {
    try {
        const result = event.searchInScopeCallback(event);
        if (result instanceof Promise) {
            result.catch(err => isDevMode() && console.error('[search-field]', err));
        }
    } catch (err) {
        isDevMode() && console.error('[search-field]', err);
    }
    return;
}

// ✅ Pure function
_autoCompleteDisplayFn = (item: SuggestionItem): string => item.value;
```
**4. Template Safety**

```
<!-- ❌ Current - Can pass undefined to pipe -->
@for (childSuggestion of filteredSuggestion?.children | suggestionMatches: inputText; track childSuggestion)

<!-- ✅ Fix -->
@for (childSuggestion of (filteredSuggestion.children ?? []) | suggestionMatches: inputText; track childSuggestion)
```

## 🟡 Improvements
Type Guards
```
private _isSuggestionItem(event: SuggestionItem | string): event is SuggestionItem {
    return typeof event === 'object' && event !== null && 'value' in event;
}
```
Production Safety
```
// Wrap all console statements
if (isDevMode()) {
    console.warn('[search-field] Warning message');
}
```
- Remove manual detectChanges() calls

## 🧪 Missing Tests
- Grouped suggestions rendering and filtering
- Search-in-scope callback execution
- Show more callback execution
- Accessibility (aria-owns, F2 key, roles)
- Error handling (callback failures, async rejections)

mikerodonnell89:
> I ran Claude review. Here are the top list of things for improvements:
> 
> ## 🔴 Critical Issues
> **1. Type Safety**
> 
> ```ts
> // Replace any types
> searchInScopeCallback?: () => any;  // ❌
> searchInScopeCallback?: (item: SuggestionItem) => void | Promise<void>;  // ✅
> ```

This is a developer-defined callback function, I don't think enforcing return types provides any benefit

mikerodonnell89:
> **2. DOM Queries in Hot Path**
> 
> ```
> // ❌ Current - Expensive DOM queries
> _inputValueMatchesFirstSuggestion(): boolean {
>     return inputEl.value === 
>         this.suggestionListEl.nativeElement.querySelector('.fd-list__item--suggestion')?.innerText;
> }
> // ✅ Fix - Use data model
> _inputValueMatchesFirstSuggestion(): boolean {
>     const first = this.suggestionsFiltered[0] ?? null;
>     return first ? this.inputText.toLowerCase() === first.value.toLowerCase() : false;
> }

Started working on this one but I think the current implementation is best. My idea for a fix was to get the list items via ViewChildren, then the title via ContentChild, but ViewChildren doesn't always return the elements in the order they appear in the dom. This function is only called once when the tab key is pressed so I don't think we need to worry about performance too much

libs/platform/search-field/search-field.component.ts
droshev: Can you migrate those @Input() Decorator → input() Signal

droshev: can you migrate @Input() with Getter/Setter → input() Signal?

libs/platform/search-field/search-field.component.ts
```diff
@@ -243,6 +280,10 @@ export class SearchFieldComponent
     @Input()
     disableSuggestionsFoundAnnouncer = false;
 
+    /** Whether to show the "All" option in the category select. */
+    @Input()
+    showCategoryAllOption = false;
+
     /** Input change event. */
     @Output()
     inputChange: EventEmitter<SearchInput> = new EventEmitter();
```
droshev: those @Output() with EventEmitter → output() Function

libs/platform/search-field/search-field.component.ts
```diff
@@ -279,31 +328,92 @@ export class SearchFieldComponent
     @ViewChild('suggestionMenuTemplate', { static: false })
     suggestionMenuTemplate: TemplateRef<any>;
 
+    /** @hidden */
+    @ViewChild('suggestionList')
+    suggestionList: ListComponent;
+
     /** @hidden */
     @ViewChildren(SearchFieldSuggestionDirective)
```
droshev: those @ViewChildren() Decorator → viewChildren() Signal

libs/platform/search-field/search-field.component.ts
droshev: those @HostListener() Decorator → host: {} Property

libs/cdk/utils/directives/auto-complete/auto-complete.directive.ts
InnaAtanasova: what did we decide for this?

libs/core/product-switch/product-switch/product-switch.component.ts
```diff
@@ -77,6 +79,9 @@ export class ProductSwitchComponent {
     /** Event emitted right before the popover is being opened. */
     readonly beforeOpen = output<void>();
 
+    /** @hidden */
+    _contentDensity: ContentDensityMode;
```
InnaAtanasova: This shouldn't be prefixed if it's not a private member.

libs/platform/search-field/search-field.component.ts
```diff
@@ -366,30 +397,82 @@ export class SearchFieldComponent
...
+            () => {
+                const isMobile = this.mobile();
+                if (isMobile) {
+                    // Defer to next tick to avoid running during construction
+                    queueMicrotask(() => {
+                        this._setUpMobileMode();
+                    });
+                }
+            },
+            { allowSignalWrites: true }
```
InnaAtanasova: This is deprecated and will produce console warnings.

libs/platform/search-field/search-field.component.ts
```diff
@@ -366,30 +397,82 @@ export class SearchFieldComponent
...
+            () => {
+                const items = this.suggestionItems();
+                // Use queueMicrotask to defer execution after view initialization
+                queueMicrotask(() => {
+                    if (items.length > 0) {
+                        this._resetKeyManager();
+                    }
+                });
+            },
+            { allowSignalWrites: true }
```
InnaAtanasova: This is deprecated and will produce console warnings.

libs/platform/search-field/search-field.component.ts
```diff
@@ -504,58 +609,65 @@ export class SearchFieldComponent
...
-        this._suggestionkeyManager = new FocusKeyManager(this.suggestionItems);
-        if (this._isOpen$()) {
+        this._isSuggestionMenuOpen$.set(true);
+        this._resetKeyManager();
+        if (this._isSuggestionMenuOpen$()) {
             return;
         }
 
-        this._isOpen$.set(true);
+        this._isSuggestionMenuOpen$.set(true);
```
InnaAtanasova:
```
openSuggestionMenu(): void {
    this._isSuggestionMenuOpen$.set(true);   // sets to true
    this._resetKeyManager();
    if (this._isSuggestionMenuOpen$()) {      // always true here!
        return;                               // always returns!
    }
    this._isSuggestionMenuOpen$.set(true);    // dead code, never reached
}

```

libs/platform/search-field/search-field.component.ts
```diff
@@ -613,16 +902,66 @@ export class SearchFieldComponent
...
+                    this._autoCompleteSuggestions.push(suggestion);
+                } else if (suggestion?.children?.length) {
+                    suggestion.children.forEach((child) => {
+                        this._autoCompleteSuggestions.push(child);
+                    });
+                }
             });
+
+            this._setAriaOwns();
         });
```
InnaAtanasova, line 918: What happened to this comment? I still see the memory/subscription leak problem.

`_getSuggestionsLength()` subscribes inside a synchronous method without cleanup
In **search-field.component.ts** (877-901), `_getSuggestionsLength()` calls `this._dropdownValues$.pipe(takeUntilDestroyed(...)).subscribe(...)` every time it's invoked. This method is called from `onValueChange()` and `_updateSearchAnnouncementText()` — potentially many times during the component's lifetime. Each call creates a new subscription that persists until destroy. This is a memory/subscription leak. The return value count is also assigned inside the async subscribe callback but returned synchronously, so it will always return 0 when `_dropdownValues$` is async.

libs/platform/search-field/search-field.component.ts
```diff
@@ -128,182 +147,179 @@ type Appearance = SearchComponent['appearance'] | undefined;
...
 
-    /** @hidden */
-    get searchFieldValue(): SearchInput {
-        return {
-            text: this.inputText,
-            category: this._currentCategory?.value || null
-        };
-    }
     /** @hidden Focus state */
     _isFocused = false;
```
InnaAtanasova: This should be changed to a signal as now this will be mutated imperatively. In a signal-based component with `OnPush`, these won't trigger change detection reliably.

libs/platform/search-field/search-field.component.ts
```diff
@@ -128,182 +147,179 @@ type Appearance = SearchComponent['appearance'] | undefined;
...
      */
-    _dropdownValues$: Observable<string[]> = of([]);
+    _dropdownValues$: Observable<SuggestionItem[]> = of([]);
 
     /**
      * Currently set category.
      * @hidden
      */
-    _currentCategory?: ValueLabelItem;
+    _currentCategory?: ValueLabelItem | null;
```
InnaAtanasova: This should be changed to a signal as now this will be mutated imperatively. In a signal-based component with `OnPush`, these won't trigger change detection reliably.

libs/platform/search-field/search-field.component.ts
```diff
@@ -326,6 +342,12 @@ export class SearchFieldComponent
...
     _clearId = '';
 
+    /** @hidden */
+    _popoverHeaderId = '';
+
+    /** @hidden */
+    _popoverBodyId = '';
+
     /** @hidden */
     isOpen = false;
```
InnaAtanasova, line 349: same. This should be changed to a signal as now this will be mutated imperatively. In a signal-based component with `OnPush`, these won't trigger change detection reliably.

libs/platform/search-field/search-field.component.ts
```diff
@@ -336,19 +358,28 @@ export class SearchFieldComponent
...
 
     /** @hidden */
-    private _categories: ValueLabelItem[];
+    _selectedSuggestionItemId: string;
+
+    /** @hidden */
+    _categorySelectDescription: string;
+
+    /** @hidden */
+    _actionButtonTabIndex: -1 | 0 = -1;
```
InnaAtanasova: These should be changed to signals.

libs/platform/search-field/search-field.component.ts
```diff
@@ -366,30 +397,82 @@ export class SearchFieldComponent
...
         public elementRef: ElementRef<HTMLElement>,
         private readonly _viewContainerRef: ViewContainerRef,
         private readonly _injector: Injector,
-        @Inject(DOCUMENT) private readonly _document: Document,
         private readonly _liveAnnouncer: LiveAnnouncer,
         readonly _dynamicComponentService: DynamicComponentService,
-        readonly contentDensityObserver: ContentDensityObserver
+        readonly contentDensityObserver: ContentDensityObserver,
+        @Optional() @Inject(FD_SHELLBAR_COMPONENT) protected _shellbar: ShellbarComponent
     ) {
```
InnaAtanasova, line 412: These should be using `inject()`

libs/platform/search-field/search-field.component.ts
```diff
@@ -336,19 +358,28 @@ export class SearchFieldComponent
...
+    _selectedSuggestionItemId: string;
+
+    /** @hidden */
+    _categorySelectDescription: string;
+
+    /** @hidden */
+    _actionButtonTabIndex: -1 | 0 = -1;
+
+    /** @hidden */
+    private _suggestions: SuggestionItem[] | Observable<SuggestionItem[]>;
```
InnaAtanasova: Is this used in the new code?

libs/platform/search-field/search-field.component.ts
```diff
@@ -366,30 +397,82 @@ export class SearchFieldComponent
     private readonly _dataSourceChanged$ = new Subject<void>();
 
     /** @hidden */
-    private _appearance: Appearance;
-
-    /** hidden */
-    private readonly _rtlService = inject(RtlService, { optional: true });
+    private _mobileComponent: ComponentRef<SearchFieldMobileComponent>;
```
InnaAtanasova, line 400: There's no cleanup in `ngOnDestroy()`. If the component is destroyed while mobile mode is active, the dynamic component leaks.

libs/platform/search-field/search-field.component.ts
```diff
@@ -336,19 +358,28 @@ export class SearchFieldComponent
     _isSearchDone = false;
 
     /** @hidden */
-    _isOpen$ = signal(false);
+    _isSuggestionMenuOpen$ = signal(false);
```
InnaAtanasova: This shouldn't have `$` if it's a signal, it's used for Observables.

libs/platform/search-field/search-field.component.html
```diff
@@ -127,17 +263,17 @@
                     }
                 </span>
             }
-            @if (!disableRefresh && !isLoading && (_isRefresh || _isSearchDone)) {
+            @if (!disableRefresh && !isLoading() && (_isRefresh || _isSearchDone)) {
```
InnaAtanasova: shouldn't `disableRefresh` be used as a signal `disableRefresh()`?

libs/platform/search-field/search-field.component.html
```diff
@@ -1,11 +1,104 @@
...
+    >
         <fd-popover-control>
             <ng-template [ngTemplateOutlet]="inputFieldTemplate"></ng-template>
         </fd-popover-control>
-        <fd-popover-body>
-            <ng-template [ngTemplateOutlet]="suggestionMenuTemplate"></ng-template>
+        <fd-popover-body
+            (keydown)="_handlePopoverKeydown($event)"
+            [style.display]="
+                (_dropdownValues$ | async | suggestionMatches: inputText()).length > 0 ||
```
InnaAtanasova: `| async | suggestionMatches: inputText()` is repeated 6 times in the template. Each invocation creates a separate async subscription and pipe evaluation. Maybe use a computed signal:

```
protected readonly _filteredSuggestions = computed(() => ...);
```

or use `@if (_dropdownValues$ | async | suggestionMatches: inputText(); as filtered)` at the top level and reuse `filtered`.

Issue #13740, opened 2026-01-19
Angular Signals Migration Phase 1


InnaAtanasova:
- [x] Icon (blocks 50+ components) - Maria [13725](https://github.com/SAP/fundamental-ngx/pull/13725)
- [x] Object Number - Maria [13719](https://github.com/SAP/fundamental-ngx/pull/13719)
- [x] Object Identifier - Inna [13731](https://github.com/SAP/fundamental-ngx/pull/13731)
- [x] Text - Deno [13696](https://github.com/SAP/fundamental-ngx/pull/13696)
- [x] Title - Deno [13698](https://github.com/SAP/fundamental-ngx/pull/13698)
- [x] Skeleton - Deno [13699](https://github.com/SAP/fundamental-ngx/pull/13699)
- [x] Busy Indicator - Maria [13727](https://github.com/SAP/fundamental-ngx/pull/13727) [13763](https://github.com/SAP/fundamental-ngx/pull/13763)
- [ ] Progress Indicator - Mike
- [ ] Generic Tag - Mike
- [x] Formatted Text - Maria
- [x] Layout Grid - Maria [13735](https://github.com/SAP/fundamental-ngx/pull/13735)
- [x] Layout Panel - Inna [13810](https://github.com/SAP/fundamental-ngx/pull/13810)
- [x] Scrollbar (directive) - Maria [13766](https://github.com/SAP/fundamental-ngx/pull/13766)
- [x] Infinite Scroll (directive) - Maria [13767](https://github.com/SAP/fundamental-ngx/pull/13767)
- [x] Scroll Spy (directive) - Maria [13770](https://github.com/SAP/fundamental-ngx/pull/13770)

PR #13766, opened 2026-01-21
fix(core): make ScrollbarDirective use signals

MariaIDineva:
## Related Issue(s)

## Description

- Makes `ScrollbarDirective` use signals.
- Replaces `ScrollbarModule` with `ScrollbarDirective` all over the codebase.


libs/core/scrollbar/scrollbar.directive.ts
```diff
@@ -37,49 +36,47 @@ let styleSheet: HTMLStyleElement | null = null;
...
     selector: '[fdScrollbar], [fd-scrollbar]',
     host: {
-        class: 'fd-scrollbar'
+        class: 'fd-scrollbar',
+        '[style.overflow-x]': '_overflowX()',
+        '[style.overflow-y]': '_overflowY()',
+        '(scroll)': 'onScroll($event)'
     },
     hostDirectives: [CdkScrollable],
     standalone: true
```
InnaAtanasova: You don't need this anymore, it's by default

libs/core/scrollbar/scrollbar.directive.ts
```diff
@@ -37,49 +36,47 @@ let styleSheet: HTMLStyleElement | null = null;
...
-    alwaysVisible = false;
+    readonly alwaysVisible = input(false, { transform: booleanAttribute });
 
     /** @hidden */
-    @HostBinding('style.overflow-x')
-    get _overflowX(): ScrollbarOverflowOptions {
-        if (this.noHorizontalScroll) {
-            return 'hidden';
-        }
+    _inPopover = false;
```
InnaAtanasova, line 57: Again, it's not in the scope of this PR, but I think we should re-evaluate this. 
First, it's prefixed with underscore, which suggest it's a private member, which is not. It's a public one and can be set from outside. This, on the other hand, poses other problems:
- Breaks encapsulation - External components are reaching in and setting internal state
- Implicit coupling - The popover components depend on this specific internal property name
- Not discoverable - There's no proper API (input signal) that documents this feature

This, in my opinion, should be an input signal instead:
```
/** Whether the scrollbar is inside a popover and should stop scroll event propagation. */
readonly inPopover = input(false, { transform: booleanAttribute });

protected onScroll(event: Event): void {
    if (this.inPopover()) {
        event.stopImmediatePropagation();
    }
}
```

Then popover components would use proper Angular bindings:
```
<div fd-scrollbar [inPopover]="true">
```
However, making this change would require updating the popover components that currently set this property directly. 
I will probably open an issue to discuss this. It's not in the scope of this PR.

InnaAtanasova: for now can you add (even if't a hidden prop)? Thanks!
/** @hidden Whether the scrollbar is inside a popover and should stop scroll event propagation. */

libs/core/scrollbar/scrollbar.directive.ts
```diff
@@ -37,49 +36,47 @@ let styleSheet: HTMLStyleElement | null = null;
...
-    _inPopover = false;
+    protected readonly _overflowY = computed<ScrollbarOverflowOptions>(() => {
+        if (this.noVerticalScroll()) {
+            return 'hidden';
+        }
 
-    /** @hidden */
-    elementRef: ElementRef<HTMLElement> = inject(ElementRef);
+        return this._overflow();
+    });
```
InnaAtanasova: Again, it's just a personal preference, and I don't want to enforce anything :D 
I think this can be optimized to
```
/** @hidden */
    protected readonly _overflowX = computed<ScrollbarOverflowOptions>(() =>
        this.noHorizontalScroll() ? 'hidden' : this.alwaysVisible() ? 'scroll' : 'auto'
    );

    /** @hidden */
    protected readonly _overflowY = computed<ScrollbarOverflowOptions>(() =>
        this.noVerticalScroll() ? 'hidden' : this.alwaysVisible() ? 'scroll' : 'auto'
    );
```

This would allow you to delete the `_overflow` signal. 
Now you have 2 computed signals, not 3 and this way you eliminate one computed signal declaration and one extra function call per change detection.

libs/core/scrollbar/scrollbar.directive.ts
InnaAtanasova: I created an issue to further discuss the `inPopover`: https://github.com/SAP/fundamental-ngx/issues/13771

PR #13770, opened 2026-01-22
fix(core): migrate ScrollSpyDirective to signals

MariaIDineva:
## Related Issue(s)

## Description

 - Migrates `ScrollSpyDirective` to signals.
 - Replaces `ScrollSpyModule` with `ScrollSpyDirective`.


libs/core/scroll-spy/scroll-spy.directive.ts
```diff
@@ -84,18 +69,21 @@ export class ScrollSpyDirective implements OnInit {
         const [firstChild] = children;
         const childrenLength = children.length;
         const targetScrollTop = target.scrollTop;
-        const targetOffsetTop = target.offsetTop + this.targetOffset - (target.offsetTop - firstChild.offsetTop);
+        const targetOffsetTop = target.offsetTop + this.targetOffset() - (target.offsetTop - firstChild.offsetTop);
```
InnaAtanasova: I know it's not in the scope of this PR, but doesn't `target.offsetTop` cancel itself in this equation? 
If we remove the brackets we get:
```
const targetOffsetTop = target.offsetTop + this.targetOffset() - target.offsetTop + firstChild.offsetTop;
```
and the whole thing is just:
```
const targetOffsetTop = firstChild.offsetTop + this.targetOffset();
```

libs/core/wizard/wizard.component.ts
```diff
@@ -208,12 +208,13 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
         if (!_fromScrollToCurrentStep) {
             this.steps.forEach((step) => {
-                if (step._stepId.toString() === $event.children[0].children[0].id) {
+                const eventStepId = $event?.children[0].children[0].id;
+                if (step._stepId.toString() === eventStepId) {
                     step.status = CURRENT_STEP_STATUS;
-                } else if (step._stepId < parseInt($event.children[0].children[0].id, 10)) {
+                } else if (eventStepId && step._stepId < parseInt(eventStepId, 10)) {
                     step.status = COMPLETED_STEP_STATUS;
                 } else {
```
InnaAtanasova: Nothing wrong here with your implementation, looks great!
Here's another way to approach this:
```
scrollSpyChange($event: HTMLElement | undefined): void {
    if (!_fromScrollToCurrentStep || !$event) {
        return;
    }
    
    const eventStepId = $event.children[0].children[0].id;
    this.steps.forEach((step) => {
        if (step._stepId.toString() === eventStepId) {
            step.status = CURRENT_STEP_STATUS;
        } else if (step._stepId < parseInt(eventStepId, 10)) {
            step.status = COMPLETED_STEP_STATUS;
        } else {
            step.status = UPCOMING_STEP_STATUS;
        }
    });
    this._shrinkWhileAnyStepIsTooNarrow();
}
```

You check for `!_fromScrollToCurrentStep || !$event` early and return immediately => avoid repeated optional chaining. 
But like I said, it's just another way to approach it.

InnaAtanasova (approved):

libs/core/wizard/wizard.component.ts
MariaIDineva: OK, I will update, but shouldn't that be:
```
        if (_fromScrollToCurrentStep || !$event) {
            return;
        }
```
as all the logic is currently inside `if (!_fromScrollToCurrentStep)`?

libs/core/scroll-spy/scroll-spy.directive.ts
MariaIDineva: Yes!

InnaAtanasova:
🚀

Issue #13769, opened 2026-01-22
multi-combobox inaccurate positioning of dropdown


SAPAndiBraimllari:
### Describe your proposal

The positioning of the multi-combobox is quite far away from the input area itself. I assume when selecting all, the position of the dropdown is not calculated again, and it is stuck on its old place. It needs to be updated, otherwise it looks off.

![Image](https://github.com/user-attachments/assets/8b320548-2f66-4ae4-b0ed-028baefa6309)

I have a working small stackblitz example.
https://stackblitz.com/edit/xqqanzjt-zgs1j5vm?file=src%2Fapp%2Fapp.component.ts
I've also added a recording of it here.
Please reopen this; I can also open a new issue if that's better for you guys.
Happy to add more information to this.

https://github.com/user-attachments/assets/1ded2667-b61a-4827-8727-5df7dde23f9c

This is a follow-up on https://github.com/SAP/fundamental-ngx/issues/12978#issuecomment-3769847647

### Which versions of Angular and Fundamental Library for Angular are affected?

Angular: 20, Fundamental: 0.57.5

### Steps to Reproduce (for bugs)

Steps to reproduce:
On a dialog with a dropdown with many elements:
- Click on search and open the country selection
- Click "select all"

### StackBlitz example(s)

https://stackblitz.com/edit/xqqanzjt-zgs1j5vm?file=src%2Fapp%2Fapp.component.ts

### Priority

None

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

PR #13774, opened 2026-01-23
fix(core): multi-combobox inaccurate positioning of dropdown

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13769

## Description

-  Reposition multi-combobox dropdown when selection changes. Now, when multi-combobox is inside a dialog which grows in height, the multi-combobox dropdown correctly follows the input and does not stay in its initial position.
- Amends the "Multi Combobox Inside Dialog" example to replicate the issue and test that it is resolved after the code change.
- Amend `track` expressions in the template to track by a unique identifier and get rid of Angular warning:

Screenshot 2026-01-23 at 15 35 30

## Screenshots

Before:
Screenshot 2026-01-23 at 15 40 15

After:
Screenshot 2026-01-23 at 15 28 20


libs/core/multi-combobox/base-multi-combobox.class.ts
```diff
@@ -533,6 +558,7 @@ export abstract class BaseMultiCombobox<T = any> {
             }
             return optionItem;
         });
+        console.log('this._suggestions', this._suggestions);
```
droshev: do we need this comment in the log?

MariaIDineva: Of course not, thanks!

juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **1 point that might be worth attention** (could be false positives, please use your judgment):

1. <strong>Refining the template tracking logic for better stability</strong>
   - [Location](https://github.com/SAP/fundamental-ngx/blob/8d8c5902e1fb9529e7586b5f435afa1810afbf7b/libs/core/multi-combobox/multi-combobox.component.html#L54): `libs/core/multi-combobox/multi-combobox.component.html:L54, L194, L221`
   - Impact: If the `id` property is optional or missing, Angular might encounter duplicate keys, which could lead to runtime errors during rendering.
   - Suggestion: Consider tracking by the item identity itself (e.g., `track token`) or using a property guaranteed to be unique and present, such as `value`.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

PR #13783, opened 2026-01-26
chore: update nx 22.4.1

droshev:
update nx 22.4.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
         "@angular/platform-browser-dynamic": "21.1.0",
         "@angular/router": "21.1.0",
         "@fundamental-styles/cx": "0.40.1",
-        "@nx/angular": "22.3.3",
+        "@nx/angular": "22.4.1",
         "@sap-theming/theming-base-content": "11.32.3",
         "@stackblitz/sdk": "1.9.0",
         "@swc/helpers": "0.5.11",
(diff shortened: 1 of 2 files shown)


libs/i18n/src/lib/translations/translations.properties
```diff
@@ -903,7 +903,7 @@ platformTable.deselectSingleRow = To deselect row, press SPACEBAR
 #XACT Text, which is read by screen reader, tells instruction of how to select a row
 platformTable.selectSingleRow = To select row, press SPACEBAR
 #XBUT Label for the button, which loads additional rows
-platformTable.loadMore = Show more
```
InnaAtanasova, line 906: Should this be part of this PR?

libs/i18n/src/lib/translations/translations.ts
```diff
@@ -548,7 +548,7 @@ export default {
         deselectAllCheckboxLongLabel: 'Checkbox, checked, click to deselect all',
         deselectSingleRow: 'To deselect row, press SPACEBAR',
         selectSingleRow: 'To select row, press SPACEBAR',
-        loadMore: 'Show more',
```
InnaAtanasova, line 551: Should this be part of this PR?

droshev: you are right. let me update iot

droshev:
addressed in #13784

PR #13801, opened 2026-01-29
docs(core): add 2 views in a dialog example

droshev:
add 2 views in dialog

### DIFF
--- a/libs/docs/core/dialog/examples/dialog-location-selection/dialog-location-selection-example.component.html
+++ b/libs/docs/core/dialog/examples/dialog-location-selection/dialog-location-selection-example.component.html
@@ -1,0 +1,21 @@
+<div>
+    <button fd-button label="Open Location Dialog" (click)="openLocationDialog()"></button>
+
+    @if (selectedLocation) {
+        <div style="margin-top: 1rem; padding: 1rem; background-color: #f4f4f4; border-radius: 4px">
+            <h3 style="margin-top: 0">Selected Location:</h3>
+            <p>
+                <strong>{{ selectedLocation.name }}, {{ selectedLocation.city }}</strong>
+            </p>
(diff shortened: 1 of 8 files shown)


libs/docs/core/dialog/examples/dialog-location-selection/location-selection-dialog.component.html
```diff
@@ -0,0 +1,121 @@
...
+            ></fd-input-group>
+        </div>
+
+        @if (!showDetails()) {
+            <!-- List View -->
+            <ul fd-list [noBorder]="true" class="location-list-container" cdkScrollable>
+                @for (location of filteredLocations(); track location.id) {
+                    <li
+                        fd-list-item
+                        style="cursor: pointer"
```
mikerodonnell89, line 51: I would just use `[interactive]="true"` so we get the hover states etc

libs/docs/core/dialog/examples/dialog-location-selection/dialog-location-selection-example.component.html
```diff
@@ -0,0 +1,21 @@
+<div>
+    <button fd-button label="Open Location Dialog" (click)="openLocationDialog()"></button>
+
+    @if (selectedLocation) {
```
MariaIDineva, line 4: Not sure what the exact intention is here, but I guess it shoud be that the selected location is displayed when the user selects a new location and clicks OK. But there is a weird bug that the new location only gets displayed when the popover is open again. Please check the video:

https://github.com/user-attachments/assets/f4c5c105-8d08-4875-8d86-05bf981077fa

The problem was fixed when I turned `selectedLocation` to a signal.

droshev:
I will close it for now and we can resurrect it once it is needed.

Issue #13775, opened 2026-01-23
User Menu keyboard navigation bug


InnaAtanasova:
It is no longer possible to navigate between main and submenu items with right and left arrow keys, as well as to open/close the submenu with them. This was working before and the component has not been updated since, so probably we modified the keyboard service and introduced the bug there.
<img width="599" height="615" alt="Image" src="https://github.com/user-attachments/assets/660c3748-0dca-48bb-8fee-905e2f669262" />

PR #13804, opened 2026-01-29
fix(core): add left/right arrow navigation and fix focus containment in user menu submenus

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13775

## Description
**Added keyboard navigation:**
- Right arrow key now opens submenu when focus is on a parent menu item
- Left arrow key closes submenu when focus is inside the submenu popover

**Fixed focus escape issue:**
- Submenu focus now stays contained within the submenu when using up/down arrow keys
- Provided isolated KeyboardSupportService instance to UserMenuSublistComponent via providers array
- This prevents parent list's keyboard manager from interfering with submenu navigation

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.59.1-rc.25",
+    "version": "0.59.1-rc.26",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.59.1-rc.25",
+    "version": "0.59.1-rc.26",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/cdk/package.json
+++ b/libs/cdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cdk",
-    "version": "0.59.1-rc.25",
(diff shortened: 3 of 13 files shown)


libs/core/user-menu/components/user-menu-list-item.component.ts
```diff
@@ -108,6 +109,38 @@ export class UserMenuListItemComponent implements KeyboardSupportItemInterface {
...
+        if (!this.hasSubmenu() || this.mobile()) {
+            return;
+        }
+
+        const popoverInstance = this.popover();
+        if (!popoverInstance) {
+            return;
+        }
+
+        if (KeyUtil.isKeyCode(event, RIGHT_ARROW) && !this.isOpen()) {
```
MariaIDineva: How about in RTL, shouldn't the logic be reversed - open with left arrow and close with right? In the method below as well.

libs/core/user-menu/components/user-menu-sublist.component.ts
```diff
@@ -69,7 +70,7 @@ export class UserMenuSublistComponent implements AfterViewInit {
 
     /** @hidden */
     ngAfterViewInit(): void {
-        this._keyboardSupportService.setKeyboardService(this._listItems, false, false);
+        this._keyboardSupportService.setKeyboardService(this._listItems, true, false);
```
MariaIDineva: Is that change on purpose, do we want to have a wrapping behaviour of the subitems?

libs/core/user-menu/components/user-menu-list-item.component.ts
InnaAtanasova: yes, needs the RTL. I added it! Thanks for catching it.

libs/core/user-menu/components/user-menu-sublist.component.ts
InnaAtanasova: Shouldn't be changed. I reverted the change. Probably leftover from prev attempts to fix the issue. Thanks

Issue #13741, opened 2026-01-19
Angular Signals Migration Phase 2


InnaAtanasova:
ICON-DEPENDENT COMPONENTS 

- [x] Object Status - Maria [13792](https://github.com/SAP/fundamental-ngx/pull/13792)
- [x] Object Marker - Maria [13808](https://github.com/SAP/fundamental-ngx/pull/13808)
- [x] Status Indicator - Maria [13809](https://github.com/SAP/fundamental-ngx/pull/13809)
- [x] Info Label - Inna [13811](https://github.com/SAP/fundamental-ngx/pull/13811)
- [x] Avatar - Inna [13812](https://github.com/SAP/fundamental-ngx/pull/13812)
- [x] Rating Indicator - Inna [13831](https://github.com/SAP/fundamental-ngx/pull/13831)
- [x] Link - Maria [13817](https://github.com/SAP/fundamental-ngx/pull/13817)
- [x] Tile - Inna [13822](https://github.com/SAP/fundamental-ngx/pull/13822)
- [x] Illustrated Message - Inna [13823](https://github.com/SAP/fundamental-ngx/pull/13823)

PR #13809, opened 2026-01-30
fix(core): migrate status-indicator to signals

MariaIDineva:
## Related Issue(s)

## Description

Migrates `StatusIndicator` to signals.

### DIFF
--- a/.github/TESTING.md
+++ b/.github/TESTING.md
@@ -85,7 +85,7 @@ act workflow_dispatch -W .github/workflows/create-release.yml -n
 ✅ `get-release-tag` - Release tag determination logic  
 ✅ `current-version` - Current version detection  
 ✅ Jest unit tests for `get-release-tag` (5 scenarios)  
-✅ Jest unit tests for `get-version` (20 scenarios: git tags, prerelease/RC handling, branch-specific versions, hotfix branches, and fallback logic)
+✅ Jest unit tests for `get-version` (23 scenarios: git tags, prerelease/RC handling, package.json fallback, failed tag recovery, branch-specific versions, hotfix branches)
 ✅ Interactive playground for release scenarios (`release-scenario.playground.js`)
 
 ### Release Scenario Playground
--- a/.github/actions/helpers/get-version.js
+++ b/.github/actions/helpers/get-version.js
@@ -44,7 +44,20 @@ module.exports = (branch = null) => {
             // semver.rsort handles prerelease comparison correctly:
             // e.g., 0.59.0-rc.0 > 0.58.1 (because 0.59.0 > 0.58.1)
             const sortedVersions = semver.rsort(validVersions);
-            return sortedVersions[0];
+            const tagVersion = sortedVersions[0];
+
+            // Also check package.json version - use whichever is higher
+            // This handles cases where a release commit bumped package.json but tag creation failed
+            try {
+                const packageVersion = getFileContents('libs/core/package.json', null).version;
+                if (semver.valid(packageVersion) && semver.gt(packageVersion, tagVersion)) {
+                    return packageVersion;
+                }
+            } catch (e) {
+                // Ignore errors reading package.json
+            }
+
+            return tagVersion;
         }
     } catch (e) {
         // Git command failed or no tags found, fall through to package.json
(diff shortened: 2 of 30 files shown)


libs/core/status-indicator/status-indicator.component.ts
```diff
@@ -31,236 +33,203 @@ export class Point {
...
 
     /** Aria defines role description for the Status Indicator. */
-    @Input()
-    ariaRoleDescription: Nullable<string>;
+    readonly ariaRoleDescription = input<Nullable<string>>();
 
     /** Aria Focusable for the Status Indicator. */
-    @Input()
-    focusable: boolean;
+    readonly focusable = input(false);
```
InnaAtanasova: Can you make this readonly focusable = input(false, { transform: booleanAttribute });

libs/core/status-indicator/status-indicator.component.html
```diff
@@ -1,100 +1,100 @@
...
-        [class.fd-status-indicator__label--lg]="labelSize === 'lg'"
-        [class.fd-status-indicator__label--xl]="labelSize === 'xl'"
-        [class.fd-status-indicator__label--top]="labelPosition === 'top'"
-        [class.fd-status-indicator__label--left]="labelPosition === 'left'"
+        [class.fd-status-indicator__label--sm]="labelSize() === 'sm'"
+        [class.fd-status-indicator__label--md]="labelSize() === 'md'"
+        [class.fd-status-indicator__label--lg]="labelSize() === 'lg'"
+        [class.fd-status-indicator__label--xl]="labelSize() === 'xl'"
+        [class.fd-status-indicator__label--top]="labelPosition() === 'top'"
+        [class.fd-status-indicator__label--left]="labelPosition() === 'left'"
```
InnaAtanasova: Check the comment for lines 26-32

libs/core/status-indicator/status-indicator.component.html
```diff
@@ -1,100 +1,100 @@
...
-        [class.fd-status-indicator__label--lg]="labelSize === 'lg'"
-        [class.fd-status-indicator__label--xl]="labelSize === 'xl'"
-        [class.fd-status-indicator__label--right]="labelPosition === 'right'"
-        [class.fd-status-indicator__label--bottom]="labelPosition === 'bottom'"
+        [class.fd-status-indicator__label--sm]="labelSize() === 'sm'"
+        [class.fd-status-indicator__label--md]="labelSize() === 'md'"
+        [class.fd-status-indicator__label--lg]="labelSize() === 'lg'"
+        [class.fd-status-indicator__label--xl]="labelSize() === 'xl'"
+        [class.fd-status-indicator__label--right]="labelPosition() === 'right'"
+        [class.fd-status-indicator__label--bottom]="labelPosition() === 'bottom'"
```
InnaAtanasova: Maybe we can get rid of all these bindings in the template and move them to the ts file.
We can have in the template `<span [class]="_labelCssClass()">{{ statusLabel() }}</span>` and in the component logic something like:
```
 protected readonly _labelCssClass = computed(() =>
        [
            'fd-status-indicator__label',
            this.labelSize() ? `fd-status-indicator__label--${this.labelSize()}` : '',
            this.labelPosition() ? `fd-status-indicator__label--${this.labelPosition()}` : ''
        ]
            .filter(Boolean)
            .join(' ')
    );
```
This way we can achieve:
- Label class logic defined once
- Computed signal caches result until dependencies change
- Changes only needed in one place
- Cleaner template

libs/core/status-indicator/status-indicator.component.ts
```diff
@@ -31,236 +34,213 @@ export class Point {
...
+            this.clickable() ? 'fd-status-indicator--link' : '',
+            this.labelPosition() === 'right' || this.labelPosition() === 'left'
+                ? 'fd-status-indicator--horizontal-label'
+                : ''
+        ]
+            .filter(Boolean)
+            .join(' ')
+    );
+
+    protected readonly _labelCssClass = computed(() =>
```
droshev: no need of `_` as it is protected

InnaAtanasova (approved):

PR #13810, opened 2026-01-30
fix(core): migrate layout-panel to signal-based patterns

InnaAtanasova:
## Related Issue(s)

## Description
Migrates the layout-panel component and its tests to use Angular 20+ signal-based patterns.

In **layout-panel.component.ts** component the following improvements are introduced:
  - Replaced `ngOnChanges`/`ngOnInit` lifecycle hooks with `computed()` signal
  - Removed manual DOM manipulation (`_applyBackgroundImage()`)
  - Added declarative `[style.background-image]` host binding
  - Removed unused `ElementRef` and related imports
  - Eliminated need for `OnChanges` and `OnInit` interfaces

### DIFF
--- a/.github/TESTING.md
+++ b/.github/TESTING.md
@@ -85,7 +85,7 @@ act workflow_dispatch -W .github/workflows/create-release.yml -n
 ✅ `get-release-tag` - Release tag determination logic  
 ✅ `current-version` - Current version detection  
 ✅ Jest unit tests for `get-release-tag` (5 scenarios)  
-✅ Jest unit tests for `get-version` (20 scenarios: git tags, prerelease/RC handling, branch-specific versions, hotfix branches, and fallback logic)
+✅ Jest unit tests for `get-version` (23 scenarios: git tags, prerelease/RC handling, package.json fallback, failed tag recovery, branch-specific versions, hotfix branches)
 ✅ Interactive playground for release scenarios (`release-scenario.playground.js`)
 
 ### Release Scenario Playground
--- a/.github/actions/helpers/get-version.js
+++ b/.github/actions/helpers/get-version.js
@@ -44,7 +44,20 @@ module.exports = (branch = null) => {
             // semver.rsort handles prerelease comparison correctly:
             // e.g., 0.59.0-rc.0 > 0.58.1 (because 0.59.0 > 0.58.1)
             const sortedVersions = semver.rsort(validVersions);
-            return sortedVersions[0];
+            const tagVersion = sortedVersions[0];
+
+            // Also check package.json version - use whichever is higher
+            // This handles cases where a release commit bumped package.json but tag creation failed
+            try {
+                const packageVersion = getFileContents('libs/core/package.json', null).version;
+                if (semver.valid(packageVersion) && semver.gt(packageVersion, tagVersion)) {
+                    return packageVersion;
+                }
+            } catch (e) {
+                // Ignore errors reading package.json
+            }
+
+            return tagVersion;
         }
     } catch (e) {
         // Git command failed or no tags found, fall through to package.json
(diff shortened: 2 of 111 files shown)


libs/core/layout-panel/layout-panel-footer/layout-panel-footer.component.ts
```diff
@@ -16,39 +7,19 @@ export type FooterPosition = 'start' | 'end';
...
+        '[class.fd-layout-panel__footer--start]': 'position() === "start"',
+        '[class.fd-layout-panel__footer--end]': 'position() === "end"'
+    }
 })
-export class LayoutPanelFooterComponent implements OnInit, OnChanges, CssClassBuilder {
-    /** Apply user custom styles */
-    @Input()
-    class: string;
-
+export class LayoutPanelFooterComponent {
```
MariaIDineva: You could mark this class as implementing `HasElementRef` which `CssClassBuilder`(now removed) implements.

libs/core/layout-panel/layout-panel-body/layout-panel-body.component.ts
```diff
@@ -1,33 +1,18 @@
...
-
-    /** Whether the edges of the panel should have bleeding padding. */
-    @Input()
-    @HostBinding('class.fd-layout-panel__body--bleed')
-    bleed = false;
+    /**
+     * Whether the edges of the panel should have bleeding padding.
+     * @default false
+     */
+    readonly bleed = input(false, { transform: booleanAttribute });
```
droshev, line 17: can we make it `protected`?

libs/core/layout-panel/layout-panel-footer/layout-panel-footer.component.ts
```diff
@@ -16,39 +7,19 @@ export type FooterPosition = 'start' | 'end';
...
-    /** @hidden */
-    ngOnInit(): void {
-        this.buildComponentCssClass();
-    }
-
-    /** @hidden */
-    ngOnChanges(): void {
-        this.buildComponentCssClass();
-    }
+    readonly elementRef = inject(ElementRef);
```
droshev, line 25: `protected` as well? less pollution

libs/core/layout-panel/layout-panel.component.ts
```diff
@@ -1,65 +1,40 @@
...
-        this._applyBackgroundImage();
-    }
-
-    /** @hidden */
-    private _applyBackgroundImage(): void {
-        if (this.backgroundImage) {
-            (this.elRef.nativeElement as HTMLElement).style['background-image'] = 'url("' + this.backgroundImage + '")';
-        }
-    }
+    protected readonly _backgroundImageUrl = computed(() => {
```
droshev: since it is not private we don't need `_`

libs/core/layout-panel/layout-panel-footer/layout-panel-footer.component.ts
InnaAtanasova: no, because it was public before and it will be a Breaking change

libs/core/layout-panel/layout-panel-body/layout-panel-body.component.ts
InnaAtanasova: protected input property?

droshev: https://angular.dev/style-guide#use-protected-on-class-members-that-are-only-used-by-a-components-template

InnaAtanasova: But this is an input. If it's not public the user won't be able to set it. What you linked to is a computed signal, which is another story and should be protected, yes. But inputs and outputs are public.

PR #13815, opened 2026-02-02
fix: migrate RtlService usage to Angular 21 signal patterns

droshev:
migrate RtlService usage to Angular 21 signal patterns

BREAKING CHANGE:
 in SelectComponent public API: Renamed `rtl$` to `isRtl` in SelectComponent and SelectComponentInterface

### DIFF
--- a/libs/cdk/utils/directives/resize/resize.directive.ts
+++ b/libs/cdk/utils/directives/resize/resize.directive.ts
@@ -7,10 +7,9 @@ import {
     Input,
     OnChanges,
     OnDestroy,
-    Optional,
     Output,
     SimpleChanges,
-    computed
+    inject
 } from '@angular/core';
 import { Observable, Subscription, fromEvent, merge } from 'rxjs';
 import { filter, map, pairwise, takeUntil, tap } from 'rxjs/operators';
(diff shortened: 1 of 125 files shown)


libs/cdk/utils/services/rtl.service.ts
```diff
@@ -1,32 +1,79 @@
...
+     */
+    readonly rtl: WritableSignal<boolean>;
 
-    /** Signal wrapper for RTL value. */
-    rtlSignal: Signal<boolean>;
+    /**
+     * Signal for RTL value.
+     * @deprecated Use `rtl()` instead. This alias will be removed in a future version.
+     */
+    get rtlSignal(): WritableSignal<boolean> {
```
droshev, line 66: I can delete it now (not used anymore)

libs/cdk/utils/directives/resize/resize.directive.ts
```diff
@@ -83,13 +82,10 @@ export class ResizeDirective implements OnChanges, AfterViewInit, OnDestroy {
...
     /** @hidden */
-    private readonly _direction$ = computed(() => (this._rtlService?.rtlSignal() ? -1 : 1));
+    private readonly _rtlService = inject(RtlService, { optional: true });
 
     /** @hidden */
-    constructor(
-        private _elementRef: ElementRef<HTMLElement>,
-        @Optional() private _rtlService: RtlService
-    ) {}
+    constructor(private _elementRef: ElementRef<HTMLElement>) {}
```
MariaIDineva, line 88: Maybe **_elementRef** can be moved outside the constructor as well, like it is moved in the other components?

PR #13817, opened 2026-02-02
fix(core): migrate link to signals, add a11y updates

MariaIDineva:
## Related Issue(s)

## Description

- Migrates `Link` component to signals:
  - Logic from `ngAfterViewInit` is now moved into an `afterNextRender`callback in an effect in the constructor. Logic is split into smaller private methods - **_clearPortals**, **_updatePortals** and **_detachPortals**.
  - Logic for [filtering only first descendants](https://github.com/SAP/fundamental-ngx/pull/13817/changes#diff-40ce869a3bdc53e2e9825e1ef6f18cfe119c3546af511438ce29ebbb68d4ba1bL142) is now replaced by `contentChildren` flag `{ descendants: false }`.
  - `buildComponentCssClass` method is removed, classes are now directly added to the host.
- Updates unit tests.
- Updates Link examples in documentation.
- Replaces all ocurrences of deprecated `LinkModule` with `LinkComponent`.
- Adopts a11y improvements coming from [fundamental-styles](https://github.com/SAP/fundamental-styles/pull/6063)
  - Additional SR-only spans for `emphasized` and `subtle` links with their corresponding translations.
  - Prefix and postfix icons now have `role="presentation" aria-hidden="true"`
  - Exposes a `touchTarget` input (false by default) to ensure larger touchable area.

### DIFF
--- a/libs/core/link/link.component.ts
+++ b/libs/core/link/link.component.ts
@@ -4,7 +4,6 @@ import {
     booleanAttribute,
     ChangeDetectionStrategy,
     Component,
-    computed,
     contentChildren,
     effect,
     ElementRef,
@@ -35,7 +34,13 @@ const LINK_ROUTER_TARGET = new InjectionToken<RouterLink | undefined>('linkRoute
     changeDetection: ChangeDetectionStrategy.OnPush,
     host: {
         '[attr.aria-disabled]': 'disabled() ? true : false',
-        '[class]': '_cssClass()'
+        class: 'fd-link',
+        '[class.fd-link--emphasized]': 'emphasized()',
+        '[class.is-disabled]': 'disabled()',
+        '[class.fd-link--inverted]': 'inverted()',
+        '[class.fd-link--subtle]': 'subtle()',
+        '[class.fd-link--undecorated]': 'undecorated()',
+        '[class.fd-link--touch-target]': 'touchTarget()'
     },
     providers: [
         {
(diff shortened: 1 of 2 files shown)


libs/core/link/link.component.ts
```diff
@@ -47,144 +52,136 @@ import { FD_LINK_COMPONENT } from './tokens';
...
+    /** @hidden */
+    protected readonly _postfixPortal = signal<Portal<any> | null>(null);
+
     /** @hidden */
-    _postfixIconName: string;
+    protected readonly _prefixIconName = signal<string>('');
 
     /** @hidden */
-    private readonly _destroyed$ = inject(DestroyRef);
+    protected readonly _postfixIconName = signal<string>('');
```
InnaAtanasova: Protected members should not be prefixed, only private. The instructions have been updated but for some reason they are ignored...

libs/core/link/link.component.ts
```diff
@@ -47,144 +52,136 @@ import { FD_LINK_COMPONENT } from './tokens';
...
-            this.subtle ? 'fd-link--subtle' : '',
-            this.undecorated ? 'fd-link--undecorated' : '',
-            this.class
-        ];
+            this.emphasized() ? 'fd-link--emphasized' : '',
+            this.disabled() ? 'is-disabled' : '',
+            this.inverted() ? 'fd-link--inverted' : '',
+            this.subtle() ? 'fd-link--subtle' : '',
+            this.undecorated() ? 'fd-link--undecorated' : '',
+            this.touchTarget() ? 'fd-link--touch-target' : ''
```
InnaAtanasova: Since there's no "calculation" like `fd-tile--${sizeValue}`, these class bindings can happen in the decorator's host object and avoid calling computed. But this is ok, too.

libs/core/link/link.component.ts
MariaIDineva: Updated, thanks!

libs/core/link/link.component.ts
MariaIDineva: Updated too, thanks!

PR #13831, opened 2026-02-05
fix(core): update Rating Indicator to signals

InnaAtanasova:
## Related Issue(s)

## Description
- migrates to signals
- add i18n
- code refactor
- update documentation
- replace module imports with component imports

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.60.0-rc.7",
+    "version": "0.60.0-rc.8",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.60.0-rc.7",
+    "version": "0.60.0-rc.8",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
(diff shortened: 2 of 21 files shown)


libs/core/rating-indicator/pipes/rating-star-label.pipe.ts
```diff
@@ -1,15 +1,32 @@
-import { Pipe, PipeTransform } from '@angular/core';
+import { inject, Pipe, PipeTransform, Signal } from '@angular/core';
+import { toSignal } from '@angular/core/rxjs-interop';
+import { FD_LANGUAGE, FdLanguage, TranslationResolver } from '@fundamental-ngx/i18n';
+import { Observable } from 'rxjs';
 
 @Pipe({
     name: 'ratingStarLabel',
-    standalone: true
+    pure: false // required to update the value when the observable is resolved
```
MariaIDineva: Alternatively, if you want to keep the pipe pure, you could pass the language as a param on line 23:
```
    transform(index: number, controlsCount: number, useHalves: boolean, currentLanguage: FdLanguage): string {
        const ofTranslation = this._translationResolver.resolve(currentLanguage, 'coreRatingIndicator.ariaLabelValue');
...
```

and inject the language in the rating indicator component:
```
    protected readonly language = toSignal(inject<Observable<FdLanguage>>(FD_LANGUAGE), {
        initialValue: {} as FdLanguage
    });
```

and then in the template you could pass the language as a last argument:
```
[attr.aria-label]="i | ratingStarLabel: rates().length : allowHalves() : language()"
```

InnaAtanasova:
@MariaIDineva thanks for the review. I updated the following:
- replaced the module import with component in other components using the Rating indicator
- added unit test for the pipe
- updated the translation as you suggested (thanks)
Let me know if I missed anything!

Issue #13833, opened 2026-02-06
Shellbar actions - support toggle buttons


droshev:
### Describe your proposal

In the shellbar we have action (icon) buttons. A new requirement is to support also toggle icon buttons. Maybe we need to bring (if something is missing) in fundamental styles too.

### Proposed Solution

offer the toggle button support

### Priority

High

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

PR #13836, opened 2026-02-06
fix(core): add option to use Toggle buttons as ShellBar actions

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13833

## Description

Added toggle buttons in Shell Bar

## Screenshots
Screenshot 2026-02-06 at 3 33 45 PM
Screenshot 2026-02-06 at 3 33 34 PM

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.60.0-rc.3",
+    "version": "0.60.0-rc.4",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.60.0-rc.3",
+    "version": "0.60.0-rc.4",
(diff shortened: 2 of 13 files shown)


libs/core/shellbar/shellbar-action/shellbar-action.component.html
```diff
@@ -1,13 +1,14 @@
 <div class="fd-shellbar__action fd-shellbar__action--desktop">
     <button
         fd-button
-        fdType="transparent"
-        class="fd-shellbar__button"
+        [fdType]="fdType() ?? 'transparent'"
```
MariaIDineva, line 4: Nullish coalescing operation is redundant here as you set a default value of the **fdType** in the component.

libs/core/shellbar/shellbar-actions-mobile/shellbar-actions-mobile.component.html
```diff
@@ -4,9 +4,10 @@
             <fd-action-sheet-control>
                 <button
                     fd-button
-                    fdType="transparent"
+                    [fdType]="fdType()"
```
MariaIDineva, line 7: Is the control supposed to support toggle as well, not only the actions?

MariaIDineva: I guess it does, as shown in the second screenshot, please disregard.

Issue #13742, opened 2026-01-19
Angular Signals Migration Phase 3


InnaAtanasova:
BASE CLASSES & SERVICES 

- [x] BaseButton (blocks Button, ButtonBar) - Inna [13847](https://github.com/SAP/fundamental-ngx/pull/13847)
- [ ] BasePopoverClass (blocks Popover, Menu, InlineHelp, ProductSwitch) - Deno
- [ ] BaseMultiCombobox (blocks MultiCombobox) - Maria [13853](https://github.com/SAP/fundamental-ngx/pull/13853) **core only**
- [ ] DialogBase (blocks Dialog, MessageBox) - Inna
- [ ] BaseTreeItem (blocks Tree) - Maria
- [ ] CardFocusItem (blocks Card)
- [ ] ContentDensityObserver  (blocks 30+ components)
- [ ] CvaDirective & CvaControl (form controls)
- [ ] PopoverService (popover ecosystem)
- [ ] MenuService (menu ecosystem)
- [ ] SelectKeyManagerService (select/combobox)
- [ ] KeyboardSupportService (list navigation)

PR #13847, opened 2026-02-09
fix(core, platform, btp): update BaseButton, Core and Platform Button to signals

InnaAtanasova:
## Related Issue(s)

## Description
Migrates BaseButton, Core ButtonComponent, Platform ButtonComponent, and BTP button components to use Angular 21+ signal-based reactive primitives.

Changes:
- Converted all @Input() to input() and @Output() to output()
- Replaced decorators with host: {} property for bindings/listeners
- Implemented computed() for derived state and linkedSignal() for mutable state
- Moved ButtonModel interface from platform to core (breaks circular dependency)
- Replaced ButtonModule/PlatformButtonModule with ButtonComponent across 40+ files
- Removed manual change detection calls (signals auto-notify)
- Fixed test compatibility with signal-based API

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.60.0-rc.4",
+    "version": "0.60.0-rc.6",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.60.0-rc.4",
+    "version": "0.60.0-rc.6",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/cdk/package.json
+++ b/libs/cdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cdk",
-    "version": "0.60.0-rc.4",
+    "version": "0.60.0-rc.6",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - CDK",
     "license": "Apache-2.0",
--- a/libs/core/action-sheet/action-sheet-item/action-sheet-item.component.html
+++ b/libs/core/action-sheet/action-sheet-item/action-sheet-item.component.html
@@ -1,8 +1,12 @@
 <button
     fd-button
-    [class]="negative ? 'fd-button--negative' : 'fd-button--transparent fd-button--text-alignment-left'"
-    role="menuitem"
     class="fd-button--full-width"
+    [class.fd-button--text-alignment-left]="!negative"
+    [class.fd-button--negative]="negative"
+    [class.fd-button--transparent]="!negative && fdType() === 'standard'"
+    role="menuitem"
+    [fdType]="fdType()"
+    [toggled]="toggled()"
     [label]="label"
     [glyph]="glyph"
     [glyphFont]="glyphFont"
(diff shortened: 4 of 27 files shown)


libs/core/button/base-button.ts
```diff
@@ -29,91 +21,174 @@ export type ButtonType =
...
-    @Input({ transform: booleanAttribute })
-    selected: BooleanInput;
+    readonly selected = input<boolean, BooleanInput>(false, { transform: booleanAttribute });
 
     /**
      * Native type of button element
      */
-    @Input()
-    type: Nullable<string> = 'button';
+    readonly type = input<Nullable<string>>('button');
```
MariaIDineva: Do we keep `Nullable`, here and below?

libs/core/button/base-button.ts
```diff
@@ -29,91 +21,174 @@ export type ButtonType =
...
+     */
+    isSelected(): boolean {
+        return this.selectedState();
+    }
+
+    /**
+     * Programmatically set the toggled state.
+     * This allows parent components to update the button's toggled state.
+     */
+    setToggled(value: boolean): void {
```
MariaIDineva, line 183: This method is unused in our example. Is the idea behind it to be an alterntive to potential apps that maybe view child(ren) and programmatically change the `toggled` field which is no longer possible now that it is an input signal?

libs/core/button/button.component.ts
```diff
@@ -94,72 +90,68 @@ export class ButtonComponent
...
+
+        if (this._disabledState() || this.ariaDisabled()) {
+            classes.push('is-disabled');
+        }
+
+        if (this.toggledState()) {
+            classes.push('fd-button--toggled');
+        }
+
+        return classes.join(' ');
```
MariaIDineva: This could as well be:
```
protected readonly cssClass = computed(() =>
    [
        'fd-button',
        this.fdTypeState() ? `fd-button--${this.fdTypeState()}` : '',
        this.fdMenu() ? 'fd-button--menu' : '',
        this._disabledState() || this.ariaDisabled() ? 'is-disabled' : '',
        this.toggledState() ? 'fd-button--toggled' : ''
    ]
        .filter(Boolean)
        .join(' ')
);
```

libs/core/button/button.component.ts
```diff
@@ -94,72 +90,68 @@ export class ButtonComponent
...
+
+    /** @hidden */
+    private _subscriptions = new Subscription();
 
     /** @hidden */
-    ngOnInit(): void {
-        this.buildComponentCssClass();
+    constructor() {
+        super();
+        this.contentDensityObserver.subscribe();
```
MariaIDineva, line 126: Maybe we need to add this subscription to `_subscriptions`?

libs/platform/button/button.component.ts
```diff
@@ -23,147 +25,168 @@ import { FDP_BUTTON } from './tokens';
     ],
     imports: [CoreButtonComponent],
     host: {
-        role: 'button'
+        role: 'button',
+        '[attr.index]': '-1'
```
MariaIDineva: Shouldn't that be `tabIndex`?

libs/core/button/base-button.ts
InnaAtanasova: yes, that's correct. It will allow `toggled` state to be updated programatically.

libs/core/button/button.component.ts
InnaAtanasova: The ContentDensityObserver.subscribe() subscription does not need to be added to this._subscriptions because it handles its own cleanup internally.

In the `content-density-observer.service.ts` (160-209), it:
- Uses takeUntilDestroyed(this._destroyRef) to automatically unsubscribe when the component is destroyed
- Calls this.complete() in its DestroyRef.onDestroy() callback, which cleans up all internal subscriptions

But it's a good point that you made, coz I didn't notice yesterday that the `_subscriptions` is not needed. I removed it from the code. Good catch!

libs/platform/button/button.component.ts
InnaAtanasova: Of course!  🤦‍♀️ 🤦‍♀️ 🤦‍♀️

Issue #13845, opened 2026-02-09
UI5 WebC Form broken


droshev:
### Describe your proposal

Broken Angular FormControl / reactive forms support
Another issue is degraded support for Angular reactive forms, which worked
correctly before the migration.

`ui5-switch`
Before migration
Works correctly with FormControl
After migration

The component no longer updates the form value
`ui5-input`
Used with formControlName inside a FormGroup
After migration, the binding stopped working correctly and throws error
This issue was observed in the development-settings component

### Which versions of Angular and Fundamental Library for Angular are affected?

latest

### Priority

High

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

PR #13850, opened 2026-02-10
fix(ui5): broken reactive forms in ui5 webc

MariaIDineva:
## Related Issue(s)

part of https://github.com/SAP/fundamental-ngx/issues/13845

## Description

Starting from the error below which was thrown when an app tried to use `ui-switch` with reactive forms, it turned out that not a 100% of UI5 form components support reactive forms integration:
Screenshot 2026-02-10 at 13 19 28

This PR:
- Extends `GenericControlValueAccessor` so that:
  - it takes into account a `CVA_CONFIG` which provides additional data related to:
    - which **property** the reactive form should bind to - in most cases that's the `value`, but it could be `checked` for example in the cases of `ui5-switch` and `ui5-checkbox`.
    - which **events** it should listen to and upon them update the corresponding property value.
    - **transformValue** function that would transform the value - for example for string values that would be `transformValue: (v) => v || ''`, for booleans like `ui5-switch` and `ui5-checkbox` that would be `transformValue: (v) => !!v`.
  - it provides support for setting initial disabled state on reactive forms;

- Fixes example for `ui5-input` with reactive forms and signals.
- Updates examples for `ui5-switch`, `ui5-checkbox` and `ui5-radio-button` to depict reactive forms usage - the examples are exactly the same as those that we have in our corresponding `core` components to make sure that the wrapper components have full compatibility with the core ones.

**NB!**  `ui5-radio-button` is a special case and there is additional logic around it because we still want to bind to the `value` property even though `checked` is also present under `members` and marked as the form property (spot the `_ui5formProperty` set to **true** in the custom elements definition for `RadioButton.js`):
```
            {
              "kind": "field",
              "name": "checked",
              "type": {
                "text": "boolean"
              },
              "default": "false",
              "description": "Defines whether the component is checked or not.\n\n**Note:** The property value can be changed with user interaction,\neither by clicking/tapping on the component,\nor by using the Space or Enter key.\n\n**Note:** Only enabled radio buttons can be checked.\nRead-only radio buttons are not selectable, and therefore are always unchecked.",
              "privacy": "public",
              "_ui5since": "1.0.0-rc.15",
              "_ui5formProperty": true,
              "_ui5formEvents": "change"
            },
```

## Screenshots

Before:

https://github.com/user-attachments/assets/ab3a7fe1-d35c-4481-8a73-331564f64e17

After:

https://github.com/user-attachments/assets/4842974b-da19-4579-a81e-1ca50623ff05

### DIFF
--- a/libs/docs/ui5-webcomponents/input/examples/input-sample.html
+++ b/libs/docs/ui5-webcomponents/input/examples/input-sample.html
@@ -245,12 +245,14 @@
                             id="bio"
                             formControlName="bio"
                             placeholder="Tell us about yourself"
+                            [valueState]="bioState()"
                             [maxlength]="maxLengthExample()"
                         ></ui5-input>
                     </div>
                 </div>
             </div>
         </form>
+
         <div class="sap-flex sap-flex--row sap-margin-top-medium sap-flex--gap-small sap-padding-x-medium">
             <ui5-button type="Submit" design="Emphasized" [disabled]="!isFormValid()"> Submit </ui5-button>
             <ui5-button design="Transparent" (ui5Click)="resetForm()"> Reset </ui5-button>
--- a/libs/docs/ui5-webcomponents/input/examples/input-sample.ts
+++ b/libs/docs/ui5-webcomponents/input/examples/input-sample.ts
@@ -3,6 +3,7 @@ import {
     AfterViewInit,
     Component,
     computed,
+    DestroyRef,
     effect,
     inject,
     Injector,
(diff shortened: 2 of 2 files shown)


libs/docs/ui5-webcomponents/check-box/examples/reactive-forms-sample.html
```diff
@@ -0,0 +1,47 @@
...
+    <small>
+        <strong>Form value:</strong> {{ registrationForm.getRawValue() | json }}<br />
+        <strong>Form Touched:</strong> {{ registrationForm.touched }}<br />
+        <strong>Form Dirty:</strong> {{ registrationForm.dirty }}<br />
+        <strong>Form Valid:</strong> {{ registrationForm.valid }}<br />
+    </small>
+</div>
+
+<style>
+    .checkbox-form {
```
droshev: can we avoid those custom CSS, or use the common-css for it?

MariaIDineva: OK, I'll see what I can do about it.

InnaAtanasova (approved):

PR #13851, opened 2026-02-10
fix(core): add fix for cut image in Illustrated Message

InnaAtanasova:
## Related Issue(s)

closes none

## Description
- brings the latest CSS code related to the recent design changes
- fix the problem with the cut image
- brings back the smallest BASE type and updates the test

**IMPORTANT**: the css changes are coming from https://github.com/SAP/fundamental-styles/pull/6237

## Screenshots
Before (image is cut):
Screenshot 2026-02-10 at 3 19 20 PM

After:
Screenshot 2026-02-10 at 3 19 03 PM

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.60.0-rc.8",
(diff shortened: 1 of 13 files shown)


InnaAtanasova:
> Maybe add a base example here as well?

Base example should not be set explicitly, it only appears on very small screens and acts as "default". If we add an example we might encourage users to use it.

PR #13853, opened 2026-02-12
fix(core): migrate BaseMultiCombobox to signals

MariaIDineva:
## Related Issue(s)

## Description

This PR migrates the `BaseMultiCombobox` class and `MultiComboboxComponent` from RxJS BehaviorSubject and mutable arrays to Angular signals.

🔄 Changes Made

**Core Signal Migrations** in **BaseMultiCombobox**

✅ Converted _selectedSuggestions from mutable array to signal<SelectableOptionItem<T>[]>
✅ Converted _suggestions from mutable array to signal<SelectableOptionItem[]>
✅ Converted _flatSuggestions from mutable array to signal<SelectableOptionItem[]>
✅ Converted _fullFlatSuggestions from mutable array to signal<SelectableOptionItem[]>
✅ Converted selectedShown$ from BehaviorSubject<boolean> to WritableSignal<boolean>
✅ Replaced direct mutations with immutable update patterns using .set() and .update()
MultiComboboxComponent ([multi-combobox.component.ts](vscode-webview://0bmbolssdisofrbbhu4iqlb2dr1jt6ik5pi6409t0n1qegbsbu4k/libs/core/multi-combobox/multi-combobox.component.ts))

✅ Migrated all properties to use signal inputs (input(), model())
✅ Converted outputs to signal-based output()
✅ Added computed() signals for derived state (e.g., isGroup)
✅ Updated all template bindings to call signals as functions (e.g., _suggestions())
✅ Converted selectedItems to use internal WritableSignal with readonly public accessor. Adds an abstract _setSelectedItems method that performs the internal writes to _selectedItem with coercion. Moves the _selectedItem field to the abstract class.
✅ Adds another abstract _inputText field that holds the value of the input "inputText" (because the input() is now immutable). We also have a new abstract method _setInputText used to programtically update the _inputText field, so now in the template we have:
```
[ngModel]="_inputText()"
(ngModelChange)="_setInputText($event); _searchTermChanged()"
```
That used to be:
```
[(ngModel)]="inputText"
(ngModelChange)="_searchTermChanged()"
```

**Template Updates** in **multi-combobox.component.html**

- Updated all signal references to use function call syntax: _suggestions(), _selectedSuggestions(), _flatSuggestions()
- Fixed conditional expressions to properly evaluate signal values
- Updated @for loops to iterate over signal values
mobile-multi-combobox.component.html
- Migrated from selectedShown$ | async to selectedShown() signal syntax
- Removed AsyncPipe dependency (no longer needed)

**Interface Updates**

- Updated **MultiCombobox** interface to reflect signal types
- Updated **MobileMultiComboboxInterface**

**Mobile Component Updates** in **mobile-multi-combobox.component.ts**

- Updated to work with signal-based parent component
- Changed selectedShown$ references to selectedShown signal
- Updated backup logic to read from signal: this._component._selectedSuggestions()
- Removed AsyncPipe from imports

🐛 **Some Bug Fixes**
- Off-by-one Error Fix - Fixed loop in _setSelectedSuggestions() that was using i <= length instead of i < length
- Immutable Updates - Ensured all array updates create new references instead of mutating existing objects
- Deselection Bug Fix - Items that are deselected properly in _setSelectedSuggestions method, they now have selected: false set, preventing stale state

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.60.1-rc.15",
+    "version": "0.60.1-rc.16",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.60.1-rc.15",
+    "version": "0.60.1-rc.16",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/cdk/package.json
+++ b/libs/cdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cdk",
-    "version": "0.60.1-rc.15",
+    "version": "0.60.1-rc.16",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - CDK",
     "license": "Apache-2.0",
--- a/libs/core/package.json
+++ b/libs/core/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/core",
-    "version": "0.60.1-rc.15",
+    "version": "0.60.1-rc.16",
     "schematics": "./schematics/collection.json",
     "ng-update": {
         "migrations": "./schematics/migrations.json"
--- a/libs/cx/package.json
+++ b/libs/cx/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cx",
-    "version": "0.60.1-rc.15",
+    "version": "0.60.1-rc.16",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - cx",
     "license": "Apache-2.0",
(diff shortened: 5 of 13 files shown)


InnaAtanasova (approved):

libs/core/multi-combobox/base-multi-combobox.class.ts
```diff
@@ -90,36 +224,55 @@ export abstract class BaseMultiCombobox<T = any> {
...
+     * In mobile mode, toggles between showing all items and showing only selected items.
+     * @hidden
+     */
+    selectedShown = signal(false);
+
+    /**
+     * Writable signal for overriding the invalid entry message with i18n translations.
+     * @hidden
+     */
+    protected readonly _invalidEntryMessageOverride = signal<Nullable<string>>(null);
```
InnaAtanasova: replace Nullable

libs/core/multi-combobox/multi-combobox.component.ts
```diff
@@ -134,55 +136,95 @@ export const FD_MAP_LIMIT = new InjectionToken<number>('Map limit≥', { factory
...
+     *  Whether the autocomplete should be enabled; Enabled by default.
+     * @defaultValue true
+     */
+    readonly autoComplete = input(true);
+
+    /**
+     * Whether list item options should be rendered as byline.
+     * @defaultValue false
+     */
+    readonly byline = input(false);
```
InnaAtanasova: the boolean inputs should be using booleanAttribute:
`readonly showSelectAll = input(false, { transform: booleanAttribute });`

InnaAtanasova (approved):
🚀

PR #13854, opened 2026-02-12
fix(core): migrate dialog system to Angular 21+ signals

InnaAtanasova:
## Related Issue(s)

## Description
Migrates all **dialog base** classes and components from decorator-based patterns to signal-based reactive architecture:

- Replace BehaviorSubject + combineLatest with computed signals
- Eliminate manual change detection (signals auto-notify)
- ResizeObserver + signals replace window resize listeners
- Computed signals cache and minimize recalculations
- Eliminate subscription management
- Replace lifecycle hooks with declarative patterns
- Convert @ContentChildren to contentChildren() signal queries
- Proper public/protected/private member ordering
- Comprehensive JSDoc documentation


libs/core/dialog/base/dialog-ref-base.class.ts
```diff
@@ -1,60 +1,183 @@
...
+        // Update signal state
+        this._statusSignal.set('dismissed');
+        this._closeResultSignal.set({
+            status: 'dismissed',
+            reason
+        });
+
+        // Maintain backward compatibility with RxJS API
+        this._endClose$.next();
+        this._endClose$.complete();
```
MariaIDineva, line 166: 👍

libs/core/title/title.component.ts
```diff
@@ -125,8 +125,19 @@ export class TitleComponent extends TitleToken {
      * @hidden
      */
     private _setHeaderSize(): void {
-        const headerSize =
-            this.headerSize() ?? this._defaultHeaderSize ?? this._elementRef.nativeElement.tagName.charAt(1);
+        let headerSize: HeaderSizes | string | null = this.headerSize() ?? this._defaultHeaderSize ?? null;
+
+        // If no explicit size provided, try to extract from element tag name
+        if (headerSize === null) {
```
MariaIDineva, line 131: Just a small question, as I don't understand this logic. What is a valid scenario where `headerSize` could actually be `null` as we always default to `_defaultHeaderSize` which is initially set and never updated?

InnaAtanasova: When a `[fd-title]` is used without:

- An explicit headerSize input
AND
- A parent component providing DEFAULT_TITLE_SIZE via DI

Then both values are null, and the fallback logic at line 131-139 kicks in to:
1. Extract the size from the element's tag name (if it's H1-H6)
2. Or default to 5 for non-heading elements

`_defaultHeaderSize` is not initially set, it's conditionally injected and will be null when no parent provider exists. This is true for standalone title components used outside of a dialog header or similar container.

InnaAtanasova:
```
@Component({
  template: `
    <h3 fd-title>My Page Title</h3>
    <div fd-title role="heading">Non-semantic heading</div>
  `
})
export class ExampleComponent {}
```

1. Line 76: `this.headerSize() → null` (as no input is provided)
2. `this._defaultHeaderSize → null` (no parent provides DEFAULT_TITLE_SIZE, so inject(..., { optional: true }) returns null)
3. Line 128: `headerSize = null ?? null ?? null` returns null
4. Line 131: Condition is true
5. Fallback logic runs:
For `<h3 fd-title>`: extracts "3" from tag name → applies fd-title--h3
For `<div fd-title>`: uses default 5 → applies fd-title--h5

MariaIDineva: Ooh, right, my bad, I thought DEFAULT_TITLE_SIZE was a constant and did not expand the whole file to see it is an injection token. Thanks for explaining!

PR #13884, opened 2026-02-16
fix: replace fast-deep-equal with fast-equals

robertIsaac:
reasons
1. fast-deep-equal doesn't support esm which causes problems with vitest
2. fast-deep-equal is not maintained anymore


droshev (approved):

robertIsaac:
when do you plan to merge this?

Issue #13859, opened 2026-02-13
Angular Signals Migration Phase 5 - Icon-Dependent Components


InnaAtanasova:
- [ ] Avatar
- [ ] GenericTag
- [ ] ObjectStatus
- [ ] ObjectMarker
- [ ] InfoLabel
- [ ] Badge
- [ ] Button
- [ ] MessageStrip
- [ ] Token

PR #13885, opened 2026-02-16
fix(core): update Generic Tag to signals

InnaAtanasova:
## Related Issue(s)

## Description
- migrates Generic Tag to signals
- updates unit tests
- improve documentation description

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.60.0-rc.13",
+    "version": "0.60.1-rc.0",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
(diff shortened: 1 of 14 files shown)


libs/core/generic-tag/generic-tag.component.ts
```diff
@@ -29,76 +28,62 @@ export type GenericTagType = 'error' | 'success' | 'warning' | 'information';
...
-
+export class GenericTagComponent implements HasElementRef {
     /**
      * The type of the Generic Tag.
      * Can be one of the following: 'error' | 'success' | 'warning' | 'information'.
      * For default Generic Tag omit this property.
      */
-    @Input()
-    type: Nullable<GenericTagType>;
+    readonly type = input<GenericTagType>();
```
MariaIDineva: Shouldn't that be `input<GenericTagType | null | undefined>()`
while the one below - `input.required<string>()`?

PR #13886, opened 2026-02-17
feat(core,platform): migrate popover and menu to Angular signals

droshev:
feat(core): migrate popover and menu to Angular signals
BREAKING CHANGE: PopoverBodyComponent.onClose is now an OutputEmitterRef instead of Subject. Use .emit() instead of .next() and wrap with outputToObservable() for RxJS operators.

Migrate popover, menu, and related components to use Angular signals:

Core Changes:
- Convert popover service to use signals for state management
- Add PopoverConfig interface for configuration objects
- Convert menu component to use signal-based inputs/outputs
- Migrate inline-help directive to use signals and effects
- Update related components (split-button, product-switch, 
  multi-combobox, shellbar menus, etc.)

Key Updates:
- Add config input for batch configuration
- Add beforeOpen event and isOpenChange output
- Fix menu focusTrapped default (restore to true)
- Fix memory leaks in submenu subscriptions

Test Updates:
- Rewrite popover.service.spec.ts for signal-based API
- Update menu component tests for new patterns
- Fix affected platform component tests


libs/core/popover/popover-mobile/popover-mobile.component.ts
```diff
@@ -79,9 +84,12 @@ export class PopoverMobileComponent extends MobileModeBase<PopoverInterface> imp
     /** @hidden */
     constructor(
         private _changeDetectorref: ChangeDetectorRef,
-        @Inject(POPOVER_COMPONENT) _popoverComponent: PopoverInterface
+        @Inject(POPOVER_COMPONENT) _popoverComponent: PopoverInterface,
+        private _injector: Injector
     ) {
         super(_popoverComponent, MobileModeControl.POPOVER);
+        // Create observable in constructor to preserve injection context for toObservable()
+        this._isOpen$ = runInInjectionContext(this._injector, () => toObservable(this._component.isOpen));
```
MariaIDineva: I think we can get rid of the **_isOpen$** observable and **_listenOnPopoverOpenChange** method if we add an effect instead, using the component **isOpen** signal as it is, not making it an observable:
```
        effect(() => {
            const isOpen = this._component.isOpen();

            // Run side effects in untracked to prevent infinite loops
            untracked(() => {
                if (isOpen) {
                    this._openDialog();
                } else {
                    this.dialogRef?.hide(true);
                }
            });
        });
```

libs/btp/navigation/components/navigation-item/navigation-list-item.component.ts
```diff
@@ -674,7 +678,13 @@ export class NavigationListItemComponent extends FdbNavigationListItem implement
         if (!this.hasChildren$()) {
             return;
         }
-        this.expanded$.update((expanded) => !expanded);
+
+        // In snapped mode, toggle popover instead of expanded state
+        if (this.navigation.isSnapped$()) {
```
MariaIDineva, line 683: I know that's how it has been and is not something introduced in this PR and we may choose to update it in a separate PR, but this `$` at the end of these fields is quite misleading to me as the Angular convention is to have the suffix only for observables while those are signals. Anyway, feel free to disregard if you find it inappropriate to update here, but I think at some point we should take care of updating those to improve readability.

libs/core/popover/popover-service/popover.service.ts
```diff
@@ -291,29 +431,91 @@ export class PopoverService extends BasePopoverClass {
...
-        this._placementContainer = config.placementContainer;
-        const onlyChanged = Object.keys(new BasePopoverClass()).filter((key) => this[key] !== config[key]);
+    refreshConfiguration(config: PopoverConfig): void {
+        // Helper function to unwrap signal values
+        const unwrap = <T>(val: T | Signal<T>): T => (isSignal(val) ? val() : val) as T;
+
+        this._placementContainerValue = unwrap(config.placementContainer) ?? this._placementContainerValue;
+
+        // Update signals if config values are provided
+        if (config.closeOnNavigation !== undefined) {
```
MariaIDineva, line 439: Maybe we can reduce the complexity of this method if we replace those `if`-s up to line 506 with something like:
```
        // Define signal property mappings
        const signalMappings: Array<{
            configKey: keyof PopoverConfig;
            signal: ReturnType<typeof signal<any>>;
        }> = [
            { configKey: 'closeOnNavigation', signal: this.closeOnNavigation },
            { configKey: 'disabled', signal: this.disabled },
            { configKey: 'maxWidth', signal: this.maxWidth },
            { configKey: 'noArrow', signal: this.noArrow },
            { configKey: 'additionalBodyClass', signal: this.additionalBodyClass },
            { configKey: 'additionalBodyComponentClasses', signal: this.additionalBodyComponentClasses },
            { configKey: 'additionalTriggerClass', signal: this.additionalTriggerClass },
            { configKey: 'closeOnEscapeKey', signal: this.closeOnEscapeKey },
            { configKey: 'disableScrollbar', signal: this.disableScrollbar },
            { configKey: 'placement', signal: this.placement },
            { configKey: 'closeOnOutsideClick', signal: this.closeOnOutsideClick },
            { configKey: 'applyOverlay', signal: this.applyOverlay },
            { configKey: 'focusTrapped', signal: this.focusTrapped },
            { configKey: 'focusAutoCapture', signal: this.focusAutoCapture },
            { configKey: 'restoreFocusOnClose', signal: this.restoreFocusOnClose },
            { configKey: 'scrollStrategy', signal: this.scrollStrategy },
            { configKey: 'cdkPositions', signal: this.cdkPositions },
            { configKey: 'fillControlMode', signal: this.fillControlMode },
            { configKey: 'appendTo', signal: this.appendTo },
            { configKey: 'placementContainer', signal: this.placementContainer },
            { configKey: 'fixedPosition', signal: this.fixedPosition },
            { configKey: 'resizable', signal: this.resizable }
        ];

        // Update all signals from config
        signalMappings.forEach(({ configKey, signal }) => {
            if (config[configKey] !== undefined) {
                signal.set(unwrap(config[configKey]));
            }
        });
```

libs/btp/navigation/components/navigation-item/navigation-list-item.component.ts
InnaAtanasova: +1

libs/btp/tool-header/src/components/tool-header-product-switch.component.ts
```diff
@@ -49,14 +48,42 @@ import { FdTranslatePipe } from '@fundamental-ngx/i18n';
...
+export class ToolHeaderProductSwitchComponent {
     /** Placement of a popover. */
-    @Input()
-    override placement: Placement = 'bottom-end';
+    readonly placement = input<Placement>('bottom-end');
 
     /** Whether the product switch is disabled. */
-    @Input()
-    override disabled = false;
+    readonly disabled = input(false);
```
InnaAtanasova, line 61: Boolean inputs should be like this:
`readonly disabled = input(false, { transform: booleanAttribute });`

libs/btp/tool-header/src/components/tool-header-product-switch.component.ts
```diff
@@ -49,14 +48,42 @@ import { FdTranslatePipe } from '@fundamental-ngx/i18n';
...
+    readonly closeOnEscapeKey = input(true);
+
+    /** Whether the popover should close when a click is made outside its boundaries. */
+    readonly closeOnOutsideClick = input(true);
+
+    /** Whether the popover should have an arrow. */
+    readonly noArrow = input(true);
+
+    /** Whether to wrap content with fd-scrollbar directive. */
+    readonly disableScrollbar = input(true);
```
InnaAtanasova, line 73: Please use `transform: booleanAttribute`

libs/btp/tool-header/src/components/tool-header-product-switch.component.ts
```diff
@@ -49,14 +48,42 @@ import { FdTranslatePipe } from '@fundamental-ngx/i18n';
...
+
+    /** The trigger events that will open/close the popover. */
+    readonly triggers = input<(string | TriggerConfig)[]>(['click']);
+
+    /** Whether the popover should be focusTrapped. */
+    readonly focusTrapped = input(false);
+
+    /** Whether the popover should automatically move focus into the trapped region. */
+    readonly focusAutoCapture = input(true);
```
InnaAtanasova, line 82: Please use `transform: booleanAttribute`

libs/core/inline-help/inline-help.directive.ts
```diff
@@ -36,43 +44,34 @@ let inlineHelpId = 0;
...
-            template: content instanceof TemplateRef ? content : null
-        };
-        this._popoverService.updateContent(text, template);
+    /** @hidden Internal body ID for ARIA */
+    protected _bodyId = '';
 
-        this._setupScreenreaderElement(content);
-    }
+    /** @hidden */
+    protected _describedBy = '';
```
InnaAtanasova: this was public before

libs/core/inline-help/inline-help.directive.ts
```diff
@@ -85,36 +84,77 @@ export class InlineHelpDirective extends BasePopoverClass implements OnInit, OnC
...
-        }
-        this._popoverService.refreshConfiguration(this);
+        // Apply additional inline help classes
+        this._applyAdditionalInlineHelpClass();
+
+        // Initialize popover configuration for inline help
+        // This ensures the proper configuration is set
+        this._popoverService.noArrow.set(false);
+        this._popoverService.closeOnEscapeKey.set(false);
+        this._popoverService.additionalBodyClass.set(this._additionalBodyClass);
```
InnaAtanasova: I don't understand why you need these here when you set them in `ngOnInit` when you configure the popover service via `config`.

libs/core/inline-help/inline-help.directive.ts
```diff
@@ -85,36 +84,77 @@ export class InlineHelpDirective extends BasePopoverClass implements OnInit, OnC
...
 
         // If connected to the icon, but not button, then apply additional class
         // That will change the arrow's position a bit
         if (this._icon) {
-            this.additionalBodyClass += ' ' + INLINE_HELP_ICON_CLASS;
+            classes += ' ' + INLINE_HELP_ICON_CLASS;
         }
+
+        this._additionalBodyClass = classes;
+        this._popoverService.additionalBodyClass.set(classes);
```
InnaAtanasova: I think this line is redundant. The classes are "static", they are read once when icon is injected and doesn't change after construction, so the value is already read.

libs/core/menu/directives/menu-trigger.directive.ts
```diff
@@ -37,6 +38,9 @@ export class MenuTriggerDirective implements OnDestroy {
     /** @hidden */
     private _menuSubscription: Subscription = new Subscription();
 
+    /** @hidden */
+    private readonly _injector = inject(Injector);
+
     /** @hidden */
     constructor(private _elementRef: ElementRef) {}
```
InnaAtanasova: this should be `private _elementRef = inject(ElementRef)` outside of the constructor.

libs/core/menu/menu-item/menu-item.component.html
```diff
@@ -4,7 +4,7 @@
         class="fd-menu__sublist"
         role="menu"
         [attr.id]="itemId"
-        [attr.aria-hidden]="!submenuVisible"
+        [attr.aria-hidden]="!submenuVisible ? 'true' : 'false'"
```
InnaAtanasova: In my opinion this is more readable:
`[attr.aria-hidden]="submenuVisible ? 'false' : 'true'"`
Double negation is always kinda confusing.

libs/core/menu/menu-mobile/menu-mobile.component.ts
```diff
@@ -102,15 +107,36 @@ export class MenuMobileComponent extends MobileModeBase<MenuInterface> implement
     /** @hidden */
     private readonly _ngZone = inject(NgZone);
 
+    /** @hidden */
+    private readonly _injector = inject(Injector);
+
     /** @hidden */
     constructor(@Inject(MENU_COMPONENT) menuComponent: MenuInterface) {
```
InnaAtanasova: can you inject menu component with `inject`?

libs/core/menu/menu-mobile/menu-mobile.component.ts
```diff
@@ -102,15 +107,36 @@ export class MenuMobileComponent extends MobileModeBase<MenuInterface> implement
...
+                    if (isOpen && !this.dialogRef) {
+                        // Only open if not already open
+                        this._openDialog();
+                    } else if (!isOpen && this.dialogRef) {
+                        // Only close if currently open
+                        this.dialogRef.close();
+                    }
+                });
+            },
+            { injector: this._injector }
```
InnaAtanasova: Why do you need an injector in the constructor?

libs/core/menu/menu-mobile/menu-mobile.component.ts
```diff
@@ -130,13 +156,22 @@ export class MenuMobileComponent extends MobileModeBase<MenuInterface> implement
...
-            container: this._elementRef.nativeElement
+        // Run inside NgZone to ensure proper change detection for dialog rendering
+        this._ngZone.run(() => {
+            const config = {
+                mobile: true,
+                disablePaddings: true,
+                ...this.dialogConfig,
+                escKeyCloseable: false,
+                backdropClickCloseable: false
+            };
```
InnaAtanasova: You need to try to get rid of NgZone.

libs/core/menu/menu-mobile/menu-mobile.component.ts
```diff
@@ -102,15 +107,36 @@ export class MenuMobileComponent extends MobileModeBase<MenuInterface> implement
     /** @hidden */
     private readonly _ngZone = inject(NgZone);
```
InnaAtanasova: we are going towards zoneless, try to see if we can get rid of it in this case.

libs/core/menu/menu.component.ts
```diff
@@ -60,329 +63,484 @@ let menuUniqueId = 0;
...
+})
+export class MenuComponent implements MenuInterface, AfterContentInit, AfterViewInit, OnDestroy {
+    /** Open submenu on hover after given milliseconds */
+    readonly openOnHoverTime = input(0);
 
-    get mobile(): boolean {
-        return this._mobile;
-    }
+    /** Aria-label for navigation */
+    readonly ariaLabel = input<Nullable<string>>(null);
```
InnaAtanasova: we decided to get rid of Nullable

libs/core/menu/menu.component.ts
```diff
@@ -60,329 +63,484 @@ let menuUniqueId = 0;
...
 
-    /** Whether the popover is disabled. */
-    @Input()
-    disabled = false;
+    /** Aria-Labelledby for element describing navigation */
+    readonly ariaLabelledby = input<Nullable<string>>(null);
```
InnaAtanasova: same as above

libs/core/menu/menu.component.ts
```diff
@@ -60,329 +63,484 @@ let menuUniqueId = 0;
...
      */
-    @Input()
-    focusAutoCapture = true;
+    readonly config = input<PopoverConfig>({});
 
-    /** Open submenu on hover after given milliseconds */
-    @Input()
-    openOnHoverTime = 0;
+    /** Whether the menu should be displayed in mobile mode */
+    readonly mobile = input(false);
```
InnaAtanasova: please use booleanAttribute

libs/core/menu/menu.component.ts
```diff
@@ -60,329 +63,484 @@ let menuUniqueId = 0;
...
-    @Input()
-    id = `fd-menu-${menuUniqueId++}`;
+    /** Whether to auto-capture focus when opened */
+    readonly focusAutoCapture = input(true);
+
+    /** Whether the menu/popover is disabled */
+    readonly disabled = input(false);
+
+    /** Whether to disable scrollbar */
+    readonly disableScrollbar = input(false);
```
InnaAtanasova: same as above

libs/core/menu/menu.component.ts
```diff
@@ -60,329 +63,484 @@ let menuUniqueId = 0;
...
+    readonly fillControlMode = input<PopoverFillMode | null>(null);
+
+    /** Whether to close on outside click */
+    readonly closeOnOutsideClick = input(true);
+
+    /** Whether to hide the arrow */
+    readonly noArrow = input(true);
+
+    /** Whether the popover should trap focus within its boundaries */
+    readonly focusTrapped = input(true);
```
InnaAtanasova: same as above

libs/core/menu/menu.component.ts
```diff
@@ -60,329 +63,484 @@ let menuUniqueId = 0;
...
-    _segmentedButtonOptions: QueryList<SegmentedButtonOptionDirective<unknown>>;
+    protected readonly _segmentedButtonOptions = contentChildren(SegmentedButtonOptionDirective);
 
     /** @hidden */
-    _navContainerRole = 'menu';
+    protected _navContainerRole = 'menu';
 
     /** @hidden */
-    _menuListContainerRole = 'none';
+    protected _menuListContainerRole = 'none';
```
InnaAtanasova: these should remain public

libs/core/popover/popover.component.ts
```diff
@@ -65,143 +68,270 @@ let cdkPopoverUniqueId = 0;
...
 
     /** @hidden - template for Dialog body content */
-    @ContentChild('popoverBodyContent')
-    popoverBodyContentTemplate: TemplateRef<any>;
+    protected readonly popoverBodyContentTemplate = contentChild<TemplateRef<any>>('popoverBodyContent');
 
     /** @hidden - template for Dialog footer content */
-    @ContentChild('popoverFooterContent')
-    popoverFooterContentTemplate: TemplateRef<any>;
+    protected readonly popoverFooterContentTemplate = contentChild<TemplateRef<any>>('popoverFooterContent');
```
InnaAtanasova: I really don't think these should be protected.

libs/btp/navigation/components/navigation-item/navigation-list-item.component.ts
```diff
@@ -783,11 +799,13 @@ export class NavigationListItemComponent extends FdbNavigationListItem implement
...
                     try {
-                        popover.popoverBody._focusFirstTabbableElement(true);
+                        const popoverBody = (popover as any).popoverBody();
+                        popoverBody?._focusFirstTabbableElement(true);
                     } catch (error) {
                         console.error('Error focusing in popover:', error);
 
-                        const popoverBodyElement = popover.popoverBody?._elementRef?.nativeElement;
+                        const popoverBody = (popover as any).popoverBody();
+                        const popoverBodyElement = popoverBody?._elementRef?.nativeElement;
```
InnaAtanasova: You can get rid of "as any" if you change popoverBody to public again in `libs/core/popover/popover.component.ts`.
It should be `readonly popoverBody = contentChild(PopoverBodyComponent);` and then you can simply use `const popoverBody = popover.popoverBody();` instead of `const popoverBody = (popover as any).popoverBody();`

libs/core/popover/popover.component.ts
InnaAtanasova: Making these protected forces type assertions aka if we want to use any of these in other components we need to use `as any`. These are readonly queries and are not mutatable so not really needed to protect them.

libs/btp/navigation/components/navigation-item/navigation-list-item.component.ts
droshev: @MariaIDineva @InnaAtanasova I will disregard it for 2 reasons:
- this will be a breaking change
- the scope of the PR is popover-base class, not the navigation-item in btp

libs/core/popover/popover-service/popover.service.ts
droshev: @MariaIDineva I see your points. They make sense. But I won't change that for 3 reasons:
- type safety: Needs careful typing to maintain type checking
- we have 2 special special cases: `isOpen` and `triggers` have special handling
- debug and testing: tracing will be harder to follow

libs/core/menu/menu-mobile/menu-mobile.component.ts
droshev: I moved it to the super. When extending a class in TypeScript, you cannot access this before calling super()

libs/core/menu/menu-mobile/menu-mobile.component.ts
droshev: removed

PR #13891, opened 2026-02-17
fix(core): migrate Message Strip to signals

InnaAtanasova:
## Related Issue(s)

## Description
- migration to signals
- update unit test
- code cleanup
- JS docs improvements

### DIFF
--- a/apps/docs/project.json
+++ b/apps/docs/project.json
@@ -246,7 +246,7 @@
                     "includePaths": ["node_modules"]
                 },
                 "scripts": ["node_modules/marked/marked.min.js"],
-                "allowedCommonJsDependencies": ["moment", "highlight.js", "fast-deep-equal", "focus-trap", "dayjs"],
+                "allowedCommonJsDependencies": ["moment", "highlight.js", "focus-trap", "dayjs"],
                 "extractLicenses": false,
                 "buildOptimizer": false,
                 "sourceMap": true,
--- a/eslint.overrides.js
+++ b/eslint.overrides.js
@@ -8,15 +8,6 @@ module.exports = [
                     paths: [
                         'rxjs/Rx',
                         {
-                            name: 'lodash',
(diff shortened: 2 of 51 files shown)


libs/core/message-strip/alert/message-strip-alert/message-strip-alert.component.ts
```diff
@@ -35,40 +35,44 @@ export type MessageStripAlertPortalType<ComponentType> =
...
     imports: [MessageStripComponent, PortalModule, AutoDismissMessageStripDirective]
 })
-export class MessageStripAlertComponent<ComponentType = unknown>
-    implements MessageStripAlert, OnDestroy, AfterViewInit
-{
+export class MessageStripAlertComponent<ComponentType = unknown> implements MessageStripAlert {
     /** @hidden */
-    @ViewChild(AutoDismissMessageStripDirective)
-    autoDismissMessageStripDirective: AutoDismissMessageStripDirective;
+    readonly autoDismissMessageStripDirective = viewChild.required(AutoDismissMessageStripDirective);
```
MariaIDineva, line 40: Is this directive required?

InnaAtanasova: Yes, the component calls `this.autoDismissMessageStripDirective().open()` in the `afterNextRender()` callback (line 91) and it would fail if the directive wasn't present.
With `viewChild.required()`, the return type is `Signal<AutoDismissMessageStripDirective>` instead of `Signal<AutoDismissMessageStripDirective | undefined>`, so we don't need to check for `undefined` before calling methods on it.
If the directive is missing from the template, Angular will throw a clear error immediately, making debugging easier than getting a runtime error when trying to call `.open()` on undefined. TypeScript knows the value will always exist, preventing null/undefined access errors.

Issue #13855, opened 2026-02-13
Angular Signals Migration Phase 1 - Base Classes & Services


InnaAtanasova:
**Platform Base Classes**

- [ ] BaseComponent - Maria
- [ ] BaseInput
- [ ] CollectionBaseInput
- [ ] InLineLayoutCollectionBaseInput
- [ ] BaseListItem
- [ ] BaseCombobox
- [ ] BaseMultiCombobox
- [ ] BaseSelect
- [ ] BaseMultiInput
- [ ] BaseStepInput
- [ ] BaseDynamicFormGeneratorControl
- [ ] BaseWizardGenerator
- [ ] BaseSettingsGeneratorLayout
- [ ] BaseSmartFilterBarConditionField
- [ ] VhdBaseTab
- [ ] IconTabBarBase
- [ ] IconTabBarPopoverBase
- [ ] ClosableIconTabBar
- [ ] BaseDataProvider

**Core Base Classes**

- [x] BasePopoverClass - Deno [13886](https://github.com/SAP/fundamental-ngx/pull/13886)
- [x] BaseButton - Inna [13847](https://github.com/SAP/fundamental-ngx/pull/13847)
- [x] CardFocusItem - Maria [13905](https://github.com/SAP/fundamental-ngx/pull/13905)
- [x] BaseTreeItem - (no changes needed, abstract class) Inna [13909](https://github.com/SAP/fundamental-ngx/pull/13909)
- [x] DialogBase - Inna [13854](https://github.com/SAP/fundamental-ngx/pull/13854)
- [x] DialogBaseService - Inna [13854](https://github.com/SAP/fundamental-ngx/pull/13854)
- [x] DialogHeaderBase - Inna [13854](https://github.com/SAP/fundamental-ngx/pull/13854)
- [x] DialogFooterBase - Inna [13854](https://github.com/SAP/fundamental-ngx/pull/13854)
- [x] DialogContentBase - Inna [13854](https://github.com/SAP/fundamental-ngx/pull/13854)
- [x] DialogRefBase - Inna [13854](https://github.com/SAP/fundamental-ngx/pull/13854)
- [x] DialogConfigBase - Inna [13854](https://github.com/SAP/fundamental-ngx/pull/13854)
- [x] MobileModeBase - Inna [13908](https://github.com/SAP/fundamental-ngx/pull/13908)
- [x] BaseMultiCombobox - Maria [13853](https://github.com/SAP/fundamental-ngx/pull/13853)
[...]

PR #13894, opened 2026-02-17
fix(core): migrate dynamic page actions to signals

droshev:
part of https://github.com/SAP/fundamental-ngx/issues/13855

migrate dynamic page actions to signals

### DIFF
--- a/libs/core/dynamic-page/dynamic-page-header/actions/dynamic-page-base-actions.ts
+++ b/libs/core/dynamic-page/dynamic-page-header/actions/dynamic-page-base-actions.ts
@@ -1,13 +1,16 @@
 import { Directive, ElementRef, inject } from '@angular/core';
+import { HasElementRef } from '@fundamental-ngx/cdk/utils';
 
 @Directive()
-export class DynamicPageBaseActions {
-    /** @hidden */
-    protected readonly elementRef = inject(ElementRef);
+export class DynamicPageBaseActions implements HasElementRef {
+    /** Element reference to the host element. */
+    readonly elementRef = inject(ElementRef);
 
-    /** @hidden */
-    protected addClassToToolbar(_class: string): void {
-        // adds global actions classes to its toolbar
+    /**
+     * Adds a CSS class to the toolbar element within this component.
+     * @param _class The CSS class name to add to the toolbar.
+     */
+    addClassToToolbar(_class: string): void {
         const toolbarEl = this.elementRef.nativeElement.querySelector('.fd-toolbar');
         if (toolbarEl) {
             toolbarEl.classList.add(_class);
--- a/libs/core/dynamic-page/dynamic-page-header/actions/dynamic-page-global-actions.component.ts
+++ b/libs/core/dynamic-page/dynamic-page-header/actions/dynamic-page-global-actions.component.ts
@@ -2,13 +2,16 @@ import {
     afterNextRender,
     ChangeDetectionStrategy,
     Component,
+    computed,
     contentChild,
+    effect,
     inject,
     Injector,
     ViewEncapsulation
 } from '@angular/core';
 import { ToolbarComponent } from '@fundamental-ngx/core/toolbar';
-import { DYNAMIC_PAGE_CLASS_NAME, DynamicPageResponsiveSize } from '../../constants';
+import { DYNAMIC_PAGE_CLASS_NAME } from '../../constants';
+import { DynamicPageService } from '../../dynamic-page.service';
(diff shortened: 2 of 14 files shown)


MariaIDineva (approved):

libs/core/dynamic-page/dynamic-page-header/actions/dynamic-page-base-actions.ts
```diff
@@ -1,11 +1,14 @@
-import { Directive, ElementRef } from '@angular/core';
+import { Directive, ElementRef, inject } from '@angular/core';
 
 @Directive()
 export class DynamicPageBaseActions {
```
InnaAtanasova: should implement `HasElementRef` to make sure we don't remove ElementRef by accident.

libs/core/dynamic-page/dynamic-page-header/actions/dynamic-page-base-actions.ts
```diff
@@ -1,11 +1,14 @@
...
     /** @hidden */
-    addClassToToolbar(_class: string, elementRef: ElementRef): void {
+    protected readonly elementRef = inject(ElementRef);
```
InnaAtanasova: shouldn't be protected, was public before so might be used anywhere and this would be a breaking changes.

libs/core/dynamic-page/dynamic-page-header/actions/dynamic-page-global-actions.component.ts
```diff
@@ -18,36 +18,42 @@ import { DynamicPageBaseActions } from './dynamic-page-base-actions';
...
+    private _handleOverflow(toolbar: ToolbarComponent, shouldBeHidden: boolean): void {
+        toolbar.forceOverflow = shouldBeHidden;
+        toolbar.shouldOverflow = shouldBeHidden;
+        toolbar.detectChanges();
+        afterNextRender(
+            () => {
+                toolbar.updateCollapsibleItems();
+            },
+            { injector: this._injector }
+        );
```
InnaAtanasova, line 67: Maybe consider this to be an effect
```
export class DynamicPageGlobalActionsComponent extends DynamicPageBaseActions {
    /** @hidden */
    readonly _toolbarComponent = contentChild(ToolbarComponent);

    /** @hidden */
    private readonly _overflowState = signal<boolean | null>(null);

    constructor() {
        super();

        // Add toolbar class after DOM is rendered (one-time initialization)
        afterNextRender(() => {
            if (this._toolbarComponent()) {
                this.addClassToToolbar(DYNAMIC_PAGE_CLASS_NAME.dynamicPageToolbar);
            }
        });

        // React to overflow state changes and update collapsible items (reactive)
        effect(() => {
            const shouldBeHidden = this._overflowState();
            const toolbar = this._toolbarComponent();
            
            if (toolbar && shouldBeHidden !== null) {
                toolbar.forceOverflow = shouldBeHidden;
                toolbar.shouldOverflow = shouldBeHidden;
                toolbar.updateCollapsibleItems();
            }
        });
    }

    /** @hidden */
    _setSize(size: DynamicPageResponsiveSize): void {
        // Update the overflow state signal to trigger the effect
        this._overflowState.set(size === 'small');
    }
}

```
The benefits of this would be:
- Automatically respond to `_overflowState` signal changes
- Handle toolbar availability changes if needed
- Eliminate manual change detection

libs/core/dynamic-page/dynamic-page-header/actions/dynamic-page-layout-actions.component.ts
```diff
@@ -19,20 +11,19 @@ import { DynamicPageBaseActions } from './dynamic-page-base-actions';
...
-    },
-    standalone: true
+    }
 })
-export class DynamicPageLayoutActionsComponent extends DynamicPageBaseActions implements AfterContentInit {
+export class DynamicPageLayoutActionsComponent extends DynamicPageBaseActions {
     /** @hidden */
-    @ContentChild(ToolbarComponent)
-    toolbarComponent: ToolbarComponent;
+    readonly toolbarComponent = contentChild(ToolbarComponent);
```
InnaAtanasova: Why do we need this `toolbarComponent` here? I don't see it's used in the class.

libs/core/dynamic-page/dynamic-page-header/actions/dynamic-page-base-actions.ts
droshev: done!

Issue #13895, opened 2026-02-18
Range Error: Maximum call stack size exceeded in Platform Settings Generator


MariaIDineva:
### Describe your proposal

When accessing **Platform Settings Generator** - https://fundamental-ngx.netlify.app/#/platform/settings-generator the following error is visible in the console:

<img width="1901" height="1089" alt="Image" src="https://github.com/user-attachments/assets/080ae302-c3b3-4cf0-b37d-8a698bb556d6" />

The issue was most probably introduced in [this PR](https://github.com/SAP/fundamental-ngx/pull/13645) as the error seems to not be present in the previous versions.

### Which versions of Angular and Fundamental Library for Angular are affected?

latest

### Affected Component

Platform Settings Generator

### Steps to Reproduce (for bugs)

1. Visit https://fundamental-ngx.netlify.app/#/platform/settings-generator
2. Open the browser console and spot the error.

### Priority

None

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

robertIsaac:
I took a look now
it seems a problem when we pass `Observable` or `FormControl` to the `deepClone` function
for Observable this comes from the fact that we support default as observable e.g.
https://github.com/SAP/fundamental-ngx/blob/2b6f375c79a0040e9564e79edb6372405e4ce878/libs/docs/platform/settings-generator/examples/default/settings-generator-default-example.component.ts#L136

the question now is, do we actually need here deep merge?

PR #13896, opened 2026-02-18
fix(cdk, platform): provide fix for max call stack size exceeded

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13895

## Description
**Problem:**
Stack overflow error (RangeError: Maximum call stack size exceeded) caused by `cloneDeep` function not handling circular references. When objects contained self-references or circular dependencies, the function would recurse infinitely. Additionally, `structuredClone` usage in form-generator service failed when cloning objects containing functions or RxJS observables.

**Solution:**

Added circular reference detection to `cloneDeep` using WeakMap to track visited objects
Replaced `structuredClone` with `cloneDeep` in form-generator service (lines 213, 288) to properly handle functions and complex objects
Added edge case handling for DOM elements, Error objects, and defensive array validation

**Changes:**

- `lodash-utils.ts`: Enhanced cloneDeep with circular reference tracking and edge case handling
- `form-generator.service.ts`: Replaced structuredClone with cloneDeep
- Updated tests to verify circular reference handling

### DIFF
--- a/libs/cdk/utils/functions/lodash-utils.spec.ts
+++ b/libs/cdk/utils/functions/lodash-utils.spec.ts
@@ -124,10 +124,38 @@ describe('lodash-utils', () => {
         });
 
         it('should handle circular references gracefully', () => {
-            const obj: any = { a: 1 };
+            const obj: any = { a: 1, b: { c: 2 } };
             obj.self = obj;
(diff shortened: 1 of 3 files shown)


MariaIDineva (commented):
The reported error is gone now, but look what happens in the UI - spot the [object Object] values in the inputs:

Image

PR #13898, opened 2026-02-18
fix(core): fix notification aria-describedby bug and migrate to Angular 21 signals

droshev:
- Fix bug where aria-describedby was not being set on action buttons (merge(...buttonsChanges$) spread operator was missing)
- Migrate NotificationGroupBaseDirective 
- Migrate NotificationGroupHeaderComponent

### DIFF
--- a/libs/core/notification/notification-group/notification-group.component.ts
+++ b/libs/core/notification/notification-group/notification-group.component.ts
@@ -93,7 +93,7 @@ export class NotificationGroupComponent implements OnInit, AfterViewInit {
 
     /** @hidden */
     ngAfterViewInit(): void {
-        this.groupHeader()?.ariaControls.set(this.groupList()?.id());
+        this.groupHeader()?.ariaControls.set(this.groupList()?.id() ?? null);
         this.groupHeader()?.expanded.set(this.expanded());
     }
 


libs/core/notification/notification-utils/notification-group-base.ts
```diff
@@ -22,40 +25,89 @@ export abstract class NotificationGroupBaseDirective implements AfterViewInit {
...
-                    .toArray()
-                    .reduce((acc, c) => acc.concat(c.buttons.changes), [] as Observable<any>[]);
-                return merge(buttonsChanges$).pipe(startWith(0));
-            })
-        );
+    private readonly _injector = inject(Injector);
+
+    constructor() {
+        // Use afterNextRender to ensure DOM is ready before setting up the observable chain
+        afterNextRender(() => this._setupAriaDescribedBy(), { injector: this._injector });
```
MariaIDineva, line 32: I don't think we need injector here since the call is inside the constructor.

libs/core/notification/notification-utils/notification-group-base.ts
```diff
@@ -22,40 +25,89 @@ export abstract class NotificationGroupBaseDirective implements AfterViewInit {
...
+     */
+    private _setupAriaDescribedBy(): void {
+        const headersChanges$ = this._createHeadersObservable();
+        const actionsChanges$ = this._createActionsObservable();
+
         merge(headersChanges$, actionsChanges$)
             .pipe(
-                debounceTime(10), // omitting extra emissions
-                filter(() => this.notificationHeader.length > 0 && this.notificationActions.length > 0),
+                auditTime(0), // batch emissions within the same tick
```
MariaIDineva, line 45: Was this change on purpose - replacing `debounceTime(10)` with `auditTime(0)`?

PR #13913, opened 2026-02-23
fix(cdk,core): deprecate fdkRepeat and use @for

droshev:
- Deprecate fdkRepeat in favor of @for control flow with range()
- Consolidate range functions from datetime into cdk/utils
- Migrate all usages in core, platform, and docs
- Add comprehensive tests for range() and loading states

### DIFF
--- a/libs/docs/btp/tool-header/examples/auto-mode-example/auto-mode-example.component.html
+++ b/libs/docs/btp/tool-header/examples/auto-mode-example/auto-mode-example.component.html
@@ -23,7 +23,7 @@
                 }"
             >
                 <h2 [style.margin]="0" fdPopoverBodyHeader>Notifications</h2>
-                @for (i of notificationRange; track i) {
+                @for (i of [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]; track i) {
                     <fd-message-strip> A new notification has arrived. {{ i }} </fd-message-strip>
                 }
             </div>
--- a/libs/docs/btp/tool-header/examples/basic-example/tool-header-basic-example.component.html
+++ b/libs/docs/btp/tool-header/examples/basic-example/tool-header-basic-example.component.html
@@ -28,7 +28,7 @@
                 }"
             >
                 <h2 [style.margin]="0" fdPopoverBodyHeader>Notifications</h2>
-                @for (i of notificationRange; track i) {
+                @for (i of [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]; track i) {
                     <fd-message-strip> A new notification has arrived. {{ i }} </fd-message-strip>
                 }
             </div>
--- a/libs/docs/core/card/examples/card-loading/card-loading-example.component.html
+++ b/libs/docs/core/card/examples/card-loading/card-loading-example.component.html
@@ -6,7 +6,7 @@
     </fd-card-header>
     <fd-card-content>
         <ul fd-list [noBorder]="true">
-            @for (_ of loadingRange; track $index) {
+            @for (_ of [0, 1, 2]; track $index) {
                 <li fd-list-item>
                     <fd-skeleton type="text" width="30%" textLines="1" styles="margin: auto 0"></fd-skeleton>
                 </li>
@@ -48,7 +48,7 @@
     </fd-card-header>
     <fd-card-content>
         <ul fd-list [noBorder]="true">
-            @for (_ of loadingRange; track $index) {
+            @for (_ of [0, 1, 2]; track $index) {
                 <li fd-list-item>
                     <fd-skeleton type="text" width="30%" textLines="1" styles="margin: auto 0"></fd-skeleton>
                 </li>
--- a/libs/docs/core/list-byline/examples/list-byline-loading-example/list-byline-loading-example.component.html
+++ b/libs/docs/core/list-byline/examples/list-byline-loading-example/list-byline-loading-example.component.html
@@ -1,7 +1,7 @@
 <button fd-button (click)="loading = !loading" [style.margin-bottom.rem]="1">Toggle loading</button>
 <ul fd-list [byline]="true">
     @if (!loading) {
-        @for (_ of loadingRange; track $index) {
+        @for (_ of [0, 1, 2]; track $index) {
             <li fd-list-item>
                 <span fd-list-thumbnail><fd-icon glyph="activate"></fd-icon></span>
                 <div fd-list-content>
@@ -21,7 +21,7 @@
             </li>
         }
     } @else {
-        @for (_ of loadingRange; track $index) {
+        @for (_ of [0, 1, 2]; track $index) {
             <li fd-list-item>
                 <fd-skeleton width="100%" height="48px" [style.margin]="'auto 0'">
                     <svg:rect x="0" y="0" width="48" height="48" rx="4" />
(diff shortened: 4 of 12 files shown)


libs/core/timeline/timeline.component.spec.ts
```diff
@@ -157,6 +200,26 @@ class TimelineTestWithTrackByComponent extends TimelineTestComponent {
     }
 }
 
+@Component({
```
MariaIDineva, line 203: Hm, it's weird to see the tested component instance (this one and below) positioned at the bottom of the test file. Not that it's wrong, I just think that in most unit tests that we have it is positioned on top of the file or at least above the `describe` testing it. 🤔

MariaIDineva (approved):

libs/core/tree/tree.component.ts
```diff
@@ -185,6 +178,9 @@ export class TreeComponent<P extends FdTreeAcceptableDataSource, T extends TreeI
      */
     _initialDataLoaded = false;
 
+    /** @hidden */
+    readonly _loadingRange = range(5);
```
InnaAtanasova: is this a new member? Because if it is there's no need to be prefixed if it's not private. If it's an existing one we can keep it to not introduce breaking changes.

libs/docs/btp/tool-header/examples/auto-mode-example/auto-mode-example.component.html
```diff
@@ -17,9 +17,9 @@
                 }"
             >
                 <h2 [style.margin]="0" fdPopoverBodyHeader>Notifications</h2>
-                <fd-message-strip *fdkRepeat="10; let index = index">
-                    A new notification has arrived. {{ index }}
-                </fd-message-strip>
+                @for (index of notificationRange; track index) {
```
InnaAtanasova: should be `$index`

libs/docs/btp/tool-header/examples/basic-example/tool-header-basic-example.component.html
```diff
@@ -28,9 +28,9 @@
                 }"
             >
                 <h2 [style.margin]="0" fdPopoverBodyHeader>Notifications</h2>
-                <fd-message-strip *fdkRepeat="10; let index = index">
-                    A new notification has arrived. {{ index }}
-                </fd-message-strip>
+                @for (index of notificationRange; track index) {
```
InnaAtanasova: should be `$index`

libs/docs/core/card/examples/card-loading/card-loading-example.component.html
```diff
@@ -46,9 +48,11 @@ <h3>Card header with avatar, subtitle, counter and badge</h3>
...
     <fd-card-content>
         <ul fd-list [noBorder]="true">
-            <li fd-list-item *fdkRepeat="3">
-                <fd-skeleton type="text" width="30%" textLines="1" styles="margin: auto 0"></fd-skeleton>
-            </li>
+            @for (_ of loadingRange; track $index) {
+                <li fd-list-item>
+                    <fd-skeleton type="text" width="30%" textLines="1" styles="margin: auto 0"></fd-skeleton>
+                </li>
+            }
```
InnaAtanasova: help me understand the reason behind `loadingRange`. From what I see it's just a static number, nothing dynamic that requires range. 
Why not just write
```
@for (_ of 3; track $index) {
    <li fd-list-item>
        <fd-skeleton type="text" width="30%" textLines="1" styles="margin: auto 0"></fd-skeleton>
    </li>
}
```
no need to import range, user can see it's looping through 1-2-3

libs/docs/core/list-byline/examples/list-byline-loading-example/list-byline-loading-example.component.html
```diff
@@ -1,30 +1,34 @@
...
     @if (!loading) {
-        <li fd-list-item *fdkRepeat="3">
-            <span fd-list-thumbnail><fd-icon glyph="activate"></fd-icon></span>
-            <div fd-list-content>
-                <div fd-list-title [wrap]="true">
-                    Long Title - Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus ut laoreet lorem.
-                    Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia curae; Aenean
-                    sagittis aliquam justo et suscipit. Nam molestie, magna at elementum pulvinar, nisi enim venenatis
-                    ante, id convallis mi neque nec risus.
+        @for (_ of loadingRange; track $index) {
```
InnaAtanasova: same comment for loadingRange

libs/platform/list/list.component.ts
```diff
@@ -319,6 +318,9 @@ export class ListComponent<T>
      */
     _firstLoadingDone = false;
 
+    /** @hidden */
+    readonly _loadingRange = range(3);
```
InnaAtanasova: should not be prefixed

libs/platform/wizard-generator/components/wizard-generator/wizard-generator.component.ts
```diff
@@ -44,4 +44,7 @@ export class WizardGeneratorComponent extends BaseWizardGenerator {
      */
     @ContentChild(WizardGeneratorReviewButtonDirective, { read: TemplateRef })
     reviewButtonTemplate: TemplateRef<HTMLElement>;
+
+    /** @hidden */
+    protected readonly _loadingRange = range(3);
```
InnaAtanasova: if you keep this (which in my opinion is redundant), should not be protected and not prefixed

libs/core/tree/tree.component.ts
droshev: removed `_`

libs/docs/core/card/examples/card-loading/card-loading-example.component.html
droshev: Angular's `@for` requires an iterable collection, not a primitive. `@for (_ of 3; ...)` won't work because `3` is not iterable.

libs/platform/wizard-generator/components/wizard-generator/wizard-generator.component.ts
droshev: `loadingRange` is only used in the template, not by external consumers => remove the prefix `_`

libs/docs/btp/tool-header/examples/basic-example/tool-header-basic-example.component.html
droshev: When the loop value is used in the template like here `{{ i }}`, i named it i and track by it. When the value is unused (_), i track by $index

libs/docs/btp/tool-header/examples/auto-mode-example/auto-mode-example.component.html
droshev: When the loop value is used in the template like here `{{ i }}`, i named it i and track by it. When the value is unused (_), i track by $index

libs/docs/core/list-byline/examples/list-byline-loading-example/list-byline-loading-example.component.html
droshev: is it like the previous comment?

Issue #13856, opened 2026-02-13
Angular Signals Migration Phase 2 - CDK Infrastructure


InnaAtanasova:
**CDK Directives**

- [x] FocusableListDirective - Maria [13914](https://github.com/SAP/fundamental-ngx/pull/13914)
- [x] FocusableItemDirective - Maria [13915](https://github.com/SAP/fundamental-ngx/pull/13915)
- [x] ClickedDirective - Maria [13921](https://github.com/SAP/fundamental-ngx/pull/13921)
- [x] DisabledBehaviorDirective - Maria [13928](https://github.com/SAP/fundamental-ngx/pull/13928)
- [ ] ReadonlyDirective
- [x] TruncateDirective - Maria [13948](https://github.com/SAP/fundamental-ngx/pull/13948)
- [x] LineClampDirective - Maria [13954](https://github.com/SAP/fundamental-ngx/pull/13954)
- [x] ResizeObserverDirective - Inna [13957](https://github.com/SAP/fundamental-ngx/pull/13957)
- [x] RepeatDirective - Deno (to be deprecated and remove from usage) [13913](https://github.com/SAP/fundamental-ngx/pull/13913)
- [x] TemplateDirective - Inna [13956](https://github.com/SAP/fundamental-ngx/pull/13956)

**CDK Forms**

- [ ] CvaControl - Maria
- [ ] CvaDirective - Maria
- [ ] Form helpers

**CDK Data Source**

- [ ] BaseDataSource
- [ ] DataProvider

**CDK Toast Infrastructure** - Inna [13965](https://github.com/SAP/fundamental-ngx/pull/13965)

- [x] BaseToastService
- [x] BaseToastDurationDismissibleService
- [x] BaseToastConfig
- [x] BaseToastRef
- [x] BaseToastContainer
- [x] BaseToastAnimatedContainer
- [x] BaseToastOverlayContainer
- [x] All toast-related base classes

PR #13914, opened 2026-02-24
fix(cdk): signalify focusable list directive

MariaIDineva:
## Related Issue(s)

## Description

This PR migrates the FocusableListDirective and related components to Angular Signals. 

1. FocusableListDirective
Migrated @Input() properties to model() and input() signals
Migrated @Output() to output() signal
Migrated @ContentChildren() to contentChildren() signal
Internal state converted to signals: _focusable, _isVisible, _tabbable, _gridPosition
Replaced ngOnChanges() with three effect() blocks for reactive updates
Added computed() for derived _focusableItems state
2. Related Components Updated
IndirectFocusableListDirective: Signal-based child handling
FocusableGridDirective: WeakSet tracking + effect-based item watching
NavigationMenuComponent: Simplified with host metadata bindings
SegmentedButtonComponent: Signal setter usage
Table Components: Array conversion for setItems()
3. Test Coverage Added

⚠️ Breaking API Changes:

setItems() now accepts ReadonlyArray<FocusableItem>
Direct property access to _focusableItems requires () invocation
Removed ngOnChanges() from parent class

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,7 +1,7 @@
 <!--
 Document: Angular 21+ Development Guidelines for Fundamental NGX
-Last Updated: February 23, 2026
-Version: 3.1
+Last Updated: February 26, 2026
+Version: 3.2
 Purpose: Comprehensive guide for AI agents and developers working with Angular 21+ in NX monorepo
 -->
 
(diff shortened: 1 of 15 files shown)


libs/cdk/utils/directives/focusable-list/focusable-list.directive.ts
```diff
@@ -163,20 +159,22 @@ export class FocusableListDirective implements OnChanges, AfterViewInit, OnDestr
...
+    private readonly _injector = inject(Injector);
+
     /** @hidden */
     private readonly _document = inject(DOCUMENT);
 
     /** @hidden */
-    @HostBinding('attr.tabindex')
     get _tabindex(): number {
-        return this._tabbable ? 0 : -1;
+        return this._tabbable() ? 0 : -1;
```
InnaAtanasova: just a question, why can't we do this directly in the host binding or make it a computed signal?

InnaAtanasova (approved):

libs/cdk/utils/directives/focusable-list/focusable-list.directive.ts
MariaIDineva: We absolutely can, I will update.

PR #13915, opened 2026-02-25
fix(cdk): migrate FocusableItemDirective to signals

MariaIDineva:
## Related Issue(s)

### Main changes in focusable-item.directive.ts

**Inputs → Signal Inputs:**

- fdkFocusableItem: Converted to input() with booleanAttribute transform
- cellFocusedEventAnnouncer: Converted to input()

**Outputs → Signal Outputs:**

- cellFocused: EventEmitter → output<FocusableItemPosition>()
- focusableChildElementFocused: EventEmitter → output<void>()
- _parentFocusableItemFocused: EventEmitter → output<void>()

**New Public API:** 

- setFocusable(state: boolean): void    //  Set focusable state programmatically.
**Internal Improvements:**

- Uses linkedSignal() to set _focusable for internal writable state that syncs with input
- Replaces manual setter logic with effect() for reactive updates

### Consumer Updates

**focusable-grid.directive.ts:**

- Wrapped outputs with outputToObservable for RxJS compatibility

**table-cell.directive.ts:**

- Updated to use new API - setFocusable(), isFocusable()

**avatar-group-overflow-button.component.ts:**

- Updated to use new API - setFocusable()

**segmented-button.component.ts:**

- Updated to use new API - setFocusable()

### ⚠️  BREAKING CHANGES
- `fdkFocusableItem` can no longer be set like so: `this.fdkFocusableItem = true;`. Use new `setFocusable()` method instead.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -198,7 +198,7 @@ module.exports = pluginTs.config(
                         },
                         {
                             sourceTag: 'scope:ui5',
-                            onlyDependOnLibsWithTags: ['scope:ui5', 'scope:cdk']
+                            onlyDependOnLibsWithTags: ['scope:ui5', 'scope:cdk', 'scope:fd']
                         }
                     ]
                 }
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.61.0-rc.10",
+    "version": "0.61.0-rc.12",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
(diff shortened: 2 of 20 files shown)


libs/cdk/utils/directives/focusable-item/focusable-item.directive.ts
```diff
@@ -82,24 +70,31 @@ export class FocusableItemDirective implements FocusableItem, HasElementRef {
 
     /** @hidden */
     protected readonly _destroyRef = inject(DestroyRef);
+
     /** @hidden */
     protected readonly _zone = inject(NgZone);
```
droshev: zone is still used

MariaIDineva: Yeah, I kept it because Claude suggested that it provides an important performance optimization (lines 124-136) and if we remove it, we might hit some problems, especially when the directive is used on a large number of items, for example in tables. And after your comment I have been playing with keeping vs removing zone to see what the impact is but I am unable to hit a scenario where the problem is reproducible, so I might just remove it.

Issue #13922, opened 2026-02-26
fdp-table always shows 'Please wait' tooltip


rkloe:
### Describe your proposal

The platform table always displays a “Please wait” tooltip, even when it isn’t loading any data. This is likely because the busy indicator is the parent of the data table and has a title attribute set to “Please wait.” Browsers display the title attribute as a tooltip.

### Which versions of Angular and Fundamental Library for Angular are affected?

Angular: >20, Fundamental Library: >0.58

### Affected Component

fdp-table

### Steps to Reproduce (for bugs)

1. Go to https://sap.github.io/fundamental-ngx/#/platform/table/basic.
2. Hover over any data cell in the first example and wait.
3. You should see the "Please wait" tooltip.

### Priority

Medium

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

PR #13925, opened 2026-02-26
fix(core,platform): set title in busy indicator only when loading

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13922

## Description

A bug introduced in https://github.com/SAP/fundamental-ngx/pull/13763 where we add default values for `title` and `aria-valuetext` attributes.

When `fd-busy-indicator` is not in loading mode, no attributes`title` and `aria-valuetext` are added.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -198,7 +198,7 @@ module.exports = pluginTs.config(
                         },
                         {
                             sourceTag: 'scope:ui5',
-                            onlyDependOnLibsWithTags: ['scope:ui5', 'scope:cdk']
+                            onlyDependOnLibsWithTags: ['scope:ui5', 'scope:cdk', 'scope:fd']
                         }
                     ]
                 }
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.61.0-rc.10",
+    "version": "0.61.0-rc.11",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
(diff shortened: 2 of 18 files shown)


libs/platform/table/table.component.html
```diff
@@ -8,6 +8,7 @@
 <fd-busy-indicator
     [loading]="(_tableService.tableLoading$ | async) === true && _dataSourceDirective._firstLoadingDone"
     [block]="true"
+    [title]="null"
```
mikerodonnell89: should this be the default for the busy indicator component so it doesn't need to be set here? Assuming that was the case before https://github.com/SAP/fundamental-ngx/pull/13763 was merged

mikerodonnell89 (changes requested):

libs/platform/table/table.component.html
MariaIDineva: Hm, I don't really know what should be the default behaviour. That was one of the functional changes implemented in https://github.com/SAP/fundamental-ngx/pull/13763 - to make the busy indicator always have a title - whether the default one `Please wait` or a custom one. I have inferred that we want that because as part of migrating the component to signals, I also tried to adopt the style changes from this PR - https://github.com/SAP/fundamental-styles/pull/6027/changes. Also, this is what I see as behaviour in UI5 - https://sdk.openui5.org/#/entity/sap.m.BusyIndicator/sample/sap.m.sample.BusyIndicator

But now that you mentioned that, I took a closer look at that PR and I see that it also introduces `attr.aria-valuetext` with a default valuetext which we might also want to have an option to unset, especially for the case with platform table.

droshev (approved):

PR #13928, opened 2026-02-27
fix(cdk): migrate DisabledBehaviorDirective to signals

MariaIDineva:
## Related Issue(s)

### Migrates DisabledBehaviorDirective to signals

1. Replaced @Input() decorators with signal-based input() functions:

- fdkDisabled → fdkDisabledInput with { alias: 'fdkDisabled', transform: booleanAttribute }
- addDisabledClass → input(true)
- disabledClass → input('is-disabled')

2. Replaced coerceBooleanProperty() with booleanAttribute transform:

3. Replaced RxJS-based reactivity with effect():

- Removed BehaviorSubject, filter(), tap(), takeUntilDestroyed()
- Removed ngAfterViewInit lifecycle hook
- Added effect() in constructor to react to signal changes
- Automatic cleanup (no manual subscriptions needed)

4. Updated signal reads to use function call syntax:

- this.disabledClass() instead of this.disabledClass
- this.addDisabledClass() instead of this.addDisabledClass

5. Maintained backward compatibility:

- Kept **fdkDisabled** getter for interface compliance
- All existing functionality preserved

### DIFF
--- a/libs/cdk/utils/directives/disabled/disabled-behavior.directive.ts
+++ b/libs/cdk/utils/directives/disabled/disabled-behavior.directive.ts
@@ -29,7 +29,7 @@ export class DisabledBehaviorDirective
     /**
      * Whether to add `disabledClass` class to the element.
      */
-    readonly addDisabledClass = input(true);
+    readonly addDisabledClass = input(true, { transform: booleanAttribute });
 
     /**
      * Disabled css class to apply to the element.


libs/cdk/utils/directives/disabled/disabled-behavior.directive.ts
```diff
@@ -22,36 +19,32 @@ import { setDisabledState } from './set-disabled-state';
...
+     * Whether the element is disabled.
+     */
+    readonly fdkDisabledInput = input(false, { alias: 'fdkDisabled', transform: booleanAttribute });
 
     /**
      * Whether to add `disabledClass` class to the element.
      */
-    @Input()
-    addDisabledClass = true;
+    readonly addDisabledClass = input(true);
```
InnaAtanasova: `transform: booleanAttribute` is missing

PR #13932, opened 2026-03-02
fix(core): fix avatar image loading

### LINKED ISSUE #13931: [BUG] Avatars are re-fetched infinitely
Very high prio bug, 
In my component i have only put
<fd-avatar image="/images/jira.svg" size="s" [contain]="true" [transparent]="true" />

This image is re-fetched infinitely (happens only when VPN is turned on)

Image

since version 0.59.1

droshev:
fix avatar image loading

fixes #13931

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.61.0-rc.8",
+    "version": "0.61.0-rc.9",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.61.0-rc.8",
+    "version": "0.61.0-rc.9",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/cdk/package.json
+++ b/libs/cdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cdk",
-    "version": "0.61.0-rc.8",
+    "version": "0.61.0-rc.9",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - CDK",
     "license": "Apache-2.0",
--- a/libs/core/avatar/avatar.component.scss
+++ b/libs/core/avatar/avatar.component.scss
@@ -5,3 +5,57 @@
 .fd-avatar.fd-avatar--sm {
     --fdAvatarFontStretch: normal;
 }
+
+// Temporary until we adopt the latest version of fundamental-styles.
+:not(.fd-avatar-group--vertical) {
+    // Consolidated overflow styling for all avatar sizes
+    > [class*='fd-avatar--'].fd-avatar--overflow {
+        width: fit-content;
+        max-width: fit-content;
+        padding-inline: 0.5rem;
+    }
+
+    // Circle border radius and focus styles per size
+    > .fd-avatar--xs.fd-avatar--overflow.fd-avatar--circle {
+        --fdAvatarBorderRadius: 1rem;
+
+        &:focus::after {
+            border-radius: 1.25rem;
+        }
+    }
+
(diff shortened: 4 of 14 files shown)


libs/core/avatar/avatar.component.ts
InnaAtanasova: This can be removed when you get rid of the signal lines 274-277

InnaAtanasova: This signal can be replaced by `'[style.background-image]': 'bgImage()'` binding in the host. It now unnecessarily triggers change detection.

libs/core/avatar/avatar.component.ts
```diff
@@ -209,6 +209,15 @@ export class AvatarComponent {
     /** @hidden Stored random color accent to prevent regeneration */
     private readonly _randomColorAccent = signal<ColorAccent>(getRandomColorAccent());
 
+    /** @hidden Track the currently loading/loaded image URL to prevent duplicate requests */
+    private _currentImageUrl: string | null = null;
+
+    /** @hidden Track if an image load is in progress to prevent duplicate requests */
+    private _isLoadingImage = false;
```
InnaAtanasova: Please double check but I think this is redundant and the whole logic can exist without it. The `_isLoadingImage` flag is redundant since the effect already prevents re-calling `_verifyImageUrl` for the same URL.
 
Why it's safe:
 
The effect's `_currentImageUrl !== imageValue` check already prevents duplicate loads for primary images
The `_loadedImageCache` handles re-use of previously loaded images
Backup images are only loaded once per error, so no duplicate risk

libs/core/avatar/avatar.component.ts
```diff
@@ -224,10 +233,17 @@ export class AvatarComponent {
         effect(() => {
             const imageValue = this.image();
             if (imageValue) {
-                this._imageLoaded.set(false);
-                this._shouldShowDefaultIcon.set(false); // Reset default icon flag
-                this._verifyImageUrl(imageValue, this._onErrorCallback.bind(this));
+                // Only reset loaded state if the URL actually changed
+                if (this._currentImageUrl !== imageValue) {
+                    this._imageLoaded.set(false);
```
InnaAtanasova, line 232: this can be removed if you get rid of `_imageLoaded`

libs/core/avatar/avatar.component.ts
```diff
@@ -308,20 +324,38 @@ export class AvatarComponent {
...
     /** @hidden */
     private _verifyImageUrl(srcValue: string, onErrorCallback: () => void): void {
-        const newBgImage = `url(${srcValue})`;
+        // Check if this URL was previously loaded successfully (component-level cache)
+        if (this._loadedImageCache.has(srcValue)) {
+            // Image was loaded before, reuse it
+            this._setBgImage(srcValue);
+            this._imageLoaded.set(true);
+            this._currentImageUrl = srcValue;
+            this._isLoadingImage = false;
```
InnaAtanasova: this can be removed if you get rid of `_imageLoaded`

libs/core/avatar/avatar.component.ts
```diff
@@ -308,20 +324,38 @@ export class AvatarComponent {
...
+            return;
+        }
 
-        // Don't load the same image if it's already loaded successfully
-        if (this.bgImage() === newBgImage && this._imageLoaded()) {
+        // Prevent duplicate requests if this URL is already being loaded
+        if (this._isLoadingImage && this._currentImageUrl === srcValue) {
             return;
         }
```
InnaAtanasova: this can be removed if you get rid of `_imageLoaded`

libs/core/avatar/avatar.component.ts
```diff
@@ -308,20 +324,38 @@ export class AvatarComponent {
...
 
+        // Track the current image URL and loading state
+        this._currentImageUrl = srcValue;
```
InnaAtanasova, line 324: this can be removed if you get rid of `_imageLoaded`

libs/core/avatar/avatar.component.ts
```diff
@@ -308,20 +324,38 @@ export class AvatarComponent {
...
+        this._isLoadingImage = true;
+
         const img = new Image();
         img.onload = () => {
             // Only set background if image loads successfully
             this._setBgImage(srcValue);
             this._imageLoaded.set(true);
+            this._isLoadingImage = false;
```
InnaAtanasova: this can be removed if you get rid of `_imageLoaded`

libs/core/avatar/avatar.component.ts
droshev: `_imageLoaded` prevents the label effect from generating new abbreviations when an
  image is already successfully loaded.

Issue #13924, opened 2026-02-26
User Menu design updates


droshev:
### Describe your proposal

User Menu design has been amended with:
- submenu opens on hover to aligned with the classic UI
- list items to accept 2 lines
- allow message strip text under the User details

### Proposed Solution

User Menu amendment

### Priority

Medium

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

PR #13936, opened 2026-03-04
fix(core): update Menu and User Menu to latest design

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/13924

## Description
Menu component:
- added new `fd-menu-content` directive to wrap the title and subtitle
- added new `fd-menu-subtitle` directive + tests
- updated `fd-menu-title` with `truncate` input + updated tests
- brought temporary CSS from fund-styles (separate scss files for easier maintenance) 
- updated doc examples

User Menu component: 
- submenu now opens on hover
- added subtitle to the items, icon after the text
- for Mobile mode, when the additional view with the subitems is closed, the focus is restored on the trigger item 

## Screenshots
**Menu:**
Screenshot 2026-03-04 at 2 51 36 PM
Screenshot 2026-03-04 at 2 51 49 PM

**User Menu:**
Screenshot 2026-03-04 at 2 12 12 PM


InnaAtanasova:
> Should the sub-menu open to the other side in RTL mode? It always opens to the left regardless of RTL
> 
> Image Image

hey Mike, yes it should always open on the left as User Menu is situated always on the right, the last element in the ShellBar, so the submenu is always on the left.

mikerodonnell89:
> hey Mike, yes it should always open on the left as User Menu is situated always on the right, the last element in the ShellBar, so the submenu is always on the left.

Right so when it's in RTL mode, the user menu will be on the left side of the screen, so shouldn't the submenu open to the right?

InnaAtanasova:
> > hey Mike, yes it should always open on the left as User Menu is situated always on the right, the last element in the ShellBar, so the submenu is always on the left.
> 
> Right so when it's in RTL mode, the user menu will be on the left side of the screen, so shouldn't the submenu open to the right?

The popover's `getCdkPlacement` handles the flipping, there's nothing needed to be done on User Menu.
The menu item is using Popover to show the submenu. There I configured the position of the popover body to be `left-start`. The Popover is now handling the flipping. If I handle it on the User menu side it will be double flipping and revert the effect of the `getCdkPlacement`.

Issue #13697, opened 2026-01-01
[BUG] fd-table-cell and fd-table-row cause performance issues


LiorCapsiSAP:
### Describe your proposal

when using fd-table with fd-table-row and fd-table-cell with a large number of rows, a significant performance hit is observed in comparison to the exact same table without those directives.

### Which versions of Angular and Fundamental Library for Angular are affected?

angular v19 and fd v0.56

### Steps to Reproduce (for bugs)

- open fd-table's stackblitz
- increase number of rows to 200 or more
- use dev tools to observe performance
- remove fd-table-row and fd-table-cell directives
- profile with devtools performance again
- observe differences

<img width="1931" height="881" alt="Image" src="https://github.com/user-attachments/assets/2f238750-6ad8-4915-bb91-bc2fa151c2b7" />
<img width="2205" height="634" alt="Image" src="https://github.com/user-attachments/assets/2020b5e4-1e96-437f-bf86-364be0510d86" />

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

mikerodonnell89:
Of course adding any Angular directive to any native element will cause some performance slowdown, especially on something like a table, where 5 columns * 200 rows = 1000 elements with Angular directives. We do work to optimize our components as best as possible, but if you are still seeing unacceptable performance in your application, I'd recommend implementing some sort of pagination feature for your table to reduce the number of rendered elements. Our platform table also comes with a variety of features for handling performance such as virtual scrolling.

PR #13946, opened 2026-03-05
fix(cdk): focusable grid/list/item performance improvements

mikerodonnell89:
part of #13697


libs/cdk/utils/directives/focusable-list/focusable-list.directive.ts
```diff
@@ -365,13 +376,19 @@ export class FocusableListDirective implements AfterViewInit, OnDestroy {
...
+                const focusableItems = this._focusableItems();
+                const focusableItemsLength = focusableItems.length;
+                for (let i = 0; i < focusableItemsLength; i++) {
+                    focusableItems[i].setTabbable(focusableItems[i] === directiveItem);
+                }
                 this._keyManager?.setActiveItem(index);
-            })
-        );
+            });
+            keydownObservables[index] = item.keydown;
```
MariaIDineva: `keydownObservables[index] = item.keydown;` is also present on line 365. Is that assignment duplicated on purpose?

PR #13948, opened 2026-03-06
fix(cdk): migrate TruncateDirective to signals

MariaIDineva:
## Related Issue(s)

## Description

- Migartes TruncateDirective to signals
- Replaces module imports in examples with imports of directives for the following directives:  `DisabledBehaviorDirective`, `FocusableItemDirective`, `FocusableListDirective`, `TruncateDirective`
- Adds missing example for `FocusableItemDirective`.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.61.2-rc.3",
+    "version": "0.61.2-rc.5",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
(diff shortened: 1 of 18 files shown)


libs/cdk/utils/directives/truncate/truncate.directive.ts
```diff
@@ -33,15 +39,7 @@ export class TruncateDirective implements OnChanges, AfterViewInit {
...
-    private takeDefaultStyleOnce = true;
-
-    /** @hidden
-     * truncation style for truncating element
-     */
-    private _truncationStyle: string;
-
-    /** @hidden */
-    constructor(private readonly _elementRef: ElementRef) {}
+    private _defaultStyleCaptured = false;
```
InnaAtanasova, line 42: why are these two not signals? Would moving to zoneless affect their change detection?

InnaAtanasova: reflecting on this, they should not be signals. They are private members that are not used in the template and host binding.

PR #13954, opened 2026-03-09
fix(cdk): migrate LineClampDirective and LineClampTargetDirective to signals

MariaIDineva:
## Related Issue(s)

## Description

1. Migrates `LineClampDirective` and `LineClampTargetDirective` from decorator-based inputs/outputs to Angular signals
- Replaces @Input() decorators with input() function and @Output() with output()
- Replaces lifecycle hooks (OnChanges, AfterViewInit) with effect() and afterNextRender()
- Removes standalone: true (now default in Angular 21+)
- Renames private methods to follow naming conventions (native() → _applyNative(), getLineHeight() → _getLineHeight())
- Fixес snake_case variable names to camelCase (padding_top → paddingTop, etc.)
 
 2. Adds a new line-clamp example page to the documentation
 3. Expands test coverage from ~40 lines to ~330 lines:
   -  Added tests for native CSS line-clamp support
   -  Added tests for line count changes and text updates
   -  Added tests for viewport resize handling with debounce
   -  Added tests for reset functionality
   -  Added tests for standalone target directive usage
   -  Added tests for disabled clamp state

**NB!**
- In `_checkLineCount` method of nstead of `let height = parseInt(style.getPropertyValue('height'), 10);` of `LineClampDirective` we now use `let height = this.rootElement.offsetHeight;` because `getComputedStyle('height')` can return 'auto' for elements that don't have an explicit height set, while `offsetHeight` always returns a concrete pixel value. Thus we avoid a case when `lineCountUpdate` can emit `NaN`.
- Adds [queueMicrotask](https://github.com/SAP/fundamental-ngx/pull/13954/changes#diff-2d5bd573a31f87083049f270cbb749b71df26fc19cd6ada9ff1673c53d885265R553) to `toggle` in `_refreshTriggerListeners` in PopoverService to avoid NG0600 error:
Screenshot 2026-03-12 at 15 25 31
The error happens when you click the - or the + sign on the step input in the new example upon reaching the min and max values.

### DIFF
--- a/libs/cdk/utils/directives/line-clamp/line-clamp.directive.spec.ts
+++ b/libs/cdk/utils/directives/line-clamp/line-clamp.directive.spec.ts
@@ -20,7 +20,7 @@ const LONG_TEXT =
             <span
                 fdkLineClampTarget
                 [fdLineClampTargetText]="text"
-                (update)="lineClampDirective?.refreshTarget($event)"
+                (update)="lineClampDirective()?.refreshTarget($event)"
             >
                 {{ text }}
             </span>


droshev:
is it worth adding an example?

MariaIDineva:
> is it worth adding an example?

I was thinkning the same thing while I was working on this. We don't expose it directly in our documentation under `cdk`, we only use it internally in the text component [Line Clamping](https://fundamental-ngx.netlify.app/#/core/text#max-lines) example. I could add an example if we are willing to expose it.

libs/docs/cdk/line-clamp/examples/default/line-clamp-default-example.component.ts
```diff
@@ -0,0 +1,54 @@
...
+    ],
+    styles: `
+        .fd-text {
+            font-size: var(--sapFontSize);
+            line-height: normal;
+            color: var(--sapTextColor);
+            font-family: var(--sapFontFamily);
+            font-weight: normal;
+            box-sizing: border-box;
+        }
```
InnaAtanasova, line 28: Added the missing CSS for fd-text

libs/cdk/utils/directives/line-clamp/line-clamp.directive.spec.ts
```diff
@@ -1,42 +1,318 @@
...
+        </div>
+    `,
     imports: [LineClampDirective, LineClampTargetDirective]
 })
 class TestComponent {
-    @Input()
-    text: string;
-    @Input()
-    rows: number;
+    @ViewChild(LineClampDirective) lineClampDirective: LineClampDirective;
```
droshev: can you use signal viewChild() instead?

libs/cdk/utils/directives/line-clamp/line-clamp.directive.ts
```diff
@@ -146,28 +167,30 @@ export class LineClampDirective implements OnChanges, AfterViewInit {
...
     }
 
-    /** @hidden */
-    refreshClamp(): void {
-        if (this.fdLineclampState && this._lineCount) {
-            this.native();
+    /** @hidden
+     * Refresh clamp styles based on state and line count
+     */
+    refreshClamp(state: boolean, lineCount: number): void {
```
droshev, line 177: i guess we need to mark this as a breaking change or make them optional?

droshev:
Can you add LineClampTargetDirective to API files (libs/docs/cdk/api-files.ts). It's a public export that's missing from the docs entry.

libs/docs/cdk/line-clamp/examples/default/line-clamp-default-example.component.html
```diff
@@ -0,0 +1,24 @@
+<div style="margin-top: 1rem">
+    <p>
+        Actual line count (without clamping): <strong>{{ actualLineCount() }}</strong>
+    </p>
+    <button fd-button (click)="toggleClamp()">{{ lineClampEnabled() ? 'Disable' : 'Enable' }} Line Clamp</button>
+</div>
+
+<div style="margin: 1rem 0">
```
droshev: can we use common-css classes instead?

libs/core/popover/popover-service/popover.service.ts
```diff
@@ -548,7 +548,9 @@ export class PopoverService {
                         }
                         const closeAction = !!trigger.closeAction;
                         const openAction = !!trigger.openAction;
-                        this.toggle(openAction, closeAction);
+                        // Defer toggle to avoid NG0600 error when triggered by focusout events
+                        // that occur during Angular's render cycle (e.g., when a button becomes disabled)
+                        queueMicrotask(() => this.toggle(openAction, closeAction));
```
droshev, line 559: Could we check whether the `queueMicrotask` in `popover.service.ts` needs to apply to all trigger types, or just the `focusout` case that causes `NG0600`? Also worth running the popover, select, and combobox specs to make sure the timing change doesn't affect them.

libs/docs/cdk/line-clamp/line-clamp-header/line-clamp-header.component.html
```diff
@@ -0,0 +1,22 @@
...
+            The directive uses CSS <code>-webkit-line-clamp</code> property when supported by the browser, with a
+            JavaScript fallback for older browsers.
+        </p>
+        <p>
+            Use <code>fdkLineClamp</code> on a container element and <code>fdLineClampTarget</code> on the text element
+            inside it.
+        </p>
+    </description>
+
+    <import module="LineClampModule" subPackage="utils"></import>
```
droshev: can we use the component?

libs/docs/cdk/line-clamp/line-clamp-header/line-clamp-header.component.ts
```diff
@@ -0,0 +1,18 @@
...
+} from '@fundamental-ngx/docs/shared';
+
+@Component({
+    selector: 'app-line-clamp-header',
+    templateUrl: './line-clamp-header.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    imports: [DocPageComponent, HeaderComponent, DescriptionComponent, ImportComponent, HeaderTabsComponent]
+})
+export class LineClampHeaderComponent {
+    falseSignal = signal(false);
```
droshev: what is the purpose of that?

libs/docs/cdk/api-files.ts
```diff
@@ -20,5 +20,6 @@ export const API_FILES = {
     focusableList: ['FocusableListDirective', 'FocusableItemDirective'],
     dnd: ['DndListDirective', 'DndItemDirective', 'DndKeyboardGroupDirective', 'DndKeyboardItemDirective'],
     initialFocus: ['InitialFocusDirective'],
+    lineClamp: ['LineClampDirective'],
```
droshev: we can add also LineClampTargetDirective.

libs/cdk/utils/directives/line-clamp/line-clamp.directive.ts
MariaIDineva: Yeah, you are right. Although... maybe that method should be private overall, not sure why it is not.

libs/docs/cdk/line-clamp/line-clamp-header/line-clamp-header.component.ts
MariaIDineva: None, I'll remove it, thanks!

PR #13959, opened 2026-03-10
fix(platform): update table pagination

### LINKED ISSUE #13840: [BUG] TableComponent Infinite Scroll Fails on Large Screens (Spy element initially visible)
### Describe your proposal

**Issue**: Infinite Scroll Breaks When Switching Tabs (fdp-table)
**Description** When using fdp-table inside tab-based navigation, scrolling breaks if the user switches from a tab containing fewer rows to a tab containing more rows.

IMPORTANT NOTE:
This issue is specific to large screens or high-resolution displays where the initial set of rows does not fill the entire height of the container.

**Root Cause** Once the table reaches the maximum number of rendered rows on the smaller dataset, switching to a tab with a larger dataset does not reset the internal pagination state. Consequently:

1. The table incorrectly determines that the maximum row count has already been reached.
2. The pageNumber stops updating.
3. Scrolling is disabled, preventing additional data from rendering.

**Actual Behavior**

- Trigger: Switching from a tab with fewer rows to a tab with more rows (where the first tab reached its "maximum rendered" state).

- pageNumber stops updating.

- The Intersection Observer (onSpyIntersect) is not triggered.

- The scroll container is removed from the DOM.

Impact: The user cannot scroll or load the rest of the dataset.

**Expected Behavior**

- The table pagination state should reset or update correctly upon tab change.

- pageNumber should continue updating to accommodate the larger dataset.

- Vertical scrolling should remain enabled.

- The Intersection Observer should trigger correctly to load subsequent data.

Angular: >=20; Fundamental Library for Angular: >0.49.2

### Affected Component

TableComponent

1. Create a page with tabs.
2. Tab A contains an fdp-table with a smaller dataset.
3. Tab B contains an fdp-table with a larger dataset.
4. Open Tab A and scroll until the maximum number of rows is reached.
5. Switch to Tab B.
6. Observe that scroll is missing and additional data is not rendered.

### StackBlitz example(s)

https://stackblitz.com/edit/5atnwhfc-uhmbbuwy

### Priority

None

### Declaration

droshev:
closes #13840

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.61.2-rc.7",
+    "version": "0.61.2-rc.8",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.61.2-rc.7",
+    "version": "0.61.2-rc.8",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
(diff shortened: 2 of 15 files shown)


droshev:
> Let's remove the example docs before merging the changes to `table.component.ts`

done, pushed unit tests too

PR #13960, opened 2026-03-10
fix(platform): setTableState should update what the user sees as well

### LINKED ISSUE #12926: Info Bar Not Displayed When Setting Filters Programmatically Using setTableState
When we use the setTableState method on the platform table to set filters
([source](https://github.com/SAP/fundamental-ngx/blob/6d19912b86895c1ced98905698dc22ec9133c71f/libs/platform/table/table.component.ts#L1056)),
the info bar with information about the applied filters is not displayed (unlike when filters are set manually by clicking).
The filters are applied to the table, but the info bar does not appear.

Stackblitz example:
https://stackblitz.com/edit/7clpjevf-m2jeq55c?file=src%2Fapp%2Fplatform-table-initial-state-example.component.ts

Image

mikerodonnell89:
fixes #12926

Previously, the publicly exposed function `setTableState` was saving the passed state to the table service, but not updating everything in the displayed table. As the function was public and documented to `Set current state/settings of the Table`, I've updated the function to call `setPreset`, which does all of that. However, I'm wondering if we should just deprecate `setTableState` with a message to simply call `setPreset`


InnaAtanasova:
> I, personally, find the name `setTableState` better than `setPreset` for a public API and for the purpose it serves. I'd rather deprecate `setPreset` and when we remove it, copy the logic from there into `setTableState`. But that's just an opinion, we can go either way.

I am +1 for this

libs/platform/table/table.component.ts
```diff
@@ -1071,8 +1071,8 @@ export class TableComponent<T = any>
 
     /** Set current state/settings of the Table. */
     setTableState(state: TableState): void {
-        this._tableService.setTableState(state);
-        this._cdr.markForCheck();
+        this.setPreset(state);
```
droshev: Few points Claude traced:

- What happens to `getCurrentPreset()` after `setTableState()` is called now that it goes through `setPreset()`? Worth tracing through `_currentPreset` to check.
- The other public methods (`sort()`, `filter()`, `group()`, etc.) all use `markForCheck()` — could you check if switching to `detectChanges()` here might cause issues when `setTableState()` is called during a lifecycle hook?
- Could you compare how many `stateChange$` / `needFetch$` emissions happen with the old `setTableState()` (single atomic `tableState$.next()`) vs going through `setPreset()` which calls `setSort`, `setFilters`, `setGroups`, etc. individually?

Its comments:
```
     1. [libs/platform/table/table.component.ts:1073] Side effect on _currentPreset may break getCurrentPreset().
     setPreset() sets this._currentPreset = data (line 1384), which is then used by getCurrentPreset() (line 1398-1411) to return a preset shaped
     like the original input. When setTableState() now calls setPreset(), it will set _currentPreset to the full TableState object. Previously,
     setTableState() had no effect on _currentPreset. This changes the return value of getCurrentPreset() for any code that calls setTableState()
     followed by getCurrentPreset(). This is an unintended behavioral change that could break consumers relying on preset tracking.
     2. [libs/platform/table/table.component.ts:1074] detectChanges() replacing markForCheck() can cause ExpressionChangedAfterItHasBeenCheckedError.
     The original code used markForCheck(), which schedules a check on the next CD cycle. The new code uses detectChanges(), which triggers
     synchronous change detection. If setTableState() is called during a lifecycle hook or within an already-running change detection cycle (e.g.,
     from an ngOnInit or a parent's template expression), detectChanges() will throw. The other public methods in the same file (sort(), addSort(),
     filter(), addFilter(), group(), etc.) all use markForCheck(). Using detectChanges() here would be inconsistent with the rest of the API surface
     and more error-prone. Note that setPreset() itself does NOT call detectChanges() or markForCheck() -- the individual service methods it calls
     (setSort, setFilters, etc.) trigger their own change detection via the markForCheck$ and detectChanges$ subjects subscribed in the component.
     Adding an explicit detectChanges() after setPreset() introduces a second synchronous CD pass on top of the debounced one already scheduled
     internally.
     3. [libs/platform/table/table.component.ts:1073] setPreset() merges with getDefaultState(), which changes semantics.
     Looking at setPreset() line 1385: const newState = Object.assign({}, this._tableService.getDefaultState(), data). This means any properties NOT
     present in the passed TableState would be filled from the default state. The original setTableState() passed the state directly to
     _tableService.setTableState() without merging defaults. While TableState is a full interface (all required fields), this merge introduces a
     subtle difference: if a caller somehow passes an object with properties explicitly set to undefined, setPreset will overwrite them with
     defaults, whereas the old code would preserve them as-is. More importantly, setPreset individually calls setSort, setFilters, setGroups,
     setColumns, search, freezeTo, and setCurrentPage, each of which emits its own stateChange$ event and needFetch$ signal. The original code set
     the entire state atomically via a single tableState$.next(). This means one call to setTableState() could now trigger up to 7 separate fetch
     requests and 7 separate state change events, which is a significant behavioral change that could cause performance issues and unexpected side
     effects in consumers listening to stateChange$.
```

droshev:
@mikerodonnell89 maybe it is worth having unit test covering the changed `setTableState` behavior

libs/platform/table/table.component.ts
mikerodonnell89: > Few points Claude traced:
> 
> * What happens to `getCurrentPreset()` after `setTableState()` is called now that it goes through `setPreset()`? Worth tracing through `_currentPreset` to check.

Looks like no changes to me

> * The other public methods (`sort()`, `filter()`, `group()`, etc.) all use `markForCheck()` — could you check if switching to `detectChanges()` here might cause issues when `setTableState()` is called during a lifecycle hook?

Switched back to markForCheck

> * Could you compare how many `stateChange$` / `needFetch$` emissions happen with the old `setTableState()` (single atomic `tableState$.next()`) vs going through `setPreset()` which calls `setSort`, `setFilters`, `setGroups`, etc. individually?
> 

Tested before and after these changes, and those are not emitted any more or less than pre-change

Issue #13962, opened 2026-03-11
Tab Icon Bar does not detect overflow


adriapulido-boyum:
### Describe your proposal

When I have to modify the structure of the icon tab bar in a tiny window, it does not group the items in the 'more' button, overflowing the items. For example, I want to change the language instantly, but it does not work well.

However, when you resize the window, it shows well again

BEFORE:
<img width="823" height="102" alt="Image" src="https://github.com/user-attachments/assets/57017fcf-37ed-4b93-96ab-1754a938d5bd" />

AFTER:
<img width="804" height="108" alt="Image" src="https://github.com/user-attachments/assets/d699233c-0b84-4a4b-b037-4101e176b0d3" />

CODE FRAGMENT
<fdp-icon-tab-bar
  iconTabSize="responsive-paddings"
  [tabsConfig]="tabsConfig()"
  [multiClick]="true"
  (iconTabSelected)="onTabSelected($event)">
</fdp-icon-tab-bar>

PD: I think that the button More has no way to be translated in other languages

### Which versions of Angular and Fundamental Library for Angular are affected?

0.60.1

### Affected Component

fdp-icon-tab-bar

### Steps to Reproduce (for bugs)

Set a few items to load in the icon tab bar.
Resize the window to show a few items.
Change the structure or set to one item to active (subitem or main item)
You won't see the more button.

### Priority

Medium

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

PR #13964, opened 2026-03-11
fix(cdk, platform): platform icon tab bar does not detect overflow

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13962

### Problem

The `OverflowListDirective` only recalculated overflow on `ResizeObserver` and `window.resize` events. When the list content changed dynamically (items added/removed, or text labels updated — e.g. a language switch), the directive did not detect that items now overflow the container. Users would see items spilling out of bounds until they manually resized the window.

### Root Cause

1. **No reaction to content changes** — the directive never subscribed to `ContentChildren.changes` (item count changes) and had no mechanism to detect text content mutations (label updates).

### Additional problems
1. **Flicker on recalculation** — the original `calculateOverflow()` used `setTimeout(0)` (a macrotask), which runs *after* the browser paints, causing a visible flash of overflowed items before they move into the "more" popover.
2. **Memory leak** — the `fromEvent(window, 'resize')` subscription was never cleaned up on destroy.

### Changes

**`overflow-list.directive.ts`**

- Added `MutationObserver` watching `childList` + `characterData` (with `subtree: true`) to detect both structural changes (items added/removed via `@for`) and text content changes (label updates, i18n switches). Attributes are intentionally excluded to avoid infinite loops from our own measurement toggles (`hidden`, `style.display`).
- Replaced `setTimeout(0)` with `Promise.resolve().then()` — a **microtask** that runs after DOM mutations but *before* the browser paints, eliminating the visible flicker.
- Added coalescing via `_recalcScheduled` flag to prevent redundant recalculations when multiple triggers fire in the same task.
- Added `takeUntilDestroyed` to the `window.resize` subscription and `DestroyRef.onDestroy` cleanup for both `ResizeObserver` and `MutationObserver`.
- Removed dead code in `calculateOverflow()` (local `totalWidth`/`overflowCount` variables were computed but never used).

**`overflow-list.directive.spec.ts`**

- Fixed silently-passing tests: assertions inside `fixture.whenStable().then()` within `fakeAsync` were swallowed by the Promise chain and never actually evaluated. Moved assertions to run synchronously after `flush()`.
- Replaced fragile `HTMLElement.prototype` property mocks (`offsetWidth`, `offsetLeft`, `clientWidth`) — which didn't produce meaningful layout in JSDOM — with `jest.spyOn(overflowListDir, 'getAmountOfExtraItems')` to test the event wiring without depending on JSDOM layout.
- Fixed `Array(100).fill(0)` → `Array.from({ length: 10 }, (_, i) => i)` to eliminate Angular NG0955 duplicate track-key warnings.
- Added unit tests to cover the new logic around `MutationObserver`.

### Examples
- Added a separate example on the `platform/icon-tab-bar` page called `Dynamic Overflow` to test and depict the issue and its fix. 
- The example allows:
  - adding new tabs dynamically so that `More` appears when needed
  - switching to longer names of the tabs so that visible tabs are recalculated dynamically and the overflowing ones are inserted into the overflow.

## Screenshots

**Before:**

https://github.com/user-attachments/assets/883fd6ef-e7b3-4ca7-9fc0-00945076c571

**After:**

https://github.com/user-attachments/assets/c5a94b5f-e51d-4e85-afa3-e9b6b188e170

### DIFF
--- a/libs/docs/platform/icon-tab-bar/examples/platform-icon-tab-bar-dynamic-overflow-example/platform-icon-tab-bar-dynamic-overflow-example.component.html
+++ b/libs/docs/platform/icon-tab-bar/examples/platform-icon-tab-bar-dynamic-overflow-example/platform-icon-tab-bar-dynamic-overflow-example.component.html
@@ -1,4 +1,4 @@
-<div style="margin-bottom: 8px; display: flex; gap: 8px">
+<div class="sap-flex sap-flex--gap-small fd-margin-bottom--sm">
     <button fd-button fdType="emphasized" (click)="toggleLabels()">
         {{ useLongLabels ? 'Short Labels' : 'Long Labels' }}
     </button>
--- a/libs/docs/platform/icon-tab-bar/examples/platform-icon-tab-bar-dynamic-overflow-example/platform-icon-tab-bar-dynamic-overflow-example.component.ts
+++ b/libs/docs/platform/icon-tab-bar/examples/platform-icon-tab-bar-dynamic-overflow-example/platform-icon-tab-bar-dynamic-overflow-example.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit } from '@angular/core';
+import { Component } from '@angular/core';
 import { ButtonComponent } from '@fundamental-ngx/core/button';
 import { IconTabBarComponent, TabConfig } from '@fundamental-ngx/platform/icon-tab-bar';
 import { textTypeConfig } from '../config-for-examples/text-type.config';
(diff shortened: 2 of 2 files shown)


libs/docs/platform/icon-tab-bar/examples/platform-icon-tab-bar-dynamic-overflow-example/platform-icon-tab-bar-dynamic-overflow-example.component.html
```diff
@@ -0,0 +1,8 @@
+<div style="margin-bottom: 8px; display: flex; gap: 8px">
```
droshev: can we use common-css instead?

libs/docs/platform/icon-tab-bar/examples/platform-icon-tab-bar-dynamic-overflow-example/platform-icon-tab-bar-dynamic-overflow-example.component.ts
```diff
@@ -0,0 +1,43 @@
...
+    imports: [IconTabBarComponent, ButtonComponent]
+})
+export class PlatformIconTabBarDynamicOverflowExampleComponent implements OnInit {
+    items: TabConfig[];
+
+    /** Whether the long labels variant is active. */
+    useLongLabels = false;
+
+    ngOnInit(): void {
+        this.items = structuredClone(textTypeConfig);
```
droshev: it can safely be a field initializer

droshev (approved):

PR #13965, opened 2026-03-11
fix(cdk, core): migrate Toast infrastructure to signals and Web   Animations API

InnaAtanasova:
## Related Issue(s)

## Description
  - Migrate BaseToastAnimatedContainerComponent to Web Animations API
  - Remove Angular animations dependencies
  - Add DestroyRef for modern cleanup patterns
  - Add Web Animations API fallback for SSR compatibility
  - Add comprehensive error handling for animation promises
  - Enhance test coverage from 3 to 16 tests
  - Remove unused animation files and exports
  - Update message-toast component to use new animation system

BREAKING CHANGE: removed BaseToastActionDismissibleRef, BaseToastOverlayContainer, BaseAnimatedToastConfig, baseToastAnimations

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.62.0-rc.7",
+    "version": "0.62.0-rc.8",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.62.0-rc.7",
+    "version": "0.62.0-rc.8",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/cdk/package.json
+++ b/libs/cdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cdk",
-    "version": "0.62.0-rc.7",
+    "version": "0.62.0-rc.8",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - CDK",
     "license": "Apache-2.0",
--- a/libs/core/package.json
+++ b/libs/core/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/core",
-    "version": "0.62.0-rc.7",
+    "version": "0.62.0-rc.8",
     "schematics": "./schematics/collection.json",
     "ng-update": {
         "migrations": "./schematics/migrations.json"
--- a/libs/cx/package.json
+++ b/libs/cx/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cx",
-    "version": "0.62.0-rc.7",
+    "version": "0.62.0-rc.8",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - cx",
     "license": "Apache-2.0",
--- a/libs/datetime-adapter/package.json
+++ b/libs/datetime-adapter/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/datetime-adapter",
-    "version": "0.62.0-rc.7",
+    "version": "0.62.0-rc.8",
     "description": "Datetime adapter for SAP Fundamentals, based on Day.js package",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/i18n/package.json
+++ b/libs/i18n/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/i18n",
-    "version": "0.62.0-rc.7",
+    "version": "0.62.0-rc.8",
     "description": "Fundamental Library for Angular - i18n",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/moment-adapter/package.json
+++ b/libs/moment-adapter/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/moment-adapter",
-    "version": "0.62.0-rc.7",
+    "version": "0.62.0-rc.8",
     "description": "Datetime adapter for SAP Fundamentals, based on Moment.js package",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/platform/package.json
+++ b/libs/platform/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/platform",
-    "version": "0.62.0-rc.7",
+    "version": "0.62.0-rc.8",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - platform",
     "license": "Apache-2.0",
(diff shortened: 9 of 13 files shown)


libs/cdk/utils/toast/base-toast.service.ts
```diff
@@ -160,9 +159,9 @@ export abstract class BaseToastService<
         toastRef._isAnchor = isAnchor;
 
         if (content instanceof TemplateRef) {
-            // TemplatePortal requires viewContainer ref
-            const viewRef = null as any as ViewContainerRef;
-            const portal = new TemplatePortal(content, viewRef, { $implicit: config, toastRef } as any);
+            // ViewContainerRef is not needed here as the portal is attached directly to the container's outlet
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            const portal = new TemplatePortal(content, null!, { $implicit: config, toastRef } as any);
```
InnaAtanasova, line 164: Before you ask :)
The second parameter of `TemplatePortal` constructor requires a `ViewContainerRef` according to Angular CDK's type signature. You can't omit it - the constructor signature looks like:
```
constructor(
    template: TemplateRef<C>,
    viewContainerRef: ViewContainerRef,  // ← Required parameter
    context?: C
  )
```

However, when you attach a TemplatePortal to a portal outlet using attachTemplatePortal(), the portal outlet provides its own ViewContainerRef internally. The ViewContainerRef passed to the constructor is essentially ignored in this specific use case.

  So we have two options:
  1. Pass null! (what we're doing) - tells TypeScript "trust me, this will be null but it's
  intentional"
  2. Try to access the actual ViewContainerRef from the container's portal outlet (more complex, unnecessary)

  The comment explains why passing null! is safe here: because the portal is being attached to an outlet that has its own ViewContainerRef, so the one we pass to the constructor doesn't actually get used.

--------------------------

  TL;DR: It's kept because TypeScript requires it, but the comment clarifies that it's not actually used at runtime since the portal outlet provides its own ViewContainerRef when the portal is attached.

  This is a common pattern in Angular CDK portal usage when you know the portal will be attached to an outlet immediately.

libs/core/message-toast/message-toast.service.spec.ts
```diff
@@ -1,29 +1,35 @@
...
+import { MessageToastRef } from './ref/message-toast.ref';
 
-// Minimal Component Setup
+// Test component for custom toast content
 @Component({
-    template: ` <ng-template #template> Template Content </ng-template> `,
-    standalone: true,
-    imports: []
+    template: `<div class="custom-toast-content">Custom Toast</div>`,
+    standalone: true
```
droshev: you can remove that

libs/core/message-toast/message-toast.service.ts
```diff
@@ -18,13 +18,7 @@ export class MessageToastService<P = any> extends BaseDismissibleToastService<Me
     /** @Hidden */
```
droshev: can you make it @hidden

libs/core/message-toast/message-toast.service.ts
```diff
@@ -18,13 +18,7 @@ export class MessageToastService<P = any> extends BaseDismissibleToastService<Me
...
     protected toastPositionStrategy = ToastBottomCenterPosition;
     /** @Hidden */
-    protected defaultConfig;
-
-    /** @hidden */
-    constructor(overlay: Overlay, injector: Injector, @Inject(MESSAGE_TOAST_CONFIG) config: MessageToastConfig) {
-        super(overlay, injector);
-        this.defaultConfig = config;
-    }
+    protected override defaultConfig = inject(MESSAGE_TOAST_CONFIG);
```
droshev, line 21: can you make it @hidden

droshev:
I guess it is worth documenting(BC) that we deleted:
       - BaseToastActionDismissibleRef
       - BaseToastOverlayContainer
       - BaseAnimatedToastConfig
       - baseToastAnimations

libs/core/message-toast/message-toast.service.ts
```diff
@@ -36,13 +30,12 @@ export class MessageToastService<P = any> extends BaseDismissibleToastService<Me
         message: string | Type<T> | TemplateRef<any>,
         config?: MessageToastConfig<P>
     ): MessageToastRef<MessageToastTextComponent | T | EmbeddedViewRef<any>> {
-        const mergedConfig = { ...this.defaultConfig, ...config };
```
MariaIDineva, line 39: Is merging the config removed on purpose?

libs/cdk/utils/toast/base-toast.service.ts
```diff
@@ -107,13 +106,13 @@ export abstract class BaseToastService<
...
     /** @hidden */
-    private _destroyRef = inject(DestroyRef);
+    protected readonly overlay = inject(Overlay);
 
     /** @hidden */
-    protected constructor(
-        public overlay: Overlay,
-        public injector: Injector
-    ) {}
+    protected readonly injector = inject(Injector);
```
MariaIDineva, line 112: Overlay and Injector used to be public and now they are protected, is that desired?

libs/core/message-toast/message-toast.service.spec.ts
```diff
@@ -33,28 +39,246 @@ describe('MessageToastService', () => {
...
+    describe('openFromComponent()', () => {
+        it('should open toast with custom component', fakeAsync(() => {
+            const config = new MessageToastConfig();
+            const toastRef = service.openFromComponent(CustomToastComponent, config);
+            tick();
+
+            expect(toastRef).toBeTruthy();
+            expect(toastRef.instance).toBeInstanceOf(CustomToastComponent);
+
+            toastRef.dismiss();
```
MariaIDineva: These final lines of the it block:
```
toastRef.dismiss();
tick();
```
 seem to be redundant since you have an `afterEach` block that calls `service.hideAll();`
Same is true for lines 66, 79, 96, 111, 125, 169, 230.

libs/cdk/utils/toast/base-toast.service.ts
InnaAtanasova: yes, the constructor was marked as `hidden` before and these were not exposed and most probably never intended as public API. This class was only extended in Message Toast and both should not be exposed publicly.

libs/core/message-toast/message-toast.service.ts
InnaAtanasova: yes,  the merging is now handled individually in each branch instead of upfront:

  1. `openFromTemplate` branch (line 34): Uses `config ?? this.defaultConfig` - if config is provided, use it; otherwise use default
  2. `openFromString` branch (line 36): Passes config directly (no default) - but `openFromString` itself handles the merging on line 47: `const
  mergedConfig = { ...this.defaultConfig, ...config }`
  3. `openFromComponent` branch (line 38): Uses `config ?? this.defaultConfig` - same as template branch

With the old approach, when `config` is undefined, spreading it would still work fine, but when config is explicitly passed (even as an empty object {}), the old code would always create a merged config object upfront.

  With the new approach :
  - `openFromTemplate` and `openFromComponent`: Only use the default config when no config is provided at all (?? this.defaultConfig)
  - `openFromString`: Still merge configs because the string data needs to be added to the config object (line 48: mergedConfig.data = message)

  This change improves performance by avoiding unnecessary object spreading when the full default config will be used anyway, and makes the behavior more explicit about when merging happens versus when defaults are used as-is.

MariaIDineva: Thanks for explaining!

libs/core/message-toast/message-toast.service.ts
InnaAtanasova: I assume you meant the old ones, not introduced in this PR, as this one is correctly marked as `@hidden`. Or I am missing something?

InnaAtanasova:
> I guess it is worth documenting(BC) that we deleted: - BaseToastActionDismissibleRef - BaseToastOverlayContainer - BaseAnimatedToastConfig - baseToastAnimations

these were internal classes, used in the Message Toast component, so technically nothing is required from the user to update, but I will add it in the description.

libs/core/message-toast/message-toast.service.spec.ts
InnaAtanasova: Updated these.

PR #13966, opened 2026-03-11
feat(platform): allow title template in Platform Table Toolbar

InnaAtanasova:
## Related Issue(s)

closes none

## Description
In Platform Table the user can now pass a template for rich content like icons and text.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.62.0-rc.5",
+    "version": "0.62.0-rc.7",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.62.0-rc.5",
+    "version": "0.62.0-rc.7",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/cdk/package.json
+++ b/libs/cdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cdk",
-    "version": "0.62.0-rc.5",
+    "version": "0.62.0-rc.7",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - CDK",
     "license": "Apache-2.0",
--- a/libs/core/package.json
+++ b/libs/core/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/core",
-    "version": "0.62.0-rc.5",
+    "version": "0.62.0-rc.7",
     "schematics": "./schematics/collection.json",
     "ng-update": {
         "migrations": "./schematics/migrations.json"
--- a/libs/cx/package.json
+++ b/libs/cx/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cx",
-    "version": "0.62.0-rc.5",
+    "version": "0.62.0-rc.7",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - cx",
     "license": "Apache-2.0",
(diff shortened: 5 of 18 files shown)


libs/platform/table/components/table-toolbar/table-toolbar.component.html
```diff
@@ -12,13 +12,28 @@
     <fd-toolbar
         fdType="transparent"
         [titleId]="tableToolbarTitleId"
-        [title]="title && !hideItemCount ? title + ' (' + counter() + ')' : title"
+        [title]="titleTemplate ? '' : title && !hideItemCount ? title + ' (' + counter() + ')' : title"
         [shouldOverflow]="shouldOverflow"
         [headingLevel]="headingLevel"
     >
+        @if (titleTemplate) {
+            <span
```
mikerodonnell89, line 20: Doesn't this kind of limit what they can put here? Why not just give them full rein here

InnaAtanasova: Because we want to guarantee the correct roles and aria attributes are applied. The user might pass a normal div or span without heading role and level, which will break the a11y. We need the CSS class "fd-toolbar__title", which now we would rely on the user to provide. We also rely on the id for `ariaLabelledBy`.

libs/platform/table/components/table-toolbar/table-toolbar.component.html
```diff
@@ -12,13 +12,28 @@
...
+                fd-title
+                role="heading"
+                [headerSize]="4"
+                [attr.aria-level]="headingLevel"
```
droshev: HeadingLevel accepts numbers and strings like `h4`, `aria-level` though accepts only number and if `h4` is passed then it will be invalid. Maybe you want to do something similar to how toolbar parses it?

libs/docs/platform/table/examples/platform-table-default-example.component.ts
```diff
@@ -28,7 +30,9 @@ import {
         PlatformTableModule,
         TableInitialStateDirective,
         ButtonComponent,
-        FdDatetimeModule
+        FdDatetimeModule,
+        CdkDragPlaceholder,
```
droshev, line 34: where is this used?

droshev:
I would mark this as a `feat` as it adds a new public `titleTemplate`

libs/platform/table/components/table-toolbar/table-toolbar.component.ts
```diff
@@ -115,6 +123,21 @@ export class TableToolbarComponent implements TableToolbarInterface {
...
+     * ```html
+     * <fdp-table-toolbar [titleTemplate]="customTitleTpl">
+     *   <ng-template #customTitleTpl let-counter="counter">
+     *     <fd-icon glyph="product"></fd-icon> Products ({{ counter() }})
+     *   </ng-template>
+     * </fdp-table-toolbar>
+     * ```
+     */
+    @Input()
+    titleTemplate: TemplateRef<TableToolbarTitleTemplateContext>;
```
droshev: how about using `input` instead? like `input<TemplateRef<TableToolbarTitleTemplateContext> | null>(null);`

libs/docs/platform/table/examples/platform-table-default-example.component.html
```diff
@@ -1,5 +1,5 @@
 <fdp-table [dataSource]="source" [trackBy]="trackBy" emptyTableMessage="No data found">
-    <fdp-table-toolbar title="Order Line Items" [shouldOverflow]="true" [disableRefresh]="true" headingLevel="3">
+    <fdp-table-toolbar [titleTemplate]="customTitle" [shouldOverflow]="true" [disableRefresh]="true" headingLevel="3">
```
droshev, line 2: would it make sense to keep the original here and showcase `titleTemplate` in a separate example instead?

InnaAtanasova: The other examples on the same page show the original way, no need of adding more examples.

libs/platform/table/components/table-toolbar/table-toolbar.component.ts
InnaAtanasova: the entire file is using the old way, this will be handled during the migration, no need to mix stuff now.

libs/docs/platform/table/examples/platform-table-default-example.component.ts
InnaAtanasova: the PR checks were failing because of this.

libs/platform/table/components/table-toolbar/table-toolbar.component.html
InnaAtanasova: I fixed this. And since I needed to use computed and signals I changed `titleTemplate` to signal as you suggested in another comment :)

PR #13968, opened 2026-03-11
feat(i18n): introduce i18n cli commands and GHAction for translation

droshev:
- introduce i18n cli commands
- removed unused keys
- fix the broken comments
- add GitHub Action for auto-generating TypeScript files
- Removed lint-staged hook for .properties files

README.md in libs/nx-plugin/src/executors/i18n-manage/README.md


libs/nx-plugin/src/executors/i18n-manage/executor.ts
```diff
@@ -0,0 +1,353 @@
...
+            case 'remove':
+                return await handleRemove(options, context);
+            case 'search':
+                return await handleSearch(options, context);
+            case 'validate':
+                return await handleValidate(options, context);
+            case 'update':
+                return await handleUpdate(options, context);
+            case 'sort':
+                return await handleSort(options, context);
```
MariaIDineva, line 38: Since we have all the existing keys sorted as part of this PR and we the add command would always insert in the correct place, do we need to keep the sort script?

MariaIDineva (approved):
💯

libs/nx-plugin/src/executors/i18n-manage/executor.ts
droshev: when we add the keys, they are inserted always at the bottom. We need to sort them after

Issue #13969, opened 2026-03-12
InlineHelp directive lost functionality


artvitk:
### Describe your proposal

After the update to v0.61 the InlineHelpDirective has lost several inputs from its parent `BasePopoverClass` (the inheritance was removed), like `additionalBodyClass` or `disabled` inputs.

Meanwhile, the demo is still using them https://sap.github.io/fundamental-ngx/#/core/inline-help#styled

Was that functionality removed by mistake? If not, is there any migration guide?

### Which versions of Angular and Fundamental Library for Angular are affected?

Angular 21, Fundamental Library 0.61+

### Affected Component

InlineHelpDirective

### Priority

Medium

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

InnaAtanasova:
hi @artvitk you can use 0.61.2, it contains a fix for the missing inputs. Thanks for reporting the issue.

PR #13973, opened 2026-03-12
fix(core): bring back missing inputs for Inline Help

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/13969

## Description
The issue was introduced in https://github.com/SAP/fundamental-ngx/pull/13886, where InlineHelpDirective stopped extending BasePopoverClass and lost the inputs that were available through it. 

Additionally to this, the Popover body didn't get an id and the correct role (tooltip). 

The PR also includes:
- Replaced constructor DI with inject()
- Moved from a one-time setAttribute to a host binding `'[attr.aria-describedby]': 'bodyId()'` so it updates if bodyId changes
- Replaced 5 individual effects + ngOnInit with one popoverConfig computed signal and one effect that calls `initialise()` on first run and r`efreshConfiguration()` on subsequent changes 
- User-provided classes are merged with the internal inline-help classes via `combinedBodyClass` computed signal

### DIFF
--- a/libs/core/inline-help/inline-help.directive.ts
+++ b/libs/core/inline-help/inline-help.directive.ts
@@ -1,21 +1,19 @@
 import {
+    booleanAttribute,
+    computed,
     Directive,
     effect,
     ElementRef,
     EmbeddedViewRef,
-    Inject,
+    inject,
     input,
-    OnInit,
-    Optional,
     Renderer2,
-    Self,
     TemplateRef,
-    Type,
     ViewContainerRef
 } from '@angular/core';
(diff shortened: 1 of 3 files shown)


MariaIDineva (commented):
While reviewing these changes I checked out the PR that introduced the reported bug and started tracing what we have lost when `InlineHelpDirective` stopped extending `BasePopoverClass`. There are quite a few inputs that are no longer present in `InlineHelpDirective`, eg - **closeOnNavigation, maxWidth, noArrow**, etc. Not sure which of these are used and should be migrated to the directive but they are definitely no longer exposed which sounds like a quite a breaking change.

PR #13983, opened 2026-03-17
fix(core): migrate TokenComponent to signals

MariaIDineva:
## Related Issue(s)

## Description

BREAKING CHANGE:
Token component migrated to Angular signals - selected is now a model signal requiring .set()/() for access, view queries like tokenWrapperElement require function calls, output subscriptions return OutputRefSubscription instead of Subscription, onCloseClick.emit() requires a value parameter, and ngOnDestroy()/_subscriptions members have been removed in favor of automatic cleanup.

`TokenComponent` migrated from decorators to Angular 21+ signals:

| Category | Before | After |
|----------|--------|-------|
| **Inputs** | `@Input()` | `input()` |
| **Outputs** | `@Output() EventEmitter` | `output()` |
| **View Queries** | `@ViewChild()` | `viewChild()` |
| **Internal State** | Plain properties | `signal()` / `computed()` |

### 1. Selected Property is Now a Model Signal

```typescript
// Before
token.selected = true;
const isSelected = token.selected;

// After
token.selected.set(true);
const isSelected = token.selected();
```

### 2. View Queries Now Require Function Calls

```typescript
// Before
token.tokenWrapperElement.nativeElement

// After
token.tokenWrapperElement()?.nativeElement
```

### 3. Output Subscriptions Return Different Type

```typescript
// Before
subscriptions: Subscription[] = [];

// After
subscriptions: OutputRefSubscription[] = [];
```

### 4. onCloseClick Emit Requires Value

```typescript
// Before
token.onCloseClick.emit();

// After
token.onCloseClick.emit(undefined);
```

### 5. Removed Members

The following members have been removed:

- `ngOnDestroy()` - cleanup is now handled automatically via `takeUntilDestroyed()`
- `_subscriptions` - no longer needed

## Files Updated

- `libs/core/token/token.component.ts` - Main migration
- `libs/core/token/token.component.html` - Template signal calls
- `libs/core/token/tokenizer.component.ts` - Consumer updates
- `libs/platform/form/multi-input/multi-input.component.ts` - Consumer updates
- `libs/core/token/token.component.spec.ts` - Test updates, added more extensive unit tests
- `libs/core/token/tokenizer.component.spec.ts` - Test updates

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.62.0-rc.16",
+    "version": "0.62.0-rc.18",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.62.0-rc.16",
+    "version": "0.62.0-rc.18",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
--- a/libs/cdk/package.json
+++ b/libs/cdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/cdk",
-    "version": "0.62.0-rc.16",
+    "version": "0.62.0-rc.18",
     "schematics": "./schematics/collection.json",
     "description": "Fundamental Library for Angular - CDK",
     "license": "Apache-2.0",
--- a/libs/core/button/base-button.ts
+++ b/libs/core/button/base-button.ts
@@ -29,63 +29,106 @@ export const defaultButtonType = 'standard' as ButtonType;
     }
 })
 export class BaseButton implements HasElementRef, ButtonModel {
-    /** Whether button is in toggled state. */
+    /**
+     * Whether the button is in a toggled state.
+     * Used for toggle buttons that maintain an on/off state.
+     * When true, adds the toggled class and sets aria-pressed="true".
+     */
     readonly toggled = input<boolean, BooleanInput>(false, { transform: booleanAttribute });
 
-    /** Whether button is selected. */
+    /**
+     * Whether the button is selected.
+     * Used in button groups or toolbars to indicate the currently selected option.
+     * When true, sets aria-selected="true".
+     */
     readonly selected = input<boolean, BooleanInput>(false, { transform: booleanAttribute });
 
     /**
-     * Native type of button element
+     * Native type attribute of the button element.
+     * Defaults to 'button' to prevent form submission.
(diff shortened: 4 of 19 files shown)


libs/core/token/token.component.ts
```diff
@@ -39,136 +38,139 @@ import { Subscription, fromEvent } from 'rxjs';
...
-    @Output()
     // eslint-disable-next-line @angular-eslint/no-output-on-prefix
-    onTokenKeydown = new EventEmitter<KeyboardEvent>();
+    readonly onTokenKeydown = output<KeyboardEvent>();
+
+    /** Emitted when token element received or lost focus. */
+    readonly elementFocused = output<boolean>();
+
+    /** Whether the token is disabled. */
+    readonly disabled = input(false);
```
InnaAtanasova: needs transform

libs/core/token/token.component.ts
```diff
@@ -39,136 +38,139 @@ import { Subscription, fromEvent } from 'rxjs';
...
 
     /**
-     * Emitted when token element received or lost focus.
+     * Whether the token is selected.
+     * @hidden Internal signal input - use the `selected` property for programmatic access.
      */
-    @Output()
-    elementFocused = new EventEmitter<boolean>();
+    // eslint-disable-next-line @angular-eslint/no-input-rename
+    readonly _selectedInput = input(false, { alias: 'selected' });
```
InnaAtanasova: Shouldn't this be private if it's internal? Or remove the underscore if it has to be public.

libs/core/token/token.component.ts
```diff
@@ -39,136 +38,139 @@ import { Subscription, fromEvent } from 'rxjs';
...
+    readonly _selectedProgrammatic = signal<boolean | null>(null);
+
+    /** @hidden Internal signal for total count (set by tokenizer) */
+    readonly _totalCount = signal<number | undefined>(undefined);
+
+    /** @hidden Internal signal for item position (set by tokenizer) */
+    readonly _itemPosition = signal<number | undefined>(undefined);
+
+    /** @hidden Effective selected state - combines input and programmatic value */
+    readonly _effectiveSelected = computed(() => {
```
InnaAtanasova: for the new signals, are they intentionally prefixed?

libs/core/token/token.component.ts
```diff
@@ -39,136 +38,139 @@ import { Subscription, fromEvent } from 'rxjs';
...
+        const programmatic = this._selectedProgrammatic();
+        return programmatic !== null ? programmatic : this._selectedInput();
+    });
 
     /** @hidden */
-    private _selected = false;
+    readonly elementRef = inject(ElementRef);
```
InnaAtanasova, line 87: should implement HasElementeRef

libs/core/token/token.component.ts
MariaIDineva: Hm, that's a interesting question. Since that's an input, it cannot be private for sure. But since it has an alias, we would expect that it be used with the aliased name `selected`. Let's talk it over.

MariaIDineva: Like we agreed, this whole complicated logic was replaced with `model()`.

PR #13984, opened 2026-03-17
fix(platform): table group row and no data message need colspan="100%"

### LINKED ISSUE #13971: [Platform] Platform-Table: Wrong colspan value for no data table wrapper column
### Describe your proposal

When using fdp-table-no-data-wrapper in platform table component for empty table or no-data message, while using semanticHighlighting, the wrapper column for the message has wrong colspan value and the table visuals look broken. May be caused by the semanticHighlighting column. Adding selectionMode="multiple" visually fixes broken layout, but if we inspect the colspan value, it's value is still lower than it should be (for a -1).

Image

Image

Angular v20, @fundamental-ngx/core v0.57.5

### Affected Component

Platform Table Component

_No response_

### StackBlitz example(s)

https://stackblitz.com/edit/5mjm2omy-paassg8p?file=src%2Fapp%2Fplatform-table-responsive-columns-example.component.html

### Priority

Low

### Declaration

mikerodonnell89:
fixes #13971

### DIFF
--- a/libs/platform/table/components/table-group-row/table-group-row.component.html
+++ b/libs/platform/table/components/table-group-row/table-group-row.component.html
@@ -4,7 +4,7 @@
     tabindex="0"
     role="gridcell"
     [focusable]="true"
-    [attr.colspan]="tableColumnsLength"
+    colspan="100%"
     [attr.aria-expanded]="row.expanded"
     [attr.data-nesting-level]="row.level + 1"
     (keydown.enter)="toggleGroupRow.emit(row)"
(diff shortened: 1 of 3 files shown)


mikerodonnell89:
> Can you keep the input, but make it accept also a string and make the default value 100%. This way we will keep the functionality, avoid breaking changes and default to 100% in most cases.

I can, but I believe the table header row component is only used internally by platform table so I figured let's rip it out entirely to avoid bloat

PR #13985, opened 2026-03-18
fix(platform): improvements to table popping row visuals

### LINKED ISSUE #13970: [Platform] Platform-Table: Column border styling issues
### Describe your proposal

Platform table border styling gets broken when the table has semantic column (semanticHighlighting), or in addition checkbox column (selectionMode="multiple") and we use [noVerticalBorders]="true".
Also, when having semanticHighlighting and selectionMode="multiple"  (with [noVerticalBorders]="true") in popping state, the table borders look broken, and the semantic column width gets same width as checkbox column. 

Issue was found on version 0.57.x but it is visible also on latest fundamental-ngx version.

Image
Image

Image

Image

Angular v20-v21, @fundamental-ngx/platform v0.57.5 - v0.61.1

### Affected Component

NGX Platform/Table

_No response_

### StackBlitz example(s)

https://stackblitz.com/edit/hyq5fhrn?file=src%2Fapp%2Fplatform-table-responsive-columns-example.component.html

### Priority

Low

### Declaration

mikerodonnell89:
fixes #13970. Further improvements (misaligned primary and secondary title and description, some vertical lines in RTL mode) should be coming from fundamental-styles update

before:
Screenshot 2026-03-18 at 1 40 04 PM

after:
Screenshot 2026-03-18 at 1 39 11 PM


libs/platform/table/table.component.scss
```diff
@@ -651,6 +651,14 @@ fdk-dynamic-portal {
...
     border-left-style: none;
 }
+.fd-table--no-vertical-borders .fd-table__cell:first-child {
+    border-left: var(--sapList_BorderWidth) solid var(--sapList_BorderColor);
+
+    @include fd-rtl() {
+        border-left: transparent;
+        border-right: var(--sapList_BorderWidth) solid var(--sapList_BorderColor);
+    }
+}
```
InnaAtanasova, line 656: I think this can be handled with 
```
.fd-table--no-vertical-borders .fd-table__cell:first-child {
    border-inline-start: var(--sapList_BorderWidth) solid var(--sapList_BorderColor);
}
```

border-inline-start would handle automatically RTL

libs/docs/platform/table/examples/platform-table-responsive-columns-example.component.ts
```diff
@@ -108,6 +108,7 @@ export interface ExampleItem {
     statusColor?: string;
     date: FdDate;
     verified: boolean;
+    semantic?: string;
```
InnaAtanasova: I know it is like this in other places, just trying to understand it for myself, this can be basically any string? Because semantic can be valid, warning, error, information... how are we making sure the correct semantic value is passes? How do we handle wrong value? Aka how do we tell the user they passed a wrong value? I couldn't find it in the code.

mikerodonnell89: We could import a semantic type here to make it more strict

PR #13995, opened 2026-03-19
fix(platform): update feed input markup

droshev:
update platfrom feed input markup

### DIFF
--- a/libs/platform/feed-input/feed-input.component.html
+++ b/libs/platform/feed-input/feed-input.component.html
@@ -1,34 +1,37 @@
-<div class="fd-feed-input" [class.is-disabled]="disabled">
-    @if (avatarShow) {
-        <fd-avatar
-            class="fd-feed-input__thumb"
-            size="s"
-            role="img"
-            [attr.aria-label]="('platformFeedInput.userTitle' | fdTranslate)()"
-            [title]="('platformFeedInput.userTitle' | fdTranslate)()"
-            [image]="avatarSrc"
-            [class.fd-avatar--placeholder]="!avatarSrc"
-            [class.sap-icon]="!avatarSrc"
-            [glyph]="!avatarSrc ? 'person-placeholder' : null"
-        >
-        </fd-avatar>
-    }
-    <textarea
-        #textAreaElement
-        fd-form-control
-        class="fd-textarea fd-feed-input__textarea"
-        [disabled]="disabled"
-        [placeholder]="placeholder"
-        [(ngModel)]="value"
-        (ngModelChange)="onChange()"
-    ></textarea>
-    <button
-        fd-button
-        class="fd-feed-input__submit-button"
-        glyph="feeder-arrow"
-        [title]="('platformFeedInput.sendButton' | fdTranslate)()"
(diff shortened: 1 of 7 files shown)


libs/platform/feed-input/feed-input.component.html
```diff
@@ -1,34 +1,37 @@
...
+                [class.fd-avatar--placeholder]="!avatarSrc"
+                [class.sap-icon]="!avatarSrc"
+                [glyph]="!avatarSrc ? 'person-placeholder' : null"
+            >
+            </fd-avatar>
+        }
+        <textarea
+            #textAreaElement
+            fd-form-control
+            [id]="textareaId"
```
InnaAtanasova, line 20: What is the purpose of this id?

libs/platform/feed-input/feed-input.component.html
```diff
@@ -1,34 +1,37 @@
...
-    ></button>
+<div class="fd-feed-input" role="group" [attr.aria-label]="ariaLabel()" [class.is-disabled]="disabled">
+    <div class="fd-feed-input__container">
+        @if (avatarShow) {
+            <fd-avatar
+                class="fd-feed-input__thumb"
+                size="s"
+                role="img"
+                [attr.aria-label]="('platformFeedInput.userTitle' | fdTranslate)()"
+                [title]="('platformFeedInput.userTitle' | fdTranslate)()"
```
InnaAtanasova, line 9: if you have aria-label you don't need title attribute.

libs/platform/feed-input/feed-input.component.ts
```diff
@@ -60,9 +63,15 @@ export class FeedInputComponent implements AfterViewInit {
...
 
+    /** Value for aria label */
+    ariaLabel = input<string>();
+
     /** @hidden Textarea entered value */
     value: string | null;
 
+    /** @hidden Unique id for the textarea element */
+    textareaId = `fdp-feed-input-textarea-${feedInputTextareaId++}`;
+
```
InnaAtanasova, line 74: Why do you need this?

libs/platform/feed-input/feed-input.component.html
droshev: it gives error in the browser
Screenshot 2026-03-19 at 11 38 06

libs/platform/feed-input/feed-input.component.ts
droshev: for https://github.com/SAP/fundamental-ngx/pull/13995#discussion_r2960890711

libs/platform/feed-input/feed-input.component.html
InnaAtanasova: Adding an id only makes the console error disappear but does not fix the problem that this warning is trying to tell. It asks for an id so you connect it with a label.

libs/platform/feed-input/feed-input.component.html
mikerodonnell89: Maybe we want the title for the tooltip? and don't need aria-label

PR #13997, opened 2026-03-19
fix(core): quick view styling fixes

mikerodonnell89:
before:
Screenshot 2026-03-19 at 11 26 17 AM

after:
Screenshot 2026-03-19 at 11 25 56 AM

### DIFF
--- a/libs/core/quick-view/quick-view-group-title/quick-view-group-title.component.html
+++ b/libs/core/quick-view/quick-view-group-title/quick-view-group-title.component.html
@@ -2,7 +2,7 @@
     class="fd-form-group__header fd-form-group__header--no-border fd-form-group__header--no-padding"
     [attr.aria-labelledby]="id"
 >
-    <span class="fd-form-group__header-text" [id]="id">
+    <span class="fd-form-group__header-text" [id]="id" role="heading" aria-level="headingLevel()">
         <ng-content></ng-content>
     </span>
 </div>
--- a/libs/core/quick-view/quick-view-group-title/quick-view-group-title.component.ts
+++ b/libs/core/quick-view/quick-view-group-title/quick-view-group-title.component.ts
@@ -1,4 +1,5 @@
-import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
+import { ChangeDetectionStrategy, Component, input, Input } from '@angular/core';
+import { HeadingLevel } from '@fundamental-ngx/core/shared';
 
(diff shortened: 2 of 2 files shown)


libs/core/quick-view/quick-view-group-title/quick-view-group-title.component.html
```diff
@@ -1,5 +1,8 @@
-<div class="fd-form-group__header" [attr.aria-labelledby]="id">
-    <h1 class="fd-form-group__header-text" [id]="id">
+<div
+    class="fd-form-group__header fd-form-group__header--no-border fd-form-group__header--no-padding"
+    [attr.aria-labelledby]="id"
+>
+    <span class="fd-form-group__header-text" [id]="id">
         <ng-content></ng-content>
-    </h1>
+    </span>
```
InnaAtanasova: This was wrong before, but now it's also not correct. It can remain a span element, but needs role="heading" and an option for the user to pass the heading level (1-6)

libs/core/quick-view/quick-view/quick-view.component.scss
```diff
@@ -4,3 +4,8 @@
 .fd-popover__body .fd-quick-view__content {
     border-radius: var(--sapElement_BorderCornerRadius);
 }
+.fd-quick-view fd-quick-view-group-title + fd-quick-view-group-item {
+    .fd-form-item {
+        padding-block-start: 0;
+    }
```
InnaAtanasova: This padding top is coming from here:
Image

You introduce additional specificity, so both rules will be present in the code. Try with just 
```
.fd-quick-view .fd-form-item {
    padding-block-start: 0;
}
```
and if it doesn't work then add specificity.

mikerodonnell89: Then the form items will have no padding between each other, which we do want

InnaAtanasova: I see. In this case 
```
fd-quick-view-group-title + fd-quick-view-group-item {
    .fd-form-item {
        padding-block-start: 0;
    }
}
```
would minimize one level of specificity (removing the parent class selector). I am being pick with this because overwriting something by the user would require a long chain of selectors and it's good to be avoided when possible. And this is something the user might decide to change and add more/less spacing.

mikerodonnell89: Tried removing the parent selector and it doesn't work

PR #13998, opened 2026-03-19
fix(core): prevent NG0600 error and site unresponsiveness

MariaIDineva:
## Related Issue(s)

Fixes a bug introduced in https://github.com/SAP/fundamental-ngx/pull/13954 

## Description

**The problem that we tried to fix in the previous PR:**

When step-input reaches its min/max value, it updates _canIncrementSignal/_canDecrementSignal, causing buttons to become disabled during template rendering. The browser synchronously fires focusout when a focused button becomes disabled. Since step-input is wrapped in fd-form-input-message-group (which uses a popover with focusin/focusout triggers), the popover tries to write to its isOpen signal while Angular is still evaluating the template — triggering NG0600.

In the PR that intrduces the bug, we added `queueMicrotask` as a solution to the problem, but that led to an infinite loop in some scenarios where the app becomes unresponsive:

1. User focuses input → focusin event fires
2. Popover service receives event → queueMicrotask(() => this.toggle(true, false)) is scheduled
3. Microtask runs → open() is called → popover opens
4. Popover opening creates DOM elements (the message popover body in the overlay)
5. This can cause focus to shift (especially with _focusFirstTabbableElement())
6. Focus shift triggers focusout on the original input
7. focusout event fires → queueMicrotask(() => this.toggle(false, true)) is scheduled
8. Microtask runs → close() is called → popover closes
9. Popover closing returns focus (via _focusLastActiveElementBeforeOpen())
10. Focus returns to input → focusin fires again
11. Goto step 2 → infinite loop!

**Another solution:**

Added _safeSetIsOpen() method in PopoverService that wraps signal writes in try-catch. If NG0600 is caught, the write is deferred via setTimeout, allowing Angular's render cycle to complete first. Non-NG0600 errors are rethrown. Normal operation remains synchronous (no unnecessary deferrals).

## Screenshots

**The bug with the site's unresponsiveness:**

https://github.com/user-attachments/assets/99b5a4b7-f0f0-4eb4-8429-3a7375e13bbb

**After the fix:**

https://github.com/user-attachments/assets/cbb0bce6-60de-4e96-b17c-23234908aece

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.62.0-rc.16",
+    "version": "0.62.0-rc.17",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.62.0-rc.16",
+    "version": "0.62.0-rc.17",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
     "homepage": "https://sap.github.io/fundamental-ngx",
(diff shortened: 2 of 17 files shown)


droshev:
other people raised the same issue https://github.com/angular/angular/issues/60247 and it doesn't look like there is a proper way of solving it

PR #14004, opened 2026-03-19
fix(platform): multi-input popover width default to 0 bug

### LINKED ISSUE #14002: Platform multi input popover body not working
> Platform multi input popover body not working. comes from https://github.com/SAP/fundamental-ngx/pull/13886
> 
> https://github.com/user-attachments/assets/d272d41e-69d4-473b-ae81-f886d669334f 

 _Originally posted by @mikerodonnell89 in [#13987](https://github.com/SAP/fundamental-ngx/issues/13987#issuecomment-4092111360)_

mikerodonnell89:
fixes #14002

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.62.0-rc.17",
+    "version": "0.62.0-rc.19",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
--- a/libs/btp/package.json
+++ b/libs/btp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/btp",
-    "version": "0.62.0-rc.17",
+    "version": "0.62.0-rc.19",
     "description": "Fundamental Library for Angular - BTP",
     "license": "Apache-2.0",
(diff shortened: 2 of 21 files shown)


libs/platform/form/multi-input/multi-input.component.html
```diff
@@ -10,7 +10,6 @@
         [focusTrapped]="true"
         [triggers]="triggers"
         [disabled]="disabled || readonly"
-        [maxWidth]="(!autoResize && minWidth) || 0"
```
InnaAtanasova, line 13: You can't remove this line. The popover needs width constraints to prevent it from:
1. Overflowing the viewport edge (extending beyond the right side of the screen)
2. Growing too wide when autoResize is disabled

You have the following two scenarios:
1. When `autoResize = true` (Dynamic sizing enabled)
- The popover should resize based on content
- BUT it still needs a limit: the available space to the right edge of the screen
- Without maxWidth, the popover could extend off-screen and get clipped

2. When `autoResize = false` (Fixed sizing)
- The popover should match the input control's width
- Sets maxWidth to minWidth (the input's width) to enforce this

You can update it to: `[maxWidth]="autoResize ? (maxWidth ?? null) : (minWidth ?? null)"`.
- autoResize ON: Use maxWidth (calculated available space from `base-multi-input.ts:594: window.innerWidth - scrollBarWidth - rect.left)`
- autoResize OFF: Use minWidth (input control width)

Issue #14005, opened 2026-03-20
fast-deep-equal and dayjs deps aren't esm


robertIsaac:
### Describe your proposal

it seems we have inter deps from fundamental for fast-deep-equal and dayjs
I think dayjs has an esm version that we need to move to
but for fast-deep-equal we need to find another library
this is causing problems for our compliant pipeline with vitest

### Which versions of Angular and Fundamental Library for Angular are affected?

angular 21 and fundamental 0.59.2

### Priority

None

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

robertIsaac:
actually fast-deep-equal is done https://github.com/SAP/fundamental-ngx/pull/13884
so only dayjs and we need to use the latest version of fundamentals

droshev:
@robertIsaac is there anything outstanding from this issue?

robertIsaac:
> [@robertIsaac](https://github.com/robertIsaac) is there anything outstanding from this issue?

right now the only outstanding is that dayjs is loaded fully even if one method is used since angular builder can't tree shake cjs libraries
but it's a low priority
I can try to make an auto migration to replace all dayjs imports to esm if that will make it a yes for you to merge

droshev:
closing it and waiting @robertIsaac to open a PR for further discussions

PR #14006, opened 2026-03-20
fix: use dayjs as esm

robertIsaac:
Fixes https://github.com/SAP/fundamental-ngx/issues/14005


robertIsaac:
The main problem is vitest only work with esm and error when esm package import cjs
Full details of the problem history has been sent to you in teams

robertIsaac:
I have also reported the issue and linked it now

droshev:
closing it for now as the issue has been solved on the app side

Issue #13990, opened 2026-03-19
[Core] [fd-list] ACC-272.1 (Level A) selection mode should change between groups using shift + tab key instead of in the same group


Meggie886:
### Describe your proposal

When focus is on a selected list item within a grouped list and the user presses Shift+Tab, the focus should move to the immediately previous focusable element in the tab order (typically the control/label preceding the group or the previous UI element), not to the first list item in the group.

https://fundamental-ngx.netlify.app/#/core/list

<img width="2619" height="1240" alt="Image" src="https://github.com/user-attachments/assets/a14e9749-c004-45b0-9c80-dad399e7dad4" />

### Proposed Solution

UI5 is correct. [https://ui5.github.io/webcomponents/components/List/](https://ui5.github.io/webcomponents/components/List/)

<img width="1099" height="1256" alt="Image" src="https://github.com/user-attachments/assets/25f97ade-eb72-4973-809d-0092531f1d54" />

### Priority

None

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

PR #14009, opened 2026-03-20
fix(core,platform): fix list item tabindex

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13990

## Description

**Problem:**
Pressing `shift + tab` would always take the user to the first item of the current list instead of a previous focusable element, eg. an item in a previous list. Upon pressing `shift+tab` again, it would take the user to the first item of the previous list even if another item in the list was selected.

**Solution:**
This PR implements roving tabindex for list components in both core and platform libraries. 
Now when `shift + tab` is pressed, the focus goes to the previous focused list item in the previous list. If no item is selected, it focuses the first item.

### Core Changes in list.component.ts

- Added `_listenOnActiveItemChange()` - subscribes to keyManager.change to update tabindexes when keyboard navigation changes the active item
- Added `_updateTabIndexes()` - refreshes tabindex on all items based on the tabbable index
- Added `_getTabbableItemIndex()` - determines which item gets tabindex="0" with priority:
  - Active item (keyboard focused)
  - First selected item
  - First item (default)
- Changed `updateItemsProperties()` to use tabbableIndex instead of hardcoded index === 0
- Rename `_isFirstItem$` to `_isTabbable$` and `setIsFirst` to `setIsTabbable` in `ListFocusItem` to reflect the new logic.

### Platform Changes list.component.ts

- Similar roving tabindex logic with `_updateTabIndexes()` and `_getTabbableItemIndex()`
- Now subscribes to _keyManager.change
- Replaced direct tabindex='0' on first item with dynamic roving logic

## Screenshots

**Before:**

https://github.com/user-attachments/assets/4d064f02-ec87-4c65-8eb9-9446356f20d9

**After:**

https://github.com/user-attachments/assets/0af648b2-4673-45ba-9054-6689e95f2ea1

### DIFF
--- a/libs/core/list/list-focus-item.model.ts
+++ b/libs/core/list/list-focus-item.model.ts
@@ -7,10 +7,10 @@ import { Subject } from 'rxjs';
 export abstract class ListFocusItem<T = any> implements KeyboardSupportItemInterface {
     @Input()
     set tabindex(value: number) {
-        this._tabIndex$.set(coerceNumberProperty(value, -1));
+        this._tabIndex.set(coerceNumberProperty(value, -1));
     }
     get tabindex(): number {
-        return this._normalizedTabIndex$();
+        return this._normalizedTabIndex();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


libs/core/list/list-focus-item.model.ts
```diff
@@ -33,14 +33,14 @@ export abstract class ListFocusItem<T = any> implements KeyboardSupportItemInter
     readonly _clicked$ = new Subject<MouseEvent>();
 
     /** @hidden */
-    protected _isFirstItem$ = signal(false);
+    protected _isTabbable$ = signal(false);
```
InnaAtanasova: is there a reason to keep the $? I see it was like this before, but we put the $ for RxJS stuff so I am wondering why we have it here.

MariaIDineva: No reason, you are right. I will remove the $.

Issue #14012, opened 2026-03-23
Breadcrumb does not respond to viewport width changes after initial render


MariaIDineva:
### Describe your proposal

Breadcrum does not seem to react to width changes. It stays as it is initially rendered. If it was rendered without overflow, when reducing the screen size, no overflow shows. If it was rendered with an overflow, increasing the width does not make items expand.

### Which versions of Angular and Fundamental Library for Angular are affected?

latest

### Affected Component

Breadcrumb

### Steps to Reproduce (for bugs)

1. Go to https://fundamental-ngx.netlify.app/#/core/breadcrumb
2. Start reducing the screen width.
3. Notice how breadcrumb items start spilling out of screen.
4. Refresh the page as it is, on a screen with small width.
5. Notice how some of the breadcrumb items are now rendered in the overflow.
6. Start increasing the screen width.
7. Notice that the items stay in the overflow even if there is plenty of space for them to be displayed.

https://github.com/user-attachments/assets/af614c6e-8454-43e4-8e27-264a28c099c4

### Priority

Medium

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

Issue #14000, opened 2026-03-19
Overflow Layout not working


InnaAtanasova:
Looks like the functionality stopped working:
<img width="1312" height="895" alt="Image" src="https://github.com/user-attachments/assets/00cae7bb-0086-41fc-b66a-004ae429fcae" />

PR #14014, opened 2026-03-23
fix(core): fix overflow functionality for Breadcrumb and Overflow Layout components

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14012
closes https://github.com/SAP/fundamental-ngx/issues/14000

## Description
both issues were connected as Breadcrumb is using overflow layout functionality.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.62.0-rc.19",
+    "version": "0.62.0-rc.21",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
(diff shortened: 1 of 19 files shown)


libs/core/overflow-layout/overflow-layout.component.scss
```diff
@@ -56,6 +56,7 @@ $fd-block: 'fd-overflow-layout';
         @include fd-reset();
         display: block;
         white-space: normal;
+        max-width: 19.5rem;
```
MariaIDineva, line 59: This is not related to the main issues referred in the description, right? Is that related to the flickering that happens when opening the overflow and items initially grow bigger?

InnaAtanasova: yes, it's the flickering that I notices when you open the overflow menu. It's because initially the width of the popover "grows" to fit all the items inside and then goes to 19.5rem width. I don't know why this width was hardcoded before, maybe it was in the design. Setting max-width as 19.5rem removes the auto resize that looks like flickering.

PR #14016, opened 2026-03-24
fix: datetime adapters migration

### LINKED ISSUE #12258: Dayjs strict mode showing local date not UTC
No

Latest version https://sap.github.io/fundamental-ngx/#/core/dayjs-datetime-adapter

- Go to DayJS datetime adaptor example
- Open the calender example in UTC
- Expected behaviour (date should be the UTC date not the local time zone date)
Screenshot 2024-08-13 at 16 39 59

It's working as expected using Moment adaptor https://sap.github.io/fundamental-ngx/#/core/moment-datetime-adapter
Screenshot 2024-08-13 at 16 39 39

Yes

Might be related to/caused by #10230 

droshev:
Modernization of the datetime adapters (`FdDatetimeAdapter`, `DayjsDatetimeAdapter`, `MomentDatetimeAdapter`)

# Key Files for Review

| File | What changed |
|------|--------------|
| `libs/core/datetime/datetime-adapter.ts` | **BREAKING**: `locale` is now `Signal<string>`, `getAmountOfWeeks` off-by-one fix |
| `libs/core/datetime/fd-datetime-adapter.ts` | `fromNow()`, locale-aware parsing, `getFirstDayOfWeek` via Intl, param name fixes |
| `libs/datetime-adapter/src/lib/dayjs-datetime-adapter.ts` | Module-level plugins, `fromNow()`, `createDate()` UTC fix, locale fallback (warn instead of throw) |

## Manual Testing Guide 

go to https://deploy-preview-14016--fundamental-ngx.netlify.app/#/core/datetime-adapters

### Locale switching

Open the Dayjs locale example. Switch between **en**, **fr**, **de**, and **ja** locales. For each locale verify:

- **Date picker**: month/day names update, input field reformats the selected date in the locale's format (e.g. `MM/DD/YYYY` for en, `DD/MM/YYYY` for fr, `DD.MM.YYYY` for de, `YYYY/MM/DD` for ja)
- **Time picker**: AM/PM labels update or disappear depending on locale (e.g. ja uses 24-hour clock)
- **Calendar**: month and weekday headers re-render in the new language, first day of week changes (Monday for fr/de, Sunday for en)

### UTC mode

Open the Dayjs options example and enable `useUtc: true`. Create a date via the date picker (e.g. select January 1, 2025). Verify:

- The displayed date matches what you selected -- it should **not** shift by your local timezone offset (e.g. selecting Jan 1 should not display Dec 31 or Jan 2)
- The ISO string output shows `2025-01-01T00:00:00` without a `Z` suffix
- Switching between UTC and local mode preserves the calendar date you selected

fixes #12258

BREAKING CHANGE:
 - `DatetimeAdapter.locale`: `DatetimeAdapter.localeChanges` Observable has been removed.`locale` is now a `Signal<string>`. 
 - deprecated `DayjsDatetimeAdapterRawModule` is removed from `dayjs-datetime-adapter.module.ts`
 - `_prepareFormat()` and `_createDayjsDate()` on DayjsDatetimeAdapter are now explicitly private - meant to be private
 - `DATE_TIME_FORMATS` injection in `DateFormatPipe` and `DateTimeFormatPipe` is now required (was `@Optional()`). Apps that use these pipes without `provideDateTime()`, `DayjsDatetimeAdapterModule`, or `MomentDatetimeAdapterModule` must provide the token explicitly via `{ provide: DATE_TIME_FORMATS, useValue: ... }`.

Bug fixes — `DayjsDatetimeAdapter`

  - `parse()` decoupled from FdDate: Non-Date objects with a `toString()` method (e.g. FdDate) are now converted to string before parsing, avoiding misinterpretation by the dayjs objectSupport plugin (which uses 0-based months and date instead of day)
  - Strict-then-loose parsing with overflow detection: `_createDayjsDate()` now tries multiple format fallbacks (original format, date-only, localized L, YYYY-MM-DD) and validates that non-strict parsing didn't silently wrap overflow values (e.g. month 13 → Feb next year)
  - `createDate()` UTC fix: When `useUtc`: true, builds dates via string format (YYYY-MM-DD) instead of native Date constructor, avoiding timezone offset shift
  - `datesEqual()` timezone normalization: Converts both dates to .local() before comparing, so UTC and local dates for the same calendar day are equal
  - `timeInput`/`dateTimeInput`parse formats: Changed from h:mm to h:mm A (and M/D/YYYY h:mm A) to include the AM/PM designator, preventing silent misparsing of 12-hour times

Bug fixes — `FdDatetimeAdapter`

- `fromNow()` implementation: Added locale-aware relative time formatting using Intl.RelativeTimeFormat
- Locale-aware `parse()`: Uses `Intl.DateTimeFormat` to detect the locale's date component order (MDY/DMY/YMD) and parses accordingly
- Regex-based time parsing: Parses time strings like 2:30 PM, 14:30:45, and combined datetime strings
- `getFirstDayOfWeek()` locale-aware: Uses `Intl.Locale.getWeekInfo()` (with weekInfo fallback) instead of hardcoded Sunday
- `toIso8601()` fix: Uses `Intl.DateTimeFormat` to preserve local time components instead of `toISOString()` which converts to UTC

Bug fixes — `MomentDatetimeAdapter`

- `toIso8601()` fix: Changed from `toISOString()` (UTC conversion) to `format('YYYY-MM-DDTHH:mm:ss')` to preserve local time

Migration guide:

  If you subscribe to localeChanges:
  // Before
```
  this.dateTimeAdapter.localeChanges.subscribe(() => {
      this.recalculate();
      this.cdr.markForCheck();
  });
```

  // After
```
  effect(() => {
      this.dateTimeAdapter.locale(); // triggers on change
      this.recalculate();
  });
```
  Note: effect() must be called in an injection context (constructor or field initializer).
  markForCheck() is no longer needed inside effects.

  If you have a custom DatetimeAdapter subclass:
  // Before
```
  this.locale   // string property
```

  // After
```
  this.locale() // signal — call it as a function
```


mikerodonnell89 (changes requested):
I noticed that changing locales on [this example](https://deploy-preview-14016--fundamental-ngx.netlify.app/#/core/datetime-adapters#dayjs-locale) changes the locale of every datetime adapter on the page, but only after I select a new date. I do not think this is intentional. Video below

https://github.com/user-attachments/assets/febba33a-6aab-4e20-9be4-261ebfb44cab

mikerodonnell89 (commented):
When I change the global example to Arabic, I'm not seeing any of the other examples on the page update, even the ones that do not specify any locale info. I think those should update?

libs/core/datetime-picker/datetime-picker.component.ts
```diff
@@ -517,6 +513,17 @@ export class DatetimePickerComponent<D>
 
         // default model value
         this.date = _dateTimeAdapter.now();
+
+        let localeEffectInitialized = false;
```
MariaIDineva, line 517: I ran a PR review through Claude and he suggests that having this `localeEffectInitialized` flag which is mutated from within the effect is an anti-pattern, so we should remove it.
Also, having just `if (this._inputFieldDate)` should do the trick since that matches the logic that was previously present in the pipe - `filter(() => this._inputFieldDate !== '')`.

libs/core/calendar/calendar-header/calendar-header.component.ts
```diff
@@ -226,7 +221,14 @@ export class CalendarHeaderComponent<D> implements OnInit, OnChanges {
         private _changeDetRef: ChangeDetectorRef,
         private _calendarService: CalendarService,
         private _dateTimeAdapter: DatetimeAdapter<D>
-    ) {}
+    ) {
+        effect(() => {
+            this._dateTimeAdapter.locale();
+            this._calculateMonthNames();
+            this._calculateLabels();
```
MariaIDineva: `_calculateMonthNames` and `_calculateLabels` are called from within the effect and from within ngOnInit which will cause initial double computation. Maybe add an additional `if (this._initiated)` check similar to the one in [calendar-aggregated-year-view.component.ts](https://github.com/SAP/fundamental-ngx/pull/14016/changes#diff-9577d0562a923e4c193203d5c3f60f4251b612cf0213a40cf07c69f58fff628a)

libs/core/datetime/datetime-adapter.ts
```diff
@@ -351,14 +358,14 @@ export abstract class DatetimeAdapter<D> {
...
-     * @param firstDayOfWeek The first day of week. 1 - Sunday, 2 - Monday...
+     * @param firstDayOfWeek The first day of week. 0 - Sunday, 1 - Monday, ..., 6 - Saturday
      * @returns Number of weeks in the given month
      */
     getAmountOfWeeks(year: number, month: number, firstDayOfWeek: number): number {
         const firstOfMonth = new Date(year, month - 1, 1);
         const lastOfMonth = new Date(year, month, 0);
 
-        const dayOffset = (firstOfMonth.getDay() - firstDayOfWeek + 8) % 7;
+        const dayOffset = (firstOfMonth.getDay() - firstDayOfWeek + 7) % 7;
```
MariaIDineva: If that change is intentional, I think we should document it as a breaking change.

libs/core/datetime/datetime-format.pipes.ts
```diff
@@ -1,18 +1,16 @@
...
 })
 export class DateFormatPipe<D> implements PipeTransform {
     /** @hidden */
-    constructor(
-        private _dateTimeAdapter: DatetimeAdapter<D>,
-        @Optional() @Inject(DATE_TIME_FORMATS) private _dateTimeFormats: DateTimeFormats
-    ) {}
+    private readonly _dateTimeAdapter = inject(DatetimeAdapter) as DatetimeAdapter<D>;
+    /** @hidden */
+    private readonly _dateTimeFormats = inject(DATE_TIME_FORMATS) as DateTimeFormats;
```
MariaIDineva, line 13: Shouldn't that be `inject(DATE_TIME_FORMATS, { optional: true })`?

libs/core/datetime/datetime-format.pipes.ts
```diff
@@ -25,15 +23,13 @@ export class DateFormatPipe<D> implements PipeTransform {
...
 })
 export class DateTimeFormatPipe<D> implements PipeTransform {
     /** @hidden */
-    constructor(
-        private _dateTimeAdapter: DatetimeAdapter<D>,
-        @Optional() @Inject(DATE_TIME_FORMATS) private _dateTimeFormats: DateTimeFormats
-    ) {}
+    private readonly _dateTimeAdapter = inject(DatetimeAdapter) as DatetimeAdapter<D>;
+    /** @hidden */
+    private readonly _dateTimeFormats = inject(DATE_TIME_FORMATS) as DateTimeFormats;
```
MariaIDineva, line 32: Same here - `{ optional: true }`?

MariaIDineva (approved):

MariaIDineva:
Just one thing that I noticed while playing with the examples for [Locale Scoping](https://deploy-preview-14016--fundamental-ngx.netlify.app/#/core/datetime-adapters#dayjs-locale-scoping). I was expecting that when I changed `Global (route-level)` locale, the other examples would change as well, shouldn't they?

https://github.com/user-attachments/assets/3139a6f0-6c72-4443-8c87-9c2d4e2855ac

droshev:
> When I change the global example to Arabic, I'm not seeing any of the other examples on the page update, even the ones that do not specify any locale info. I think those should update?

@mikerodonnell89 All the other examples are using a separate instance of the Adapter. so they should not update.

droshev:
> Just one thing that I noticed while playing with the examples for [Locale Scoping](https://deploy-preview-14016--fundamental-ngx.netlify.app/#/core/datetime-adapters#dayjs-locale-scoping). I was expecting that when I changed `Global (route-level)` locale, the other examples would change as well, shouldn't they?
> 
>  Screen.Recording.2026-04-14.at.10.49.32.mov

it should change if they don't have a local adapter, which they have

droshev:
@mikerodonnell89 and @MariaIDineva thank you for the feedback! the fact that both of you were expecting something made me think i had not explained that well. I updated the examples and removed the wording `Global`. I hope it is clearer now

libs/core/datetime-picker/datetime-picker.component.ts
droshev: The flag is meant to skip the first run, only react to future locale changes. Your suggestion to check if (this._inputFieldDate) instead doesn't work because the input field already has a value at that point - the check would pass and we'd do the unnecessary work anyway.

libs/core/datetime/datetime-format.pipes.ts
droshev: On main, these pipes had `@Optional()` but never handled the null case: they would crash at runtime with a confusing error. Making the injection required surfaces a clear Angular error at startup instead. These pipes can't work without formats, so requiring them is correct.

libs/core/datetime/datetime-format.pipes.ts
droshev: same here

libs/core/calendar/calendar-header/calendar-header.component.ts
droshev: yes, makes sense!

libs/core/datetime/datetime-adapter.ts
droshev: You're right! The original +8 was correct: it expects 1-based input (1=Sunday), matching how all callers use it. I got confused because `getFirstDayOfWeek()` returns 0-based( and no tests). Reverted now with correct 1-based tests.

MariaIDineva:
> @mikerodonnell89 and @MariaIDineva thank you for the feedback! the fact that both of you were expecting something made me think i had not explained that well. I updated the examples and removed the wording `Global`. I hope it is clearer now

Yes, at least to me, it makes more sense now. Thanks!

MariaIDineva (dismissed):

InnaAtanasova:
This is what Claude provided as a review:
1. [libs/core/datetime/datetime-format.pipes.ts] Date format pipes are pure but need to react to locale signal changes
       - The `DateFormatPipe` and `DateTimeFormatPipe` call `adapter.format()`, which internally reads `this.locale()`
       - Pure pipes only re-run when their inputs change by reference, but locale changes won't trigger the pipe because the date reference stays the same
       - The `TranslateDayPeriodPipe` is correctly marked pure: false, but the other format pipes are not
       - Fix: Add pure: false to DateFormatPipe and DateTimeFormatPipe metadata, with a comment explaining they react to locale signal changes

2. [libs/core/calendar/calendar-header/calendar-header.component.ts:228-236] Effect calls `markForCheck` unnecessarily
       - The effect reads `_dateTimeAdapter.locale()` and calls `markForCheck()` after recalculating labels
       - Per CLAUDE.md: "DO NOT use `signal.set()` then `markForCheck()` when the signal is read in the template or a host binding -- Angular automatically tracks and marks the component"
       - While this effect doesn't write to signals, the recalculation methods update properties (selectMonthLabel, etc.) that are bound in the template
       - The _initiated flag pattern is correct to avoid running during construction
       - Question: Are selectMonthLabel, selectYearLabel, etc. signals or plain properties? If plain properties bound in the template, the markForCheck() is appropriate. If signals, remove it.
    
3. [libs/core/datetime/datetime-adapter.ts:364-372] getAmountOfWeeks off-by-one fix uses inconsistent modulo
       - Line 368: const dayOffset = (firstOfMonth.getDay() - firstDayOfWeek + 8) % 7;
       - The commit message for 8d1f9a236e shows the correct fix used + 7, but this PR changed it to + 8
       - The + 8 appears to be a workaround that masks the real issue: firstDayOfWeek parameter convention is inconsistent
       - The JSDoc says "1 - Sunday, 2 - Monday" (1-based), but callers pass 0-based values (0=Sunday)
       - Fix: Either:
           - Use + 7 and update JSDoc to clarify the parameter is 0-based, OR
         - Keep + 8 but add a comment explaining the conversion from 1-based to 0-based convention

mikerodonnell89 (commented):
Could be good to add Japan to the segmented button to demonstrate the "ja uses 24-hour clock" thing

mikerodonnell89 (commented):
Image

Looks like this PR doesn't actually address the 5-row problem in the description. The description says "July 2017", this is June

mikerodonnell89 (commented):
Doesn't appear to be unique to your PR but I noticed the (dateChange) event isn't firing for fd-datetime-picker

mikerodonnell89 (changes requested):
Editing the input text with a valid date shows an error state on this example:

https://github.com/user-attachments/assets/a8d11551-0426-45a3-85f2-ad4dfc9e8641

droshev:
> Image
> Looks like this PR doesn't actually address the 5-row problem in the description. The description says "July 2017", this is June

that was wrong to be checked, i deleted that part from the description

droshev:
> Doesn't appear to be unique to your PR but I noticed the (dateChange) event isn't firing for fd-datetime-picker

reported it as a separate issue: https://github.com/SAP/fundamental-ngx/issues/14258

droshev:
> Editing the input text with a valid date shows an error state on this example:
> 
>  Screen.Recording.2026-05-28.at.12.33.27.PM.mov

fixed

mikerodonnell89 (approved):

Issue #13992, opened 2026-03-19
[Platform][fdp-table]ACC-264.1 (Level A) Should not read  additional info "empty" when there is an button inside a table cell


Meggie886:
### Describe your proposal

<img width="1389" height="636" alt="Image" src="https://github.com/user-attachments/assets/562706ee-f03b-407f-b64d-d71b954eb3b7" />

### Proposed Solution

When a user navigates to the ‘Delete’ button in the “Actions” column of the table, the screen reader should announce only the relevant label and role of the button, for example: “Delete, button”. No additional or unrelated content such as “Empty” should be read.

### StackBlitz example(s)

https://stackblitz.com/edit/pm4e2ycz?file=src%2Fapp%2Fplatform-table-custom-width-example.component.html

### Priority

None

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

PR #14017, opened 2026-03-24
fix(platform): fdp-table should not read additional info "empty" when there is a button inside a table cell

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13992

## Description

Fixes screen reader incorrectly announcing "Empty" for table cells containing action buttons (icon-only buttons with accessible labels).

### Changes
**Table Row Component (table-row.component.ts)**
- Added **_hasAccessibleContent()** method that checks if a cell contains elements with aria-label or aria-labelledby attributes.
- Modified **_setAnnounceEmptyCell()** to skip the "Empty" announcement when the cell contains accessible interactive content.

**Platform Button Component (button.component.ts, button.component.html)**
- Added **_accessibleName** computed property with a fallback chain for accessible names:
  - ariaLabel input
  - title input (commonly used for icon-only buttons)
  - label input (visible text)
  - glyph name as last resort fallback
- Updated template to use **_accessibleName()** instead of just **ariaLabel()** for the inner button.

### Problem
When a table cell contained only an icon button (e.g., a delete button with title="delete" or glyph="decline"), the screen reader would announce "Empty" because:

1. The cell's **innerText** was empty (icons have no text content).
2. The **_setAnnounceEmptyCell** logic didn't account for elements with ARIA labels.

### Solution
1. Table cells now check for [aria-label] or [aria-labelledby] attributes before announcing "Empty".
2. Platform buttons now always compute an accessible name from available inputs, ensuring icon-only buttons have proper ARIA labels.

## Screenshots

Screen reader text output before the change:

Screenshot 2026-03-24 at 10 25 25

Screen reader text output after the change:

Screenshot 2026-03-24 at 10 24 43

### DIFF
--- a/libs/platform/button/button.component.ts
+++ b/libs/platform/button/button.component.ts
@@ -129,7 +129,7 @@ export class ButtonComponent implements ButtonModel {
      * 4. glyph name as fallback (e.g., "decline" -> "decline")
      * @hidden
      */
-    protected readonly _accessibleName = computed(
+    protected readonly accessibleName = computed(
         () => this.ariaLabel() || this.title() || this.label() || this.glyph() || null
     );
 
--- a/libs/platform/button/button.component.html
+++ b/libs/platform/button/button.component.html
@@ -3,7 +3,7 @@
     fd-button
     [attr.id]="id()"
     [attr.name]="name()"
-    [ariaLabel]="_accessibleName()"
+    [ariaLabel]="accessibleName()"
     [attr.aria-controls]="ariaControlsId()"
     [attr.aria-expanded]="ariaExpanded()"
     [attr.aria-disabled]="ariaDisabled()"


libs/platform/button/button.component.ts
```diff
@@ -120,6 +120,19 @@ export class ButtonComponent implements ButtonModel {
...
+     * Follows ARIA accessible name computation priority:
+     * 1. aria-label input
+     * 2. title input (commonly used for icon-only buttons)
+     * 3. label input (visible text)
+     * 4. glyph name as fallback (e.g., "decline" -> "decline")
+     * @hidden
+     */
+    protected readonly _accessibleName = computed(
+        () => this.ariaLabel() || this.title() || this.label() || this.glyph() || null
+    );
```
InnaAtanasova: This shouldn't be prefixed, because it's not private.

Issue #14001, opened 2026-03-19
Disabled Segmented Button receives touched event and gets an outline


InnaAtanasova:
Before interaction with the component
<img width="234" height="152" alt="Image" src="https://github.com/user-attachments/assets/09522114-3b3b-4abd-bacb-3b479caf2b51" />

After:

<img width="241" height="148" alt="Image" src="https://github.com/user-attachments/assets/b3b3ebe3-4fd0-4dd3-97e6-6ae4c045bf0d" />

PR #14026, opened 2026-03-24
fix(core): remove outline and event from disabled segmented button

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/14001

## Description
Fixed incorrect behavior in segmented button reactive forms where  disabled controls were responding to click events and triggering touch state changes.                                               

**TypeScript (segmented-button.component.ts)**                                       
  - Added early return check in `@HostListener('click')` to prevent `onTouched()` from being called when component is disabled
  - Removed buggy code block in `_toggleDisableButtons()` that was re-registering event listeners when buttons were disabled (lines 393-398). This code was redundantly setting `role="option"` (already set in `_setButtonAttributes()`)                    More critically, it was re-adding click/keyboard event listeners instead of removing them.                                 
                  
**CSS (segmented-button.component.scss)**                                      
  - Added rules to prevent :focus-visible outline on disabled segmented buttons

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.62.0-rc.22",
+    "version": "0.62.0-rc.24",
     "conventionalCommits": true,
     "tagVersionPrefix": "v",
     "message": "chore(release): publish %v",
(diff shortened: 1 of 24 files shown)


InnaAtanasova:
fix in fund-styles that brings this CSS there, too: https://github.com/SAP/fundamental-styles/pull/6262

Issue #13989, opened 2026-03-19
fdp-table-toolbar does not update well when modifying a title


adriapulido-boyum:
### Describe your proposal

There is an issue when I have to modifiy the title of the table toolbar. I have to update the title because of translations or other dynamic data I get from an API REST. However, when I update the title, the actions of the toolbar does not collapse, causing that some actions are hidden when the table is small. Or when the title is so long, I cannot see the toolbar actions

This works well when I have to not do an update to the title or when I'm modifying angular code (it refreshes automatically the web app an it looks ok)

I consider that this is a bug and the actions (at least when they're collapsed, with the 3 dots button) should appear always.

<img width="524" height="124" alt="Image" src="https://github.com/user-attachments/assets/eeb5dc21-66e7-45c0-a420-6a5125ccabae" />

### Which versions of Angular and Fundamental Library for Angular are affected?

0.61.1

### Affected Component

fdp-table-toolbar and related

### Steps to Reproduce (for bugs)

Add a logic that you can modify a title after rendering the table or add a very long title

### Priority

None

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

PR #14040, opened 2026-03-26
fix(platform,core): fdp-table-toolbar does not update well when modifying a title

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13989

### Main issues

- When a title is dynamically updated and takes wider space than intial render, some items move out of screen instead of getting in the overflow.
- If title is too long, it would never truncate, possibly take the whole screen and not leave space for anything else.

### Fixes

- Adds toolbar overflow recalculation when title content changes. 
- Implements text truncation with ellipsis for long toolbar titles to allow space for the overflow button (the three dots).
- Adds missing title and aria-label attributes to the toolbar title for accessibility and hover tooltip support.

### Detailed changes
**Core Toolbar (libs/core/toolbar/)**
- Added `ResizeObserver` on the title element to trigger toolbar overflow recalculation when title content/size changes (previously only observed the toolbar container, not the title itself).
- Added `title` and `aria-label` attributes to the title element for accessibility
- Added `.fd-toolbar__title--truncate` CSS class for text truncation with ellipsis.
- Unit tests added to cover the new logic.

**Platform Table Toolbar (libs/platform/table/components/table-toolbar/)**
- Added `title` and `aria-label` attributes to the custom titleTemplate span
- Implemented `MutationObserver` to extract text content from rendered title template and keep it in sync with DOM changes. This ensures the tooltip and screen reader text stay updated even when the title content changes dynamically.
- Unit tests added to cover the new logic.

## Screenshots

**Before the change:**
- Items get hidden when title becomes longer.
- Only after manually changing screen size, items move to the overflow.
- No tooltip on the title.

https://github.com/user-attachments/assets/44fec64b-e48d-4ba7-830b-9f1fe71b61c5

**After the change:**
- Items move to the overflow immediately when title becomes longer.
- In case title is too long, it gets truncated.
- A tooltip is shown upon hover.

https://github.com/user-attachments/assets/8ea7394b-9d11-4333-92b4-4d9f2731a155


MariaIDineva:
You can test the fix by adding similar changes (that simulate a dynamic title change) to a table example of your choice:
Screenshot 2026-03-26 at 16 00 26

Issue #14035, opened 2026-03-26
Remove NgZone usage across the project for Angular 21 zoneless compatibility


droshev:
## Problem

Angular 21 defaults to zoneless change detection. Code that relies on `NgZone` events like `onMicrotaskEmpty` and `onStable` **will silently break** — these observables never emit in zoneless mode.

We already hit this in the **Carousel component**, fixed in #14028: buttons and indicators were missing because `NgZone.onMicrotaskEmpty` never fired.

The same class of bug exists in **27 other files** across the project.

## High Risk: `onMicrotaskEmpty` / `onStable` subscribers (17 files)

These will **break in zoneless mode** — the subscriptions never fire, causing missing UI, broken initialization, or stale state.

| # | File | Usage | Line(s) | Assignee | PR |
|---|------|------|--------|---------|----|
| 1 | `libs/cdk/utils/directives/initial-focus/initial-focus.directive.ts` | `onMicrotaskEmpty.pipe(take(1))` | :74 | Inna | [#14044](https://github.com/SAP/fundamental-ngx/pull/14044)|
| 2 | `libs/cdk/utils/directives/disabled/fdk-disabled-provider.service.ts` | `firstValueFrom(ngZone.onStable)` | :70 | Maria | |
| 3 | `libs/cdk/utils/directives/readonly/fdk-readonly-provider.service.ts` | `firstValueFrom(ngZone.onStable)` | :75 | Maria | |
| 4 | `libs/core/tabs/tab-list.component.ts` | `onStable` + `onMicrotaskEmpty` | :379, :384 | Deno | #14042 |
| 5 | `libs/core/tabs/tab-panel/tab-panel.component.ts` | `onStable.pipe(first())` | :160 | Deno | #14042 |
| 6 | `libs/core/form/form-item/form-item.component.ts` | `onStable.pipe(first())` | :63 | Inna | [#14045](https://github.com/SAP/fundamental-ngx/pull/14045) |
| 7 | `libs/core/table/table.component.ts` | `onStable.pipe(first())` | :141 | Mike | |
| 8 | `libs/core/timeline/components/timeline-node-body/timeline-node-body.component.ts` | `onStable.pipe(first())` | :36 | Maria | |
| 9 | `libs/core/menu/directives/segmented-button/segmented-button-option.directive.ts` | `delayWhen(() => ngZone.onStable)` | :58 | Deno | #14043 |
[...]

PR #14045, opened 2026-03-26
fix(core): remove NgZone usage from Form Item

InnaAtanasova:
## Related Issue(s)

## Description
Because of this bug form item input was not connected to the label via aria-labelledby.

## Screenshots
**Before (missing aria-labelledby value):**
Screenshot 2026-03-26 at 12 33 57 PM

**After:**
Screenshot 2026-03-26 at 12 34 27 PM

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
     "$schema": "node_modules/lerna/schemas/lerna-schema.json",
     "packages": ["libs/*"],
-    "version": "0.62.0-rc.40",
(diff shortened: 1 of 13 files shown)


InnaAtanasova:
> I don't know if it is related and if it should be addressed in this PR, but it seems like (before and after the changes) `aria-label` is present on the input as an attribute but has no value.
> 
> Image

I fixed it in `libs/core/form/form-control/form-control.component.ts`

MariaIDineva (approved):

PR #14051, opened 2026-03-26
fix(platform): remove NgZone usage from icon-tab-bar for zoneless compatibility

droshev:
remove NgZone usage from icon-tab-bar for zoneless compatibility

### DIFF
--- a/libs/platform/icon-tab-bar/components/icon-tab-bar-process-type/icon-tab-bar-process-type.component.ts
+++ b/libs/platform/icon-tab-bar/components/icon-tab-bar-process-type/icon-tab-bar-process-type.component.ts
@@ -104,7 +104,7 @@ export class IconTabBarProcessTypeComponent extends ClosableIconTabBar {
         this._selectItem(selectedItem);
 
         queueMicrotask(() => {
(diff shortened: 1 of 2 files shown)


libs/platform/icon-tab-bar/components/icon-tab-bar-process-type/icon-tab-bar-process-type.component.ts
```diff
@@ -104,7 +103,7 @@ export class IconTabBarProcessTypeComponent extends ClosableIconTabBar {
         }
         this._selectItem(selectedItem);
 
-        this._ngZone.onMicrotaskEmpty.pipe(take(1)).subscribe(() => {
+        queueMicrotask(() => {
             if (this.overflowDirective) {
```
MariaIDineva: Claude hint: Maybe add a `&& !this._destroyed` check, similar to the base class so that we don't end up in a situation where `this._cd.detectChanges();` is called on a destroyed view.

droshev: updated

PR #14053, opened 2026-03-26
fix(core): remove NgZone usage from Segmented Button

InnaAtanasova:
## Related Issue(s)

## Description
`setDisabledState` is called by Angular's reactive forms infrastructure (via ControlValueAccessor), which already executes inside the zone. The `_ngZone.run()` wrapper was redundant — the code was never running outside the zone to begin with. Additionally, `detectChanges()` is called immediately after, which explicitly triggers change detection regardless of zone context. In zoneless mode (Angular 21 default), `NgZone.run()` is a no-op anyway since there's no zone to re-enter.

### DIFF
--- a/libs/core/segmented-button/segmented-button.component.ts
+++ b/libs/core/segmented-button/segmented-button.component.ts
@@ -10,7 +10,6 @@ import {
     Host,
     HostListener,
     Input,
-    NgZone,
     OnChanges,
     OnDestroy,
     QueryList,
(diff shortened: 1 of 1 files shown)


MariaIDineva (approved):

MariaIDineva:
Actualy, maybe we should talk about removing `this._zone.run` completely before merging this PR as per [Angular documentation ](https://angular.dev/guide/zoneless)"`removing these calls can lead to performance regressions for libraries that are used in applications that still rely on ZoneJS.`" Although is this case we have `this._changeDetRef.detectChanges();` right after, so we might be safe.

InnaAtanasova:
> Actualy, maybe we should talk about removing `this._zone.run` completely before merging this PR as per [Angular documentation ](https://angular.dev/guide/zoneless)"`removing these calls can lead to performance regressions for libraries that are used in applications that still rely on ZoneJS.`" Although is this case we have `this._changeDetRef.detectChanges();` right after, so we might be safe.

`_ngZone.run()` wrapper was redundant, the code was never running outside the zone to begin with. `detectChanges()` is called immediately after, which explicitly triggers change detection regardless of zone context.

PR #14054, opened 2026-03-26
fix(platform): remove NgZone usage in auto complete directive

InnaAtanasova:
## Related Issue(s)

## Description
Removed **NgZone** and the **runOutsideAngular** wrapper. The event listeners (keyup, compositionstart, compositionend) only do direct DOM manipulation (setting `input.value`, `setSelectionRange`) and update internal flags (`_isComposing`, `inputText`). None of these trigger Angular change detection, so there was no CD overhead to avoid. The `@Output() onComplete` emit (which does need CD) happens through `handleKeyboardEvent` → `_sendCompleteEvent`, which is covered by Angular's event binding mechanics.


MariaIDineva:
Something weird is happening, though, which does not seem to be related to the changes in this PR as it is reprodicble in main as well. Look how I type "A" and then spot how many times I type "p" initially and nothing happens. It seems like after the first keystroke the focus moves away and only when I focus back on the input field, I could start typing again and autocomplete starts working.

https://github.com/user-attachments/assets/e6046f1b-4c8e-4524-93c0-fedb8a36e47c

Maybe I should log a separate issue for that, what do you think?

MariaIDineva:
Aand maybe we should talk about removing `this._zone.runOutsideAngular` completely before merging this PR as per [Angular documentation ](https://angular.dev/guide/zoneless)"`removing these calls can lead to performance regressions for libraries that are used in applications that still rely on ZoneJS.`"

PR #14055, opened 2026-03-26
fix(core): add fix for flickering Inline Help

InnaAtanasova:
## Description
**Problem**
When hovering near the edge of an inline-help trigger, the popover rapidly opens and closes (flickering). This happens because mouseleave fires on the trigger as the cursor moves into the popover body, causing an immediate close() → overlay disposed → cursor back over trigger → mouseenter → open() → loop. During this rapid cycle, the popover body also renders without its CSS classes (missing padding, arrow styles) because the CDK position change event never fires before the overlay is disposed.

https://github.com/user-attachments/assets/99e1070d-9aa4-4ffb-9d0d-86a5f1a94fe5

**Fix**

Introduced a 50ms delayed close for hover triggers (mouseenter/mouseleave), giving the cursor time to travel from the trigger into the popover body.
Added mouseenter/mouseleave listeners on the CDK overlay element itself — entering the overlay cancels the pending close, leaving it schedules a close.
All other trigger types (click, focusin/focusout, etc.) remain unaffected and continue to toggle immediately.

### DIFF
--- a/libs/core/popover/popover-service/popover.service.ts
+++ b/libs/core/popover/popover-service/popover.service.ts
@@ -172,6 +172,12 @@ export class PopoverService {
     /** @hidden */
     private _ignoreTriggers = false;
 
+    /** @hidden Timer for delayed hover close, allows mouse to travel from trigger to overlay body. */
+    private _hoverCloseTimer: ReturnType<typeof setTimeout> | null = null;
+
+    /** @hidden Listeners for mouseenter/mouseleave on the overlay element. */
+    private _overlayHoverListeners: (() => void)[] = [];
+
     /** @hidden */
     private _modalBodyClass = 'fd-overlay-active';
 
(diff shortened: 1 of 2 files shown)


libs/core/popover/popover-service/popover.service.ts
```diff
@@ -542,15 +554,30 @@ export class PopoverService {
...
                         const closeAction = !!trigger.closeAction;
                         const openAction = !!trigger.openAction;
-                        this.toggle(openAction, closeAction);
+
+                        // For hover triggers, use delayed close to prevent flickering
+                        // when the mouse travels between the trigger and the popover body.
+                        if (hasHoverTrigger && trigger.trigger === 'mouseenter' && openAction) {
+                            this._clearHoverCloseTimer();
+                            if (!this.isOpen()) {
+                                this.toggle(true, false);
```
MariaIDineva: Shall we use openAction and closeAction instead of hardcoded booleans here?

InnaAtanasova: yes, makes sense. Will fix it!

Issue #14064, opened 2026-03-29
DataSourceDirective provider emits empty data before consumers can configure it


droshev:
## Problem

When `DataSourceDirective` creates a new data source provider (via its `dataSource` setter), two things happen in sequence:

1. The new provider is created with a `BehaviorSubject<[]>` — which immediately emits an empty array to any subscriber
2. The `dataSourceChanged` event is emitted, telling consumers "hey, there's a new provider — configure it"

The problem is that step 1 happens *before* step 2. So any component subscribed to the data stream receives an empty `[]` emission before it even has a chance to configure the new provider with matching settings and trigger the real data fetch.

## Why this matters

Components that react to empty data (e.g., showing "no results" states, reverting input text, marking entries as invalid) will incorrectly trigger those behaviors during what is actually just a normal data source swap. The empty emission isn't real — it's just the `BehaviorSubject`'s initial value before `match('*')` populates it with actual data.

## Reproduction

You don't need a full component to see this. Here's what happens with just `DataSourceDirective` and a subscriber:

```typescript
// Simplified consumer that uses DataSourceDirective as a host directive
@Component({
    selector: 'my-search',
    hostDirectives: [{ directive: DataSourceDirective, inputs: ['dataSource'] }]
})
class MySearchComponent implements OnInit {
    private dataSourceDirective = inject(DataSourceDirective);
    private log: string[] = [];

    ngOnInit() {
        // Listen for data source swaps
        this.dataSourceDirective.dataSourceChanged.subscribe(() => {
            this.log.push('dataSourceChanged — new provider, configuring...');
            // Configure the new provider with matching settings
            this.dataSourceDirective.dataSourceProvider.dataProvider.setMatchingBy({ firstBy: displayFn });
            // Trigger initial data fetch
            this.dataSourceDirective.dataSourceProvider.match(new Map([['query', '*']]));
        });

        // Listen for data arriving
        this.dataSourceDirective.dataChanged$.subscribe((data) => {
            this.log.push(`dataChanged$ — received ${data.length} items`);
            if (data.length === 0) {
                this.log.push('  → showing "no results" state ← THIS IS THE BUG');
            }
        });
    }
}
```

When the user switches the data source at runtime (e.g., clicking a "Load fruits" button), the log shows:

```
dataChanged$ — received 0 items        ← spurious [] from new BehaviorSubject
  → showing "no results" state          ← wrong! data hasn't loaded yet
dataSourceChanged — new provider, configuring...
[...]

PR #14065, opened 2026-03-29
fix(core): reconfigure multi-combobox data provider when dataSource changes

### LINKED ISSUE #14061: multi box search stop working if the data changed
### Describe your proposal

if the dataSource of the `fd-multi-combobox` value changed, the search stop working

0.61.1 and 21.1.0

### Affected Component

_No response_

1. open the `stackblitz` link
2. try to search (it doesn't work)
3. comment the setTimeout
4. try to search (now it works)

### StackBlitz example(s)

https://stackblitz.com/edit/qwlvjyfj?file=src%2Fapp%2Fmulti-combobox-datasource-example.component.ts

### Priority

High

### Declaration

droshev:
reconfigure multi-combobox data provider when dataSource changes

fixes #14061

The bug: When you swap the data source of fd-multi-combobox at runtime (e.g., loading new data from a server), search stops working. Typing in the search field returns no results because the new data source provider doesn't know how to match items.

When the data source changes, DataSourceDirective creates a brand new provider. But the multi-combobox only configured the provider once — during initialization. We cannot match items or fetch the initial data. 

Fix: listen for data source changes and, each time one happens, we:

  1. Reconfigure the new provider with the same matching settings (display function, lookup key, matching strategy)
  2. Fetch the initial data (match('*'))
  3. Reconnect the loading/data-received event forwarding so those events come from the new provider, not the old one
  4. Skip the false "empty data" emission that the new provider fires before real data loads (a BehaviorSubject starts with []) — without this guard, the component would briefly flash a "no results" state.

things not covered:

  1. Timing gap on async data sources — DataSourceDirective emits data before dataSourceChanged, so the guard
   doesn't fully protect Observable/HTTP sources. Benign for arrays (the reported bug). See #14064.
  2. skip(0) no-op — out of scope, separate cleanup.
  3. No async data source test — requires significant test infrastructure; timing gap documented in #14064.

### DIFF
--- a/libs/core/multi-combobox/base-multi-combobox.class.ts
+++ b/libs/core/multi-combobox/base-multi-combobox.class.ts
@@ -663,6 +663,10 @@ export abstract class BaseMultiCombobox<T = any> {
      * Prepares the data stream and subscribes to it.
      */
     protected _openDataStream(matchingStrategy: MatchingStrategy): void {
+        if (!this.dataSourceDirective.dataSourceProvider) {
+            throw new Error(`[dataSource] source did not match an array, Observable, or DataSource`);
+        }
+
         this._matchingStrategy = matchingStrategy;
         this._configureDataSourceProvider(matchingStrategy);
 


libs/core/multi-combobox/base-multi-combobox.class.ts
```diff
@@ -357,10 +357,11 @@ export abstract class BaseMultiCombobox<T = any> {
 
     /**
      * Flag indicating whether the data source has changed.
-     * Set to true when a new data source is assigned, triggering a refresh of suggestions.
+     * Guard flag: true while waiting for a new data source's first real data emission.
+     * Prevents the provider's initial empty emission from being treated as "no results."
      * @hidden
      */
-    private _dataSourceChanged = false;
+    private _dataSourceChanged = true;
```
MariaIDineva, line 364: Shouldn't that be "false" and set to "true" only when the data source changes?

libs/core/multi-combobox/base-multi-combobox.class.ts
```diff
@@ -795,4 +770,74 @@ export abstract class BaseMultiCombobox<T = any> {
...
+
+    /**
+     * @hidden
+     * Configures the data source provider with matching settings.
+     */
+    private _configureDataSourceProvider(matchingStrategy: MatchingStrategy): void {
+        const dataSourceProvider = this.dataSourceDirective.dataSourceProvider;
+
+        if (!dataSourceProvider) {
+            return;
```
MariaIDineva, line 786: There used to be an error:
```
throw new Error(`[dataSource] source did not match an array, Observable, or DataSource`);
```
 and now we return silently. Is that intentional?

libs/core/multi-combobox/base-multi-combobox.class.ts
```diff
@@ -662,60 +663,34 @@ export abstract class BaseMultiCombobox<T = any> {
...
-                this._dataSourceChanged = true;
-            });
+        this.dataSourceDirective.dataSourceChanged.pipe(takeUntilDestroyed(this._destroyRef)).subscribe(() => {
+            this._dataSourceChanged = true;
+            this._subscribeToDsEvents();
+            this._configureDataSourceProvider(this._matchingStrategy);
+            this._fetchInitialData();
+        });
 
         this.dataSourceDirective.dataChanged$.pipe(skip(0), takeUntilDestroyed(this._destroyRef)).subscribe((data) => {
```
MariaIDineva, line 685: I know it was not added in this PR, but `skip(0)` seems redundant and maybe we should remove it.

libs/core/multi-combobox/base-multi-combobox.class.ts
droshev: The old code used `false` + `startWith(true)`, so the flag was effectively true before any data arrived. This PR just inlines that `true` directly, same net result. It needs to start as true because `_fetchInitialData()` triggers the provider's `BehaviorSubject` to emit [] before real data loads — the guard skips that empty emission so we don't flash "no results."

libs/core/multi-combobox/base-multi-combobox.class.ts
droshev: that's why i wrote it in things not covered.

libs/core/multi-combobox/base-multi-combobox.class.ts
droshev: Now `_openDataStream` throws on misconfiguration at init, while `_configureDataSourceProvider` still silently returns during transient null states on data source swaps.

PR #14066, opened 2026-03-29
fix(core): bring back more inline help inputs

droshev:
bring back more inline help inputs


libs/core/inline-help/inline-help.directive.spec.ts
```diff
@@ -38,42 +69,119 @@ describe('InlineHelpDirective', () => {
         expect(component).toBeTruthy();
     });
 
-    it('should show the inline help on hover', () => {
+    it('should show the inline help on hover', fakeAsync(() => {
```
MariaIDineva, line 72: I don't think all those `fakeAsync/tick` calls are needed because they don't test any timer-based logic. Also, tests pass without them.

libs/core/inline-help/inline-help.directive.ts
```diff
@@ -71,6 +71,27 @@ export class InlineHelpDirective {
...
+    readonly closeOnEscapeKey = input(false, { transform: booleanAttribute });
+
+    /** Whether to close the popover on router navigation start. */
+    readonly closeOnNavigation = input(true, { transform: booleanAttribute });
+
+    /** Whether to restore focus to the previously focused element when the popover closes. */
+    readonly restoreFocusOnClose = input(true, { transform: booleanAttribute });
+
+    /** The element to which the overlay is attached. By default it is body. */
+    readonly appendTo = input<Nullable<ElementRef | Element>>(null);
```
InnaAtanasova: Didn't we decide to get rid of Nullable?

libs/core/inline-help/inline-help.directive.ts
```diff
@@ -71,6 +71,27 @@ export class InlineHelpDirective {
...
+
+    /** Whether position should remain fixed when approaching page corners. */
+    readonly fixedPosition = input(false, { transform: booleanAttribute });
+
+    /** Maximum width of popover body in px. */
+    readonly maxWidth = input<Nullable<number>>(null);
```
InnaAtanasova: readonly maxWidth = input<number | null | undefined>(null);

libs/core/inline-help/inline-help.directive.ts
```diff
@@ -71,6 +71,27 @@ export class InlineHelpDirective {
     /** aria-role for the Inline Help Popover body */
     readonly bodyRole = input('tooltip');
 
+    /** Whether the popover should have an arrow. */
+    readonly noArrow = input(false, { transform: booleanAttribute });
+
+    /** Whether the popover should close when the escape key is pressed. */
+    readonly closeOnEscapeKey = input(false, { transform: booleanAttribute });
```
mikerodonnell89, line 78: Testing these, this one does not appear to work

libs/core/inline-help/inline-help.directive.spec.ts
droshev: The fakeAsync/tick(50) is needed — PopoverService.[_scheduleHoverClose](https://github.com/SAP/fundamental-ngx/blob/62f27705ebcac1d69e583a718650d5e87e9e7697/libs/core/popover/popover-service/popover.service.ts#L647)() uses a 50ms setTimeout to debounce close on mouseleave/focusout. Without tick, assertions run before the timer fires, making tests flaky.

libs/core/inline-help/inline-help.directive.ts
droshev: updates

libs/core/inline-help/inline-help.directive.ts
droshev: updated

mikerodonnell89 (approved):

Issue #14024, opened 2026-03-24
Type 'MenuInterface' does not satisfy the constraint 'MobileMode'


LukasHeimann:
### Describe your proposal

```
  Error: R] TS2344: Type 'MenuInterface' does not satisfy the constraint 'MobileMode'.
    Types of property 'mobile' are incompatible.
      Type 'InputSignalWithTransform<boolean, unknown>' is not assignable to type 'boolean'. [plugin angular-compiler]
  
      node_modules/@fundamental-ngx/core/types/fundamental-ngx-core-menu.d.ts:685:57:
        685 │ ...mponent extends MobileModeBase<MenuInterface> implements OnInit {
            ╵                                   ~~~~~~~~~~~~~
```

This is my tsconfig:

```json
{
  "compilerOptions": {
    "forceConsistentCasingInFileNames": true,
    "esModuleInterop": true,
    "sourceMap": true,
    "declaration": false,
    "experimentalDecorators": true,
    "moduleResolution": "bundler",
    "importHelpers": true,
    "target": "ES2022",
    "module": "ES2022"
  },
  "angularCompilerOptions": {
    "strictInjectionParameters": true,
    "strictInputAccessModifiers": true,
    "strictTemplates": true,
    "strictStandalone": true
  }
}
```

If I add `skipLibCheck` to my `angularCompilerOptions`, I don't get the error anymore, but I feel like I should not be doing that...

### Which versions of Angular and Fundamental Library for Angular are affected?

Angular 21.2.4, Fundamentals 0.61.1

### Affected Component

MenuInterface

### Priority

None

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

InnaAtanasova:
Hi @LukasHeimann , can you test with `0.61.4`? We released a fix for this issue. Let us know if you are still experiencing problems.

PR #14069, opened 2026-03-30
fix(core): support both signal and plain property types in MobileMode interface

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/14024

## Description
**Problem**                                                                     
Users upgrading to 0.61.1+ with strict TypeScript settings encounter a compilation error:                                                            
                                                                                
`Type 'MenuInterface' does not satisfy the constraint 'MobileMode'. Types of property 'mobile' are incompatible. Type 'InputSignalWithTransform<boolean, unknown>' is not assignable to type 'boolean'.`                                                               
                                                                                
The `MobileMode` interface requires mobile: boolean, but components migrating to signals use `InputSignalWithTransform<boolean, unknown>`, creating a type incompatibility. Users had to add `skipLibCheck` to work around this.           
                                                                              
**Solution**
Updated MobileMode interface to accept both plain properties and signal functions:
```                                                       
  export interface MobileMode {                                               
      readonly mobile: boolean | (() => boolean);                               
      readonly mobileConfig: MobileModeConfig | (() => MobileModeConfig);       
  }                                                                             
```
                                                                            
Added runtime type checks in SelectKeyManagerService to handle both patterns  when accessing the mobile property.                                         This maintains backward compatibility with existing components using plain properties while supporting new signal-based implementations.


libs/core/select/select-key-manager.service.ts
```diff
@@ -121,8 +123,12 @@ export class SelectKeyManagerService {
             event.preventDefault();
             manager.activeItem._selectViaInteraction();
             this._component.blur();
-        } else if (!isTyping && KeyUtil.isKeyCode(event, [TAB]) && this._component.mobile) {
-            event.preventDefault();
+        } else if (!isTyping && KeyUtil.isKeyCode(event, [TAB])) {
+            const isMobile =
```
MariaIDineva: This logic could be extracted into a reusable function.

Issue #14037, opened 2026-03-26
Audit setTimeout/setInterval usage for zoneless compatibility


droshev:
## Problem

Related to #14035 (NgZone removal tracking).

In Angular 21, `zone.js` is no longer included by default. One side effect: `setTimeout` and `setInterval` are no longer monkey-patched by zone.js, which means they **no longer trigger change detection**.

Any `setTimeout`/`setInterval` callback that mutates template-bound state will silently stop updating the UI in zoneless mode. Unlike `NgZone.onStable` (which clearly breaks), these are harder to catch because the code still *runs* — it just doesn't trigger rendering.

## Scope

**85 occurrences across 50 files** in the project use `setTimeout` or `setInterval`.

Not all are problematic — only those that update state visible in templates. Each call site needs to be individually audited.

## What to look for

A `setTimeout`/`setInterval` call is **safe** if it:
- Only performs DOM manipulation (e.g., focus, scroll)
- Only calls external APIs with no template-bound side effects
- Is already followed by `markForCheck()` or `detectChanges()`

A `setTimeout`/`setInterval` call **needs fixing** if it:
- Sets a component property bound in the template
- Modifies a signal or observable that feeds the template
- Mutates shared state without notifying Angular's change detection

## Suggested fixes

| Pattern | Fix |
|---------|-----|
| `setTimeout(() => { this.visible = true; })` | Add `this._cdr.markForCheck()` after mutation, or convert to a signal |
| `setTimeout(() => el.focus(), 0)` | Safe as-is (DOM-only, no CD needed) |
| `setInterval(() => { this.counter++; }, 1000)` | Use `signal` + `effect` or add `markForCheck()` in the callback |

## References

- Parent issue: #14035
- Angular zoneless guide: https://angular.dev/guide/experimental/zoneless

droshev:
## Audit Results

### P1: `libs/cdk` — All safe, no changes needed

All 3 call sites audited — DOM-only, signal writes, or consumer-handled CD.

### P2: `libs/core` — 4 components need fixes (out of ~35 call sites across 25 files)

#### NEEDS FIX

| # | Component | File | Problem | Fix |
|---|-----------|------|---------|-----|
| 1 | `CarouselItemComponent._visibility` | `carousel-item.component.ts:89` | Plain property with `@HostBinding('style.visibility')`. Parent's `markForCheck()` doesn't dirty the child in zoneless mode. | Convert `_visibility` to `signal`, move to `host: {}` |
[...]

PR #12521, opened 2024-10-07
fix(core): remove OnPush from dialog container

mikerodonnell89:
fixes #12471

PR #14071, opened 2026-03-31
fix(core,platform,btp,cx): migrate remaining Default strategy components to OnPush

### LINKED ISSUE #14039: Audit ChangeDetectionStrategy.Default components for zoneless compatibility
## Problem

Related to #14035 (NgZone removal tracking).

Components using `ChangeDetectionStrategy.Default` rely on zone.js to trigger change detection on every async event (HTTP calls, timers, DOM events). In zoneless Angular 21, these components **won't update automatically** — they need either `OnPush` with explicit `markForCheck()` calls, or signals.

## Scope

**15 files** currently use `ChangeDetectionStrategy.Default` (explicitly set or inherited):

These were found by searching for explicit `ChangeDetectionStrategy.Default` declarations. Additionally, any component that doesn't declare a change detection strategy at all inherits `Default` — those are harder to enumerate but equally affected.

## What to look for

A `Default` strategy component is **safe** if:
- All its template bindings come from signals or observables with `async` pipe
- It has no async state mutations (everything is synchronous input-driven)

A `Default` strategy component **needs fixing** if:
- It mutates class properties in async callbacks (HTTP, timers, event listeners) that are bound in the template
- It relies on parent component CD runs to propagate changes downward

## Suggested fix

For most components in this project, the fix is straightforward:
1. Change to `ChangeDetectionStrategy.OnPush`
2. Add `markForCheck()` where needed, or migrate to signals

This is lower priority than the NgZone and setTimeout audits, since `Default` components will still *mostly* work via parent CD propagation — but they'll have subtle, hard-to-reproduce bugs.

## References

- Parent issue: #14035
- setTimeout/setInterval audit: #14037
- Angular zoneless guide: https://angular.dev/guide/experimental/zoneless

droshev:
migrate remaining Default strategy components to OnPush

fixes #14039

Summary

  Audit and migrate all remaining ChangeDetectionStrategy.Default components to OnPush for zoneless
  compatibility (#14039).

  Changes across 52 files (48 components, 3 specs, 1 base class):

  Core changes (applied to all 48 components):
  - Add ChangeDetectionStrategy.OnPush
  - Replace detectChanges() with markForCheck()
  - Remove standalone: true (default since Angular 19)

  FilterCustomComponent (libs/platform/table/.../filter-custom.component.ts):
  - Migrate from ngDoCheck polling to afterEveryRender — the only component that needed special handling
  because it detects mutations on a plain object passed by reference into a user template
  - Add markForCheck() after emission for OnPush parent notification
  - New spec file with 20 tests covering mutation detection, circular references, filterBy reset, template
  integration, and content density

  Icon Tab Bar (libs/platform/icon-tab-bar/):
  - Remove NgZone.onMicrotaskEmpty dependency from base class and process-type component
  - Replace with afterNextRender for zoneless compatibility
  - Preserve async _selectExtraItem(): Promise<void> signature (no breaking change)

  Affected libraries: core, platform, cdk, btp, cx

  Out of scope (intentionally not changed):

  - setTimeout() → afterNextRender() replacement (tracked in #14037)
  - @HostBinding / @HostListener → host: {} migration (future ticket)
### Review findings addressed

  | Concern | Analysis |
  |---------|----------|
  | OnPush on dialog/message-box containers may regress #12521 (dynamic content not updating) | Low risk. `_attached()` now calls `markForCheck()` on portal content. Child internal state changes are governed by the child's own CD, not the container's. |
  | Redundant `markForCheck()` after `signal.set()` on `@HostBinding` | Not redundant. `@HostBinding` getter is not a signal-reactive context — `markForCheck()` is needed. The CLAUDE.md rule applies to `host: {}` and template bindings only. |
  | Missing `markForCheck()` in `list-navigation-item` after mutating `@HostBinding` props | Not needed. `_handleExpandedChanges()` is only called from `@HostListener` handlers — DOM events automatically trigger CD on OnPush components. |
  | `afterEveryRender` detects mutations one frame later than `ngDoCheck` | Valid but inconsequential. Filter dialog requires user to click "OK" to apply — one-frame delay is imperceptible. |

### DIFF
--- a/libs/btp/search-field/src/search-field.component.ts
+++ b/libs/btp/search-field/src/search-field.component.ts
@@ -1,5 +1,7 @@
 import {
     AfterViewInit,
+    ChangeDetectionStrategy,
+    ChangeDetectorRef,
     Component,
     DestroyRef,
     ElementRef,
@@ -28,7 +30,8 @@ import { of, tap } from 'rxjs';
     hostDirectives: [CvaDirective],
     providers: [CvaControl],
     imports: [IconComponent, FormsModule, FdTranslatePipe, ButtonComponent, NestedButtonDirective],
-    encapsulation: ViewEncapsulation.None
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class SearchFieldComponent implements AfterViewInit, HasElementRef {
     /** Placeholder for the input field. */
@@ -60,6 +63,9 @@ export class SearchFieldComponent implements AfterViewInit, HasElementRef {
     private _destroyRef = inject(DestroyRef);
 
     /** @hidden */
+    private _cdr = inject(ChangeDetectorRef);
+
+    /** @hidden */
     focus(): void {
         this._searchInputField?.nativeElement.focus();
     }
@@ -68,7 +74,10 @@ export class SearchFieldComponent implements AfterViewInit, HasElementRef {
         this._cvaControl.listenToChanges();
         (this._cvaControl.cvaDirective?.ngControl?.valueChanges || of(undefined))
             .pipe(
-                tap((value: string) => (this._value = value)),
+                tap((value: string) => {
+                    this._value = value;
+                    this._cdr.markForCheck();
+                }),
                 takeUntilDestroyed(this._destroyRef)
             )
             .subscribe();
--- a/libs/core/side-navigation/side-navigation.component.ts
+++ b/libs/core/side-navigation/side-navigation.component.ts
@@ -1,16 +1,21 @@
 import {
     AfterContentInit,
     AfterViewInit,
+    ChangeDetectionStrategy,
+    ChangeDetectorRef,
     Component,
     ContentChild,
+    DestroyRef,
     HostBinding,
     HostListener,
+    inject,
     Input,
     OnInit,
     QueryList,
     ViewChildren,
     ViewEncapsulation
 } from '@angular/core';
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { Nullable, warnOnce } from '@fundamental-ngx/cdk/utils';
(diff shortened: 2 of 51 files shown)


libs/btp/search-field/src/search-field.component.ts
```diff
@@ -68,7 +74,10 @@ export class SearchFieldComponent implements AfterViewInit, HasElementRef {
         this._cvaControl.listenToChanges();
         (this._cvaControl.cvaDirective?.ngControl?.valueChanges || of(undefined))
             .pipe(
-                tap((value: string) => (this._value = value)),
+                tap((value: string) => {
+                    this._value = value;
+                    this._cdr.markForCheck();
```
mikerodonnell89: I think this component's `focus()` function needs markForCheck too

libs/platform/value-help-dialog/components/base-tab/vhd-base-tab.component.ts
```diff
@@ -1,12 +1,12 @@
...
 
 import { VhdTab } from '../../models';
 
 @Component({
     template: '',
-    standalone: true
+    changeDetection: ChangeDetectionStrategy.OnPush
 })
-// eslint-disable-next-line @angular-eslint/component-class-suffix
+
```
MariaIDineva: Redundant blank line

libs/core/side-navigation/side-navigation.component.ts
```diff
@@ -76,12 +82,18 @@ export class SideNavigationComponent implements AfterContentInit, AfterViewInit,
...
 
+    /** @hidden */
+    private readonly _cdr = inject(ChangeDetectorRef);
+
+    /** @hidden */
+    private readonly _destroyRef = inject(DestroyRef);
+
     /** @hidden */
     constructor(
         private keyboardService: NestedListKeyboardService,
```
MariaIDineva, line 93: Maybe inject those outside the constructor for consistency.

PR #14077, opened 2026-04-01
fix(core): remove Angular animation dependncy

### LINKED ISSUE #13091: move to css animation
it seems that Angular is moving away from animation, since new projects in Angular 19.2 won't have it installed by default and Angular material migrated all of their components to pure css animations and dropped the animations as a dependency 
looking at the code it seems there are only 4 animations used in 5 components so it should be feasible to do the same here
of course it's not something urgent since it's not officially deprecated and there isn't an official eof date but it should be planned

droshev:
## Summary

closes #13091

- Remove `@angular/animations` dependency from `@fundamental-ngx/core`
- Replace all Angular animation triggers with CSS transitions, Web Animations API, or Angular 21's `animate.enter`/`animate.leave`
- Redesign docs example cards with density toggle, responsive preview, and keyboard shortcut hints
- Remove dead code: `DirectionalityComponent`, `ExampleBackgroundComponent`, `CopyService`, collapse animations
- Remove `add-animations` schematic (no longer needed)

## Animation Migration Rationale

This PR removes `@angular/animations` and replaces all animation triggers with either CSS transitions, the Web Animations API, or Angular 21's `animate.enter`/`animate.leave`.

### Why not use `animate.enter`/`animate.leave` everywhere?

Angular 21's native animation API (`animate.enter`/`animate.leave`) works when elements are inserted/removed via template control flow (`@if`/`@for`). We verified each animated component against this requirement:

| Component | Creation method | Can use `animate.enter`/`animate.leave`? | Replacement |
|-----------|----------------|------------------------------------------|-------------|
| `DialogContainerComponent` | Programmatic via CDK Overlay/Portal | No — no template hooks available | Web Animations API |
| `MessageBoxContainerComponent` | Programmatic via CDK Overlay/Portal | No — same as dialog | Web Animations API |
| `BaseToastAnimatedContainerComponent` | Programmatic via CDK Overlay/Portal | No — also needs observable-based completion events | Web Animations API |
| `MessageViewComponent` | Always in DOM (list/details slide) | No — not a DOM insertion/removal scenario | Web Animations API |
| `CodeExampleComponent` | Template-driven `@if (isOpen())` | **Yes** | `animate.enter`/`animate.leave` |

### Dialog, Message-box, Toast — Web Animations API

These components are created programmatically by their respective services via `ComponentPortal` + `OverlayRef.attach()`. Since they're never inserted via `@if` in a template, Angular's `animate.enter`/`animate.leave` cannot bind to them. They also require:

- **Imperative start/cancel**: `_enter()` and `_exit()` are called by the service at specific lifecycle points
- **Completion callbacks**: Animation `.finished` promise drives state transitions (`_animationStateSignal`) and observable emissions (`onEnter$`/`onExit$`)
- **Cancellation on destroy**: Stored `Animation` references are cancelled in `DestroyRef.onDestroy()`

The Web Animations API (`element.animate()`) provides the imperative control these components need while staying zoneless-compatible and avoiding `@angular/animations`.

### Message-view — Web Animations API

The list ↔ details screen transition animates two sections that are always present in the DOM — it's a coordinated slide/fade, not an insertion/removal. `animate.enter`/`animate.leave` doesn't apply to this pattern.

### Code-example — `animate.enter`/`animate.leave`

The code panel expand/collapse is template-driven (`@if (isOpen())`), making it the right candidate for Angular 21's native API. The previous pure-CSS approach had a timing issue: `@if` removed DOM immediately on close, so the content fade-out never completed. With `animate.leave`, Angular now waits for the CSS animation to finish before removing the element.

```html
<div class="docs-code-example-content"
     animate.enter="docs-code-enter"
     animate.leave="docs-code-leave">
```

This is the first adoption of `animate.enter`/`animate.leave` in the codebase.

## Docs Example Redesign

- **Component-example card**: Signal-based state, always-visible action strip (RTL, BG, compact density, responsive preview with device presets), SAP design tokens throughout
- **Density toggle**: Uses `hostDirectives: [ContentDensityDirective]` + `ContentDensityObserver` with `restrictChildContentDensity: true` to propagate compact/cozy to all child components
- **Responsive preview**: Inline device presets (phone 375px, tablet 768px, full), CSS `resize: horizontal` drag handle, dot-grid canvas background
- **Keyboard shortcut hints**: `<kbd>` badge on Copy button showing ⌘C (Mac) or Ctrl C (other)
- **Dead code removed**: `DirectionalityComponent`, `ExampleBackgroundComponent`, `CopyService`, per-snippet copy buttons

## Breaking Changes

`@angular/animations` is no longer required by `@fundamental-ngx/core`. The `add-animations` schematic has been removed.

## Test plan

- [ ] Dialog open/close fade animation works
- [ ] Message-box open/close fade animation works
- [ ] Toast slide-in/fade-out animation works
- [ ] Message-popover list ↔ details slide transition works


apps/docs/src/styles.scss
```diff
@@ -27,28 +27,28 @@ body {
...
 
 .container {
-    max-width: 960px;
+    max-width: 60rem;
     margin-left: auto;
     margin-right: auto;
-    padding-left: 15px;
-    padding-right: 15px;
+    padding-left: 1rem;
+    padding-right: 1rem;
```
InnaAtanasova, line 34: both should be in one rule: `padding-inline: 1rem`. 
And you can change the margins to `margin-inline: auto`

apps/docs/src/styles.scss
```diff
@@ -27,28 +27,28 @@ body {
...
 }
 
 .row {
     display: flex;
-    margin-right: -15px;
-    margin-left: -15px;
+    margin-right: -1rem;
+    margin-left: -1rem;
```
InnaAtanasova, line 40: `margin-inline: -1rem`

apps/docs/src/styles.scss
```diff
@@ -27,28 +27,28 @@ body {
...
 }
 
 .col {
-    padding-right: 15px;
-    padding-left: 15px;
+    padding-right: 1rem;
+    padding-left: 1rem;
```
InnaAtanasova, line 45: `padding-inline: 1rem`

apps/docs/src/styles.scss
```diff
@@ -27,28 +27,28 @@ body {
...
     flex: 1;
 }
 
 .navbar {
-    background-color: #eee;
-    margin-bottom: 20px;
+    background-color: var(--sapPageHeader_Background);
+    margin-bottom: 1.25rem;
```
InnaAtanasova, line 51: `margin-block-end: 1.25rem`

apps/docs/src/styles.scss
```diff
@@ -87,15 +87,15 @@ fd-card {
 .fd-scroll-spy-example-icon-class {
     position: relative;
     display: inline-block;
-    margin-right: 4px;
+    margin-right: 0.25rem;
```
InnaAtanasova, line 90: `margin-inline-end: 0.25rem`

apps/docs/src/styles.scss
```diff
@@ -87,15 +87,15 @@ fd-card {
...
 
     @at-root {
         [dir='rtl'] &,
         &[dir='rtl'] {
             margin-right: 0;
-            margin-left: 4px;
+            margin-left: 0.25rem;
```
InnaAtanasova, line 96: then this won't be needed

apps/docs/src/styles.scss
```diff
@@ -188,12 +180,12 @@ pre {
 
 .fd-playground--markdown {
     #fundamental-ngx {
-        margin-top: calc(2rem + 10px);
+        margin-top: calc(2rem + 0.625rem);
```
InnaAtanasova, line 183: `margin-block-start`

libs/docs/shared/src/lib/core-helpers/code-example/code-example.component.scss
```diff
@@ -1,10 +1,75 @@
...
+    // Bottom fade gradient — hints at scrollable content
+    &::after {
+        content: '';
+        position: sticky;
+        bottom: 0;
+        left: 0;
+        right: 0;
+        display: block;
+        height: 2rem;
+        margin-top: -2rem;
```
InnaAtanasova, line 64: `margin-block-start: -2rem;`

libs/docs/shared/src/lib/core-helpers/code-example/code-example.component.scss
```diff
@@ -13,26 +78,49 @@
...
+    text-align: center;
+    padding: var(--sapContent_Space_S);
+}
+
 code-example {
+    display: block;
+
     ul.fd-tabs {
         margin-bottom: 0;
         margin-block-end: 0;
```
InnaAtanasova, line 107: these 2 are basically the same

apps/docs/src/styles.scss
droshev: addressed

apps/docs/src/styles.scss
droshev: addressed

apps/docs/src/styles.scss
droshev: addressed

apps/docs/src/styles.scss
droshev: addressed

apps/docs/src/styles.scss
droshev: addressed

apps/docs/src/styles.scss
droshev: addressed

apps/docs/src/styles.scss
droshev: addressed

libs/docs/shared/src/lib/core-helpers/code-example/code-example.component.scss
droshev: addressed

libs/docs/shared/src/lib/core-helpers/code-example/code-example.component.scss
droshev: addressed

droshev:
> @droshev said the comments will be addressed in a separate PR for the documentation.

@InnaAtanasova I addressed all the comments in the PR #14084

PR #14084, opened 2026-04-03
fix(ci): harden workflow security against injection and over-permissioning

droshev:
harden workflow security against injection and over-permissioning


.github/workflows/create-release.yml
```diff
@@ -6,9 +6,6 @@ on:
         branches:
             - main
             - tmp_hotfix_branch
-permissions:
```
MariaIDineva, line 9: Since permissions were moved from root level to **nx_agents** and **create_release** jobs, shouldn't we add permissions to **gh_pages** and **stop_agents** as well?

PR #14088, opened 2026-04-06
fix(core): fix avatar group accessibility

### LINKED ISSUE #13991: (Core)[fd-avatar-group]ACC-264.1 (Level A)  should give information for how to navigate and item count during a avatar group
### Describe your proposal

Should give information for how to navigate and item count during a avatar group
https://fundamental-ngx.netlify.app/#/core/avatar-group

### Proposed Solution

Reference UI5
https://ui5.github.io/webcomponents/components/AvatarGroup/

### Proposed Alternatives

_No response_

### Organization

_No response_

### StackBlitz example(s)

_No response_

### Priority

None

### Declaration

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13991

### Changes:

`fd-avatar-group` component

- Group Type
  - Added `ariaLabel` input for the group type; falls back to a generated i18n label via `_defaultAriaLabelGroup` (computed from visible/hidden counts)
  - Group type `fd-avatar-group-host` now has `role="button"`, `tabindex="0"`, and the aria-label bound
  - Added `[fdkFocusableItem]="false"` on fd-avatar-group-overflow-button inside #overflowButtonTemplate (group type only) — this prevents the overflow button from being Tab-focusable in group mode, where the entire host is one button
- Individual Type
  - Added a fall-back i18n-generated label via  `_defaultAriaLabelIndividual` computed property using the new `coreAvatarGroup.ariaLabelIndividual` i18n key.
  - Added `LiveAnnouncer` and `_onItemFocused(event: FocusableListItemFocusedEvent) ` for individual type import to have screen reader announce `"N of total"`. **Why this approach?**: aria-posinset/aria-setsize are valid on fd-avatar but the individual type wraps each avatar in fd-popover > fd-popover-control, which breaks any ancestor role="listitem" propagation so screen readers need `LiveAnnouncer` to announce `"N of total"`.
  - Inlined the overflow button content directly inside the individual type's `fdAvatarGroupInternalOverflowButton` template (instead of delegating to ngTemplateOutlet overflowButtonTemplate).  **Why this approach?**:The overflow button was rendered through two layers of `ngTemplateOutlet`, which prevented `fdkFocusableList's contentChildren(FDK_FOCUSABLE_ITEM_DIRECTIVE, { descendants: true })` from discovering it. The fix removes one layer of indirection for the individual type and now the button can be reached via arrow keys.
- Popover
  - Both overflow popovers get `[bodyAriaLabel]` so screen readers announce **"dialog, With N items"** on open
  - Escape key inside the overflow body now closes the popover (passes popoverRef down to the body)

`fd-avatar-group-host`

- Sets `aria-posinset / aria-setsize` on every item after each layout calculation

`fd-avatar-group-overflow-button`

- Added `keydown.enter` and `keydown.space` handlers so the overflow button is keyboard-activatable

`fd-default-avatar-group-overflow-body`

- Added `role="list"` on the avatar list container
- Passes `posInSet / setSize` through to items re-rendered inside the popup
- Overrides `AVATAR_GROUP_HOST_CONFIG` so items are always individually **focusable** inside the popup (fixes the group-type case where they'd otherwise be un-tabbable)
- Added `escapePressed` output, wired via capture-phase listener (bypasses stopPropagation from fdkFocusableList)
- **Improved focus management**: on avatar click, focus moves into the detail view; on back navigation, focus returns to the clicked avatar

`fd-avatar-group-item-renderer` directive

- Added `posInSet / setSize` inputs that write `aria-posinset / aria-setsize` to the rendered DOM element

`fd-popover` (component + service + body)

- New `bodyAriaLabel` input on fd-popover, wired through PopoverConfig → PopoverService → PopoverBodyComponent._bodyAriaLabel signal
- `popover.component.ts` now propagates bodyAriaLabel changes to the service unconditionally (not only when a trigger element is set)

`libs/i18n`

- Added three new translation keys — `coreAvatarGroup.ariaLabelGroup`, `coreAvatarGroup.ariaLabelIndividual` and `coreAvatarGroup.popupBodyAriaLabel`

## Screenshots
Screenshot 2026-04-07 at 11 43 39
Screenshot 2026-04-07 at 11 42 57
Screenshot 2026-04-07 at 11 43 08
Screenshot 2026-04-16 at 13 31 33


InnaAtanasova (approved):

libs/core/avatar-group/components/avatar-group-overflow-button.component.ts
droshev, line 38: can we remove this?

droshev, line 60: can we migrate this to to host too?

libs/core/avatar-group/directives/avatar-group-item-renderer.directive.ts
droshev, line 29: we keep it @Input?

MariaIDineva: Yeah, since this PR is about fixing some accessibility issues in avatar group, I have not put any effort into migrating the components to signals. Especially since we have this work planned as part of other issues.
For the same reason I did not migrate any @HostBindings

droshev (approved):

mikerodonnell89 (commented):
They didn't really provide details on what exactly they expect, but is "Case 1" from this comment addressed in this pr? Maybe I am just missing it https://github.com/SAP/fundamental-ngx/issues/13991#issuecomment-4161349261

MariaIDineva:
> They didn't really provide details on what exactly they expect, but is "Case 1" from this comment addressed in this pr? Maybe I am just missing it [#13991 (comment)](https://github.com/SAP/fundamental-ngx/issues/13991#issuecomment-4161349261)

I am really not sure what exactly is desired either, but I thought it was related to having the screen reader pronounce the phrase `"activate for complete list"` which is circled in red in the first screenshot of the comment you have linked. And we do have it now.

mikerodonnell89 (changes requested):
Not unique to this PR but we should fix the example code not loading

MariaIDineva:
> Not unique to this PR but we should fix the example code not loading

Hm, interesting. It loads for me... It might be some intermittent error.

Issue #13860, opened 2026-02-13
Angular Signals Migration Phase 6 - Simple Containers


InnaAtanasova:
**Containers**

- [ ] Panel - Maria [14109](https://github.com/SAP/fundamental-ngx/pull/14109)
- [ ] Bar - Maria [14101](https://github.com/SAP/fundamental-ngx/pull/14101)

**Layout Components**

- [x] LayoutGrid  - Maria [13735](https://github.com/SAP/fundamental-ngx/pull/13735)
- [x] Scrollbar - Maria [13766](https://github.com/SAP/fundamental-ngx/pull/13766)
- [x] ScrollSpy - Maria [13770](https://github.com/SAP/fundamental-ngx/pull/13770)

PR #14101, opened 2026-04-07
fix(core): signalify bar component

MariaIDineva:
## Related Issue(s)

part of https://github.com/SAP/fundamental-ngx/issues/13860

## Description

- Migrated `BarComponent` from `@Input()` decorators + `CssClassBuilder`/`@applyCssClass` pattern to signal inputs (`input()`) with a `computed()` CSS class binding
- Migrated `BarElementDirective` from `@HostBinding`/`@Input()` to signal `input()` with `host: {}` bindings
- Removed lifecycle hooks (`OnInit`, `OnChanges`, `OnDestroy`) and the `Subscription` teardown that were only needed to drive `buildComponentCssClass()`
- Removed `standalone: true` (redundant since Angular 19)
- Expanded `BarComponent` spec
- Updated `BarElementDirective` spec to import the directive directly instead of `BarModule`
- Removed `NoopAnimationsModule` from `MenuComponent` and `PopoverComponent` mobile-mode specs (no longer needed)

## Changes

| File | Change |
|------|--------|
| `libs/core/bar/bar.component.ts` | Replace `@Input()` + `CssClassBuilder` with signal inputs + `computed()` class |
| `libs/core/bar/directives/bar-element.directive.ts` | Replace `@Input()` + `@HostBinding` with signal `input()` + `host: {}` |
| `libs/core/bar/bar.component.spec.ts` | Add comprehensive class/role/attribute tests |
| `libs/core/bar/directives/bar-element.directive.spec.ts` | Import `BarElementDirective` directly |
| `libs/core/menu/menu.component.spec.ts` | Remove unnecessary `NoopAnimationsModule` |
| `libs/core/popover/popover.component.spec.ts` | Remove unnecessary `NoopAnimationsModule` |


libs/core/bar/bar.component.ts
```diff
@@ -39,55 +35,67 @@ export type BarDesignType = 'header' | 'subheader' | 'header-with-subheader' | '
     ],
     host: {
         '[attr.role]': 'role()',
-        '[class.fd-bar--initial-suggestion-title]': 'initialSuggestionTitle',
-        '[class.fd-bar--initial-suggestion-subline]': 'initialSuggestionSubline'
-    },
-    standalone: true
+        '[class]': '_cssClass()',
+        '[class.fd-bar--initial-suggestion-title]': 'initialSuggestionTitle()',
+        '[class.fd-bar--initial-suggestion-subline]': 'initialSuggestionSubline()'
```
InnaAtanasova: Can you add these 2 to the `_cssClass()`/.

libs/core/bar/bar.component.ts
```diff
@@ -39,55 +35,67 @@ export type BarDesignType = 'header' | 'subheader' | 'header-with-subheader' | '
...
+    }
 })
-export class BarComponent implements OnChanges, OnInit, CssClassBuilder, OnDestroy {
+export class BarComponent {
     /** user's custom classes */
-    @Input()
-    class: string;
+    readonly class = input<string>('');
```
InnaAtanasova: We don't need this. We tested this before and custom class is applied on the element.

libs/core/bar/bar.component.ts
```diff
@@ -39,55 +35,67 @@ export type BarDesignType = 'header' | 'subheader' | 'header-with-subheader' | '
...
+            inPage && !size ? 'fd-bar--page' : '',
+            inPage && size ? `fd-bar--page-${size}` : '',
+            inHomePage && !size ? 'fd-bar--home-page' : '',
+            inHomePage && size ? `fd-bar--home-page-${size}` : '',
+            this.clear() ? 'fd-bar--clear' : '',
+            this.class()
+        ]
+            .filter(Boolean)
+            .join(' ');
+    });
```
InnaAtanasova: You can reduce the number of lines and combine the class binding in the host object. Also no need of the underscore prefix.
```
protected readonly cssClass = computed(() => {
      const barDesign = this.barDesign();                                                        
      const inPage = this.inPage();                                                              
      const inHomePage = this.inHomePage();                                                      
      const size = this.size();                                                                  
                                                                                                 
      return [                                                                                   
          'fd-bar',                                                                              
          this._contentDensityObserver.isCompactSignal() ||                                      
  this._contentDensityObserver.isCondensedSignal()                                               
              ? ''                                                                               
              : 'fd-bar--cozy',                                                                  
          barDesign && `fd-bar--${barDesign}`,                                                   
          inPage && `fd-bar--page${size ? `-${size}` : ''}`,                                     
          inHomePage && `fd-bar--home-page${size ? `-${size}` : ''}`,                            
          this.clear() && 'fd-bar--clear',                                                       
          this.initialSuggestionTitle() && 'fd-bar--initial-suggestion-title',                   
          this.initialSuggestionSubline() && 'fd-bar--initial-suggestion-subline',               
          this.class()                                                                           
      ]                                                                                          
          .filter(Boolean)                                                                       
          .join(' ');                                                                            
  });      
```

libs/core/bar/bar.component.ts
MariaIDineva: yeah, right, thanks!

InnaAtanasova (approved):

libs/core/bar/bar.component.ts
```diff
@@ -39,55 +35,62 @@ export type BarDesignType = 'header' | 'subheader' | 'header-with-subheader' | '
...
-        '[class.fd-bar--initial-suggestion-subline]': 'initialSuggestionSubline'
-    },
-    standalone: true
+        '[class]': 'cssClass()'
+    }
 })
-export class BarComponent implements OnChanges, OnInit, CssClassBuilder, OnDestroy {
-    /** user's custom classes */
-    @Input()
-    class: string;
```
droshev, line 50: so this is not a breaking change in the wide meaning?

MariaIDineva: Yes, correct. If the consuming applications adds a custom class to our tag, it will be kept even if we don't have such an input explicitly declared.
Screenshot 2026-04-09 at 9 19 25
Screenshot 2026-04-09 at 9 21 53

PR #14118, opened 2026-04-10
fix(core): add a11y improvements for Card component

### LINKED ISSUE #14096: [Core][fd-card] ACC-270 (Level A) Card items should be navigable with arrow keys instead of Tab key
### Component
`fd-card` (Core)

### Accessibility Checkpoint
ACC-270.3 (Level A)

### Problem Description

When using keyboard navigation, individual `fd-card` items in a card list/group can only be navigated using the **Tab key**, but according to accessibility standards (ACC-270.3, Level A), navigation within a group of control elements should be possible using **arrow keys**.

**Current behavior:**
Cards within a card group are navigated via Tab key.

**Expected behavior:**
Navigation between cards within the same group/region should be possible with arrow keys, consistent with standard keyboard navigation patterns for composite widgets.

### Why This Matters

Keyboard navigation is a prerequisite for assistive technologies (e.g., screen readers like JAWS) to support users with disabilities. Using the correct navigation keys (arrow keys for intra-component navigation, Tab for inter-component navigation) is essential for WCAG 2.x Level A compliance.

### References
- WCAG 2.x Success Criterion related to keyboard accessibility
- ACC-270.3 (Level A) keyboard navigation guideline

### LINKED ISSUE #14106: [Core][fd-card] ACC-264.1 (Level A) Card badge content not announced by screen readers
## Summary

When navigating to an `fd-card` component using a screen reader, badge content within the card is not announced. This leaves assistive technology users without the full context available to sighted users.

## Observed Behavior

When a user navigates to an `fd-card` using a screen reader (e.g. JAWS), badge elements rendered inside the card are not read aloud. The screen reader skips over badge content entirely, failing to convey the information they represent.

I can set `aria-describedby`attribute for the `fd-card` to read the subtitle and content. But when using the tab key to navigate the action button, it will read the whole `aria-describedby` content which is redundancy.

## Why aria-describedby is read when tabbing to "More Actions"(Provided by AI)
Root cause: fd-card has both tabindex="0" (from [interactive]="true") and aria-describedby, making the card itself a focusable element.

## Expected Behavior

All badge content within an `fd-card` should be accessible and announced by screen readers. Users relying on assistive technology should receive the same information as sighted users when navigating to or through a card.

## WCAG Criterion Violated

- **WCAG 2.1 — 1.3.1 Info and Relationships (Level A)**  
  Information, structure, and relationships conveyed through presentation must be available in text or programmatically determinable.

## Reproduction Steps

1. Open a page containing one or more `fd-card` components that include badge elements.
2. Navigate to the card using a screen reader (tested with JAWS 2026).
3. Observe that badge content inside the card is not announced.

## Severity

**Level A** — This is a fundamental accessibility barrier that prevents screen reader users from receiving badge information that is visually presented within the card.

## Environment

- Browser: Google Chrome (latest)
- Screen Reader: JAWS 2026
- OS: Windows 11

InnaAtanasova:
## Related Issue(s)
 
closes https://github.com/SAP/fundamental-ngx/issues/14106
closes https://github.com/SAP/fundamental-ngx/issues/14096

### Summary
This PR enhances the Card component's accessibility features and modernizes its implementation by migrating to Angular 21+ patterns. Key improvements include screen reader support, internationalization, RTL keyboard navigation, and signal-based state management.

### Changes

  **Accessibility Enhancements:**
  - Added screen reader support with visually hidden content for card descriptions and role descriptions
  - Implemented comprehensive aria-describedby that includes subtitle, counter, and badge IDs
  - Added ariaDescribedby input to allow additional descriptive content references
  - Enhanced badge accessibility with proper ARIA labels and ID references
  - RTL keyboard navigation support for card lists via withHorizontalOrientation()

  **Internationalization:**
  - Integrated FdTranslatePipe and resolveTranslationSignal for i18n support
  - Added translated defaults for ariaRoledescription and ariaDescription
  - Translation keys support interpolation (e.g., {cardType} for different card types)
  - Added translations for 35+ languages

  **Modernization:**
  - Migrated from CssClassBuilder + @applyCssClass to computed() + host: { '[class]': } pattern
  - Migrated @HostListener to host: { '(keydown)': } binding
  - Removed OnChanges, OnInit, OnDestroy lifecycle hooks in favor of signals
  - Replaced BehaviorSubject patterns with signal() and computed()
  - Used inject() for dependency injection instead of constructor parameters

  **Tests:**
  - Added 7 new test cases covering:
    - Badge IDs in aria-describedby
    - Subtitle, second subtitle, and counter ID inclusion
    - Additional ariaDescribedby input
    - Translation interpolation with different card types
    - Custom ariaRoledescription override

  ⚠️  None - This PR has no breaking changes to the public API.

###  Analysis:
  - `ariaRoledescription` and `ariaDescription` inputs now default to undefined instead of hardcoded strings ('Card' and 'Activate for action or navigation'), but this enhances functionality by providing translated defaults that adapt to the user's language. Applications not providing custom values will now get properly translated text instead of English-only strings.
  - Removed internal `getCardModifierClassNameByCardType()` function was never exported in the public API.
  - Component lifecycle interfaces (OnChanges, OnInit, OnDestroy) and `CssClassBuilder` were implementation details, not part of the public API contract.
  - All component inputs remain available with the same semantics; signal-based inputs are backward compatible.


mikerodonnell89 (changes requested):
Arrow keys are not working properly in RTL mode. Right arrow moves focus to the left

libs/core/card/content/card-content.component.ts
```diff
@@ -1,29 +1,15 @@
...
-
-    /** @hidden */
-    constructor(public readonly elementRef: ElementRef<HTMLElement>) {}
-
-    /** @hidden */
-    @applyCssClass
-    buildComponentCssClass(): string[] {
-        return [CLASS_NAME.cardContent];
+    host: {
+        class: 'fd-card__content'
```
MariaIDineva: Could we use the constant `CLASS_NAME.cardContent` here since we have it declared anyway?

libs/core/card/header-elements/card-subtitle.directive.ts
```diff
@@ -1,35 +1,27 @@
...
-    /** @hidden */
-    constructor(public readonly elementRef: ElementRef<HTMLElement>) {}
-
-    /** @hidden */
-    @applyCssClass
-    buildComponentCssClass(): string[] {
-        return [CLASS_NAME.cardSubtitle];
+    ],
+    host: {
+        class: 'fd-card__subtitle',
```
MariaIDineva: Also here, shall we use `CLASS_NAME.cardSubtitle`?

libs/core/card/header-elements/card-title.directive.ts
```diff
@@ -1,43 +1,27 @@
...
     selector: '[fd-card-title]',
-    standalone: true,
     providers: [
         {
             provide: FD_CARD_TITLE,
             useExisting: CardTitleDirective
         }
     ],
     host: {
+        class: 'fd-card__title',
```
MariaIDineva: And here - `CLASS_NAME.cardTitle`?
And for all the classes below that used to use constants and are now hardcoded - `CLASS_NAME.cardAnalyticsContent`, `CLASS_NAME.cardAnalyticsText`, `CLASS_NAME.cardAnalytics`, `CLASS_NAME.cardAnalyticalArea`, etc.

libs/core/card/card.component.ts
```diff
@@ -43,43 +50,52 @@ let cardId = 0;
...
+    cardCounter = contentChild<CardCounterDirective>(FD_CARD_COUNTER);
+
     /** @hidden */
     cardMediaHeading = contentChild<CardMediaHeadingDirective>(FD_CARD_MEDIA_HEADING);
 
     /**
      * text for the card badge
      */
-    badge = input<Nullable<string>>();
+    badge = input<string | null | undefined>();
```
MariaIDineva: Those contentChildren and inputs can be all marked as `readonly`

MariaIDineva (approved):

mikerodonnell89 (approved):

MariaIDineva (approved):

PR #14119, opened 2026-04-10
fix(platform): table pop-in keyboard improvements

### LINKED ISSUE #13174: Platform Table: Keyboard Navigation Not Working in "Hiding Columns Depending on Table Size" Example
In the Table example: **_Hiding columns depending on the table size_**, keyboard navigation is not functioning correctly.

**Steps to Reproduce:**

Tab to focus on the first row of the table.

Attempt to navigate to other rows using the arrow keys or Tab.

**Expected Behavior:**

Arrow keys should allow navigation between rows.

Tab should leave the table.

**Actual Behavior:**

Arrow keys do not navigate to other rows.

mikerodonnell89:
fixes #13174

Allows the pop-in table to be navigated via arrow keys


MariaIDineva:
I am trying to test the changes by navigating through the table but when I tab into the table, more specifically when I tab into the first row for the first time, the visible focus disappears and then I cannot navigate to the next cell or to the next row using the arrows. The focus moves out of the table on the next tab, but arrows don't seem to work inside unless I click on a cell with the mouse. Then arrows navigate correctly.

mikerodonnell89:
> I am trying to test the changes by navigating through the table but when I tab into the table, more specifically when I tab into the first row for the first time, the visible focus disappears and then I cannot navigate to the next cell or to the next row using the arrows. The focus moves out of the table on the next tab, but arrows don't seem to work inside unless I click on a cell with the mouse. Then arrows navigate correctly.

just pushed a fix, thanks

InnaAtanasova (approved):

MariaIDineva (approved):

PR #14126, opened 2026-04-15
fix(platform): platform combobox can now be used inside dialog

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14107

## Description

When `fdp-combobox` is placed inside `fd-dialog`, pressing the Down Arrow key to navigate into the dropdown list caused focus to immediately snap back to the search input. This happened because the dialog's focus trap was active while the combobox dropdown (rendered in a CDK overlay outside the dialog DOM) tried to receive focus.

The fix follows the existing pattern used by `fd-combobox`, `fd-multi-combobox`, `fd-multi-input`, and `fd-date-picker`: pause the dialog's focus trap when the dropdown opens, and unpause it when the dropdown closes.

### Before:
Pressing the Down Arrow key would move focus to the first item, but then back to the input, making it impossible to select any other item from the dropdown.

https://github.com/user-attachments/assets/12c28e8a-9db3-4115-b037-7e625448eb15

### After:

Pressing the Down Arrow key makes it possible to move though the items in the dropdown.

https://github.com/user-attachments/assets/b486a383-188b-480d-8e88-5441e229b328


InnaAtanasova (approved):
🚀

PR #14128, opened 2026-04-15
fix(platform): table info toolbar bugs when using outerScroll

### LINKED ISSUE #12855: Filter Info Bar overlaps sticky header when the page is scrolled
If we set outerScroll to the platform table and apply a filter, scrolling does not work properly. The filter information overlaps the table header.

bug

no

Angular (18)
newest (0.54)

- Create a table with filters and an outer scroll feature.
- Set a filter so that the information is shown below the table header.
- Scroll the table.

https://github.com/user-attachments/assets/3f7ce91d-4315-4f33-b927-f3b40e65c717

https://stackblitz.com/edit/vtw1marn?file=src%2Fapp%2Fplatform-table-initial-state-example.component.ts,src%2Fapp%2Fplatform-table-initial-state-example.component.html

mikerodonnell89:
fixes #12855 

before:

https://github.com/user-attachments/assets/826ef4e6-176c-4a8e-b162-880ff7d28104

after:

https://github.com/user-attachments/assets/9c5f0bcb-1408-48d9-826d-b2c93ddd98af


libs/platform/table/table.component.scss
```diff
@@ -583,9 +583,18 @@ fdk-dynamic-portal {
...
+        .fd-toolbar:not(.fd-toolbar--info) {
+            z-index: 10;
+        }
+        .fd-toolbar.fd-toolbar--info {
+            z-index: 9;
+            top: 2.75rem;
+        }
+        .fd-toolbar.fd-toolbar--info + fd-busy-indicator thead {
+            top: 4.75rem;
+        }
```
InnaAtanasova, line 597: A minor thing, but would be easier to find and maintain stuff in future, can you make these nested in the .fd-toolbar?
```
.fd-toolbar {
    position: sticky;
    top: 0;

    &:not(.fd-toolbar--info) {
            z-index: 10;
     }
        
     &.fd-toolbar--info {
            z-index: 9;
            top: 2.75rem;
     }

     &.fd-toolbar--info + fd-busy-indicator thead {
            top: 4.75rem;
     }
}

```

InnaAtanasova (approved):

MariaIDineva (approved):

PR #14129, opened 2026-04-15
fix(core): a11y improvements for Breadcrumb component

### LINKED ISSUE #14095: [Core][fd-breadcrumb] ACC-270 (Level A) Breadcrumb items should be navigable with arrow keys instead of Tab key
### Component
`fd-breadcrumb` (Core)

### Accessibility Checkpoint
ACC-270.3 (Level A)

### Problem Description

When using keyboard navigation, the breadcrumb links inside `fd-breadcrumb` can only be navigated using the **Tab key**, but according to accessibility standards (ACC-270.3, Level A), navigation within control elements (such as breadcrumb items) should be possible using **arrow keys**.

**Current behavior:**
Breadcrumb items are navigated via Tab key.

**Expected behavior:**
Navigation within breadcrumb items should be possible with arrow keys, consistent with standard keyboard navigation patterns for composite widgets.

### Why This Matters

Keyboard navigation is a prerequisite for assistive technologies (e.g., screen readers like JAWS) to support users with disabilities. Using the correct navigation keys (arrow keys for intra-component navigation, Tab for inter-component navigation) is essential for WCAG 2.x Level A compliance.

### References
- WCAG 2.x Success Criterion related to keyboard accessibility
- ACC-270.3 (Level A) keyboard navigation guideline

### LINKED ISSUE #14097: [Core][fd-breadcrumb] ACC-264 (Level A) Breadcrumb item position not announced by screen reader
### Component
`fd-breadcrumb` (Core)

### Accessibility Checkpoint
ACC-264.1 (Level A)

### Problem Description

When navigating through breadcrumb links (e.g., first level, second level, third level), the screen reader does not announce the item's position within the breadcrumb trail (e.g., "1 of 3", "2 of 3", "3 of 3").

**Current behavior:**
Screen reader announces only the link label (e.g., "first level, link") without its position within the set.

**Expected behavior:**
Screen reader should announce both the label and its position within the set (e.g., "first level, link, 1 of 3").

### Why This Matters

Announcing item count and position is essential for users relying on assistive technologies (e.g., JAWS screen reader) to understand their location within a navigation structure. This is required for WCAG 2.x Level A compliance (ACC-264.1).

The `fd-breadcrumb` component should expose proper  `aria-posinset`, and `aria-setsize` attributes on breadcrumb items so screen readers can announce position information correctly.

### References
- WCAG 2.x Success Criterion related to name, role, value
- ACC-264.1 (Level A) item count announcement guideline

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14097
closes https://github.com/SAP/fundamental-ngx/issues/14095

### Problem:
Breadcrumb items could only be navigated using the Tab key. Accessibility standards (ACC-270.3, Level A) require arrow key navigation within composite widgets like breadcrumbs.

#### Arrow key navigation:
- Added Left/Right/Up/Down arrow key handling on `fd-breadcrumb` host to move focus between visible breadcrumb items
- Right/Down moves to the next item, Left/Up moves to the previous 
- RTL-aware: Left and Right swap direction in RTL mode

#### Roving tabindex
- Implemented roving tabindex pattern: only the first visible item has `tabindex="0"`, all others have `tabindex="-1"`
- Tab enters the breadcrumb on the first item, then exits — arrow keys navigate within
- Tabindex updates are deferred via Promise.resolve() to run after the overflow layout finishes calculating visibility

#### Current page span focusability
- `<span aria-current="page">` is now included in the focus sequence with a managed tabindex

#### Href-less links focusability
- Links without href (e.g. click-only handlers) now receive tabindex so they are programmatically focusable

#### Overflow menu keyboard fix
- Fixed Enter/Space on the overflow "..." button: was calling toggle() on keydown which raced with the popover service's click trigger, causing double-toggle (open then immediately close)
- Now uses keyup to open the menu, ensuring the key is fully released before focus moves to menu items

### Files changed
`breadcrumb.component.ts` — arrow key handler, roving tabindex, overflow menu keyboard fix
`breadcrumb.component.html` — overflow button keyboard bindings
`breadcrumb.component.spec.ts` — tests for arrow navigation (Left, Right, Up, Down, aria-current span, boundary clamping)

#### No breaking changes
- Tab key behavior unchanged (enters/exits breadcrumb)
- No public API additions or removals
- Arrow keys add focus movement between already-focusable elements


libs/core/breadcrumb/breadcrumb.component.html
```diff
@@ -52,15 +52,17 @@
...
                     aria-haspopup="menu"
-                    tabindex="0"
+                    tabindex="-1"
                     class="fd-breadcrumb__collapsed"
-                    (keydown.enter)="_keyDownHandle($event)"
-                    (keydown.space)="_keyDownHandle($event)"
+                    (keyup.enter)="openOverflowMenu($event)"
+                    (keyup.space)="openOverflowMenu($event)"
+                    (keydown.enter)="$event.preventDefault()"
+                    (keydown.space)="$event.preventDefault()"
```
InnaAtanasova, line 65: These and `#menuTriggerSpan` on line 55 are fixing the problem where when we have an overflow button with menu, when we open the menu with the keyboard, the first link executes either the click or the href. This was not reported as an issue, but I discovered it while working on the other issues.

libs/core/breadcrumb/breadcrumb.component.ts
```diff
@@ -224,16 +273,98 @@ export class BreadcrumbComponent implements AfterViewInit, HasElementRef {
...
+     * - querySelectorAll naturally returns elements in DOM (visual) order, which is
+     *   exactly what arrow key navigation needs, without manual sorting.
+     */
+    private _getVisibleLinks(): HTMLElement[] {
+        const el = this.elementRef.nativeElement;
+        return Array.from(
+            el.querySelectorAll<HTMLElement>(
+                '.fd-overflow-layout__item:not(.fd-overflow-layout__item--soft-hidden) a.fd-link, .fd-overflow-layout__item:not(.fd-overflow-layout__item--soft-hidden) [aria-current], .fd-breadcrumb__collapsed'
+            )
+        );
```
InnaAtanasova, line 310: I tried a more "Angular" approach, but since the overflow button is not part of the component template I had to add a really complex logic, including portals, so this as ugly as it is was the simplest way I managed to find.

libs/core/breadcrumb/breadcrumb.component.ts
```diff
@@ -224,16 +273,98 @@ export class BreadcrumbComponent implements AfterViewInit, HasElementRef {
...
+    }
+
+    /**
+     * @hidden
+     * Applies roving tabindex: sets all focusable breadcrumb elements to tabindex="-1",
+     * then promotes the first visible one to tabindex="0".
+     * This ensures Tab enters/exits the breadcrumb as a single stop,
+     * and non-href elements (click-only links, aria-current spans) are programmatically focusable.
+     */
+    private _updateTabIndexes(): void {
```
InnaAtanasova, line 320: This logic here is a bit complicated due to a few things:
in the second documentation example our first link has a click event, and also we can an overflow button at the beginning. These should get the first focus. We have a span as the last element, which is not an interactive element. It needs to get a tabindex because we need to focus it. On top, the Tab should only go to the first item and leaves the Breadcrumb on second Tab.

libs/core/breadcrumb/breadcrumb.component.ts
```diff
@@ -224,16 +273,98 @@ export class BreadcrumbComponent implements AfterViewInit, HasElementRef {
...
+            const positionLabel = this._translationResolver.resolve(lang, 'coreBreadcrumb.positionLabel', {
+                current: position,
+                total: totalCount
+            });
+
+            // Set aria-label on link if present
+            if (item.breadcrumbLink) {
+                const linkEl = item.breadcrumbLink.elementRef.nativeElement;
+                const linkText = linkEl.textContent?.trim() || '';
+                item.breadcrumbLink.ariaLabel.set(`${linkText}, ${positionLabel}`);
```
InnaAtanasova: this logic is for the link elements

libs/core/breadcrumb/breadcrumb.component.ts
```diff
@@ -224,16 +273,98 @@ export class BreadcrumbComponent implements AfterViewInit, HasElementRef {
...
+            } else {
+                // Set aria-label on direct child span with aria-current if present
+                const itemElement = item.elementRef.nativeElement;
+                const directChildSpan = Array.from(itemElement.children).find(
+                    (child) => child.tagName === 'SPAN' && child.hasAttribute('aria-current')
+                ) as HTMLElement;
+                if (directChildSpan) {
+                    const spanText = directChildSpan.textContent?.trim() || '';
+                    directChildSpan.setAttribute('aria-label', `${spanText}, ${positionLabel}`);
```
InnaAtanasova, line 365: This logic is for the last item. It should be a direct child, because links can have nested spans, so we need to make sure we don't get them. I added an additional protection with 'aria-current' but just in case...

droshev (approved):

libs/core/breadcrumb/breadcrumb.component.ts
```diff
@@ -199,17 +201,64 @@ export class BreadcrumbComponent implements AfterViewInit, HasElementRef {
...
-    /** @hidden */
-    _keyDownHandle(event: Event): void {
-        this._menuComponent.toggle();
+    /**
+     * @hidden
+     * Opens the overflow menu on keyboard activation.
+     * Bound to keyup (not keydown) so that by the time the menu opens and captures focus,
+     * the key has been fully released and no stale key events reach the first menu item.
+     */
+    openOverflowMenu(event: Event): void {
```
MariaIDineva, line 226: Maybe that method should be protected.

libs/core/breadcrumb/breadcrumb.component.ts
```diff
@@ -199,17 +201,64 @@ export class BreadcrumbComponent implements AfterViewInit, HasElementRef {
...
         event.preventDefault();
+        if (!this._menuComponent.isOpen()) {
+            this._menuComponent.open();
+        }
+    }
+
+    /** @hidden */
+    _onKeydown(event: KeyboardEvent): void {
```
MariaIDineva, line 234: Maybe this one should be protected as well and not prefixed with an underscore.

libs/core/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -175,4 +181,121 @@ describe('BreadcrumbComponent', () => {
...
+    describe('arrow key navigation', () => {
+        function createKeyboardEvent(keyCode: number, key: string): KeyboardEvent {
+            return new KeyboardEvent('keydown', { key, bubbles: true });
+        }
+
+        it('should move focus to next link on ArrowRight', async () => {
+            await whenStable(fixture);
+            const links = fixture.nativeElement.querySelectorAll('a.fd-link') as NodeListOf<HTMLElement>;
+            if (links.length < 2) {
+                return;
```
MariaIDineva, line 194: What is the purpose of those `return`s in the tests? Wouldn't they cause the test to pass silently in case of a problem with the setup?

libs/core/link/link.component.ts
```diff
@@ -91,16 +91,16 @@ export class LinkComponent implements HasElementRef {
     readonly routerLink = inject<RouterLink>(LINK_ROUTER_TARGET);
 
     /** @hidden */
-    protected readonly prefixPortal = signal<Portal<any> | null>(null);
+    readonly prefixIconName = signal<string>('');
 
     /** @hidden */
-    protected readonly postfixPortal = signal<Portal<any> | null>(null);
+    readonly postfixIconName = signal<string>('');
```
MariaIDineva, line 97: Just a note - converting those to protected is a breaking change, which we should document in my opinion.

MariaIDineva (approved):

libs/core/link/link.component.ts
InnaAtanasova: they were protected, now they are not

MariaIDineva: oh, yes, sorry, my bad.

PR #14137, opened 2026-04-17
fix(core): limit rendered tokens in multi-input to visible count

### LINKED ISSUE #13100: multi-input component is not optimized.
Image

https://github.com/SAP/fundamental-ngx/blob/main/libs/core/multi-input/multi-input.component.html#L70

This implementation is not optimized, considering that there may be a large list of selected items. In our application, this component allows selecting a large number of items. As soon as we try to render such a component, rendering takes a significant amount of time because the component renders each item individually. My suggestion is to implement virtual scrolling for this element. This way, we can eliminate performance problems.

droshev:
## Related Issue(s)

closes #13100

## Problem

`MultiInputComponent` rendered **every** selected `<fd-token>` into the DOM unconditionally. With large selections (100–1000 items), the Angular change-detection cycle for the `@for` loop caused a measurable paint delay on every keystroke and every focus event, degrading perceived performance for power users.

The regression was introduced via `viewModel.selectedOptions` being bound directly in the template with no render gate.

## Key files

| File | Role |
|---|---|
| `libs/core/multi-input/multi-input.component.ts` | Core logic — new signals, computed slices, focus/blur/click handlers |
| `libs/core/multi-input/multi-input.component.html` | Template — `_renderedTokens()` binding + "+N more" indicator |

## Solution — Architecture: multi-input owns token virtualization

The fix is entirely inside `MultiInputComponent`. The tokenizer (`TokenizerComponent`) is **not touched** — its public/internal API is stable and unchanged.

### What changed

| Area | Change |
|---|---|
| `maxVisibleTokens` input | New `input<number>(12)` signal input. Controls the collapsed-view cap. |
| `_renderAllTokens` | Private `signal(false)`. Flipped to `true` on input focus or "+N more" click. |
| `_renderedTokens` | Protected `computed()`. Returns the full list when `_renderAllTokens` is true or selection ≤ cap; otherwise returns a slice of `maxVisibleTokens`. |
| `_hiddenCount` | Protected `computed()`. Drives the "+N more" indicator visibility. |
| `_selectedOptionsSignal` | Private signal bridging the existing RxJS `_viewModel$` stream to the signal graph via `toSignal()`. |
| `_onInputFocus()` | Replaces the direct `tokenizer._showAllTokens()` call on the input's `(focus)` binding. Sets `_renderAllTokens` before expanding the tokenizer so all tokens are materialized before measurement. |
| `_onMoreClicked()` | Click handler on the "+N more" indicator. Expands the token list, expands the tokenizer, and focuses the search input. |
| `_focusOut()` (existing) | Extended to reset `_renderAllTokens` and call `tokenizer._hideTokens()` when focus leaves the host. |
| Template `@for` | Changed from `viewModel.selectedOptions` to `_renderedTokens()`. |
| Template `@if (_hiddenCount() > 0)` | Adds the "+N more" `<span fd-form-control class="fd-tokenizer-more">` indicator when tokens are hidden. |

### Why `fd-form-control` on the `<span>`

The `<fd-tokenizer>` template has two `<ng-content>` slots: `select="fd-token"` and `select="[fd-form-control]"`. The `+N more` span needs to project into the second slot so it sits inline with the input. `fd-form-control` as an **attribute** routes the projection correctly — the `FdFormControlComponent` directive selector is scoped to `input` and `textarea`, so it does **not** activate on a `<span>`, meaning zero behavioral side effect.

## What is NOT changed

- `TokenizerComponent` — zero modifications.
- `PlatformMultiInputComponent` — follow-up change to be done separately if needed.
- No breaking changes to existing `MultiInputComponent` inputs/outputs.


libs/core/multi-input/multi-input.component.ts
```diff
@@ -441,6 +451,34 @@ export class MultiInputComponent<ItemType = any, ValueType = any>
...
     readonly _viewModel$: Observable<ViewModel<ItemType, ValueType>> = this._getViewModel();
 
+    /** @hidden */
+    readonly _contentDensityObserver = inject(ContentDensityObserver);
+
+    /** @hidden */
+    readonly elementRef = inject<ElementRef<HTMLElement>>(ElementRef);
+
+    /** @hidden — slice (or full list) of options rendered into <fd-tokenizer>. */
+    protected readonly _renderedTokens = computed(() => {
```
MariaIDineva: I think our convention is to not use an underscore prefix for protected members, so these should be `renderedTokens` and `hiddenCount`.

MariaIDineva (dismissed):
That seems like an optimization to the component in terms of page load, but I am not sure it resolves the reported problem completely. Now a page with a huge number of items will likely load faster, but once the multi input is open, all elements would still be rendered again and would cause delay/buggy user experience as there is no virtual scrolling inside the dropdown itself.

libs/core/multi-input/multi-input.component.html
```diff
@@ -66,7 +66,7 @@
                 fdMultiAnnouncer
                 [multiAnnouncerOptions]="filterFn(dropdownValues, searchTerm)"
             >
-                @for (option of viewModel.selectedOptions; track valueFn(option.item)) {
+                @for (option of _renderedTokens(); track valueFn(option.item)) {
```
InnaAtanasova: We should not have prefixed members in the template. Only private members are prefixed with an underscore.

libs/core/multi-input/multi-input.component.html
```diff
@@ -77,6 +77,17 @@
                         <span [innerHtml]="option.label"></span>
                     </fd-token>
                 }
+                @if (_hiddenCount() > 0) {
```
InnaAtanasova: Same here

libs/core/multi-input/multi-input.component.html
```diff
@@ -77,6 +77,17 @@
...
+                    <span
+                        fd-form-control
+                        class="fd-tokenizer-more"
+                        role="button"
+                        tabindex="-1"
+                        (click)="_onMoreClicked()"
```
InnaAtanasova: I have a few concerns about this approach. I assume you used `fd-form-control` so that it is routed through the tokenizer's `<ng-content select="[fd-form-control]">` projection slot. While this technically works, it's semantically incorrect. The `fd-form-control` directive is intended for actual form controls (input, textarea), and I don't see how this is a form control.

InnaAtanasova: Consider adding a dedicated projection slot to TokenizerComponent for custom content (e.g., <ng-content select=".fd-tokenizer-more">), or simply project without a specific selector and handle positioning with CSS.

InnaAtanasova: You have both `role="button"` and `tabindex="-1"`, which makes it keyboard-inaccessible. If it's intended to be clickable, it should be keyboard-accessible with tabindex="0" and support Enter/Space key events.

libs/core/multi-input/multi-input.component.ts
```diff
@@ -422,6 +425,13 @@ export class MultiInputComponent<ItemType = any, ValueType = any>
     @ViewChild(TokenizerComponent)
     tokenizer: TokenizerComponent;
 
+    /**
+     * Maximum number of selected tokens to render when the input is not focused.
+     * When the selection exceeds this limit, the remaining count is shown as a
+     * "+N more" indicator. Defaults to 12.
+     */
+    readonly maxVisibleTokens = input<number>(12);
```
InnaAtanasova: `readonly maxVisibleTokens = input(12);`

libs/core/multi-input/multi-input.component.ts
```diff
@@ -441,6 +451,34 @@ export class MultiInputComponent<ItemType = any, ValueType = any>
     /** @hidden */
     readonly _viewModel$: Observable<ViewModel<ItemType, ValueType>> = this._getViewModel();
 
+    /** @hidden */
+    readonly _contentDensityObserver = inject(ContentDensityObserver);
```
InnaAtanasova: no need to be prefixed

InnaAtanasova (changes requested):
**Additional comments from Claude review, which I verified and think are applicable:**


- [multi-input.component.ts:545-547] The `_onInputFocus()` method could benefit from a JSDoc comment explaining why `_renderAllTokens.set(true)` must happen before `tokenizer._showAllTokens()` (so all tokens are materialized before the tokenizer measures them). The PR description mentions this, but the code doesn't.

- [multi-input.component.spec.ts:299-302] The helper function `setSelected()` has an unnecessary await fixture.whenStable() at the start before doing anything. This adds ~5ms per test for no benefit.

- [multi-input.component.ts:432] The JSDoc comment for `maxVisibleTokens` says "When the selection exceeds this limit, the remaining count is shown as a '+N more' indicator." However, it would be clearer to mention that this limit only applies when the input is not focused. When focused, all tokens are shown regardless of this limit.

- [multi-input.component.spec.ts:322-331] The test "'+N more' click expands the slice and calls tokenizer._showAllTokens()" is testing implementation details (spying on tokenizer._showAllTokens()). Tests should verify observable behavior, not internal method calls.

**Better assertion:** Verify that after clicking "+N more", the token count increases to 50 and the indicator disappears. The internal implementation detail of calling _showAllTokens() is not user-observable and creates brittle coupling.

mikerodonnell89 (changes requested):
I think this is something we'd probably want for any component which uses the tokenizer - changes should come from that component in my opinion

mikerodonnell89 (changes requested):
Testing with 1000 items, seeing double "___ more" text on resize

Image

droshev:
> I think this is something we'd probably want for any component which uses the tokenizer - changes should come from that component in my opinion

keeping ownership in multi-input rather than tokenizer was a deliberate decision: multi-input gets used with hundreds or thousands of selections, and the fix is to not render that many token elements in the first place. That's about how multi-input manages its data, not about how tokens are laid out.
Tokenizer is used in places where selection counts are naturally small. Pushing this into tokenizer would give every consumer an API they don't need, for a problem only multi-input has.

droshev:
> That seems like an optimization to the component in terms of page load, but I am not sure it resolves the reported problem completely. Now a page with a huge number of items will likely load faster, but once the multi input is open, all elements would still be rendered again and would cause delay/buggy user experience as there is no virtual scrolling inside the dropdown itself.

Thanks! to set expectations on what this PR does and doesn't cover:

**What it fixes:** initial render and focus-time render with large selections. A page with one or many multi-inputs each holding 1000 items now renders fast, because we don't materialize all 1000 token elements upfront.

**What it doesn't fix:** interaction cost on a focused multi-input with a large selection. If the user focuses a multi-input with 1000 items and wants to remove item #874, that interaction still walks the full list. Same for the dropdown rendering — at 1000 items, the list itself has the same eager-render problem the tokens used to have.

mikerodonnell89:
> > I think this is something we'd probably want for any component which uses the tokenizer - changes should come from that component in my opinion
> 
> keeping ownership in multi-input rather than tokenizer was a deliberate decision: multi-input gets used with hundreds or thousands of selections, and the fix is to not render that many token elements in the first place. That's about how multi-input manages its data, not about how tokens are laid out. Tokenizer is used in places where selection counts are naturally small. Pushing this into tokenizer would give every consumer an API they don't need, for a problem only multi-input has.

Multi-input and multi-combobox are so similar, I don't see how this could only happen in multi-input but not multi-combobox. And if it can happen in multi-input, couldn't it happy in any custom component which utilizes tokenizer, built by an application developer?

droshev:
> > > I think this is something we'd probably want for any component which uses the tokenizer - changes should come from that component in my opinion
> > 
> > 
> > keeping ownership in multi-input rather than tokenizer was a deliberate decision: multi-input gets used with hundreds or thousands of selections, and the fix is to not render that many token elements in the first place. That's about how multi-input manages its data, not about how tokens are laid out. Tokenizer is used in places where selection counts are naturally small. Pushing this into tokenizer would give every consumer an API they don't need, for a problem only multi-input has.
> 
> Multi-input and multi-combobox are so similar, I don't see how this could only happen in multi-input but not multi-combobox. And if it can happen in multi-input, couldn't it happy in any custom component which utilizes tokenizer, built by an application developer?

@mikerodonnell89 Good question - and I get the instinct, they do look similar from the outside. But I'd flip the framing: we ship them as two separate components precisely because they target different usages. If they covered the same ground they'd be one component, and if they had the same features the implementations wouldn't have diverged the way they have.

> Multi-combobox's own guidelines back this up:
> The multi-combo box lets users select one or more options from a predefined list.
> Don't use the multi-combo box if you want to display large data sets. In this case, use a multi-input field with a select dialog or value help dialog (for more than 1000 items).

So multi-combobox's docs send the >200-item case over to multi-input - which is exactly why I created virtualization there. Adding the same machinery to multi-combobox would optimize for a scenario it's contractually not supposed to be in. Happy to revisit if a real-world case crosses that line, but I'd rather not pre-build for it.

On the "any custom tokenizer consumer" angle - tokenizer stays a width-driven layout primitive on purpose. The coordination piece is already generalized: the new hasInternalOverflowIndicator computed is public, so any consumer that genuinely needs count-driven virtualization can wire it up the same way multi-input does, without widening tokenizer's surface for everyone.

two components = two use cases. Scope stays multi-input; pattern's reusable if/when someone needs it elsewhere.

mikerodonnell89:
> > > > I think this is something we'd probably want for any component which uses the tokenizer - changes should come from that component in my opinion
> > > 
> > > 
> > > keeping ownership in multi-input rather than tokenizer was a deliberate decision: multi-input gets used with hundreds or thousands of selections, and the fix is to not render that many token elements in the first place. That's about how multi-input manages its data, not about how tokens are laid out. Tokenizer is used in places where selection counts are naturally small. Pushing this into tokenizer would give every consumer an API they don't need, for a problem only multi-input has.
> > 
> > 
> > Multi-input and multi-combobox are so similar, I don't see how this could only happen in multi-input but not multi-combobox. And if it can happen in multi-input, couldn't it happy in any custom component which utilizes tokenizer, built by an application developer?
> 
> @mikerodonnell89 Good question - and I get the instinct, they do look similar from the outside. But I'd flip the framing: we ship them as two separate components precisely because they target different usages. If they covered the same ground they'd be one component, and if they had the same features the implementations wouldn't have diverged the way they have.
> 
> > Multi-combobox's own guidelines back this up:
> > The multi-combo box lets users select one or more options from a predefined list.
> > Don't use the multi-combo box if you want to display large data sets. In this case, use a multi-input field with a select dialog or value help dialog (for more than 1000 items).
> 
> So multi-combobox's docs send the >200-item case over to multi-input - which is exactly why I created virtualization there. Adding the same machinery to multi-combobox would optimize for a scenario it's contractually not supposed to be in. Happy to revisit if a real-world case crosses that line, but I'd rather not pre-build for it.
> 
> On the "any custom tokenizer consumer" angle - tokenizer stays a width-driven layout primitive on purpose. The coordination piece is already generalized: the new hasInternalOverflowIndicator computed is public, so any consumer that genuinely needs count-driven virtualization can wire it up the same way multi-input does, without widening tokenizer's surface for everyone.
> 
> two components = two use cases. Scope stays multi-input; pattern's reusable if/when someone needs it elsewhere.

Ok, I'm still not totally sold here as you could have dozens of multi-comboboxes with 200 tokens on the same page, rendering thousands of tokens. Also doesn't platform multi-input need this functionality too?

droshev:
> > > > > I think this is something we'd probably want for any component which uses the tokenizer - changes should come from that component in my opinion
> > > > 
> > > > 
> > > > keeping ownership in multi-input rather than tokenizer was a deliberate decision: multi-input gets used with hundreds or thousands of selections, and the fix is to not render that many token elements in the first place. That's about how multi-input manages its data, not about how tokens are laid out. Tokenizer is used in places where selection counts are naturally small. Pushing this into tokenizer would give every consumer an API they don't need, for a problem only multi-input has.
> > > 
> > > 
> > > Multi-input and multi-combobox are so similar, I don't see how this could only happen in multi-input but not multi-combobox. And if it can happen in multi-input, couldn't it happy in any custom component which utilizes tokenizer, built by an application developer?
> > 
> > 
> > @mikerodonnell89 Good question - and I get the instinct, they do look similar from the outside. But I'd flip the framing: we ship them as two separate components precisely because they target different usages. If they covered the same ground they'd be one component, and if they had the same features the implementations wouldn't have diverged the way they have.
> > > Multi-combobox's own guidelines back this up:
> > > The multi-combo box lets users select one or more options from a predefined list.
> > > Don't use the multi-combo box if you want to display large data sets. In this case, use a multi-input field with a select dialog or value help dialog (for more than 1000 items).
> > 
> > 
> > So multi-combobox's docs send the >200-item case over to multi-input - which is exactly why I created virtualization there. Adding the same machinery to multi-combobox would optimize for a scenario it's contractually not supposed to be in. Happy to revisit if a real-world case crosses that line, but I'd rather not pre-build for it.
> > On the "any custom tokenizer consumer" angle - tokenizer stays a width-driven layout primitive on purpose. The coordination piece is already generalized: the new hasInternalOverflowIndicator computed is public, so any consumer that genuinely needs count-driven virtualization can wire it up the same way multi-input does, without widening tokenizer's surface for everyone.
> > two components = two use cases. Scope stays multi-input; pattern's reusable if/when someone needs it elsewhere.
> 
> Ok, I'm still not totally sold here as you could have dozens of multi-comboboxes with 200 tokens on the same page, rendering thousands of tokens. Also doesn't platform multi-input need this functionality too?

@mikerodonnell89 i measured 10 multi-comboboxes × 200 tokens each (2000 total selected) on a single page, production build, Apple M4 Pro / Chrome 148:

| Scenario | DOM `<fd-token>` elements | Cold load | Focus-to-paint p95 |
|---|---:|---:|---:|
| 10 × 200 tokens, all collapsed | **120** (12 per instance) | 220 ms | — |
| 10 × 200 tokens, one focused | 120 (unchanged) | — | 19 ms |
| 10 × 10 tokens, all collapsed | 100 (10 per instance) | 204 ms | — |
| 10 × 10 tokens, one focused | 100 (unchanged) | — | 14 ms |

Two things the data shows:

**1. The "thousands of tokens" rendering case doesn't happen.** Tokenizer's width-driven collapse caps each instance at ~12 DOM token elements regardless of selection size — 188 of every 200 selected tokens never materialize as DOM. The 200-token and 10-token examples produce essentially the same render cost because the rendered count is governed by available width, not by selection size.

**2. Multi-combobox has no expansion-on-focus path.** It hardcodes `[showOverflowPopover]="false"` on its internal tokenizer and doesn't call `_showAllTokens()` on input focus. The DOM token count is identical before and after focus. This is the structural reason multi-combobox can't reach the failure mode the multi-input fix addresses — multi-input *does* materialize every selected token on focus, which is the 1000-element paint event we virtualized away. Multi-combobox doesn't have that paint event to begin with.

So the pattern from this PR is a fix for a code path multi-combobox doesn't have. Adding it there would protect against a state that's structurally unreachable today. If the contract changes — if multi-combobox ever grows an expanded view — that's the trigger to revisit, and we'll have the pattern ready.

**Platform multi-input is a different story.** It does call `tokenizer._showAllTokens()` on focus (template line 84), so it has the same expansion path multi-input had. the follow-up is filed and queued to open once this PR merges. You were right to ask.

mikerodonnell89 (changes requested):
With 1000 tokens selected, at certain screen sizes, the "more" count is still way off:

Image

mikerodonnell89 (changes requested):
Also I agree with @MariaIDineva 's earlier comment, this doesn't fully solve the issue, if there are 1000 tokens and the component is focused, it's quite slow. It does need some true virtual scrolling. However this is a great start so I am OK with moving forward (once the above bug is fixed), but I don't think it can be considered to properly close the linked issue

droshev:
> 
Screenshot 2026-06-09 at 08 46 17
Screenshot 2026-06-09 at 08 46 03
ully solve the is

addressed the issue with the wrong +X more and also improved the performance on focus for 1000 tokens

mikerodonnell89 (changes requested):
Seeing the occasional "0 More" when removing tokens (this does appear to be unique to this PR):

https://github.com/user-attachments/assets/7a393d39-ba18-4f2d-a87c-f42cac87a973

mikerodonnell89 (approved):

InnaAtanasova (approved):

PR #14155, opened 2026-04-25
fix: adopt minor styles changes

droshev:
## Summary

Adopts CSS classes and modifiers from fundamental-styles that were missing or outdated in the Angular components

## Icon (`fd-icon`)

- **Added `sap-icon` base class** to `fdBuildIconClass` output — previously only the glyph modifier (`sap-icon--{name}`) was emitted, now the base class is included as required by fundamental-styles
- **Added `size` input** (`IconSize`: `'sm' | 'md' | 'lg' | 'xl' | 'xxl'`) — maps to `sap-icon--{size}` modifier
- **Added `role` and `aria-hidden` accessibility defaults** — decorative icons (no `ariaLabel`) get `role="presentation"` and `aria-hidden="true"`; icons with `ariaLabel` get `role="img"` and no `aria-hidden`
- **No breaking changes** — `ariaLabel`, `ariaHidden`, and existing inputs are unchanged

## Form Label (`fd-form-label`)

- **Added `disabled` signal input** — applies `fd-form-label--disabled` class on host and `aria-disabled` on inner label span
- **Added `unitDescription` signal input** — applies `fd-form-label--unit-description` class (e.g., "kg", "USD" labels)
- **Added `standAlone` signal input** — applies `fd-form-label--stand-alone` class
- **Removed stale SCSS overrides** — 25 lines of CSS marked "remove after adopting fd-styles version 0.39" (current version is 0.41+)

## Form Group (`fd-form-group`)

- **Added `withSpacing` input** — applies `fd-form-group--with-spacing` modifier

## Form Legend (`fd-form-legend`)

- **Added `disabled` signal input** — applies `is-disabled` state class
- **Migrated from `@HostBinding` to `host: {}`** — the directive was trivial (one class binding), so the migration is minimal

### DIFF
--- a/libs/core/form/form-group/form-group.component.ts
+++ b/libs/core/form/form-group/form-group.component.ts
@@ -27,8 +27,7 @@ import { CssClassBuilder, applyCssClass } from '@fundamental-ngx/cdk/utils';
     template: `<ng-content></ng-content>`,
     encapsulation: ViewEncapsulation.None,
     styleUrl: './form-group.component.scss',
-    changeDetection: ChangeDetectionStrategy.OnPush,
-    standalone: true
+    changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class FormGroupComponent implements CssClassBuilder, OnChanges, OnInit {
     /** @hidden */
@@ -40,6 +39,10 @@ export class FormGroupComponent implements CssClassBuilder, OnChanges, OnInit {
      */
     @Input()
     isInline: boolean;
+
+    /** Adds gap spacing between form items. */
+    @Input()
+    withSpacing = false;
 
     /** @hidden */
     class: string;
@@ -53,7 +56,7 @@ export class FormGroupComponent implements CssClassBuilder, OnChanges, OnInit {
      */
     @applyCssClass
     buildComponentCssClass(): string[] {
-        return [this.isInline ? 'fd-form-group--inline' : ''];
+        return [this.isInline ? 'fd-form-group--inline' : '', this.withSpacing ? 'fd-form-group--with-spacing' : ''];
     }
 
     /** @hidden */
--- a/libs/core/form/form-group/form-group.component.ts
+++ b/libs/core/form/form-group/form-group.component.ts
@@ -6,7 +6,8 @@ import {
     Input,
     OnChanges,
     OnInit,
-    ViewEncapsulation
+    ViewEncapsulation,
+    booleanAttribute
 } from '@angular/core';
 import { CssClassBuilder, applyCssClass } from '@fundamental-ngx/cdk/utils';
 
@@ -41,7 +42,7 @@ export class FormGroupComponent implements CssClassBuilder, OnChanges, OnInit {
     isInline: boolean;
 
     /** Adds gap spacing between form items. */
-    @Input()
+    @Input({ transform: booleanAttribute })
     withSpacing = false;
 
     /** @hidden */
--- a/libs/core/form/form-label/form-label.component.ts
+++ b/libs/core/form/form-label/form-label.component.ts
@@ -35,7 +35,6 @@ let formLabelIdCount = 0;
     styleUrl: './form-label.component.scss',
     encapsulation: ViewEncapsulation.None,
     changeDetection: ChangeDetectionStrategy.OnPush,
-    standalone: true,
     imports: [LinkComponent, IconComponent, InlineHelpDirective, NgTemplateOutlet],
     host: {
         '[class.fd-form-label--disabled]': 'disabled()'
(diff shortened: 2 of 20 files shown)


libs/docs/core/input/examples/input-example.component.html
```diff
@@ -23,3 +23,14 @@ <h4 fd-form-header>Examples</h4>
...
+    <label fd-form-label [disabled]="true" for="input-5">Disabled Label</label>
+    <input fd-form-control type="text" id="input-5" placeholder="Field placeholder text" disabled />
+</div>
+<div fd-form-item>
+    <label fd-form-label [unitDescription]="true" for="input-6">kg</label>
+    <input fd-form-control type="text" id="input-6" placeholder="Enter weight" />
+</div>
+<div fd-form-item>
+    <label fd-form-label [standAlone]="true">Stand-Alone Label</label>
+</div>
```
MariaIDineva: Hm, is there some special idea behind adding these examples here? I, personally, don't really see any benefit to that because we have examples with `placeholder` and `disabled` in the `Input States with Messages Components` section. But maybe I am missing the point. I also don't see why we would add an example of a `Stand-Alone Label`.

Remove the compact example

MariaIDineva (approved):

MariaIDineva:
Oh, just one more thing that misled me when I started reviewing the PR - the description says `in the Angular wrappers.` and since we would normally use "wrappers" for the webcomponents wrappers I was looking for changes related to them while there are none. So maybe we can change that to `in Fundamental NGX` or just leave it the phrase out.

libs/core/icon/icon.component.ts
```diff
@@ -109,6 +117,9 @@ export class IconComponent implements HasElementRef {
     /** Icon color */
     readonly background = input<IconColor | null>(null);
 
+    /** Icon size */
+    readonly size = input<IconSize | null>(null);
```
MariaIDineva, line 121: It might be a good idea to add an example in the docs with the new input.

libs/core/form/form-group/form-group.component.ts
```diff
@@ -41,6 +40,10 @@ export class FormGroupComponent implements CssClassBuilder, OnChanges, OnInit {
     @Input()
     isInline: boolean;
 
+    /** Adds gap spacing between form items. */
+    @Input()
+    withSpacing = false;
```
InnaAtanasova, line 46: this can use boolean transform

libs/core/form/form-label/form-label.component.ts
```diff
@@ -120,6 +125,15 @@ export class FormLabelComponent implements OnChanges {
...
     }
 
+    /** Whether the label is disabled. */
+    readonly disabled = input(false, { transform: booleanAttribute });
+
+    /** Whether this label is a unit description (e.g., "kg", "USD"). */
+    readonly unitDescription = input(false, { transform: booleanAttribute });
+
+    /** Whether this label stands alone (self-centering, no end margin). */
+    readonly standAlone = input(false, { transform: booleanAttribute });
```
InnaAtanasova: I wouldn't block this PR for the naming, but `standAlone` is very close to Angular’s `standalone` concept that it could create confusion. I would recommend another name for the input, like selfCentered, independent, centered, etc.

libs/core/icon/icon.component.ts
```diff
@@ -124,6 +135,18 @@ export class IconComponent implements HasElementRef {
     /** Whether or not this icon is for a list navigation item. */
     readonly _navigationItemIcon = signal(false);
 
+    /** @hidden Computed role — "presentation" for decorative, "img" when ariaLabel is set. */
+    protected readonly _role = computed(() => (this.ariaLabel() ? 'img' : 'presentation'));
+
+    /** @hidden Default aria-hidden to true for decorative icons (no ariaLabel). */
+    protected readonly _effectiveAriaHidden = computed<Nullable<boolean>>(() => {
```
InnaAtanasova: protected members are not prefixed with underscore.

InnaAtanasova: we decided to go without Nullable

libs/docs/core/inline-help/examples/inline-help-example.component.html
```diff
@@ -24,7 +24,7 @@
         glyph="question-mark"
         fd-inline-help="Inline Help Tooltip Lorem ipsum, dolor sit amet consectetur adipisicing elit. Officia vitae ipsum facilis quasi similique! Aperiam incidunt perspiciatis quam architecto. Culpa minima odio adipisci tempora reprehenderit necessitatibus, repellendus dolor tempore pariatur?"
         placement="bottom"
-        tabindex="0"
+        [tabindex]="0"
```
InnaAtanasova: You don't need binding for static values. The previous syntax works correctly.

InnaAtanasova:
Comments from Claude review:
- [libs/core/form/form-label/form-label.component.ts:36] standalone: true should be removed (default in
     Angular 19+).
- [libs/core/form/form-group/form-group.component.ts] No test for withSpacing input
- [libs/docs/core/icon/] No example demonstrating size input
- [libs/docs/core/form/] No example demonstrating FormGroup withSpacing

InnaAtanasova (approved):

Issue #14020, opened 2026-03-24
[Core][fd-dynamic-page-header] [fd-card] [fdp-table] ACC-263.3 (Level A)  The title aria-level should can be configured 


Meggie886:
## Summary

The title's aria-level should can be change in fd-dynamic-page-header. Mostly, the fd-dynamic-page-header should be the page's title and the level should be level 1.

Together with fd-card, settings dialog in fdp-table, the aria-level should can be configured

<img width="1148" height="1033" alt="Image" src="https://github.com/user-attachments/assets/a08e2270-34be-4a6d-b177-1745ae3b1123" />
<img width="2207" height="1299" alt="Image" src="https://github.com/user-attachments/assets/672003c2-7ddd-441a-96d7-d7e143abcd79" />

## Expected Behavior

 Provide an attribute to config the title and subtitle's level

<img width="2814" height="904" alt="Image" src="https://github.com/user-attachments/assets/b3d79153-dfd1-4fc9-8cd2-e513042ab9b4" />

## Proposed Solution
 Provide an attribute to config the title and subtitle's level

InnaAtanasova:
Hi @Meggie886 and thanks for creating the issue. I am struggling to understand the problem as it mentions a lot of components so please feel free to add more info to it. 

Let me start with Card component. Card title and Card Subtitle are styled by Angular directives `fd-card-title` and `fd-card-subtitle`. These directives can be added to any heading element and even div, span element. The heading level is left to the application developer to handle. 
In out example we have:
```
<fd-card>
    <h2 fd-card-title>Card Title</h2>
    <h3 fd-card-subtitle>Card Subtitle</h3>
</fd-card>
```
where we just show that in this case the Card has h2 for title and h3 for subtitle. This could easily be anything depending on the project page structure:

```
<fd-card>
    <h3 fd-card-title>Card Title</h2>
    <h4 fd-card-subtitle>Card Subtitle</h3>
</fd-card>
```

Please let me know if this addresses your concern with Card component. 

<img width="1170" height="818" alt="Image" src="https://github.com/user-attachments/assets/480baf79-14ee-44d3-ad1e-553111d7a615" />

InnaAtanasova:
Regarding Platform Dynamic Page, in your example, the heading level is not applied correctly. The heading level is set via `headingLevel` input attribute on `fdp-dynamic-page` level:

[...]

PR #14161, opened 2026-04-28
fix(platform): improve a11y for Platform Table View Settings dialog

InnaAtanasova:
## Related Issue(s)

related to https://github.com/SAP/fundamental-ngx/issues/14020

### The user reported the following:
In the “Settings” and “Filter” dialogs, group labels such as “Sort Order”, “Sort by”, and “Filter By” are marked up as H3 headings, even though they function as form/group labels rather than true section headings. This creates an inaccurate heading structure for screen reader users.
Group labels inside dialogs should be implemented as proper group names/labels (e.g., using for grouped controls or role="group" with aria-labelledby) instead of headings. Headings (H1–H6) should be used only for actual section headings and must follow a logical hierarchy within the dialog/page.
Remove the heading level for “Sort Order”, “Sort by”, and “Filter By”

### The fix provided:
  Replaced `role="heading"` with semantic `<fieldset>` + `<legend>` for "Sort Order" and "Sort by"  groups in the table sorting dialog. Group labels were incorrectly marked as H3 headings, creating a false document structure for screen reader users.                             
                                                                                             
  Changes:                                                                                   
  - Use `<fieldset>` with `<legend>` for proper form control grouping
  - Remove `role="heading"` and `aria-level` attributes                                          
  - Remove unnecessary `aria-describedby` references (fieldset handles this natively)   
  - custom CSS to achieve the old look     
                                                                                             
  Accessibility improvements:                                                                
  - Correct semantic HTML for form grouping                                                  
  - No artificial heading hierarchy                                                          
  - Better screen reader announcements (group context provided by fieldset)  
  
## Screenshots
Screenshot 2026-04-28 at 4 35 38 PM
Screenshot 2026-04-28 at 4 35 16 PM
Screenshot 2026-04-28 at 4 34 36 PM

### DIFF
--- a/libs/platform/table/components/table-view-settings-dialog/filtering/filters-list-step.component.html
+++ b/libs/platform/table/components/table-view-settings-dialog/filtering/filters-list-step.component.html
@@ -5,27 +5,26 @@
 </ng-template>
 
 <ng-template #bodyTemplate>
-    <span
-        role="heading"
-        [attr.aria-level]="headingLevel + 1"
-        fd-list-group-header
-        id="platformTableInternalFilterListGroupHeader"
-    >
-        {{ ('platformTable.filterDialogFilterBy' | fdTranslate)() }}
-    </span>
-    <ul fd-list [navigationIndicator]="true" aria-labelledby="platformTableInternalFilterListGroupHeader">
-        @for (filter of filters; track filter) {
-            <li
-                fd-list-item
-                [interactive]="true"
-                (click)="selectFilter.emit(filter)"
-                (keydown.space)="selectFilter.emit(filter)"
-                (keydown.enter)="selectFilter.emit(filter)"
-            >
-                <a fd-list-link [navigationIndicator]="true">
-                    <span fd-list-title>{{ filter.label }}</span>
-                </a>
-            </li>
-        }
-    </ul>
+    <fieldset>
+        <legend fd-list-group-header>
+            <span fd-list-title id="platformTableInternalFilterListGroupHeader">
+                {{ ('platformTable.filterDialogFilterBy' | fdTranslate)() }}
+            </span>
+        </legend>
+        <ul fd-list aria-labelledby="platformTableInternalFilterListGroupHeader" [navigationIndicator]="true">
+            @for (filter of filters; track filter) {
+                <li
+                    fd-list-item
+                    [interactive]="true"
+                    (click)="selectFilter.emit(filter)"
+                    (keydown.space)="selectFilter.emit(filter)"
+                    (keydown.enter)="selectFilter.emit(filter)"
+                >
+                    <a fd-list-link [navigationIndicator]="true">
+                        <span fd-list-title>{{ filter.label }}</span>
+                    </a>
+                </li>
+            }
+        </ul>
+    </fieldset>
 </ng-template>
--- a/libs/platform/table/components/table-view-settings-dialog/filtering/filters-list-step.component.html
+++ b/libs/platform/table/components/table-view-settings-dialog/filtering/filters-list-step.component.html
@@ -5,26 +5,24 @@
 </ng-template>
 
 <ng-template #bodyTemplate>
-    <fieldset>
-        <legend fd-list-group-header>
-            <span fd-list-title id="platformTableInternalFilterListGroupHeader">
-                {{ ('platformTable.filterDialogFilterBy' | fdTranslate)() }}
-            </span>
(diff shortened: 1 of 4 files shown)


libs/platform/table/components/table-view-settings-dialog/grouping/grouping.component.html
```diff
@@ -1,56 +1,61 @@
...
-    @for (column of columns(); track column) {
-        <li fd-list-item [selected]="column.key === field()" [attr.aria-describedby]="groupByHeaderId">
+<fieldset>
+    <legend fd-list-group-header>
+        <span fd-list-title [attr.id]="groupOrderHeaderId">
+            {{ ('platformTable.groupDialogGroupOrderHeader' | fdTranslate)() }}
+        </span>
+    </legend>
+    <ul fd-list [selection]="true">
+        <li fd-list-item [selected]="direction() === SORT_DIRECTION.ASC" [attr.aria-describedby]="groupOrderHeaderId">
```
MariaIDineva: Maybe all those `aria-describedby` attributes on the `li` items are not needed any more since they lead to duplicate screen reader announcements:

Image

Same for the `aria-describedby` attributes on the sorting component below.

libs/platform/table/components/table-view-settings-dialog/sorting/sorting.component.html
```diff
@@ -1,63 +1,67 @@
-<span role="heading" [attr.aria-level]="headingLevel() + 1" fd-list-group-header>
```
MariaIDineva, line 1: Since we are no longer using `headingLevel` in the template we might as well remove the input.

libs/platform/table/components/table-view-settings-dialog/grouping/grouping.component.html
```diff
@@ -1,56 +1,61 @@
-<span role="heading" [attr.aria-level]="headingLevel() + 1" fd-list-group-header>
```
MariaIDineva, line 1: Since we are no longer using headingLevel in the template we might as well remove the input.

libs/platform/table/components/table-view-settings-dialog/filtering/filters-list-step.component.html
```diff
@@ -5,25 +5,24 @@
...
-                [interactive]="true"
-                (click)="selectFilter.emit(filter)"
-                (keydown.space)="selectFilter.emit(filter)"
-                (keydown.enter)="selectFilter.emit(filter)"
-            >
-                <span fd-list-title>{{ filter.label }}</span>
-            </li>
-        }
-    </ul>
+    <fieldset>
```
MariaIDineva: And one comment from Claude:

> The list in this template is a navigation list (each <li fd-list-item [interactive]="true"> fires selectFilter.emit(filter) to drill into a sub-filter view). It contains no form controls. Wrapping it in "**fieldset/legend**" is semantically incorrect — **fieldset** is for grouping form controls. The appropriate semantic for a labelled, non-form list group is role="group" + aria-labelledby, or simply keeping the existing aria-labelledby on the **ul** without **fieldset**. This is an accessibility regression compared to the original structure for screen readers that communicate fieldset boundaries with "group start/end" cues when encountering form widgets.

libs/platform/table/components/table-view-settings-dialog/sorting/sorting.component.html
InnaAtanasova: `headingLevel` is needed for the title of the dialog. The user needs to be able to set this based on the entire page structure. Before we used this to set the list group heading levels, which I removed and you see in the template. But the `headingLevel` is still needed for the dialog heading.

libs/platform/table/components/table-view-settings-dialog/grouping/grouping.component.html
InnaAtanasova: answered above.

libs/platform/table/components/table-view-settings-dialog/filtering/filters-list-step.component.html
InnaAtanasova: Claude was very convincing fieldset is the best approach for this problem :D

MariaIDineva: Oh, interesting... 😆

MariaIDineva (approved):
🚢

PR #10158, opened 2023-07-11
chore(tooling): introduced new release creation ci workflow

g-cheishvili:
## Related Issue(s)

closes #10068

## Description

Introduced new CI workflow for creation of the hotfixes and regular releases

PR #14170, opened 2026-04-29
chore(ci): exclude current version tag from deletion in release notes generation

droshev:
The closest-version helper included the current version's own tag in tagsTillClosest. index.js then deleted it locally, leaving conventional-changelog with no anchor for releaseCount:1 — producing empty release notes. This was latent since PR #10158 (Jul 2023) but only surfaced with v0.62.0's 103 RC tags.

### DIFF
--- a/.github/actions/generate-conventional-release-notes/action.yml
+++ b/.github/actions/generate-conventional-release-notes/action.yml
@@ -2,8 +2,8 @@ name: Generate changelog
 description: Generates changelog from nearest release to current commit
 
 outputs:
-    generatedChangelog:
-        description: Generated changelog
+    generatedReleaseNotes:
+        description: Generated release notes body
 
 runs:
     using: node20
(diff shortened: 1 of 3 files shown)


MariaIDineva (approved):
A comment from Claude:

> **[.github/actions/generate-conventional-release-notes/index.js:9]** Pre-existing issue (not introduced here): git tag -d ${tagsTillClosest.join(' ')} will fail with **fatal: tag name required** when tagsTillClosest is empty. The new test case "no RC tags between current and closest" returns tagsTillClosest: [], making this path more likely to be hit. Consider guarding with if (tagsTillClosest.length > 0) in index.js. This is out of scope for this PR but worth tracking.

InnaAtanasova (approved):

droshev:
> A comment from Claude:
> 
> > **[.github/actions/generate-conventional-release-notes/index.js:9]** Pre-existing issue (not introduced here): git tag -d ${tagsTillClosest.join(' ')} will fail with **fatal: tag name required** when tagsTillClosest is empty. The new test case "no RC tags between current and closest" returns tagsTillClosest: [], making this path more likely to be hit. Consider guarding with if (tagsTillClosest.length > 0) in index.js. This is out of scope for this PR but worth tracking.

addressed

PR #14181, opened 2026-05-04
fix(platform): table with selection reading select all column header …

### LINKED ISSUE #14140: [P2][Sourcing][Accessibility] Platform Table Check box incorrect information read issue
### Describe your proposal

On accessing any check box within the tables, same information “Select all” is read which is incorrect. Here only header row check box is a ‘Select all’ check box. Information, state, properties of all elements should be clearly read to the users when accessed.
Replicable: https://docs-0-57--fundamental-ngx.netlify.app/#/platform/table#tree

ng: 20, F-NGX platform: 0.57.6

### Affected Component

_No response_

On accessing any check box within the tables, same information “Select all” is read which is incorrect. Here only header row check box is a ‘Select all’ check box. Information, state, properties of all elements should be clearly read to the users when accessed.
Replicable: https://docs-0-57--fundamental-ngx.netlify.app/#/platform/table#tree

### StackBlitz example(s)

_No response_

### Priority

High

### Declaration

mikerodonnell89:
fixes #14140 

Fixes a bug where the JAWS screenreader would read "Select all" or "Deselect all" when focusing any row with selection on the platform table. This was because JAWS would attempt to announce a column header for this row, in this case, grabbing the title attribute from the column header cell. Also fixes a bug where "Select all" or "Deselect all" would be read on the single-selection platform table column header


libs/platform/table/components/table-header-row/table-header-row.component.ts
```diff
@@ -161,6 +161,9 @@ export class TableHeaderRowComponent extends TableRowDirective implements OnInit
         this.setItems(items.toArray());
     }
 
+    /** @hidden */
+    _selectionColumnHeaderDisableTitle = true;
```
MariaIDineva: I would use a signal in this case, since it's a field used in the template. Maybe also add a comment to specify why this field is added as it is not immediately obvious.

MariaIDineva (approved):

mikerodonnell89:
hotfix will be published to v0.58.7

droshev:
the main comment from claude:

Findings

  Minor — Title hiding via mouse/focus events is fragile

  The approach of toggling _selectionColumnHeaderDisableTitle via mouseenter/mouseleave/focusin/focusout means:

  1. Keyboard-only users who navigate to the checkbox via Tab will see the title (focusin fires), but if focus moves to a child element and back, the
   signal toggles rapidly.
  2. If focus moves directly from within the <th> to outside (e.g., skip-nav or programmatic focus), focusout fires correctly. But if a
  screenreader's virtual cursor lands on the <th> without triggering a true focusin DOM event, the title remains empty and the screenreader gets no
  column-level context.

  Suggestion: Consider using aria-label on the <th> element itself (which screen readers prefer for column headers) instead of title. This would
  eliminate the need for the toggle signal entirely — aria-label is read as the column header when navigating cells, but unlike title, it doesn't get
   echoed in data rows by JAWS. If title must stay for tooltip purposes, the current approach is acceptable but should have a test validating the
  hover/focus toggle behavior.

Minor (M3) — Empty string vs null for [attr.title]

  selectionMode === SELECTION_MODE.SINGLE || _selectionColumnHeaderDisableTitle()
      ? ''
      : ...

  When [attr.title]="", the attribute is rendered as title="" in the DOM. For [attr.title], Angular removes the attribute entirely when the value is
  null. An empty title="" is technically still present in the DOM and some assistive tech may interpret it differently than an absent attribute.
  Consider using null instead of '' to fully remove the attribute:

  selectionMode === SELECTION_MODE.SINGLE || _selectionColumnHeaderDisableTitle()
      ? null
      : ...

  Same applies to aria-description

mikerodonnell89:
> If title must stay for tooltip purposes, the current approach is acceptable but should have a test validating the hover/focus toggle behavior.

Yep, that is the reason for using title and not aria-label

libs/platform/table/components/table-header-row/table-header-row.component.ts
```diff
@@ -161,6 +162,12 @@ export class TableHeaderRowComponent extends TableRowDirective implements OnInit
         this.setItems(items.toArray());
     }
 
+    /**
+     * @hidden
+     * Used to remove the selection column title, to prevent screenreader from reading "Select All" when focusing non-header cells
+     * */
+    _selectionColumnHeaderDisableTitle = signal(true);
```
InnaAtanasova: This should not be prefixed with an underscore if it's not a private member.

PR #14190, opened 2026-05-08
fix(core): update touch area for Switch component

### LINKED ISSUE #14185: margin-inline-end in  .fd-switch CSS is overwritten in the same style
### Describe your proposal

In fundamental-styles 0.41.5, the .fd-switch CSS rule sets margin-inline-end: .5rem early, but then margin-inline: 0 later in the same rule overwrites it — removing the gap between the switch and its label.

angular 21 FD 62.1

### Affected Component

fd-switch

add a fd-switch, see the styles inside '.fd-switch' and see the overwrite

### StackBlitz example(s)

_No response_

### Priority

Low

### Declaration

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/14185


droshev (approved):

MariaIDineva (approved):

InnaAtanasova:
closing because we will bring the latest version of fund-styles: https://github.com/SAP/fundamental-ngx/pull/14198

PR #14192, opened 2026-05-10
feat(core): add headingLevel support to grid-list title bar

### LINKED ISSUE #13537: Add option to set heading level for Toolbar in Grid List
Currently, the API doesn't allow to set heading level for the Toolbar title in Grid List

droshev:
closes https://github.com/SAP/fundamental-ngx/issues/13537

**Key files to review:**

1. **`grid-list-title-bar.component.ts`** — Core change. New `headingLevel` signal input + `normalizedLevel` computed that strips the `h` prefix (e.g. `'h4'` → `4`). Check that the default value (`2`) and normalization logic are correct.

2. **`grid-list-title-bar.component.html`** — `<label>` replaced with `<span>` + `role="heading"` + `[attr.aria-level]`. Verify the a11y semantics are correct (implicit heading via ARIA role instead of native `<h2>`–`<h6>`).

**How to verify:**

- Check the docs example at `/grid-list` → "Heading Level" tab (`yarn start` then navigate)
- Inspect the rendered DOM: the title `<span>` should have `role="heading"` and `aria-level="<N>"`

### DIFF
--- a/libs/core/grid-list/components/grid-list-title-bar/grid-list-title-bar.component.spec.ts
+++ b/libs/core/grid-list/components/grid-list-title-bar/grid-list-title-bar.component.spec.ts
@@ -1,24 +1,174 @@
+import { Component } from '@angular/core';
 import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
-import { GridListModule } from '../../grid-list.module';
+import { By } from '@angular/platform-browser';
 import { GridListTitleBarComponent } from './grid-list-title-bar.component';
 
+@Component({
+    template: `<fd-grid-list-title-bar [title]="title"></fd-grid-list-title-bar>`,
+    imports: [GridListTitleBarComponent]
+})
+class GridListTitleBarDefaultTestComponent {
+    title = 'Products';
+}
+
+@Component({
+    template: `<fd-grid-list-title-bar [title]="title" [headingLevel]="headingLevel"></fd-grid-list-title-bar>`,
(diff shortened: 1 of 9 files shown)


libs/core/grid-list/components/grid-list-title-bar/grid-list-title-bar.component.spec.ts
```diff
@@ -1,24 +1,127 @@
...
+        const titleBar = fixture.debugElement.query(By.directive(GridListTitleBarComponent));
+        expect(titleBar).toBeTruthy();
+    });
+
+    it('should render role="heading" on the title label', () => {
+        const label = fixture.debugElement.query(By.css('[fd-toolbar-label]'));
+        expect(label.nativeElement.getAttribute('role')).toBe('heading');
+    });
+
+    it('should render aria-level="2" by default', () => {
```
MariaIDineva: This test is rather misleading. It does not really test that 2 is the default heading level, because we explicitly set it to 2 on line 12 of this file. If we were to test the default value for real, we should not be passing any value to the `headingLevel` input.
Also, we could test edge cases like setting the input to an invalid value, eg. `0` or `h7` and test that those also default to 2.

MariaIDineva (approved):

libs/core/grid-list/components/grid-list-title-bar/grid-list-title-bar.component.spec.ts
droshev: updated

Issue #14158, opened 2026-04-27
Evaluate and improve @fundamental-ngx/mcp server + create complementary Skills


droshev:
## Summary

The `@fundamental-ngx/mcp` server provides AI coding assistants with structured access to 1000+ components across 8 libraries. It has 9+ tools covering component APIs, examples, design tokens, accessibility, migration, and comparison. It has not yet been evaluated through real-world usage in an application.

This issue covers: use the MCP server to build real Angular UI, identify gaps, create complementary Skills for procedural knowledge the MCP server cannot provide, and propose improvements.

## Background

**What the MCP server does today:**
- Tools: `list_components`, `search_components`, `get_component_api`, `get_component_examples`, `recommend_components`, `get_migration_guide`, `get_design_tokens`, `get_accessibility_guide`, `compare_components`, `get_usage_guide`
- Metadata from Custom Elements Manifests (CEM) + TypeDoc, covering ~1000 components
- 131 tests across 5 test suites

**What the MCP server does NOT cover:**
MCP tools answer "what exists?" and "what does this component's API look like?" They do not encode "how do I compose a master-detail page" or "how do I migrate from @Input decorators to signal inputs across a library." That procedural knowledge belongs in Skills.

The repo already has excellent agent guides in `docs/agents/` and `.claude/rules/` — the question is whether some of that knowledge should also be packaged as installable Skills for consumers outside the repo.

## Part 1: Real-World Usage Test

Set up the MCP server in a fresh Angular project and attempt to build real UI.

### Setup
- [x] Scaffold a new Angular 21+ app
- [x] Configure the MCP server in Claude Code or VS Code
- [x] Verify all tools respond

### Build Scenarios
- [x] **Simple form** — "Build a login form with email, password, and submit using @fundamental-ngx/core"
- [x] **Data table** — "Create a platform table with sorting, filtering, and row selection"
- [x] **Dialog with form** — "Build a dialog that contains a reactive form with validation states"
- [x] **Page layout** — "Create a page with dynamic page header, side navigation, and content area"
- [x] **Complex composition** — "Build a master-detail layout with a platform list on the left and detail view on the right"
- [x] **UI5 Web Components** — "Build a form using UI5 Web Component wrappers (ui5-input, ui5-button)"

For each: did the AI use MCP tools? Was the generated code correct (imports, standalone components, signal patterns)? What was missing?

## Part 2: Tool-by-Tool Evaluation

[...]

PR #14208, opened 2026-05-14
feat(mcp): MCP server evaluation part 2

MariaIDineva:
## Related Issue(s)

part of https://github.com/SAP/fundamental-ngx/issues/14158

## Description

# Part 1: Real-World Usage Test — Findings

**Date:** 2026-05-13  
**Tester:** Claude Code (automated)  
**MCP server version:** 0.62.0-rc.67  
**Method:** Node.js MCP client connecting to the built server via stdio; 6 scenarios, ~4 tool calls each.

---

## Scenarios Tested

| #   | Scenario                             | Tools exercised                                                                         |
| --- | ------------------------------------ | --------------------------------------------------------------------------------------- |
| 1   | Login form                           | `recommend_components`, `get_usage_guide` ×3                                            |
| 2   | Data table                           | `recommend_components`, `get_usage_guide`, `compare_components`, `get_component_api`    |
| 3   | Dialog with form                     | `get_usage_guide`, `get_accessibility_guide`                                            |
| 4   | Page layout (DynamicPage)            | `search_components`, `get_usage_guide`, `get_component_api`, `recommend_components`     |
| 5   | Master-detail (FlexibleColumnLayout) | `search_components`, `get_component_api`, `recommend_components`, `get_usage_guide`     |
| 6   | UI5 Web Component form               | `list_components`, `get_component_api`, `compare_components`, `get_accessibility_guide` |

---

### Curated usage guides are excellent

`get_usage_guide` for `dialog`, `table`, `button`, `layout-grid`, and `card` all returned sharp decision trees with concrete code examples and realistic pitfalls. The dialog guide covers all three API surfaces (template-ref, component, object-based) and correctly distinguishes `fd-dialog` vs `ui5-dialog` vs `fd-message-box`. The table guide cleanly separates fd-table (manual), fdp-table (feature-rich), and ui5-table (Fiori-native). These guides would save a developer 20–30 minutes of digging.

### `recommend_components` handles most patterns correctly

"A filterable sortable data table with pagination" → `fdp-table`, `fdb-search-field`, `fd-pagination`. Exactly right.  
"Master detail split view with list and detail panel" → `fd-flexible-column-layout`, `ui5-flexible-column-layout`. Correct.

### `compare_components` delivers immediate signal

The input count diff (fd-table: 7 inputs vs fdp-table: 44 inputs, 0 shared outputs) communicates the complexity tradeoff without reading the full API. The summary line is concise and useful.

### `list_components` with category filtering works well

`library=ui5, category=Form` returned 25 correctly-categorised components with zero noise.

### `get_accessibility_guide` for UI5 components is solid

`ui5-input` returned 4 correct accessible inputs (`accessibleName`, `accessibleNameRef`, `accessibleDescription`, `accessibleDescriptionRef`) with actionable, specific tips.

---

### 1. `fd-input` silently resolves to `fd-input-group` — HIGH ✅ Fixed

`get_usage_guide("fd-input")` and `compare_components("ui5-input", "fd-input")` both resolved to `fd-input-group`. Investigation revealed there is no `fd-input` element selector in the library — the plain input field is `input[fd-form-control]`. The silent misdirection came from `findComponent`'s partial-match filter: `"fd-input-group".includes("fd-input")` is `true`, so `InputGroupComponent` was the top-scoring partial hit.

**Reproduction:**

```
get_usage_guide("fd-input")
→ Was returning fd-input-group, not a "not found" response
```

**Fix (`libs/mcp-server/src/server.ts`):**

Added a `selectorHasBoundaryMatch(selector, query)` helper that tests the query against the selector using a negative lookahead — the query must not be immediately followed by a `-`:

```typescript
function selectorHasBoundaryMatch(selector: string, query: string): boolean {
    const escaped = query.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
    return new RegExp(escaped + '(?!-)').test(selector);
}
```

The partial-match filter in `findComponent` now uses this instead of `String.includes`:

```typescript
// Before
(c) => c.selector.toLowerCase().includes(lower) || ...

// After
(c) => selectorHasBoundaryMatch(c.selector.toLowerCase(), lower) || ...
```

This prevents prefix-segment false matches (`fd-input` → `fd-input-group`, `fd-table` → `fd-table-cell`) while leaving all legitimate partial matches intact. The result is that `get_usage_guide("fd-input")` now returns an honest "not found" message, prompting the caller to use `search_components` to discover what is actually available.

Three regression tests were added to `server.spec.ts` covering: `fd-input`/`fd-input-group`, `fd-table`/`fd-table-cell`, and the positive case (querying `fd-input-group` by full name still resolves correctly).

### 2. `search_components` ranking breaks for multi-word queries — HIGH ✅ Fixed

`search_components("master detail flexible column")` returned 0 results (after fix 1) or wrong list sub-directives (before). Root cause: `scoreMatch` looks for the full query string as a single substring — no selector or description ever contains `"master detail flexible column"` verbatim, so every component scores 0.

**Reproduction:**

```
search_components("master detail flexible column")
→ count: 0, results: []
```

**Fix (`libs/mcp-server/src/server.ts`):**

In the `search_components` handler, multi-word queries now split on whitespace and **sum the per-word scores**. Components matching more words rank higher; the full-string path is preserved for single-word queries:

```typescript
const queryWords = lowerQuery.split(/\s+/).filter((w) => w.length > 2);
const isMultiWord = queryWords.length > 1;

const scored = components
    .map((c) => ({
        component: c,
        score: isMultiWord
            ? queryWords.reduce((sum, word) => sum + scoreMatch(c, word), 0)
            : scoreMatch(c, lowerQuery)
    }))
    ...
```

After the fix, `search_components("master detail flexible column")` returns 20 results with `fd-flexible-column-layout` (score 284) at # 1 and `ui5-flexible-column-layout` (score 242) at # 2. Four tests were added: two fixture-based (multi-word ranking, zero-score guard) and two against the real catalog (`"flexible column"` and the full four-word query).

### 3. `search_components` surfaces sub-components above primary components — MEDIUM ✅ Fixed (side effect of # 2)

`search_components("dynamic page")` previously returned `[fdDynamicPageHeaderSubtitle]` and `[fdDynamicPageHeaderTitle]` as the top results, not `fd-dynamic-page`. Helper directives and slot directives consistently outranked the component they belong to.

**Reproduction:**

```
search_components("dynamic page")
→ Was: top result [fdDynamicPageHeaderSubtitle], not fd-dynamic-page
→ Now: fd-dynamic-page (301) #1, fdp-dynamic-page (299) #2, sub-directives #3+
```

**Why fix # 2 resolved this:** "dynamic page" is a two-word query, so the new code scores each word independently. The existing length-proximity bonus in `scoreMatch` then favours shorter selectors — `fd-dynamic-page` (14 chars) gets a much larger bonus for matching "dynamic" (7 chars) than `[fdDynamicPageHeaderSubtitle]` (29 chars) does, because the query covers a larger fraction of the primary selector.

### 4. `recommend_components` false positives via substring matching — MEDIUM ✅ Fixed

"A login form with username and password" triggered the `avatar|user|profile` pattern because "username" contains "user". This added `fd-avatar`, `fd-avatar-group`, and `ui5-avatar` to a login-form recommendation. The keyword fallback in `recommend_components` needs whole-word matching instead of `lowerDesc.includes(word)`.

**Reproduction:**

```
recommend_components("a login form with username and password")
→ Was: Recommends fd-avatar, fd-avatar-group, ui5-avatar
→ Now: No avatar components
```

**Fix (`libs/mcp-server/src/server.ts`):**

Added a `patternKeywordMatches(keyword, text)` helper that applies word-boundary regex for single-word keywords and falls back to plain `includes` for multi-word phrases:

```typescript
function patternKeywordMatches(keyword: string, text: string): boolean {
    if (keyword.includes(' ')) {
        return text.includes(keyword);
    }
    const escaped = keyword.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
    return new RegExp(`\\b${escaped}\\b`).test(text);
}
```

The keyword check in the `recommend_components` UI_PATTERNS loop was updated from `lowerDesc.includes(kw)` to `patternKeywordMatches(kw, lowerDesc)`. This prevents `"user"` from firing on `"username"` while still matching `"user profile page"` correctly. Six unit tests were added covering: whole-word match, prefix rejection (`"user"` / `"username"`), embedded rejection (`"form"` / `"platform"`), punctuation boundary, multi-word phrase match, and multi-word partial miss.

### 5. `get_accessibility_guide` for `fd-dialog` returns generic fallback — MEDIUM ✅ Fixed

`get_accessibility_guide("fd-dialog")` returned zero ARIA inputs and the generic tip "This component has no explicit ARIA inputs. It likely handles accessibility internally." But the dialog usage guide itself lists `ariaLabelledBy` as a required pitfall. The ARIA input detector only matches properties whose name starts with `aria` or equals `role` or starts with `accessible`; anything defined on child components or slots is missed.

**Root cause:** `fd-dialog` exposes a single input — `dialogConfig: DialogConfig<any>`. All ARIA configuration (including `ariaLabelledBy`) is passed through that config object. The CEM metadata does not flatten `DialogConfig` properties into the component's inputs, so the detector finds nothing.

**Reproduction:**

```
get_accessibility_guide("fd-dialog")
→ Was: ariaInputs: [], tips: ["...handles accessibility internally..."]
→ Now: ariaInputs: [], tips: [
    "ARIA warnings — always provide ariaLabelledBy pointing to the dialog title element ID.",
    "Accessibility properties (ariaLabelledBy, ariaDescribedBy, etc.) are set on the \"dialogConfig\" input via DialogConfig. Check the DialogConfig interface for available ARIA properties."
  ]
```

**Fix (`libs/mcp-server/src/server.ts`):**

Two supplementary lookups were added inside `get_accessibility_guide` after the existing `ariaInputs` filter:

1. **Curated guide pitfalls** — looks up the curated usage guide for the component (trying `selector` key first, then the selector with its library prefix stripped: `"fd-dialog"` → `"dialog"`). Any pitfall containing `"aria"` or `"accessible"` is appended to `tips`. For `fd-dialog`, this surfaces: _"ARIA warnings — always provide ariaLabelledBy pointing to the dialog title element ID."_

2. **Config-input hint** — when `ariaInputs` is empty and the component has an input whose type contains `"Config"`, a targeted tip is added that names the config input and type, directing the developer to that interface instead of the generic fallback.

The generic "handles accessibility internally" message is now only shown when both lookups also find nothing. Five unit tests were added covering: curated pitfall extraction, stripped-key lookup, empty case, config-input detection, and real-catalog validation of `fd-dialog`.

### 6. `fd-dynamic-page` exposes `collapsed: Signal<boolean>` as an input type — MEDIUM ✅ Fixed

The component API for `fd-dynamic-page` lists `collapsed` with type `Signal<boolean>`. Signal wrappers are implementation details, not valid `input()` types. This is a metadata extraction artifact from TypeDoc reading signal property wrappers as type annotations. An AI consuming this API would try to bind `[collapsed]="someSignal"` instead of `[collapsed]="someBoolean"`.

**Root cause:** `Signal<T>` and `WritableSignal<T>` are public state-signal properties (internal reactive state readable by consumers, not bindable `@Input()` members). TypeDoc exposes them as public properties, and the extractor's `isLikelyLegacyInput` heuristic classified them as legacy `@Input()` values because they were not in the `INTERNAL_REFERENCE_NAMES` blocklist. In total, 41 such properties existed across 16 components in the catalog.

**Reproduction:**

```
get_component_api("fd-dynamic-page")
→ Was: inputs includes { name: "collapsed", type: "Signal<boolean>" }
→ Now: collapsed not in inputs list
```

**Fix (two-layer):**

1. **`libs/mcp-server/src/extractors/typedoc-extractor.ts`** — `Signal` and `WritableSignal` were added to `INTERNAL_REFERENCE_NAMES`. Future catalog regenerations will not include these properties as inputs. Angular's actual signal-input types (`InputSignal<T>`, `InputSignalWithTransform<T,U>`, `ModelSignal<T>`) remain handled correctly.

2. **`libs/mcp-server/src/server.ts`** — At catalog load time, all inputs whose type matches `/^(Writable)?Signal</` are filtered out. This fixes the 41 already-committed bad entries without needing to regenerate `components.json`. A new `isSignalWrapperType(type)` helper encapsulates the pattern check.

Seven unit tests were added: 6 covering `isSignalWrapperType` directly (positive, negative, and edge cases) and 1 real-catalog test confirming `fd-dynamic-page` has signal inputs in the raw data that are then removed by the filter.

### 7. No curated guide for `fd-flexible-column-layout` — LOW ✅ Fixed

`get_usage_guide("fd-flexible-column-layout")` falls through to the auto-generated path, which returns only `component: "FlexibleColumnLayoutComponent"` with a minimal slot stub.

**Fix (`libs/mcp-server/src/data/usage-guides.ts`):**

Added a `'fd-flexible-column-layout'` entry to `USAGE_GUIDES` covering:

- Decision tree for 1/2/3-column layouts with concrete layout enum values and example templates
- `fd-flexible-column-layout` vs `ui5-flexible-column-layout` decision node
- All 10 `FlexibleColumnLayout` enum values listed in `commonPitfalls` for reference
- Content projection pattern — `<ng-template #startColumn>`, `<ng-template #midColumn>`, `<ng-template #endColumn>` (direct children, NOT named slot attributes)
- All 6 required title inputs (`collapseTitle`, `collapseTitleStartBtn`, `collapseTitleEndBtn`, `expandTitle`, `expandTitleStartBtn`, `expandTitleEndBtn`) and `separatorAriaLabel` flagged as required pitfalls
- `layoutDefinitions` (required `FlexibleColumnLayoutDefinition[]`) documented in pitfalls
- Full `compositionPattern` with signal-based `layout` and import statement

Five unit tests were added to `server.spec.ts` verifying: guide existence, all 10 enum values present, template reference projection in composition pattern, required-input pitfall coverage, and `ui5-flexible-column-layout` in related components.

### 8. Auto-generated pitfall over-reports required inputs — LOW ✅ Fixed

`get_usage_guide("fd-form-group")` produces: "Required inputs with no default: isInline. Omitting these will cause runtime errors." `isInline: boolean` is almost certainly optional (defaults to `false` — the description even says so). The required-input detection is too aggressive and will generate misleading warnings for most boolean inputs that happen to have no explicit `defaultValue` in the CEM metadata.

**Root cause:** The filter in `buildPitfalls` (and `buildTemplate`) was `i.required && !i.defaultValue`. This flags any input marked `required: true` in the catalog with no stored default, including every boolean input that naturally defaults to `false`.

**Fix (`libs/mcp-server/src/utils/selector-utils.ts`):**

Added `isTriviallyOptionalType(type)` helper and applied it to both filter sites:

```typescript
function isTriviallyOptionalType(type: string | undefined): boolean {
    if (!type) return false;
    return type.includes('boolean') || type.includes('undefined');
}

// In buildPitfalls and buildTemplate:
const requiredInputs = component.inputs.filter(
    (i) => i.required && !i.defaultValue && !isTriviallyOptionalType(i.type)
);
```

- `boolean` → excluded: Angular treats a missing boolean binding as `false`, never a crash
- union with `undefined` → excluded: the type author explicitly marked it as optional

Four unit tests were added: boolean excluded, `boolean | undefined` excluded, `string | undefined` excluded, and a positive case (`DataSource` type still fires the pitfall).

---

### DIFF
--- a/libs/mcp-server/src/data/usage-guides.ts
+++ b/libs/mcp-server/src/data/usage-guides.ts
@@ -358,5 +358,121 @@ this._dialogService.open(MyDialogComponent, {
   </fd-card-content>
 </fd-card>`,
         relatedComponents: ['fd-card-header', 'fd-card-content', 'fd-avatar', 'fd-tile', 'ui5-card']
+    },
+
+    'fd-flexible-column-layout': {
+        component: 'fd-flexible-column-layout',
+        summary:
+            'SAP Fiori master-detail layout that transitions between 1, 2, and 3 columns. Controlled by the FlexibleColumnLayout enum via the layout input. Content is projected via #startColumn, #midColumn, and #endColumn template reference variables.',
+        decisionTree: [
+            {
+                question: 'How many columns does your layout need?',
+                options: [
+                    {
+                        answer: 'Single full-screen column (initial/focused state)',
+                        recommendation:
+                            'Use layout="OneColumnStartFullScreen", "OneColumnMidFullScreen", or "OneColumnEndFullScreen" depending on which panel is active.',
+                        example: `<fd-flexible-column-layout layout="OneColumnStartFullScreen"
+  [layoutDefinitions]="layoutDefs"
+  collapseTitle="Collapse" collapseTitleStartBtn="Collapse start"
+  collapseTitleEndBtn="Collapse end"
+  expandTitle="Expand" expandTitleStartBtn="Expand start"
+  expandTitleEndBtn="Expand end"
+  separatorAriaLabel="Column separator">
+  <ng-template #startColumn>
+    <!-- Master list -->
+  </ng-template>
+</fd-flexible-column-layout>`
+                    },
+                    {
+                        answer: 'Two-column master-detail',
+                        recommendation:
+                            '"TwoColumnsStartExpanded" gives 67/33 split (list dominant). "TwoColumnsMidExpanded" gives 33/67 split (detail dominant). "TwoColumnsEndExpanded" gives 33/67 with end panel.',
+                        example: `<fd-flexible-column-layout [layout]="layout()"
+  [layoutDefinitions]="layoutDefs"
+  collapseTitle="Collapse" collapseTitleStartBtn="Collapse start"
+  collapseTitleEndBtn="Collapse end"
+  expandTitle="Expand" expandTitleStartBtn="Expand start"
+  expandTitleEndBtn="Expand end"
+  separatorAriaLabel="Column separator">
+  <ng-template #startColumn><!-- Master list --></ng-template>
+  <ng-template #midColumn><!-- Detail view --></ng-template>
+</fd-flexible-column-layout>`
+                    },
+                    {
+                        answer: 'Three-column master-detail-extra',
+                        recommendation:
+                            '"ThreeColumnsMidExpanded" gives 25/50/25. "ThreeColumnsEndExpanded" gives 25/25/50. "ThreeColumnsStartMinimized" gives 0/67/33. "ThreeColumnsEndMinimized" gives 67/33/0.',
+                        example: `<fd-flexible-column-layout [layout]="layout()"
+  [layoutDefinitions]="layoutDefs"
+  collapseTitle="Collapse" collapseTitleStartBtn="Collapse start"
+  collapseTitleEndBtn="Collapse end"
+  expandTitle="Expand" expandTitleStartBtn="Expand start"
+  expandTitleEndBtn="Expand end"
+  separatorAriaLabel="Column separator">
+  <ng-template #startColumn><!-- Master list --></ng-template>
+  <ng-template #midColumn><!-- Detail --></ng-template>
+  <ng-template #endColumn><!-- Extra detail --></ng-template>
+</fd-flexible-column-layout>`
+                    }
+                ]
+            },
+            {
+                question: 'fd-flexible-column-layout vs ui5-flexible-column-layout?',
+                options: [
+                    {
+                        answer: 'Angular-native with full control over layout percentages',
+                        recommendation:
+                            'Use fd-flexible-column-layout. Customize column widths via layoutDefinitions (FlexibleColumnLayoutDefinition array).',
+                        example: `import { FlexibleColumnLayout } from '@fundamental-ngx/core/flexible-column-layout';`
+                    },
+                    {
+                        answer: 'SAP UI5 design system compliance out of the box',
+                        recommendation:
+                            'Use ui5-flexible-column-layout from @fundamental-ngx/ui5-webcomponents. Slots are start-column, mid-column, end-column attributes.',
+                        example: `<ui5-flexible-column-layout layout="TwoColumnsMidExpanded">
+  <div slot="startColumn">Master</div>
+  <div slot="midColumn">Detail</div>
+</ui5-flexible-column-layout>`
+                    }
+                ]
+            }
+        ],
+        commonPitfalls: [
+            'All six title inputs are required for accessibility: collapseTitle, collapseTitleStartBtn, collapseTitleEndBtn, expandTitle, expandTitleStartBtn, expandTitleEndBtn. Omitting any causes runtime errors.',
+            'separatorAriaLabel is required for screen reader accessibility of the column resize handles.',
+            'layoutDefinitions input is required — provide a FlexibleColumnLayoutDefinition[] that maps each FlexibleColumnLayout value to column width percentages.',
+            'Content projection uses template reference variables (#startColumn, #midColumn, #endColumn) inside ng-template elements — NOT named slot attributes. The ng-template must be a direct child of fd-flexible-column-layout.',
+            '#midColumn and #endColumn are only rendered when the active layout includes those columns. Do not assume they are always mounted.',
+            'FlexibleColumnLayout enum values: OneColumnStartFullScreen | OneColumnMidFullScreen | OneColumnEndFullScreen | TwoColumnsStartExpanded | TwoColumnsMidExpanded | TwoColumnsEndExpanded | ThreeColumnsMidExpanded | ThreeColumnsEndExpanded | ThreeColumnsStartMinimized | ThreeColumnsEndMinimized.',
+            'Import FlexibleColumnLayout enum from @fundamental-ngx/core/flexible-column-layout, not a deep path.'
+        ],
+        compositionPattern: `import { FlexibleColumnLayout } from '@fundamental-ngx/core/flexible-column-layout';
+
+// In component
+layout = signal<FlexibleColumnLayout>(FlexibleColumnLayout.OneColumnStartFullScreen);
+
+// Template
+<fd-flexible-column-layout
+  [layout]="layout()"
+  [layoutDefinitions]="layoutDefs"
+  collapseTitle="Collapse panel"
+  collapseTitleStartBtn="Collapse start panel"
+  collapseTitleEndBtn="Collapse end panel"
+  expandTitle="Expand panel"
+  expandTitleStartBtn="Expand start panel"
+  expandTitleEndBtn="Expand end panel"
+  separatorAriaLabel="Resize column separator">
+  <ng-template #startColumn>
+    <!-- Master / list panel -->
+  </ng-template>
+  <ng-template #midColumn>
+    <!-- Detail panel -->
+  </ng-template>
+  <ng-template #endColumn>
+    <!-- Extra detail panel (3-column layouts only) -->
+  </ng-template>
+</fd-flexible-column-layout>`,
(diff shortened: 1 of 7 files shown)


libs/mcp-server/src/data/usage-guides.ts
```diff
@@ -358,5 +358,121 @@ this._dialogService.open(MyDialogComponent, {
   </fd-card-content>
 </fd-card>`,
         relatedComponents: ['fd-card-header', 'fd-card-content', 'fd-avatar', 'fd-tile', 'ui5-card']
+    },
+
+    'fd-flexible-column-layout': {
+        component: 'fd-flexible-column-layout',
+        summary:
+            'SAP Fiori master-detail layout that transitions between 1, 2, and 3 columns. Controlled by the FlexibleColumnLayout enum via the layout input. Content is projected via #startColumn, #midColumn, and #endColumn template reference variables.',
```
InnaAtanasova, line 367: for the summary and esp for the decision tree maybe give the AI also this link to the UX specs: https://www.sap.com/design-system/fiori-design-web/v1-145/page-types/page-layouts/flexible-column-layout

droshev (approved):

Issue #13640, opened 2025-11-29
[FSM][P2] Table P13N Dialog for Column Selection, Sorting, Grouping, Filtering, Custom


D034394:
Dear fundamentals-ngx Team,

Currently Fundamentals provides two ways of managing table settings (p13n):

1. The personalization dialog: https://sap.github.io/fundamental-ngx/#/platform/table/p13-dialog-table -> The challenge we see here is that every option (Columns, Sorting, Filtering, Grouping) opens separate dialogs.
2. Combined Table Settings Dialog: https://sap.github.io/fundamental-ngx/#/platform/table/settings-dialog-table -> the combined table settings dialog provides an all in one dialog but lacks the Select Column feature.

What we try to achieve is very common in SAPUI5 - having **one dialog with multiple tabs for column settings, sorting, grouping, filtering** (see also sample screenshot at the bottom):
- https://sapui5.netweaver.ondemand.com/sdk/#/entity/sap.m.p13n.Popup/sample/sap.m.sample.p13n.Popup
- https://sapui5.netweaver.ondemand.com/sdk/#/entity/sap.m.p13n.Engine

Additionally, there shall be the option to add additional tabs for **custom table settings**.

We need this feature add least for v0.56.

### Which versions of Angular and Fundamental Library for Angular are affected?

Angular v19, Fundamentals v0.56

<img width="665" height="512" alt="Image" src="https://github.com/user-attachments/assets/4ab84546-e4de-4b70-b7c4-4aec40a421d6" />

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

droshev:
@D034394 We will be able to take a look on this issue as early as in February 2026

droshev:
@D034394 As this is a new feature (and you have all the functionality available already) we won't downport it and only plan and implement it to the latest release.

mikerodonnell89:
We will address this issue over the course of several PR's, which will do the following:

1. The "Combined Table Settings" dialog, which uses a segmented button to allow the user to configure sorting, filtering, and grouping all from the same dialog, will add an additional button to allow for column show/hide and reordering.
2. This split button will be converted to a tab interface.
3. Sorting needs to support sorting on multiple properties, and the user needs to be able to prioritize those properties. See [here](https://sapui5.netweaver.ondemand.com/sdk/#/entity/sap.m.p13n.Engine/sample/sap.m.sample.p13n.Engine)
[...]

PR #14212, opened 2026-05-15
feat(platform): add column configuration to table combined settings dialog

mikerodonnell89:
part of #13640

Adds a columns button to the segmented button on the combined table settings dialog (bottom example on [this](https://deploy-preview-14212--fundamental-ngx.netlify.app/#/platform/table/settings-dialog-table) page). You may note some duplicate logic from the p13 dialog. I deliberately chose not to abstract this functionality, as I intend on deprecating the existing p13 dialog in favor of the combined settings dialog, and I'd like to avoid pointless abstraction of classes when there is only one class extending the abstract class.

**BREAKING CHANGES**
The combined settings dialog now has `width: auto` (previously `width: 340px`) to ensure that all dialog contents are rendered.


libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
```diff
@@ -0,0 +1,381 @@
...
+    headingLevel = input<number>(2);
+
+    /** Event emitter for columns changes */
+    columnsChange = output<SettingsColumnsDialogResultData>();
+
+    /** Event emitter for reset availability changes */
+    resetAvailabilityChange = output<boolean>();
+
+    /** All available columns for interacting */
+    _selectableColumns = signal<SelectableColumn[]>([]);
```
MariaIDineva: I think our convention is not to prefix non-private fields with an underscore. Also, those fields used only in the template and not meant for external use, [Angular suggest](https://angular.dev/style-guide#use-protected-on-class-members-that-are-only-used-by-a-components-template) to be marked as `protected`.

Same goes for the methods below. Those that are not private should not be underscored and those used in the template should be marked as `protected`.

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.html
```diff
@@ -93,6 +103,15 @@
                     (resetAvailabilityChange)="onResetAvailabilityChange($event)"
                 ></fdp-grouping>
             }
+            @case ('columns') {
+                <fdp-columns
+                    [headingLevel]="headingLevel"
+                    [columnsData]="columnsData()!"
+                    [initialColumns]="_initialColumns()"
+                    (columnsChange)="onColumnsChange($event)"
+                    (resetAvailabilityChange)="onResetAvailabilityChange($event)"
```
MariaIDineva, line 109: Hm, there seems to be something weird going on with the reset button as it does not seem to be working for 'Columns', 'Sort' and 'Group'.

https://github.com/user-attachments/assets/cb6a4ee6-d99f-4a1c-b2e8-bb74443c70ba

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
```diff
@@ -170,6 +184,24 @@ export class SettingsDialogComponent implements Resettable {
...
+                // Reset columns to initial state
+                this.columnsData.update((data) => {
+                    if (!data) {
+                        return data;
+                    }
+                    return {
+                        ...data,
+                        columns: data.columns.map((col) => ({
+                            ...col,
+                            visible: initialColumns.visibleColumns.includes(col.key)
```
MariaIDineva: I think we should take care not  only of the visibility but also of column order here.

libs/platform/table/components/table-view-settings-dialog/table-view-settings-dialog.component.ts
```diff
@@ -24,18 +24,25 @@ import {
...
+    SettingsColumnsDialogData,
+    SettingsGroupDialogData,
+    SettingsSortDialogData
+} from './table-view-settings.model';
 
 export const dialogConfig: DialogConfig = {
     responsivePadding: false,
     verticalPadding: false,
-    width: '340px'
+    width: '368px'
```
MariaIDineva: Is there a specific requirement for the width of the dialog? I spotted that in some cases in the new 'Columns' tab a horizontal scroll appears and I am not sure that's the best user experience. Do you think we could make it a little wider?

https://github.com/user-attachments/assets/9b625c37-9351-4742-aacf-275a309e2ab8

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
```diff
@@ -0,0 +1,381 @@
...
+            this._initiateColumns(data);
+        }
+
+        this._compareInitialColumns();
+    }
+
+    /** @hidden */
+    _toggleSelectAll(selectAll: boolean): void {
+        this._selectableColumns.update((columns) => {
+            columns.forEach((column) => (column.selected = selectAll));
```
MariaIDineva: Instead of mutating the object in place, we could:
```
this._selectableColumns.update((columns) =>
    columns.map((col) => ({ ...col, selected: selectAll }))
);
```

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    /** @hidden */
+    _toggleShowAll(): void {
+        this._showAllItems.update((val) => !val);
+    }
+
+    /** @hidden */
+    _setActiveColumn(column: SelectableColumn | null): void {
+        this._selectableColumns.update((columns) => {
```
MariaIDineva: Same here:
```
this._selectableColumns.update((columns) =>
    columns.map((col) => ({ ...col, active: col === column }))
);

```

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
```diff
@@ -0,0 +1,376 @@
...
+})
+export class ColumnsComponent {
+    /** Input data for columns configuration */
+    columnsData = input<SettingsColumnsDialogData>();
+
+    /** @hidden Initial columns state */
+    initialColumns = input<Nullable<SettingsColumnsDialogResultData>>();
+
+    /** @hidden Heading level */
+    headingLevel = input<number>(2);
```
droshev: this is not used input, maybe you missed it? looks like a dead input

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
```diff
@@ -208,6 +232,19 @@ export class SettingsDialogComponent implements Resettable {
...
+    /**
+     * Handle columns change event and update columns data.
+     * @param event Updated columns data.
+     */
+    onColumnsChange(event: SettingsColumnsDialogResultData): void {
+        // Simply use the columns array from the event which is already
+        // in the correct order with correct visibility flags
+        this.columnsData.set({
+            ...this.columnsData()!,
+            columns: event.columns
```
droshev: rewrites columnsData with a new object reference on every toggle/move. The child's effect treats that as a real change and calls _initiateColumns, which clears the search query, resets "Show Selected" back to "Show All", and forces the active row to index 0.

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    /** @hidden Track the last data reference to detect real changes */
+    private _lastColumnsDataRef: SettingsColumnsDialogData | undefined;
+
+    /** @hidden */
+    constructor() {
+        effect(() => {
+            // Recalculate button states when filtered columns change
+            this.filteredColumns();
```
droshev: could these be in computed() signals instead of an effect()? From Angular: "Effects should be the last API you reach for"

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
mikerodonnell89: It is used, see settings-dialog.component.html

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
```diff
@@ -0,0 +1,366 @@
...
+            const data = this.columnsData();
+            if (data && data !== this._lastColumnsDataRef) {
+                this._lastColumnsDataRef = data;
+                this._initiateColumns(data);
+                // Only compare if initialColumns is actually set
+                if (this.initialColumns()) {
+                    this._compareInitialColumns();
+                }
+            }
+        });
```
InnaAtanasova: The effect has a conditional dependency: `if (data && data !== this._lastColumnsDataRef)` means the data signal is read conditionally, creating invisible dependency gaps. The proper pattern is to use computed() for derived state or handle initialization in ngOnInit() / input transforms.

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
```diff
@@ -0,0 +1,366 @@
...
+
+    /** @hidden */
+    protected setActiveColumn(column: SelectableColumn | null): void {
+        this.selectableColumns.update((columns) =>
+            columns.map((col) => ({
+                ...col,
+                active: col === column
+            }))
+        );
+    }
```
InnaAtanasova, line 171: This method updates the signal using .update() and maps over columns to create new objects, which is correct. However, the implementation creates a new array with new objects on every update, which could be expensive if the column list is large. Consider marking only the active property as mutable.

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
```diff
@@ -0,0 +1,366 @@
...
+class SelectableColumn {
+    constructor(
+        /** Selected */
+        public selected: boolean,
+        /** Active */
+        public active: boolean,
+        /** Table Column it belongs to */
+        public column: SettingsColumnsDialogColumn
+    ) {}
+}
```
InnaAtanasova: Why is this a class and not an interface?

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
```diff
@@ -208,6 +238,16 @@ export class SettingsDialogComponent implements Resettable {
...
     }
 
+    /**
+     * Handle columns change event and update columns data.
+     * @param event Updated columns data.
+     */
+    onColumnsChange(event: SettingsColumnsDialogResultData): void {
+        // Store the current result without updating the columnsData signal
+        // This prevents triggering the child component's effect and resetting its state
+        this._currentColumnsResult = event;
```
InnaAtanasova: Here the result is stored in `_currentColumnsResult` instead of updating `columnsData` signal to avoid triggering the child component's effect. This is a workaround for the effect anti-pattern. If the child used proper patterns (computed/linkedSignal), this wouldn't be necessary.

InnaAtanasova (commented):
Just some suggestions...

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
MariaIDineva: Hm, I advised this change so that we avoid mutating objects in place, not considering any possible performance implications. This method was previously returning [...columns] like so:
```
this.selectableColumns.update((columns) => {
  columns.forEach((col) => { col.active = col === column; });
  return [...columns];
});
```
So now that you are raising this question, I am unsure what the correct approach is really...

I ran the question through Claude and here's what he said:

> The reviewer's comment is incorrect for this codebase. The current approach is right, and the suggestion would break Angular signals reactivity.
> 
> The relevant landmine from the project's own CLAUDE.md:
> 
> DO NOT mutate objects/arrays in place then call signal.set() with the same reference — signals use reference equality (===), so same-reference updates are silently ignored.
> 
> If you follow the reviewer's advice and mutate col.active in place, the inner objects change but the array reference stays the same. The signal won't detect a change and the UI won't update.
> 
> The "optimization" the reviewer imagines would look like this broken pattern:
> 
> 
> // BROKEN — same array reference, signal ignores the update
> this.selectableColumns.update((columns) => {
>   columns.forEach((col) => { col.active = col === column; });
>   return columns; // same ref → signal sees no change
> });
>
> You could return a shallow [...columns] copy to force a reference change while keeping mutated inner objects, but that approach:
> 

- > Still iterates every column (same O(n) cost)
- > Mutates the previous signal value as a side effect (messy)
- > Saves only the cost of object spreads, which are near-zero for a typical column list (5–50 items)

>
> Advice: Keep the current implementation. The spread-to-new-objects pattern is the idiomatic Angular signals approach and is already what the project guidelines require. The performance concern is premature — object spread over a small list is negligible compared to a change detection cycle.

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
droshev: yes, you are passing it from there but the column component doesn't use it after

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
InnaAtanasova: The effect reads initialColumns() inside a conditional block, so Angular only tracks it as a dependency if that if statement runs. This means if initialColumns changes later but the condition isn't met, the effect won't re-run—creating silent bugs. Effects should be for side effects (DOM, logging), not for deriving component state; use computed() for derived values or handle initialization in ngOnInit() instead.

InnaAtanasova: "Invisible dependency gaps" means Angular doesn't know the effect depends on initialColumns() because it's only read conditionally—so when initialColumns changes, Angular doesn't re-run the effect, leaving a gap in the reactive chain that you can't see by looking at the code.

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
mikerodonnell89: Oh interesting, looks like the other components ('fdp-sorting' etc) aren't using it either. I'll update those as well

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
```diff
@@ -0,0 +1,366 @@
...
+        FormsModule,
+        ListTitleDirective,
+        FdTranslatePipe
+    ]
+})
+export class ColumnsComponent {
+    /** Input data for columns configuration */
+    columnsData = input<SettingsColumnsDialogData>();
+
+    /** @hidden Initial columns state */
```
droshev, line 69: Can you document the inputs as this is public API?

mikerodonnell89: This component is meant to be internal use only

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
```diff
@@ -170,6 +197,16 @@ export class SettingsDialogComponent implements Resettable {
                 direction: this._initialGrouping()!.direction
             });
         }
+        if (this.activeTab() === ActiveTab.COLUMNS && this.columnsData()) {
+            const initialColumns = this._initialColumns();
+            if (initialColumns) {
+                // Reset columns to initial state (both order and visibility)
+                this.columnsData.set({
+                    ...this.columnsData()!,
+                    columns: initialColumns.columns
```
droshev, line 206: reset() sets columnsData back to initial but doesn't clear _pendingColumnsChanges.
you can test it: open dialog → reorder/hide a column → click Reset → click Confirm. It doesn't reset

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
```diff
@@ -104,38 +110,59 @@ export class SettingsDialogComponent implements Resettable {
...
+        if (this.showColumns) {
+            this.columnsData.set(data.columnsData);
+        }
         this.headingLevel = data.headingLevel;
         this.activeTab.set(this._getInitialActiveTab());
         this._shouldRenderSubheader();
         this._setInitialSorting();
         this._setInitialFilters();
         this._setInitialGrouping();
+        this._setInitialColumns();
```
droshev: it is called unconditionally even when allowColumnConfiguration = false. It calls getTableColumns() and builds a full column array that's never used.

droshev (changes requested):
I added 2 comments

MariaIDineva:
Hm, it seems like there is still a problem with reseting those settings that come as radio buttons. I tested that on main as well and the bug is aslo present there, so if you prefer, we could create a separate issue to fix that. Whatever you decide.

https://github.com/user-attachments/assets/01762391-6ac0-4f42-b747-8d3b2883a2c1

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
```diff
@@ -0,0 +1,309 @@
...
+
+    /**
+     * Compare initial columns with current state on next tick.
+     * Scheduled via setTimeout to avoid triggering during signal computation.
+     * @hidden
+     */
+    private _compareInitialColumnsNextTick(): void {
+        setTimeout(() => {
+            const initialColumns = this.initialColumns();
+            if (!initialColumns) {return;}
```
InnaAtanasova: This will cause lint errors

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.ts
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /** @hidden Move active column to target index */
+    private _moveActiveColumn(event: Event, targetIndex: number, refocusButton = false): void {
+        event.stopPropagation();
+        event.preventDefault();
+        this._moveColumnInFilteredListByIndex(this._activeColumnIndexInFilteredList(), targetIndex);
+
+        if (refocusButton) {
+            setTimeout(() => (event.target as HTMLElement | null)?.focus(), 0);
```
InnaAtanasova, line 213: Can you add a comment why the setTimeout is needed in this case for future reference?

libs/platform/table/components/table-view-settings-dialog/table-view-settings-dialog.component.ts
```diff
@@ -24,18 +24,25 @@ import {
...
+    SettingsColumnsDialogData,
+    SettingsGroupDialogData,
+    SettingsSortDialogData
+} from './table-view-settings.model';
 
 export const dialogConfig: DialogConfig = {
     responsivePadding: false,
     verticalPadding: false,
-    width: '340px'
+    width: 'auto'
```
InnaAtanasova, line 38: This is a visual breaking change that should be mentioned in the PR description. Users who relied on the fixed width for layout calculations will be affected.

libs/docs/platform/table/child-docs/settings-dialog/settings-dialog-docs.component.html
```diff
@@ -59,13 +59,14 @@
...
-        The <code>fdp-table-view-settings-dialog</code> component allows users to manage sorting, filtering, and
-        grouping settings in one convenient dialog.
+        The <code>fdp-table-view-settings-dialog</code> component allows users to manage column order, column hiding,
+        sorting, filtering, and grouping settings in one convenient dialog.
     </p>
     <p>
         By using the properties <code>[sortable]="true"</code>, <code>[filterable]="true"</code>, and
         <code>[groupable]="true"</code>, developers can enable the respective functionalities directly in the table
-        columns.
+        columns. Column order and column hiding is made available by setting the
```
InnaAtanasova: "Column order and column hiding are made available"

mikerodonnell89:
> Hm, it seems like there is still a problem with reseting those settings that come as radio buttons. I tested that on main as well and the bug is aslo present there, so if you prefer, we could create a separate issue to fix that. Whatever you decide.

fixed in this PR, thanks!

MariaIDineva (dismissed):

InnaAtanasova (dismissed):

libs/platform/table/components/table-view-settings-dialog/columns/columns.component.spec.ts
```diff
@@ -0,0 +1,444 @@
...
+    describe('initialization', () => {
+        it('should initialize with columns data', () => {
+            const columnsData: SettingsColumnsDialogData = {
+                columns: mockColumns
+            };
+
+            fixture.componentRef.setInput('columnsData', columnsData);
+            fixture.detectChanges();
+
+            expect(component.selectableColumns().length).toBe(4);
```
droshev: doesn't TS complain that you are calling protected method here (the same issue applies to showAllItems, filteredColumns, moveUpDisabled, moveDownDisabled, selectedColumnsCount)

libs/platform/table/components/table-view-settings-dialog/table-view-settings-dialog.component.ts
droshev: since you introduce the settings, shouldn't you check above also for `allowColumnConfiguration`?

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.spec.ts
```diff
@@ -171,4 +176,149 @@ describe('SettingsDialogComponent', () => {
...
+            };
+            fixture = TestBed.createComponent(SettingsDialogComponent);
+            component = fixture.componentInstance;
+            fixture.detectChanges();
+
+            expect(component.showSubheader()).toBe(true);
+
+            // Check DOM - should render icon tab bar
+            const nativeElement = fixture.nativeElement as HTMLElement;
+            const iconTabBar = nativeElement.querySelector('fdp-icon-tab-bar');
```
droshev: where do we have icon-tab-bar? i think it is segmented-button ?

droshev (approved):

Issue #14085, opened 2026-04-03
Not support parallel flows in serial.


zhuwentao0724:
### Describe your proposal

When there are two or more parallel paths within a serial flow, the workflow is not displaying. Details in https://product-jira.ariba.com/browse/SS-75104

### Which versions of Angular and Fundamental Library for Angular are affected?

@fundamental-ngx/platform": "^0.57.6

### Affected Component

fdp-approval-flow

### Steps to Reproduce (for bugs)

**1. Below is the approval flow and the node values passed to the fdp-approval-flow (approvalProcess.nodes) is correct.**
**flow:**

<img width="1117" height="305" alt="Image" src="https://github.com/user-attachments/assets/40e02972-4875-4a21-a377-3b1168ad13b0" />

**node values:**
[
    {
        "id": "AAAAAFjJS3p8",
        "name": "A/P - Expenses",
        "description": "",
        "approvers": [
            {
                "id": "A/P - Expenses",
                "teamId": "A/P - Expenses",
                "name": "A/P - Expenses",
                "description": "Team User",
                "imgUrl": ""
            }
        ],
        "status": "Pending",
        "targets": [
            "AAAAAFjJS3qC",
            "AAAAAFjJS3qI"
        ],
        "isEveryoneApprovalNeeded": true,
        "disableActions": false
    },
    {
        "id": "AAAAAFjJS3qF",
        "name": "A/P - Supplier Payment",
        "description": "",
        "approvers": [
            {
                "id": "A/P - Supplier Payment",
                "teamId": "A/P - Supplier Payment",
                "name": "A/P - Supplier Payment",
                "description": "Team User",
                "imgUrl": ""
            }
        ],
        "status": "Pending",
        "targets": [
            "AAAAAFjJS3p!"
        ],
        "isEveryoneApprovalNeeded": true,
        "disableActions": false
    },
    {
        "id": "AAAAAFjJS3p!",
        "name": "BOM Manager",
        "description": "",
        "approvers": [
            {
                "id": "BOM Manager",
                "teamId": "BOM Manager",
                "name": "BOM Manager",
                "description": "Team User",
                "imgUrl": ""
            }
        ],
        "status": "Pending",
        "targets": [
            "AAAAAFjJS3qC",
            "AAAAAFjJS3qI"
        ],
        "isEveryoneApprovalNeeded": true,
        "disableActions": false
    },
    {
        "id": "AAAAAFjJS3qC",
        "name": "Category Buyer",
[...]

PR #14214, opened 2026-05-17
fix(platform): add fix for the exception that occurs in fdp-approval-flow for multiple parallel blocks

zhuwentao0724:
…ltiple parallel blocks.

## Related Issue(s)

https://github.com/SAP/fundamental-ngx/issues/14085

closes

## Description

When there are two or more parallel paths within a serial flow, the workflow is not displaying.

## Screenshots

image


droshev:
@zhuwentao0724 you need to join SAP org first, then you will be able to open a PR from a branch

PR #14237, opened 2026-05-26
[WIP]feat(core): implement 2 months calendar

### LINKED ISSUE #13416: Two-Month Calendar Horizontal View
### Describe your proposal

adopt https://github.com/SAP/fundamental-styles/issues/6131

latest

_No response_

### StackBlitz example(s)

_No response_

### Declaration

droshev:
## Summary

fixes #13416

  - Implement multi-month calendar view with new `FdCalendarContainerComponent` that manages 2+ calendars in horizontal or vertical layout

Files for review:
  1. `libs/core/calendar/calendar-container/calendar-container.component.ts` - the new container component that orchestrates multi-month logic, layout switching, and event coordination between calendar instances
  2. `libs/core/calendar/calendar.component.ts` - API changes to work with the container

  ## Test plan

  - [ ] Horizontal layout displays calendars side-by-side; vertical layout stacks them vertically
  - [ ] Month navigation arrow visibility is correct (previous hidden on non-first, next hidden appropriately)
  - [ ] Date range selection works across both visible months (hover effects, selection state)
  - [ ] Responsive collapse to vertical layout triggers below 895px width
  - [ ] Calendar header displays correct month/year for each displayed month
  - [ ] Hover over dates works in both months (fix(core): hover over dates works in all months)


InnaAtanasova (changes requested):
1. I don't think it's a good idea to have in the documentation (and in general as functionality) an option for more than 2 months. It's called "Two-Month Calendar" in the specs. 
2. The selection of the year and month happens in a wrong way. Currently the month and year view replaces the days view 

Image
Image

The month/year view should be a popover (specs:

Image

3. Missing Phone example where the Calendar popover turns into full screen dialog with only one month and taking back the adjacent month days. This is something the component needs to provide as logic, not like in the normal calendar where it can be an input property. Please refer to the specs:

Image

droshev:
> 1. I don't think it's a good idea to have in the documentation (and in general as functionality) an option for more than 2 months. It's called "Two-Month Calendar" in the specs.
> 2. The selection of the year and month happens in a wrong way. Currently the month and year view replaces the days view
> 
> Image Image
> The month/year view should be a popover (specs:
> 
> Image
> 3. Missing Phone example where the Calendar popover turns into full screen dialog with only one month and taking back the adjacent month days. This is something the component needs to provide as logic, not like in the normal calendar where it can be an input property. Please refer to the specs:
> 
> Image

1. Fair point - I will update the docs
2. Agreed 
3. Agreed on the gap, and agreed the container should provide the logic rather than expecting consumers to wire it. Placement-wise this belongs in fd-date-picker.

droshev:
closing it for now, i will reopen it once i am back from vac.

PR #14239, opened 2026-05-27
docs: add a skills page to documentation

MariaIDineva:
## Related Issue(s)

closes none

## Description

This PR consists of two main changes:

1. Adds a new `Skills` page to the documentation website. The `Skills` page now appears in the sidebar as "Skills" and navigates to `/skills/overview`, showing all 14 skills categorized as `Building & Setup`, `Code Quality & Review`, and `Migration & Maintenance`.
2. Updates the `MCP` page on the documentation website so that it resembles the one that we have in fundamental-styles. Also adds information how to use the `modelcontextprotocol` inspector if a user wants to play with the MCP manually and provides some example queries.

### DIFF
--- a/apps/docs/src/assets/skills/README.md
+++ b/apps/docs/src/assets/skills/README.md
@@ -1,0 +1,189 @@
+# Skills
+
+Agent Skills for **@fundamental-ngx** provide AI assistants with step-by-step guidance for building, testing, migrating, and reviewing Angular components with this library. These skills help AI agents (like Claude Code) complete complex multi-step tasks accurately without hallucinating APIs.
+
+## What are Skills?
+
+Skills are markdown-based task playbooks specifically designed for AI assistants. Each skill contains:
+
+- **Phased workflows** — Step-by-step instructions that guide the AI from discovery through implementation
+- **API lookup patterns** — How to query the MCP server for accurate, up-to-date component APIs before writing code
+- **Project conventions** — Enforced Angular 21+ patterns, NX workspace commands, and library-specific rules
+- **Validation steps** — How to verify the output compiles and behaves correctly
+
+## Available Skills
+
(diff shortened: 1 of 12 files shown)


droshev (dismissed):

droshev (approved):

apps/docs/src/assets/skills/README.md
```diff
@@ -0,0 +1,189 @@
...
+- **preflight** — Run all local quality gates (format, lint, build, test) before opening a PR
+
+### Migration & Maintenance (4 skills)
+
+Keep the codebase up to date with evolving standards and dependencies:
+
+- **migrate** — Migrate a component or directive from decorator-based patterns to Angular 21+ signal-based patterns
+- **update-docs** — Verify that documentation examples are in sync with a component's current public API and fix any drift
+- **adopt-styles** — Apply breaking changes from `fundamental-styles` into the Angular component templates and class bindings
+- **i18n-manage** — Add, rename, or remove i18n translation keys across `FdLanguage`, `.properties` files, and generated types
```
InnaAtanasova, line 45: Are these skills exposed to the users? In my opinion they make more sense to be "internal" for the project and not exposed. What do you think about it?

InnaAtanasova (approved):

apps/docs/src/assets/skills/README.md
MariaIDineva: Yes, all of them are exposed. Now a researched if we could hide those, but Clause claims that there is no way (a flag or setting) that could make them not be exported by `npx skills add`. We could only add `user-invocable: false` to the metadata to hide a certain skill from slash commands and/or add `disable-model-invocation: true` to now allow Claude to use it. Do you mean any of those? Or do you mean we should not add them to the readme?

PR #14244, opened 2026-06-01
fix(platform): use tabs instead of segmented button in combined table settings dialog

mikerodonnell89:
part of #13640

Changes the combined table settings dialog to use tabs instead of a segmented button to switch from column ordering/sorting/filtering/grouping


libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
```diff
@@ -264,11 +272,56 @@ export class SettingsDialogComponent implements Resettable {
...
 
     /**
-     * Handle changes to the reset availability.
-     * @param event Boolean indicating if reset is available.
+     * Handle tab selection changes from the icon tab bar.
+     * @param event The selected tab item.
      */
-    onResetAvailabilityChange(event: boolean): void {
-        this.isResetAvailable$.set(event);
+    onTabSelected(event: any): void {
```
MariaIDineva: That event should be of type `IconTabBarItem`.

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
```diff
@@ -116,9 +117,15 @@ export class SettingsDialogComponent implements Resettable {
...
     protected showColumns = false;
 
+    /** @hidden Expose ActiveTab enum to template */
+    protected readonly ActiveTab = ActiveTab;
+
     /** @hidden Pending columns changes (not applied to signal to avoid triggering child effects) */
     private _pendingColumnsChanges: Nullable<SettingsColumnsDialogResultData> = null;
 
+    /** @hidden Tracks reset availability per tab */
+    private _resetAvailabilityByTab = signal<Record<string, boolean>>({});
```
MariaIDineva, line 122: One from Claude, related to this field, which I think is valid:

> In settings-dialog.component.ts, after reset() runs, isResetAvailable$ is set to false, but _resetAvailabilityByTab[activeTab] still holds the old true value. If the user switches to another tab and back, updateResetAvailability() will restore the stale true from _resetAvailabilityByTab, re-enabling the Reset button even though the data was already reset. The reset() method should also call this._resetAvailabilityByTab.update(state => ({ ...state, [this.activeTab()!]: false })) (or call updateResetAvailability(this.activeTab()!, false)).

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
```diff
@@ -148,6 +155,7 @@ export class SettingsDialogComponent implements Resettable {
         this._setInitialSorting();
         this._setInitialFilters();
         this._setInitialGrouping();
+        this.updateResetAvailability();
```
MariaIDineva: I think this call is redundant as it will set `isResetAvailable$` to `false` which is already initially set to `false`.

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
MariaIDineva: Also, since we are only calling this method from inside the template, maybe we should mark it as `protected`? Same goes for `updateResetAvailability`. 
And then in the unit tests those can be accessed like this: `component['updateResetAvailability']`

MariaIDineva (commented):
One more comment from Claude:

> FormsModule is imported but no longer used. The [(ngModel)] on fd-segmented-button is removed in this PR and nothing in the new template requires FormsModule.

I think there are other imports that can be removed as well. My IDE shows `FocusableItemDirective` and `ButtonComponent`.

MariaIDineva (dismissed):

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.spec.ts
```diff
@@ -315,10 +310,219 @@ describe('SettingsDialogComponent', () => {
...
+            });
+
+            component.activeTab.set(ActiveTab.COLUMNS);
+            component.reset();
+
+            // Signal should be reset
+            expect(component.columnsData()?.columns).toEqual(initialColumns.columns);
+
+            // Pending changes should also be reset
+            component.confirm();
```
droshev: don't you want to spy before confirm?

droshev (approved):

PR #14251, opened 2026-06-03
fix(core, datetime-adapter): fix datetime picker manual input

### LINKED ISSUE #14250: Issue with Editing Bidding Start Time in Guided Sourcing
### Describe your proposal

Customer has reported an issue when editing the bid start timings (Hours). User enters timing manually in 24HR format as per attached recording and upon saving, the value is not saved. 

https://github.com/user-attachments/assets/b2401b49-00aa-4ea9-a93a-5a9b0a1079c0

    "@angular/core": "21.1.6", "@fundamental-ngx/datetime-adapter": "^0.57.6",

### Affected Component

_No response_

1. Go to Dayjs Datetime Adapter
2. Type menually date 
3.  it is not save time proply in 24 hour setting 

### StackBlitz example(s)

https://stackblitz.com/edit/elhabvbj?file=src%2Fapp%2Fdayjs-datetime-formats-example.component.ts,src%2Fapp%2Fdayjs-datetime-formats-example.component.html

### Priority

None

### Declaration

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14250
 
#### 1. Manual 24h time input silently discarded (`DayjsDatetimeAdapter`)

**File:** `libs/datetime-adapter/src/lib/dayjs-datetime-adapter.ts`

**What was broken:** When a user typed a time value in 24h notation (e.g. `5/25/2025 15:30`) into a datetime picker whose format used 12h tokens (`h:mm A`), the adapter's fallback parse chain exhausted every sensible format without success. The last fallback, `'YYYY-MM-DD'`, would non-strictly match the string and return a technically-valid `Dayjs` object — but with a garbage date (e.g. year 2027, day 15) and `hour = 0`. Because `isValid()` returned `true`, no validation error was shown and `onChange` emitted the wrong value. From the user's perspective, "the time was not saved."

**Fix:** Two changes to the fallback list in `_createDayjsDate`:

1. Added `HH:mm` and `H:mm` variants as fallbacks when the configured format contains lowercase 12h tokens (`h:mm A`/`h:mm a`). Typing `15:30` without AM/PM now matches the 24h fallback and the correct hour is preserved.
2. Added `HH:mm` and `H:mm` variants that strip the AM/PM meridiem marker when the format already uses uppercase 24h tokens (`HH:mm A`/`H:mm A`). This handles the case where the format is contradictory (24h hour + meridiem) and the user omits the AM/PM suffix.
3. Guarded the `'YYYY-MM-DD'` last-resort fallback behind a `/^\d{4}-/` check. It now only fires when the input string actually starts with a 4-digit year followed by a dash, preventing US-locale datetime strings from silently matching as a garbage ISO date.

---

#### 2. `customDateTimeFormat` ignored when parsing typed input (`DatetimePickerComponent`)

**File:** `libs/core/datetime-picker/datetime-picker.component.ts`

**What was broken:** `_parseDate` always used `this._dateTimeFormats.parse.dateTimeInput` (from the injected `DATE_TIME_FORMATS` provider) regardless of whether `[customDateTimeFormat]` was set on the picker. So a picker with `[customDateTimeFormat]="'YYYY-MM-DD h:mm A'"` would display in that format but parse typed input using the provider's format — a mismatch that caused the model not to update when the user typed directly into the input field.

**Fix:** `_parseDate` now uses `customDateTimeFormat` as the parse format when it is set, falling back to the provider format otherwise.

```ts
// before
return this._dateTimeAdapter.parse(date, this._dateTimeFormats.parse.dateTimeInput);

// after
const format = this.customDateTimeFormat ?? this._dateTimeFormats.parse.dateTimeInput;
return this._dateTimeAdapter.parse(date, format);
```

---

#### `DAYJS_DATETIME_FORMATS_24H` preset

**File:** `libs/datetime-adapter/src/lib/dayjs-datetime-formats.ts`

Added a ready-to-use `DateTimeFormats` preset for applications that need 24h time input and display. Previously consumers had to hand-roll the format object; now they can simply provide `DAYJS_DATETIME_FORMATS_24H` from `@fundamental-ngx/datetime-adapter`.

```ts
providers: [
    { provide: DATE_TIME_FORMATS, useValue: DAYJS_DATETIME_FORMATS_24H },
    { provide: DatetimeAdapter, useClass: DayjsDatetimeAdapter }
];
```

Use with `[meridian]="false"` on the picker to hide the AM/PM spinner.

---

### Documentation

- Added a **"24-hour time format"** example section to the Dayjs datetime adapter docs page, demonstrating `DAYJS_DATETIME_FORMATS_24H` with `[meridian]="false"`.
- Added a `Selected` date span to all format examples to display the selected value.

### Before:

https://github.com/user-attachments/assets/056e0608-4e73-48b0-a038-7d3e772d4b0c

### After:

https://github.com/user-attachments/assets/78c3b274-bbed-471d-a50f-11c29e38c068


libs/docs/core/dayjs-datetime-adapter/examples/date-picker-dayjs-adapter-example.component.html
```diff
@@ -11,3 +11,6 @@
 </fd-segmented-button>
 <br />
 <fd-date-picker [(ngModel)]="date"> </fd-date-picker>
+<br />
+<br />
+<span>Selected: {{ date.format('L HH:mm') }}</span>
```
droshev: you don't need to show the time(as it will always show 00:00) maybe you can write `Selected: {{ date.format('L') }}`

libs/datetime-adapter/src/lib/dayjs-datetime-adapter.ts
```diff
@@ -421,10 +421,15 @@ export class DayjsDatetimeAdapter extends DatetimeAdapter<Dayjs> {
                 format, // original
                 format.replace(/ ?[Hh]:?mm[aA]?/, ''), // remove time
                 format.replace(/ ?[Hh]:?mm/, ''), // remove time (no meridiem),
+                format.replace(/h:mm ?[aA]/i, 'HH:mm'),
```
droshev: the `/i` flag which makes it case-insensitive - so lowercase `h` and uppercase `H` both match. Take this scenario: you have a datetime picker configured with a format that uses HH:mm A — for example MM/DD/YYYY HH:mm A.

  When a user types 05/25/2025 15:30, the adapter builds the fallback list. The broken line:

  format.replace(/h:mm ?[aA]/i, 'HH:mm')
  // input:  'MM/DD/YYYY HH:mm A'
  // result: 'MM/DD/YYYY HHH:mm'   ← invalid

  dayjs doesn't know what HHH means, so that fallback always fails. The H:mm fallback has the same bug:

  format.replace(/h:mm ?[aA]/i, 'H:mm')
  // input:  'MM/DD/YYYY HH:mm A'
  // result: 'MM/DD/YYYY HH:mm'   ← actually fine here

  So the second replacement accidentally works — but the first one (HH:mm) is supposed to be the more specific/correct match and it produces garbage.

  To reproduce: configure a datetime picker with [customDateTimeFormat]="'MM/DD/YYYY HH:mm A'", type a 24h time like 05/25/2025 15:30 directly into the input field, and confirm. The value won't be saved correctly because the broken fallback
  HHH:mm causes that parse attempt to fail, and the adapter may fall through to a less accurate fallback or return invalid.

mikerodonnell89 (dismissed):

droshev (approved):

PR #14270, opened 2026-06-09
fix(core): expose bodyRole + bodyAriaLabelledBy inputs on fd-popover (#14260)

### LINKED ISSUE #14260: fix(core): fd-popover-body role=dialog without accessible name causes aria-dialog-name violation
## Bug

`PopoverBodyComponent` sets `role="dialog"` by default (added in #12907), but provides no accessible name. This causes axe-core `aria-dialog-name` WCAG 2.1 AA violations whenever a popover renders without an explicit label.

## Affected components

- `fd-form-input-message-group` / `fdp-input-message-group` — became observable after #14217 removed `focusAutoCapture`.
- Any `fd-popover` without `bodyAriaLabel`.

## Root cause

`libs/core/popover/popover-body/popover-body.component.ts` defaults `_bodyRole` to `'dialog'` with no accessible name.

## Proposed fix

Add translated `bodyAriaLabel` to `fd-form-input-message-group` and/or expose `bodyRole` as an input on `PopoverComponent`.

## Temporary workaround

Wizard-generator routes suppressed in `a11y-suppressions.json` until fixed.

droshev:
Fixes #14260

## Summary

Exposes `bodyRole` and `bodyAriaLabelledBy` as template inputs on `<fd-popover>` so consumers can set the correct ARIA role without going through the imperative `PopoverConfig` path.

**Background:** PR #12907 changed `PopoverBodyComponent._bodyRole` default from `null` → `'dialog'` to fix screen reader announcements. However, `bodyRole` was only settable via `PopoverConfig` (imperative API). Template consumers (`<fd-popover>` in HTML) had no way to override it.

**This PR:** Adds template inputs so all consumers can set the correct role:
```html
<fd-popover [bodyRole]="'region'" [bodyAriaLabelledBy]="'header-id'">
```

**Default behavior unchanged.** All 94 existing popover consumers remain byte-identical: `bodyRole` defaults to `'dialog'`. `null` is reachable — `[bodyRole]="null"` renders no role attribute at all, which is the correct ARIA shape for non-modal disclosure widgets where the trigger's `aria-haspopup` already carries the relationship.

### Primitive surface (`<fd-popover>`)

- `bodyRole = input<PopoverBodyRole | (string & {}) | null>('dialog')` — new input with type union for autocomplete + escape hatch
- `bodyAriaLabelledBy = input<string | null>(null)` — new input
- `PopoverBodyRole` type exported: `'dialog' | 'region' | 'menu' | 'listbox' | 'tooltip' | 'alertdialog'`
- Wired through `_effectiveConfig` → `PopoverService.refreshConfiguration` → `PopoverBodyComponent` signals
- `bodyRole` read path trusts the input directly (no `?? cfg ?? 'dialog'` fallback chain — that would swallow explicit `null`). Sibling inputs (`bodyAriaLabel`, `bodyAriaLabelledBy`) keep the chain since their default is `null` and fall-through is meaningful.

### Consumer patches

- `fd-form-input-message-group`: binds `[bodyAriaLabel]` to new i18n key `coreFormInputMessageGroup.popoverAriaLabel`
- `fdp-input-message-group`: binds `[bodyAriaLabel]` to new i18n key `platformInputMessageGroup.popoverAriaLabel`

## Out of scope (filed separately)

A follow-up audit will (a) classify the remaining 92 popover consumers by ARIA semantics (disclosure → `null`, listbox → `'listbox'`, menu → `'menu'`, region → `'region'` + name, true dialog → `'dialog'` + name) and (b) flip `bodyRole`'s default from `'dialog'` to `null`. Bundling that into this PR would balloon the diff and force a 94-consumer baseline regen.


libs/core/popover/popover.component.spec.ts
```diff
@@ -516,6 +516,88 @@ describe('PopoverComponent service stub tests', () => {
...
+        <fd-popover #popover [bodyRole]="bodyRole" [bodyAriaLabelledBy]="bodyAriaLabelledBy">
+            <fd-popover-control>
+                <button>Open Popover</button>
+            </fd-popover-control>
+            <fd-popover-body>
+                <div>Popover Content</div>
+            </fd-popover-body>
+        </fd-popover>
+    `,
+    standalone: true,
```
MariaIDineva, line 530: Maybe no need to set `standalone: true`

libs/core/popover/popover.component.ts
```diff
@@ -182,6 +182,34 @@ export class PopoverComponent implements AfterViewInit, AfterContentInit, OnDest
...
+     * 'tooltip' — match the role to the popover's actual semantics.
+     *
+     * @remarks When set via this input (including when unbound — defaults to 'dialog'),
+     * the PopoverConfig object's `bodyRole` field is ignored. Config-only usage
+     * is not supported for `bodyRole` because the input has a concrete default.
+     * To override, bind the input explicitly: `[bodyRole]="'region'"` or `[bodyRole]="null"`.
+     *
+     * See https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/ for guidance.
+     */
+    readonly bodyRole = input<string | null>('dialog');
```
InnaAtanasova: Right now the API documentation suggests a specific set of supported values (`'dialog', 'region', 'menu', 'listbox', 'tooltip', null`), but the type is `string | null`, which means consumers get no autocomplete, no compile-time validation, and can pass invalid values.

If those are indeed the only supported values, a union type would improve the developer experience:
```
type PopoverBodyRole = 'dialog' | 'region' | 'menu' | 'listbox' | 'tooltip';

readonly bodyRole = input<PopoverBodyRole | null>('dialog');
```

droshev: Added PopoverBodyRole union type with the 6 most common roles + (string & {}) escape hatch for edge cases like 'tree', 'status', 'alert'. Gives autocomplete without gatekeeping the ARIA spec. Also updated PopoverConfig.bodyRole for consistency

MariaIDineva (dismissed):

InnaAtanasova (dismissed):

PR #14295, opened 2026-06-16
fix(cdk): fix autocomplete race condition with fast typing

### LINKED ISSUE #14282: Regression bug in fd-combobox/fd-multi-input autocomplete after upgrade
### Describe your proposal

After upgrading Angular to v21 and fundamental-ngx from 0.59.3 to 0.62.3, we’re seeing a regression in fd-combobox and fd-multi-input with autocomplete. When typing quickly, the control sometimes selects the last character the user just typed, so the next keystroke overwrites it. Example: typing “Week24” fast can end up as “Wek24” or "Wee24". We can reproduce this manually, and many of our automated tests (which type quickly) now fail.
 
It looks like a race in the autocomplete behavior between updating the value and setting the highlighted suffix, causing the last typed character to be included in the selection.

### Reproduction Rate
Manual fast typing: **intermittent, but reliably reproducible within a few attempts.**
Automated tests (fast input): **~consistently failing across multiple suites.**

Angular 21.2.9 , Fundamental Library: 0.62.3

### Affected Component

fd-combobox , fd-multi-input

1. Open the page containing the fd-combobox / fd-multi-input.
2. Click into the input to give it focus.
3. Quickly type the string: Week24 (type at a fast, natural typing speed — do not pause between keystrokes).
4. Observe the value rendered in the input field after typing completes.

### Expected Result
The input contains exactly: **Week24**

### Actual Result
The input contains a corrupted value where one of the typed characters has been overwritten by the autocomplete suffix selection, e.g.:
- Wek24
- Wee24

The defect is timing-sensitive and reproduces more reliably the faster the user types. It is consistently reproduced by automated UI tests that use programmatic fast typing (even with 0.5 seconds per-key delay).

### StackBlitz example(s)

_No response_

### Priority

High

### Declaration

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/14282

### Summary
Fixes a race condition in `AutoCompleteDirective` where typing quickly caused characters to be lost.

### Problem
When typing fast (< 50ms between keystrokes), the native DOM input value updates immediately, but Angular's model value (`inputText`) lags behind due to async change detection. The autocomplete logic was using the stale model length to set the selection range, which caused it to select the last typed character. The next keystroke would then overwrite that character, resulting in missing letters.

Example: typing "Strawberry" fast would result in "Strawerry" or "Straw".

### Solution
Pass the current native input length to `_typeahead()` instead of reading the stale `this.inputText.length`. This ensures the selection range starts after all characters the user has actually typed.

### Changes
- Modified `_typeahead()` signature to accept `currentInputLength: number` parameter
- Updated call site to pass `currentNativeValue.length` instead of `this.inputText.length`
- Added regression tests demonstrating the race condition and verifying the fix

### Test Plan
1. Unit tests: Two new regression tests in `auto-complete.directive.spec.ts`
2. Manual testing: Added test buttons to combobox example in docs app
3. E2E: Existing Playwright tests continue to pass

### Breaking Change
`_typeahead()` is a private method, so this only affects consumers who extended `AutoCompleteDirective`.

## Screenshots
In Combobox component added 3 buttons to test fast typing Strawberry, Pineapple and Apple
Screenshot 2026-06-16 at 4 06 18 PM


libs/docs/core/combobox/examples/combobox-example.component.ts
```diff
@@ -17,5 +25,47 @@ export class ComboboxExampleComponent {
     searchTermFour = '';
     searchTermFive = 'Kiwi';
     searchTermSix = '';
-    fruits = ['Apple', 'Pineapple', 'Banana', 'Kiwi', 'Strawberry'];
+    fruits = ['Apple', 'Apricot', 'Aprium', 'Banana', 'Kiwi', 'Strawberry', 'Tomato', 'Pineapple'];
+
+    async testFastTyping(textToType: string): Promise<void> {
+        const input = document.getElementById('comboboxID1') as HTMLInputElement;
```
droshev, line 31: you cover this in the unit test and e2e, so maybe we can skip it from here? also it is very fragile with the getElementById() with hardcoded id

droshev (dismissed):

libs/cdk/utils/directives/auto-complete/auto-complete.directive.spec.ts
```diff
@@ -122,4 +122,51 @@ describe('AutoCompleteDirective', () => {
         // Should autocomplete to "Banana"
         expect((<any>directive)._elementRef.nativeElement.value).toBe('Banana');
     });
+
+    it('should handle fast typing without selecting last typed character', () => {
+        // Regression test for: https://github.com/SAP/fundamental-ngx/issues/XXXX
```
droshev: did you mean to write the issue link? maybe not needed, just the explanation is ok

libs/docs/core/combobox/examples/combobox-example.component.ts
InnaAtanasova: My intention was to keep this until the person reported the issue test it and verify the fix. After that I will remove it.

libs/cdk/utils/directives/auto-complete/auto-complete.directive.spec.ts
InnaAtanasova: True. I will remove this. Thanks

MariaIDineva (dismissed):
Wow, that was a tricky one... 🧠 but what a nice solution!

droshev (approved):

PR #14298, opened 2026-06-17
fix(core): fix msg box focus trap

### LINKED ISSUE #14173: focusTrapped in Message Box dosn't work
### Describe your proposal

The focusTrapped option is not working, because even though I set the value to false, the button still gets the "focus" and is highlighted.

v0.61.0

Image

v.0.62.0

Image

Angular 21.1.1, @fundamental-ngx/core 0.62.2

### Affected Component

_No response_

1. Go to message box documentation in the specific version
2. Run the stackblitz example

### StackBlitz example(s)

_No response_

### Priority

None

### Declaration

droshev:
## Summary
  - Fixes `focusTrapped: false` being ignored in `MessageBoxService.open()` — the approve button was always receiving focus on open regardless of the config value. Closes #14173.
  - Removes a duplicate `setupFocusTrap()` call in `MessageBoxComponent.ngAfterViewInit` that was redundant with the `afterNextRender` call already in `DialogBase`.

  ## Test plan
1. Open the MessageBox docs page → "Object-based" example → click "Open from object" → confirm focus moves to the "Ok" button.
2. Click "Open without focus trap" → confirm focus stays on the trigger button, not the "Ok" button.

### DIFF
--- a/libs/core/message-box/message-box-default/message-box-default.component.html
+++ b/libs/core/message-box/message-box-default/message-box-default.component.html
@@ -18,6 +18,7 @@
                 @if (_messageBoxContent.approveButton) {
                     <fd-button-bar
                         fdkInitialFocus
+                        [enabled]="_messageBoxConfig.focusTrapped !== false"
                         fdType="emphasized"
                         [label]="_messageBoxContent.approveButton"
                         [ariaLabel]="_messageBoxContent.approveButtonAriaLabel"
--- a/libs/core/message-box/message-box-default/message-box-default.component.spec.ts
+++ b/libs/core/message-box/message-box-default/message-box-default.component.spec.ts
@@ -1,7 +1,11 @@
 import { ChangeDetectorRef } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { RouterTestingModule } from '@angular/router/testing';
+import { FD_DIALOG_FOCUS_TRAP_ERROR } from '@fundamental-ngx/core/dialog';
 
 import { MessageBoxConfig } from '../utils/message-box-config.class';
 import { MessageBoxContent } from '../utils/message-box-content.class';
+import { MessageBoxRef } from '../utils/message-box-ref.class';
 import { MessageBoxDefaultComponent } from './message-box-default.component';
 
 describe('MessageBoxDefaultComponent', () => {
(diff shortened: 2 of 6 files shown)


MariaIDineva (approved):

libs/core/message-box/message-box-default/message-box-default.component.html
```diff
@@ -18,6 +18,7 @@
                 @if (_messageBoxContent.approveButton) {
                     <fd-button-bar
                         fdkInitialFocus
+                        [enabled]="_messageBoxConfig.focusTrapped !== false"
```
mikerodonnell89, line 21: Two things - not a fan of the double negative (does not equal false - better to just say true). And does this disable the buttons if focusTrapped is false?

mikerodonnell89 (changes requested):

libs/core/message-box/message-box-default/message-box-default.component.html
droshev: On the double negative — focusTrapped is typed as boolean | undefined (it's an optional config property, not a required input). Replacing it with focusTrapped === true would silently break the default case: any caller that doesn't set focusTrapped at all would get undefined === true → false, meaning the approve button would never receive initial focus. The !== false idiom is the correct way to express "enabled unless explicitly opted out" for an optional boolean that defaults to true.

On [enabled] disabling the button — [enabled] is an input on fdkInitialFocus, not on fd-button-bar. It controls only whether the directive fires a programmatic .focus() call on open. The button remains fully interactive regardless of this value.

mikerodonnell89 (changes requested):
Why are the buttons in the bar showing as compact regardless of what the content density is set to, but only in this example? Also a little margin between these buttons would be a good look

droshev:
> bar showing as compact regardless of what the content density is set to, but only in this example? Also a little margin between these buttons would be a good look

this is a surgical-focused fix for the focus trap of message box.

mikerodonnell89:
> > bar showing as compact regardless of what the content density is set to, but only in this example? Also a little margin between these buttons would be a good look
> 
> this is a surgical-focused fix for the focus trap of message box.

True, that is also not unique to this PR, apologies i should've checked

mikerodonnell89 (approved):

PR #14308, opened 2026-06-21
fix(core): prevent duplicate subscription in segmented-button button changes listener

### LINKED ISSUE #13358: BUG: Possible unintended multiple subscriptions in segmented-button.component.ts
### Describe your proposal

Hello,
there's a 'subscribe' inside _listenToButtonChanges which is called more than once (ngAfterViewInit and setDisabledState). This could lead to unintended multiple subscriptions.

Thank you,
CDP team.

Angular: 19.2.14 . FD: 0.56.4

_No response_

### StackBlitz example(s)

_No response_

### Declaration

droshev:
`_listenToButtonChanges()` was called from both `ngAfterViewInit` and `setDisabledState(false)`, each creating a new outer subscription to `QueryList.changes` without tearing down the previous one. After N enable/disable cycles, every change emission re-ran the inner work N+1 times.

Track the outer subscription as a field; unsubscribe before resubscribing. `takeUntilDestroyed` on the pipe handles destroy-time cleanup.

Keep the change surgical only

Closes #13358

### DIFF
--- a/libs/core/segmented-button/segmented-button.component.spec.ts
+++ b/libs/core/segmented-button/segmented-button.component.spec.ts
@@ -1,5 +1,5 @@
-import { Component, ElementRef, ViewChild } from '@angular/core';
-import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import { Component, ElementRef, ViewChild, signal } from '@angular/core';
+import { ComponentFixture, TestBed, fakeAsync, tick, waitForAsync } from '@angular/core/testing';
 import { RtlService } from '@fundamental-ngx/cdk/utils';
 import { ButtonComponent } from '@fundamental-ngx/core/button';
 import { runValueAccessorTests } from 'ngx-cva-test-suite';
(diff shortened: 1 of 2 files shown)


InnaAtanasova (commented):
Here are some suggestions from Claude review:

`[segmented-button.component.spec.ts:32]` Test suite name inconsistency: The first new test suite is named `SegmentedButtonComponent` — duplicate subscription fix but uses the wrong test component class. Line 32 creates `fixture = TestBed.createComponent(HostComponent)`, but the test is specifically about the duplicate subscription fix and should use `HostWithDynamicButtonComponent` (which is defined at line 250 but not used until the second suite). The test at line 40 calls `component.segmentedButton` as any which works because `HostComponent` also has a segmentedButton property, but the test intent would be clearer if it used a dedicated test component for this specific regression.

`[segmented-button.component.ts:119]` Field initialization: `_buttonChangesSub` is typed as `Subscription | null = null` but could use a more idiomatic pattern. Consider adding a comment explaining why null is needed vs undefined, or initialize to Subscription.EMPTY (which is safe to call .unsubscribe() on without a null check). Current approach works fine but the nullable check at line 249 could be avoided.

`[segmented-button.component.spec.ts:40-49]` The duplicate subscription test uses a tight loop with manual tick() calls that assumes specific RxJS scheduler behavior. While this effectively proves the fix, consider adding a comment explaining why tick(0) is necessary to flush the asyncScheduler delay from startWith(1). This helps future maintainers understand the test's timing requirements.

droshev:
> Here are some suggestions from Claude review:
> 
> `[segmented-button.component.spec.ts:32]` Test suite name inconsistency: The first new test suite is named `SegmentedButtonComponent` — duplicate subscription fix but uses the wrong test component class. Line 32 creates `fixture = TestBed.createComponent(HostComponent)`, but the test is specifically about the duplicate subscription fix and should use `HostWithDynamicButtonComponent` (which is defined at line 250 but not used until the second suite). The test at line 40 calls `component.segmentedButton` as any which works because `HostComponent` also has a segmentedButton property, but the test intent would be clearer if it used a dedicated test component for this specific regression.
> 
> `[segmented-button.component.ts:119]` Field initialization: `_buttonChangesSub` is typed as `Subscription | null = null` but could use a more idiomatic pattern. Consider adding a comment explaining why null is needed vs undefined, or initialize to Subscription.EMPTY (which is safe to call .unsubscribe() on without a null check). Current approach works fine but the nullable check at line 249 could be avoided.
> 
> `[segmented-button.component.spec.ts:40-49]` The duplicate subscription test uses a tight loop with manual tick() calls that assumes specific RxJS scheduler behavior. While this effectively proves the fix, consider adding a comment explaining why tick(0) is necessary to flush the asyncScheduler delay from startWith(1). This helps future maintainers understand the test's timing requirements.

comments 1 and 3: addressed.

about comment 2: Subscription.EMPTY has closed: true, which would misrepresent "no subscription yet" as "subscription completed".

InnaAtanasova (approved):

PR #14309, opened 2026-06-22
feat(core): shift-click multi-date selection for fd-calendar / fd-date-picker

### LINKED ISSUE #13502: Implement shift-click selection for fd-date-picker with [allowMultipleSelection]="true"
### Describe your proposal

Hello, Fundamental-NGX team,

Our application receives a lot of user feedbacks, that in the sapui5 version of the application it was possible to select a date range with shift key pressed. Is it possible to implement this kind of feature in fd-date-picker component when [allowMultipleSelection]="true" set?

It would enhance the UX.

"@angular/core": "^18.2.5", "@fundamental-ngx/core": "0.54.1",

1. Go to https://sapui5.hana.ondemand.com/#/entity/sap.ui.unified.Calendar/sample/sap.ui.unified.sample.CalendarMultipleDaySelection
2. Select a date
3. Press shift key
4. Select on another date

As a result a range of dates between those dates will be selected.

### StackBlitz example(s)

https://sapui5.hana.ondemand.com/#/entity/sap.ui.unified.Calendar/sample/sap.ui.unified.sample.CalendarMultipleDaySelection

### Declaration

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/13502

## Summary

- When `[allowMultipleSelection]="true"`, holding **Shift** while clicking a date now fills all dates between the last plain-clicked anchor and the shift-clicked date — matching SAP UI5 Calendar behaviour
- Plain-click toggle behaviour and all existing outputs are unchanged

## Changes

- **`CalendarDayViewComponent`**: two new focused outputs added for the shift-click feature:
    - `multiDateAnchorChange: EventEmitter<D>` — emitted on every plain click in multi single-selection mode; `CalendarComponent` uses it to track the shift anchor
    - `shiftMultiDateSelected: EventEmitter<D>` — emitted only when Shift is held; `CalendarComponent` uses it to fill the date range
    - `_toggleMultiDate` and `selectedMultipleDatesChange` are unchanged — plain clicks still go through the existing path with no breaking change
- **`CalendarComponent`**: owns `_shiftAnchorDate`, `handleMultiDateAnchorChange`, `handleShiftMultiDateSelected`, `_fillDateRange` (DatetimeAdapter-based cross-month arithmetic); anchor resets on `writeValue` and `allowMultipleSelection`/`calType` changes; `selectedMultipleDatesChanged` only emits `closeCalendar` when `allowMultipleSelection` is false
- **`calendar.component.html`**: wires the two new outputs
- **Docs**: shift-click behaviour is documented as a sentence in the existing "Multi Calendar" and "Multi Date Picker" sections (no separate example added)

## Test Plan

- [x] `nx run core:test` — all tests pass
- [x] `nx affected:lint` — no errors
- [x] Manual: open docs app, navigate to Calendar → "Multi Calendar", click a date, navigate to the next month, Shift+click — verify all dates in between are selected
- [x] Manual: verify plain clicks still toggle individual dates and `selectedMultipleDatesChange` fires as before
- [x] Manual: verify `fd-date-picker` popover stays open while building a multi-date selection

## Screenshots

Screenshot 2026-06-24 at 11 26 32

Screenshot 2026-06-24 at 11 23 42


mikerodonnell89 (commented):
Maybe I'm just confused but this bit:

> CalendarDayViewComponent.selectedMultipleDatesChange no longer fires on individual date clicks when allowMultipleSelection=true. Single-date clicks in multi-mode now emit the new selectedMultipleDateWithShiftChange

So `selectedMultipleDateWithShiftChange` fires with a single-date (no shift) click on the multi-mode? Seems a bit of a misnomer, if that is the case. I am wondering if breaking existing functionality is truly needed here?

mikerodonnell89 (commented):
Also, I don't think you need an entire additional example for the "shift" functionality, which can be demonstrated in the multi-mode selection example. I think the more we cloud the documentation, the more likely people will miss things. And I think there will likely be some confusion around the difference between multi-mode and range selection, so the fewer unnecessary demonstrations the better

mikerodonnell89 (commented):
Datetime picker coming later I'm guessing?

MariaIDineva:
> Also, I don't think you need an entire additional example for the "shift" functionality, which can be demonstrated in the multi-mode selection example. I think the more we cloud the documentation, the more likely people will miss things. And I think there will likely be some confusion around the difference between multi-mode and range selection, so the fewer unnecessary demonstrations the better

Agreed, thanks! I'll remove the separate examples and just add a sentence to the existing ones for multi-mode selection, describing the new feature.

MariaIDineva:
> Maybe I'm just confused but this bit:
> 
> > CalendarDayViewComponent.selectedMultipleDatesChange no longer fires on individual date clicks when allowMultipleSelection=true. Single-date clicks in multi-mode now emit the new selectedMultipleDateWithShiftChange
> 
> So `selectedMultipleDateWithShiftChange` fires with a single-date (no shift) click on the multi-mode? Seems a bit of a misnomer, if that is the case. I am wondering if breaking existing functionality is truly needed here?

You are not confused, you got it quite right. And, yeah, I agree that it's a misnomer and I did spend quite some time trying to figure out a better name, but whetever I came up with, sounded pretty much like the old `selectedMultipleDatesChange` event and did not bring any idea about the possibility to carry the "shift" click information, so I decided to keep that name.

But, you raise a very good point that it might be possible to implement it without a breaking change. I can keep the old `selectedMultipleDatesChange` that would fire on individual clicks and I can create separate other events - one for the "anchor" click (the one that the date range would start from) and one for the "shift-click" itself. That way I will manage to extend the functionality only by adding new events, fully preserving backwards compatibility.

MariaIDineva:
> Datetime picker coming later I'm guessing?

The datetime-picker doesn't expose `allowMultipleSelection`. Maybe you meant date-picker?

If that's the case `fd-date-picker` is already covered — it passes `allowMultipleSelection` straight through to `fd-calendar` and listens to the same event from the calendar `selectedMultipleDatesChange`, so shift-click works there without requiring any changes.

mikerodonnell89:
> > Datetime picker coming later I'm guessing?
> 
> The datetime-picker doesn't expose `allowMultipleSelection`. Maybe you meant date-picker?
> 
> If that's the case `fd-date-picker` is already covered — it passes `allowMultipleSelection` straight through to `fd-calendar` and listens to the same event from the calendar `selectedMultipleDatesChange`, so shift-click works there without requiring any changes.

 Ah yes apologies, I forgot the datetime picker doesn't support multi selection

mikerodonnell89 (approved):

PR #14310, opened 2026-06-22
fix(core): prevent inline-help text from being announced twice by screen readers

### LINKED ISSUE #14241: fd-inline-help directive causes Voice Over to repeat text
### Context in a nutshell

`fd-inline-help` directive is rendering the text twice -- one is visible the other is hidden -- and as a consequence, the Voice Over reads the text twice which is quite annoying for users who rely on such accessibility tool.

0.58.4 at least (maybe 0.62.3 as well, haven't tested)

### Affected Component

fd-inline-help

Look at the following StackBlitz code example.
https://stackblitz.com/edit/7xraz4ae?file=src%2Fapp%2Finfo-label-example.component.html

Turn on Voice Over and then click on the (?) button to get the inline help text. You will notice that the Voice Over feature reads the value twice. As a result, we have an accessibility issue. I received a ticket from the last Acessibility Audit for one of our MFEs and when I went to check how other MFEs deal with this I noticed they actually have the same issue. So although it only impacts a few people -- only Voice Over users -- they probably get this for all components (or at least many) that are using that directive.

### More Context

I tested this on the latest that we can use for Angular 20 (i.e. Fundamental v0.58.4), and it's what the MFEs are all using for the SAP product I'm working on (I won't disclose which one because this repo is public). I haven't tested for the Angular 21 versions (e.g. 0.62.3) so I don't know whether it's impacted or not.

Some more context on my findings:
When using the fd-inline-help directive, it ends up rendering the text twice -- one is the real value that's displayed on the screen and the other is a hidden text for the voice over. Since we have the text displayed on screen, the hidden one is unnecessary.

Image

### Priority

Low

### Declaration

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14241

### Problem
The `fd-inline-help` directive was causing screen readers (VoiceOver, JAWS) to announce the help text twice when the popover was open:

1. Once from the hidden screen-reader element (added for JAWS linear reading mode support in #9276)
2. Once from the visible popover content (via aria-describedby)

### Solution
Conditionally set `aria-describedby` based on the popover's open state:

When closed: `aria-describedby` references the popover body (enables linear reading mode access via the hidden element)
When open: `aria-describedby` is removed (visible content is naturally accessible, prevents duplication)

This preserves the fix for #9276 (JAWS linear reading mode) while eliminating the duplicate announcement in #14241.

### DIFF
--- a/libs/core/inline-help/inline-help.directive.ts
+++ b/libs/core/inline-help/inline-help.directive.ts
@@ -40,7 +40,7 @@ let inlineHelpId = 0;
     providers: [PopoverService],
     host: {
         '[class.fd-inline-help__trigger]': 'true',
-        '[attr.aria-describedby]': 'bodyId()'
+        '[attr.aria-describedby]': 'ariaDescribedBy()'
     }
 })
 export class InlineHelpDirective {
@@ -124,6 +124,13 @@ export class InlineHelpDirective {
         bodyRole: this.bodyRole,
         bodyId: this.bodyId
     });
+
+    /** @hidden Computed aria-describedby - only set when popover is closed to avoid duplication */
+    protected readonly ariaDescribedBy = computed(() =>
+        // When popover is open, the visible content is already accessible
+        // Only reference bodyId when closed for linear reading mode support
+        this._popoverService.isOpen() ? null : this.bodyId()
+    );
 
     /** @hidden */
     private readonly _popoverService = inject(PopoverService);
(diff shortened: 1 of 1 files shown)


MariaIDineva (dismissed):

MariaIDineva (approved):

droshev (approved):
🚀

PR #14312, opened 2026-06-23
feat(platform): update sorting functionality for platform table combined settings dialog

mikerodonnell89:
part of #13640 

Updates the "Sort" page on the combined table settings dialog to allow the user to sort on multiple properties simultaneously, and allows the user to prioritize the sort criteria.


libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
```diff
@@ -160,8 +166,12 @@ export class SettingsDialogComponent implements Resettable {
             ? { columns: this._pendingColumnsChanges.columns }
             : this.columnsData();
 
+        // Build sorting result with sortBy array
+        const sortingData = this.sortingData();
+        const sortingResult = sortingData ? { sortBy: this._pendingSortBy } : null;
```
MariaIDineva: There is a bug here that Claude caught:

> 
> The _pendingSortBy field is initialized as [] and is only populated when onSortChange fires. If the user opens the dialog, makes no changes to sorting, and clicks OK, sortingResult will be { sortBy: [] } — clearing all existing sort configuration. This is a data-loss regression on every "OK without sort changes" action.
> 
> The fix: initialize _pendingSortBy from sortingData()?.sortBy in ngOnInit or the constructor, or derive it from sortingData() directly in confirm() instead of caching in a field.
>

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
```diff
@@ -218,13 +236,23 @@ export class SettingsDialogComponent implements Resettable {
      * @param event Updated sorting data.
      */
     onSortChange(event: SettingsSortDialogResultData): void {
+        // Store the sortBy array for multi-column sorting
+        this._pendingSortBy = event.sortBy ?? [];
```
MariaIDineva: Another comment that I find meaningful:

> _pendingSortBy is a mutable plain array but is also set on sortingData via onSortChange
> 
> The sortingData signal already holds the sortBy array after onSortChange fires. The separate _pendingSortBy field is redundant and the source of the bug above. confirm() should read this.sortingData()?.sortBy ?? [] directly instead of maintaining a separate parallel field.
>

libs/platform/table/components/table-view-settings-dialog/sorting/sorting.component.html
```diff
@@ -1,63 +1,73 @@
...
-                    name="sort-by"
-                    [value]="column.key"
-                    state="default"
-                    [ngModel]="field()"
-                    (ngModelChange)="_sortFieldChange($event)"
-                ></fd-radio-button>
-                <span fd-list-title>{{ column.label }}</span>
-            </li>
+                    <fd-segmented-button
+                        style="margin-inline-end: 2rem"
```
MariaIDineva: Here as well?

libs/platform/table/components/table-view-settings-dialog/sorting/sorting.component.html
```diff
@@ -1,63 +1,73 @@
...
-        </li>
-    </ul>
-</fieldset>
+    <ul fd-list>
+        @for (criteria of sortCriteriaList(); track $index; let rowIndex = $index) {
+            @if (rowVisibility()[rowIndex]) {
+                <li fd-list-item class="fdp-sorting__row">
+                    <fd-combobox
+                        #sortCombobox
+                        style="margin-inline-end: 1rem"
```
MariaIDineva: Is it OK to have inline styles instead of using fundamental-styles?

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
```diff
@@ -97,6 +97,9 @@ export class SettingsDialogComponent implements Resettable {
     /** Signal to track reset availability */
     readonly isResetAvailable$ = signal(false);
 
+    /** Signal to track if the dialog is valid (OK button can be clicked) */
+    readonly isDialogValid$ = signal(true);
```
MariaIDineva: I don't think we need a `$` at the end of the name here since this is a signal, not an observable.

libs/platform/table/components/table-view-settings-dialog/settings-dialog/settings-dialog.component.ts
MariaIDineva: You might want to cover that case with a test as well.

libs/platform/table/components/table-view-settings-dialog/sorting/sorting.component.ts
```diff
@@ -79,42 +157,491 @@ export class SortingComponent {
...
+            if (column === '') {
+                // Empty string - clear any error state and don't update selection
+                // This can happen when combobox re-renders and emits empty value
+                shouldUpdate = false;
+                isInvalid = false;
+            } else {
+                const matchingColumn = this.columns().find((col) => col.label === column);
+                fieldKey = matchingColumn?.key ?? null;
+
+                if (rowIndex === 0) {
```
MariaIDineva: Another one from Claude:

>  _comboboxTextChanged is only tracked for rowIndex === 0 but resetAvailabilityChange uses it as a general "any row changed" indicator
> 
> The code only sets _comboboxTextChanged for rowIndex === 0. If a non-first row has an invalid text entry, resetAvailabilityChange may not reflect it correctly.

InnaAtanasova:
## Missing
### Tests:

- No E2E tests added for the new multi-sort UI interactions (move up/down, delete, add criterion)
- No tests for keyboard navigation through the new multi-row sort interface
- No tests verifying that invalid combobox input blocks the OK button
- No tests for the snapshot mechanism in TableInitialStateDirective

### Docs:

- The example at `libs/docs/platform/table/examples/platform-table-sortable-example.component.html` only adds `[sortable]="true"` to one column. Should there be a more comprehensive example demonstrating multi-column sorting capabilities?
- No documentation update explaining how to use the new multi-sort feature or how the priority order works

### Breaking Changes:

Interface changes to `SettingsSortDialogData` and `SettingsSortDialogResultData` remove field and direction properties, replacing them with `sortBy` array. This is a breaking change for anyone consuming these interfaces directly.

mikerodonnell89:
> * The example at `libs/docs/platform/table/examples/platform-table-sortable-example.component.html` only adds `[sortable]="true"` to one column. Should there be a more comprehensive example demonstrating multi-column sorting capabilities?

There are several other sortable properties in that table, I added `[sortable]="true"` to one column (Status) where many rows have identical values (Available, Out of Stock etc), to help the user more easily be able to see the prioritized sorting functionality. So they can sort by Status first, then Price second, for example, which would show all "Available" rows in the first few rows, sorted by price within that group.

> ### Breaking Changes:
> Interface changes to `SettingsSortDialogData` and `SettingsSortDialogResultData` remove field and direction properties, replacing them with `sortBy` array. This is a breaking change for anyone consuming these interfaces directly.

These are for internal use only

InnaAtanasova (dismissed):

InnaAtanasova (dismissed):

InnaAtanasova (approved):

Issue #14249, opened 2026-06-02
FlexibleColumn component has no accessible name


guilhermerenna:
### Describe your proposal

The expand/collapse buttons rendered inside the FlexibleColumn use only the title attribute to convey their purpose in Fundamental 0.58.x. Screen readers (e.g. VoiceOver on macOS) do not compute title as an accessible name when the button receives focus — they treat it as a tooltip only. As a result, the button has no accessible name and is announced as ? by VoiceOver.

Important note:
in 0.62.3 it works perfectly fine, so 0.58.x can probably use the same fix available in the newest version.

Unfortunately we cannot use the newest library because it's Angular 21, and we're stuck in Angular 20.

Image of the issue reproduced generically:

<img width="1766" height="1553" alt="Image" src="https://github.com/user-attachments/assets/f710c578-0cee-411d-b6fd-805a71428a35" />

### Which versions of Angular and Fundamental Library for Angular are affected?

Angular: 20, Fundamental Library: 0.58.4

### Affected Component

FlexibleColumn

### Steps to Reproduce (for bugs)

1. Open the VoiceOver accessibility feature or equivalent (In MacOS it's under Settings, you can search for VoiceOver and turn it on)
2. Navigate to a page that contains the FlexibleColumn design
3. Tab until you select the button that expands/shrinks the columns
4. Look/listen to what the reader announces

Actual result:
VoiceOver says "?, button, group"

Expected result:
VoiceOver should say "Expand first column, button, group"

### Priority

Low

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

droshev:
@guilhermerenna have you tested it with JAWS? this is the official supported tool? If not, can you try to reproduce it (or not) and come back to us?

guilhermerenna:
Hi deno!
This is the same as the other one: I haven't tested with JAWS myself, but the reps who did the Audit did test used JAWS 2026.2602.49.
As the other one, I cannot copy it directly here since it's internal docs, but it's all available inside the internal Jira CXMC-58864

InnaAtanasova:
hi @guilhermerenna  I added a fix for the issue and downported it. It's available with version 0.58.8. Please check and let us know if the issue is resolved on your side.

guilhermerenna:
Hi Inna,
sorry, I missed your last message! 

[...]

PR #14313, opened 2026-06-23
fix(core): add accessible names to FlexibleColumnLayout separator buttons

InnaAtanasova:
## Related Issue(s)

related to https://github.com/SAP/fundamental-ngx/issues/14249

## Summary

Fixes expand/collapse separator buttons in FlexibleColumnLayout not having accessible names for screen readers.

## Problem

The separator buttons only used the `title` attribute, which screen readers treat as a tooltip rather than an accessible name. This caused VoiceOver to announce the buttons as "?" instead of describing their purpose (e.g., "Expand first column").

## Solution

- Added `[ariaLabel]` bindings to both left and right separator buttons
- Implemented fallback chain: specific button title → general title → empty string
- Used button's `[ariaLabel]` input to prevent override by component's host binding

## Testing

- Added 4 accessibility tests verifying proper `aria-label` values
- Tests cover specific titles, fallback titles, and both separator buttons

### DIFF
--- a/libs/core/flexible-column-layout/flexible-column-layout.component.html
+++ b/libs/core/flexible-column-layout/flexible-column-layout.component.html
@@ -25,10 +25,15 @@
                 fdCompact
                 fdType="transparent"
                 class="fd-flexible-column-layout__button"
+                [ariaLabel]="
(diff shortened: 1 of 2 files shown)


droshev (dismissed):

droshev (approved):
🚀

PR #14314, opened 2026-06-24
[WIP] fix(platform): add missing text alternatives to fdp-table group button and sort icon

### LINKED ISSUE #14021: [platform][fdp-table] ACC-255.1/264.1 (Level A) Missing text alternatives on group button/sort icon
### 1. Missing text alternative for row Expand/Collapse buttons (ACC-255.1, Level A)
Row-level expand/collapse buttons in `fdp-table` have no accessible text alternative, making them unidentifiable to screen reader users.

Image

**Request:** Provide accessible `aria-label` for row expand/collapse buttons (e.g., "Expand row" / "Collapse row").

### 2. Sort icon buttons have no text alternative (ACC-255.1, Level A)
Sort buttons generated by column headers have no text alternative.

Image

**Request:** Column sort buttons should include accessible labels (e.g., "Sort by [column name] ascending/descending").

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14021

## Description

Fixes ACC-255.1 / ACC-264.1 (WCAG Level A) — missing text alternatives on the group row expand/collapse cell and column sort icon in `fdp-table`.

### 1. Group row expand/collapse cell (`table-group-row`)

The interactive `<td>` that toggles group row expansion had no accessible name, so screen readers announced the state (`aria-expanded`) without identifying what the element was.

- Added `[attr.title]` to the group row `<td>` using the existing `platformTable.expandRowButtonTitle` / `platformTable.collapseRowButtonTitle` i18n keys, matching the pattern already in place for tree rows.
- Used `title` instead of `aria-label` to avoid the accessible name propagating into the row's computed name (which caused VoiceOver to repeat "Collapse row" twice).
- Imported `FdTranslatePipe` into `TableGroupRowComponent`.

### 2. Column sort icon (`table-header-cell-content`)

The sort direction icon in column headers was `aria-hidden`, giving screen reader users no indication of which column is sorted or in which direction.

- Replaced `[ariaHidden]="true"` on the sort `fd-icon` with `[ariaLabel]` bound to a new translated string. The `fd-icon` component automatically sets `role="img"` and clears `aria-hidden` when `ariaLabel` is provided.
- The label reflects the current sort direction: **"Sort by {columnName}, ascending"** or **"Sort by {columnName}, descending"**.
- Added `[ariaHidden]="true"` to the filter icon (previously missing) — it is a purely decorative status indicator whose state is already communicated by the visible "Filtered by: …" bar, so hiding it from AT is correct.
- The sort icon intentionally receives `ariaLabel` instead, because the `<th>`'s `aria-sort` attribute alone does not include the column name.
- Imported `FdTranslatePipe` into `TableHeaderCellContentComponent`.

### 3. New i18n keys (`platformTable` namespace)

Two new keys added across all 37 locales (English default, pending translation):

| Key | Default value |
|-----|--------------|
| `platformTable.columnSortedAscLabel` | `Sort by {columnName}, ascending` |
| `platformTable.columnSortedDescLabel` | `Sort by {columnName}, descending` |

## Test

Open the **Platform Table → Column Sorting** example in the docs app (`yarn start`). With VoiceOver (macOS) or NVDA (Windows):

- Focus a sorted column header — the sort icon should be announced as e.g. **"Sort by Description, ascending"**.
- Open the **Platform Table → Groupable** example. Focus a collapsed/expanded group row — it should be announced as **"Expand row"** or **"Collapse row"** without duplication.

## Screenshots

Screenshot 2026-06-24 at 15 19 27
Screenshot 2026-06-24 at 15 18 29

### DIFF
--- a/libs/i18n/src/lib/models/fd-language-key-identifier.ts
+++ b/libs/i18n/src/lib/models/fd-language-key-identifier.ts
@@ -380,6 +380,8 @@ export type FdLanguageKeyIdentifier =
     | 'platformTable.cancelBtnLabel'
     | 'platformTable.clearFilters'
     | 'platformTable.collapseRowButtonTitle'
+    | 'platformTable.columnSortedAscLabel'
+    | 'platformTable.columnSortedDescLabel'
     | 'platformTable.confirmBtnLabel'
     | 'platformTable.defaultEmptyMessage'
     | 'platformTable.deselectAllCheckboxLabel'
--- a/libs/i18n/src/lib/models/fd-language.ts
+++ b/libs/i18n/src/lib/models/fd-language.ts
@@ -764,6 +764,8 @@ export interface FdLanguage {
         collapseRowButtonTitle: FdLanguageKey;
         expandRowButtonTitle: FdLanguageKey;
         rowNavigateButtonTitle: FdLanguageKey;
+        columnSortedAscLabel: FdLanguageKey<{ columnName: string }>;
+        columnSortedDescLabel: FdLanguageKey<{ columnName: string }>;
     };
     platformWizardGenerator: {
         summarySectionEditStep: FdLanguageKey;
--- a/libs/i18n/src/lib/translations/translations.properties
+++ b/libs/i18n/src/lib/translations/translations.properties
@@ -752,6 +752,10 @@ platformTable.cancelBtnLabel = Cancel
 platformTable.clearFilters = Clear Filters
 #XBUT: Label for the button, which collapses the clicked row in the table
 platformTable.collapseRowButtonTitle = Collapse row
+#XACT: ARIA label for sort icon when column is sorted ascending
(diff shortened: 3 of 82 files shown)


mikerodonnell89:
- "Sort up" is already being read by the screenreader, this screenshot is from main. But I guess we do need a space after "collapsed"
Screenshot 2026-06-24 at 1 38 58 PM

- I think the "Sort up" text would be better suited in the title attribute rather than aria-label
- The filter icon is not mentioned by the screenreader. I know the linked issue doesn't explicitly mention it by text, but they did highlight it in their screenshot. Guessing they want that as well

InnaAtanasova (approved):

MariaIDineva:
As per [latest comment](https://github.com/SAP/fundamental-ngx/issues/14021#issuecomment-5128468233) on the issue, no change is required on our side.

PR #14324, opened 2026-06-30
fix(platform): use live announcer for screenreader when expand all / collapse all buttons are clicked in platform table

### LINKED ISSUE #14113: [P3][Sourcing][Accessibility]: Platform table Expand and collapse buttons screenreader issue
### Describe your proposal

On activating ‘Expand all/Collapse all’ button, there is no information conveyed for the screen reader user about the successful operation.
Replicable: https://docs-0-57--fundamental-ngx.netlify.app/#/platform/table#tree

ng: 20, F-NGX platform: 0.57.6

### Affected Component

_No response_

On activating ‘Expand all/Collapse all’ button, there is no information conveyed for the screen reader user about the successful operation.
Replicable: https://docs-0-57--fundamental-ngx.netlify.app/#/platform/table#tree

### StackBlitz example(s)

https://docs-0-57--fundamental-ngx.netlify.app/#/platform/table#tree

### Priority

Medium

### Declaration

mikerodonnell89:
fixes #14113


libs/i18n/src/lib/translations/translations.properties
```diff
@@ -752,6 +752,8 @@ platformSwitch.ariaLabel = Switch input
 platformTable.cancelBtnLabel = Cancel
 #XBUT: Label for the button, which clears the filters
 platformTable.clearFilters = Clear Filters
+#XMSG: Announcement when collapse all button is clicked
+platformTable.collapseAllAnnouncementLabel=Collapse All clicked
```
MariaIDineva, line 756: I think something like `All rows expanded/collapsed` would sound more natural.

MariaIDineva (commented):
I am not sure why but when I test with voiceover, I don't hear the expected phrases pronounced. Maybe I am doing something wrong. I am testing `#/platform/table/basic#tree`. Any idea why I don't get the output?

mikerodonnell89:
> I am not sure why but when I test with voiceover, I don't hear the expected phrases pronounced. Maybe I am doing something wrong. I am testing `#/platform/table/basic#tree`. Any idea why I don't get the output?

Voiceover seems to want to prioritize explaining the toolbar, if you're just moving focus in to it for the first time. But subsequent clicks on the expand/collapse buttons will announce the "clicked" message

https://github.com/user-attachments/assets/50e3378a-4a83-40cf-8740-6c892b3e344d

libs/i18n/src/lib/translations/translations.properties
mikerodonnell89: I decided to have it announce "expand all clicked"/"collapse all clicked" because tree node expansion can sometimes mean making additional API calls (see "Handling large amounts of data" example [here](https://fundamental-ngx.netlify.app/#/platform/table/advanced)), which could potentially fail, in which case having the screenreader say "All rows expanded/collapsed" wouldn't really be true

MariaIDineva (approved):

InnaAtanasova (approved):

PR #14327, opened 2026-07-03
fix(core): improve menu ARIA attributes and VoiceOver compatibility

### LINKED ISSUE #14105: [Core][fd-menu] ACC-264 (Level A) Menu item count is not announced by screen reader when navigating
## Summary

When a user activates a button that opens a `fd-menu` and navigates through the menu items using arrow keys, the screen reader does not announce item count information (e.g., "Item 1 of 3"). Users are unable to determine how many items are in the menu or their current position within it.

## Current Behavior

When navigating through menu items inside a `fd-menu` using arrow keys, the screen reader only reads the item label — no count information (e.g., "X of Y") is announced.

## Expected Behavior

When navigating through menu items, the screen reader should announce item count information such as "Item 1 of 3" or "X items" so users understand the total number of options and their current position.

## Component

- `fd-menu` / `fd-menu-item` in `@fundamental-ngx/core`

## Accessibility Checkpoint

ACC-264 (Level A)

## Steps to Reproduce

1. Use a screen reader (e.g., JAWS Version 2026.2510.195)
2. Open a page with a `fd-menu`
3. Activate a button to open the menu
4. Navigate through menu items using arrow keys
5. Observe that item count information is **not** announced

## Environment

- Browser: Google Chrome
- Screen Reader: JAWS Version 2026.2510.195
- OS: Windows 11 x64

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14105

## Description
- **Added ARIA position attributes**: Menu items now have `aria-posinset` and `aria-setsize` set recursively for all menu levels
- **Dynamic role switching**: Parent menu items with submenus use `role="menuitem"` when collapsed (ARIA compliant) and `role="none"` when expanded (fixes VoiceOver counting)
- **Removed static role attributes**: Removed hardcoded `role="menuitem"` from `MenuItemComponent` host bindings in favor of setting it on interactive elements
- **Fixed @for tracking**: Updated menu example templates to track by stable identifiers instead of object references

### Technical Details

**New method `_setAriaPositionAttributesRecursive`**:
- Recursively sets ARIA position attributes on all menu items
- Filters out submenu descendants to only process direct children at each level
- Resubscribes to submenu changes to update attributes dynamically
- Runs in `afterNextRender` to ensure `@Input()` bindings are processed

**Dynamic role on `MenuInteractiveComponent`**:
- Returns `role="menuitem"` by default
- Returns `role="none"` when parent item is expanded (has submenu + `aria-expanded="true"`)
- Maintains ARIA compliance while fixing VoiceOver behavior

**Menu example fixes**:
- Changed tracking: `track active.itemId`, `track item.title`, `track subItem.title`

### DIFF
--- a/libs/core/menu/menu-interactive.component.ts
+++ b/libs/core/menu/menu-interactive.component.ts
@@ -21,9 +21,6 @@ import { MenuItemInputDirective } from './directives/menu-item-input.directive';
         <ng-template cdkPortalOutlet></ng-template>
         <ng-content></ng-content>
     `,
-    host: {
-        role: 'menuitem'
-    },
     changeDetection: ChangeDetectionStrategy.OnPush,
     imports: [PortalModule]
 })
@@ -40,6 +37,14 @@ export class MenuInteractiveComponent implements HasElementRef {
         ) {
             this._startAddonInstance = addon;
         }
+    }
+
+    /** @hidden */
+    @HostBinding('attr.role')
+    get role(): string {
+        // Parent items with submenus should not have role="menuitem" as it causes VoiceOver
+        // to count them in the position calculation. Only leaf items should be menuitems.
+        return this.ariaHaspopup ? 'none' : 'menuitem';
     }
 
     /** @hidden */
--- a/libs/core/menu/menu-interactive.component.ts
+++ b/libs/core/menu/menu-interactive.component.ts
@@ -42,9 +42,8 @@ export class MenuInteractiveComponent implements HasElementRef {
     /** @hidden */
     @HostBinding('attr.role')
     get role(): string {
-        // Parent items with submenus should not have role="menuitem" as it causes VoiceOver
-        // to count them in the position calculation. Only leaf items should be menuitems.
-        return this.ariaHaspopup ? 'none' : 'menuitem';
+        // Parent items switch to role="none" when expanded so VoiceOver doesn't count them.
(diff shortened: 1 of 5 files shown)


libs/core/menu/menu-interactive.component.ts
```diff
@@ -42,6 +39,13 @@ export class MenuInteractiveComponent implements HasElementRef {
         }
     }
 
+    /** @hidden */
+    @HostBinding('attr.role')
+    get role(): string {
```
MariaIDineva, line 44: Just a suggestion, feel free to ignore. Do you think we could turn this into:
```
host: {
  '[attr.role]': 'ariaHaspopup && ariaExpanded ? "none" : "menuitem"'
}
```
so that we use `host` instead of `@HostBinding`?

MariaIDineva (approved):

libs/core/menu/menu-interactive.component.ts
InnaAtanasova: `ariaHaspopup` and `ariaExpanded` are plain properties (not signals) and the expression in host: {} will only be evaluated once during component initialization. It won't update when those properties change later.

With @HostBinding getter - getters are re-evaluated on every change detection cycle, so they pick up property changes.

MariaIDineva: Ooh, right, they are not signals, that's why you did it this way. Thanks for clarifying!

PR #14331, opened 2026-07-07
fix: ng update

robertIsaac:
before: it report this error `Migrations for package (@fundamental-ngx/core) were not found.`
after: it passes


InnaAtanasova:
hi @robertIsaac ! Thanks for creating the PR. Could you write in the description what problem the proposed solution is addressing? Thank you!

InnaAtanasova (approved):

PR #14332, opened 2026-07-07
chore: upgrade to Angular 22

### LINKED ISSUE #14254: Support for Angular v22
this is just to track angular 22 dependencies progress, once they are supported I will create a PR

- [x] https://github.com/angular-eslint/angular-eslint/issues/3052
- [x] https://github.com/prettier/prettier/issues/18974
- [x] https://github.com/nrwl/nx/pull/35851
- [x] https://github.com/thymikee/jest-preset-angular/issues/3811

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14254

### 1. Replace deprecated `ChangeDetectionStrategy.Default` with `ChangeDetectionStrategy.Eager`

`ChangeDetectionStrategy.Default` is the deprecated alias for `ChangeDetectionStrategy.Eager` in Angular 22. All occurrences of `changeDetection: ChangeDetectionStrategy.Default` have been updated to use the non-deprecated `ChangeDetectionStrategy.Eager` instead.

### 2. Remove `ComponentFactoryResolver` usage

`ComponentFactoryResolver` is deprecated since Angular 14. Replaced with the modern `createComponent` API and direct `ViewContainerRef.createComponent(Type)`.

**`libs/cdk/utils/dynamic-component/dynamic-component.service.ts`**

- Replaced `componentFactoryResolver.resolveComponentFactory(type) + factory.create(injector)` with `createComponent(type, { environmentInjector, elementInjector })`
- Removed `ComponentFactoryResolver` from imports and constructor
- Migrated constructor injection to `inject()` fields (project convention for new/modified code)
- Marked `createDynamicModule` as `@deprecated` — it has no internal callers and relies on the deprecated NgModule-era APIs (`Compiler`, `NgModuleFactory`, `compileModuleAsync`). It will be removed in a future release.

**`libs/core/notification/notification/notification.component.ts`**

- Replaced `componentFactoryResolver.resolveComponentFactory(content) + containerRef.createComponent(factory)` with `containerRef.createComponent(content)` directly
- Removed `ComponentFactoryResolver` from imports and injected fields

### 3. Remove deprecated `downlevelIteration` compiler option

`downlevelIteration: true` was needed to correctly iterate ES6 iterables when targeting older JS environments. With `"target": "es2023"` already set in `tsconfig.base.json`, native iteration support is guaranteed and the flag is dead config. Removed.

### 4. Add ambient module declarations for side-effect imports

**`apps/docs/src/typings.d.ts`**, **`apps/e2e-harness/src/typings.d.ts`**

With `moduleResolution: bundler` and stricter TypeScript 6 type checking, bare side-effect imports now require ambient type declarations (TS2882):

- **CSS imports** (`import 'fundamental-styles/dist/margins.css'`) in `libs/docs/ui5-webcomponents` examples — covered by `declare module '*.css'`
- **Moment locale imports** (`import 'moment/locale/bg'` etc.) in the calendar i18n example — covered by `declare module 'moment/locale/*'`

Both declarations added to the typings files for both apps (`docs` and `e2e-harness`), since each is a separate Angular compilation that lazy-loads the same calendar example component.

### 5. Keep `baseUrl` and add `ignoreDeprecations` to `tsconfig.base.json`

**`tsconfig.base.json`**

`"baseUrl": "."` and `"moduleResolution": "node"` are both deprecated in TypeScript 6.0. Both are retained and suppressed via `"ignoreDeprecations": "6.0"` for two reasons:

1. **`baseUrl`** — the `ui5-webcomponents-base` and sibling ui5 libs have local `paths` entries pointing to `dist/libs/` output (pre-built declarations). These paths are relative to `baseUrl` and are inherited correctly by child tsconfigs. Removing `baseUrl` breaks Angular's ng-packagr compiler in partial compilation mode (TS2307 errors) even when `./`-prefixed paths are used. The correct long-term fix is TypeScript project references, which would eliminate the need for these `dist/` path aliases entirely.

2. **`moduleResolution: "node"`** — migrating to `"bundler"` causes regressions: Jest config files require `allowImportingTsExtensions: true` and `@types/jest` namespace resolution breaks. This migration is blocked until the project references work is done.

Both are tracked as a single follow-up item in https://github.com/SAP/fundamental-ngx/issues/14335.

### 6. Suppress `extendedDiagnostics` in tsconfig files (Angular 22 migration)

Angular 22 promotes two template diagnostics to warnings/errors:

- `nullishCoalescingNotNullable` — `??` used where the left-hand side is already non-nullable
- `optionalChainNotNullable` — `?.` used where the operand is already non-nullable

These are suppressed via `"extendedDiagnostics"` in affected tsconfig files to unblock the upgrade. Addressing each instance is tracked as follow-up tech debt in https://github.com/SAP/fundamental-ngx/issues/14334.

### 7. Fix `rootDir` violations in `moduleResolution: bundler` tsconfigs

TypeScript 6 + `moduleResolution: bundler` enforces `rootDir` strictly when following `paths` aliases into source files. Libs like `i18n` that import `@fundamental-ngx/cdk/utils` would pull in source files from outside their inferred `rootDir`, causing TS6059 errors.

Added `"rootDir": "../../"` to all affected `tsconfig.lib.json`, `tsconfig.json`, `tsconfig.spec.json`, and app tsconfig files using `moduleResolution: bundler`, making the rootDir the repo root so cross-lib source resolution is always within scope. The proper long-term fix (TypeScript project references) is tracked as a follow-up.

### 8. Fix `isOpen / isOpenChange` conflict in MenuComponent, PopoverComponent, and ProductSwitchComponent

**Files touched**
- libs/core/menu/menu.component.ts, libs/core/menu/menu.interface.ts, libs/core/menu/directives/menu-trigger.directive.ts
- libs/core/popover/popover.component.ts, libs/core/popover/popover.interface.ts, libs/core/popover/base/base-popover.class.ts
- libs/core/product-switch/product-switch/product-switch.component.ts, libs/core/product-switch/product-switch/product-switch.component.html

Angular 22 reports a conflict when a component has both:
1. `isOpen = model(false)` (which implicitly provides template output `isOpenChange`)
2. an explicit Angular output also named `isOpenChange`

That conflict was resolved by keeping `model(false)` for `isOpen` and using a non-Angular programmatic stream for compatibility:

```typescript
readonly isOpen = model(false);
readonly isOpenChange = new Subject<boolean>();
```

An effect now bridges model state to the Subject, emitting only on real open-state transitions (not on unrelated reactive reruns).

For Product Switch, the explicit template forwarding `(isOpenChange)="isOpenChange.emit($event)"` was removed from `libs/core/product-switch/product-switch/product-switch.component.html` to avoid duplicate binding/diagnostic issues. Programmatic subscribers are now supported via the Subject bridge in `libs/core/product-switch/product-switch/product-switch.component.ts`.

Menu details:
- `libs/core/menu/menu.component.ts`: `isOpen` remains a model; `isOpenChange` is a Subject; emission is transition-only.
- `libs/core/menu/menu.interface.ts`: contract uses `isOpen: ModelSignal<boolean>` and `isOpenChange: Subject<boolean>`.
- `libs/core/menu/directives/menu-trigger.directive.ts`: continues subscribing to `isOpenChange` for ARIA updates.

Popover details:
- `libs/core/popover/popover.component.ts`: same model + Subject pattern with transition-only emission.
- `libs/core/popover/popover.interface.ts`: aligned to `isOpenChange: Subject<boolean>`.
- `libs/core/popover/base/base-popover.class.ts`: guidance text updated to reflect the new pattern.

Result:
- Template consumers using `[(isOpen)]` and `(isOpenChange)` continue to work via the model-generated Angular output.
- Programmatic subscribers to `component.isOpenChange` continue to work via Subject.
- No duplicate isOpenChange Angular output collision.

### 9. Fix Angular 22 signal input propagation across all test suites

**Affected files: `libs/cdk`, `libs/core`, `libs/platform`, `libs/cx` — ~216 tests across 67 suites**

Angular 22 introduced a breaking change in zone-based tests: mutating a plain property on a wrapper component and calling `fixture.detectChanges()` no longer reliably propagates new values into signal inputs on child `OnPush` components. In Angular 21, `detectChanges()` would re-evaluate all template bindings in the tree and push new values down to child inputs synchronously. In Angular 22, signal inputs on `OnPush` children are only updated when Angular considers the child dirty — and a plain property mutation on the parent no longer consistently marks the child dirty in the zone-based test scheduler. Overriding the child to `ChangeDetectionStrategy.Eager` does not fix this because `Eager` only controls _when_ Angular checks a view, not _how_ template binding values are written into signal input nodes.

The fix was applied uniformly across all affected specs using one of four patterns:

1. **Wrapper property → `input()` signal:** Convert the wrapper's plain property to `readonly prop = input(defaultValue)`, update the template to call it (`[binding]="prop()"`), and update test mutations to `fixture.componentRef.setInput('prop', value)`. Signal-to-signal propagation via template binding still works in Angular 22; only the plain-property-to-signal path broke.
2. **Direct fixture:** For tests that don't need content children, bypass the wrapper entirely with `TestBed.createComponent(Component)` and call `setInput` on the component directly.
3. **ngModel bindings:** For `[(ngModel)]`-bound values, `component.ngControl.setValue(val)` fires the RxJS `valueChanges` observable that the directive subscribes to — direct property mutation does not.
4. **Injected service mutations:** For `OnPush` components reading from an injected service (not an input), fix the component to call `markForCheck()` after mutating state (see `AvatarGroupLegacyComponent` bug fix below).

Also upgraded `jest-preset-angular` from `16.0.0` to `17.0.0` (Angular 22 peer dependency range) and bumped the full jest suite to latest: `jest@30.4.2`, `jest-environment-jsdom/node/util@30.4.1`, `jest-extended@7.0.0`, `ts-jest@29.4.11`, `@nx/jest@23.0.1`, `ngx-cva-test-suite@2.0.1`, `ng-packagr@22.0.1`.

### 10. Fix `AvatarGroupLegacyComponent._onResize()` not marking view for check

**`libs/core/avatar-group-legacy/avatar-group-legacy.component.ts`**

`_onResize()` mutates `overflowItemsCount` and hides/shows avatar elements, but did not call `markForCheck()` afterward. With `OnPush` change detection, this meant the overflow button `@if (overflowItemsCount > 0)` would not re-evaluate unless Angular happened to mark the view dirty through the zone (which it does in production when the viewport ruler fires inside a zone-tracked subscription, but not if `_onResize` is ever called outside the zone).

Fixed by injecting `ChangeDetectorRef` and calling `this._cdr.markForCheck()` at the end of `_onResize()`. The test was also simplified — the `markForCheck()` workaround in the spec is removed now that the component handles it correctly.

### 11. Fix TypeScript 6 strict catch typing across schematics and executors

TypeScript 6 makes `catch (e)` variables typed as `unknown` by default (`useUnknownInCatchVariables: true`). Several files across schematics and executors were passing the caught value directly to APIs expecting `Error` or `string`. Fixed using the safe pattern `error instanceof Error ? error.message : String(error)` throughout, which handles thrown strings and non-Error objects correctly.

**Files affected:**
- `libs/core/schematics/add-styles/index.ts` — `as Error` cast replaced with `instanceof` guard
- `libs/core/schematics/add-dependencies/add-dependencies.spec.ts` — two `catch (e)` clauses annotated `e: unknown` with `(e as Error).message` cast
- `libs/nx-plugin/src/executors/i18n-manage/executor.ts` and `operations/*.ts` — same pattern applied across executor and all operation files
- `libs/webc-generator/src/executors/generate/executor.ts` — same fix
- `libs/core/schematics/add-theming/index.ts` — `context.logger.debug(e)` wrapped with `String(e)`

Also added `"ignoreDeprecations": "6.0"` to the standalone schematics tsconfigs (`libs/core/schematics/tsconfig.json`, `libs/platform/schematics/tsconfig.json`, `libs/cx/schematics/tsconfig.json`) — these don't extend `tsconfig.base.json` and had their own `baseUrl` and `moduleResolution: "node"` producing TypeScript 6 deprecation errors.

### 12. Fix `core-schematics` test suite compilation under TypeScript 6 + ts-jest 29

**`libs/core/schematics/tsconfig.spec.json`**

Two issues required fixes:

1. **ts-jest legacy compiler failure** — without `isolatedModules: true`, ts-jest 29 uses its full TypeScript program compiler. Under TypeScript 6, this failed with `"outDir is '' or '.'"` when resolving the output directory for the schematics project. Added `"isolatedModules": true` to switch ts-jest to the transpile-only path, consistent with every other library's `tsconfig.spec.json`.

2. **TS5011 rootDir violation** — `isolatedModules: true` causes TypeScript 6 to enforce `rootDir` strictly. Each spec file lives in its own subdirectory (`add-dependencies/`, `add-styles/`, `add-theming/`), so TypeScript inferred a different `rootDir` per file and rejected them all with TS5011. Added `"rootDir": "."` to pin it to the schematics root, covering all subdirectories.


InnaAtanasova:
This is Claude PR review summary:

🚨 Blocking Issues
1. **88 occurrences of `$safeNavigationMigration()` in production templates** - These are temporary compiler migration markers that must be manually removed. Either remove the `?.` where non-nullable is confirmed, or suppress the diagnostic per-line if needed.

2. **MenuComponent missing dedicated `isOpenChange` emission effect** - Unlike PopoverComponent and ProductSwitchComponent which got dedicated constructor effects, **MenuComponent** reuses an existing complex effect. This creates inconsistency and risk if that effect has conditional logic that skips emission.

3. **ProductSwitchComponent effect won't emit initial state** - The effect initializes `previousIsOpen = false` but only emits on changes, so programmatic subscribers won't get the initial value. This breaks parity with Angular 21 output behavior.

💡 Suggestions
1. **Document `baseUrl` retention in tsconfig.base.json** - Add comment explaining ng-packagr dependency and link to #14335
2. **Add comments to all `extendedDiagnostics` blocks** - Link to #14334 follow-up issue
3. **Improve `createDynamicModule` deprecation message** - Current message is circular ("removed when removed")
4. **Add tests for `isOpenChange` Subject subscriptions** - Verify programmatic subscribers still work correctly

📋 Missing
1. Tests verifying programmatic `isOpenChange.subscribe()` behavior for `menu/popover/product-switch`
2. JSDoc documenting the `isOpenChange` type change (OutputEmitterRef → Subject) and behavior

-----------------------------------------------------------------------------------
Additionally to this I ran Copilot review and this is the summary:
**1. High: Duplicate isOpenChange emissions in ProductSwitch for user-driven open/close transitions**

**Evidence:** [product-switch.component.html:14](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) forwards popover isOpenChange directly to the Subject, while [product-switch.component.ts:95](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to [product-switch.component.ts:103](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) emits again from the effect when isOpen changes.

**Impact:** Programmatic subscribers to ProductSwitch isOpenChange can receive two events for one UI toggle, causing duplicated side effects (analytics calls, state sync, API calls, etc.).

**Suggested fix:** Remove template forwarding at [product-switch.component.html:14](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and keep the effect as the single emission path.


**2. Medium: createDynamicModule changed replacement semantics and now keeps existing views in container**

**Evidence:** [dynamic-component.service.ts:55](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to [dynamic-component.service.ts:67](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) creates into the provided container without clearing it first.

**Impact:** Repeated calls can accumulate multiple dynamic components in the same container where previous behavior replaced content. Even though deprecated, this remains a callable public method and can regress existing consumers.

**Suggested fix:** Re-introduce container clear before create, or explicitly document this as a breaking behavior change.

**3. Low: Missing regression test for duplicate emission path**

**Evidence:** [product-switch.component.spec.ts:67](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to [product-switch.component.spec.ts:80](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) only checks that true/false were emitted, not emission count per transition, and does not exercise popover event forwarding path.

**Impact:** The duplicate-emission regression above can pass current tests undetected.

**Suggested fix:** Add an assertion for exact emission sequence/count when toggling via popover output path.

InnaAtanasova:
Avatar Group Legacy. This is just something I noticed while testing and mention it because there were changes done on this component. I can see that before this PR the component had a similar behavior.

https://github.com/user-attachments/assets/ff00c130-e254-4032-87dd-5f578a0001dc

mikerodonnell89 (commented):
There are lots of mentions of "Angular 21" in the docs we'll probably want to bump up

InnaAtanasova:
Platform Settings Generator doesn't render in the first example:
Screenshot 2026-07-13 at 3 23 29 PM

On click on the Save button it renders, but there's a console error:
Screenshot 2026-07-13 at 3 24 10 PM

InnaAtanasova:
I tested all components in Core and Platform, randomly tested 10+ components from the wrappers and I couldn't find more issues that the ones mentioned above.

mikerodonnell89:
> Platform Settings Generator doesn't render in the first example

Snapshot testing should've caught this right?

MariaIDineva:
> This is Claude PR review summary:
> 
> 🚨 Blocking Issues
> 
> ...
## Claude AI review summary response

### 🚨 Blocking Issue 1: `$safeNavigationMigration()` in production templates

**Assessment: Not a blocker for this PR.**

`$safeNavigationMigration()` is an Angular compiler migration helper emitted by `ng migrate` for `optionalChainNotNullable` diagnostics. All 79 occurrences are suppressed via `extendedDiagnostics` in the affected tsconfigs — the build, tests, and runtime are unaffected. They were introduced in this PR as part of suppressing the Angular 22 `optionalChainNotNullable` / `nullishCoalescingNotNullable` extended diagnostics. Tracked as tech-debt alongside #14334.

---

### 🚨 Blocking Issue 2: MenuComponent missing dedicated `isOpenChange` emission effect

**Addressed.** The effect now guards the emission with `if (openState !== previousIsOpen)`, identical to the pattern in `ProductSwitchComponent`. The emission is no longer unconditional — it only fires when the open state actually changes, eliminating the risk of spurious emissions even though it remains co-located with the `beforeOpen` and service-sync logic.

---

### 🚨 Blocking Issue 3: ProductSwitchComponent effect won't emit initial state

**Assessment: Correct behavior, not a bug.**

`isOpen` is initialized to `false`. Since the component starts closed and `previousIsOpen` is also initialized to `false`, the effect deliberately skips the initial emission (there is no state change). This matches standard Angular `output()` behavior where outputs do not emit on construction. Programmatic subscribers subscribing _before_ any state change will receive all subsequent transitions. If a subscriber needs the current state synchronously, they can read `component.isOpen()` directly.

---

### 💡 Suggestion 1: Document `baseUrl` retention in tsconfig.base.json

**Fixed.** Added an inline comment to `tsconfig.base.json` explaining why `baseUrl` and `moduleResolution: node` are retained and linking to #14335.

---

### 💡 Suggestion 2: Add comments to `extendedDiagnostics` blocks

**Fixed.** Added `// TODO(#14334): suppress Angular 22 optionalChainNotNullable and nullishCoalescingNotNullable diagnostics` above every `extendedDiagnostics` block across all 55 affected tsconfig files.

---

### 💡 Suggestion 3: Improve `createDynamicModule` deprecation message

**Fixed.** Updated the JSDoc on `createDynamicModule` in [dynamic-component.service.ts](libs/cdk/utils/dynamic-component/dynamic-component.service.ts) to explain the replacement pattern (`ViewContainerRef.createComponent()` with dynamic `import()`), that it has no internal callers, and that its `NgModuleFactory`/`Compiler` dependencies are themselves deprecated since Angular 14.

---

### 💡 Suggestion 4: Add tests for `isOpenChange` Subject subscriptions

**Fixed.** All three components now have dedicated `isOpenChange` Subject subscription tests:
- `ProductSwitchComponent` — "should not emit isOpenChange on initial render" + "should emit isOpenChange when open state changes" (exact sequence `[true, false]`)
- `MenuComponent` — "should emit isOpenChange when open state changes" + "should not emit isOpenChange when open state does not change"
- `PopoverComponent` — "should emit isOpenChange when open state changes" + "should still emit isOpenChange and beforeOpen events in mobile mode"

---

### 📋 Missing: JSDoc documenting `isOpenChange` type change

**Fixed.** All three components have JSDoc on `isOpenChange` noting both the purpose and the backwards-compatibility rationale (`Subject<boolean>` for programmatic subscribers). `MenuInterface` also reflects the updated `Subject<boolean>` type.

---

## Copilot review summary response

### High: Duplicate isOpenChange emissions in ProductSwitch

**Fixed.** This was a real bug. The template forwarded `(isOpenChange)="isOpenChange.next($event)"` from the inner `fd-popover`, while the constructor effect _also_ emits on every `isOpen` change. For UI-driven transitions, this caused double emission.

**Fix:** Removed the template `(isOpenChange)` binding from `product-switch.component.html`. The effect in the constructor is now the sole emission path. The two-way `[(isOpen)]` binding on `fd-popover` ensures `isOpen` model signal stays in sync with popover state changes, which the effect then forwards to `isOpenChange`.

**Also updated the spec** to assert `toEqual([true, false])` (exact sequence/count) instead of `toContain(true)` / `toContain(false)`, so this regression cannot silently pass again.

---

### Medium: `createDynamicModule` changed replacement semantics — no container clear

**Fixed.** The original implementation did call `containerRef.clear()` before creating the component — this was accidentally dropped during the `ComponentFactoryResolver` removal. `containerRef.clear()` has been restored, preserving the original replace-not-accumulate semantics.

---

### Low: Missing regression test for duplicate emission path

**Fixed.** The updated spec now uses `toEqual([true, false])` to assert the exact emission sequence. This catches any double-emission regression regardless of which code path triggers it.

MariaIDineva:
> Platform Settings Generator doesn't render in the first example: Screenshot 2026-07-13 at 3 23 29 PM
> 
> On click on the Save button it renders, but there's a console error: Screenshot 2026-07-13 at 3 24 10 PM

That's a very nice catch, thank you!

Here's a short explanation of the root cause and the fix:

`SettingsGeneratorDefaultExampleComponent` built its `schema` (bound to the child `fdp-settings-generator`'s `[settings]` input) as a plain property assigned inside `ngAfterViewInit()`, relying on an implicit follow-up change-detection tick to propagate it into the child's `@Input()` setter. That implicit tick reliably happened under Angular 21/zone.js, but is no longer guaranteed for this pattern under Angular 22 — so the example rendered blank until an unrelated event (the Save click) forced a fresh CD pass, which also surfaced a console error from the late mutation.

Refactored the example to the Angular 21+ signal pattern instead of patching it with `ChangeDetectorRef`: `@ViewChild()` → `viewChild.required()` signal queries, and the `schema` property → a `computed()` signal built from them. The template now binds `[settings]="schema()"`. Since `computed()`/`viewChild()` are part of Angular's reactive graph, the UI updates automatically as soon as the queried templates resolve — no manual `detectChanges()`/`markForCheck()`, no lifecycle-timing race.

Files: `settings-generator-default-example.component.ts` / `.html` (default example only — other 3 settings-generator examples already guarded against this via `ChangeDetectorRef.detectChanges()` and are unaffected).

As a follow-up, I created another [issue](https://github.com/SAP/fundamental-ngx/issues/14364) to investigate if we have any other similar examples (or even components) where change detection might be causing bugs.

P.S. The error in the console was not at all the culprit. That was also fixed, though.

MariaIDineva:
> > Platform Settings Generator doesn't render in the first example
> 
> Snapshot testing should've caught this right?

You were quite right to put that question on the table. And I initially thought that maybe we did not have e2e tests for this component but I was wrong.

After my fix on the Platform Settings Generator, some e2e tests related to it started failing so I spent some time investigating those failures. We had an accessibility test and a visible test failing.

Interestingly, it seems like we had a [blank screen](https://github.com/SAP/fundamental-ngx/blob/main/apps/e2e-harness/e2e/snapshots/darwin/chromium/platform-settings-generator-default.png) uploaded for one of the Platform Settings Generator e2e tests from the start which only now started failing because this is what we see after my latest changes (and what was actually on main):
Screenshot 2026-07-16 at 14 29 24

So now I am really wondering why this was not failing on main, since this bug is only present in the current PR, after upgrading to Angular 22...

And I am currently investigating two more blank screenshots of Platform Settings Generator, which also seem to have been there ever since we introduced e2e tests. [One of them](https://github.com/SAP/fundamental-ngx/blob/main/apps/e2e-harness/e2e/snapshots/darwin/chromium/platform-settings-generator-custom-layout.png) is used on a route that is [skipped](https://github.com/SAP/fundamental-ngx/blob/main/apps/e2e-harness/e2e/config/e2e.skip.json#L272) and [the other one](https://github.com/SAP/fundamental-ngx/blob/main/apps/e2e-harness/e2e/snapshots/darwin/chromium/platform-settings-generator-custom-control.png) is used in a test that is actually ran and gives the false impression that the component is OK and the tests pass. But when I run the e2e harness `npx nx serve e2e-harness` and navigate to the route that potentially tests the component, this is what I get:
Screenshot 2026-07-16 at 15 11 29

Soo, I'm trying to figure out why those tests were introduced and how to proceed with these 2 problematic screenshots.

MariaIDineva:
So I've updated the two failing e2e tests and I've created [a separate issue](https://github.com/SAP/fundamental-ngx/issues/14366) to fix the other problems I encountered with the Platform Settings Generator e2e tests as they are not at all related to Angular 22 upgrade.

InnaAtanasova (approved):

PR #14333, opened 2026-07-07
fix(core): address Shell Bar design review comments

### LINKED ISSUE #14225: Shellbar desing update/review
Image

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14225

## Description
- verified the changes that come from fund-styles
- fix the problem with Shell Bar buttons not applying `fd-shellbar__button` class when the type is default (transparent)
- removed the assistive tools from the example

Comments on the screenshot from the issue 14225:
- Shell Bar shadow is fixed, comes from the updated of fund-styles
- action buttons now have `fd-shellbar__button` class, except the buttons that explicitly have type set. See [issue](https://github.com/SAP/fundamental-ngx/issues/13833) and related [PR](https://github.com/SAP/fundamental-ngx/pull/13836). 
- Product identifier and separator spacings are updated in fund-styles
- The comment "Overflow has wrong style" is not relevant as discussed with the designers. There will be a complete redesign of this in the future.


libs/core/shellbar/shellbar-action/shellbar-action.component.html
```diff
@@ -1,8 +1,8 @@
 <div class="fd-shellbar__action fd-shellbar__action--desktop">
     <button
         fd-button
-        [fdType]="fdType() ?? 'transparent'"
```
InnaAtanasova, line 4: ?? is not needed as the default type is set to transparent

libs/core/shellbar/shellbar-action/shellbar-action.component.html
```diff
@@ -1,8 +1,8 @@
...
-        [class.fd-shellbar__button]="!fdType()"
```
InnaAtanasova, line 5: when the type is 'transparent' aka the user has not specified a type, it should apply the class `fd-shellbar__button`. This allows Shell Bar to have Toggle buttons with states.

mikerodonnell89 (dismissed):

mikerodonnell89 (approved):

PR #14348, opened 2026-07-09
fix(ui5): do not attempt to write undefined value to ui5-wc generated component

### LINKED ISSUE #14337: bug(ui5): `RouterLink` on `ui5-link` / `ui5-breadcrumbs-item` doesn't render `href`, breaking new-tab / hover-preview / modifier-click semantics
### Describe your proposal

# bug(ui5): `RouterLink` on `ui5-link` / `ui5-breadcrumbs-item` doesn't render `href`, breaking new-tab / hover-preview / modifier-click semantics

## Describe the bug

When Angular's `RouterLink` directive is applied to a `ui5-*` component from `@fundamental-ngx/ui5-webcomponents` (e.g. `ui5-link`, `ui5-breadcrumbs-item`), the resolved URL is **not** propagated to the host element's `href` attribute. As a result, the shadow-DOM anchor that UI5 renders for the item has no `href`, and every browser affordance that depends on a real link is broken:

- Right-click → **"Open link in new tab"** is missing from the context menu.
- Hovering the element does **not** show the target URL in the browser status bar.
- **Middle-click** and **Cmd/Ctrl/Shift-click** do **not** open the link in a new tab — they SPA-navigate in the current tab (because the click event is a synthesized `CustomEvent` without modifier keys, see below).
- Browser extensions that operate on `<a href>` (link previews, "copy link", "open all links") don't see the element as a link.

There is a **second, related** problem: even if the consumer sets `[href]` explicitly (bypassing `RouterLink`), the wrapper's generated input→property `effect()` **clobbers** the value on the next microtask when the input signal is undefined, silently wiping any `href` written imperatively (or by an accessibility helper, `RouterLinkActive`, etc.).

## Environment

- `@fundamental-ngx/ui5-webcomponents`: `0.63.0`
- `@ui5/webcomponents`: `2.23.1`
- Angular: `22.0.4`
- Chrome / Safari / Firefox — behaviour is browser-agnostic (root cause is DOM state, not UA-specific)

### (1) `RouterLink` is not translated to `href`

`RouterLink` sets `href` on the host element only when the directive resolves the element to a **standard `<a>` anchor** — its `_setUpAttributes` special-cases the anchor tag. For any other element type (including UI5 custom elements), `RouterLink` skips the attribute write and relies on the framework catching the click event.

That's fine for a plain `<button>`, but UI5 components render **their own anchor inside shadow DOM** and expect a real `href` on the host so they can propagate it into the shadow anchor. The wrapper is the natural place to bridge this gap.

### (2) The generated input-sync `effect()` clobbers external `href` writes

Each wrapper component defines an input signal like `href = input<string | undefined>()` and, in `ngAfterViewInit`, registers roughly:

```ts
effect(() => {
  (this.wcElement as any).href = this.href();
});
```

When the consumer doesn't bind `[href]`, `this.href()` is `undefined`. The effect fires on every consumer update and on setup, writing `undefined` onto the element. Any `href` set imperatively (by the app, by an integration wrapper, by `RouterLink` if we route around root cause (1)) is wiped on the next microtask.

Concretely: property writes on the host (`element.href = url`) get clobbered by the parent's input-sync `effect()` on the next microtask. Setting the attribute via `setAttribute('href', url)` survives longer, but the same `effect()` re-assigns the property, which the browser reflects back to the attribute — so even attribute writes get cleared unless defended by a `MutationObserver` watchdog.

### (3) `stopImmediatePropagation` + synthesized `CustomEvent` strips modifier keys

Related but downstream — `ui5-breadcrumbs-item` (and other UI5 child elements per [UI5/webcomponents#13304](https://github.com/UI5/webcomponents/issues/13304)) intercepts the native `click`, calls `stopImmediatePropagation()`, and dispatches a **`CustomEvent('click')`** instead. `RouterLink#onClick` reads `ctrlKey`/`metaKey`/`shiftKey`/`button` off the event — those properties don't exist on a `CustomEvent`, so `RouterLink` treats every click as a plain unmodified left-click and does a same-tab SPA navigation, ignoring the user's intent to open in a new tab.

This is the piece [UI5/webcomponents#13330](https://github.com/UI5/webcomponents/issues/13330) / [PR #13353](https://github.com/UI5/webcomponents/pull/13353) partially addressed by exposing modifier flags on `event.detail`, but the wrappers don't currently forward those to `RouterLink`.

## Proposed fixes

Ordered by impact:

1. **Emit `href` from the generator when `RouterLink` is present.** In the `ui5wc` Nx generator (introduced in [#13511](https://github.com/SAP/fundamental-ngx/pull/13511)), for components whose UI5 API has an `href` property, inject a small piece of Angular integration:
   - Detect `RouterLink` on the host via `inject(RouterLink, { optional: true, self: true })`.
   - Compute the URL via the public `routerLink.urlTree` getter + `LocationStrategy.prepareExternalUrl` (base-href aware).
   - Set it on the host with `setAttribute('href', url)`.
   - Re-run on `router.events` filtered to `NavigationEnd` (for relative-path `[routerLink]` inputs whose resolution changes with the active route).
   - Ideally also re-run when the `[routerLink]` input reference changes — an `effect()` reading `routerLink.urlTree` won't track cleanly because `.urlTree` uses `untracked()` internally; a small `afterEveryRender` or explicit subscription is fine.

2. **Stop clobbering imperatively-set attributes.** In the generated input→property sync `effect`, skip the write when the input signal is `undefined` **and** the current host property/attribute is truthy. This makes it safe for consumers (or Angular integrations like RouterLink) to write `href` imperatively without racing the wrapper. Better still, only write when the input signal value has actually changed since the previous read, so idempotent absence doesn't clear existing state.

3. **Forward modifier keys to synthesized click handlers.** For components where UI5 fires a `CustomEvent('click')` (Breadcrumbs, Link, TableRow, etc.), the wrapper should surface the `ctrlKey`/`metaKey`/`shiftKey`/`button` flags from `event.detail` onto the event the framework sees, so `RouterLink` (and any application code) can honour "open in new tab" intent. Alternatively, when the wrapper detects `RouterLink` and a modifier click, call `window.open(href, '_blank', 'noopener,noreferrer')` and cancel the SPA navigation.

4. **Document the workaround / recommended pattern.** At minimum, document how consumers should combine `[routerLink]` with `ui5-link` / `ui5-breadcrumbs-item` today. The current docs suggest they "just work" with `[href]`, which is misleading given (2).

## Reference workaround

For anyone hitting this before an upstream fix lands, the following pattern works but is significantly more ceremony than most consumers would reasonably expect `[routerLink]` to require:

- Extend the wrapper component (`Ui5BreadcrumbsItemComponent`, `Ui5LinkComponent`, etc.) in a thin custom Angular component with the same selector.
- In `ngAfterViewInit`, call `super.ngAfterViewInit()` first so the wrapper's input-sync `effect()`s are still installed for `[href]`/`[target]`/... on non-routerLink usages.
- Read `RouterLink` via `inject(RouterLink, { optional: true, self: true })`. If absent, bail — the wrapper is transparent.
- If present, compute the URL via `routerLink.urlTree` + `LocationStrategy.prepareExternalUrl` and write it via `setAttribute('href', url)`. Property writes (`element.href = url`) get clobbered by the parent's input-sync `effect()`; the attribute path is more resilient but still needs defence.
- Use `queueMicrotask` for the initial write so the parent's `undefined`-writing `effect()` flushes first.
- Re-run the sync on `router.events` filtered to `NavigationEnd` (relative `[routerLink]` inputs resolve differently under a different active route).
- Install a `MutationObserver` on the host's `href` attribute as a watchdog — if anything (notably the parent's `effect()`) clears it, re-apply. Guard against re-entry with an equality check inside the sync function so the observer becomes a no-op once the attribute matches.
- For click routing, override the `item-click` / `click` handler and read `event.detail.ctrlKey/metaKey/shiftKey`. On a modifier click, `window.open(href, '_blank', 'noopener,noreferrer')`. On a plain click, synthesize a native click on the item element so `RouterLink#onClick` sees a real `MouseEvent` (`button === 0`, no modifiers) and performs a normal SPA navigation.

This is a lot of ceremony for what most consumers would reasonably expect `[routerLink]` to Just Do — hence this issue.

## Related

- Umbrella: [#13176](https://github.com/SAP/fundamental-ngx/issues/13176) — Merge fundamental-ngx and ui5-webcomponents-ngx.
- Generator: [#13511](https://github.com/SAP/fundamental-ngx/pull/13511) — feat: bring ui5wc generator.
- Open questions tracker: [#14078](https://github.com/SAP/fundamental-ngx/issues/14078), [#13899](https://github.com/SAP/fundamental-ngx/issues/13899).
- Upstream semantic click event: [UI5/webcomponents#13304](https://github.com/UI5/webcomponents/issues/13304), [#13330](https://github.com/UI5/webcomponents/issues/13330), [PR #13353](https://github.com/UI5/webcomponents/pull/13353).

Angular: `22.0.4`, `@fundamental-ngx/ui5-webcomponents`: `0.63.0`, `@ui5/webcomponents`: `2.23.1`

### Affected Component

@fundamental-ngx/ui5-webcomponents: `ui5-link` and `ui5-breadcrumbs-item`

```html
<ui5-breadcrumbs>
  <ui5-breadcrumbs-item [routerLink]="['/things']">Things</ui5-breadcrumbs-item>
  <ui5-breadcrumbs-item [routerLink]="['/things', thing.id]">
    {{ thing.name }}
  </ui5-breadcrumbs-item>
</ui5-breadcrumbs>

<ui5-link [routerLink]="['/items', item.id]">
  {{ item.name }}
</ui5-link>
```

**Expected:** the resulting DOM has `<ui5-breadcrumbs-item href="/things">` / `<ui5-link href="/items/123">`; the inner shadow anchor inherits a real href; right-click, hover-preview, middle-click, Cmd/Ctrl-click all behave as normal browser links.

**Actual:** no `href` attribute is set on the host. The inner shadow anchor has no href either. Only same-tab SPA navigation works, via UI5's `item-click`/`click` `CustomEvent` path.

### StackBlitz example(s)

_No response_

### Priority

Medium

### Declaration

mikerodonnell89:
fixes #14337

### DIFF
--- a/libs/webc-generator/src/executors/generate/component-template.spec.ts
+++ b/libs/webc-generator/src/executors/generate/component-template.spec.ts
@@ -503,4 +503,38 @@ describe('componentTemplate', () => {
             expect(output).toContain("events: ['change', 'input']");
         });
     });
+
+    describe('input synchronization', () => {
+        const linkDeclaration: CEM.CustomElementDeclaration = {
+            kind: 'class',
+            name: 'Link',
+            tagName: 'ui5-link',
+            customElement: true,
+            members: [
+                {
+                    kind: 'field',
+                    name: 'href',
+                    privacy: 'public',
+                    type: { text: 'string | undefined' },
+                    default: 'undefined'
+                } as CEM.ClassField
+            ]
+        };
+
+        it('generates guard code to prevent clobbering imperatively-set properties', () => {
+            const output = componentTemplate(linkDeclaration, [], PACKAGE);
+            // This test verifies the GENERATOR produces the guard code.
+            // The guard prevents the effect from writing undefined when a property
+            // (like href) has been set imperatively by RouterLink or other code.
+            expect(output).toContain('// Skip writes when the signal is undefined and the property is already truthy');
+            expect(output).toContain('if (value === undefined && wcElement[inputName]) {');
+            expect(output).toContain('return;');
+        });
+
+        it('generates code that still writes non-undefined values', () => {
+            const output = componentTemplate(linkDeclaration, [], PACKAGE);
+            // After the guard, the effect should still write when value is not undefined
+            expect(output).toContain('wcElement[inputName] = value;');
+        });
+    });
 });
--- a/libs/webc-generator/src/executors/generate/component-template.ts
+++ b/libs/webc-generator/src/executors/generate/component-template.ts
@@ -405,6 +405,11 @@ export function componentTemplate(
             // Read the signal value
             const value = this[signalName]();
             if (wcElement) {
+              // Skip writes when the signal is undefined and the property is already truthy
+              // This prevents clobbering imperatively-set values (e.g., from RouterLink)
+              if (value === undefined && wcElement[inputName]) {
+                return;
+              }
               // Write the value to the Web Component's property
               wcElement[inputName] = value;
             }


MariaIDineva (approved):
Looks good to me. I created a few examples and tested a few scenarios and it seems to be working fine, so let's 🚢 it.

InnaAtanasova (approved):

PR #14354, opened 2026-07-13
fix(core): product switch design updates

### LINKED ISSUE #14220: Product Switch component design updates for mobile mode, truncation, avatar integration, and accessibility
### Description
Following the latest design review and alignment with the design team, the Angular implementation of the Product Switch component needs to adopt the updated structure and responsive behavior introduced in Fundamental-styles.

### Tasks

- Adopt the updated markup structure related to the Avatar component
- Align Angular implementation with all related Fundamental-styles changes where applicable
- Implement responsive logic controlling the number of visible vertical items
- Update mobile mode responsiveness behavior
- Apply accessibility improvements and semantic updates

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14220

### Docs — aligned examples with fundamental-styles reference

- Added **Large** example (`fd-product-switch-body` with 14 items) showing the 4-column grid layout used on large desktop screens.
- Added **Medium** example (`fd-product-switch-body` with 5 items) showing the automatic 3-column layout used when fewer than 7 products are present.
- Removed the old "Less Than 7 Items" section (replaced by Medium).
- Replaced deprecated `ProductSwitchModule` with direct imports of `ProductSwitchComponent` and `ProductSwitchBodyComponent` in all three examples.

### Accessibility — implemented requirements from design spec

**`product-switch-body`**

- `<ul>` now carries `role="menu"`.
- Each `<li>` now carries `role="menuitem"`, `aria-posinset` (1-based index), `aria-setsize` (total item count), and a computed `aria-label` (`"Title Subtitle"`, with a navigation-target suffix appended when `target` is set).

**`product-switch` (trigger button)**

- Added `aria-haspopup="menu"` to the grid button so assistive technologies announce _"Products, Menu"_ on focus.

**Popover body**

- Added `aria-modal="false"` to the product-switch popover body — the popover is non-modal (focus is not trapped) and this attribute tells assistive technologies that background content remains accessible.
- Added `ariaModal` input to `PopoverBodyComponent` (backed by a `linkedSignal`) and wired it through `PopoverConfig` / `PopoverService` so other consumers can also control this attribute.

**`ProductSwitchItem` interface**

- Added optional `url` property — when set, the item renders an `<a>` element internally so the browser handles navigation natively (including `target` behaviour, middle-click, right-click → "Open in new tab", etc.). The `<a>` uses `display: contents` so it is invisible to the layout engine and adds no visual styling.
- Added optional `target` property (`'_self' | '_blank' | '_parent' | '_top'`) mirroring the HTML anchor target attribute. Used both for the `<a>` `target` attribute and to append the matching i18n suffix to the `aria-label`.

**Focus management**

- When the product switch popover opens, focus moves to the last focused item, or the first item on first open.
- When the popover closes (including via Escape), focus returns to the trigger button.
- The last focused item is tracked continuously via a focus event listener so the index is always accurate regardless of overlay teardown timing.

**Keyboard navigation**

- Enter/Space on a focused item with `url` now activates the inner `<a>` directly, so the browser handles navigation (including `target` window behaviour). Items without `url` are unchanged.

### Tests

- **`product-switch-body.component.spec.ts`** — expanded from 7 to 30 tests: rendering, ARIA attributes (`aria-posinset`, `aria-setsize`), URL anchor rendering, `isOpen` model, `itemClicked` / `productsChange` outputs, and grid/list keyboard navigation including circular wrap.
- **`product-switch.component.spec.ts`** — expanded from 8 to 25 tests: default input values, `disabled` host class binding, template structure, `booleanAttribute` coercion, and custom button projection.

### i18n

- Added three new keys to `coreProductSwitch` (`targetBlank`, `targetParent`, `targetTop`) with English values matching the design spec annotations (_"opens in a new browser tab"_, _"opens in the parent frame"_, _"opens in the full browser window"_).
- Updated `FdLanguageKeyIdentifier` union and all 38 translation files (`.ts` + `.properties`).


libs/core/popover/popover-body/popover-body.component.html
```diff
@@ -14,6 +14,7 @@
     [attr.role]="_bodyRole()"
     [attr.aria-label]="_bodyAriaLabel()"
     [attr.aria-labelledby]="_bodyAriaLabelledBy()"
+    [attr.aria-modal]="_bodyAriaModal() !== null ? _bodyAriaModal() : null"
```
InnaAtanasova: isn't this just `[attr.aria-modal]="_bodyAriaModal()"`?

libs/core/product-switch/product-switch/product-switch.component.ts
```diff
@@ -81,11 +85,28 @@ export class ProductSwitchComponent {
     readonly beforeOpen = output<void>();
 
     /** Event emitted when the product switch open state changes. */
-    readonly isOpenChange = output<boolean>();
+    readonly isOpenChange = new Subject<boolean>();
```
InnaAtanasova, line 87: We need to double check if this would actually be considered a breaking change. 
Claude's opinion:
`isOpenChange` converted from output<boolean>() to Subject<boolean> changes the public API. Template consumers using (isOpenChange)="handler($event)" will break because Angular won't recognize the Subject as an output.

libs/core/product-switch/product-switch-body/product-switch-body.component.ts
```diff
@@ -65,11 +96,25 @@ export class ProductSwitchBodyComponent implements OnInit, OnDestroy {
...
+    private _lastFocusedIndex = -1;
+
     /** @hidden */
     constructor(
         private _viewportRuler: ViewportRuler,
         private readonly _cdr: ChangeDetectorRef
-    ) {}
+    ) {
+        effect(() => this._syncFocusWithOpenState());
+        this._elementRef.nativeElement.addEventListener('focusin', (event: FocusEvent) =>
```
InnaAtanasova: This is never removed. It will leak if the component is destroyed while the listener is still attached. Maybe use `DestroyRef` or `afterNextRender() + effect()` scoped to the component lifecycle.

libs/core/product-switch/product-switch-body/product-switch-body.component.html
```diff
@@ -5,57 +5,78 @@
...
     >
-        @for (product of products; track product) {
+        @for (product of products; track product; let i = $index) {
             <li
                 fd-dnd-item
                 tabindex="0"
+                role="menuitem"
                 class="fd-product-switch__item"
                 [class.selected]="product.selected"
+                [attr.aria-label]="_itemAriaLabel()(product)"
```
InnaAtanasova, line 21: That's an interesting thing :)

libs/core/product-switch/product-switch-body/product-switch-body.component.ts
```diff
@@ -53,9 +59,34 @@ export class ProductSwitchBodyComponent implements OnInit, OnDestroy {
...
+        if (item.subtitle) {
+            parts.push(item.subtitle);
+        }
+        if (item.target === '_blank') {
+            parts.push(this._targetBlank());
+        } else if (item.target === '_parent') {
+            parts.push(this._targetParent());
+        } else if (item.target === '_top') {
+            parts.push(this._targetTop());
+        }
```
InnaAtanasova: Here we create three separate translation signal calls (_targetBlank, _targetParent, _targetTop) that are read on every aria-label evaluation. This means if there are 10 product items, we're calling this._targetBlank() 10 times per render cycle (once per item) even if no item has target="_blank". 

Can we use a single computed() that returns an object with all three, then destructure in `_itemAriaLabel`? Or something like that?

libs/core/product-switch/product-switch-body/product-switch-body.component.ts
MariaIDineva: Nice catch, thanks!

libs/core/product-switch/product-switch/product-switch.component.ts
MariaIDineva: You are right, that is a breaking change. These are the ways to handle this that Claude advised:

1. Remove the explicit `isOpenChange` output entirely and let consumers migrate to the model's built-in `isOpenChange` (breaking, but clean).
2. Rename one of them (e.g. `openStateChange`) to avoid the conflict — also breaking.
3. Keep the pre-existing conflict as-is and file a separate issue to resolve it properly in a future major version.

I actually thought that this was an error and that the app would not build unless the error is addressed but it seems like it's a warning and everything builds just fine, so we can go with the third approach and introduce the breaking change (we should decide which way to go, though) with the Angular update, what do you think?

libs/core/product-switch/product-switch-body/product-switch-body.component.html
MariaIDineva: Yeah, it is, that's because computed does not accept any arguments...

InnaAtanasova (dismissed):

InnaAtanasova (approved):

PR #14365, opened 2026-07-15
fix(platform): table header cell popover aria improvements

### LINKED ISSUE #14104: [Platform][fdp-table] ACC-264.1 (Level A) Screen reader announces more than once for the sort dialog inside of the table header
## Description

When opening the sort dialog (popover) from the table header in `fdp-table`, JAWS (Version 2026.2510.195) announces the menu item text multiple times. The expected behavior is that each menu item is announced only once.

Image
---

## Root Cause Analysis(Provided by AI)

The HTML structure rendered by `fdp-table`'s sort popover contains several accessibility issues that cause JAWS to announce the content repeatedly.

### 1. Duplicate `role="menu"` on nested elements

Both the `<fdp-list>` component and the inner rendered `<div fd-list>` carry `role="menu"`, causing JAWS to announce "menu" twice when entering the list.

```html
<fdp-list role="menu">                         <!-- role="menu" #1 -->
  <div fd-list="" role="menu" ...>             <!-- role="menu" #2 (duplicate) -->
    ...
  </div>
</fdp-list>
```

### 2. Triple-nested interactive roles on a single item

Each list item renders three nested elements that each carry an interactive ARIA role:

```html
<li role="menuitem">                           <!-- role #1 -->
  <div fd-list-item="" role="menuitem">        <!-- role #2 (duplicate) -->
    <div fd-list-link="" role="button">        <!-- role #3 (redundant) -->
      <span>Sort Ascending</span>
    </div>
  </div>
</li>
```

JAWS traverses each role-bearing ancestor when announcing a focusable element, causing the item text and "menu" context to be read once per role layer.

### 3. `role="button"` nested inside `role="menuitem"`

The innermost `<div fd-list-link="" role="button">` is a direct child of `role="menuitem"`. When JAWS focuses this button, it walks up the ancestor chain and re-announces the parent "menu" context, producing the repeated "Sort Ascending menu" output. A `menuitem` is already an activatable interactive role — wrapping its content in a `role="button"` is redundant and causes this re-announcement.

### 4. `cdk-focus-trap-anchor` with `tabindex="0"` and `aria-hidden="true"`

The CDK focus trap injects sentinel elements with `tabindex="0"` and `aria-hidden="true"`. JAWS sometimes ignores `aria-hidden` on focusable elements, which can cause focus to loop back into the popover and re-trigger the container role announcement.

```html
<div tabindex="0" class="cdk-visually-hidden cdk-focus-trap-anchor" aria-hidden="true"></div>
<!-- popover content -->
<div tabindex="0" class="cdk-visually-hidden cdk-focus-trap-anchor" aria-hidden="true"></div>
```

---

## Steps to Reproduce

1. Open a page with an `fdp-table` component that has sortable columns.
2. Enable JAWS (tested with Version 2026.2510.195).
3. Navigate to a table column header and activate the sort popover.
4. Listen to the screen reader output as focus enters the popover menu.

---

### Fix 1 — Remove `role="button"` from `fd-list-link`

`role="menuitem"` is already an activatable role. The inner link element should not additionally carry `role="button"`.

```html
<!-- Before -->
<div fd-list-link="" role="button" tabindex="-1" class="fd-list__link--focusable fd-list__link">

<!-- After -->
<div fd-list-link="" tabindex="-1" class="fd-list__link--focusable fd-list__link">
```

### Fix 2 — Ensure `role="menu"` is rendered only once

Remove the `role="menu"` from the `<fdp-list>` host element if the inner `<div fd-list>` already carries it, so that only one element in the tree declares this role.

### Fix 3 — Ensure `role="menuitem"` is declared only once per item

Either the `<li>` or the `<div fd-list-item="">` should carry `role="menuitem"`, not both.

---

## Accessibility Standard Reference

- WCAG 2.1 Success Criterion **4.1.2 Name, Role, Value** (Level A)
- WAI-ARIA Authoring Practices — [Menu and Menubar Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/)
  - A `menuitem` must not contain another interactive widget role such as `button`

---

## Environment

- **Component:** `fdp-table` (Platform Table), sort column header popover
- **Screen Reader:** JAWS Version 2026.2510.195

mikerodonnell89:
fixes #14104


InnaAtanasova (changes requested):
With this solution the element is now a list with list items:
Image

A `list` is a static container that groups related items. It simply tells assistive technology that several related items belong together.

`Menu` represents a collection of commands or actions, which is this case: we have actions group together [aria specs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menu_role). You need to modify the `fdp-list` to assign role menu to the proper html element and also have the option to to do the same for `fdp-free-content-list-item`. By removing the roles you are getting the default values. 

There's also a repetition of the role `listitem` for parent-child elements:

Image

Here's how the roles are assigned to the elements in UI5:

Image

libs/platform/list/free-content-list-item/free-content-list-item.component.ts
```diff
@@ -10,9 +11,19 @@ import { BaseListItem } from '../base-list-item';
...
+     */
+    @Input()
+    @HostBinding('attr.role')
+    role: Nullable<string>;
+
+    /**
+     * The ARIA role of the fd-list-item.
+     */
+    @Input()
+    listItemRole: Nullable<string>;
```
InnaAtanasova: we decided against Nullable. Should be `string | null` or `string | undefined` or `string | null | undefined` based on the case.

InnaAtanasova:
The SR is still not announcing properly the menu items. It announce each item as a group and doesn't say 
"Item X of X"

https://github.com/user-attachments/assets/f1a40993-63aa-45ce-bcab-2bd4888fb160

InnaAtanasova (approved):

PR #14374, opened 2026-07-16
fix(core): remove icon for default Micro process flow element

### LINKED ISSUE #14373: [Core] Micro Process Flow default item not showing the icon
Only for default `fd-micro-process-flow-item`

https://github.com/user-attachments/assets/57306f39-d88c-4c01-a47d-25f1c20ef7d9

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/14373

## Description
- render the icon only if glyph value is set
- update examples (remove glyph values for default steps)


MariaIDineva:
I don't fully understand what is desired here. Now the icon is not present at all and to me it looks a little bit broken.
Screenshot 2026-07-17 at 11 35 19

**Edit:** 
Like we discussed during the daily, that is the way it should be when there is no status, so all good.

MariaIDineva (approved):

PR #14381, opened 2026-07-17
docs(core): align dayjs date-picker i18n example with adapter model

InnaAtanasova:
## Related Issue(s)

closes none, defect hunting before release

## Description
This fixes the core date-picker i18n docs example after the datetime adapter changes.

The example was providing `DayjsDatetimeAdapter` but initializing the bound model with `FdDate.getNow()`. That created a model/adapter mismatch. After the recent adapter changes, that mismatch now fails during date-picker initialization with:

`DayjsDatetimeAdapter: Cannot format invalid date.`

### Root Cause

The Dayjs adapter expects Dayjs model values.

The i18n example was still binding an `FdDate` value while also overriding `DatetimeAdapter` with `DayjsDatetimeAdapter`. In the date-picker `writeValue()` flow, the incoming model is parsed and then formatted immediately. With the custom Dayjs parse format used by the example, the `FdDate` seed value no longer round-trips into a valid Dayjs instance, so formatting throws.

### Changes

- Updated the core date-picker i18n example to use `dayjs()` / `Dayjs` instead of `FdDate.getNow()`
- Updated the injected adapter generic type to `DatetimeAdapter<Dayjs>`
- Added a documentation note to the Dayjs adapter docs clarifying that `DayjsDatetimeAdapter` must be paired with Dayjs model values
- Added the same note to the date-picker i18n docs section

### Impact

This is not a general breaking change for all date-picker i18n users.

Consumers only need to update if they were mixing:
- `DayjsDatetimeAdapter` with `FdDate` model values

Consumers do not need changes if they already use:
- `DayjsDatetimeAdapter` with Dayjs values
- `FdDatetimeAdapter` with `FdDate` values

### Before
Screenshot 2026-07-17 at 2 36 49 PM

### After
Screenshot 2026-07-17 at 3 03 34 PM

### DIFF
--- a/libs/docs/core/date-picker/examples/date-picker-i18n-example.component.ts
+++ b/libs/docs/core/date-picker/examples/date-picker-i18n-example.component.ts
@@ -2,11 +2,12 @@ import { Component, LOCALE_ID } from '@angular/core';
 import { FormsModule } from '@angular/forms';
 import { ButtonComponent } from '@fundamental-ngx/core/button';
 import { DatePickerComponent } from '@fundamental-ngx/core/date-picker';
-import { DATE_TIME_FORMATS, DatetimeAdapter, FdDate, FdDatetimeModule } from '@fundamental-ngx/core/datetime';
+import { DATE_TIME_FORMATS, DatetimeAdapter, FdDatetimeModule } from '@fundamental-ngx/core/datetime';
 import { FormLabelComponent } from '@fundamental-ngx/core/form';
 import { SegmentedButtonModule } from '@fundamental-ngx/core/segmented-button';
 import { DAYJS_DATETIME_FORMATS, DayjsDatetimeAdapter } from '@fundamental-ngx/datetime-adapter';
 import { patchLanguage } from '@fundamental-ngx/i18n';
+import dayjs, { Dayjs } from 'dayjs';
 import 'dayjs/locale/bg';
 import 'dayjs/locale/de';
 import 'dayjs/locale/fr';
@@ -77,10 +78,10 @@ const CUSTOM_DATETIME_FORMATS = {
     ]
 })
 export class DatePickerI18nExampleComponent {
-    date = FdDate.getNow();
+    date: Dayjs = dayjs();
     locale = 'fr';
 
-    constructor(private datetimeAdapter: DatetimeAdapter<FdDate>) {}
+    constructor(private datetimeAdapter: DatetimeAdapter<Dayjs>) {}
 
     setLocale(locale: string): void {
         this.locale = locale;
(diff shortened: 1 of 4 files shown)


libs/docs/core/date-picker/examples/date-picker-i18n-example.component.ts
```diff
@@ -77,10 +78,10 @@ const CUSTOM_DATETIME_FORMATS = {
     ]
 })
 export class DatePickerI18nExampleComponent {
-    date = FdDate.getNow();
+    date: Dayjs = dayjs();
```
MariaIDineva, line 74: After this change we can now remove `FdDatetimeModule` from the imports. Per Claude:

>  After the PR it is dead code on two counts: (1) its FdDatetimeAdapter provider (from FdDatetimeAdapterModule) is overridden by the component's own DayjsDatetimeAdapter provider, and (2) none of its pipe exports (DateFormatPipe, DateTimeFormatPipe, etc.) are used in the template. Additionally, FdDatetimeModule is marked @deprecated in the source. It should be removed from both the imports array and the import statement (the FdDatetimeModule symbol on line 5 of the diff)

MariaIDineva (dismissed):
Also, I think this console error has nothing to do with the Angular 22 update as it does not touch anything related to dates. I can see it in version 0.63 as well, where we are on Angular 21. The issue was rather introduced in our [datetime adapters migration](https://github.com/SAP/fundamental-ngx/pull/14016). Per Claude:

> The real cause is commit 6be78b57 — the DayjsDatetimeAdapter migration that made _createDayjsDate() and _prepareFormat() explicitly private and fixed parsing bugs. That commit tightened the adapter's parse/format behavior such that an FdDate value can no longer silently round-trip through the Dayjs adapter. The example was already broken-by-design before Angular 22; the stricter adapter just surfaced it.

Not that it matters much, but I was curious to investigate because I really did not touch anything related to the adapter as part of Angular 22 migration.

InnaAtanasova:
> Also, I think this console error has nothing to do with the Angular 22 update as it does not touch anything related to dates. I can see it in version 0.63 as well, where we are on Angular 21. The issue was rather introduced in our [datetime adapters migration](https://github.com/SAP/fundamental-ngx/pull/14016). Per Claude:
> 
> > The real cause is commit [6be78b5](https://github.com/SAP/fundamental-ngx/commit/6be78b57ac60b16b893059856ce6d79902ffce4c) — the DayjsDatetimeAdapter migration that made _createDayjsDate() and _prepareFormat() explicitly private and fixed parsing bugs. That commit tightened the adapter's parse/format behavior such that an FdDate value can no longer silently round-trip through the Dayjs adapter. The example was already broken-by-design before Angular 22; the stricter adapter just surfaced it.
> 
> Not that it matters much, but I was curious to investigate because I really did not touch anything related to the adapter as part of Angular 22 migration.

that's true

libs/docs/core/date-picker/examples/date-picker-i18n-example.component.ts
InnaAtanasova: removed `FdDatetimeModule`

MariaIDineva (approved):

Issue #14364, opened 2026-07-15
Settings Generator (and similar) doc examples don't render until an unrelated event fires


MariaIDineva:
## Summary

Reported in PR [#14332](https://github.com/SAP/fundamental-ngx/pull/14332#issuecomment-4961757973)
(chore: upgrade to Angular 22): the Platform Settings Generator **default** doc
example rendered blank on load. Clicking the "Save" button made it render, but
logged a console error (`ExpressionChangedAfterItHasBeenCheckedError` /
`NG0100`-class warning).

## Root cause

`SettingsGeneratorDefaultExampleComponent` built a data structure (`schema`) —
consumed by a child `@Input()` (`fdp-settings-generator`'s `settings`) — as a
**plain property**, assigned inside `ngAfterViewInit()`, using `@ViewChild()`
decorator results (`privacyContent`, `termsOfServiceContent`) that are only
available at that point in the lifecycle.

```ts
@ViewChild('privacyContent') privacyContent: TemplateRef<any>;

schema: SettingsModel;

ngAfterViewInit(): void {
    // ... build using this.privacyContent, this._theming, etc.
    this.schema = { ... };
    // no markForCheck()/detectChanges() call
}
```

This relies on an **implicit follow-up change-detection tick** happening after
`ngAfterViewInit()` mutates `schema`, so that the parent's template binding
(`[settings]="schema"`) is re-evaluated and pushed into the child's `@Input()`
setter. In Angular 21 (zone.js) that follow-up tick reliably happened (timers,
zone stabilization, etc.). In Angular 22, that implicit tick is no longer
guaranteed for this pattern — so the child's input setter never re-runs with
the real data, and whatever internal logic depends on it (e.g. dynamically
creating a layout component) never executes. The page only "renders" once an
unrelated Angular-bound DOM event (the Save button click) forces a fresh
change-detection pass.

**Confirmed by comparison:** three sibling examples
(`message-popover`, `custom-control`, `dialog`, `custom-layout`) already guard
against this by calling `this._cdr.detectChanges()` at the end of
`ngAfterViewInit()` and using `ChangeDetectionStrategy.OnPush` explicitly.
The `default` example was the only one missing both — which is exactly why it
was the only one broken.

## Fix applied

Rather than patch it with `ChangeDetectorRef.detectChanges()` (works, but
imperative/legacy), refactored `SettingsGeneratorDefaultExampleComponent` to
the Angular 21+ idiomatic reactive pattern:

[...]

PR #14386, opened 2026-07-20
fix(cdk): filter disabled tabbables and stabilize tabbable docs example

InnaAtanasova:
## Related Issue(s)

related to https://github.com/SAP/fundamental-ngx/issues/14364

## Description
The docs example under CDK Tabbable started showing an empty result instead of the expected enabled button class.
### Root Cause
1. Disabled states were not fully filtered in tabbable detection, so elements marked as non-interactive via accessibility/state classes could still be selected.
2. The docs example update flow was not consistently reflected in the rendered output in the current runtime setup.

### Changes
1. Updated tabbable detection logic in [tabbable-element.service.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html):

- Excludes elements that are disabled by:
- disabled attribute
- aria-disabled="true"
- is-disabled class
- Keeps first-match traversal semantics while respecting skipSelf and focusLastElement behavior.

2. Updated docs example in [tabbable-default-example.component.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html):

- Uses signal-backed state for the displayed class value.
- Runs detection in a stable two-pass flow (initial + microtask) to avoid stale/empty display.

3. Updated template binding in [tabbable-default-example.component.html](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html):

- Reads signal value directly in template.

### Before
Screenshot 2026-07-20 at 4 15 29 PM

### After
Screenshot 2026-07-20 at 4 15 39 PM

### DIFF
--- a/libs/cdk/utils/services/tabbable-element.service.ts
+++ b/libs/cdk/utils/services/tabbable-element.service.ts
@@ -39,11 +39,7 @@ export class TabbableElementService {
     }
 
     private _isTabbableElement(element: HTMLElement): boolean {
-        return (
-            this._checker.isTabbable(element) &&
-            !this._isElementDisabled(element) &&
-            this._checker.isFocusable(element, { ignoreVisibility: true })
-        );
+        return this._checker.isTabbable(element) && !this._isElementDisabled(element);
     }
 
     private _isElementDisabled(element: HTMLElement): boolean {
--- a/libs/cdk/utils/services/tabbable-element.service.ts
+++ b/libs/cdk/utils/services/tabbable-element.service.ts
@@ -12,9 +12,11 @@ export class TabbableElementService {
 
     /** Get the first tabbable element from a DOM subtree (inclusive). */
     getTabbableElement(root: HTMLElement, focusLastElement = false, skipSelf = false): HTMLElement | null {
-        if (!skipSelf && this._checker.isFocusable(root) && this._checker.isTabbable(root)) {
+        if (!skipSelf && this._isTabbableElement(root)) {
             return root;
         }
+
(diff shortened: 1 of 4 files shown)


libs/cdk/utils/services/tabbable-element.service.ts
```diff
@@ -35,4 +37,20 @@ export class TabbableElementService {
 
         return null;
     }
+
+    private _isTabbableElement(element: HTMLElement): boolean {
+        return (
+            this._checker.isTabbable(element) &&
+            !this._isElementDisabled(element) &&
+            this._checker.isFocusable(element, { ignoreVisibility: true })
```
MariaIDineva: Claude:

> _isTabbableElement ends with this._checker.isFocusable(element, { ignoreVisibility: true }). Angular CDK's isTabbable already calls isFocusable internally — isTabbable() returning true guarantees isFocusable() would also return true. The third condition is therefore always true when reached, making it dead code. Worse, passing { ignoreVisibility: true } is a silent behavioral change from the original: the original isFocusable(root) respected visibility, and now the explicit isFocusable check ignores it. In practice the impact is negligible (invisible elements won't pass isTabbable anyway), but the intention is unclear and the extra CDK call adds overhead. Remove the third condition — isTabbable already covers focusability.

InnaAtanasova: fixed

MariaIDineva (approved):

Issue #14363, opened 2026-07-14
User Menu + Notifications design review findings


InnaAtanasova:
## User Menu 
- User Menu Enter key not working
- Avatar+Edit icon not getting focus
- Mobile focus order

## Notifications:
- Acc issue for overflow button not closing on 
- update text to Important
- Show more should be More
- Remove margin below message strip
- remove the numbers in the title [Notifications(30)]
- Tab and Arrow Keys should navigate through Notifications
- Label for close button should be per ACC specs

PR #14391, opened 2026-07-21
fix(core): add User menu a11y improvements

InnaAtanasova:
## Related Issue(s)
part of https://github.com/SAP/fundamental-ngx/issues/14363

## Description
Fixes four keyboard accessibility issues in the user menu component:

### 1. Enter/Space keys not opening mobile dialog
The user menu control now responds to Enter and Space keys in mobile mode, matching standard button behavior.

**Test:** Tab to the avatar control in mobile mode, press Enter or Space → dialog opens.

### 2. Avatar focus in documentation
Fixed examples by adding `[clickable]="true"` to make avatars keyboard-accessible.

### 3. Initial focus on menu open
When the popover/dialog opens, focus now moves directly to the first `fd-user-menu-list-item` instead of other interactive elements in the header (like edit buttons or close button).

Screenshot 2026-07-21 at 5 23 09 PM
Screenshot 2026-07-21 at 5 22 56 PM

### 4. Focus reset when re-entering the menu (discovered bug)
**Problem:** The menu uses roving tabindex to preserve focus position (important for submenus). When a user navigated with arrow keys to item 3, then tabbed out, the browser would return focus to item 3 when tabbing back in instead of starting at item 1.

**Fix:** Added `focusout` listener that resets tabindex to the first item when focus leaves the menu entirely, while preserving roving tabindex behavior within the menu.

Before:

https://github.com/user-attachments/assets/e5a1ec75-dbd4-4c79-9856-0aeb175b18ae

After:

https://github.com/user-attachments/assets/6e96a152-aa49-46a3-a601-fb3b034a17da

### DIFF
--- a/libs/core/user-menu/components/user-menu-control.component.ts
+++ b/libs/core/user-menu/components/user-menu-control.component.ts
@@ -13,7 +13,7 @@ import { ChangeDetectionStrategy, Component, ElementRef, EventEmitter, Output, V
     imports: []
 })
 export class UserMenuControlComponent {
-    /** Event emitted event when control element is clicked */
+    /** Event emitted when control element is clicked */
     @Output()
     clicked: EventEmitter<void> = new EventEmitter<void>();
 
@@ -25,10 +25,16 @@ export class UserMenuControlComponent {
         this.clicked.emit();
     }
 
-    /** @hidden */
+    /**
+     * Handles keyboard activation (Enter or Space).
+     * Prevents default to avoid scrolling on Space and stops propagation
+     * to prevent unintended bubbling to parent elements.
+     * @hidden
+     */
     onKeydown(event: KeyboardEvent): void {
         if (event.key === 'Enter' || event.key === ' ') {
             event.preventDefault();
+            event.stopPropagation();
             this.clicked.emit();
         }
     }
--- a/libs/core/user-menu/components/user-menu-control.component.ts
+++ b/libs/core/user-menu/components/user-menu-control.component.ts
@@ -33,6 +33,15 @@ export class UserMenuControlComponent {
     }
 
     /** @hidden */
+    @HostListener('keydown', ['$event'])
+    onKeydown(event: KeyboardEvent): void {
+        if (event.key === 'Enter' || event.key === ' ') {
+            event.preventDefault();
+            this.clicked.emit();
+        }
+    }
+
+    /** @hidden */
     focus(): void {
         this.el.nativeElement.focus();
     }
--- a/libs/core/user-menu/components/user-menu-control.component.ts
+++ b/libs/core/user-menu/components/user-menu-control.component.ts
@@ -27,14 +27,14 @@ export class UserMenuControlComponent {
 
     /**
      * Handles keyboard activation (Enter or Space).
-     * Prevents default to avoid scrolling on Space and stops propagation
-     * to prevent unintended bubbling to parent elements.
+     * Prevents default to avoid scrolling on Space.
+     * Emits clicked event for mobile mode; in popover mode, the event
+     * bubbles to the popover component which handles the toggle.
      * @hidden
      */
     onKeydown(event: KeyboardEvent): void {
         if (event.key === 'Enter' || event.key === ' ') {
             event.preventDefault();
-            event.stopPropagation();
             this.clicked.emit();
         }
     }
--- a/libs/core/user-menu/components/user-menu-control.component.ts
+++ b/libs/core/user-menu/components/user-menu-control.component.ts
@@ -1,12 +1,4 @@
-import {
-    ChangeDetectionStrategy,
-    Component,
-    ElementRef,
-    EventEmitter,
-    HostListener,
-    Output,
-    ViewEncapsulation
-} from '@angular/core';
+import { ChangeDetectionStrategy, Component, ElementRef, EventEmitter, Output, ViewEncapsulation } from '@angular/core';
 
 @Component({
     selector: 'fd-user-menu-control',
@@ -14,7 +6,9 @@ import {
     encapsulation: ViewEncapsulation.None,
     changeDetection: ChangeDetectionStrategy.OnPush,
     host: {
-        tabindex: '0'
+        tabindex: '0',
+        '(click)': 'onClick()',
+        '(keydown)': 'onKeydown($event)'
     },
     imports: []
 })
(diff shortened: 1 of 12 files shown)


libs/core/user-menu/components/user-menu-control.component.ts
```diff
@@ -1,37 +1,44 @@
...
+    /**
+     * Handles keyboard activation (Enter or Space).
+     * Prevents default to avoid scrolling on Space and stops propagation
+     * to prevent unintended bubbling to parent elements.
+     * @hidden
+     */
+    onKeydown(event: KeyboardEvent): void {
+        if (event.key === 'Enter' || event.key === ' ') {
+            event.preventDefault();
+            event.stopPropagation();
```
MariaIDineva: It seems like after adding this change the menu in default mode stopped opening on enter/space.

MariaIDineva (commented):
There's another bug that I found while reviewing the PR. This has nothing to do with the current changes as it is reproducible on main as well, so maybe we can deal with it in a separate PR. When the user navigates through the menu using the tab key and opens one or more submenus, if they click outside, the submenus stay open:

https://github.com/user-attachments/assets/92068763-fd2f-4c03-a41d-fe995dd80c85

libs/core/user-menu/user-menu.component.ts
```diff
@@ -216,16 +203,7 @@ export class UserMenuComponent implements AfterViewInit {
...
-        if (this._listItems().length > 0) {
-            this._listItems().forEach((item) => {
-                item.isOpen.set(false);
-                item._elementRef?.nativeElement.querySelector('.fd-menu__link')?.classList.remove('is-active');
-            });
-
-            this._resetListFocus();
-        }
-
+        this._closeAllSubmenus();
```
MariaIDineva: Do we need to have all the logic under `_closeAllNestedSubmenus` in `libs/core/user-menu/components/user-menu-list-item.component.ts`  when we have the logic in `_closeAllSubmenus` here?

libs/core/user-menu/components/user-menu-list-item.component.ts
```diff
@@ -220,13 +233,24 @@ export class UserMenuListItemComponent implements KeyboardSupportItemInterface {
 
     /** @hidden Open submenu on host mouseenter (non-mobile only) */
     onHostMouseEnter(): void {
-        if (!this.hasSubmenu() || this.mobile()) {
+        if (this.mobile()) {
```
MariaIDineva, line 236: Claude:

> [libs/core/user-menu/components/user-menu-list-item.component.ts ~233] onHostMouseEnter() calls this.focus() unconditionally on every hover
> 
> The original guard was if (!this.hasSubmenu() || this.mobile()) return. The PR drops the !this.hasSubmenu() check, so this.focus() and this._closeSiblingSubmenus() now run for all items on every mouseenter, not just items with submenus. Programmatically calling focus() on mouseenter is an a11y anti-pattern: screen readers announce focus changes, so hovering with a mouse causes noisy, unexpected announcements for AT users. WCAG 2.4.3 / 3.2.1 specifically prohibit moving focus as a side-effect of hover. This is an accessibility regression in a PR specifically targeting accessibility. The roving-tabindex _tabIndex$ signal can be updated without calling focus() — update the signal only, or restrict the focus() call to hasSubmenu() items.

In this particular case I am not sure if Claude is 100% right, but I thought it was worth flagging it because of this part: `Programmatically calling focus() on mouseenter is an a11y anti-pattern`, so maybe it is worth checking.

libs/core/user-menu/user-menu.component.ts
MariaIDineva: Also, `isOpenChangeHandle` calls `_closeAllSubmenus`, so if we call `_closeAllSubmenus` here and then in `isOpenChangeHandle` wouldn't that be a duplicate call? Maybe we don't need this one here.

libs/core/user-menu/user-menu.component.spec.ts
```diff
@@ -40,4 +66,92 @@ describe('UserMenuComponent', () => {
...
+        it('should only subscribe to control clicks when in mobile mode', fakeAsync(() => {
+            // Set mobile on the actual user menu component
+            userMenu.mobile = jest.fn().mockReturnValue(true) as any;
+            fixture.detectChanges();
+
+            // Trigger ngAfterViewInit
+            tick();
+
+            const control = userMenu['userMenuControl']();
+            expect(control).toBeDefined();
```
MariaIDineva: I don't understand this expectation. It does not seem to test what the it block claims.

libs/core/user-menu/user-menu.component.spec.ts
```diff
@@ -40,4 +66,92 @@ describe('UserMenuComponent', () => {
...
+            userMenu.open();
+            fixture.detectChanges();
+
+            // Wait for requestAnimationFrame
+            tick(20);
+
+            expect(firstItemFocusSpy).toHaveBeenCalled();
+        }));
+
+        it('should not focus first item when opening in mobile mode', fakeAsync(() => {
```
MariaIDineva: > should **not** focus first item

 Really, is that the requirement? Actually, when I test both in desktop and in mobile mode we do have the first list item focused? 🤔

libs/core/user-menu/user-menu.component.ts
InnaAtanasova: Re: the first comment 
`_closeAllSubmenus()` (parent component) closes all menu items when the main menu closes, accessing items via contentChildren
`_closeAllNestedSubmenus()` (list item component) closes only the nested descendants of a specific submenu item, accessing them via DOM queries
The parent component handles top-level closure, while each list item handles its own recursive cleanup. (There can be multiple nested submenus.)
Screenshot 2026-07-23 at 12 43 04 PM

Re: the second comment - you're right! In non-mobile mode, we were calling `_closeAllSubmenus()` twice. I've updated it to:
```
close(): void {
    this.isOpenChangeHandle(false);
    if (this.mobile()) {
        this._closeAllSubmenus();
    }
    this._clearSubmenu();
    this._dialogRef?.close();
}
```

libs/core/user-menu/components/user-menu-list-item.component.ts
InnaAtanasova: Claude is right about the a11y violation, but that's the design requirement for this component. It's the same behavior (hover and focus the list items on hover). It's like this in ui5 web components as well: https://ui5.github.io/webcomponents/components/fiori/UserMenu/

libs/core/user-menu/user-menu.component.spec.ts
InnaAtanasova: Good catch! This test only verified that the control exists, not really something related to subscription. I've removed it since it wasn't adding value.

libs/core/user-menu/user-menu.component.spec.ts
InnaAtanasova: It is a very confusing description, I agree. I updated it. 
The design requires to focus the first `list` item, but in Popover and Dialog we focus by default the first focusable item. So this test is checking if we overwrite this behavior.

MariaIDineva (dismissed):

droshev (approved):

PR #14392, opened 2026-07-22
chore(i18n): consolidate translation pipeline into i18n-manage sync

### LINKED ISSUE #14350: i18n - we should only update translations.properties
Some of the translation team has notified me that our updating of every *.properties file when we add new translations to the library is breaking something on their end, so we need to modify our library and build task such that we (fundamental-ngx team) need to ONLY update the translations.properties file, and don't touch the other *.properties files. We can have the task check for the strings it needs in the existing .properties file for the language it's translating, and if it's not there, just use whatever is in translations.properties and build that to the *.ts file. For example:

1. We need to add "Hello, world!" to the core shellbar in the library.
2. We add `coreShellbar.helloWorld = Hello, world!` to translations.properties
3. The build task is building translations_es.ts. It looks at translations_es.properties, but `coreShellbar.helloWorld`isn't there. So it just looks at translations.properties and grabs it from there. `coreShellbar.helloWorld = Hello, world!` is added to translations_es.ts.
4. Later, the translation delivery service gives us translations_es.properties, with `coreShellbar.helloWorld = ¡Hola Mundo!`. Now this will be built into translations_es.ts in all future builds.

mikerodonnell89:
fixes #14350 

The translation team had reached out to us and asked that we stop touching the translated properties files - they only want to see us modify the default (English) `translation.properties` file. So this PR modifies our translation pipeline such that developers now do not edit any `.properties` or `.ts` files themselves - instead, they use `i18n-manage` to make changes - and these changes are only propagated to `translations.properties`, `translations.ts` and the `translations_*.ts` files for each language, without touching any of the `translations_*.properties` files. The translation team will deliver us the `translations_*.properties` files, and the `i18n-manage --command=sync` will build these into their corresponding `translations_*.ts` files.

This PR also takes the logic from `transform-translations` and brings it into the `i18n-manage` ecosystem, to bring all the translation scripting under one umbrella. `i18n-manage --command=sync` is effectively the new `transform-translations`


droshev:
About the type generation: Right now FdLanguageKeyIdentifier is built by parsing the FdLanguage interface in fd-language.ts with a hand-written brace-counting parser (with a special case for FdLanguageKey<{...}>). That's the part I'd worry about maintaining - it'll probably trip up the next time someone formats the interface a bit differently. Before this PR the key list came from a .properties file, i.e. reading plain data instead of parsing code. Could we point it back at translations.properties? Simpler, and nothing to special-case.

Second, and related: `add` writes the key to translations.properties and the .ts files, but not to fd-language.ts - which is where the types are generated from. So right after add, the type union is missing the key until someone edits the interface by hand. (The README even still lists that as a manual step, which reads oddly next to "developers don't edit .properties or .ts files.") Good news: this gap just disappears if we do the first thing - generate types from translations.properties and there's only one source to keep in sync.

mikerodonnell89:
> About the type generation: Right now FdLanguageKeyIdentifier is built by parsing the FdLanguage interface in fd-language.ts with a hand-written brace-counting parser (with a special case for FdLanguageKey<{...}>). That's the part I'd worry about maintaining - it'll probably trip up the next time someone formats the interface a bit differently. Before this PR the key list came from a .properties file, i.e. reading plain data instead of parsing code. Could we point it back at translations.properties? Simpler, and nothing to special-case.

> Second, and related: `add` writes the key to translations.properties and the .ts files, but not to fd-language.ts - which is where the types are generated from. So right after add, the type union is missing the key until someone edits the interface by hand. (The README even still lists that as a manual step, which reads oddly next to "developers don't edit .properties or .ts files.") Good news: this gap just disappears if we do the first thing - generate types from translations.properties and there's only one source to keep in sync.

I certainly can point the generation of the FdLanguageKeyIdentifier back to translations.properties, but I thought this new setup would actually be advantageous - it helps the dictionary file actually serve more of a purpose. Enforcing that the dictionary file and the translations.properties files each have the same keys will prevent a side effect we had before this PR, where the properties files had 80+ keys that weren't present in fd-language.ts. At that point what was even the purpose of fd-language.ts? This makes it so the developer is keenly aware of any stragglers in their properties files, and they're more deliberate when adding translations or modifying existing ones

InnaAtanasova:
Since this PR removes the `transform-translations` executor (from `executors.json` and `nx.json`), should we document this as a breaking change for any external workflows or scripts that invoke nx run i18n:transform-translations?

InnaAtanasova:
I think you should shorten the PR title, we have a limit of 72 characters and it's currently 149.
Maybe something like `chore(i18n): consolidate translation pipeline into i18n-manage sync`.

InnaAtanasova (commented):
1. `libs/nx-plugin/src/executors/i18n-manage/utils/comment-generator.spec.ts` Missing tests for `extractJSDocComment` function (new feature added in this PR)

2. `libs/nx-plugin/src/executors/i18n-manage/operations/sync.ts` Add explicit validation that `translations.properties` exists before attempting to read it (currently fails with generic error).

3. `.github/workflows/i18n-auto-generate.yml` Consider adding syntax validation for properties files to prevent code injection via fork PRs

4. All executor files -  Inconsistent console messages - some say "Modified X .properties files" vs "Modified X TypeScript translation files". Use consistent terminology throughout.

5. `.claude/skills/i18n-manage/SKILL.md` Workflow order unclear - doesn't explain when to update `fd-language.ts` relative to running CLI commands. Should clarify: update `fd-language.ts` first, then run i18n-manage --command=add.

6. Missing migration guide at `docs/migrations/i18n-transform-to-sync.md` with before/after examples for:

> - `project.json` target configuration
> - CI/CD script updates
> - Rollback instructions

mikerodonnell89:
> 4. All executor files -  Inconsistent console messages - some say "Modified X .properties files" vs "Modified X TypeScript translation files". Use consistent terminology throughout.

Just went through the console logging and comments in the scripts and I think they are all actually accurate. Perhaps you had an nx cache thing going on? Working on addressing your other points now

> 5. `.claude/skills/i18n-manage/SKILL.md` Workflow order unclear - doesn't explain when to update `fd-language.ts` relative to running CLI commands. Should clarify: update `fd-language.ts` first, then run i18n-manage --command=add.

This actually doesn't matter. Either can be done first, they just both need to be done before attempting to run the app

6. Missing migration guide at `docs/migrations/i18n-transform-to-sync.md` with before/after examples for:

> - `project.json` target configuration
> - CI/CD script updates
> - Rollback instructions

I think the changes have been thoroughly explained in the existing readmes and no additional "migration guide" is necessary

droshev:
> Since this PR removes the `transform-translations` executor (from `executors.json` and `nx.json`), should we document this as a breaking change for any external workflows or scripts that invoke nx run i18n:transform-translations?

Actually, i would disagree here: transform-translations was never part of the library consumer API - it was an internal repo maintenance tool. They don't install @fundamental-ngx/nx-plugin and they certainly don't configure an NX executor in their own project. That command only ever ran inside this monorepo to regenerate .ts files. Ideally looking this could be a breaking change but only for those running nx run i18n:transform-translations in their own workflow - the question is who are those?

mikerodonnell89:
> > Since this PR removes the `transform-translations` executor (from `executors.json` and `nx.json`), should we document this as a breaking change for any external workflows or scripts that invoke nx run i18n:transform-translations?
> 
> Actually, i would disagree here: transform-translations was never part of the library consumer API - it was an internal repo maintenance tool. They don't install @fundamental-ngx/nx-plugin and they certainly don't configure an NX executor in their own project. That command only ever ran inside this monorepo to regenerate .ts files. Ideally looking this could be a breaking change but only for those running nx run i18n:transform-translations in their own workflow - the question is who are those?

Agreed, more of an internal breaking change. I'll remove the note

droshev (dismissed):

libs/i18n/src/lib/models/fd-language-key-identifier.ts
```diff
@@ -3,580 +3,580 @@
...
-    | 'coreCalendar.calendarLegendLabel'
-    | 'coreCalendar.calendarRoleDescription'
-    | 'coreCalendar.calendarMonthViewDescription'
-    | 'coreCalendar.calendarYearsRangeViewDescription'
-    | 'coreCalendar.calendarYearsViewDescription'
-    | 'coreCalendar.closeCalendarLabel'
-    | 'coreCalendar.dateSelectedLabel'
-    | 'coreCalendar.dateSelectionLabel'
-    | 'coreCalendar.dayInPastLabel'
+    | 'coreActionSheet.toolbarLabel'
```
MariaIDineva: Why all these changes? It seems like the keys are now not ordered alphabetically.

MariaIDineva (commented):
Great idea to add security validation for `.properties` files. 👏 

And here are just a couple comments from the review I did with Claude that I think are worth addressing, although not blocking:

**[`add-key.ts`, `remove-key.ts`, `rename-key.ts`, `update-key.ts`] Significant code duplication in TS generation loop**

All four operations contain an identical ~30-line `for (const tsFile of tsFiles)` loop that merges properties, runs Prettier, and writes `.ts` files. This is the same logic as `sync.ts`. Each operation should modify `translations.properties` then call `sync({ propertiesPath })` directly, instead of duplicating the generation logic inline. As written, a bug fix or behavior change to the generation would need to be applied in five places.

**[`validate.ts:5530-5569`] Comment validation was silently removed**

The old validate operation checked that every key in `translations.properties` had a valid SAP-format comment header (`#XBUT:`, `#XTIT:`, `#XFLD:`, etc.). This validation no longer runs — tests that previously asserted `success: false` for missing or invalid comment types now assert `success: true`. The quality gate is gone with no mention in the PR description. If this is intentional (the comment format is now an authoring-time concern, not a validation-time one), it should be called out explicitly. Otherwise it may be an unintended regression.

**[`shared/update-typings.ts:7278-7281`] Dead code: `updateTypings(tree, keys)` is never called**

The NX-Tree-based `updateTypings` export (lines ~275-281 in the new file) was carried over from `transform-translations/update-typings.ts`, but nothing in the new codebase calls it. The filesystem-based `updateFdLanguageKeyIdentifier` is used everywhere. The Tree overload should be removed, or a comment should explain what it's kept for.

MariaIDineva (dismissed):

MariaIDineva (approved):

InnaAtanasova (approved):

PR #14395, opened 2026-07-23
fix: attempt number 3 to fix ng update

robertIsaac:
it seems the previous attempt did stop the error, but didn't actually update all deps
now hopefully it will be fully fixed


InnaAtanasova (approved):

robertIsaac:
I tested it now with 0.64.1 and it seems to work now :)

PR #14396, opened 2026-07-23
fix(ui5): force-construct per-package theme providers to eliminate "Theme not registered" errors

### LINKED ISSUE #14394: when mix ui5 and core components together, the dark theme is broken
### Describe your proposal

we can't mix between ui5 and core components, even in different page

angular 22 fundamental 0.64

### Affected Component

all ui5 components

1. clone https://github.com/robertIsaac/fundamental-theme-ui5-reprod
2. run `npm i` and `npm start`
3. go to http://localhost:4200/?theme=sap_horizon_dark (it's okay)
4. go to http://localhost:4200/buggy?theme=sap_horizon_dark (it's not okay)
5. go back to okay page (it's still broken)

### StackBlitz example(s)

_No response_

### Priority

High

### Declaration

droshev:
closes #14394

When mixing @fundamental-ngx/ui5-webcomponents and @fundamental-ngx/core components in the same app, two related bugs caused non-default themes (`sap_horizon_dark`, `sap_horizon_hcb`, `sap_horizon_hcw`) to not apply correctly.

---

**Bug 1** - `provideUi5ThemingBridge()` never constructing the base token provider

**Symptom**: UI5 Web Components always rendered in light `sap_horizon` regardless of the active theme. The `--sap*` CSS custom properties on `:root` never switched.

**Root cause**: `provideUi5ThemingBridge()` wired the Fundamental→UI5 theme bridge to the registry (`Ui5ThemingService`) but never constructed `Ui5WebcomponentsThemingService` (the base token provider). The registry's `effect()` early-returns when `_providers` is empty, so `setTheme()` was never forwarded to the UI5 runtime.

**Fix**: `provideUi5ThemingBridge()` now force-constructs `Ui5WebcomponentsThemingService` via the same `provideEnvironmentInitializer`, so the provider self-registers and the full chain (`ThemingService` → `Ui5ThemingService` → `effect` → `provider` → `UI5 setTheme()`) works with a single consumer call.

---
**Bug 2** - per-component shadow DOM styles never registered

**Symptom**: After bug 1 is fixed, a Theme [`sap_horizon_dark`] not registered for package [`@ui5/webcomponents`] console error fires on every page load with a non-default theme. UI5 component shadow DOM stylesheets don't fully switch.

**Root cause**: `Ui5WebcomponentsMainThemingService` and `Ui5WebcomponentsFioriThemingService` are `providedIn: 'root'` but never injected - their constructors (which register per-package theme loaders via `registerThemes()`) never run.

**Fix**: New theming-bridge/ secondary entrypoints in `@fundamental-ngx/ui5-webcomponents` and `@fundamental-ngx/ui5-webcomponents-fiori` export `provideUi5Webcomponents()` and `provideUi5WebcomponentsFiori()`, which force-construct the respective services.


InnaAtanasova (dismissed):

MariaIDineva (commented):
Since we now have `provideUi5Webcomponents` and `provideUi5WebcomponentsFiori`, shouldn't we also have `provideUi5WebcomponentsAi`?

Also, if we are to use those instead of `provideEnvironmentInitializer(() => inject(Ui5WebcomponentsMainThemingService))` and `provideEnvironmentInitializer(() => inject(Ui5WebcomponentsFioriThemingService))`, I think we should:
1. Update the docs to make the new approach visible for the users - i.e. `theming-docs.html` and `GETTING_STARTED.md`
2. Update our `apps/docs/src/app.config.ts` to use the new pattern.

libs/ui5-webcomponents/theming-bridge/index.ts
```diff
@@ -0,0 +1,10 @@
+import { EnvironmentProviders, inject, provideEnvironmentInitializer } from '@angular/core';
+import { Ui5WebcomponentsMainThemingService } from '@fundamental-ngx/ui5-webcomponents/theming';
+
+export { Ui5WebcomponentsMainThemingService };
```
MariaIDineva: Is this export needed?

libs/ui5-webcomponents-fiori/theming-bridge/index.ts
```diff
@@ -0,0 +1,10 @@
+import { EnvironmentProviders, inject, provideEnvironmentInitializer } from '@angular/core';
+import { Ui5WebcomponentsFioriThemingService } from '@fundamental-ngx/ui5-webcomponents-fiori/theming';
+
+export { Ui5WebcomponentsFioriThemingService };
```
MariaIDineva: Also, is this export needed too?

libs/ui5-webcomponents-ai/theming-bridge/index.ts
```diff
@@ -0,0 +1,10 @@
+import { EnvironmentProviders, inject, provideEnvironmentInitializer } from '@angular/core';
+import { Ui5WebcomponentsAiThemingService } from '@fundamental-ngx/ui5-webcomponents-ai/theming';
+
+export { Ui5WebcomponentsAiThemingService };
```
MariaIDineva: And this one?

libs/ui5-webcomponents-fiori/theming-bridge/ui5-webcomponents-fiori-theming-bridge.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+            providers: [provideUi5WebcomponentsFiori()]
+        });
+
+        await flushAsyncEffects();
+
+        const ui5Theming = TestBed.inject(Ui5ThemingService);
+        expect(ui5Theming.getAvailableThemes().length).toBeGreaterThan(0);
+    });
+
+    it('should call UI5 setTheme through the full bridge chain when provideUi5WebcomponentsFiori() is used', async () => {
```
MariaIDineva, line 62: Claude:

**libs/ui5-webcomponents/theming-bridge/ui5-webcomponents-theming-bridge.spec.ts:521, fiori and ai analogues**
 The integration tests titled "should call UI5 setTheme through the full bridge chain" manually wire up `Ui5WebcomponentsThemingService` and `Ui5ThemingBridgeService` directly instead of using `provideUi5ThemingBridge()`. This does not test the actual expected usage pattern `(provideUi5ThemingBridge() + provideUi5Webcomponents())`. Add at least one test per per-package spec that wires them via their public provider functions together, so the contract the docs describe is verified end-to-end.

libs/ui5-webcomponents-ai/theming/index.ts
```diff
@@ -0,0 +1,14 @@
+import { Injectable } from '@angular/core';
+import { WebcomponentsThemingProvider } from '@fundamental-ngx/ui5-webcomponents-base/theming';
+
+@Injectable({ providedIn: 'root' })
+class Ui5WebcomponentsAiThemingService extends WebcomponentsThemingProvider {
+  name = 'ui-5-webcomponents--ai-theming-service';
```
MariaIDineva: `ui-5-webcomponents...` - this is weird, shoudn't that be `ui5-webcomponents..`?
 Same for the other services.

MariaIDineva (commented):
Two issues that Claude flagged:

**libs/ui5-webcomponents/theming/index.ts, libs/ui5-webcomponents-fiori/theming/index.ts, libs/ui5-webcomponents-ai/theming/index.ts** - `CLAUDE.md` strictly prohibits hand-editing files in `libs/ui5-webcomponents*/`: these are auto-generated by `libs/webc-generator`. The theming/ subdirectory and its index.ts + ng-package.json are produced by generateThemingFiles() in the generator executor using theming-service-template.tpl. The PR commits hand-created copies that happen to match the template output today, but now there are two sources of truth. The next nx run ui5-webcomponents{,-fiori,-ai}:generate will overwrite them and diverge if the template ever changes. The correct fix is to update the generator to also emit theming-bridge/ secondary entry points so all files under these directories stay generated.

**libs/ui5-webcomponents/theming-bridge/index.ts, libs/ui5-webcomponents-fiori/theming-bridge/index.ts, libs/ui5-webcomponents-ai/theming-bridge/index.ts** - Same violation: new hand-created files under auto-generated library directories. The theming-bridge/ entry points are new infrastructure — they should be added to the webc-generator (a new template + generateThemingBridgeFiles() step) rather than committed manually. Without this, theming-bridge files live outside the generator's knowledge and will silently diverge from any future generator evolution.

InnaAtanasova:
The `name` property values use inconsistent dash patterns:

- `'ui-5-webcomponents--theming-service'` (double dash before theming)
- `'ui-5-webcomponents--fiori-theming-service'` (double dash)
- `'ui-5-webcomponents--ai-theming-service'` (double dash)

libs/ui5-webcomponents/theming-bridge/index.ts
droshev: not anymore

libs/ui5-webcomponents-ai/theming-bridge/index.ts
droshev: same

libs/ui5-webcomponents-ai/theming/index.ts
droshev: the naming was the same before the change, but i decided to update it per your recommendation

droshev:
> The `name` property values use inconsistent dash patterns:
> 
> * `'ui-5-webcomponents--theming-service'` (double dash before theming)
> * `'ui-5-webcomponents--fiori-theming-service'` (double dash)
> * `'ui-5-webcomponents--ai-theming-service'` (double dash)

the naming was the same before the change, but i decided to update it per your recommendation

MariaIDineva (approved):
:shipit:

PR #14408, opened 2026-07-27
fix(core): updates for Notifications after design review

InnaAtanasova:
## Related Issue(s)

part of https://github.com/SAP/fundamental-ngx/issues/14363

## Description

Adds keyboard interaction support to notification components and updates doc examples for consistency (per issue description)

### Features

- **Action key callbacks**: New `actionKeyHandler` input on `fd-notification` fires callbacks for Enter (activate) and Delete/Backspace (dismiss) keys
- **Arrow navigation**: Navigate between notifications in groups using Up/Down arrows while preserving focus on the same child element (e.g., stay on buttons when moving between notifications)
- **Home/End keys**: Jump to first/last notification in a group

### API
```typescript
<fd-notification [actionKeyHandler]="onAction">
  <!-- content -->
</fd-notification>

onAction(event: KeyboardEvent, key: 'enter' | 'delete'): void {
  // Handle Enter key press or Delete/Backspace
}
```

### Action Sheet update
The overflow actions menu in Notifications is using Action Sheet component. Per a11y specs it should close with Tab key:

Screenshot 2026-07-27 at 10 54 36 AM

### Testing
Added 10 new unit tests covering keyboard interactions

### DIFF
--- a/libs/core/action-sheet/action-sheet-body/action-sheet-body.component.ts
+++ b/libs/core/action-sheet/action-sheet-body/action-sheet-body.component.ts
@@ -68,6 +68,9 @@ export class ActionSheetBodyComponent implements AfterViewInit {
     @ContentChildren(ActionSheetItemComponent)
     private readonly _items: QueryList<ActionSheetItemComponent>;
 
+    /** Emits when the Tab key is pressed inside the action sheet body. */
+    readonly tabKeyDown = new Subject<void>();
+
     /** @hidden */
     private _refresh$ = new Subject<void>();
 
@@ -89,6 +92,11 @@ export class ActionSheetBodyComponent implements AfterViewInit {
     /** @hidden */
     @HostListener('keydown', ['$event'])
     keyDownHandler(event: KeyboardEvent): void {
+        if (event.key === 'Tab') {
+            event.preventDefault();
+            this.tabKeyDown.next();
+            return;
+        }
         if (this._keyboardSupportService.keyManager) {
             this._keyboardSupportService.onKeyDown(event);
         }
--- a/libs/core/action-sheet/action-sheet-body/action-sheet-body.component.ts
+++ b/libs/core/action-sheet/action-sheet-body/action-sheet-body.component.ts
@@ -1,3 +1,4 @@
+import { TAB } from '@angular/cdk/keycodes';
 import {
     AfterViewInit,
     ChangeDetectionStrategy,
@@ -5,7 +6,6 @@ import {
     ContentChildren,
     DestroyRef,
     ElementRef,
-    HostListener,
     Input,
     QueryList,
     ViewChild,
@@ -13,11 +13,11 @@ import {
     inject
 } from '@angular/core';
 
-import { KeyboardSupportService, Nullable } from '@fundamental-ngx/cdk/utils';
+import { KeyUtil, KeyboardSupportService, Nullable } from '@fundamental-ngx/cdk/utils';
 
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { ContentDensityObserver, contentDensityObserverProviders } from '@fundamental-ngx/core/content-density';
-import { Subject, merge, startWith, takeUntil } from 'rxjs';
+import { Observable, Subject, merge, startWith, takeUntil } from 'rxjs';
 import { ActionSheetItemComponent } from '../action-sheet-item/action-sheet-item.component';
 
 let actionSheetBodyUniqueIdCounter = 0;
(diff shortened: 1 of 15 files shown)


libs/core/action-sheet/action-sheet-body/action-sheet-body.component.ts
```diff
@@ -41,7 +41,11 @@ let actionSheetBodyUniqueIdCounter = 0;
     encapsulation: ViewEncapsulation.None,
     changeDetection: ChangeDetectionStrategy.OnPush,
     providers: [KeyboardSupportService, contentDensityObserverProviders()],
-    standalone: true
+    standalone: true,
```
droshev: not needed anymore?

libs/core/notification/notification/notification.component.ts
```diff
@@ -189,6 +197,24 @@ export class NotificationComponent extends AbstractFdNgxClass implements OnInit,
         }
     }
 
+    /** @hidden Calls the optional action callback when Enter or Delete/Backspace is pressed. */
+    _handleNotificationActionKeys(event: KeyboardEvent): void {
```
droshev: does this need to be public or private?

droshev (dismissed):

libs/core/action-sheet/action-sheet-body/action-sheet-body.component.ts
InnaAtanasova: I initially kept it for consistency with other Action Sheet components and directives, which still have it and are out of this PR scope. I removed it for this component.

libs/core/notification/notification/notification.component.ts
InnaAtanasova: It's public, but I initially named it with an underscore for consistency with the other methods in the component. Anyways, renamed the old method as well :) 
Screenshot 2026-07-29 at 3 36 46 PM

InnaAtanasova: Added also the additional unit test that was suggested.

libs/core/notification/notification-group-list/notification-group-list.component.ts
```diff
@@ -55,4 +63,101 @@ export class NotificationGroupListComponent implements AfterViewInit {
             notification.ariaLevel.set(2);
         });
     }
+
+    /** @hidden Handles list keyboard navigation and moves focus to the target notification. */
+    protected _onListKeydown(event: KeyboardEvent): void {
```
MariaIDineva: Since this method is not private, it should not be prefixed with an underscore.

libs/core/action-sheet/action-sheet.component.spec.ts
```diff
@@ -13,6 +17,35 @@ class TestComponent {
...
+            <fd-action-sheet-control>
+                <button fd-button>Action Sheet</button>
+            </fd-action-sheet-control>
+            <fd-action-sheet-body>
+                <fd-action-sheet-item>Option 1</fd-action-sheet-item>
+                <fd-action-sheet-item>Option 2</fd-action-sheet-item>
+            </fd-action-sheet-body>
+        </fd-action-sheet>
+    `,
+    standalone: true,
```
MariaIDineva: no need of `standalone: true`

libs/core/notification/notification/notification.component.ts
```diff
@@ -183,12 +191,30 @@ export class NotificationComponent extends AbstractFdNgxClass implements OnInit,
...
     /** @hidden Listen and close notification on Escape key */
-    _closeNotificationEsc(event: KeyboardEvent): void {
+    closeNotificationEsc(event: KeyboardEvent): void {
         if (this.escKeyCloseable && KeyUtil.isKeyCode(event, ESCAPE) && this._notificationRef) {
             this._notificationRef.dismiss('escape');
         }
     }
 
+    /** @hidden Calls the optional action callback when Enter or Delete/Backspace is pressed. */
+    handleNotificationActionKeys(event: KeyboardEvent): void {
```
MariaIDineva: Shouldn't `handleNotificationActionKeys` and `closeNotificationEsc` be protected?

libs/core/notification/notification-group-list/notification-group-list.component.ts
```diff
@@ -55,4 +63,101 @@ export class NotificationGroupListComponent implements AfterViewInit {
...
+        if (!isNavigationKey) {
+            return false;
+        }
+
+        const target = event.target;
+        if (!(target instanceof HTMLElement)) {
+            return false;
+        }
+
+        if (!this._hostElement.nativeElement.contains(target)) {
```
MariaIDineva: One from Claude:

**libs/core/notification/notification-group-list/notification-group-list.component.ts:51,508** `_hostElement` injection is redundant

_hostElement is injected solely for this._hostElement.nativeElement.contains(target) inside _shouldHandleNavigation. Because the (keydown) listener is registered on the host element itself, any event the handler receives was already dispatched from within that element — the contains check is unconditionally true. Remove the injection and the guard.

droshev (approved):

PR #14409, opened 2026-07-27
fix(core): change Product Switch name to App Launcher

InnaAtanasova:
## Related Issue(s)

closes none

## Description

The designers informed us that Product Switch is now called App Launcher externally. 

- added the new name as an alternative to Product Switch (kept both names)
- updated docs and links
- updated translation files for the aria label and title props of the button

### DIFF
--- a/libs/docs/core/docs-data.json
+++ b/libs/docs/core/docs-data.json
@@ -221,7 +221,7 @@
         },
         {
             "url": "core/product-switch",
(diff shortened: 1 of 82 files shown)


droshev (approved):

MariaIDineva (approved):
Is there anything we need to do in order to notify the translations team that they will need to translate the property anew?

PR #14420, opened 2026-07-29
fix(core): improve aria-description for special types Buttons

### LINKED ISSUE #14415: ButtonComponent: auto-generated aria-description from fdType is not a valid ARIA value (WCAG Level A)
### Describe your proposal

ButtonComponent auto-generates an `aria-description` attribute whose value is the visual style name of the button type (e.g. `aria-description="emphasized"` for `fdType="emphasized"`). This happens for the four "special" types: `emphasized`, `positive`, `negative`, `attention`.

A visual style name is not a meaningful screen-reader description. ARIA validators (and JAWS screen reader) flag it as an invalid ARIA property value, violating WCAG Level A checkpoint ACC-264.1 ("Ensure ARIA roles, states and properties are valid").

Root cause: `button.component.ts` — the `buttonAriaDescription` computed getter returns `fdTypeState()` when no explicit `ariaDescription` input is provided and the type is in `SPECIAL_BUTTON_TYPES`.

Expected behavior: No `aria-description` should be auto-generated from the button type, or the value should be a meaningful, localizable string — not a CSS style name.

Note: This was introduced intentionally in #10292 / #10461, but the chosen value (`"emphasized"` etc.) does not convey useful information to screen reader users.

Angular: 19.x, Fundamental Library: 0.62.3

### Affected Component

ButtonComponent (button[fd-button])

1. Render any button with fdType="emphasized" (or positive/negative/attention) without an explicit ariaDescription input.
2. Inspect the DOM — the button will have aria-description="emphasized".
3. Run an ARIA validator or use JAWS screen reader — it flags the value as invalid.

### StackBlitz example(s)

_No response_

### Priority

Medium

### Declaration

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14415

## Description
Replaces direct aria-description attribute usage on special button types with a hidden-span + aria-describedby pattern, fixing ACC-264.1 (duplicate aria-description).

### Changes:

- Special button types (Emphasized, Positive, Negative, Attention) now render a hidden `fd-button__sr-only` span for their type description, referenced via `aria-describedby`
- Added ariaDescription` input to supply custom context descriptions (e.g. "Deletes permanently"), also rendered as a hidden span — supports Angular input binding, native `aria-description` attribute, and bound `[attr.aria-description]`
- Native `aria-description` attribute is captured once after render and removed from the DOM to avoid duplicate ARIA attributes
- Added ariaTypeDescription` input to override the default i18n type descriptions (Emphasized → "Default Action", Positive → "Positive Action", Negative → "Negative Action", Attention → "Warning")
- Default type descriptions are auto-translated via i18n; custom `ariaTypeDescription` values are used as-is
- Updated unit tests and docs

## Screenshots
Testtin scenarios:
- **Standard Button**: no special type button and no ariaDescription passed => should not render hidden spans and `aria-describedby` attr.
- **Emphasized Button**: special type and no ariaDescription passed => should only render the default type description and `aria-describedby` attr linking to the hidden span that contains the default type description.
- **Ghost button**: same as Standard btn, no no special type button but tests `[attr.aria-description]="ghostButtonAriaDescription"` and if the binding would take effect with `ngOnInit` lifecycle.
- **Positive Button**: special type and `[attr.aria-description]="'This adds a new record'"`. Should render 2 hidden span that are linked in `aria-describedby` attr. `aria-description` should be removed from the button and the value should be placed in a hidden span.
- **Negative button**: special type and `ariaDescription="This permanently deletes the record"`. The value of `ariaDescription` should be in a hidden span and linked with the type description in `aria-describedby`.
- **Attention Button**: testing overwriting the default type description if the user wants to do it by setting ` ariaTypeDescription="This requires immediate attention"`
- **Transparent Button**: this is a no special button so by default it shouldn't have an aria-description, but in case the user wants to provide one, they can still do it by `aria-description="Transparent Button Description"` (or `ariaDescription`). This also tests if `aria-description` is removed from the button and placed in a hidden span.

Screenshot 2026-07-29 at 1 33 51 PM

-----------------------------------------

**While reviewing keep in mind this requirement:**
DISCUSS_THIS

-----------------------------------------

The user can choose to completely overwrite the `aria-describedby` values by passing an id to an element in their markup. In this case the hidden spans with default values we provide should not be in the DOM. This is tested in the buttons with icons. In **Emphasized Button** we provide `aria-describedby` and no hidden spans are present. In **Positive Button** we illustrate the default behavior.

Screenshot 2026-07-29 at 2 58 47 PM


libs/core/button/button.component.spec.ts
MariaIDineva, line 1: Just a small comment from Claude on the unit tests:

**libs/core/button/button.component.spec.ts** Missing test: verify that after ngOnInit, the native aria-description attribute has been removed from the DOM element. The Ghost button scenario in the docs example relies on this removal but there is no unit assertion for it.

**libs/core/button/button.component.spec.ts** Missing test: verify that when ariaDescribedBy is provided (or a native aria-describedby attribute is present), the hidden description spans are not rendered in the DOM. The template guards are there (!ariaDescribedBy() && !nativeAriaDescribedBy()), but the negative case is untested.

MariaIDineva (approved):

droshev (approved):

PR #14429, opened 2026-07-31
feat(i18n): add auto-correct command to i18n-manage executor

### LINKED ISSUE #14025: review both i18n READMEs
### Problem 1:

We currently have two README-s about i18n:

https://github.com/SAP/fundamental-ngx/blob/main/libs/i18n/README.md?plain=1#L201
https://github.com/SAP/fundamental-ngx/blob/main/libs/nx-plugin/src/executors/i18n-manage/README.md

This step in the first readme is misleading:

**Step 2: Add to ALL .properties Files**
 
Add to **all** files in `libs/i18n/src/lib/translations/` (use English text everywhere):

**TODO:** Figure out if we still need both README-s. Make sure the one that stays contains the correct information.

### Problem 2:

As part of **adding** a new translation key with the command `nx run i18n:i18n-manage --command=add --key=coreButton.submit --value="Submit" --commentType=XBUT --comment="Submit button"` it seems like the `FdLanguage` interface located in `libs/i18n/src/lib/models/fd-language.ts` is not updated and needs to be updated manually. The script should update that as well.
**Rename** and **remove** scripts should update that interface as well.

droshev:
Fixes #14025

## Summary

Implement complete i18n workflow simplification: **`add` now auto-updates fd-language.ts** (one-step), **new `remove` command** with automatic interface cleanup, and **interface-updater utility** for safe TypeScript manipulation.

### New Features
- **`remove` command**: Deletes translation keys from properties and fd-language.ts interface. Auto-removes empty component sections. Regenerates all TypeScript files.
- **Interface-updater utility**: Safely parses and updates fd-language.ts with proper indentation, duplicate prevention, and JSDoc generation.

### Improvements
- **`add` command**: Now uses interface-updater for automatic one-step fd-language.ts + properties update. Validates multi-dot keys, supports auto-generated comments.
- **Documentation**: Refactored translations.md for safer workflows (Files section moved to top, Remove workflow reordered for safety, Reference material separated).

### Bug Fixes
- Fixed hardcoded fd-language.ts path in add-key operation
- Removed unused `fdLanguagePath` parameter from schema
- Added missing SAP UI5 comment types in variant properties files

## How to Test

**Quick test of new features:**

1. **Add command (one-step workflow):**
   ```bash
   nx run i18n:i18n-manage --command=add \
     --key=coreButton.testSave \
     --value="Test Save" \
     --comment="Test save button"
   ```
   ✅ Verify: `translations.properties` AND `fd-language.ts` both updated

2. **Remove command (auto-cleanup):**
   ```bash
   nx run i18n:i18n-manage --command=remove --key=coreButton.testSave
   ```
   ✅ Verify: Key removed from both files, empty sections cleaned up

3. **Validate all changes:**
   ```bash
   nx run i18n:i18n-manage --command=validate
   ```
   ✅ Verify: No errors, all 39 translation files valid

### Automated Testing
```bash
# Run all nx-plugin tests (includes i18n operations)
nx run nx-plugin:test

# Run specific test files
nx run nx-plugin:test --testFile=add-key.spec.ts
nx run nx-plugin:test --testFile=remove-key.spec.ts
nx run nx-plugin:test --testFile=interface-updater.spec.ts
```

## Breaking Changes
None.


libs/nx-plugin/src/executors/i18n-manage/operations/add-key.ts
```diff
@@ -62,6 +63,12 @@ export async function addKey(options: AddKeyOptions): Promise<AddKeyResult> {
         };
     }
 
+    // Step 1.5: Validate that comment is provided (unless commentType is NOTR)
+    // Note: comment is optional; if not provided, it will be auto-generated by generateComment()
+    if (!comment && commentType === 'NOTR') {
+        // NOTR type explicitly requires no comment
+    }
```
InnaAtanasova: What is this validation? It's an empty if block.

libs/nx-plugin/src/executors/i18n-manage/schema.d.ts
```diff
@@ -23,4 +23,5 @@ export interface I18nManageExecutorSchema {
     commentType?: CommentType;
     propertiesPath?: string;
     fdLanguagePath?: string;
```
InnaAtanasova: `fdLanguagePath` is no longer used as it's now hardcoded in `add-key.ts`.

libs/nx-plugin/src/executors/i18n-manage/operations/correct.ts
```diff
@@ -0,0 +1,272 @@
...
+                applied.push(`Fixed malformed comment for key "${section.key}"`);
+            }
+        }
+
+        // Trim whitespace in value (only leading, preserve trailing for intent)
+        const trimmedValue = section.value.trimStart();
+        if (trimmedValue !== section.value) {
+            section.value = trimmedValue;
+            applied.push(`Trimmed whitespace in key "${section.key}"`);
+        }
```
InnaAtanasova: Why is the trimming only at the beginning of the value? Shouldn't we trim both (start and end)? In which case the trailing whitespace would make sense?

libs/nx-plugin/src/executors/i18n-manage/operations/remove-key.ts
```diff
@@ -3,6 +3,141 @@ import { readFileSync, writeFileSync } from 'fs';
...
+        removeCount = 2;
+    }
+
+    interfaceLines.splice(removeStart, removeCount);
+
+    // Check if component section is now empty
+    const updatedComponentEnd = component.endLine - removeCount;
+    if (!hasComponentContent(interfaceLines, component.startLine, updatedComponentEnd)) {
+        interfaceLines.splice(component.startLine, updatedComponentEnd - component.startLine + 1);
+    }
```
InnaAtanasova, line 136: **comment from Claude:**

Empty section removal might be too aggressive

The removeKeyFromInterface function removes entire component sections when they become empty (lines 133-136). While this is clean, it could surprise users who expect the section structure to match their mental model of the codebase even when empty.

Recommendation: Consider keeping empty component sections with a comment, or document this auto-cleanup behavior explicitly in `translations.md`.

InnaAtanasova (commented):
Left some comments

libs/nx-plugin/src/executors/i18n-manage/operations/add-key.ts
droshev: Removed

libs/nx-plugin/src/executors/i18n-manage/schema.d.ts
droshev: Removed

libs/nx-plugin/src/executors/i18n-manage/operations/correct.ts
droshev: Changed from `trimStart()` to `trim()`

libs/nx-plugin/src/executors/i18n-manage/operations/remove-key.ts
droshev: updated

InnaAtanasova (dismissed):

InnaAtanasova (dismissed):

mikerodonnell89 (changes requested):
` nx run i18n:i18n-manage --command=validate` is failing due to missing comments on some product switch translations, which I think was what inspired this PR in the first place if i'm not mistaken, let's add those

mikerodonnell89 (commented):
I have a bit of an odd feeling about automatically updating fd-language.ts, having the AI make assumptions if a comment is missing, and this auto-correct feature in general. Seems like it invites contributors to just not bother with comments, and the AI will inevitably get something wrong when generating a comment, which will cause the translation team to not fully grasp the meaning of the key they're translating. And we'll have heaps of translations to different languages that don't make any sense. Just my personal opinion - it is better to have the contributors to this library be very deliberate about adding keys, adding comments, and yes we should do plenty of verification then to make sure translations.properties has everything. Doing all of that should properly safeguard us from receiving PRs from the translation team that might be missing comments etc

droshev:
> ` nx run i18n:i18n-manage --command=validate` is failing due to missing comments on some product switch translations, which I think was what inspired this PR in the first place if i'm not mistaken, let's add those

Actually, my goal is to leave it to the translation team to fix that

droshev:
> I have a bit of an odd feeling about automatically updating fd-language.ts, having the AI make assumptions if a comment is missing, and this auto-correct feature in general. Seems like it invites contributors to just not bother with comments, and the AI will inevitably get something wrong when generating a comment, which will cause the translation team to not fully grasp the meaning of the key they're translating. And we'll have heaps of translations to different languages that don't make any sense. Just my personal opinion - it is better to have the contributors to this library be very deliberate about adding keys, adding comments, and yes we should do plenty of verification then to make sure translations.properties has everything. Doing all of that should properly safeguard us from receiving PRs from the translation team that might be missing comments etc

Mike, understand your concern. Few points:

1. `add` doesn't force comments - only auto-generates if missing. Opt-in.
2. `fd-language.ts` sync prevents manual errors - CI validation catches mismatches.
3. correct is CI safety-net, not encouragement - still requires review before merge.
4. Translation team workflow unchanged - same `.properties` files, same comments (human or sensible defaults).

Goal: reduce busy work, catch errors early.

mikerodonnell89:
> > ` nx run i18n:i18n-manage --command=validate` is failing due to missing comments on some product switch translations, which I think was what inspired this PR in the first place if i'm not mistaken, let's add those
> 
> Actually, my goal is to leave it to the translation team to fix that

If they're missing from translations.properties we shouldn't expect the translation team to fix them. But hopefully with your PR here moving forward, we shouldn't be missing them from translations.properties

droshev:
> > > ` nx run i18n:i18n-manage --command=validate` is failing due to missing comments on some product switch translations, which I think was what inspired this PR in the first place if i'm not mistaken, let's add those
> > 
> > 
> > Actually, my goal is to leave it to the translation team to fix that
> 
> If they're missing from translations.properties we shouldn't expect the translation team to fix them. But hopefully with your PR here moving forward, we shouldn't be missing them from translations.properties

Exactly. PR prevents incomplete .properties from ever reaching translation team - validate catches it, correct fixes it in CI. They always get complete data. It is up to us(reviewers) whether we want to accept/or not the comment

mikerodonnell89 (approved):

PR #14439, opened 2026-08-04
fix(platform): remove ngZone from platform table

### LINKED ISSUE #14435: fix(platform): remove NgZone from the platform Table for zoneless compatibility
## Problem

The platform table stack still depends on `NgZone` across five files (`platform/table/table.component.ts`, `table-column-resizer`, `table-row`, and the three `table-helpers` directives). Under zoneless change detection (the Angular default since v20; this workspace is on `@angular/core` 22.0.5), two of those uses are silent correctness breaks and the rest are dead no-ops. `NgZone.onMicrotaskEmpty` **never emits** and `ngZone.run()` **no longer triggers change detection** in a zoneless app.

Relates to #14035 (closed COMPLETED — the table was left unmigrated). Blocks #14038 (zoneless bootstrap). Sibling of #14434 (core table).

## What breaks (scenarios)

**Scenario 1 — `tableRowsSet` output stops emitting + infinite scroll stalls.** `_onZoneFree()` (line 2489) waits on `onMicrotaskEmpty`:

```ts
private _onZoneFree(callback: () => void): void {
  this._ngZone.onMicrotaskEmpty.pipe(take(1)).subscribe(() => { callback(); });
}
```

- Caller :1148 (`ngAfterViewChecked`) fires the **public `tableRowsSet` output** → zoneless, the output silently never emits (consumer-facing API regression).
- Caller :2483 region drives **infinite-scroll next-page loading** via the IntersectionObserver spy row → zoneless, scroll pagination stalls.

**Scenario 2 — scroll-triggered pagination doesn't repaint.** Line 1838:

```ts
this._ngZone.run(() => { this.setCurrentPage(currentPage + 1); });
```

Zoneless, `.run()` executes the callback but does **not** schedule change detection. The new page loads into state but the view may not repaint until an unrelated tick.

## What needs to be fixed

1. **`table.component.ts:2489` (`_onZoneFree`):** replace the primitive with `afterNextRender` (the "wait for DOM to settle, run once" intent maps directly). Verify both callers (`tableRowsSet` emit at :1148, spy-row check at :2483) still fire post-render.
2. **`table.component.ts:1838`:** remove the `.run()` wrapper. Ensure `setCurrentPage` triggers CD through signal-driven state (preferred) or `markForCheck()`. *(Verify `setCurrentPage` internals first — if already signal-driven, this is already safe and drops to a no-op removal.)*
3. **Cleanup:** remove the `runOutsideAngular(...)` wrappers around listener setup (dead ceremony under zoneless), then strip the `NgZone` import/injection from all five files:
   - `platform/table/table.component.ts:2443` (mouseleave → hide resizer)
   - `table-column-resizer.component.ts:58,102` (mousedown/mouseup resize)
   - `table-row.component.ts:250` (keydown / DnD)
   - `table-scrollable.directive.ts:53,198` (scroll stream + setTimeout scroll-top init)
   - `table-cell-resizable.directive.ts:72` (mousemove)
   - `table-header-resizer.directive.ts:59` (keydown header resize)

   **Caveat (table-row:250, table-header-resizer:59):** these keydown handlers currently run outside the zone. Confirm any state they mutate is signal-driven or calls `markForCheck()` — if a handler relies on an implicit tick, removing the wrapper turns it into a correctness break. Flag on discovery.

## How to validate

- [ ] `grep -rn NgZone libs/platform/table libs/platform/table-helpers --include="*.ts"` (excluding specs) returns nothing.
- [ ] In a **zoneless** build: `tableRowsSet` output emits after rows render.
- [ ] In a **zoneless** build: infinite scroll loads the next page when the spy row intersects.
- [ ] In a **zoneless** build: scroll-triggered pagination repaints the new page without an external tick.
- [ ] No regression in resize / column-reorder / row-DnD / scroll interactions.

## Testing note

The correctness-break reproductions run through `onMicrotaskEmpty` / `.run()` / CD timing — browser-only. jsdom does not run the zone-stability engine, so a TestBed spec asserting "output emitted" or "page repainted" passes regardless (false green). The RED must be an E2E / running-zoneless check; lock-in unit tests only where they can genuinely fail.

mikerodonnell89:
fixes #14435


libs/platform/table-helpers/directives/table-scrollable.directive.ts
```diff
@@ -81,8 +69,7 @@ export class TableScrollableDirective implements TableScrollable, OnInit, OnDest
     /** @hidden */
     constructor(
         public elementRef: ElementRef<HTMLElement>,
-        protected scrollDispatcher: TableScrollDispatcherService,
-        protected ngZone: NgZone
```
droshev, line 85: keep the constructor parameter but mark it @deprecated and don't use it otherwise we have to mark this as a breaking change

droshev:
@mikerodonnell89 2 comments from claude's review:

```
  [Major] _onZoneFree → queueMicrotask changes emission timing under zone.js

  Before: ngZone.onMicrotaskEmpty.pipe(take(1)) fires after Angular has drained its microtask queue and CD is stable.
  After: queueMicrotask fires as a raw microtask — before zone.js CD has settled.

  The tableRowsSet emitter fires earlier in the tick now. Under zone.js this may cause subscribers to receive the event before the
  table's internal state is fully flushed. This needs a real-browser repaint test (not TestBed — TestBed false-greens this class of
  timing issue). Please provide before/after evidence that tableRowsSet consumers see consistent state at the point they receive the
  emission.

  ---
  [Major] ngZone.run(() => setCurrentPage(...)) removed — infinite-scroll regression risk on zone.js

  table.component.ts previously wrapped the infinite-scroll page advance in this._ngZone.run(). This forced a CD-triggering zone
  re-entry, which matters if the scroll callback is outside Angular's zone. After the removal, if the caller is outside the zone under
  zone.js, the page advance won't trigger CD.

  Please confirm: is the call site (_onScrolledToEnd or equivalent) guaranteed to be inside Angular's zone? If it's wired via fromEvent
  inside runOutsideAngular, removing ngZone.run() is a regression on the zoneful path. Show the call chain.
```

Can you verify it?

Also

after looking at the ticket more carefully - the ticket was wrong on one point. the instruction to remove all runOutsideAngular wrappers was too broad.

runOutsideAngular on scroll, mousemove, and the setTimeout in initializeScrollTop should stay. here's why: under zone.js, every event that runs inside Angular's zone triggers change detection. scroll and mousemove are high-frequency - on a large table you can get dozens of CD cycles per second just from the user moving the mouse or scrolling. the original wrappers were there specifically to avoid that cost.

the good news is keeping them doesn't hurt zoneless at all - runOutsideAngular is effectively a no-op under NoopNgZone, so zoneless behavior is unchanged.

the actual zoneless problems from the ticket (onMicrotaskEmpty never emitting, ngZone.run() not triggering CD) are correctly fixed in this PR. just put runOutsideAngular back on those three high-frequency listeners and we're good.

libs/platform/table-helpers/directives/table-cell-resizable.directive.ts
```diff
@@ -65,11 +65,11 @@ export class PlatformTableCellResizableDirective
     });
 
     /** @hidden */
-    private readonly _zone = inject(NgZone);
+    private readonly _ngZone = inject(NgZone);
```
InnaAtanasova: I am confused. Shouldn't we be removing NgZone entirely?

InnaAtanasova (commented):
The PR lacks description of what is the purpose of this PR and the changes being done. 

The issue it reference has these as requirements
- `table.component.ts` (_onZoneFree): replace the primitive with afterNextRender (the "wait for DOM to settle, run once" intent maps directly).

- `table.component.ts` remove the `.run()` wrapper. Ensure `setCurrentPage` triggers CD through signal-driven state (preferred) or `markForCheck()`. 

- Cleanup: remove the `runOutsideAngular(...)` wrappers around listener setup (dead ceremony under zoneless), then strip the `NgZone` import/injection from all five files:

    - platform/table/table.component.ts (mouseleave → hide resizer)
    - table-column-resizer.component.ts (mousedown/mouseup resize)
    - table-row.component.ts (keydown / DnD)
    - table-scrollable.directive.ts (scroll stream + setTimeout scroll-top init)
    - table-cell-resizable.directive.ts (mousemove)
    - table-header-resizer.directive.ts (keydown header resize)

but I still see `runOutsideAngular(...)`  and imports of `NgZone`

libs/platform/table/table.component.ts
```diff
@@ -1145,9 +1144,16 @@ export class TableComponent<T = any>
...
-        });
+        // Use afterNextRender to defer emission until after Angular CD is stable AND browser layout is complete.
+        // This guarantees that subscribers can safely query DOM (scroll positions, element dimensions, focus).
+        // Unlike queueMicrotask, which runs before zone.js CD settles, afterNextRender preserves the timing
+        // guarantee of the original ngZone.onMicrotaskEmpty approach.
+        afterNextRender(
+            () => {
+                emitter.emit();
+            },
+            { injector: this.injector }
```
InnaAtanasova: Just a note about this section:

The replacement of `ngZone.onMicrotaskEmpty` with `afterNextRender()` changes timing semantics. The original code waited for the zone to be stable (all microtasks complete), while `afterNextRender()` fires after the next render cycle. `afterNextRender` runs after browser layout/paint, which is later than microtask empty. This could cause race conditions.

If the goal is to preserve the original "wait for zone to settle" behavior, use queueMicrotask() instead:

queueMicrotask(() => {
    emitter.emit();
});

If the new timing is intentional, then it can stay with `afterNextRender`

InnaAtanasova (commented):
Just a minor comment to consider. The optional NgZone logic we discussed looks great

InnaAtanasova (dismissed):

droshev (approved):

PR #14444, opened 2026-08-05
fix(core, platform): resolve zoneless change detection issues in Flexible Column Layout, Textarea (Platform) and Multi Input (Platform) 

### LINKED ISSUE #14436: fix(core,platform): fix off-tick template-bound writes that break zoneless (non-table)
## Problem

Three components mutate **template-bound plain (non-signal) fields** from an off-tick callback (a window-resize subscription or a `setTimeout`) with **no `markForCheck()`/`detectChanges()`**. Under zone.js the automatic tick repainted them; under zoneless change detection (the Angular default since v20; this workspace is on `@angular/core` 22.0.5) there is no tick, so the state changes but the view does not repaint. These are the same failure class as the table `NgZone` breaks (#14434, #14435) — but because none of them touch `NgZone`, a `grep NgZone` never finds them. Each was verified against source.

Blocks #14038 (zoneless bootstrap) alongside the table issues.

## What breaks (scenarios)

**A1 — `libs/core/flexible-column-layout/flexible-column-layout.component.ts:399`**
`fromEvent(window,'resize').pipe(debounceTime(100)).subscribe(() => this._responsiveLayoutChangeHandler())`. The handler sets plain `layout`, `_leftColumnSeparator`, `_rightColumnSeparator` — all template-bound. The only CD in the path is a deferred `layoutChange.emit` (an output, not a self-repaint).
→ **Resizing the window does not re-lay-out the columns.**

**A2 — `libs/platform/form/multi-input/base-multi-input.ts:584`**
`fromEvent(window,'resize').subscribe(() => this._getOptionsListWidth())`. `_getOptionsListWidth()` sets plain `this.maxWidth` / `this.minWidth`, bound at `multi-input.component.html:13`.
→ **The options popover does not resize on window resize.**

**A3 — `libs/platform/form/text-area/text-area.component.ts:263`**
`setTimeout(() => this.updateCounterInteractions())` on the paste path. `updateCounterInteractions()` sets plain `counterExcessOrRemaining`, `exceededCharCount`, `_textAreaCharCount`, all bound in the OnPush template.
→ **The character counter does not repaint after a paste.**

## What needs to be fixed

For each, convert the mutated fields to `signal()` (preferred — the counter/width/layout values are read in the template, so a signal is the correct model and gives durable zoneless-safety). Minimal alternative: add `markForCheck()` at the end of each handler. Prefer the signal conversion.

- A1: `layout`, `_leftColumnSeparator`, `_rightColumnSeparator` → signals (or `markForCheck()` in `_responsiveLayoutChangeHandler`).
- A2: `maxWidth`, `minWidth` → signals (or `markForCheck()` in `_getOptionsListWidth`).
- A3: counter fields → signals (or `markForCheck()` in/after `updateCounterInteractions`).

These are three small, independent changes — they can land in parallel and in any order.

## How to validate (browser-only)

- [ ] **A1:** in a zoneless build, render a flexible-column-layout, resize the window across a breakpoint → columns re-lay-out and separators update.
- [ ] **A2:** in a zoneless build, open a multi-input's options popover, resize the window → popover width tracks the new viewport.
- [ ] **A3:** in a zoneless build, paste text over the max length into a text-area → the character counter updates immediately.
- [ ] Lock-in unit tests asserting the fields are signal-consumed (only where genuinely failable).

## Testing note

All three reproductions require a real repaint under a running zoneless bootstrap — jsdom will not show the missed repaint (false green in a TestBed CD assertion). The RED for each is browser / E2E; add a signal-consumption lock-in spec where it can actually fail.

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14436

## Description

In zoneless change detection with `OnPush` strategy, component state must be reactive. Properties used in templates need to be signals so Angular can track changes automatically without `markForCheck()` calls. The text-area component was also writing to signals during template rendering (via `getUpdatedState()` method binding), which is forbidden in Angular 22+.

### 1. flexible-column-layout (core)

- Migrated `_leftColumnSeparator` and `_rightColumnSeparator` to signals
- Column separators now update reactively on window resize in zoneless mode
- Updated unit tests to call separator signals with () syntax

### 2. multi-input (platform)

- Migrated `_maxWidth` and `_minWidth` to signals
- Token width calculations now update reactively on window resize in zoneless mode
- No test updates needed (properties are private/protected)

### 3. text-area (platform)

- Converted `counterExcessOrRemaining`, `exceededCharCount`, and `hasTextExceeded` from **writable** signals to **computed** signals
- All counter properties now derive from single source: `_textAreaCharCount` signal
- Character counter now updates reactively after paste events in zoneless mode
- Removed all signal writes during template rendering


InnaAtanasova:
closing in favor of https://github.com/SAP/fundamental-ngx/pull/14465

PR #14463, opened 2026-08-11
fix(core): fix layout grid reflow inside flexible colum

### LINKED ISSUE #12332: Layout Grid Column responsiveness lost in Flexible Column Layout
bug

#### Describe your proposal.
I have a layout grid with responsive columns: https://sap.github.io/fundamental-ngx/#/core/layout-grid. When I open a new column with https://sap.github.io/fundamental-ngx/#/core/flexible-column-layout, the layout grids columns do not react as I would expect. 
They do not respond to the changes in the flexible column width as they would in case the window is resize.

"@angular/core": "17.3.12"
"@fundamental-ngx/core": "0.50.3",

See belows stackblitz example. 
Some screenshots:

Initially opening example with layout-grid
image

decreasing window size to trigger column responsiveness:
image

increasing window size and opening the second column where I would expect realignment of layout-grid columns but instead they are truncated:
image

A prude forward merge of two fundamentals stackblitz examples:
https://stackblitz.com/edit/amu28b 

( used stackblitz: https://sap.github.io/fundamental-ngx/#/core/layout-grid#responsive-columns 
 and stackblitz: https://sap.github.io/fundamental-ngx/#/core/flexible-column-layout#default )

yes

yes

no

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/12332

## Description
The fundamental-styles layout grid uses viewport-based @media queries, so grid items did not reflow when an FCL column changed width — they only resized proportionally.

Fix: add `container-type: inline-size` to `.fd-flexible-column-layout__column` and override the fd-col-md/lg/xl breakpoint classes with @container queries that fire based on the FCL column width instead of the viewport width.

The container query rules use SCSS @for loops to generate max-width/min-width overrides for all 12 column spans. In the medium (600–1023 px) container, lg/xl overrides are declared before md overrides so that the md value wins the cascade for elements that carry both classes (e.g. fd-col-md--6 + fd-col-lg--4).

Updated the default FCL example to demonstrate the fix with a responsive layout grid that correctly shows 1 → 2 → 3 items per row as columns open.

## Screenshots
Test the first example of Flexible Column Layout
Screenshot 2026-08-11 at 4 54 14 PM
Screenshot 2026-08-11 at 4 54 03 PM


apps/e2e-harness/e2e/specs/interaction/flexible-column-layout.spec.ts
```diff
@@ -0,0 +1,149 @@
+import { expect, test } from '../../fixtures/base.fixture';
+
+async function getContentWidth(page: any): Promise<number> {
```
MariaIDineva: Shouldn't this (and in the function below) be of type `Page`?

MariaIDineva (dismissed):

MariaIDineva (approved):

PR #14467, opened 2026-08-12
fix(platform): restore initial sidebar focus in settings generator dialog

### LINKED ISSUE #13827: [QUESTION] [platform] [Settings Generator]  Is the initial focus in the dialog changed ? It was the first focusable element in the content
In version v0.58.0  and before The initial focus is the first focusable element
Image

In version v0.59.0 The initial focus is in the footer rather than the content, and also v0.58.0-rc.40
Image

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/13827

## Description
When the Settings Generator was rendered in a dialog, initial focus was landing on the footer submit button. The expected behavior is to place focus on the first item in the sidebar, which was the behavior in earlier versions.

- restore initial focus to the first selected sidebar item when Platform Settings Generator is opened inside a dialog
- prevent focus from falling through to the footer submit action
- simplify the focus targeting logic by using Angular view queries for sidebar items

### DIFF
--- a/libs/platform/settings-generator/layouts/settings-generator-sidebar-layout/settings-generator-sidebar-layout.component.ts
+++ b/libs/platform/settings-generator/layouts/settings-generator-sidebar-layout/settings-generator-sidebar-layout.component.ts
@@ -8,9 +8,8 @@ import {
     HostBinding,
     inject,
     OnInit,
-    QueryList,
     ViewChild,
-    ViewChildren,
+    viewChildren,
     ViewEncapsulation
 } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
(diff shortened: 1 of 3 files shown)


libs/platform/settings-generator/layouts/settings-generator-sidebar-layout/settings-generator-sidebar-layout.component.ts
```diff
@@ -68,6 +70,10 @@ export class SettingsGeneratorSidebarLayoutComponent
     @ViewChild('listElement', { read: ElementRef })
     private _listElement: ElementRef;
 
+    /** @hidden */
+    @ViewChildren('sidebarItem', { read: ElementRef })
```
MariaIDineva: Is it possible to use `viewChildren()` instead?

MariaIDineva (commented):
Shouldn't we add some unit test(s)?

MariaIDineva (approved):

mikerodonnell89 (approved):

PR #14471, opened 2026-08-14
feat(i18n): lazy language registry with provideFundamentalTranslations

### LINKED ISSUE #14174: i18n: Add secondary entry points per language to enable tree-shaking (1.2MB bundle impact)
## Problem

The `@fundamental-ngx/i18n` package bundles all 36 languages into a single FESM module (~1,284 KB). Any app that imports from `@fundamental-ngx/core` pulls in the entire i18n package as a transitive dependency — even if the app only needs English.

**Impact:** 1.2 MB of unused translation data in the initial bundle for single-language apps. In our case, this was 59% of a 2.7 MB initial bundle.

### Root causes

1. **No secondary entry points** — there is no way to import `@fundamental-ngx/i18n/en` or `@fundamental-ngx/i18n/languages/english`. The only entry point is the root, which re-exports everything.
2. **Barrel export re-exports all languages** — `src/lib/languages/index.ts` does `export *` for all 36 language files, preventing tree-shaking.
3. **No `exports` field in package.json** — modern bundlers cannot resolve subpaths.
4. **Core components depend on the full i18n package** — e.g., `@fundamental-ngx/core/busy-indicator`, `toolbar`, and `message-strip` import from `@fundamental-ngx/i18n`, which pulls all languages into the initial chunk.

### Bundle analysis evidence

```
node_modules/@fundamental-ngx/i18n/fesm2022/fundamental-ngx-i18n.mjs  — 1,284 KB (in initial chunk)
```

Each individual language is ~30-35 KB, and the core tokens/pipes/resolver are ~5 KB. The remaining ~1,100 KB is 35 languages that aren't used.

### 1. Add secondary entry points per language

```
@fundamental-ngx/i18n           → core only (tokens, pipes, resolver, types — ~5 KB)
@fundamental-ngx/i18n/en        → English translations
@fundamental-ngx/i18n/de        → German translations
@fundamental-ngx/i18n/fr        → French translations
...
```

### 2. Add `exports` field to package.json

```json
{
  "exports": {
    ".": { "default": "./fesm2022/fundamental-ngx-i18n.mjs" },
    "./en": { "default": "./fesm2022/fundamental-ngx-i18n-en.mjs" },
    "./de": { "default": "./fesm2022/fundamental-ngx-i18n-de.mjs" }
  }
}
```

### 3. Decouple core component imports from language data

Core components should only import the injection tokens and types from `@fundamental-ngx/i18n` — not the language constants. The `FD_LANGUAGE_SIGNAL` factory should use a minimal English fallback (or empty object) without importing the full `FD_LANGUAGE_ENGLISH` constant from the barrel.

### 4. Consider dynamic language loading by default

The `FD_LANGUAGE_SIGNAL` factory could detect `LOCALE_ID` at runtime and `import()` the matching language chunk lazily, making all language data lazy-loaded by default:

```typescript
// Instead of eagerly importing FD_LANGUAGE_ENGLISH
factory: () => {
  const locale = inject(LOCALE_ID);
  const lang = signal(MINIMAL_FALLBACK);
  import(`@fundamental-ngx/i18n/${locale}`).then(m => lang.set(m.default));
  return lang;
}
```

## Current workaround

We replaced the FESM in `node_modules` with an English-only shim (~31 KB) via a postinstall script. This reduced the initial bundle from 2.70 MB to 1.37 MB. The shim exports the same public API surface but only includes English translations.

## Expected bundle impact of the fix

| Scenario | Current | With fix |
|----------|---------|----------|
| Single-language app (en) | 1,284 KB | ~35 KB |
| Two languages (en + de) | 1,284 KB | ~70 KB |
| All languages | 1,284 KB | 1,284 KB (no regression) |

## Additional context

- Angular 21 + `@angular/build:application` (Vite + esbuild)
- `@fundamental-ngx/i18n` version 0.62.x
- tsconfig `paths` aliases do NOT work as a workaround because esbuild resolves `node_modules` before checking paths for packages that exist on disk

droshev:
Closes #14174.

i18n was bundling all 37 languages eagerly, which is about 1.29 MB even for an English-only app. This changes the registry to start with English only, and apps register whatever else they need:

```ts
provideFundamentalTranslations(FD_LANGUAGE_GERMAN, ...)  // just the ones you use
provideAllFundamentalLanguages()                          // all 37, old behavior
```

Heads up, this is breaking: if you had a non-English LOCALE_ID and no language provider, you'll now fall back to English. There's a dev-mode console.warn that tells you which locale is missing and how to fix it. Setting FD_LANGUAGE_SIGNAL yourself with .set() still works the same.

The actual change is 3 files, everything else is docs and tests:
- utils/detect-language.ts is the core, the English-seeded registry and the resolution order
- providers.ts has the two providers. The Chinese aliases match on the locale string now, not object identity, so a cloned/patched language object still works
- utils/tokens.ts is just the dev warning, guarded by isDevMode() so it costs nothing in prod

To validate: try it by hand: non-English LOCALE_ID with no provider gives you English + one warning, then add provideFundamentalTranslations(<lang>) and it resolves correctly with no warning.

Stuff you don't need to review: the registry being module-scoped is on purpose (that's what makes it lazy, and it was already a singleton before this). The init-order thing is documented and there's a test pinning it. Ignore the docs/CHANGELOG churn. And there are 3 pre-existing lint warnings that are unrelated, tracked separately.


InnaAtanasova (approved):

libs/i18n/MIGRATION.md
```diff
@@ -0,0 +1,67 @@
...
+    providers: [provideAllFundamentalLanguages()]
+};
+```
+
+This registers all 37 built-in languages at bootstrap, exactly matching the pre-lazy
+behavior. All languages remain in the bundle. Use this if you need zero-effort migration.
+
+### Resolution order (unchanged)
+
+`detectLanguage` still applies the same 5-step resolution for _registered_ languages:
```
InnaAtanasova: should be 6-step

InnaAtanasova (dismissed):

droshev:
Closing it as it needs more topics to be covered from the issue

PR #14472, opened 2026-08-14
fix(platform): focus issues on the table-view-settings-dialog filters

### LINKED ISSUE #14007: [Platform][fdp-table]ACC-273.2 (Level A) In setting dialogs, should focus the list item or next time after selected
### Describe your proposal

When a user selects a status list item in the “Filter by: Status” dialog, focus should remain on the selected item (or move in a predictable way within the status list, such as to the next item) so users can continue interacting with the list/filter options. 

Image

### Proposed Solution

When a user selects a status list item in the “Filter by: Status” dialog, focus should remain on the selected item (or move in a predictable way within the status list, such as to the next item) so users can continue interacting with the list/filter options. 

Image

### Proposed Alternatives

_No response_

### Organization

_No response_

### StackBlitz example(s)

https://stackblitz.com/run?file=src%2Fapp%2Fplatform-table-settings-dialog-example.component.ts

### Priority

None

### Declaration

mikerodonnell89:
fixes #14007

### DIFF
--- a/libs/platform/table/components/table-view-settings-dialog/filtering/filter-custom.component.ts
+++ b/libs/platform/table/components/table-view-settings-dialog/filtering/filter-custom.component.ts
@@ -1,8 +1,10 @@
 import {
     afterEveryRender,
+    AfterViewInit,
     ChangeDetectionStrategy,
     ChangeDetectorRef,
     Component,
+    ElementRef,
     EventEmitter,
     inject,
     Input,
@@ -33,7 +35,7 @@ import { TableViewSettingsFilterComponent } from '../table-view-settings-filter.
     providers: [contentDensityObserverProviders()],
     imports: [NgTemplateOutlet]
 })
-export class FilterCustomComponent {
+export class FilterCustomComponent implements AfterViewInit {
     /** ViewSettingsFilter options the filter is created from */
     @Input()
     filter: TableViewSettingsFilterComponent;
@@ -75,8 +77,27 @@ export class FilterCustomComponent {
     private readonly _cdr = inject(ChangeDetectorRef);
 
     /** @hidden */
+    private readonly _elementRef = inject(ElementRef);
+
+    /** @hidden */
     constructor(private contentDensityObserver: ContentDensityObserver) {
         afterEveryRender(() => this._checkValueChanges());
+    }
+
+    /** @hidden */
+    ngAfterViewInit(): void {
+        // Focus the first focusable element in the custom template
+        const focusableSelectors = [
+            'input:not([disabled])',
+            'select:not([disabled])',
+            'textarea:not([disabled])',
+            'button:not([disabled])',
+            'a[href]',
+            '[tabindex]:not([tabindex="-1"])'
+        ].join(', ');
+
+        const firstFocusable = this._elementRef.nativeElement.querySelector(focusableSelectors);
+        firstFocusable?.focus();
     }
 
     /** @hidden */
--- a/libs/platform/table/components/table-view-settings-dialog/filtering/filter-custom.component.ts
+++ b/libs/platform/table/components/table-view-settings-dialog/filtering/filter-custom.component.ts
@@ -1,5 +1,6 @@
 import {
     afterEveryRender,
+    afterNextRender,
     AfterViewInit,
     ChangeDetectionStrategy,
     ChangeDetectorRef,
@@ -7,6 +8,7 @@ import {
     ElementRef,
     EventEmitter,
     inject,
+    Injector,
     Input,
     Output
 } from '@angular/core';
@@ -80,13 +82,15 @@ export class FilterCustomComponent implements AfterViewInit {
     private readonly _elementRef = inject(ElementRef);
 
     /** @hidden */
+    private readonly _injector = inject(Injector);
+
+    /** @hidden */
     constructor(private contentDensityObserver: ContentDensityObserver) {
         afterEveryRender(() => this._checkValueChanges());
     }
 
     /** @hidden */
     ngAfterViewInit(): void {
-        // Focus the first focusable element in the custom template
         const focusableSelectors = [
             'input:not([disabled])',
             'select:not([disabled])',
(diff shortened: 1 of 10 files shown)


libs/platform/table/components/table-view-settings-dialog/filtering/filter-custom.component.ts
```diff
@@ -74,11 +76,30 @@ export class FilterCustomComponent {
...
+            'input:not([disabled])',
+            'select:not([disabled])',
+            'textarea:not([disabled])',
+            'button:not([disabled])',
+            'a[href]',
+            '[tabindex]:not([tabindex="-1"])'
+        ].join(', ');
+
+        const firstFocusable = this._elementRef.nativeElement.querySelector(focusableSelectors);
+        firstFocusable?.focus();
```
InnaAtanasova: Maybe wrap this logic in `afterNextRender` block as this can cause `NG0100` (Expression Changed After Check) if the focus triggers state changes in parent or sibling components.

libs/platform/table/components/table-view-settings-dialog/filtering/filter-custom.component.ts
```diff
@@ -74,11 +76,30 @@ export class FilterCustomComponent {
...
+        const focusableSelectors = [
+            'input:not([disabled])',
+            'select:not([disabled])',
+            'textarea:not([disabled])',
+            'button:not([disabled])',
+            'a[href]',
+            '[tabindex]:not([tabindex="-1"])'
+        ].join(', ');
+
+        const firstFocusable = this._elementRef.nativeElement.querySelector(focusableSelectors);
```
InnaAtanasova: `querySelector()` will find the first matching element in the entire subtree, which might not be what you want if there are nested focusable elements.

libs/platform/table/components/table-view-settings-dialog/filtering/filter-custom.component.ts
```diff
@@ -74,11 +76,30 @@ export class FilterCustomComponent {
...
+    /** @hidden */
+    ngAfterViewInit(): void {
+        // Focus the first focusable element in the custom template
+        const focusableSelectors = [
+            'input:not([disabled])',
+            'select:not([disabled])',
+            'textarea:not([disabled])',
+            'button:not([disabled])',
+            'a[href]',
+            '[tabindex]:not([tabindex="-1"])'
```
InnaAtanasova, line 100: No visibility check: Hidden elements (display: none, visibility: hidden, opacity: 0, off-screen) will match but shouldn't be focused.
There'a also a potential race condition
If the template contains dynamically loaded content (e.g., @if, @defer, async components), the focusable element might not exist yet when `ngAfterViewInit` runs.

libs/platform/table/components/table-view-settings-dialog/filtering/filter-custom.component.ts
```diff
@@ -74,11 +76,30 @@ export class FilterCustomComponent {
...
+        ].join(', ');
+
+        const firstFocusable = this._elementRef.nativeElement.querySelector(focusableSelectors);
+        firstFocusable?.focus();
+    }
+
```
InnaAtanasova, line 135: maybe a summary for all the comments above :D 
Just a suggestion
```
ngAfterViewInit(): void {
    // Use afterNextRender to avoid NG0100
    afterNextRender(() => {
        const focusableSelectors = [
            'input:not([disabled]):not([hidden])',
            'select:not([disabled]):not([hidden])',
            'textarea:not([disabled]):not([hidden])',
            'button:not([disabled]):not([hidden])',
            'a[href]:not([hidden])',
            '[tabindex="0"]' // Only explicit tab order, not arbitrary values
        ].join(', ');

        const firstFocusable = this._elementRef.nativeElement.querySelector(
            focusableSelectors
        ) as HTMLElement | null;
        
        // Check if element is actually visible and interactive
        if (firstFocusable && this._isVisible(firstFocusable)) {
            firstFocusable.focus();
        }
    }, { injector: this._injector });
}

private _isVisible(element: HTMLElement): boolean {
    const rect = element.getBoundingClientRect();
    return rect.width > 0 && rect.height > 0 && 
           getComputedStyle(element).visibility !== 'hidden';
}

```

libs/platform/table/components/table-view-settings-dialog/filtering/filter-multi-select.component.ts
```diff
@@ -41,6 +51,10 @@ export class FilterMultiSelectComponent {
     @Output()
     valueChange: EventEmitter<any[]> = new EventEmitter();
 
+    /** @hidden */
+    @ViewChildren(ListItemComponent)
+    listItems: QueryList<ListItemComponent>;
```
InnaAtanasova: this can be a signal

libs/platform/table/components/table-view-settings-dialog/filtering/filter-single-select.component.ts
```diff
@@ -35,6 +44,10 @@ export class FilterSingleSelectComponent {
     @Output()
     valueChange: EventEmitter<any[]> = new EventEmitter();
 
+    /** @hidden */
+    @ViewChildren(ListItemComponent)
+    listItems: QueryList<ListItemComponent>;
```
InnaAtanasova: can you make this a signal?

libs/platform/table/components/table-view-settings-dialog/filtering/filters.component.ts
```diff
@@ -73,8 +78,31 @@ export class FiltersComponent implements AfterViewInit, OnInit {
...
 
+    /** @hidden */
+    private readonly _elementRef = inject(ElementRef);
+
+    /** @hidden */
+    private readonly _injector = inject(Injector);
+
     /** hidden */
-    constructor(private readonly _cd: ChangeDetectorRef) {}
+    constructor(private readonly _cd: ChangeDetectorRef) {
```
InnaAtanasova: `_cd` can also be injected with `inject` call

libs/platform/table/components/table-view-settings-dialog/filtering/filters.component.ts
```diff
@@ -73,8 +78,31 @@ export class FiltersComponent implements AfterViewInit, OnInit {
...
+        // Focus first focusable element when returning to the filters list
+        effect(() => {
+            const step = this.activeStep();
+            if (step === ACTIVE_STEP.SELECT_FILTER) {
+                afterNextRender(
```
InnaAtanasova: Every time `activeStep()` changes, the effect runs and schedules another `afterNextRender` callback. These callbacks never get cleaned up. If the user switches steps repeatedly, you're queueing up dozens/hundreds of render callbacks.


effect(() => {
    if (step === ACTIVE_STEP.SELECT_FILTER) {
        afterNextRender(() => { /* ... */ });  // ⚠️ NEW CALLBACK EVERY TIME
    }
});

libs/platform/table/components/table-view-settings-dialog/filtering/filters.component.ts
```diff
@@ -73,8 +78,31 @@ export class FiltersComponent implements AfterViewInit, OnInit {
...
+                    () => {
+                        // Query from the root element to find the list items rendered via template outlet
+                        const firstListItem = document.querySelector('li[fd-list-item]');
```
InnaAtanasova: wouldn't this find the first match in the entire document, not just the component?

InnaAtanasova (changes requested):

libs/platform/table/components/table-view-settings-dialog/filtering/filter-custom.component.ts
mikerodonnell89: I think this is what we want - what would be the alternative? I think we'd have to hand the reigns over to the application developer (which would be fine, but I think it's nice for the library to attempt to focus the first item by default)

InnaAtanasova (approved):

PR #14473, opened 2026-08-14
fix(core, platform): update form messages a11y

### LINKED ISSUE #13222: ACC264.1 (Screen Reader) [fd-form-label][fd-form-message][fd-form-control] Screen reader cannot get the messages
Is it a bug ?
 Yes 

In the form, only UI name  is being read but other information such as ‘Error message’ ‘Error icon’ is not being read by the screen reader

Example: 

https://stackblitz.com/edit/cvhgsabu?file=src%2Fapp%2Fform-messaging-state-example.component.html

Expect:

1. Read the required status instead of reading star
2.Read the error status of the input if the  type is error 
3. Read the error messgae 

This is raised by Acessebility Team with high level

Affected Check Point:	[ACC-264](https://wiki.wdf.sap.corp/wiki/display/Acc/ACC-264+Details#ACC264Details-acc2641)

Expected Behavior:

Along with the name and the UI type screen reader should also convey other information (Such as ‘Error message’, ‘Error icon’, etc.….)to the users.

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/13222

## Description
TBD


libs/docs/core/form-message/form-message-header/form-message-header.component.html
```diff
@@ -4,9 +4,43 @@
...
+            Form messages are announced to screen readers with <code>aria-live="assertive"</code> and
+            <code>aria-atomic="true"</code>, ensuring assistive technologies immediately announce changes.
+        </p>
+        <ul>
+            <li>
+                <b>Automatic ARIA linking</b>: When using <code>fd-form-input-message-group</code>, the component
+                automatically links the form control to the message using the appropriate ARIA attributes:
+                <ul>
+                    <li>
+                        For <b>error</b> messages: sets both <code>aria-errormessage</code> (pointing to the message ID)
```
MariaIDineva: Wouldn't it be confusing for a developer to set an `aria-errormessage` which we would replace with `aria-describedby`. Maybe they would expect to see `aria-errormessage` in the dom, but it will be missing. Shouldn't we just ask the dev to pass `aria-describedby` directly?

libs/core/form/form-input-message-group/form-input-message-group.component.ts
```diff
@@ -100,10 +106,96 @@ export class FormInputMessageGroupComponent {
...
+
+    /** @hidden Query for the projected form message */
+    private readonly _formMessage = contentChild(FormMessageComponent);
+
+    /** @hidden */
+    ngAfterContentInit(): void {
+        // Set up ARIA attributes linking the control to the form message
+        effect(
+            () => {
+                const formMessage = this._formMessage();
```
MariaIDineva, line 126: Per Claude:

The effect does not clean up ARIA attributes when the fd-form-message is conditionally removed (e.g. wrapped in @if). When this._formMessage() becomes undefined the effect returns early without removing the previously set aria-describedby and aria-invalid from the control element. A screen reader would then try to resolve a describedby reference pointing to a detached or non-existent element. The early-return block needs to remove the previously managed attributes:

```
if (!formMessage || !controlElement) {
    // Clean up previously managed attributes
    this._renderer.removeAttribute(controlElement, 'aria-describedby');
    this._renderer.removeAttribute(controlElement, 'aria-invalid');
    return;
}
```

libs/core/form/form-input-message-group/form-input-message-group.component.ts
```diff
@@ -100,10 +106,96 @@ export class FormInputMessageGroupComponent {
...
+                const controlElement = this._getControlElement();
+
+                if (!formMessage || !controlElement) {
+                    return;
+                }
+
+                const messageId = formMessage.id();
```
MariaIDineva, line 143: Per Claude:

**libs/platform/form/form-group/form-field/form-field.component.html, lines 32 and 57** — The platform `fdp-input-message-group` extends FormInputMessageGroupComponent and will inherit the new ngAfterContentInit. But the platform form-field sets ids on its nested fd-form-message elements via [attr.id]="'fdp-form-message-grouped-errors-' + id". [attr.id] sets the DOM attribute directly — it does not feed the new id signal input. The component's host: { '[attr.id]': 'id()' } binding will then overwrite those custom ids with the auto-generated default (e.g. fd-form-message-id-3). The new _formMessage.id() value read inside the effect will therefore be the auto-id, not the intended platform id, breaking any ARIA association the platform was relying on. The platform template must be changed to use [id] (the input binding) rather than [attr.id]:

```
<!-- before -->
<fd-form-message [attr.id]="'fdp-form-message-grouped-errors-' + id">
<!-- after -->
<fd-form-message [id]="'fdp-form-message-grouped-errors-' + id">
```
 This is how you can test:
Image
Image

InnaAtanasova:
The fix in Platform with the correct IDs
Screenshot 2026-08-19 at 3 13 36 PM
Screenshot 2026-08-19 at 3 12 56 PM

libs/platform/form/form-group/form-field/form-field.component.html
```diff
@@ -22,14 +22,14 @@
     </span>
 </ng-template>
 <ng-template #withFormMessage>
-    <fdp-input-message-group class="fd-col" [class]="_fieldColumnLayoutClass" [attr.aria-labelledby]="_getLabelledBy()">
+    <fdp-input-message-group class="fd-col" [class]="_fieldColumnLayoutClass">
```
InnaAtanasova, line 25: We don't need aria-labelledby on this level. 
We need to connect the input with the label, which we do. The input has an `aria-labelledby` that points to the group label and `aria-describedby` that points to the form message.

Image
Image

libs/platform/form/form-group/form-field/form-field.component.html
```diff
@@ -22,14 +22,14 @@
...
         <ng-template #triggerItem>
             <ng-content></ng-content>
         </ng-template>
 
         @if (hasErrors() && groupedErrors.length > 0) {
             @for (error of groupedErrors; track error) {
-                <fd-form-message [type]="error.directive.type" [attr.id]="'fdp-form-message-grouped-errors-' + id">
+                <fd-form-message [type]="error.directive.type">
```
InnaAtanasova, line 32: removed these as we provide defaults in the Core Input Group logic. No need to provide another default IDs

libs/platform/form/form-group/form-field/form-field.component.ts
```diff
@@ -771,19 +771,6 @@ export class FormFieldComponent
...
-        let retVal = 'fdp-form-label-' + this.id;
-        if (this.hasErrors() && this.groupedErrors.length > 0) {
-            retVal = retVal + ' fdp-form-message-grouped-errors-' + this.id;
-        }
-        if (this.groupedErrors.length === 0 && this.hasErrors() && this.i18Strings) {
-            retVal = retVal + ' fdp-form-message-error-' + this.id;
-        }
-
-        return retVal;
-    }
```
InnaAtanasova, line 785: This logic is removed as we don't need the method in the template anymore.

MariaIDineva (dismissed):

mikerodonnell89 (approved):

PR #14487, opened 2026-08-21
feat(core): product-switch design review updates

MariaIDineva:
## Related Issue(s)

closes none

## Description

- Add busy state input to ProductSwitchBodyComponent with fd-busy-indicator overlay
- Add two-column layout mode for < 3 items (.fd-product-switch__body--col-2)
- Add URL navigation support with configurable target per item
- Move max-height: 33rem (4-row cap) into component SCSS per spec
- Add popover border-radius token override
- Add list-mode item min-height: 6rem per spec
- Update unit tests: fix two stale assertions, add coverage for new functionality
- Add docs examples for busy state, two-column layout, and URL navigation
- Document SAP Start first-item convention in Basic section

## Screenshots

Screenshot 2026-08-24 at 16 43 29
Screenshot 2026-08-24 at 16 43 42
Screenshot 2026-08-24 at 16 43 51


libs/core/product-switch/product-switch-body/product-switch-body.component.scss
```diff
@@ -1,5 +1,21 @@
...
+    border-radius: var(--sapPopover_BorderCornerRadius, var(--sapElement_BorderCornerRadius));
+    max-height: 33rem;
+}
+
+// 2-column layout for fewer than 3 items: 2 × 11.25rem max-width + 0.5rem gap
+.fd-product-switch__body--col-2 .fd-product-switch__list {
+    width: 23rem;
+}
+
+// List-mode item min-height: spec requires 6rem (96px), fundamental-styles ships 5rem (80px)
```
mikerodonnell89, line 14: So shouldn't this be changed in fundamental-styles?

mikerodonnell89 (changes requested):
Noticed some intermittent problems with the items that have navigation sometimes not working. FYI I was seeing this with all of the links, not just the "new tab" ones but that's all I captured on video:

https://github.com/user-attachments/assets/f7c39e71-5136-4aec-956e-83908882db6e

libs/core/product-switch/product-switch-body/product-switch-body.component.scss
MariaIDineva: Yes, I have a PR there as well - https://github.com/SAP/fundamental-styles/pull/6377

MariaIDineva:
> Noticed some intermittent problems with the items that have navigation sometimes not working. FYI I was seeing this with all of the links, not just the "new tab" ones but that's all I captured on video:
> 
>  Screen.Recording.2026-08-24.at.3.16.13.PM.mov

That was a very interesting bug, thanks for catching it, Mike!

**The problem:**

 The `<a>` element inside each item was responsible for navigation — it would follow its `href` whenever it received a click. But sometimes when `itemClicked` emitted and closed the popover, the panel closed too fast and destroyed the `<a>` before the browser could follow the link — so navigation silently failed.

**The fix:**

- Navigation is now handled explicitly in the click handler via `window.open` instead of relying on the `<a>`'s default behavior.
- Because those items are draggable, a flag `_wasDragged` is set when a drag starts (using CDK's own started event). The click handler checks this flag and skips navigation if a drag just happened.
- `preventDefault()` is added to the `<a>`'s click handler so it never navigates on its own. It keeps its href only for right-click / middle-click support.

InnaAtanasova (dismissed):

mikerodonnell89 (changes requested):
Need to update the PR to remove the snapshot changes

MariaIDineva:
> Need to update the PR to remove the snapshot changes

Done, thanks!

InnaAtanasova (approved):

mikerodonnell89 (approved):

PR #14495, opened 2026-08-25
feat(i18n): lazy registry + secondary entry points for tree-shaking

### LINKED ISSUE #14174: i18n: Add secondary entry points per language to enable tree-shaking (1.2MB bundle impact)

droshev:
Closes #14174.

Splits `@fundamental-ngx/i18n` into per-language secondary entry points so bundlers can tree-shake unused translations.

**Before:** any import from `@fundamental-ngx/i18n` pulled the full 1.3 MB FESM into every app.
**After:** apps load only what they register — 258 KB for a single-language app, 64 KB for English-only.

```ts
// Specific languages — only those bundles are loaded (~33 KB each)
import { provideFundamentalTranslations } from '@fundamental-ngx/i18n';
import { FD_LANGUAGE_GERMAN } from '@fundamental-ngx/i18n/de';
import { FD_LANGUAGE_FRENCH } from '@fundamental-ngx/i18n/fr';

providers: [provideFundamentalTranslations(FD_LANGUAGE_GERMAN, FD_LANGUAGE_FRENCH)]

// Restore old behavior — all 37 languages, one line
import { provideAllFundamentalLanguages } from '@fundamental-ngx/i18n/all';
```

**Review focus (3 files, rest is scaffolding):**
- `libs/i18n/src/lib/providers.ts` — `provideFundamentalTranslations` (Chinese alias fix included)
- `libs/i18n/src/lib/utils/detect-language.ts` — lazy registry, shared alias constants
- `libs/i18n/all/providers-all.ts` — `provideAllFundamentalLanguages` in its own entry

**Validate:** `nx run i18n:test` → 49 suites / 23134 green. `nx run core:build` → clean.

⚠️ **Breaking:** `FD_LANGUAGE_*` constants and `provideAllFundamentalLanguages` are no longer exported from the root entry. See `MIGRATION.md` for import path changes. `provideFundamentalTranslations`, `FD_LANGUAGE_SIGNAL`, all tokens, pipes, and utilities stay in the root.

**Skip:** the 166 `libs/i18n/<locale>/` scaffolding files (index.ts + ng-package.json + translation symlink per language) — they're mechanical and generated. The `package.json` lint-staged change skips symlinks in the prettier hook.


libs/i18n/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@fundamental-ngx/i18n",
-    "version": "0.64.2-rc.45",
+    "version": "0.64.2-rc.43",
```
MariaIDineva: Why is this package downgrade? Maybe that is a mistake and we'd want to correct that?

mikerodonnell89 (dismissed):

InnaAtanasova (dismissed):

libs/i18n/MIGRATION.md
```diff
@@ -0,0 +1,116 @@
+# @fundamental-ngx/i18n — Migration Guide
+
+## Lazy language registry (0.64.x → minor)
```
MariaIDineva, line 3: Per Claude:

**[libs/i18n/MIGRATION.md] Version headers are contradictory for a single release.**
The file has two sections: ## Lazy language registry (0.64.x → minor) and ## Secondary entry points — language import paths changed (0.65.x → minor). Both changes land in the same PR. A reader arriving from a changelog will find the split confusing — one section suggests the secondary entry points shipped a release later. Either merge them under one version header, or add a note clarifying that both changes ship together.

droshev: I have reviewed the documentation and this was edited. I hope the updated version is better?

MariaIDineva (dismissed):

InnaAtanasova (dismissed):

InnaAtanasova (approved):

PR #14497, opened 2026-08-26
fix(core): improve Slider aria attributes

### LINKED ISSUE #14491: Slider: aria-valuemin / aria-valuemax / aria-valuenow contain text instead of numeric values (invalid ARIA)
### Describe your proposal

```html
  <div role="slider"
       aria-valuetext="Current value is 3"
       aria-valuenow="Current value is 3"   <!-- should be: 3 -->
       aria-valuemax="Value is 100"          <!-- should be: 100 -->
       aria-valuemin="Value is 0"            <!-- should be: 0 -->
       aria-label="Slider minimum value is 0, maximum value is 100"
       tabindex="0" class="fd-slider__handle" role="slider"></div>
```

  Root cause

  In libs/core/slider/slider.component.html, the numeric ARIA attributes are bound through the fdTranslate pipe, which wraps the value in a
  translation string ('Value is {value}', 'Current value is {value}'). This affects all three handles:

  - single mode — lines ~84–101
  - range handle 1 — lines ~138–155
  - range handle 2 — lines ~175–192
```html
  [attr.aria-valuemin]="('coreSlider.singleValueminDetails' | fdTranslate: { value: minValue })()"
  [attr.aria-valuemax]="('coreSlider.singleValuemaxDetails' | fdTranslate: { value: maxValue })()"
  [attr.aria-valuenow]="_useSliderValuePrefix
      ? ('coreSlider.singleValueNowDetails' | fdTranslate: { value: getValuenow(...) })()
      : getValuenow(...)"
```

  The internal _useSliderValuePrefix flag defaults to true and only becomes false momentarily during drag, so the resting DOM (what screen
  readers and scanners see) always exposes the text form. There is no @Input to opt out of this.

  Proposed fix

  Bind the raw numeric values to aria-valuemin / aria-valuemax / aria-valuenow, and keep the human-readable string only in aria-valuetext —
  which is the spec-correct place for it and preserves the screen-reader announcement:
```html
  [attr.aria-valuemin]="minValue"
  [attr.aria-valuemax]="maxValue"
  [attr.aria-valuenow]="getValuenow(_position, _sliderValueTargets.SINGLE_SLIDER)"
  [attr.aria-valuetext]="('coreSlider.singleValueNowDetails' | fdTranslate: { value: getValuenow(_position, _sliderValueTargets.SINGLE_SLIDER)
  })()"
```

According to WAI-ARIA 1.2, `aria-valuenow`, `aria-valuemin` and `aria-valuemax` take a `<number>` value ("a decimal number, between the minimum and maximum values"), whereas the human-readable string belongs in `aria-valuetext`. Using text in the numeric attributes violates WCAG 2.1 SC 4.1.2 Name, Role, Value (Level A) and is flagged by the axe-core rule `aria-valid-attr-value`.

found in: 0.63.0 / Angular 21.2.19; also can repro in (0.64.2-rc.43)

### Affected Component

core / Slider (SliderComponent)

  1. Add a core `<fd-slider>` to a template with numeric bounds, e.g.:
     ```html
     <fd-slider [min]="0" [max]="100" [ngModel]="3" tooltipMode="readonly"></fd-slider>
  2. Render the component and inspect the slider handle element (div.fd-slider__handle[role="slider"]) in the DOM.
  3. Observe the aria-* value attributes on the handle.

  Expected:
```html
  <div role="slider" aria-valuemin="0" aria-valuemax="100" aria-valuenow="3" aria-valuetext="Current value is 3">
```

  Actual:
```html
  <div role="slider" aria-valuemin="Value is 0" aria-valuemax="Value is 100" aria-valuenow="Current value is 3" aria-valuetext="Current value is 3">
```
  Running any ARIA validator (e.g. axe-core aria-valid-attr-value) against the page reports the handle's aria-valuemin / aria-valuemax /
  aria-valuenow as invalid, since they are not numeric.

### StackBlitz example(s)

_No response_

### Priority

Medium

### Declaration

InnaAtanasova:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/14491

### Issue
Slider component was setting `aria-valuemin`, `aria-valuemax`, and `aria-valuenow` to translated text strings (e.g., "Value is 100") instead of numeric values, violating `WCAG 2.1 SC 4.1.2` and the `aria-valid-attr-value` rule.

### Changes
- Bind raw numeric values directly to `aria-valuemin`, `aria-valuemax`, and `aria-valuenow`
- Keep human-readable text in `aria-valuetext`.  Removed the default translated strings and added it as an input for the user to specify.
- i18n keys: removed unused verbose translation keys (coreSlider.singleValueminDetails, etc.) and add single coreSlider.ariaLabel for default label ('Slider Handle')

NOTE: 
We have the option to provide custom values for the min and max values (see the screenshot). These values need to be strings. Only for this case the returned value will be a string. In the other cases will be a number. For this reason we have

```
get minValue(): string | number {
     return this.customValues.length > 0 ? this.customValues[this.min as number].label : this.min;
}

get maxValue(): string | number {
    return this.customValues.length > 0 ? this.customValues[this.max as number].label : this.max;
}
```
Screenshot 2026-08-26 at 2 08 19 PM


mikerodonnell89 (dismissed):
What's the reason behind not giving aria-valuetext a default `Value is {value}`?

droshev (dismissed):

InnaAtanasova:
> What's the reason behind not giving aria-valuetext a default `Value is {value}`?

> The `aria-valuetext` attribute provides a human-readable text **_alternative_** for a slider's numeric value when the bare number alone does not make sense to screen reader users.

It's an alternative, it should not have a default value if not necessary. For example, you might have `aria-valuemin="1"`, 
`aria-valuemax="3"` and `aria-valuenow="2"` and you want `aria-valuenow="2"` to be read by the screenreader as "Medium", then you provide `aria-valuetext="Medium"`. 
Another example can be values 1, 2, 3, 4, 5, 6 and 7 that represent the days. Then you can have `aria-valuetext="Wednesday"` for `aria-valuenow="3"` and so on.
We can't know what these alternative would be, that's why I exposed this as an input the user can set.

libs/core/slider/slider.component.html
```diff
@@ -72,34 +72,18 @@
...
-        [attr.aria-label]="
-            ariaLabel || ('coreSlider.singleMinMaxDetails' | fdTranslate: { min: minValue, max: maxValue })()
-        "
+        [attr.aria-label]="ariaLabel || ('coreSlider.ariaLabel' | fdTranslate)()"
         [attr.aria-labelledby]="ariaLabelledBy"
         [fdPopoverTrigger]="tooltipMode ? singleSliderPopover : null"
-        [attr.aria-valuemin]="('coreSlider.singleValueminDetails' | fdTranslate: { value: minValue })()"
-        [attr.aria-valuemax]="('coreSlider.singleValuemaxDetails' | fdTranslate: { value: maxValue })()"
-        [attr.aria-valuenow]="
-            _useSliderValuePrefix
```
MariaIDineva, line 87: `_useSliderValuePrefix` is not used anymore and should be removed from the component.
There is also this private method `_resetPrefix` which is redundant now as well.

InnaAtanasova: removed them. Thanks!

droshev (approved):

PR #14499, opened 2026-08-27
fix(deps): ui5-webcomponents version 2.26.0

mikerodonnell89:
fixes none


MariaIDineva (commented):
Will there be a separate PR to display the new features?

droshev:
> Will there be a separate PR to display the new features?

I would say we should add them in the same PR, no?

MariaIDineva:
> > Will there be a separate PR to display the new features?
> 
> I would say we should add them in the same PR, no?

We have set this rule:
Screenshot 2026-08-28 at 14 54 31

So maybe for any bug fixes or additional features of already existing components, we should update the examples as part of this PR. If there are completely new ones - open an issue to build a new page with the corresponding examples

InnaAtanasova (dismissed):

libs/docs/ui5-webcomponents-fiori/user-settings-dialog/examples/basic-sample.html
```diff
@@ -35,7 +35,7 @@
             tooltip="User Account"
             headerText="User Account"
         >
-            <ui5-user-settings-account-view>
+            <ui5-user-settings-account-view [showEditButton]="true">
```
droshev: maybe you can enrich the example with linking also an event for editing?

InnaAtanasova (approved):

droshev (approved):

PR #14505, opened 2026-08-28
fix(platform): harden icon-tab-bar stackContent active-tab fix (#14501)

### LINKED ISSUE #14501: bug(platform): [IconTabBar] active property on fdp-icon-tab-bar-tab is ignored when stackContent="true"
### Describe your proposal

When using `<fdp-icon-tab-bar>` from the platform module, setting the `[active]="true"` property on a specific `<fdp-icon-tab-bar-tab>` works correctly if `[stackContent]="false"`. 

However, when `[stackContent]="true"` is applied to the parent component, the `[active]` input on the child tab is completely ignored upon the initial render, and the component forces the first tab (index 0) to be active instead.

Additionally, using `[stackContent]="true"` in Angular 17/18 throws a performance warning in the browser console: 
`NG0956: The configured tracking expression (track by identity) caused re-creation of the entire collection of size 2...`
This suggests that the internal `@for` loop or `*ngFor` used to render the stacked tabs lacks an optimal `track` function.

**Expected Behavior:**
The tab with `[active]="true"` should be the one displayed and visually selected upon the initial render, regardless of whether `stackContent` is true or false.

Angular Version: 22.x.x Fundamental NGX Version: 0.61.x (or your current version)

### Affected Component

Platform Icon Tab Bar (fdp-icon-tab-bar and fdp-icon-tab-bar-tab)

1. Create a platform Icon Tab Bar with `[stackContent]="true"`.
2. Add two `<fdp-icon-tab-bar-tab>` child components.
3. Apply `[active]="true"` to the **second** tab component (or dynamically bind it to return true for the second tab).
4. Render the page.
5. Notice that the first tab is displayed as active, ignoring the `[active]` property on the second tab.
6. Open the browser console to see the `NG0956` tracking performance warning.

### StackBlitz example(s)

N/A (Can be easily reproduced by adding [stackContent]="true" to any existing platform icon tab bar documentation example).

### Priority

Medium

### Declaration

droshev:
closes #14501

### DIFF
--- a/libs/platform/icon-tab-bar/icon-tab-bar.component.ts
+++ b/libs/platform/icon-tab-bar/icon-tab-bar.component.ts
@@ -1,5 +1,6 @@
 import { NgTemplateOutlet } from '@angular/common';
 import {
+    afterNextRender,
     booleanAttribute,
     ChangeDetectionStrategy,
     ChangeDetectorRef,
@@ -11,12 +12,14 @@ import {
     ElementRef,
     HostBinding,
     inject,
+    Injector,
     Input,
     input,
     model,
     OnInit,
     Optional,
     output,
+    runInInjectionContext,
     signal,
     TemplateRef,
     viewChild,
@@ -242,6 +245,12 @@ export class IconTabBarComponent implements OnInit, TabList {
     private readonly _rtlService = inject(RtlService, { optional: true });
 
     /** @hidden */
+    private readonly _injector = inject(Injector);
+
+    /** @hidden Maps stable tab key → uId string, allocated once per tab identity */
+    private _tabUidMap = new Map<string, string>();
+
+    /** @hidden */
     constructor(
         private _cd: ChangeDetectorRef,
         @Optional() private _contentDensityService: ContentDensityService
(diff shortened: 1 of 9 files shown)


libs/platform/icon-tab-bar/icon-tab-bar.component.ts
```diff
@@ -266,6 +275,30 @@ export class IconTabBarComponent implements OnInit, TabList {
...
+         * risk (see popover ancestor-observer loop). Track as separate issues if needed.
+         */
+        if (this.stackContent) {
+            runInInjectionContext(this._injector, () => {
+                afterNextRender(() => {
+                    const activeTab = this._flatTabs$().find((t) => t.active);
+                    if (activeTab) {
+                        const panel = this.tabDirectives().find((tab) => tab.uId() === activeTab.uId);
+                        this._scrollToPanel(panel ?? null);
+                    }
```
InnaAtanasova: This violates 
"DO NOT use `afterNextRender()` / `afterRender()` to set properties on child components -- we hit NG0100 when FormItemComponent used `afterNextRender` to set `ariaLabelledBy` on a content child (PR #14045). Moving to `ngAfterContentInit` fixed it. Angular docs recommend `afterNextRender` for DOM operations (measuring layout, third-party library init), not for modifying component state."

The scroll operation queries this.tabDirectives() which depends on content projection being stable. Try to move this logic to `ngAfterViewInit()` or `ngAfterContentInit()` lifecycle hook instead. If the scroll must wait for layout, wrap just the `scrollIntoView()` call in `afterNextRender()`, not the entire tab lookup + panel query.

libs/platform/icon-tab-bar/icon-tab-bar.component.ts
```diff
@@ -266,6 +275,30 @@ export class IconTabBarComponent implements OnInit, TabList {
...
+         *    asynchronously (e.g. deferred content), the active panel may not exist
+         *    yet and the initial scroll will be silently skipped.
+         * 2. A runtime `false → true` transition of `[stackContent]` does not re-run
+         *    this scroll — the active panel is not scrolled into view and the scroll-spy
+         *    may select the first visible panel instead.
+         * Do NOT add a retry-on-mutation observer for either case — self-feeding observer
+         * risk (see popover ancestor-observer loop). Track as separate issues if needed.
+         */
+        if (this.stackContent) {
+            runInInjectionContext(this._injector, () => {
```
InnaAtanasova: Why is `runInInjectionContext` necessary? 
`ngOnInit()` already runs within an injection context. `runInInjectionContext()` is only needed when calling DI-dependent APIs outside Angular's normal execution flow (e.g., in a setTimeout callback, event handler, or third-party library callback).

droshev: Good point. I moved the query lookup to ngAfterViewInit() and kept only the layout-dependent _scrollToPanel() call inside afterNextRender().

libs/platform/icon-tab-bar/icon-tab-bar.component.ts
droshev: Agreed. The callback does not set properties on child components; afterNextRender() is now used only for the layout-dependent scroll operation.

InnaAtanasova (approved):

PR #14506, opened 2026-08-29
fix(core, platform): update multi-combobox tokens and reset selected flags on programmatic selectedItems change (#13553)

### LINKED ISSUE #13553: [BUG] Multi Combobox: View Not Updating Programmatically When Setting Value to Form or Changing Selected Items
### Describe your proposal

When I set the value of the Multi Combobox programmatically—either by updating the form value or by changing the selected items in code—the view does not update to reflect the new selection. The UI only updates when we interacts with the component manually.

Angular 20.1.4 ,Fundamental:0.57.3

Steps to Reproduce 1:

Open the example.
Change the selection in the "Select some fruit" multi-combobox (add or remove items).
In the onSelect handler for "Select some fruit," we programmatically set the selected items for "Select some vegetables" to:
[
  { "name": "Banana", "type": "Fruits" },
  { "name": "Pineapple", "type": "Fruits" `}`
]
However, the "Select some vegetables" control still has its initial value ([]).
Issue:
The selected values for "Select some vegetables" do not update in the UI unless the combobox dropdown is opened.

Steps to Reproduce 2:

Open the example.
Change the selection in the "Select some vegetables" multi-combobox (add or remove items).
In the onSelect handler for "Select some vegetables," we programmatically set the selected items for "Select some vegetables1" to:

[]
However, the "Select some vegetables" control still has its initial value.
Selected Item:

[
  { "name": "Strawberry", "type": "Fruits" },
  { "name": "Broccoli", "type": "Vegetables" }
]
Issue:
The selected values for "Select some vegetables" do not update in the UI unless the combobox dropdown is opened.

### StackBlitz example(s)

https://stackblitz.com/edit/h5aamdvb-qk3nx7fb?file=src%2Fapp%2Fmulti-combobox-forms-example.component.ts,src%2Fapp%2Fmulti-combobox-forms-example.component.html

### Declaration

droshev:
## Summary

- Keep multi-combobox selections synchronized across bound values, rendered tokens, and dropdown checkbox states after programmatic updates and user interaction.
- Synchronize delayed and initial `FormControl` values in Core; synchronize Platform's full and visible option collections using value-based matching, and emit raw selected values to forms.
- Update both Reactive Forms and signal-backed documentation examples with programmatic Select All/Clear actions; use stable tracking for tokens and option lists.
- Closes #13553

## Test plan

- [x] In `/#/core/multi-combobox` and `/#/platform/multi-combobox`, use the Reactive Forms Select All/Clear controls and confirm tokens and dropdown checks match the displayed form value.
- [x] In both Pure Signal-based examples, add an item through the dropdown, then replace, clear, filter, and remove a token; confirm the signal value, tokens, and checked rows remain aligned.

### DIFF
--- a/libs/core/multi-combobox/base-multi-combobox.class.ts
+++ b/libs/core/multi-combobox/base-multi-combobox.class.ts
@@ -730,6 +730,10 @@ export abstract class BaseMultiCombobox<T = any> {
             this._flatSuggestions.set(this.isGroup() ? flattenGroups(currentSuggestions) : currentSuggestions);
             this._fullFlatSuggestions.set(this._flatSuggestions());
 
+            if (!this._selectedItems().length && this._cva.value?.length) {
+                this._setSelectedItems(this._cva.value);
+            }
+
             this._setSelectedSuggestions();
 
             this._mapAndUpdateModel();
--- a/libs/core/multi-combobox/multi-combobox.component.ts
+++ b/libs/core/multi-combobox/multi-combobox.component.ts
@@ -54,7 +54,7 @@ import { PopoverFillMode } from '@fundamental-ngx/core/shared';
 import { contentDensityObserverProviders } from '@fundamental-ngx/core/content-density';
 import { TokenComponent, TokenizerComponent } from '@fundamental-ngx/core/token';
 import { Subject } from 'rxjs';
-import { debounceTime } from 'rxjs/operators';
+import { debounceTime, filter } from 'rxjs/operators';
 import { BaseMultiCombobox } from './base-multi-combobox.class';
 import { MobileMultiComboboxComponent } from './mobile/mobile-multi-combobox.component';
 import { MULTI_COMBOBOX_COMPONENT } from './multi-combobox.token';
@@ -576,6 +576,17 @@ export class MultiComboboxComponent<T = any> extends BaseMultiCombobox<T> implem
         this._assignCustomTemplates();
 
         this._initWindowResize();
+
+        this._cva.stateChanges
+            .pipe(
+                filter((event) => event === 'writeValue'),
+                takeUntilDestroyed(this._destroyRef)
+            )
+            .subscribe(() => {
+                if (this._fullFlatSuggestions().length) {
+                    this._setSelectedItems(this._cva.value);
+                }
+            });
     }
 
     /** @hidden */
@@ -927,6 +938,9 @@ export class MultiComboboxComponent<T = any> extends BaseMultiCombobox<T> implem
      */
     protected _setSelectedItems(value: T[]): void {
         this._selectedItems.set(coerceArraySafe(value));
+        if (this._fullFlatSuggestions().length) {
+            this._setSelectedSuggestions();
+        }
     }
 
     /**
--- a/libs/core/multi-combobox/multi-combobox.component.ts
+++ b/libs/core/multi-combobox/multi-combobox.component.ts
@@ -577,6 +577,9 @@ export class MultiComboboxComponent<T = any> extends BaseMultiCombobox<T> implem
 
         this._initWindowResize();
 
+        // Sync selected items when FormControl.setValue() is called (CVA.writeValue event).
+        // This catches programmatic value updates even without a [selectedItems] binding.
+        // Requires datasource to be populated for the lookup to succeed.
         this._cva.stateChanges
             .pipe(
                 filter((event) => event === 'writeValue'),
--- a/libs/docs/core/multi-combobox/examples/multi-combobox-forms/multi-combobox-forms-example.component.ts
+++ b/libs/docs/core/multi-combobox/examples/multi-combobox-forms/multi-combobox-forms-example.component.ts
@@ -51,12 +51,10 @@ export class MultiComboboxFormsExampleComponent {
     selectAllFruits(): void {
         const fruits = this.dataSource.filter((item) => item.type === 'Fruits');
         this.reactiveFormControl.setValue(fruits);
-        this.customForm.updateValueAndValidity();
     }
 
     clearReactiveForm(): void {
         this.reactiveFormControl.setValue([]);
-        this.customForm.updateValueAndValidity();
     }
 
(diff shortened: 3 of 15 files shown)


libs/core/multi-combobox/base-multi-combobox.class.ts
```diff
@@ -730,6 +730,10 @@ export abstract class BaseMultiCombobox<T = any> {
             this._flatSuggestions.set(this.isGroup() ? flattenGroups(currentSuggestions) : currentSuggestions);
             this._fullFlatSuggestions.set(this._flatSuggestions());
 
+            if (!this._selectedItems().length && this._cva.value?.length) {
+                this._setSelectedItems(this._cva.value);
+            }
+
```
InnaAtanasova, line 736: This will fail to handle the replace scenario (e.g., [Apple] → [Banana]). When _selectedItems() already has items, this condition is false, so programmatic replacements via CVA won't trigger _setSelectedItems(). The fix only covers the initial load case, not runtime updates.

libs/platform/form/multi-combobox/commons/base-multi-combobox.ts
```diff
@@ -631,14 +638,17 @@ export abstract class BaseMultiCombobox extends CollectionBaseInput implements O
     protected _setSelectedSuggestions(): void {
         this._selectedSuggestions = [];
 
+        this._fullFlatSuggestions.forEach((item) => (item.selected = false));
+        (this._suggestions || []).forEach((item) => (item.selected = false));
```
InnaAtanasova, line 646: The fix resets all selected flags to false before re-applying, which is correct. However, the finder function on line 651 uses `shallowEqual(item.value, selectedItem)` which will work for object references, but the old code on the right side of the || still uses string equality (`item.label === selectedItem`). This creates an asymmetry: the new code assumes objects, but the fallback assumes strings. Consider whether the string path is still needed or should also use shallowEqual.

libs/platform/form/multi-combobox/multi-combobox/multi-combobox.component.ts
```diff
@@ -361,7 +361,7 @@ export class MultiComboboxComponent extends BaseMultiCombobox implements OnInit,
     /** @hidden */
     private _propagateChange(emitInMobile?: boolean): void {
         if (!this.mobile || emitInMobile) {
-            this.onChange(this._selectedSuggestions);
+            this.onChange(this._selectedSuggestions.map(({ value }) => value));
```
InnaAtanasova, line 364: This is correct and fixes the wrapper leak bug. However, verify that all callers of `_propagateChange()` expect raw values, not wrappers. If any internal code relies on the old shape, this could be breaking.

libs/core/multi-combobox/multi-combobox.component.ts
```diff
@@ -576,6 +576,17 @@ export class MultiComboboxComponent<T = any> extends BaseMultiCombobox<T> implem
         this._assignCustomTemplates();
 
         this._initWindowResize();
+
+        this._cva.stateChanges
```
InnaAtanasova, line 583: Maybe add a comment to explain why this is needed. Something like "Sync selected items when `FormControl.setValue()` is called after datasource populates."

libs/docs/core/multi-combobox/examples/multi-combobox-forms/multi-combobox-forms-example.component.ts
```diff
@@ -33,13 +34,38 @@ export class MultiComboboxFormsExampleComponent {
...
+
+    onPureSignalChange(item: MultiComboboxSelectionChangeEvent): void {
+        this.pureSignalSelection.set(item.selectedItems);
+    }
+
+    // Programmatic updates via FormControl
+    selectAllFruits(): void {
+        const fruits = this.dataSource.filter((item) => item.type === 'Fruits');
+        this.reactiveFormControl.setValue(fruits);
+        this.customForm.updateValueAndValidity();
```
InnaAtanasova: Why is this line needed? `setValue()` already triggers validation. This line should be redundant unless there's custom async validation. Should be documented why it's needed.

InnaAtanasova (changes requested):

InnaAtanasova:
Since you are making changes in both libs, the PR title should be updated to include `core`.

libs/docs/core/multi-combobox/examples/multi-combobox-forms/multi-combobox-forms-example.component.ts
droshev: removed

libs/core/multi-combobox/multi-combobox.component.ts
droshev: added comments why we need it

libs/platform/form/multi-combobox/multi-combobox/multi-combobox.component.ts
droshev: This is the correct fix and verified no broken callers. _propagateChange() is private, so the only callers are internal selection handlers

libs/platform/form/multi-combobox/commons/base-multi-combobox.ts
droshev: added some comments. The asymmetry is intentional: item.label === selectedItem matches string - based selections (dataSource of primitives), shallowEqual(item.value, selectedItem) matches object-based selections. Since labels are always strings, === is sufficient there; only the value side needs deep equality.

libs/core/multi-combobox/base-multi-combobox.class.ts
droshev: Verified it live: the condition does fire correctly on replace. !changes['selectedItems'].firstChange is true precisely on the second and later changes, which is when replace happens (first change = initial load, skipped intentionally).

InnaAtanasova (approved):

InnaAtanasova (approved):

PR #14531, opened 2026-09-10
fix(platform): platform table cell title performance improvements

### LINKED ISSUE #14503: Table slows down the app when having loaded data
### Describe your proposal

## What we see

## The line

```html
<div
  class="fd-table__cell..."
  [attr.title]="
  "
  #tableTextContainer
>
```

### Affected Component

_No response_

## Our patch, for reference

```diff
-            [attr.title]="
-                tableTextContainer.offsetWidth < tableTextContainer.scrollWidth ? tableTextContainer.textContent : null
-            "
```

### StackBlitz example(s)

_No response_

### Priority

Very High

### Declaration

mikerodonnell89:
fixes #14503


libs/platform/table/components/table-row/table-row.component.ts
```diff
@@ -350,6 +366,11 @@ export class TableRowComponent<T> extends TableRowDirective implements OnInit, A
         }
     }
 
+    /** @hidden Get cached cell title for the given column index */
+    protected _getCellTitle(columnIndex: number): string | null {
```
InnaAtanasova, line 370: protected should not get a prefix.

libs/platform/table/components/table-row/table-row.component.ts
```diff
@@ -269,6 +277,14 @@ export class TableRowComponent<T> extends TableRowDirective implements OnInit, A
...
             }
         });
+
+        // Measure cell overflow after initial render to populate title attributes
+        afterNextRender(
+            () => {
+                this._updateCellTitles();
+            },
+            { injector: this._componentInjector }
+        );
```
InnaAtanasova, line 287: The comment says the logic happens on initial render (which is prob correct as it's afterNextRender), but what would happen when the cell content changes dynamically, for example in sorting, filtering, pagination, cell editing, column resizing?

libs/platform/table/components/table-row/table-row.component.ts
```diff
@@ -417,4 +438,24 @@ export class TableRowComponent<T> extends TableRowDirective implements OnInit, A
...
+            '.fd-table__text, .fd-table__cell--truncate-txt'
+        );
+        const newTitles = new Map<number, string | null>();
+
+        textContainers.forEach((container, index) => {
+            // Check if content overflows
+            if (container.offsetWidth < container.scrollWidth) {
+                newTitles.set(index, container.textContent?.trim() || null);
+            } else {
+                newTitles.set(index, null);
```
InnaAtanasova, line 455: This is something Claude caught:
The `_updateCellTitles()` method queries by CSS class selectors (.fd-table__text, .fd-table__cell--truncate-txt) but uses the resulting node list index as the map key. This breaks when:

1. Selection mode is enabled (adds a checkbox cell before content cells)
2. Semantic highlighting is enabled (adds a status indicator cell)
3. Navigation column is shown (adds a cell at the end)

The DOM node index doesn't match the column index that `_getCellTitle($index)` receives from the template.

libs/platform/table/components/table-row/table-row.component.html
```diff
@@ -155,9 +155,7 @@
             [class.fd-table__text]="column.applyText"
             [class.fd-table__text--no-wrap]="column.noWrap"
             [class.fd-table__cell--truncate-txt]="column.applyText"
-            [attr.title]="
-                tableTextContainer.offsetWidth < tableTextContainer.scrollWidth ? tableTextContainer.textContent : null
-            "
+            [attr.title]="_getCellTitle($index)"
```
InnaAtanasova, line 158: The template calls `_getCellTitle($index)` where `$index` is the column index in `visibleColumns$()`, but the map is keyed by DOM node index from querySelectorAll. These are different coordinate systems.

libs/platform/table/components/table-row/table-row.component.ts
```diff
@@ -269,6 +277,14 @@ export class TableRowComponent<T> extends TableRowDirective implements OnInit, A
                 this._tableRowService.removeEditableCells(this.row);
             }
         });
+
+        // Measure cell overflow after initial render to populate title attributes
+        afterNextRender(
+            () => {
+                this._updateCellTitles();
+            },
+            { injector: this._componentInjector }
```
InnaAtanasova, line 286: I don't think you need the injector inside the constructor.

droshev (changes requested):
Recycled cells after virtual scroll keep whatever tooltip the row that first occupied the slot produced - a data grid showing another row's value in a tooltip. Same staleness for in-place updates (editable-cell commit, consumer row updates, data re-emit) without virtualization.

nothing re-measures when widths change

if you change the data and the text gets truncated there is no tooltip (you can easily create such example)

if you had truncated text then resize to increase the width so the text becomes fully visible, the tooltip still appears

libs/platform/table/components/table-row/table-row.component.ts
mikerodonnell89: Yea, I'll just use the text from the cell itself

libs/platform/table/components/table-row/table-row.component.html
mikerodonnell89: Moving away from using column index

PR #14536, opened 2026-09-13
fix(core,platform): correct combobox leave behavior and popover sync

### LINKED ISSUE #14469: Combobox blur with autocompleted text - incorrect behavior
### Describe your proposal

When I start typing something in the combobox (both core and platform) and the autocomplete fills out a term for me, when I then blur the combobox via tab or outside click, the text in the input is cleared. What should happen is that the combobox selects the item it has autocompleted to. You can observe the correct behavior on the webcomponents combobox, or at openui5. This bug was introduced with https://github.com/SAP/fundamental-ngx/pull/14306, so a fix will need to keep that issue in mind as well

Introduced with https://github.com/SAP/fundamental-ngx/pull/14306

### LINKED ISSUE #14534: fd-combobox: freestyle text input reverts to last confirmed value on blur
### Describe your proposal

**fd-combobox** is designed as a selection control, but it silently reverts any arbitrary text typed by the user back to the last value confirmed from the dropdown list whenever the input
  loses focus. This behavior is caused by the internal _lastConfirmedValue mechanism: _close() is called on outside click/blur, which compares the current inputText against
  displayFn(_lastConfirmedValue) via _isInputDirtyVsModel(), and if they differ, it resets the DOM input to the last confirmed value.

  This makes fd-combobox unusable in any scenario where the user should be allowed to type a custom value not present in the suggestion list (i.e. "select or type" UX pattern).

  Additionally, when used with Angular signal forms via the [formField] directive, a race condition occurs: writeValue() (called by ControlValueAccessor after [formField] syncs the model)
  updates _lastConfirmedValue, which then causes _close() to revert the input to the model value instead of preserving what the user typed.

Fundamental - 0.64.1, Angular - 22.1.0

### Affected Component

 fd-combobox (ComboboxComponent) — @fundamental-ngx/core/combobox

  1. Add fd-combobox with a list of options bound via [formField] or ngModel
  2. Open the dropdown and select any option — value is set correctly
  3. Click into the input and type a custom string not present in the options list
  4. Click outside the input (blur)
  5. Observed: input reverts to the previously selected option
  6. Expected: input retains the custom text the user typed

### StackBlitz example(s)

_No response_

### Priority

_No response_

### Declaration

droshev:
## What changed

Fixes #14469
Fixes #14534

- Corrected Core and Platform combobox behavior when leaving with Tab or pointer.
- Commits autocomplete candidates exactly once and preserves valid custom text.
- Keeps `tabOutStrategy="close"` and `closeOnOutsideClick="false"` behavior intact.
- Synchronizes projected Popover configuration correctly without duplicating `isOpen` synchronization.

> Core object mode intentionally restores the last valid object; freestyle text is supported by Platform combobox.

## Manual test

1. In Platform Combobox Forms, enter unmatched text and leave with Tab or an outside click; verify the custom value is preserved.
2. Accept an autocomplete suggestion and leave; verify the correct form value is emitted exactly once.
3. Reopen the combobox and press Tab; verify it closes without changing or re-emitting the value.
4. With `closeOnOutsideClick="false"`, click outside; verify the popover remains open and the value is unchanged.

## Review focus

Focus on CVA emission count, Tab/blur/pointer event ordering, `closeOnOutsideClick`, and projected Popover configuration synchronization.


libs/platform/form/combobox/combobox/combobox.component.ts
```diff
@@ -159,7 +208,7 @@ export class ComboboxComponent extends BaseCombobox implements ComboboxInterface
     }
 
     /** @hidden Method to set selected item */
-    selectOptionItem(item: OptionItem, shouldClosePopover = true): void {
+    selectOptionItem(item: OptionItem, shouldClosePopover = true, focusInput = true): void {
```
MariaIDineva, line 211: Since that method is inherited from the base class maybe we should update the signature there as well?

MariaIDineva (approved):

libs/platform/form/combobox/combobox/combobox.component.ts
droshev: The new optional parameter is implementation details, while the base class only uses selectOptionItem(item), so its signature can remain unchanged.

InnaAtanasova (approved):

PR #14537, opened 2026-09-14
fix(core): add Home/End keyboard support for menu navigation

### LINKED ISSUE #14532: Core Menu: MenuService does not handle Home/End keys, menu not fully keyboard-navigable (WCAG 2.1.1)
### Describe your proposal

The `MenuService` keyboard handler in `@fundamental-ngx/core` Menu handles Arrow,
Enter, Space and Escape, but does **not** handle the `Home` and `End` keys. Per
WAI-ARIA Authoring Practices for the `menu` pattern, `Home` should move focus to the
first menu item and `End` to the last item. Without this, a menu opened from its
trigger is not fully operable by keyboard, which we assess as a WCAG 2.1.1 (Keyboard)
gap for the menu pattern.

Because the menu content renders inside a CDK overlay, the keydown does not bubble to
the `fd-menu` host, so consumers cannot easily polyfill this on the host element
either, we had to add a document-level capture-phase listener as a workaround.

Expected behavior:
- With focus inside an open `fd-menu`, pressing `Home` moves focus to the first
  enabled menu item.
- Pressing `End` moves focus to the last enabled menu item.
- Subsequent Arrow navigation continues correctly from the new position (i.e. the
  focus change is mirrored into `MenuService` focus state).

Actual behavior:
- `Home`/`End` do nothing; focus stays on the current item.

Suggested fix:
- Handle `Home`/`End` in `MenuService` keyboard handling alongside the existing
  Arrow/Enter/Space/Escape cases, moving focus to the first/last enabled item and
  updating `setFocused(...)` accordingly.

Angular: 21.2.17, Fundamental Library: 0.63.1

### Affected Component

Core Menu (fd-menu / MenuComponent, MenuService)

1. Render an `fd-menu` with several `fd-menu-item`s, opened from a trigger button.
2. Open the menu and move focus into it (e.g. ArrowDown).
3. Press `Home`, observe focus does not jump to the first item.
4. Press `End`, observe focus does not jump to the last item.

### StackBlitz example(s)

https://stackblitz.com/edit/v8rdqnug?file=src%2Fapp%2Fmenu-examples.component.ts

### Priority

Medium

### Declaration

InnaAtanasova:
## Related Issue(s)
closes https://github.com/SAP/fundamental-ngx/issues/14532

## Description
This PR adds missing Home and End keyboard handling for core menu navigation so keyboard users can quickly jump to the first or last enabled item inside an open menu.

### Problem
Menu keyboard support handled Arrow, Enter, Space, and Escape, but did not handle Home and End.
That left menu keyboard behavior incomplete against WAI-ARIA menu authoring practices and created a WCAG 2.1.1 keyboard gap.

### Changes
- Added Home key handling to focus the first enabled item.
- Added End key handling to focus the last enabled item.
- Ensured focus state updates through the existing focus APIs so subsequent Arrow navigation continues from the new position.
- Added unit tests for Home and End behavior.
- Added unit tests verifying disabled items are skipped for boundary navigation.

To test with your Mac keyboard which doesn't have Home and End keys use:
- Home: Fn + Left Arrow
- End: Fn + Right Arrow


MariaIDineva (approved):
Nice that you also added the logic to MenuKeyboardService.

PR #14538, opened 2026-09-14
[WIP] fix(platform): platform table cell title performance improvements

### LINKED ISSUE #14503: Table slows down the app when having loaded data

mikerodonnell89:
fixes #14503


libs/platform/table/directives/table-cell-overflow.directive.ts
```diff
@@ -0,0 +1,65 @@
+import { DestroyRef, Directive, ElementRef, Input, Renderer2, afterNextRender, inject } from '@angular/core';
+
+/**
+ * Directive that sets the title attribute on an element when its text content overflows.
+ * Uses ResizeObserver to detect overflow without forcing layout reflows during change detection.
+ */
+@Directive({
+    selector: '[fdpTableCellOverflow]'
+})
+export class TableCellOverflowDirective {
```
InnaAtanasova, line 10: I think this directive should also be exported in the `index.ts` file

libs/platform/table/directives/table-cell-overflow.directive.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    /** Whether the directive is enabled. */
+    @Input()
+    set fdpTableCellOverflow(value: boolean) {
+        this._enabled = value;
+        if (!this._enabled && this._resizeObserver) {
+            this._resizeObserver.disconnect();
+            this._renderer.removeAttribute(this._elementRef.nativeElement, 'title');
+        }
+    }
```
InnaAtanasova, line 25: Can this be done with input signal and effect?
it can be something like:
```
readonly fdpTableCellOverflow = input(false);

constructor() {
    effect(() => {
        this._enabled = this.fdpTableCellOverflow();
        if (!this._enabled && this._resizeObserver) {
            this._resizeObserver.disconnect();
            this._renderer.removeAttribute(this._elementRef.nativeElement, 'title');
        }
    });
    // ... rest of constructor
}

```

libs/platform/table/components/table-row/table-row.component.html
```diff
@@ -155,9 +155,7 @@
             [class.fd-table__text]="column.applyText"
             [class.fd-table__text--no-wrap]="column.noWrap"
             [class.fd-table__cell--truncate-txt]="column.applyText"
-            [attr.title]="
-                tableTextContainer.offsetWidth < tableTextContainer.scrollWidth ? tableTextContainer.textContent : null
-            "
+            [fdpTableCellOverflow]="column.noWrap"
```
InnaAtanasova, line 158: I can't fully understand this. 
The previous implementation showed tooltip on overflow: `[attr.title]="tableTextContainer.offsetWidth < tableTextContainer.scrollWidth ? tableTextContainer.textContent : null"`
The new directive adds title when `column.noWrap` is true AND content overflows
But what if `column.noWrap` is false? The cell might still overflow with wrapping.
Am I missing something?

libs/platform/table/directives/table-cell-overflow.directive.ts
```diff
@@ -0,0 +1,65 @@
...
+ */
+@Directive({
+    selector: '[fdpTableCellOverflow]'
+})
+export class TableCellOverflowDirective {
+    private readonly _elementRef = inject(ElementRef<HTMLElement>);
+    private readonly _renderer = inject(Renderer2);
+    private readonly _destroyRef = inject(DestroyRef);
+    private _resizeObserver?: ResizeObserver;
+    private _enabled = false;
```
InnaAtanasova, line 15: These need ` /** @hidden */`

droshev:
I wrote my comments in team

PR #14539, opened 2026-09-16
fix(core): prevent avatar-group width cache corruption when hidden via display:none

MariaIDineva:
## Related Issue(s)

Part of #12369 (bug b).

## Description

When an `fd-avatar-group` is hidden by a CSS `display:none` ancestor (for example, an inactive
tab that keeps DOM content alive instead of using `@if`), the `ResizeObserver` on the host fires
with `contentRect.width = 0`. At that point `AvatarGroupItemRendererDirective.width` runs for each
portal-attached item: `getBoundingClientRect()` returns `0` for elements inside a hidden subtree,
but `!isNaN(0)` is `true`, so the guard passes and `_lastSavedWidth` is overwritten with `~0`.

When the container is shown again, any item that was moved to `hiddenItems` during the hidden
period has `visible = false` (portal detached). The `width` getter skips `getBoundingClientRect`
for detached items and returns the stale near-zero cache. `_calculateVisibility` then believes all
items fit inside the real container, collapses the overflow count to zero, and shows every avatar
as visible — even though they no longer fit.

## Fix

Change the `_lastSavedWidth` update guard from `!isNaN(width)` to `width > 0`. A zero width from
a hidden element is indistinguishable from a missing measurement, so it must not overwrite a
previously valid cached value.

## How to reproduce (before fix)

1. Open the **Avatar Group → Group Type** docs example.
2. Note the overflow count — some avatars should be hidden.
3. Click **Hide** to apply `display:none` to the container.
4. Click **Show**.
5. The overflow count drops to zero (wrong).

After the fix, step 5 shows the correct overflow count.

## Screenshots

Before:

https://github.com/user-attachments/assets/9199d1dd-34b1-4074-ae93-88ac46f59e6a

After:

https://github.com/user-attachments/assets/8be21cc6-d55e-46ba-bbd8-8dfa3e5a1c09


InnaAtanasova (dismissed):

InnaAtanasova:
Looks great 🚀

droshev (approved):

InnaAtanasova (approved):

Issue #14504, opened 2026-08-28
Table a11y improvements


InnaAtanasova:
This concerns Core and Platform Tables.
____________________________________________________________________

Role="table" (Standard Data Table)

A table is for displaying and reading data. Screen readers navigate by rows:
User experience:
* Arrow keys move between rows (↓ goes to next row)
* User hears the entire row's content at once
* It's a read-only presentation of data
* Think: spreadsheet you can only view
Example use case: A list of employees with their names, departments, and salaries. Users just need to read the information.
____________________________________________________________________

Role="grid" (Interactive Data Grid)
A grid is for interacting with individual cells. It works like Excel:

1. Cell-by-cell arrow key navigation
2. Editable cells
3. Multi-cell selection

____________________________________________________________________

In our case, all the tables we have are grid tables (we have cell by cell navigation).

What we currently have as markup:
```
<table role="grid">
    <tr><th>...</th></tr>
</table>
```

Screen readers encounter conflicting semantics:

- The native <table> element has implicit role="table"
- But we are overriding with role="grid"
- Child elements like `<th>`, `<tr>`, `<td>` still announce as table semantics

This causes JAWS to announce both "table" AND "grid".

For role="grid", we must use divs with explicit ARIA roles throughout:

-  role="grid" on container
- role="row" for each row
- role="columnheader" for header cells
- role="gridcell" for data cells
- role="rowgroup" for thead/tbody equivalents (optional but recommended)

[...]

PR #14542, opened 2026-09-16
fix(core, platform): refactor Core Table 

### LINKED ISSUE #14434: fix(core): remove NgZone from the core Table for zoneless compatibility
## Problem

`libs/core/table/table.component.ts` still depends on `NgZone`. Under zoneless change detection (the Angular default since v20; this workspace is on `@angular/core` 22.0.5), one of those uses is a silent correctness break and the rest are dead no-ops. `NgZone.onStable` **never emits** in a zoneless app, so code waiting on it never runs.

Relates to #14035 (closed COMPLETED — the table was left unmigrated). Blocks #14038 (zoneless bootstrap).

## What breaks (scenario)

**Cell focusability is never set** — keyboard/focus navigation across table cells breaks.

`_updateCells()` (line 141) defers cell setup to `onStable`:

```ts
private _updateCells(): void {
  this._ngZone.onStable.pipe(first()).subscribe(() => {
    this._cells?.forEach((cell) => (cell.focusable = cell.focusable || this.allCellsFocusable));
  });
}
```

Called from `ngAfterViewInit` (:92) and on cell-query changes (:119). Zoneless → `onStable` never fires → `cell.focusable` stays unset → arrow-key/Tab navigation across cells does not work. No build error, no lint warning, invisible in jsdom.

## What needs to be fixed

1. **`table.component.ts:141`:** replace `onStable.pipe(first())` with `afterNextRender(() => …)` (one-shot post-render, the exact semantics being approximated). If it must re-run on cell-query changes, use `afterRenderEffect`. Use `DestroyRef` for teardown.
2. **Cleanup:** remove any `runOutsideAngular(...)` wrappers in core table (dead ceremony under zoneless — "outside the zone" is already the default), then delete the now-unused `NgZone` import/injection from `libs/core/table/table.component.ts`.

## How to validate

- [ ] `grep -n NgZone libs/core/table/table.component.ts` returns nothing.
- [ ] In a **zoneless** docs build: open a core table, Tab into it, and confirm arrow keys move focus across all cells. This reproduction is **browser-only** — jsdom cannot run the focus engine or `onStable` timing.
- [ ] No regression in existing core-table keyboard-nav interactions.

## Testing note

The reproduction runs through `onStable` + the browser focus engine — both are browser-only. A jsdom TestBed spec asserting "cell.focusable is set" passes regardless (false green). The RED must be an E2E / running-zoneless check; add a lock-in unit test only where it can genuinely fail.

InnaAtanasova:
## Related Issue(s)

closes of https://github.com/SAP/fundamental-ngx/issues/14504
closes https://github.com/SAP/fundamental-ngx/issues/14434

## Description
1. refactor all Table directives and components to signals
2. get rid of Zone js 
3. Add appropriate roles to table and its children to comply with grid table and the pattern in https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/grid_role
4. Update all tests for Table

Our tables in Core and Platform are grid tables (Cell-by-cell arrow key navigation)
For role="grid", we now use divs with explicit ARIA roles throughout:

-  role="grid" on container
- role="row" for each row
- role="columnheader" for header cells
- role="gridcell" for data cells
- role="rowgroup" for thead/tbody equivalents (optional but recommended)

```
<table role="grid" aria-labelledby="...">
  <thead role="rowgroup">
    <tr role="row">
      <td></td>
      <th role="columnheader" aria-label="">...</th>
      <th role="columnheader" aria-label="">...</th>
      ....
    </tr>
  </thead>
  <tbody role="rowgroup">
    <tr role="row">
      <th scope="row" role="rowheader">...</th>
      <td>...</td>
      <td>...</td>
      <td role="gridcell" tabindex="-1">1</td>
    </tr>
    <tr role="row">
      <th scope="row" role="rowheader">...</th>
      <td role="gridcell">...</td>
      <td role="gridcell">...</td>
    </tr>
  </tbody>
</table>
```

## Screenshots
Screenshot 2026-09-16 at 4 09 36 PM


libs/core/table/directives/table-icon.directive.ts
```diff
@@ -1,65 +1,36 @@
...
-    }
-
     /** @hidden */
-    ngOnInit(): void {
-        this.buildComponentCssClass();
-    }
+    protected readonly cssClass = computed(() =>
+        [
+            'fd-table__icon',
+            this.glyph ? fdBuildIconClass(this.glyphFont(), this.glyph()) : '',
```
MariaIDineva: That should be `this.glyph()` as it is now a signal.

libs/core/table/directives/table-cell.directive.ts
```diff
@@ -94,32 +118,36 @@ export class TableCellDirective extends FocusableItemDirective implements AfterC
     }
 
     /** @hidden */
-    @HostListener('focusin')
+    ngOnInit(): void {
+        // Sync focusable input with parent's focusable state
+        this.setFocusable(this.focusable());
```
MariaIDineva: One comment from Claude which I think is worth addressing:

**[table-cell.directive.ts:117-120] focusable input not reactively wired to setFocusable()**

The original @Input() set focusable(value) { this.setFocusable(value); } updated the FocusableItemDirective internal state on every change. The new code only calls this.setFocusable(this.focusable()) once in ngOnInit. After that, changes to the focusable input (e.g. when TableComponent sets allCellsFocusable = true and later false) won't propagate to isFocusable(), which is what drives the fd-table__cell--focusable class binding and the focusability behavior. Note also that TableComponent's new effect only calls cell.setFocusable(true) when allFocusable is true but never resets to false, which also differs from the original cell.focusable = cell.focusable || allCellsFocusable.

Fix: add an effect in TableCellDirective's constructor:
```
constructor() {
    super();
    effect(() => this.setFocusable(this.focusable()));
}
```

And fix TableComponent's effect to also handle the false case:
```
cells.forEach((cell) => cell.setFocusable(allFocusable || cell.focusable()));
```

libs/core/table/directives/table-container.directive.ts
```diff
@@ -59,15 +34,51 @@ export class TableContainerDirective implements OnDestroy {
...
+    private readonly _destroyRef = inject(DestroyRef);
 
     /** @hidden */
-    private _resizeSubscription = new Subscription();
+    constructor() {
+        // Handle toolbar sticky positioning when outerScroll is enabled
+        toObservable(this.toolbar)
+            .pipe(
+                switchMap((toolbarElRef) => {
+                    const outerScroll = this.outerScroll();
```
MariaIDineva: Another one from Claude:

**[table-container.directive.ts:679] outerScroll not reactive**

outerScroll() is read inside a switchMap callback triggered only when toolbar emits. If a consumer changes outerScroll dynamically after the toolbar is already in the DOM, the sticky logic won't update. Consider:

```
toObservable(this.toolbar).pipe(
    combineLatestWith(toObservable(this.outerScroll)),
    switchMap(([toolbarElRef, outerScroll]) => { ... })
)
```
(The original @ContentChild setter had the same limitation, so this is not a regression, but the signal migration is a good opportunity to fix it.)

libs/core/table/directives/table-status-indicator.directive.ts
```diff
@@ -1,22 +1,22 @@
...
     selector: '[fdTableStatusIndicator], [fd-table-status-indicator]',
     host: {
-        class: 'fd-table__cell--status-indicator',
-        '[class.fd-table__cell--status-indicator--valid]': 'status === "valid"',
-        '[class.fd-table__cell--status-indicator--warning]': 'status === "warning"',
-        '[class.fd-table__cell--status-indicator--information]': 'status === "information"',
-        '[class.fd-table__cell--status-indicator--error]': 'status === "error"'
+        '[class]': 'cssClass()'
     },
     standalone: true
```
MariaIDineva: `standalone: true` not needed any more.

libs/core/table/directives/table-cell.directive.spec.ts
```diff
@@ -1,74 +1,177 @@
...
+        });
+
+        it('should set role="rowheader" and scope="row" for th in tbody', () => {
+            const tbodyHeader = fixture.nativeElement.querySelector('tbody th');
+            expect(tbodyHeader.getAttribute('role')).toBe('rowheader');
+            expect(tbodyHeader.getAttribute('scope')).toBe('row');
+        });
     });
 
-    it('should handle focus events', () => {
```
MariaIDineva, line 63: Maybe we should keep some unit test for focus events.

libs/core/table/directives/table-row.directive.spec.ts
```diff
@@ -1,120 +1,130 @@
...
-
-        component.keys.pop();
-
-        keys = component.keys;
-
-        (<any>component.tableRow)._resetCells(keys);
-
-        fixture.detectChanges();
-
-        expect(getVisibleCells()).toEqual(keys);
```
MariaIDineva, line 118: Maybe we should keep the `should sort elements` and  `should hide elements` tests as well as this functionality will be left untested otherwise?

MariaIDineva (commented):
I don't remember what we finally decided but when merging the PR I think we should mark the migration to signals as a breaking change for potential consumers that access the migrated properties programmatically.

libs/core/table/directives/table-cell.directive.ts
InnaAtanasova: That's how I initially implemented it. After I opened the PR I ran the review-pr skill and Clause said it's against our guidelines to have this functionality with an effect in the constructor. I will see what I can do :D

droshev (approved):

MariaIDineva (approved):

mikerodonnell89 (commented):
I believe some of the roles you're explicitly setting ("row" on `tr`, "columnheader" on any `th` inside of a `thead`), are implicitly set, is there a reason to set it explicitly

PR #14544, opened 2026-09-16
fix(core): emit datetime picker dateChange events

### LINKED ISSUE #14258: bug(core): fd-datetime-picker `dateChange` output declared but never emitted
## Summary

`DatetimePickerComponent` declares `dateChange` as a public `@Output` but **never emits it from anywhere in the component**. Consumers binding `(dateChange)="…"` on `<fd-datetime-picker>` are silently never called.

## Reproduction

```html
<fd-datetime-picker
    [(ngModel)]="value"
    (dateChange)="onDateChange($event)">
</fd-datetime-picker>
```

```ts
onDateChange(date: FdDate): void {
    console.log('dateChange fired', date); // never runs
}
```

Pick any date / time using the picker UI, or type a valid value and submit. `(dateChange)` is never emitted. `[(ngModel)]` and `selectedDateChange` (on the underlying calendar) work as expected — only the documented `dateChange` output on the picker itself is dead.

## Evidence

`libs/core/datetime-picker/datetime-picker.component.ts:357`:

```ts
readonly dateChange: EventEmitter<D> = new EventEmitter<D>();
```

Output is declared and exported as part of the component's public surface.

```bash
$ rg 'dateChange\.emit|this\.dateChange' libs/core/datetime-picker/datetime-picker.component.ts
$ # zero matches
```

No `.emit()` call anywhere in the file. The handler at line 692 (`handleDateChange`) writes to a private `_tempDate` and submits, but does not emit `dateChange`:

```ts
handleDateChange(date: D): void {
    this._tempDate = date;
    if (!this.showFooter) {
        this.submit();
    }
}
```

`submit()` does not emit `dateChange` either. The same component does emit other outputs correctly (`activeViewChange`, `selectedDateChange`, etc.), so the contract pattern is established — `dateChange` is just not wired up.

## Expected behaviour

`dateChange` should fire whenever the picker's bound date value changes — at minimum on:

- a successful date selection from the calendar popover,
- a successful manual entry submitted via Enter / OK,
- a programmatic `writeValue` from `ngModel`.

The semantics should mirror what consumers reasonably expect by analogy with `MatDatepicker`'s `dateChange`. (Note: `fd-date-picker` itself does not declare a `dateChange` output — it relies on `[(ngModel)]` for value notification. So the question for maintainers is whether to **wire up `dateChange` properly** or **deprecate and remove it** in favour of `ngModel`. Removing would be a breaking change but the API is currently broken regardless.)

## Suggested fix shape

Emit from `submit()` (after the model is updated), and in `writeValue` when the new value differs from the previous one. Something along the lines of:

```ts
submit(): void {
    // …existing logic that resolves currentDate/currentTime…

    const previous = this.date;
    this.date = combined;
    if (!this._dateTimeAdapter.datesEqual(previous, combined)) {
        this.dateChange.emit(combined);
    }
    // …existing logic that closes the popover, clears temp state…
}
```

A unit test asserting that `(dateChange)` fires after a calendar selection and after manual-input submit should land alongside the fix (RED-then-GREEN). The fix should also confirm there is no double-emission on `[(ngModel)]` round-trips.

## Scope

- This is **not** introduced by PR #14016 (datetime adapter migration). It is pre-existing on `main` — confirmed by `git grep` against `main`'s tip showing zero `dateChange.emit` sites.
- Fix is independent of the adapter migration and should land as its own PR with its own test.

## Open question for maintainers

Two possible resolutions:

1. **Wire up `dateChange`** (additive, non-breaking). Emit from `submit()` and on programmatic value changes.
2. **Deprecate and remove `dateChange`** (breaking). Direct consumers to `[(ngModel)]` / `valueChanges` on the underlying control.

Either is defensible; pick before assigning the fix.

## References

- Original review observation: https://github.com/SAP/fundamental-ngx/pull/14016#pullrequestreview-4382858274
- Declaration site: `libs/core/datetime-picker/datetime-picker.component.ts:357`

droshev:
## Description

  Fixes #14258.

  - Emits `dateChange` for accepted user changes.
  - Emits `null` when an allowed value is cleared.
  - Prevents duplicate and no-op emissions.
  - Keeps programmatic writes, invalid input, and cancellation silent.

  ## Breaking change

  `dateChange` now emits `D | null`. Event handlers must support `null`.

  ## How to test

  Open **Core → Datetime Picker** in the docs app.

  1. Inspect a `<fd-datetime-picker>` element and run:
     ```js
     const element = document.querySelector('fd-datetime-example fd-
     datetime-picker');
     const picker = ng.getComponent(element);
     picker.dateChange.subscribe(value => console.log('dateChange:',
     value));

  2. Change the date/time and click OK - one event should appear.
  3. Clear the Simple picker - one null event should appear.
  4. Click Change in Programmatic Change - no event should appear.
  5. In Update Value on Blur, type a value, press Enter, then blur - only one event should appear.

---
for later: 
BREAKING CHANGE: DatetimePickerComponent.dateChange now emits D | null. Event handlers must accept null when allowNull is enabled.


libs/core/datetime-picker/datetime-picker.component.spec.ts
```diff
@@ -7,6 +8,15 @@ import { runValueAccessorTests } from 'ngx-cva-test-suite';
 import { DatetimePickerComponent } from './datetime-picker.component';
 import { DatetimePickerModule } from './datetime-picker.module';
 
+@Component({
+    template: '<fd-datetime-picker [date]="date" (dateChange)="dateChanges.push($event)"></fd-datetime-picker>',
+    imports: [DatetimePickerModule]
```
MariaIDineva, line 13: Minor: We can use DatetimePickerComponent instead of the module.

MariaIDineva (approved):

InnaAtanasova (approved):

PR #14546, opened 2026-09-17
feat(core): avatar group feature requests

### LINKED ISSUE #12369: [BUG] fd-avatar-group focus, overflow bugs

MariaIDineva:
## Related Issue(s)

closes https://github.com/SAP/fundamental-ngx/issues/12369

## Description

The following feature requests from the issue are addressed in this PR:

### Feature d) Open group overflow popover from the center of the avatar group

The group-type `fd-popover` had no explicit `placement` binding, so it defaulted to `'bottom-start'`
(anchored to the left edge of the avatar strip). The `Placement` type in `@fundamental-ngx/core/shared`
already maps `'bottom'` to `{ originX: 'center', originY: 'bottom', overlayX: 'center', overlayY: 'top' }`
via CDK overlay — exactly the centered-arrow layout shown in the request.

**Change:** a new `popoverPlacement` signal input on `AvatarGroupComponent` (only applies to `type="group"`),
defaulting to `'bottom'` so the popover opens centered out of the box. Consumers who need the previous
left-aligned behaviour can pass `[popoverPlacement]="'bottom-start'"`.

A docs example was added showing `'bottom'` (default) alongside `'bottom-end'` (right-aligned) to illustrate
the range of options.

Screenshot 2026-09-17 at 11 07 25

### Feature e) Circle-shaped overflow button

The `+N` overflow button uses a square avatar shape by default. The internal `AvatarGroupOverflowButtonComponent`
already had a `circle` input and the `fd-avatar--circle` CSS binding — it just wasn't wired through the public API.

**Change:** added `readonly overflowButtonShape = input<'circle' | 'square'>('square')` to `AvatarGroupComponent`.
The template binds `[circle]="overflowButtonShape() === 'circle'"` on both overflow button usages (individual type
and group type). Passing `overflowButtonShape="circle"` makes the overflow button match the circular avatars.

A docs example was added showing the default square variant alongside the circle variant.

Unit tests were added covering both individual and group type: default shape is square, `'circle'` adds the
`fd-avatar--circle` CSS class, and the class updates reactively when the input changes.

Screenshot 2026-09-17 at 11 07 48

### Feature f) Cap visible avatars at a fixed count

The default visibility algorithm fills the container width greedily — there was no way to say "show at most 5 avatars
regardless of how wide the container is."

**Change:** added `readonly maxVisibleItems = input<number | null>(null)` to `AvatarGroupComponent` and a corresponding
`@Input() maxVisibleItems: number | null = null` to `AvatarGroupHostComponent`. `_calculateVisibility` now delegates to
one of two private methods depending on whether `maxVisibleItems` is set:

- **`_calculateVisibilityByWidth`** — the original algorithm, extracted verbatim and used when `maxVisibleItems` is
  `null`.
- **`_calculateVisibilityWithMaxItems`** — new method that respects both the count cap and the container width:
    1. Separates items into `forcedVisible` (`forceVisibility = true`) and `regular`.
    2. Computes `maxRegularSlots = max(0, maxVisibleItems - forcedVisible.length)`.
    3. Iterates up to `min(regular.length, maxRegularSlots)`, accumulating widths; breaks as soon as an item would
       exceed `containerWidth`. Items beyond that point (whether cut by count or by width) become hidden.
    4. Applies the same "make room for overflow button" step as the width-based path: if any items are hidden and at
       least one regular item is visible, the last visible regular item is moved to the front of the hidden list.

**Behaviour:** wide container → shows exactly up to N avatars (count is the limiting factor). Narrow container →
width constraint kicks in first within the cap; items move to overflow as the container shrinks. `forceVisibility`
items are always shown and reduce the number of regular slots available.

A docs example was added showing the default width-based behaviour alongside a fixed-count variant
(`[maxVisibleItems]="5"` with `overflowButtonShape="circle"`). Both examples include an `fd-slider` that
controls the container width interactively, so readers can observe both the count-cap and the
container-shrink behaviours without leaving the docs page.

Unit tests cover: binding propagation, null fall-through to the width-based path, count-based hiding, equal-to-total
(no overflow), max=0 (all hidden), `forceVisibility` accounting, reactive updates when the input changes, group type,
and width-constraint scenarios within the cap (container shrinks, overflow-button reservation, forced-visible width
accounting). Tests call `_calculateVisibility` directly via type cast because `ResizeObserver` never fires in JSDOM,
making the `combineLatest` pipeline untriggerable in tests.

Screenshot 2026-09-17 at 10 32 22


InnaAtanasova:
This is a very nice enhancement for the component! The code also looks great!
While testing the examples I found some issues:
1. For the **Maximum Visible Avatars**, the popover is not visible, but also it focusses the wrong trigger.

https://github.com/user-attachments/assets/2134d952-631a-4fe8-a4f3-d9edde9fddbf

2. For the same example, the overflow does not respect the window resize
Screenshot 2026-09-17 at 2 04 11 PM

3. I tried adding `orientation="vertical"` to make the group vertical and it didn't work for `maxVisibleItems` case
4. For the **Group Popover Placement** example, when I add `orientation="vertical"`, the overflow functionality is broken:
Screenshot 2026-09-17 at 2 04 33 PM

InnaAtanasova:
I was able to reproduce this issue in main as well. If you can fix it in this PR would be great. If it would take too much I can open a separate issue. Just noticed it while testing the new changes.

https://github.com/user-attachments/assets/dd22fa4a-f64a-4cd3-a344-2532ade1fd10

MariaIDineva:
- Regarding **point 1**, I updated the docs by adding details templates. Since they were not present in the examples, there was nothing to display in the popover. The same problem was present in the `"Circular Overflow Button"` example is now fixed there as well. The change is in the docs only.

- Regarding **point 2**, I fixed it by making the container where the avatar groups are places responsive. It's width was fixed before that. The change is in the docs only.

- As for **points 3 and 4**, it turned out that we don't really have orientation `vertical` support when it comes to avatar groups of type `"group"`. It was a problem in code that was not introduced in this PR and is now fixed, but from what I see in the designs, we have to support overflow in vertical mode as well, which we don't and which we explicitly declare in our docs:

Screenshot 2026-09-18 at 15 43 42

Maybe we should consider implementing it as a separate task.

As part of addressing 3 and 4 I updated the `Vertical Avatar Group Usage` section in the docs by including a vertical example of group avatars of type `group`.

- The last problem that you found, the one with the focus, is also fixed. We just had to make `fd-popover` size itself to its content (match the avatar width) instead of stretching the whole width for vertical avatar groups.

**Edit:** Still working on the focus bug from point 1.

MariaIDineva:
Like we discussed, here are the two separate issues that I created:
- [Wrong focus when selecting an avatar in an avatar group](https://github.com/SAP/fundamental-ngx/issues/14549)
- [Add support for overflow in vertical orientation avatar group](https://github.com/SAP/fundamental-ngx/issues/14550)

libs/core/avatar-group/components/avatar-group-host.component.ts
```diff
@@ -72,13 +66,26 @@ export class AvatarGroupHostComponent
     @Input()
     items: QueryList<AvatarGroupItemDirective>;
 
+    /**
+     * Maximum number of avatars to show before the overflow button.
+     * When set, overrides the width-based visibility calculation.
+     **/
+    @Input()
+    maxVisibleItems: number | null = null;
```
InnaAtanasova: maybe a signal input?

InnaAtanasova (dismissed):

droshev (approved):

InnaAtanasova (approved):

PR #14548, opened 2026-09-17
fix(platform): platform table cell title performance improvements

### LINKED ISSUE #14503: Table slows down the app when having loaded data

mikerodonnell89:
fixes #14503 

Opening a new PR as this is a different, simpler and cleaner approach than the previous PR. Title is updated whenever a resize is done on a column or whenever the MutationObserver detects changes to a cell's text content (which will usually happen via virtual scroll, sorting, filtering, etc)

This PR also modifies the `[noWrap]` example on the "Basic Examples" docs to feature sorting and filtering, so you can test if the titles update after modifying the table contents


libs/platform/table/directives/table-cell-overflow.directive.ts
```diff
@@ -0,0 +1,79 @@
...
+    /** Whether the directive is enabled. */
+    @Input()
+    set fdpTableCellOverflow(value: boolean) {
+        this._enabled = value;
+        if (!this._enabled) {
+            this._resizeObserver?.disconnect();
+            this._mutationObserver?.disconnect();
+            this._renderer.removeAttribute(this._elementRef.nativeElement, 'title');
+        }
+    }
```
InnaAtanasova: what's the reasoning behind @Input? Can this be achieved with a signal and effect?

InnaAtanasova (commented):
LGTM. I tested and saw the performance improvement.

libs/platform/table/directives/table-cell-overflow.directive.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+    });
+
+    it('should not set title attribute when text does not overflow', () => {
+        component.text = 'Short';
+        component.width = 200;
+        fixture.detectChanges();
+
+        // Wait for ResizeObserver callback
+        setTimeout(() => {
+            expect(divElement.hasAttribute('title')).toBe(false);
```
MariaIDineva: We should call `done()` here.

libs/docs/platform/table/examples/platform-table-wrap-example.component.ts
```diff
@@ -84,17 +110,54 @@ export class TableDataProviderExample extends TableDataProvider<ExampleItem> {
         }
 
         return items.filter((item) => {
-            const valuesForSearch = keysToSearchBy.map((key) => getNestedValue(key, item));
+            const valuesForSearch = keysToSearchBy.map((key) => item[key as keyof ExampleItem]);
```
MariaIDineva: Something is wrong with the search here. After this change searching by price is now broken.

libs/platform/table/directives/table-cell-overflow.directive.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+
+        setTimeout(() => {
+            // Change the text
+            component.text = 'Different long text that also overflows the container width completely';
+            fixture.detectChanges();
+
+            // Wait for MutationObserver callback
+            setTimeout(() => {
+                const hasTitle = divElement.hasAttribute('title');
+                if (hasTitle) {
```
MariaIDineva: I removed a couple of `if` checks (that I find weird to have in a unit test) and the assertions failed. This means that those assertions within if blocks that we have in this file are silently skipped, i.e. those unit test do not really serve us.

libs/docs/platform/table/examples/platform-table-wrap-example.component.ts
mikerodonnell89: I'll remove searching from this example as I don't think it adds any value and simply further clutters the docs.

InnaAtanasova (approved):

libs/platform/table/directives/table-cell-overflow.directive.spec.ts
```diff
@@ -14,20 +14,27 @@ class TestComponent {
...
+ * - The directive sets up ResizeObserver and MutationObserver
+ * - The directive responds to the `enabled` input
+ * - The directive calls setAttribute/removeAttribute on the element
+ *
+ * What requires E2E testing in a real browser:
+ * - Actual overflow detection (offsetWidth < scrollWidth)
+ * - Observer callbacks triggering on real layout changes
+ * - Title attribute appearing/disappearing based on actual overflow
+ *
+ * Your coworker was correct that the original `if` guards made tests pass silently
```
MariaIDineva, line 34: Maybe we don't need this part of the comment stating "Your coworker..."

MariaIDineva (approved):

droshev (approved):

PR #14558, opened 2026-09-21
fix(core): synchronize multi-combobox token and option selection

### LINKED ISSUE #13308: core multi-combobox: deleting tokens in the input does not update dropdown selections
When deleting a token from the input tokenizer, the dropdown selection should also update. See video for current behavior:

https://github.com/user-attachments/assets/0f038953-94b2-4ba4-89a4-f86d699df240

droshev:
## Related Issue(s)

closes #13308

## Description

Synchronizes MultiCombobox tokens and dropdown option states when selections change.

- Updates all suggestion collections when a token is removed.
- Keeps the open dropdown checkbox state synchronized.
- Restores staged mobile selections on Cancel without updating the form.
- Commits mobile selections once on Approve.
- Prevents duplicate `selectionChange` emissions.
- Supports primitive, object, grouped, filtered, and range selections.


libs/core/multi-combobox/base-multi-combobox.class.ts
```diff
@@ -434,37 +434,26 @@ export abstract class BaseMultiCombobox<T = any> {
...
-        }
-
-        const fullFlatSuggestions = this._fullFlatSuggestions();
-        const selectedSuggestions: SelectableOptionItem<T>[] = [];
-
-        // Update immutably
-        const updatedSuggestions = fullFlatSuggestions.map((item) => {
-            const isSelected = selectedItems.some(
-                (selectedItem) => item.label === selectedItem || item.value === selectedItem
+    protected _setSelectedSuggestions(selectedItems = this._selectedItems()): void {
```
InnaAtanasova, line 437: Signal mutation creates stale reference issue: The new `_setSelectedSuggestions` method now accepts an optional `selectedItems` parameter (defaults to `this._selectedItems()`), but the implementation has a critical bug with signal tracking. When `_setSelectedSuggestions()` is called with an explicit `selectedItems` array parameter (lines in `_toggleSelection` and `_dialogDismiss`), it bypasses reading `this._selectedItems()` signal. 

This means:

1. In `_toggleSelection` `this._setSelectedSuggestions(updatedSelectedSuggestions.map(({ value }) => value))` passes an explicit array, so `this._selectedItems()` is never read
2. The subsequent `this._propagateChange()` call writes to the model via `_mapAndUpdateModel()`, which reads `this._selectedSuggestions()` and calls `this.writeValue(selectedItems)`
3. `writeValue()` updates `this._selectedItems` signal internally
4. But `_setSelectedSuggestions` already completed using the OLD selectedItems array, creating a desynchronization.

The flow creates a race condition: `_setSelectedSuggestions` uses stale data (the parameter passed in), then `_propagateChange` → `writeValue` updates `_selectedItems` signal, but the suggestion lists were already updated based on the stale data.

Fix: Change the signature to make it explicit when we're restoring a backup vs. synchronizing current state:
```
protected _setSelectedSuggestions(selectedItems?: T[]): void {
    const items = selectedItems ?? this._selectedItems();
    // ... rest of implementation
}
```

And ensure `_toggleSelection` updates `_selectedSuggestions `first, then calls `_setSelectedSuggestions()` without parameters to sync all lists from the updated signal.

libs/core/multi-combobox/multi-combobox.component.ts
```diff
@@ -652,7 +650,7 @@ export class MultiComboboxComponent<T = any> extends BaseMultiCombobox<T> implem
         if (event) {
             event.preventDefault();
         }
-        const optionItem = this._flatSuggestions().find((s) => s.value === token.value);
+        const optionItem = this._flatSuggestions()[getTokenIndexByIdlOrValue(token, this._flatSuggestions())];
```
InnaAtanasova, line 653: If `getTokenIndexByIdlOrValue` returns `-1` (token not found), this accesses `array[-1]` which returns `undefined`. The old code used `.find()` which explicitly returns `undefined` when not found. The subsequent if (optionItem) check handles `undefined`, but accessing `array[-1]` is unclear intent.

Fix: Store the index first and add an explicit check:
```
const idx = getTokenIndexByIdlOrValue(token, this._flatSuggestions());
if (idx === -1) return;
const optionItem = this._flatSuggestions()[idx];
if (optionItem) {
    this._toggleSelection(optionItem, true);
    // ...
}
```

libs/core/multi-combobox/base-multi-combobox.class.ts
droshev: I am not sure how you tested the signal-tracking issue.
_setSelectedSuggestions is a synchronous method(neither an effect, nor computed). That's why whether _selectedItems is read as a default param won't affect signal tracking.

In _toggleSelection() the flow is:
1. read the current _selectedSuggestions
2. create updatedSelectedSuggestions
3. pass those values to _setSelectedSuggestions
4. _setSelectedSuggestions updates right away _selectedSuggestions, _suggestions, _flatSuggestions, and _fullFlatSuggestions
5. _propagateChange calls _mapAndUpdateModel(which reads the updated _selectedSuggestions)
6. writeValue received those same values and synchronizes the signal again.

libs/core/multi-combobox/multi-combobox.component.ts
droshev: Good point. It was safe, but unclear. I added an explicit check for -1. All MultiCombobox tests pass.

InnaAtanasova (approved):

